PR #1, opened 2022-02-15
Address typographical errors

englishm-ietf:
Just a quick pass to fix some typos I noticed while reading.


kixelated:
Oh I just noticed this, thanks!

Issue #12, opened 2022-10-20
TRACK message


kixelated:
I think there should be a TRACK message containing optional metadata about a track. I don't think this should be required for decoding (ex. keep the SPS/PPS in a layer), although it's probably required for some capabilities exchange. Each track should have a unique identifier for sure.

suhasHere:
Having a track message that captures the details of the media being sent will be helpful. It does need to tie to media capabilities or manifest. +1 on having unique track identifier

vasilvv:
I think we should require the sender to declare upfront what tracks are being sent, with ability to declare more later.  For instance, in case of ingestion, I would like to be able to immediately reject unsupported configuration with an error.

As for SPS/PPS, ideally, the layer system supports this by design?  E.g. you declare SPS/PPS on an `order=0` layer, and make all subsequent video frames depend on it.  We should make sure that having a permanent `order=0` layer does not prevent subsequent layers from being discarded.

suhasHere:
@vasilvv agree. I think there are couple of issues that are related to
- media capabilities 
- tracks
- manifest 
all these need a similar thinking at a higher level  which is - "way to setup and modify media capabilites between the publishers and consumers.":

kixelated:
A few features are missing but this is CATALOG.

PR #43, opened 2022-11-15
Initial track PUBLISH/SUBSCRIBE

### LINKED ISSUE #13: PLAY message
Currently, there's no way to determine which layers/tracks get pushed. We probably need a new message to control which tracks the consumer wants. We need to support both simulcast (send all tracks), ABR (send one track of this group), and manual track selection (send just these tracks).

kixelated:
Closes #13 #12 

This covers a few use-cases:

1. The sender can advertise support for multiple codecs/containers.
2. The receiver can request a specific track, for example: english or japannese audio.
3. The receiver can request one of many tracks, so the server can perform ABR. For example: one of 480p, 720p, or 1080p.
4. The receiver can request multiple tracks of the same rendition for multicast. For example: 480p, 720p, and 1080p.
5. The receiver can request only tracks it supports. For example: no 1080p due to an unsupported profile.
6. The receiver can seamlessly switch between enabled tracks. For example: send me the ad track while enabled, then switch to the content track.
7. The sender can pre-emptively send tracks before the PLAY message, avoiding a round trip.

There's a few use-cases that aren't covered:

1. The receiver wants to specify when the track should be switched (timestamp).
2. The receiver wants to go rewind and request older content for a track.

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -329,7 +329,7 @@ These are similar to QUIC and HTTP/3 frames, but called messages to avoid the me
 Each stream MUST start with a `HEADERS` message ({{headers}}) to indicates how the stream should be transmitted.
 
 Messages SHOULD be sent over the same stream if ordering is desired.
-For example, `PAUSE` and `PLAY` messages SHOULD be sent on the same stream to avoid a race.
+For example, `PLAY` messages SHOULD be sent on the same stream to avoid a race.
 
 ## Prioritization
 Warp utilizes stream prioritization to deliver the most important content during congestion.
@@ -471,7 +471,7 @@ The sender advertises tracks that are available.
 The receiver chooses which tracks to `PLAY` ({{play}}) based on this information.
 
 Tracks MAY be updated by sending a new `TRACK` message with an existing `track_id`.
-The most recently received `TRACK` message takes presendence, so they SHOULD be sent over the same control to preserve ordering.
+Messages with the same `track_id` SHOULD be sent over the same stream to preserve ordering.
 
 General properties:
 
@@ -511,31 +511,27 @@ An optional integer height in pixels.
 * `frame_rate`
 An optional decimal frame rate.
 
+
 Note that none of information is required to decode the media bitstream; that information is contained in init segments.
 A decoder MAY choose to display media before receiving the cooresponding `TRACK` message.
 
 
 ## PLAY
-The `PLAY` message is sent by the receiver to indicate which tracks it wishes to receive.
+The `PLAY` message is sent by the receiver to indicate that it wishes to receive a track.
 This is a hint; the sender SHOULD try to respect the request but it MAY transmit any tracks that it wants.
 
+* `play_id`.
+A identifier for the PLAY message.
+The next PLAY message with the same `play_id` will override this message.
+Messages with the same `play_id` SHOULD be sent over the same stream to preserve ordering.
+
 * `track_ids`.
-An optional list of track identifiers.
+A list of track identifiers in order of precidence.
+If the list is not empty, the sender SHOULD transmit at least one of the tracks based on availability and network conditions.
 
-* `group_ids`.
-An optional list of group identifiers.
-The sender SHOULD transmit at least one track within each group and it SHOULD be the track with the highest sustainable `max_bitrate`, which can change.
-
-* `timestamp`.
-An optional timestamp indicating when this should take effect.
-A negative value is relative to the maximum timestamp known by the sender.
-The sender SHOULD transmit any segments newer than the timestamp, and MAY transmit any segments older than the timestamp.
-If the provided timestamp is in the future, the sender SHOULD follow prior `PLAY` messages until then.
-
-The receiver MAY perform receiver-side ABR by requesting one `track_id` in each group according to network conditions.
-However, the sender has direct access to the congestion controller, so the receiver SHOULD request `group_ids` to perform sender-side ABR instead.
-
-The sender MAY transmit default track(s) prior to the first `PLAY` message to avoid a round-trip on connect.
+Before the receipt of the first `PLAY` message, the sender MAY choose default tracks to transmit.
+This both avoids a round-trip of startup latency and starts warming the connection.
+Upon the receipt of the first `PLAY` message, the sender SHOULD stop transmitting these default tracks.
 
 
 ## GOAWAY
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -248,7 +248,7 @@ A segment:
 * MAY overlap with other segments. This means timestamps may be interleaved between segments.
 * MAY reference frames in other segments, but only if listed as a dependency.
 
-Segments are packaged using a specified container ({{containers}}).
+Segments are encoded using a specified container ({{containers}}).
 
 ## Delivery Order
 Media is produced with an intended order, both in terms of when media should be presented (PTS) and when media should be decoded (DTS).
@@ -326,7 +326,7 @@ See messages ({{messages}}) for the list of messages and their encoding.
 These are similar to QUIC and HTTP/3 frames, but called messages to avoid the media terminology.
 
 Messages SHOULD be sent over the same stream if ordering is desired.
-For example, `PLAY` messages SHOULD be sent on the same stream to avoid a race.
+For example, TRACK\_PUBLISH and TRACK\_SUBSCRIBE messages SHOULD be sent on the same stream to avoid a race.
 
 ## Prioritization
 Warp utilizes stream prioritization to deliver the most important content during congestion.
@@ -398,17 +398,17 @@ Messages consist of a type identifier followed by contents, depending on the mes
 
 TODO document the encoding
 
-|------|---------------------------|
-| ID   | Messages                  |
-|-----:|:--------------------------|
-| 0x0  | SEGMENT ({{segment}})     |
-|------|---------------------------|
-| 0x1  | PUBLISH ({{publish}})     |
-|------|---------------------------|
-| 0x2  | SUBSCRIBE ({{subscribe}}) |
-|------|---------------------------|
-| 0x10 | GOAWAY ({{goaway}})       |
-|------|---------------------------|
+|------|----------------------------------------|
+| ID   | Messages                               |
+|-----:|:---------------------------------------|
+| 0x0  | SEGMENT ({{segment}})                  |
+|------|----------------------------------------|
+| 0x1  | TRACK\_PUBLISH ({{track-publish}})     |
+|------|----------------------------------------|
+| 0x2  | TRACK\_SUBSCRIBE ({{track-subscribe}}) |
+|------|----------------------------------------|
+| 0x10 | GOAWAY ({{goaway}})                    |
+|------|----------------------------------------|
 
 
 ## SEGMENT
@@ -417,8 +417,12 @@ A SEGMENT message contains a single segment associated with a specified track, a
 Each SEGMENT message starts with a header, containing information useful for a relay:
 
 * `track_id`.
-The track identifier. {{track}}
-A decoder MUST block until the cooresponding PUBLISH message has been received.
+The track identifier. {{track-publish}}
+When using a container with an initialization payload, the decoder MUST block until the cooresponding TRACK\_PUBLISH message ({{track-publish}}) has been received.
+
+* `segment_id`.
+A unique identifier for each segment within a track.
+It is RECOMMENDED that this is an monotonically increasing sequence number, but the receiver MUST NOT assume that it is.
 
 * `order`.
 An integer indicating the delivery order ({{delivery-order}}).
@@ -426,28 +430,29 @@ This field is optional; the default value is 0.
 
 * `depends`.
 An list of dependencies by `segment_id` ({{dependencies}}).
-This field is optional; the default value is an empty array.
-
-* `uri`.
-An identifier for the payload, useful for backwards compatibility with HTTP caches.
-This field is optional.
+This field is optional; an empty array indicates the segment is independent.
 
 The remainder of the SEGMENT message is a payload depending on the track `mime_type`.
 This contains a media bitstream intended for the decoder and SHOULD NOT be processed by a relay.
 See the containers section ({{containers}}).
 
 
-## PUBLISH
-The sender advertises an available track via the PUBLISH message.
-The receiver can ask for tracks via the SUBSCRIBE ({{subscribe}}) message based on this information.
+## TRACK\_PUBLISH {#track-publish}
+The sender advertises an available track via the TRACK\_PUBLISH message.
+The receiver can ask for tracks via the TRACK\_SUBSCRIBE ({{track-subscribe}}) message based on this information.
 
-Tracks MAY be updated by sending a new PUBLISH message with an existing `track_id`.
-PUBLISH messages with the same `track_id` SHOULD be sent over the same stream to preserve ordering.
+Tracks MAY be updated by sending a new TRACK\_PUBLISH message with an existing `track_id`.
+TRACK\_PUBLISH messages with the same `track_id` SHOULD be sent over the same stream to preserve ordering.
 
-Each PUBLISH message starts with a header, containing information useful for a relay:
+Each TRACK\_PUBLISH message starts with a header, containing information useful for a relay:
 
-* `track_id`.
+* `publish_id`.
 A unique identifier for the track.
+
+* `active`.
+An boolean indicating that the track can be served.
+An inactive track SHOULD NOT be requested via TRACK\_SUBSCRIBE.
+A track MAY be inactive because is has not started yet, or it has ended.
 
 * `group_id`.
 An identifier indicating that this track is part of a group.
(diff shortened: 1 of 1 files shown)


draft-lcurley-warp.md
```diff
@@ -455,6 +465,78 @@ This is useful for metadata that would otherwise have to be shoved into the medi
 Relays MUST NOT differentiate between streams containing `SEGMENT` and `APP` frames.
 The same forwarding and caching behavior applies to both as specified in the`HEADERS` frame.
 
+## TRACK
+The `TRACK` message contains information about a specific track.
+The sender advertises tracks that are available.
+The receiver chooses which tracks to `PLAY` ({{play}}) based on this information.
+
+Tracks MAY be updated by sending a new `TRACK` message with an existing `track_id`.
+The most recently received `TRACK` message takes presendence, so they SHOULD be sent over the same control to preserve ordering.
```
afrind, line 474: control stream?

draft-lcurley-warp.md
```diff
@@ -455,6 +465,78 @@ This is useful for metadata that would otherwise have to be shoved into the medi
...
+* `frame_rate`
+An optional decimal frame rate.
+
+Note that none of information is required to decode the media bitstream; that information is contained in init segments.
+A decoder MAY choose to display media before receiving the cooresponding `TRACK` message.
+
+
+## PLAY
+The `PLAY` message is sent by the receiver to indicate which tracks it wishes to receive.
+This is a hint; the sender SHOULD try to respect the request but it MAY transmit any tracks that it wants.
```
afrind, line 520: Should there be a way to request the sender to stop sending a given track?

kixelated: Absolutely; I was thinking you would send a PLAY with fewer (perhaps none) tracks.

kixelated: But after thinking about this some more, I think a PLAY/PAUSE message per track/group is just simpler to encode.

draft-lcurley-warp.md
kixelated: oops yeah

draft-lcurley-warp.md
```diff
@@ -455,6 +465,78 @@ This is useful for metadata that would otherwise have to be shoved into the medi
...
+
+* `track_ids`.
+An optional list of track identifiers.
+
+* `group_ids`.
+An optional list of group identifiers.
+The sender SHOULD transmit at least one track within each group and it SHOULD be the track with the highest sustainable `max_bitrate`, which can change.
+
+* `timestamp`.
```
kixelated, line 529: I don't think this timestamp field is particularly well thought out. I think we remove it for now and specify something later once the use-cases are better explored.

vasilvv:
I guess I have two thoughts regarding the track listing.

First of all, I am not sure we should be describing media parameters ourselves. MP4 already has fields for video width and length (in fact, a typical moov atom would have those repeated at least twice), so ideally we would just defer to those, since otherwise we're stuck recording more and more media-specific fields, which is something deferring to an existing container was supposed to avoid.

Second, this is definitely the solution that covers the most complex scenario (delivery with multiple formats), but I would also like us to support the simplest case (which would probably be ingestion where all formats are fixed at the start), and then extend up to the complex solution where the use case requires it.  That probably means we should make adding tracks mid-session a negotiated extension.

kixelated:
> First of all, I am not sure we should be describing media parameters ourselves. MP4 already has fields for video width and length (in fact, a typical moov atom would have those repeated at least twice), so ideally we would just defer to those, since otherwise we're stuck recording more and more media-specific fields, which is something deferring to an existing container was supposed to avoid.

That's a great idea. We could combine the `INIT` and `TRACK` message. The only change is that senders MUST push `INIT` messages for all tracks, whereas before it was only needed after `PLAY`.


For those not aware, here's an example of what an [MP4 parser](https://github.com/gpac/mp4box.js) will spit out. We're using fragmented MP4 in our case so durations are 0, but it's very similar.
```
{
  "duration":360002,
  "timescale":600,
  "isFragmented":false,
  "isProgressive":true,
  "hasIOD":true,
  "brands":["isom"],
  "created":"2014-04-15T18:24:40.000Z",
  "modified":"2014-04-15T18:24:40.000Z",
  "tracks":[
    {
	  "id":2,
	  "created":"2012-02-13T23:07:31.000Z",
	  "modified":"2014-04-16T12:22:56.000Z",
	  "movie_duration":360000,
	  "layer":0,
	  "alternate_group":0,
	  "volume":0,
	  "track_width":320,
	  "track_height":180,
	  "timescale":25000,
	  "duration":15000000,
	  "bitrate":120000,
	  "codec":"avc1.42c00d",
	  "video":{
	    "width":320,"height":180
	  },
	  "language":"und",
	  "nb_samples":15000
	},
	{
	  "id":3,
	  "created":"2012-09-12T11:14:57.000Z",
	  "modified":"2014-04-16T12:22:56.000Z",
	  "movie_duration":360002,
	  "layer":0,
	  "alternate_group":0,
	  "volume":1,
	  "track_width":0,
	  "track_height":0,
	  "timescale":44100,
	  "duration":26460160,
	  "bitrate":60000,
	  "codec":"mp4a.40.2",
	  "audio":{
	    "sample_rate":44100,
		"channel_count":1,
		"sample_size":16
	  },
	  "language":"und",
	  "nb_samples":25840
	}
  ]
}
```

Technically, MP4 already contains a `track_id`. We could reuse it, but I think it's prudent to make a new transport ID so it's easier to stitch without re-encoding.

The one downside with MP4 is that the INIT segment are relatively large (looks like 800 bytes for single h.264 track). I definitely agree that we should specify an alternative container in the future to tackle the large size.


> Second, this is definitely the solution that covers the most complex scenario (delivery with multiple formats), but I would also like us to support the simplest case (which would probably be ingestion where all formats are fixed at the start), and then extend up to the complex solution where the use case requires it. That probably means we should make adding tracks mid-session a negotiated extension.

I think we could expand upon the idea of "default tracks", where the sender just chooses what to push and possibly ignores any PLAY messages. This was mostly added to avoid a round-trip of latency but it also works for simple contribution scenarios where there's no choice involved.

draft-lcurley-warp.md
```diff
@@ -474,6 +512,29 @@ The client:
 * SHOULD establish the connection in parallel which MUST use different QUIC connection.
 * SHOULD remain connected for two servers for a short period, processing segments from both in parallel.
 
+# Containers
```
kixelated, line 515: I can also split this out into another PR.

draft-lcurley-warp.md
```diff
@@ -400,60 +396,102 @@ Messages consist of a type identifier followed by contents, depending on the mes
...
-| 0x2  | APP ({{app}})         |
-|------|-----------------------|
-| 0x10 | GOAWAY ({{goaway}})   |
-|------|-----------------------|
+|------|---------------------------|
+| ID   | Messages                  |
+|-----:|:--------------------------|
+| 0x0  | SEGMENT ({{segment}})     |
+|------|---------------------------|
+| 0x1  | TRACK ({{track}})         |
```
kixelated, line 404: PUBLISH could also work but it's a bit more awkward.

draft-lcurley-warp.md
```diff
@@ -400,60 +396,102 @@ Messages consist of a type identifier followed by contents, depending on the mes
...
-This field is optional and the default value is 0.
+This field is optional; the default value is 0.
 
 * `depends`.
-An list of dependencies by stream identifier ({{dependencies}}).
-This field is optional and the default value is an empty array.
+An list of dependencies by `segment_id` ({{dependencies}}).
+This field is optional; the default value is an empty array.
 
+* `uri`.
```
kixelated, line 429: @suhasHere The idea is that a CDN can cache the payload using this path/URI. We might need a byte offset too.

draft-lcurley-warp.md
suhasHere: I know naming is always challenging .. SEGMENT makes sense for video but for audio, it wasn't very obvious on initial read.  One of the thoughts I had was called any media data like `Moq Media Data` which is made up of one more Media Object/Media Frames .. Should we call it as `Warp Media Data ` ?? not sure..

On the track, there are 2 scenarios
1. Update track mid-session. For this use-case having Track as a standalone message makes sense
2. I want to request for sending media and advertise my tracks at the same time. I was referring to this as `PUBLISH_REQUEST` message which has along with other things one or more tracks defined via TrackInfo/TrackDescriptors.

I think we need both the standalone track message as well as sometning where you can embed your tracks a sender as the first message when you are about to start publishing and sent via the control stream

draft-lcurley-warp.md
```diff
@@ -402,136 +396,102 @@ Messages consist of a type identifier followed by contents, depending on the mes
...
 
-Media fragments can be packaged at any frequency, causing a trade-off between overhead and latency.
-It is RECOMMENDED that a media fragment consists of a single frame to minimize latency.
-
-## APP
-The `APP` message contains arbitrary contents.
-This is useful for metadata that would otherwise have to be shoved into the media bitstream.
+The remainder of the SEGMENT message is a payload depending on the track `mime_type`.
+This contains a media bitstream intended for the decoder and SHOULD NOT be processed by a relay.
+See the containers section ({{containers}}).
```
suhasHere, line 435: On the Media Data Message/Segment Message, we need to support a single wire encoding that works where a sender can operate in one or 2 modes 
- one frame per stream ( rush)
   This needs one media frame per media data
- one gop per stream (warp)
    This needs one or more media frames per media data

```
Segment {
...
... 
MediaFrame(...) ...  // list of one frame for rush, list of one or more frames for warp
}
```
Something like this can support both the cases. Also how the media frame/media object itself is represented is specific to the underlying container format (fmp4/something else).

I might be mis-reading, but I am unable to understand how to support these sender modes in a generic way

draft-lcurley-warp.md
```diff
@@ -402,136 +396,102 @@ Messages consist of a type identifier followed by contents, depending on the mes
...
 
 * `track_id`.
 A unique identifier for the track.
 
 * `group_id`.
-An optional identifier indicating that this track is part of a group.
+An identifier indicating that this track is part of a group.
 Tracks within the same group contain the same content but alternate encodings.
 For example: different resolutions, bitrates, languages, or codecs.
+This field is optional.
```
suhasHere, line 454: just trying to understand my vocabulary here. There are these things

- Different rendition - HD Video @ 30fps, HD video @60fps ... [ I was thinking these as track ]
- One or more Group of pictures  .... [ I was thinking each of these as identified by a groupId .. so equivalent to how CDN stores the segments like  1.mp4, 2.mp4  ( so one file per gop) ]
- Each GOP has several media frames insider it (I, P, ... ) .. [ I was thinking each such thing inside a GOP as MediaFrame/Media Object]


could you please help me understand if by group above, you mean the same or something different ?

suhasHere:
Thanks Luke for starting this PR and it is very helpful to get going.

draft-lcurley-warp.md
kixelated: In HLS/DASH, they're known as audio segments. I don't think we need different terminology for similar concepts just because it's audio or video (ex. audio frames).

kixelated: I'd like to optimize startup latency, and I think we can just send media without an explicit handshake.

Broadcast use-case:
```
sender           receiver
----------------------------
 CONNECT     -->
 TRACK (1)   -->
 TRACK (2)   -->
 SEGMENT (1) -->
 SEGMENT (2) -->
             <--  CONNECT
 SEGMENT (1) -->
 SEGMENT (2) -->
```

Viewer use-case:
```
sender           receiver
----------------------------
             <-- CONNECT
 CONNECT     -->
 TRACK (1)   -->
 TRACK (2)   -->
 TRACK (3)   -->
 TRACK (4)   -->
 SEGMENT (1) -->
 SEGMENT (2) -->
             <--  PLAY (1,4)
 SEGMENT (1) -->
 SEGMENT (4) -->
```

kixelated: If you require a handshake or acknowledgement for each subscription, it looks like:

Broadcast use-case:
```
sender           receiver
----------------------------
 CONNECT     -->
 TRACK (1)   -->
 TRACK (2)   -->
             <--  CONNECT
             <--  PLAY (1,2)
 SEGMENT (1) -->
 SEGMENT (2) -->
```

Viewer use-case:
```
sender           receiver
----------------------------
             <-- CONNECT
 CONNECT     -->
 TRACK (1)   -->
 TRACK (2)   -->
 TRACK (3)   -->
 TRACK (4)   -->
             <--  PLAY (1,4)
 SEGMENT (1) -->
 SEGMENT (4) -->
```

kixelated: The extra round-trip sucks for the broadcast scenario because it's obvious 99% of the time. In general, if the sender guesses wrong, then receiver can issue a PLAY message and potentially STOP_SENDING the segments it doesn't care about.

draft-lcurley-warp.md
kixelated: Yeah that encoding is pretty much how containers already work. 

* A fMP4 segment has any number of fragments, each of which can contain any number of frames.
* A TS segment has any number of frames.

kixelated: I don't think there's a compelling reason to hoist GoP/frame semantics into the transport layer. 

For example, does a relay need to know if two segments are part of the same GoP? Not really; it just needs to know which order to deliver them.

draft-lcurley-warp.md
kixelated: `rendition == track` when it's the same content, but different encoding
`rendition != track` when it's different content.

Each track consists of GoPs, which consist of segments, which consist of frames.

I might just rename segment to "fragment" though, since that's really what it's called in CMAF.

draft-lcurley-warp.md
```diff
@@ -400,60 +396,102 @@ Messages consist of a type identifier followed by contents, depending on the mes
...
-| 0x10 | GOAWAY ({{goaway}})   |
-|------|-----------------------|
+|------|---------------------------|
+| ID   | Messages                  |
+|-----:|:--------------------------|
+| 0x0  | SEGMENT ({{segment}})     |
+|------|---------------------------|
+| 0x1  | TRACK ({{track}})         |
+|------|---------------------------|
+| 0x2  | SUBSCRIBE ({{subscribe}}) |
```
afrind, line 408: Is SUBSCRIBE the same as PLAY then?

draft-lcurley-warp.md
```diff
@@ -400,60 +396,102 @@ Messages consist of a type identifier followed by contents, depending on the mes
...
+These messages SHOULD be sent over the same stream to preserve ordering.
+
+* `track_ids`.
+A list of track identifiers in order of precidence.
+If the list is not empty, the sender SHOULD transmit at least one of the tracks based on availability and network conditions.
+
+
+Prior to the receipt of the first SUBSCRIBE message, the sender MAY choose default tracks to transmit.
+This can avoid a round-trip of startup latency when the decision is obvious, or at the very least start warming the connection.
+Upon the receipt of the first SUBSCRIBE message, the sender SHOULD stop transmitting these default tracks.
```
afrind, line 495: I like the idea of explicitly cancelling them via STOP_SENDING

afrind (commented):
General comment is that PUBLISH / SUBSCRIBE / SEGMENT are kind of misusing pub/sub terminology, right?  Shouldn't it be:

ADVERTISE / SUBSCRIBE / PUBLISH

draft-lcurley-warp.md
kixelated: > General comment is that PUBLISH / SUBSCRIBE / SEGMENT are kind of misusing pub/sub terminology, right? Shouldn't it be:
> 
> ADVERTISE / SUBSCRIBE / PUBLISH

Yeah, true. I wanted to capture some of the QUICR terminology, but afraid we're encroaching on generic verbs here. I much prefer the crude media terms and can revert it if you guys agree.

RTMP does [publish](https://rtmp.veriskope.com/docs/spec/#7226publish), [play](https://rtmp.veriskope.com/docs/spec/#7221play), [play2](https://rtmp.veriskope.com/docs/spec/#7222play2), [pause](https://rtmp.veriskope.com/docs/spec/#7228pause), [seek](https://rtmp.veriskope.com/docs/spec/#7227seek). It has separate messages for [audio](https://rtmp.veriskope.com/docs/spec/#714audio-message-8) and [video](https://rtmp.veriskope.com/docs/spec/#715video-message-9).

`TRACK` = `publish`
`PLAY` = `play`, `play2`, `pause`
`SEGMENT` = `video`, `audio`

kixelated:
Some open TODOs for this PR or the next one:

1. Do we need acknowledgements?
2. Add some timestamps so you can SUBSCRIBE to a relative or absolute timestamp.
3. Signal when a track has ended.
4. Naming.

vasilvv (approved):

draft-lcurley-warp.md
```diff
@@ -143,7 +143,7 @@ Producer:
 
 Rendition:
```
afrind, line 144: Is it a rendition or a variant?

draft-lcurley-warp.md
```diff
@@ -481,37 +466,121 @@ A SEGMENT message contains a single segment associated with a specified track, a
...
 
 ~~~
-SEGMENT Header {
-  Header Name Length (8),
-  Header Name (..),
-  Header Value Length (i),
-  Header Value (..),
-}
-
 SEGMENT Message {
```
afrind, line 469: So per the object model PR, a SEGMENT corresponds to a "media object"?  We can reconcile after one or both of them land.

draft-lcurley-warp.md
```diff
@@ -481,37 +466,121 @@ A SEGMENT message contains a single segment associated with a specified track, a
...
-* `depends`.
-An list of dependencies by stream identifier ({{dependencies}}).
-This field is optional and the default value is an empty array.
+* Dependency Count:
+The number of dependency segment IDs that follow.
+A synchronization point (ex. i-frame) will have zero dependencies.
+
+* Dependency Segment IDs:
+An list dependencies by Segment ID ({{dependencies}}), each of which is a variable integer.
+The decoder SHOULD block until the cooresponding SEGMENT messages have been received.
```
afrind, line 499: We don't need to add this now, but we'll probably want some language like QPACK has regarding flow control:

1) leave the payload bytes in the flow control window until the deps are satisfied
2) beware flow control deadlocks

draft-lcurley-warp.md
```diff
@@ -481,37 +466,121 @@ A SEGMENT message contains a single segment associated with a specified track, a
...
+* Track Count
+The number of track IDs that follow.
+
+* Track IDs
+A list of track identifiers, each of which is a variable integer.
+The receiver SHOULD arrange the list in preferred order; for example 1080p takes precidence over 480p.
+The sender SHOULD transmit at least one track based on this preferred order, the track availablity, and network conditions.
+
+
+When no tracks are specified, the receiver indicates that it wants to cancel the subscription with the same ID.
```
afrind, line 567: Probably clearer to just have UNSUBSCRIBE_TRACK?

draft-lcurley-warp.md
```diff
@@ -481,37 +466,121 @@ A SEGMENT message contains a single segment associated with a specified track, a
...
+
+* Subscribe ID:
+An identifier for the subscription.
+A subscription MAY be updated by sending a SUBSCRIBE\_TRACK message with the same Subscribe ID.
+
+* Track Count
+The number of track IDs that follow.
+
+* Track IDs
+A list of track identifiers, each of which is a variable integer.
```
afrind, line 562: The object model says track IDs are unique within a broadcast, not necessarily globally unique, so I think you need to specify the broadcast URL also?

draft-lcurley-warp.md
```diff
@@ -481,37 +466,121 @@ A SEGMENT message contains a single segment associated with a specified track, a
...
+The receiver SHOULD arrange the list in preferred order; for example 1080p takes precidence over 480p.
+The sender SHOULD transmit at least one track based on this preferred order, the track availablity, and network conditions.
+
+
+When no tracks are specified, the receiver indicates that it wants to cancel the subscription with the same ID.
+
+When a single track is specified, the receiver indicates it wants that specific track.
+This is useful for indicating user preference (ex. send japanese audio) and receiver-side ABR (ex. send 480p)
+
+When multiple tracks are specified, the receiver indicates that the sender should choose one.
```
afrind, line 572: Not all tracks are variants of each other, and it sounds like this mode of subscription would only be valid when that is true.  eg:

```
SUBSCRIBE_TRACK:
  broadcast url: <x>
  tracks: <1080p>, <480p>
SUBSCRIBE_TRACK:
  broadcast url: <x>
  tracks: <japanese audio>
```

So presumably the PUBLISH_TRACK message has some identifier which helps the receiver understand which tracks are variants (track group)?

Is there more nuance to what needs to be specified for "automatic sender-side variant selection based on congestion" than is covered here?

Since it sounds like some relay providers are not yet onboard for this feature, it is something that would need to be negotiated and fall back to client ABR selection?  Even if it is we might just want to note that for now and revisit it later.

draft-lcurley-warp.md
```diff
@@ -481,37 +466,121 @@ A SEGMENT message contains a single segment associated with a specified track, a
...
+This is useful for sender-side ABR (ex. send 1080p otherwise 480p) and track stitching (ex. send an advertisement otherwise the original content).
+
+The receiver MAY have multiple active subscriptions.
+The sender MUST avoid sending the same tracks twice when they overlap.
```
afrind, line 576: `MUST avoid sending` seems slightly ambiguous -- `MUST NOT send` perhaps?

What about the broadcast case where the broadcaster is a transcoding relay feeding a CDN?

draft-lcurley-warp.md
```diff
@@ -481,37 +466,121 @@ A SEGMENT message contains a single segment associated with a specified track, a
...
+
+Prior to the receipt of the first SUBSCRIBE\_TRACK message, the sender MAY choose default tracks to transmit.
+This can avoid a round-trip of startup latency when the decision is obvious, or at the very least start warming the connection.
+
+Upon the receipt of the first SUBSCRIBE\_TRACK message, the sender SHOULD stop transmitting these default tracks.
```
afrind, line 581: This feels racy but I don't have an example yet.

draft-lcurley-warp.md
```diff
@@ -577,11 +654,12 @@ Each media fragment consists of a Movie Fragment Box (moof) followed by a Media
 The Media Fragment Box (moof) MUST contain a Movie Fragment Header Box (mfhd) and Track Box (trak) with a Track ID (`track_ID`) matching a Track Box in the initialization fragment.
 A Common Media Application Format Segment {{CMAF}} meets all these requirements.
 
-Each SEGMENT message ({{segment}}) MUST start with an initialization segment, or MUST depend on a SEGMENT that does. The rest of the SEGMENT message MAY be a media segment.
-
 Media fragments can be packaged at any frequency, causing a trade-off between overhead and latency.
 It is RECOMMENDED that a media fragment consists of a single frame to minimize latency.
 
+The PUBLISH\_TRACK message ({{publish-track}}) payload MUST be an initization segment.
+The SEGMENT message ({{segment}}) payload MUST be a media segment, which consists of any number of media fragments.
```
afrind, line 661: Are you referring to any SEGMENT messages sent with a matching Track ID?

afrind (commented):
This feels close, thanks for the work here.

kixelated:
Closed in favor of #63

PR #46, opened 2022-11-16
Initial proposal for the wire format

vasilvv:
This also merges HEADERS/SEGMENT and removes APP.

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -79,6 +79,10 @@ Bitstream:
 
 : A continunous series of bytes.
 
+Client:
+
+: The party initiating a Warp session.
+
 Codec:
 
 : A compression algorithm for audio or video.
@@ -139,13 +143,19 @@ Rendition:
 
 : One or more tracks with the same content but different encodings.
 
+Server:
+
+: The party accepting an incoming Warp session.
+
 Slice:
 
 : A section of a video frame. There may be multiple slices per frame.
 
 Track:
 
-: An encoded bitstream, representing a single video/audio component that makes up the larger broadcast.
+: An encoded bitstream, representing a single video/audio component that makes up the larger broadcast. See {{tracks}}.
+
+This document uses the conventions detailed in Section 1.3 of {{!RFC9000}} when describing the binary encoding.
 
 
 # Motivation
@@ -276,7 +286,7 @@ This creates hard and soft dependencies that need to be respected by the transpo
 See the appendex for an overview of media encoding ({{appendix.encoding}}).
 
 A segment MAY depend on any number of other segments.
-The encoder MUST indicate these dependecies on the wire via the `HEADERS` message ({{headers}}).
+The encoder MUST indicate these dependecies on the wire via the segment headers.
 
 The sender SHOULD NOT use this list of dependencies to determine which segment to transmit next.
 The sender SHOULD use the delivery order instead, which MUST respect dependencies.
@@ -317,14 +327,9 @@ The application SHOULD use the CONNECT request for authentication.
 For example, including a authentication token and some identifier in the path.
 
 ## Streams
-Warp endpoints communicate over unidirectional QUIC streams.
-The application MAY use bidirectional QUIC streams for other purposes.
+Warp endpoints communicate over QUIC streams. Every stream is a sequence of messages, framed as described in {{messages}}.
 
-A stream consists of sequential messages.
-See messages ({{messages}}) for the list of messages and their encoding.
-These are similar to QUIC and HTTP/3 frames, but called messages to avoid the media terminology.
-
-Each stream MUST start with a `HEADERS` message ({{headers}}) to indicates how the stream should be transmitted.
+The first stream opened is a bidirectional stream where the peers exchange INIT messages ({{init}}). The subsequent streams MAY be either unidirectional and bidirectional. For exchanging media, an application would typically send a unidirectional stream containing a single SEGMENT message ({{segment}}).
 
 Messages SHOULD be sent over the same stream if ordering is desired.
 For example, `PAUSE` and `PLAY` messages SHOULD be sent on the same stream to avoid a race.
@@ -356,8 +361,7 @@ The sender MAY cancel streams in response to congestion.
 This can be useful when the sender does not support stream prioritization.
 
 ## Relays
-Warp encodes the delivery information for each stream via a `HEADERS` frame ({{headers}}).
-This MUST be at the start of each stream so it is easy for a relay to parse.
+Warp encodes the delivery information for each stream via segment headers.
 
 A relay SHOULD prioritize streams ({{prioritization}}) based on the delivery order.
 A relay MAY change the delivery order, in which case it SHOULD update the value on the wire for future hops.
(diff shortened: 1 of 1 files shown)


draft-lcurley-warp.md
```diff
@@ -317,14 +327,9 @@ The application SHOULD use the CONNECT request for authentication.
...
-Warp endpoints communicate over unidirectional QUIC streams.
-The application MAY use bidirectional QUIC streams for other purposes.
-
-A stream consists of sequential messages.
-See messages ({{messages}}) for the list of messages and their encoding.
-These are similar to QUIC and HTTP/3 frames, but called messages to avoid the media terminology.
+Warp endpoints communicate over QUIC streams. Every stream is a sequence of messages, framed as described in {{messages}}.
 
-Each stream MUST start with a `HEADERS` message ({{headers}}) to indicates how the stream should be transmitted.
+The first stream opened is a bidirectional stream where the peers exchange INIT messages ({{init}}). The subsequent streams MAY be either unidirectional and bidirectional. For exchanging media, an application would typically send a unidirectional stream containing a single SEGMENT message ({{segment}}).
```
kixelated, line 332: Which endpoint opens the bidirectional stream?

draft-lcurley-warp.md
```diff
@@ -395,28 +399,105 @@ The application SHOULD use a non-zero error code to indicate a fatal error.
...
 |------|-----------------------|
 | ID   | Messages              |
 |-----:|:----------------------|
-| 0x0  | HEADERS ({{headers}}) |
-|------|-----------------------|
-| 0x1  | SEGMENT ({{segment}}) |
+| 0x0  | SEGMENT ({{segment}}) |
 |------|-----------------------|
-| 0x2  | APP ({{app}})         |
+| 0x1  | INIT ({{init}})       |
```
kixelated, line 441: Nit: can we call it `SETUP`, `SETTINGS`, or something else instead? "init" is somewhat used in media terminology already.

draft-lcurley-warp.md
```diff
@@ -474,6 +544,48 @@ The client:
...
+
+: Only the server is expected to send media on the connection. This is commonly referred to as *the delivery case*.
+
+0x03:
+
+: Both the client and the server are expected to send media.
+
+The client MUST send a ROLE parameter with one of the three values specified above. The server MUST close the connection if the ROLE parameter is missing, is not one of the three above-specified values, or it is different from what the server expects based on the application in question.
+
+## INITIAL_TRACKS parameter
```
kixelated, line 573: This is fine for now but we do need a way to update/add/remove tracks, and then this will be redundant.

draft-lcurley-warp.md
kixelated: There's also some latency considerations here. Do endpoints need to wait for the `INIT` message before transmitting media?

draft-lcurley-warp.md
```diff
@@ -395,28 +399,105 @@ The application SHOULD use a non-zero error code to indicate a fatal error.
...
+~~~
+Track Descriptor {
+  Track ID (i),
+  Track Format (i),
+  Format-Specific Metadata (...),
+}
+~~~
+{: #warp-track-descriptor title="Warp Track Descriptor"}
+
+Here, Track ID is the unique ID that identifies the track within the broadcast. Track Format idetifies the format used by the track; for the ISOBMFF-based format described in this document, 0x00 is specified. In the case 0x00 is used, Format-Specific Metadata contains a `ftyp` box followed by a `moov` box.
```
kixelated, line 417: To be more specific, this is a init segment, or more correctly a `CMAF header`. There's already text for this that you can cut/paste from the SEGMENT message. The terminology is lifted straight from the HLS specification.

> Each segment MUST start with an initialization fragment, or MUST depend on a segment with an initialization fragment. An initialization fragment consists of a File Type Box (ftyp) followed by a Movie Box (moov). This Movie Box (moov) consists of Movie Header Boxes (mvhd), Track Header Boxes (tkhd), Track Boxes (trak), followed by a final Movie Extends Box (mvex). These boxes MUST NOT contain any samples and MUST have a duration of zero. Note that a Common Media Application Format Header [CMAF] meets all these requirements.

And yeah, we don't need each SEGMENT to start with this init segment now that track descriptor exists.

draft-lcurley-warp.md
```diff
@@ -395,28 +399,105 @@ The application SHOULD use a non-zero error code to indicate a fatal error.
...
+
+The INIT parameters are described in the {{init-parameters}} section.
+
+## SEGMENT
+A `SEGMENT` message contains a single segment associated with a specified track, as well as associated metadata required to deliver, cache, and forward it.
+
+The format of the SEGMENT message is as follows:
+
+~~~
+SEGMENT Header {
```
kixelated, line 484: I think we should define a concrete encoding rather than arbitrary key/value pairs. I'm not sure though; relays running different versions complicates things.

draft-lcurley-warp.md
kixelated: And yeah I know this is a TODO; CMAF doesn't need to be done in this PR.

draft-lcurley-warp.md
```diff
@@ -395,28 +399,105 @@ The application SHOULD use a non-zero error code to indicate a fatal error.
...
+~~~
+INIT Parameter {
+  Parameter Key (i),
+  Parameter Value Length (i),
+  Parameter Value (..),
+}
+
+Client INIT Message Payload {
+  Number of Supported Versions (i),
+  Supported Version (i) ...,
```
kixelated, line 461: What's the plan when relays running different versions are thrown into the mix? Implementations should not need to support every permutation of ingesting version x and distributing version y.

I don't think there's much we can do except to be backwards-compatible, but I want to make sure we think about it.

draft-lcurley-warp.md
```diff
@@ -474,6 +544,48 @@ The client:
...
+If the peer is expected to send media according to the negotiated ROLE, it MUST send an INITIAL_TRACKS parameter enumerating the tracks it intends to send. The value of the parameter SHALL be formatted as follows:
+
+~~~
+INITIAL_TRACKS Track Listing {
+  Track Descriptor Length (i),
+  Track Descriptor (..),
+}
+
+INITIAL_TRACKS Parameter Value {
+  Track Listings (..) ...,
```
jordicenzano, line 584: Perhaps another name than TRACK? It seems that in this paper TRACK is "A broadcast consists of one or more tracks. Each track has a type (audio, video, caption, etc)". That reminds me more to "rendition" or "lane"

Some food for thought:
- How can we send (eficiently) separate audio/caption tracks (english, spanish, ...) with this proposal?

jordicenzano (commented):
Great progress! Added a couple of comments

draft-lcurley-warp.md
```diff
@@ -474,6 +544,48 @@ The client:
 * SHOULD establish the connection in parallel which MUST use different QUIC connection.
 * SHOULD remain connected for two servers for a short period, processing segments from both in parallel.
 
+# INIT Parameters
+
+The INIT message ({{init}}) allows the peers to exchange arbitrary parameters before any media is exchanged. It is the main extensibility mechanism of Warp. The peers MUST ignore unknown parameters. TODO: describe GREASE for those.
+
+Every parameter MUST appear at most once within the INIT message. The peers SHOULD verify that and close the connection if a parameter appears more than once.
+
+The ROLE parameter is mandatory for the client. The INITIAL_TRACKS parameter is mandatory for any parry that is expected to send media. All of the other parameters are optional.
```
afrind, line 553: s/parry/party/

afrind (commented):
High level comment -- is it possible to transfer more than one piece of media at a time, each with their own TRACKs and SEGMENTs?

draft-lcurley-warp.md
kixelated: Since we're relying on WebTransport, we could also put this information in the CONNECT request (headers?).

draft-lcurley-warp.md
kixelated: Good point about the relay being able to forward headers it doesn't know about.

jordicenzano: AFAIK the way CDNs are configured (for HTTP headers) is to whitelist  / blacklist the ones you want to forward (You could do `*`), that allows to send private info between origin and relays without leaking it to client (or saving bytes to the final client, the ones that matters most)

draft-lcurley-warp.md
vasilvv: Clarified this is opened by the client. Regarding latency, as far as I can tell, there should be no difference for the delivery case, since the server can always speak as soon as it receives the version; I believe we do take an RTT for ingestion, unless we come up with something clever.

draft-lcurley-warp.md
vasilvv: Renamed to SETUP.

draft-lcurley-warp.md
vasilvv: Removed until tracks PR lands.

draft-lcurley-warp.md
vasilvv: I imagine this works the same way as HTTP: the semantics of the objects involved are the same across any protocol versions (i.e. you have things with key-value headers attached), but the wire encoding is different, and it's up to relay to figure out how to translate one into another as necessary.

draft-lcurley-warp.md
vasilvv: I agree this might be confusing. I'm using "track" since that's what currently in the draft; we should probably discuss renaming this.

draft-lcurley-warp.md
```diff
@@ -395,28 +399,105 @@ The application SHOULD use a non-zero error code to indicate a fatal error.
...
+A track descriptor may appear in both the INIT message and a subsequent TRACK message.
+
+
 # Messages
-Messages consist of a type identifier followed by contents, depending on the message type.
+Both unidirectional and bidirectional Warp streams are sequences of length-deliminated messages.
 
-TODO document the encoding
+~~~
+Warp Message {
```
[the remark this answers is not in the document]
vasilvv: Everything is TLV, except for the rare cases when there is no length and we then spend a year bikeshedding that.

draft-lcurley-warp.md
vasilvv: Resolved by removing this.

PR #47, opened 2022-11-18
Remove APP and merge HEADERS+SEGMENT

### LINKED ISSUE #28: APP message
This sounds a bit like an extension that just gets forwarded when not understood by a relay?  How does APP interact with caches, for example?  Can a single QUIC stream have SEGMENT and APP data interleaved?

Might be too vague to leave in

kixelated:
Shared between #43 and #46

Closes #28


kixelated:
Merging because I don't think this is controversial, and we have two PRs doing the same thing.

PR #48, opened 2022-11-18
Add containers section

kixelated:
Should be easy to define new containers now. Victor will make a registry in his PR (#46) instead of using the mimetype.


kixelated:
Going to merge because this is just moving stuff around, and doesn't seem controversial. Although we might need to hash out naming: `container` versus `format`.

PR #55, opened 2022-11-28
Describe Warp object model

vasilvv:
This is mostly based on the discussions we had the last week, with some terminology choices slightly adjusted by myself to make things clearer. Terminology is very much subject to discussion and further refinement.


draft-lcurley-warp.md
```diff
@@ -159,6 +159,19 @@ Track:
 
 This document uses the conventions detailed in Section 1.3 of {{!RFC9000}} when describing the binary encoding.
 
+# Model
+
+The basic element of Warp is *a media object*. A media object is a single addressable cacheable unit that may either contain a sequence of media samples, or some media-specific metadata, and may have relay-related attributes such as TTL or delivery priority associated with it. Typically a media object is a sequence of bytes, however, in some formats, it may be a sequence of datagrams. A Warp media object is similar in function to what is often referred to as "segments" or "chunks" in other media protocols; however, they are different from the traditional notion of chunks. The first key distinction is that Warp media objects are not always expected to be fully available, and thus any relays have to be able to convey partial media chunks. The second key distinction is that Warp media objects may not be fully decodable by themselves; an object will contain a description of the prerequisites if that is the case.
```
afrind, line 164: How can an object be defined as a sequence of datagrams?  It seems like it's defined as a sequence of bytes that can be conveyed by streams or datagrams, the latter being ok if you don't mind holes in your media objects.  Or am I missing something?

draft-lcurley-warp.md
```diff
@@ -159,6 +159,19 @@ Track:
...
+
+*A media feed* in Warp is a combination of *an init object* and a sequence of media objects. An init object is a format-specific self-contained description of the feed that is required to decode any media object contained within the feed, but can also be used as the metadata for feed selection. If two media feeds carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
+
+*A Warp broadcast* is a collection of multiple media feeds produced by a single origin. When subscribing to a broadcast, a peer has an option of subscribing to one, many or all media feeds within the broadcast.
+
+A Warp broadcast is globally identifiable via a URI. Within the broadcast, every media feed is identified via *a feed name*, an application-specific binary string. Within a single media feed, every media object is identified by its *sequence number*.
```
afrind, line 170: This may be a naming bikeshed, but `sequence number` implies a total ordering of media objects within a feed, but some of those media objects may overlap in timestamp for example.  Maybe just a media object identifier?

afrind, line 170: Why a binary string as opposed to an ASCII string or a number?

afrind (commented):
Overall I like this direction

draft-lcurley-warp.md
kixelated: A media object cannot have gaps. That's the point; it's the smallest decodable unit. If an object could be decoded out of order, then it should be split into multiple objects.

Media objects can be fragmented into IP packets via QUIC streams, and potentially datagrams. Either way, the media object MUST be delivered to the decoder in order.

I would say: "A media object is an ordered bitstream that is transmitted relative to other objects, based on a delivery order and list of dependencies."

Personally I would call it a "media fragment". It's similar, but makes it clear that it's part of a whole, and it's used by CMAF.

draft-lcurley-warp.md
kixelated, line 170: I really don't like the "application-specific binary string". We're obviously trying to encode data into this string with the intent that the broadcaster/viewer pre-negotiate an encoding structure, which just hurts interop.

draft-lcurley-warp.md
```diff
@@ -159,6 +159,19 @@ Track:
...
+
+As an example, consider a scenario where `example.org` hosts a simple live stream that anyone can subscribe to. That live stream would be a single Warp broadcast identified by the URL `https://example.org/livestream`. In the simplest implementation, it would provide only two media feeds with names `audio` and `video`. In more complicated scenarios, it could provide multiple video formats, with names like `360p_h264` or `1080p_av1`; those feeds would be *variants* of each other. Note that the feed names are opaque on the Warp level; if the player is not aware of the naming scheme for the media feeds in advance, it would have to request the broadcast description first.
```
kixelated, line 174: I like the idea of pre-negotiating feed identifiers, kind of like how WebRTC can pre-negotiate data channel stream IDs to avoid a round trip. It doesn't mean the identifier needs to be a string label though.

draft-lcurley-warp.md
vasilvv: Well, track names can be really application specific, in a way that we may not foresee. For instance, if we're sending combined security camera footage, I can imagine it having video feeds called "front door", "garage", etc.

draft-lcurley-warp.md
vasilvv: Would you prefer integers? That works for me (though that does not necessarily preclude people encoding information into the identifiers).

draft-lcurley-warp.md
vasilvv: Could you elaborate? Typically, I would expect people to order their objects by DTS.

kixelated: Yeah, I don't think we need a sequence number, but rather a unique ID + presentation timestamp.

afrind: What happens if a video and audio media object start at the same timestamp?

vasilvv: Those would belong to a separate feed, and the sequence numbers are unique per feed.

afrind: So there's an assumption that a given feed doesn't have duplicate timestamps?  That's probably worth adding, as well as some examples?

draft-lcurley-warp.md
vasilvv: Removed the mention of datagrams, at least for now.

draft-lcurley-warp.md
vasilvv: Changed to just an "object ID".

vasilvv: Made track IDs opaque integers.

vasilvv: Made track IDs opaque integers.

draft-lcurley-warp.md
vasilvv: Made track IDs opaque integers.

draft-lcurley-warp.md
```diff
@@ -159,6 +159,19 @@ Track:
 
 This document uses the conventions detailed in Section 1.3 of {{!RFC9000}} when describing the binary encoding.
 
+# Model
+
+The basic element of Warp is *a media object*. A media object is a single addressable cacheable unit that may either contain a sequence of media samples, or some media-specific metadata, and may have relay-related attributes such as TTL or delivery priority associated with it. A media object is a sequence of bytes with a fixed length. A Warp media object is similar in function to what is often referred to as "segments" or "chunks" in other media protocols; however, they are different from the traditional notion of chunks. The first key distinction is that Warp media objects are not always expected to be fully available, and thus any relays have to be able to convey partial media chunks. The second key distinction is that Warp media objects may not be fully decodable by themselves; an object will contain a description of the prerequisites if that is the case.
```
afrind, line 164: > fixed length

Roberto just mentioned on list about appendable objects, so maybe this is fungible.  But for now we can keep it.

afrind, line 164: > convey partial media chunks

media objects?

Does this mean that a relay MUST be able to forward stream data out of order if a media object is sent on a QUIC stream?

afrind (commented):
Looks good with a couple minor questions.

draft-lcurley-warp.md
vasilvv: I think that was regarding the manifests, not the objects themselves.

vasilvv: No, but if it has the first N bytes of the object, it has to forward them without waiting for the rest.  That way, even if you decide to send your media by sending a large GOP per object, as long as the encoder actually flushes frames onto the wire, they will relayed ASAP to the client.

afrind: That thread also suggested manifests are themselves objects.  But if there's not consensus there, this is fine for now, and we can always change later.

afrind: > Warp media objects are not always expected to be fully available, and thus any relays have to be able to convey partial media chunks.

Should we change this to say:

> Warp media objects can be incomplete, and thus any relays have to be able to convey the beginning of a media object without waiting for the remainder.

afrind: Or maybe use the RFC 6919 normative "REALLY OUGHT TO"

kixelated (commented):
I like it; just some nits.

draft-lcurley-warp.md
```diff
@@ -159,6 +159,19 @@ Track:
 
 This document uses the conventions detailed in Section 1.3 of {{!RFC9000}} when describing the binary encoding.
 
+# Model
+
+The basic element of Warp is *a media object*. A media object is a single addressable cacheable unit that may either contain a sequence of media samples, or some media-specific metadata, and may have relay-related attributes such as TTL or delivery priority associated with it. A media object is a sequence of bytes with a fixed length. A Warp media object is similar in function to what is often referred to as "segments" or "chunks" in other media protocols; however, they are different from the traditional notion of chunks. The first key distinction is that Warp media objects are not always expected to be fully available, and thus any relays have to be able to convey partial media objects. The second key distinction is that Warp media objects may not be fully decodable by themselves; an object will contain a description of the prerequisites if that is the case.
```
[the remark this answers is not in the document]
VMatrix1900: Sometimes the length is known at the beginning. The media encoder produces the media object at once and then handle it to the transport, instead of writing the object into the stream?

draft-lcurley-warp.md
```diff
@@ -159,6 +159,19 @@ Track:
...
+
+*A media track* in Warp is a combination of *an init object* and a sequence of media objects. An init object is a format-specific self-contained description of the track that is required to decode any media object contained within the track, but can also be used as the metadata for track selection. If two media tracks carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
+
+*A Warp broadcast* is a collection of multiple media tracks produced by a single origin. When subscribing to a broadcast, a peer has an option of subscribing to one, many or all media tracks within the broadcast.
+
+A Warp broadcast is globally identifiable via a URI. Within the broadcast, every media track is identified via *a track ID*, an unsigned 62-bit integer unique within the broadcast. Within a single media track, every media object is identified by an *object ID*, an unsigned 62-bit integer unique within the track.
```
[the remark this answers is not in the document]
vasilvv: Removed.

draft-lcurley-warp.md
```diff
@@ -159,6 +159,19 @@ Track:
...
+
+Depending on the profile of the application using it, Warp supports both a mode of operation where the peer unilaterally sends a broadcast with the media tracks of its choice, and a mode where the peer has to explicitly subscribe to a broadcast and select media tracks it wishes to receive. In the latter case, the peer requesting a subscription provides a URI identifying the broadcast, as well as the list of the names of tracks it wishes to subscribe to; if the names of the tracks are not known to the client in advance, it may request the list of all tracks available within the broadcast (which is known as *the broadcast description*).
```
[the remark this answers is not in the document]
vasilvv: I removed the second sentence here.  What I'm trying to say is that in some cases, a peer can pub without prompting (ingestion), and in some cases it needs a sub request before sending anything (delivery).

afrind (approved):
lgtm

PR #56, opened 2022-11-28
Terminology draft.

kixelated:
Still not sure if this is the best format, or if we should combine it with the other draft.


afrind:
Would it be better to include the terminology and definitions in the base protocol draft or have a separate document?  This list is not so long that it would bloat the core protocol draft.  I think the downside might be that other drafts might have to include the protocol as an informative reference?

kixelated:
I don't think this is useful by itself.

PR #61, opened 2023-01-04
s/segment/object

kixelated:


kixelated:
Going to merge because these are non-controversial naming changes. I still like "fragments" instead of "objects" but we'll litigate that in the WG.

PR #62, opened 2023-01-09
Remove the dependency list.

kixelated:
It's difficult to populate and has fringe benefit at best. Let's revisit this in the future if it's actually useful.

removed after discussion in #58

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -251,7 +251,6 @@ Warp works by splitting media into objects that can be transferred over QUIC str
 
 * The encoder determines how to fragment the encoded bitstream into objects ({{media}}).
 * Objects are assigned an intended delivery order that should be obeyed during congestion ({{delivery-order}})
-* Objects can be dependent on other objects, in which case reordering is required ({{dependencies}}).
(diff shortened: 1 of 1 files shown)


kixelated:
To explain a bit more in-depth:

The goal of the dependency list was to give relays the information required to know which frames are droppable. This doesn't actually accomplish that goal since a frame could be a _reference_ for a future frame. It's a lot of state to keep track of and it's unlikely to be used by even the smartest relay. It's much simpler and probably a better user experience for a relay to obey the delivery order.

The decoder doesn't actually need this list since the references are stored in the codec itself. A mismatch between the Warp dependency list and the codec reference list sounds pretty awful. Accurately populating the Warp dependency list requires parsing the codec bitstream, which is more difficult and expensive than it needs to be.

The only immediate use-case for the dependency list is to mark init segments. We could have a separate field for that, or put the init segment in the CATALOG itself. There's a possible use-case for SVC but again, a relay can just follow delivery order and the decoder _should_ be able to piece together dependencies from the underlying codec.

kixelated:
I went ahead and merged because I haven't heard a positive reaction to the dependency list. We can always make a PR to add it back and actually discuss the merits.

Issue #68, opened 2023-01-22
Data Model needs to expand to support different use-cases


suhasHere:
The currently proposed data model in -03 is a good start, but it does seem to lack things needed to various use-cases as defined in (https://datatracker.ietf.org/wg/moq/about/)

  -  querying/subscribing to data from Relay at various granularities in a application agnostic way 
  -  represent different forms of  media objects grouping - gops, slices,  audio groups, something new
  -   describe dependencies between objects within a group enabling relays appropriate caching strategies.

vr000m:
Does this issue also cover one publisher per broadcast or is that another issue?

afrind:
Does the now-merged model of track/group/object resolve this issue?

afrind:
Data model has been updated

Issue #57, opened 2022-12-27
Slice as segment


ZmGorynych:
In some ultra low latency use cases a video frame is encoded in multiple slices (which are independently decodable complete NAL units) and the upper slices are ready earlier than the ones on the bottom. In case of gradual decoder refresh, some slices may be intra-coded and some can be inter-coded.
A frequent optimization is to send slices as they are ready as opposed to waiting till the end of the frame. It looks like this use case is not supported in the draft

suhasHere:
@ZmGorynych  I have also been thinking around some of these application specific grouping in next generation use-cases like (AR/VR) and have been playing around with  a data model that might help in this situation .. 

I will get back here once I have something written

kixelated:
Yeah, I think breaking the media into slices will eventually be necessary. I don't have much experience with gradual decoder refresh but it makes a ton of sense for smoothing out IDR bitrate spike.

acbegen:
I suppose the subject should be "slice as a stream" and this will certainly make sense in certain low-latency apps, especially in interactive ones.

suhasHere:
@ZmGorynych, does the proposed update help  #69  . I have seem many low latency use-cases needing sort of grouping this issue raises

huitema:
I think it would be just fine to encode each slice as a separate object, and use metadata to manage transmission order, priorities, etc. For example, upper slices would be marked at higher priority than slices with fine details, and would be sent first. In case of congestion, relays might drop the details.

fluffy:
I agree an applications should be able to choose to do an object per slice.  If a slice is lost, some video devices want to display the rest of the frame that was received. Some application choose not to show the whole frame if any part of it was not received. I think that should be an application choice and it would probably change how the application decided to to put slices or full frames into single objects.

suhasHere:
moq-usages draft address some of the discussions around such an encoding. Agree with @ianswett , this is application defined and not specific to MOQT.

Issue #52, opened 2022-11-23
Subscribe with time offset


suhasHere:
It would be good if a subscriber can specify he is interested to in the media being sent starting from a given time offset

For the first phase, we can keep it focussed on time offsets that is closer to live edge and may be expand eventually to support DVR like long running time offsets.

kixelated:
side note; I'd like to see this use the presentation timestamp instead of the GoP sequence number. It should be possible to use an absolute timestamp (ex. 3:12) and a relative timestamp (ex. 5s before live).

afrind:
@suhasHere : this is kind of an old issue, has it beed overtaken by events?  The current draft allows for subscribing against track, track+group, or track+object+group.  There is also a generic Track Parameters struct in the SUBSCRIBE REQUEST that would allow for extensions, such as querying by Timestamp *if* the publisher supported that.

PR #69, opened 2023-01-22
Updated Data Model 

suhasHere:
This PR attempts to address #68

the goal of this PR is to
- Define terminologies that is generic (hopefully with less ambiguities) across application domains
- Define relationships between different resources that make up  the data model 
- Specify interactions between entities of the MoQ architecture when dealing with the data model

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -353,7 +353,7 @@ To that extent, the proposed data model enables following ways for the subscribe
        receive everything being published and all future media objects under a 
        given representation.
 
-    -  Request media to be delivered everything in group currenly being published and 
+    -  Request media to be delivered everything in group currently being published and 
        all future media objects starting from  "a specific group in a WarpStream" 
        thus allowing subscribers to control a time scale closer the beginning 
        of a given group, say beginning of a GOP sequence.
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -178,9 +178,15 @@ x (b):
                                                                                                       
 # Data Model - MediaSessions, MediaStream, MediaGroups and MediaObjects {#data-model}
 
->Note: This section expands on the currently specified model  by
-adding further elements that are relay friendly and defining generic terminology 
-that is applicable across multiple application domains envisioned by the MoQ Working Group.
+>Note: This section auguments the currently specified model by
+- adding terminology that can be generically applicable across multiple MOQ application domains
+- Clarifying the relationships between the entities of the data model
+- Definining object grouping that enables variety of use-cases (low latency publish and retrieval,
+   allowing for alternate forms of grouping (slices, tiles in a VR application) and so on
+- Allow Relays to operate on data model independent of application logic 
+
+>Note: Once the data model concepts are discussed, the PR will apply the decisions to the
+       rest of the document as appropriate
 
 This section defines a data model that expresses the relationship between various WARP 
 resources for realizing the client (publishers and subscribers) and relay functionality
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -178,7 +178,7 @@ x (b):
                                                                                                       
 # Data Model - MediaSessions, MediaStream, MediaGroups and MediaObjects {#data-model}
 
->Note: This section auguments the currently specified model by
+>Note: This section address some of the shortcomings with the currently specified model by
 - adding terminology that can be generically applicable across multiple MOQ application domains
 - Clarifying the relationships between the entities of the data model
 - Definining object grouping that enables variety of use-cases (low latency publish and retrieval,
@@ -288,7 +288,8 @@ and controlled. Some examples of how this grouping might be defined:
 * Each video frame boundary is mapped to a MediaGroup. 
   There would be a multiple MediaObjects in each mediaGroup, each containing a slice of a video frame. 
 
-* A single MediaGroup is mapped to the entire GOP sequence spanning the lifetime of the MediaStream. Each Media Object contains a slice of that media Group. 
+* A single MediaGroup is mapped to the entire GOP sequence spanning the lifetime of the MediaStream. Each 
+Media Object contains a slice of that media Group. 
 
 * Each audio frame is mapped to a MediaGroup. In this grouping, each 
    MediaGroup has a single audio frame as the MediaObject.
@@ -305,7 +306,8 @@ Examples of valid Group IDs
 ## Media Object
 
 Media Objects are the children of MediaGroups and carry a binary payload. Payload examples would include encoded and 
-encrypted media data, or caption data or the acceleration data of a racing car. Media Objects have associated header/metadata that is authenticated (but not end-to-end encrypted). The metadata contains priority/delivery order, 
+encrypted media data, or caption data or the acceleration data of a racing car. Media Objects have associated 
+header/metadata that is authenticated (but not end-to-end encrypted). The metadata contains priority/delivery order, 
 time to live, and other information aiding the caching/forwarding decision at the Relays. 
 
 Each MediaObject is identified by a sequentially increasing integer, called MediaObjectId, starting at 0. 
@@ -316,7 +318,8 @@ caching and retrieval, a Media Object is fully identified by the combination of 
 `
 Object cache key :=   MediaSessionID | MediaStreamId | MediaGroupId | MediaObjectId
 `
-These identitiers are stored as separate fields in the header of each MediaObject, to facilitate unambiguous extraction by each relay. 
+These identitiers are stored as separate fields in the header of each MediaObject, to facilitate unambiguous 
+extraction by each relay. 
 
 ## Scope / Goals of the Data Model
 
@@ -331,15 +334,18 @@ To that extent, the proposed data model enables following ways for the subscribe
 (end-points or Relays) to ask for the data:
 
     -  Request a specific MediaSession.  Such a request enables subscribers to 
-       receive all future mediastreams,  mediagroups and mediaobjects under a 
+       receive all future MediaStreams, MediaGroups and MediaoOjects under a 
        given MediaSession.
     -  Request a specific MediaStream. Such a request enables subscribers to 
-       receive all future media groups and objects under a 
+       receive all future MediaGroups and MediaObjects under a 
        given MediaStream.
     -  Request a specific MediaStream starting at a specific MediaGroup. Since the MediaGroup 
        identification is ordinal, subscribers may request for example content starting from 
        the last, or 3rd last, MediaGroup available in the cache, or from the first one for 
        which MediaGroupID  > N
+    - Request a specific MediaSession. Such a request enables subscribers to 
+       receive all future MediaStreams, MediaGroups and MediaObjects under a 
+       given MediaSession.
        
 The MediSession ID to given to each subscriber via an out-of-band mechanism. The subscriber uses this 
 MediaSessionID to request the CATALOG, which in turn provides information to the client about the 
@@ -409,7 +415,9 @@ pub: acme.tv/broadcasts/channel8/alice/4k            sub: acme.tv/brodcasts/chan
 
 ~~~~                                                                                                                             
   
-The relay does not intercept and parse the CATALOG messages, therefore it does not know the entireity of the content being produced by Alice. It simply aggregates and forwards all subscription requests that it receives. 
+The relay does not intercept and parse the CATALOG messages, therefore it does not know the entireity of 
+the content being produced by Alice. It simply aggregates and forwards all subscription requests that 
+it receives. 
 
 Similarly, below example shows an Interactive media session 
 
@@ -465,7 +473,8 @@ Publishers generate media objects within a MediaSession. A
 publisher starts by obtaining a `MediaSessionId` via an out-of-band 
 mechanism such as a session bootstrap with the Origin server.
 
-Each MediaObject produced by a publisher contains header fields which identify the MediaSession and MediaStream to which it belongs. 
+Each MediaObject produced by a publisher contains header fields which identify the MediaSession 
+and MediaStream to which it belongs. 
 
 ## Terminology Mappings
 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -176,42 +176,42 @@ x (b):
 
 
                                                                                                       
-# Data Model - Session, Stream, Groups and Objects {#data-model}
+# Data Model - MediaSessions, MediaStream, MediaGroups and MediaObjects {#data-model}
 
 >Note: This section expands on the currently specified model  by
 adding further elements that are relay friendly and defining generic terminology 
-that is applicable across multiple application domans envisioned by MoQ Working Group.
+that is applicable across multiple application domains envisioned by the MoQ Working Group.
 
-This section defines data model that expresses the relationship between various WARP 
+This section defines a data model that expresses the relationship between various WARP 
 resources for realizing the client (publishers and subscribers) and relay functionality
 for executing the protocol defined in this specification. This section doesn't define 
 mapping of the proposed data model to its transport mappings, but capture the entities 
-that form core part of WARP media delivery.                                                                                            
+that form a core part of WARP media delivery.                                                                                            
 
 
 ~~~~                                                                                          
                     ┌────────────────────────┐                                                
-                    │     WarpMediaSession   │                                                
+                    │     MediaSession       │                                                
                     └────────────────────────┘                                                
                       │                                                                       
                       │                                                                       
-                      │                                         WarpMediaStream                
+                      │                                                        
    ┌─────────────────┐│                                                                       
-   │Representation-1 │├───┬───┬──┬───┬─────────────┬───┬────────────────▶  t                  
+   │MediaStream-1     │├───┬───┬──┬───┬─────────────┬───┬────────────────▶  t                  
    └─────────────────┘│   │G1 │  │G2 │  ◉  ◉  ◉    │GN │                                      
                       │   └───┘  └───┘             └───┘                                      
                       │     ◉      ◉                 ◉                                        
                       │     ◉      ◉                                                          
-                      │                             WarpMediaStream                            
+                      │                                                       
    ┌─────────────────┐│                                                                       
-   │Representation-2 │├───┬───┬─────┬───┬────┬───┬──────▶  t                                  
+   │MediaStream-2     │├───┬───┬─────┬───┬────┬───┬──────▶  t                                  
    └─────────────────┘│   │G1 │     │G2 │    │G3 │                                            
                       │   └─┬─┘     └─┬─┘    └─┬─┘                                            
                       │     │         │        │                                              
                       │   ┌─┴─┐     ┌─┴─┐    ┌─┴─┐                                            
-                      │   │O1 │     │O1 │    │O1 │                          WarpMediaStream    
+                      │   │O1 │     │O1 │    │O1 │                             
    ┌─────────────────┐│   └───┘     └───┘    └───┘                                            
-   │Representation-3 ││─────┬──────────┬───────┬──────────┬────────────┬──────────┬───▶   t   
+   │MediaStream-3     ││─────┬──────────┬───────┬──────────┬────────────┬──────────┬───▶   t   
    └─────────────────┘│     │ Group-1  │       │ Group-2  │   ◉  ◉  ◉  │ Group-N  │           
                       │     └────┬─────┘       └─────┬────┘            └─────┬────┘           
                       │          │                   │                       │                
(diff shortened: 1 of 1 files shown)


fluffy:
This looked like a good generic starting point to me.

draft-lcurley-warp.md
```diff
@@ -175,19 +175,345 @@ x (b):
...
+`
+MediaObjectId :=   WarpStreamId  | GroupId | ObjectId
+`
+
+## Scope / Goals of the Data Model
+
+### Subscribers
+
+One of the goals of the proposed data model is to facilitate subscribers, 
+within a WarpMediSession, to request the right granularity of the media
```
VMatrix1900, line 345:
```suggestion
within a WarpMediaSession, to request the right granularity of the media 
```

draft-lcurley-warp.md
```diff
@@ -175,19 +175,345 @@ x (b):
...
+given transport mapping.
+
+To that extent, the proposed data model enables following ways for the subscribers 
+(end-points or Relays) to ask for the data:
+
+    -  Request a specific WarpStream. Such a request enables subscribers to 
+       receive everything being published and all future media objects under a 
+       given representation.
+
+    -  Request media to be delivered everything in group currenly being published and
```
VMatrix1900, line 356:
```suggestion
    -  Request media to be delivered everything in group currently being published and 
```

draft-lcurley-warp.md
```diff
@@ -175,19 +175,345 @@ x (b):
...
+                                                      sub: acme.tv/brodcasts/channel8/alice/sd/group12                                                                          .    
+                                                               ─────.                               
+                                                      ┌──────(  S1   )                              
+                                                      │       `─────'                               
+                                                      │                                             
+            sub: acme.tv/broadcasts/channel8          │   
+┌──────────────┐                ┌──────────────┐      │   
+│              │                │              │      │  sub: acme.tv/brodcasts/channel8/alice/
+|              |                |              |      |       4k        
+│   Ingest     │ ◀──────────────┤  Relay-Edge  │◀─────┘       .─────.
```
VMatrix1900, line 415: Is Ingest a relay? Is Ingest a publisher side Relay-Edge? By using Relay-edge, are you implying other types of relay such as Relay-Origin/Relay-Central?

draft-lcurley-warp.md
```diff
@@ -175,19 +175,345 @@ x (b):
...
+session or an Interactive conference meeting session or some other applicaiton
+defined session semantics.
+
+`
+Example WarpMediaSession URL for a broadcast 
+  acme.tv/broadcasts/channel8/
+  acme.tv/broadcasts/channel8/
+`
+
+## Media Representation
```
VMatrix1900, line 251: What is the relationship between the Media Representation and WarpStream? Do they map 1:1? Looks like the Media Representation URI is a verbose ID corresponding to WarpStreamID.

vasilvv:
This PR has two parts; I will go over each part individually.

------

The first part is a rewrite of the existing model section that replaces the broadcast/track/object model with a five-level model.  Since the PR does not diff easily, I am going to attempt to summarize the changes while trying to express my opinion on them one-by-one.

1. The old concept of "broadcast" has been renamed into WarpMediaSession.  I think I like the term "media session" (though I'm not so sure about using CamelCase for all the terms).
1. What was formerly known as "tracks" is now split into "media representation" and "WARP stream".  Those appear to be in one-to-one correspondence, though I do not understand what is the actual difference between those two.  I can't say I am particularly fond of the term "representation", and I object to use of word "stream" for anything that is not a QUIC stream, since that makes things confusing.  If the issue is that "tracks" are too specific to live media, we can use "media feeds", which I think was in the original model PR.
1. A layer of "object groups" is added that reflects the dependencies between objects.  I am not sure an extra level here is useful compared to the previous approach of just annotating the dependency properties directly in the objects themselves, but I do agree that we should have something of this nature.
1. The media objects are still referred to as media objects, though there is some text about encryption and authentication of those.  I don't believe that text belongs in the model, since encryption is fundamentally an issue specific to a particular media format.

------

The second part of this PR is the "Scope / Goals of the Data Model" section that introduces the notion of publishers, subscribers and relays into the draft.  I don't believe this should be a part of the model by itself (it is probably more appropriate to the section describing how the relays work), for two reasons.

The first reason is that this is not applicable to all use cases. For example, it focuses on the notion that the media receivers need to subscribe to things before receiving them, but there are plenty of cases where one side can just open a connection and start sending media (live ingestion, streaming media from a laptop to a smart device, etc). It also suggests a handshake mechanism between the publisher and the origin server, which is not a distinction that has to exist in all scenarios.

The second reason is that I believe that the fact that the media is being relayed should be mostly opaque to the endpoints producing and consuming media.  Setting authentication question aside, a generic HLS player can play media based on a playlist URL alone, not knowing where it's served from (this applies to HTTP in general); similarly, a generic RTMP ingestion client can send media to an RTMP server based on a server URL (plus auth key) alone, without caring what happens next to it.  As player and ingestion client vendors are likely going to constitute the majority of the target audience of this document long-term, I believe the model section should minimize description of parts that are not immediately relevant to them.

--------

My general issue while reviewing this PR is that it tries to address a lot of problems at once, and since it does not explain exactly what are the specific problems it is trying to address, I mostly have to infer those myself.  It would be helpful if the issue #68 specifically spelled out what the issues are; as in, what is the use case affected, what is needed for that specific use case, and why the current model does not allow that to happen.

draft-lcurley-warp.md
suhasHere: @VMatrix1900 .. Ingest is a publisher in this context and by Relay-Edge , i meant to indicate the relay at the edge. It might be better to term it as just the Relay as you have noticed

draft-lcurley-warp.md
suhasHere: Media Representation is application level concept and Warp Stream was the protocol level idea where it represent a flow of media objects for a given representation. The separation was to emphasize the fact that each application has their own way to represent such representations but at the MOQ protocol level, they are uniformly called Warp Streams ..

suhasHere:
Responding to  _A layer of "object groups" is added that reflects the dependencies between objects. I am not sure an extra level here is useful compared to the previous approach of just annotating the dependency properties directly in the objects themselves, but I do agree that we should have something of this nature._

Applications group media objects - like WARP does in one group per gop, RUSH does one video frame, some other examples where they group think based on the VR tiles and so on. The dependency relation is application defined and is implicit when an application puts one or more objects in a group.  There is no explicit dependency list carried and is not needed as well.
But having a way to identify group boundaries will help subscribers ask the media at the right granularity and for relays to provide the media objects without having to know about all possible application dependencies. 
The PR has few examples of such grouping and the model is flexible to extend for new class of applications that are evolving in this space.

suhasHere:
Responding to _What was formerly known as "tracks" is now split into "media representation" and "WARP stream". Those appear to be in one-to-one correspondence, though I do not understand what is the actual difference between those two. I can't say I am particularly fond of the term "representation", and I object to use of word "stream" for anything that is not a QUIC stream, since that makes things confusing. If the issue is that "tracks" are too specific to live media, we can use "media feeds", which I think was in the original model PR._

The name tracks is overloaded, representation is somewhere in between (DASH uses it) , Stream for sure is unusable, but not WarpStream. I am not too attached to names, as long as the names are generic to MoQ applications and not just one class of MoQ applications.

suhasHere:
Responding to _The second reason is that I believe that the fact that the media is being relayed should be mostly opaque to the endpoints producing and consuming media. Setting authentication question aside, a generic HLS player can play media based on a playlist URL alone, not knowing where it's served from (this applies to HTTP in general); similarly, a generic RTMP ingestion client can send media to an RTMP server based on a server URL (plus auth key) alone, without caring what happens next to it. As player and ingestion client vendors are likely going to constitute the majority of the target audience of this document long-term, I believe the model section should minimize description of parts that are not immediately relevant to them._

I agree and disagree on the points raised here.  Relays are part of the MoQ architecture. some applications might be totally fine with having no relays but that's not the case for all the applications. Relays might be optional in some architectures and application combinations , in which case the corresponding section of the draft doesn't apply but for the cases where Relays play a role, there is definitely a need for the Relays to understand both the protocol and the data model. This PR addresses that fact.

suhasHere:
Responding to _ As player and ingestion client vendors are likely going to constitute the majority of the target audience of this document long-term, I believe the model section should minimize description of parts that are not immediately relevant to them._

That's not my read of the charter.

draft-lcurley-warp.md
VMatrix1900: Is the Media Session application level concept or protocol level concept? By data model are we talking about the application level concept or protocol level concept or both?

vasilvv:
> Relays are part of the MoQ architecture. some applications might be totally fine with having no relays but that's not the case for all the applications. Relays might be optional in some architectures and application combinations , in which case the corresponding section of the draft doesn't apply but for the cases where Relays play a role, there is definitely a need for the Relays to understand both the protocol and the data model. 

I think we should be careful to distinguish between the MoQ data model and the MoQ architecture.

Let me explain what I mean here, using HTTP as an example.  HTTP has a fairly simple data model (the protocol exchanges request/response messages that have a method, a target/status code, some header/trailer fields and a body).  It has three wire protocols, all of which are at their heart, just an expression of that data model.  But I don't think there really is such a thing as a single "HTTP architecture".  HTTP is used in an extremely wide range of scenarios, that include direct connections, a wide array of intermediaries, caches, etc.  RFC 9110 acknowledges a lot of them, provides some mechanisms that are only applicable to specific roles (e.g. Section 7.6 describes mechanisms specific to intermediaries), but ultimately, rather than defining a unified "HTTP architecture", it merely describes the common approaches and how they fit together.

Similarly, when talking about MoQ, rather than thinking about a single MoQ architecture, I think our work here mostly revolves around creating the building blocks that one can architect the solution they need.  More specifically, we should:
1. Figure out what are the things being exchanged in the protocol (objects/tracks/broadcasts/sessions/object groups/etc) -- this is the *data model*.
1. Provide some common methods of exchanging those objects (direct push, subscriptions) -- this is also the part of the data model, though there is more flexibility around some parts of this being optional.
1. Enumerate the ways in which we expect those to be used, explaining how you build systems made of boxes that speak MoQ, and what those boxes would do with the data items and methods defined above.

Note that in this scenario, everyone has to understand and implement the first part, but they can only implement the relevant methods and boxes for the architecture they care about.  The only people who have to care about all the possible boxes are people designing the protocol (i.e. us).

(it's 4am on Saturday, so this is me mostly writing my thoughts on this down before I forget -- I'll almost certainly need to elaborate more at the interim)

**TL;DR** "what does MoQ do?" and "what does one do with MoQ?" are two different questions, those should be described in different sections, and relays are mostly an answer to the second question

kixelated (commented):
I agree with Victor. It's very difficult to figure out what you want to address and why. Could you try to break it up into smaller issues/PRs for the sake of discussion?

My best guess is:

1. Rename "broadcast" to "WarpMediaSession".
2. Add "Representation", which is member of a broadcast and contains a single track...?
3. Rename "track" to "WarpMediaStream"?
4. Add "WarpObjectsGroup", which is a member of a track and contains objects.
5. Require a globally unique broadcast URL. #70 
6. Add the ability to use an object per slice. #57 
7. Add a sequence number for objects.
8. Add the ability to subscribe starting at a GOP boundary. #52 kinda?
9. Add the ability to subscribe starting at a specific frame?

suhasHere:
> > Relays are part of the MoQ architecture. some applications might be totally fine with having no relays but that's not the case for all the applications. Relays might be optional in some architectures and application combinations , in which case the corresponding section of the draft doesn't apply but for the cases where Relays play a role, there is definitely a need for the Relays to understand both the protocol and the data model.
> 
> I think we should be careful to distinguish between the MoQ data model and the MoQ architecture.
> 
> Let me explain what I mean here, using HTTP as an example. HTTP has a fairly simple data model (the protocol exchanges request/response messages that have a method, a target/status code, some header/trailer fields and a body). It has three wire protocols, all of which are at their heart, just an expression of that data model. But I don't think there really is such a thing as a single "HTTP architecture". HTTP is used in an extremely wide range of scenarios, that include direct connections, a wide array of intermediaries, caches, etc. RFC 9110 acknowledges a lot of them, provides some mechanisms that are only applicable to specific roles (e.g. Section 7.6 describes mechanisms specific to intermediaries), but ultimately, rather than defining a unified "HTTP architecture", it merely describes the common approaches and how they fit together.
> 
> Similarly, when talking about MoQ, rather than thinking about a single MoQ architecture, I think our work here mostly revolves around creating the building blocks that one can architect the solution they need. More specifically, we should:
> 
> 1. Figure out what are the things being exchanged in the protocol (objects/tracks/broadcasts/sessions/object groups/etc) -- this is the _data model_.
> 2. Provide some common methods of exchanging those objects (direct push, subscriptions) -- this is also the part of the data model, though there is more flexibility around some parts of this being optional.
> 3. Enumerate the ways in which we expect those to be used, explaining how you build systems made of boxes that speak MoQ, and what those boxes would do with the data items and methods defined above.
> 
> Note that in this scenario, everyone has to understand and implement the first part, but they can only implement the relevant methods and boxes for the architecture they care about. The only people who have to care about all the possible boxes are people designing the protocol (i.e. us).
> 
> (it's 4am on Saturday, so this is me mostly writing my thoughts on this down before I forget -- I'll almost certainly need to elaborate more at the interim)
> 
> **TL;DR** "what does MoQ do?" and "what does one do with MoQ?" are two different questions, those should be described in different sections, and relays are mostly an answer to the second question

The protocol draft is supposed to talk about protocol elements, data model and how the entities (publishers/subscribers/relays) interact to understand the protocol. If we do it in one section or 3 sections , it is up to us , but the spec should address these things to align with the charter. If we don't thjnk that's the case, may be good thing to raise it to the WG ?

draft-lcurley-warp.md
```diff
@@ -175,19 +175,345 @@ x (b):
...
+
+* Each audio frame is mapped to a WarpObjectGroup. In this grouping, each 
+   WarpObjectGroup has a single audio frame as the WarpMediaObject.
+
+
+## WarpMediaObject/MoqMediaObject/Media Object
+
+Media Objects within Warp belong to a group and carry encoded and 
+encrypted media data. Each object within a group, is identified by 
+sequentially increasing integer, called ObjectId, starting at 0.
```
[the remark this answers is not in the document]
suhasHere: > I agree with Victor. It's very difficult to figure out what you want to address and why. Could you try to break it up into smaller issues/PRs for the sake of discussion?
> 
> My best guess is:
> 
> 1. Rename "broadcast" to "WarpMediaSession".
> 2. Add "Representation", which is member of a broadcast and contains a single track...?
> 3. Rename "track" to "WarpMediaStream"?
> 4. Add "WarpObjectsGroup", which is a member of a track, contains objects, and is independently decodable?
> 5. Require a globally unique broadcast URL. #70
> 6. Add the ability to use an object per slice. #57
> 7. Add a sequence number for objects.
> 8. Add the ability to subscribe starting at a GOP boundary. #52 kinda?
> 9. Add the ability to subscribe starting at a specific frame?


We need to work on a data model and it needs to be dealt in one place. Splitting across the PRs will just cause more confusions and it's hard to have discussions made coherently.

draft-lcurley-warp.md
```diff
@@ -175,19 +175,345 @@ x (b):
...
+to audio stream from alice, whereas S2 subscribers
+to all the streams being published. Relay at the Edge
+forwards the unique subscriptions to the downstream Realys,
+as needed, to setup the delivery network.
+
+
+### Publishers
+
+Publishers generate media objects with in a WarpMediaSession. A given
+publishers start by obtaining `WarpStreamId` via out-of-band
```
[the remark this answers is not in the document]
suhasHere: There is no negotiation here. can you provide details on the flow of steps that happens today with these generic clients...

draft-lcurley-warp.md
```diff
@@ -175,19 +175,345 @@ x (b):
...
+wherever applicable.
+
+A WarpMediaSession can represent a Broadcast for a live media 
+session or an Interactive conference meeting session or some other applicaiton
+defined session semantics.
+
+`
+Example WarpMediaSession URL for a broadcast 
+  acme.tv/broadcasts/channel8/
+  acme.tv/broadcasts/channel8/
```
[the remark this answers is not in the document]
suhasHere: That should have been 2 different examples. Copy paste error

draft-lcurley-warp.md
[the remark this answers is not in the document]
kixelated: I guess today publishers will get a URL. This would require a URL (for WebTransport) and another URL for the `WarpStreamId`. Actually aren't there multiple `WarpStreamIds`? Do you need to pre-negotiate one for every rendition?

draft-lcurley-warp.md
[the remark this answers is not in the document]
kixelated: I want to discuss each of these separately. 

A few of them are good changes (6, 7, 8). Some of the changes are up for debate and deserve their own issue, like (1, 4, 5, 9). Some of the changes are just questionable (2, 3) and really just drag the rest of the PR down. 

But you absolutely don't need to rewrite the object model. It's perfectly acceptable to create a PR that simultaneously amends the object model while adding the corresponding wire format. I really don't think there's a huge discrepancy between the current object model and the one that you're proposing.

suhasHere:
Responding to _As player and ingestion client vendors are likely going to constitute the majority of the target audience of this document long-term, I believe the model section should minimize description of parts that are not immediately relevant to them_

This might be a good thing to bring to the WG and the list ..

draft-lcurley-warp.md
suhasHere: Here is what I had in mind, do let me know if it helps or should we adjust it differently.

Each MoQ application will have their own way to identify publisher media capabilities that list media types, their quality attributes (codec, bitrates and so on). They are typically some sort of textual descriptions. Then for each of those there will be stream of grouped media objects that is actually sent.

I totally understand the reason why this might cause confusion. It might be better served, it we can just say 

```
Warp Media Streams identify media objects belonging to a publisher's media capability/representation. A given Media Representation, in an application defined way,  identifies capabilities of a publisher's media source(s) within a WarpMediaSession. Representations identifies things such as media source type, codec info, bitrate and other attribute qualities.

```

Would this work ?

kixelated: Yeah that won't work. We need to support generic clients. #71

Any information required to broadcast or view needs to be part of the protocol. That can be as simple as specifying container/manifest formats with enough information, but it cannot be left up to the application.

draft-lcurley-warp.md
```diff
@@ -175,19 +175,329 @@ x (b):
...
+       receive all future MediaGroups and MediaObjects under a 
+       given MediaStream.
+    -  Request a specific MediaStream starting at a specific MediaGroup. Since the MediaGroup 
+       identification is ordinal, subscribers may request for example content starting from 
+       the last, or 3rd last, MediaGroup available in the cache, or from the first one for 
+       which MediaGroupID  > N
+    - Request a specific MediaSession. Such a request enables subscribers to 
+       receive all future MediaStreams, MediaGroups and MediaObjects under a 
+       given MediaSession.
+
```
afrind, line 349: This text appears duplicated with 336-338

draft-lcurley-warp.md
```diff
@@ -175,19 +175,329 @@ x (b):
...
+       identification is ordinal, subscribers may request for example content starting from 
+       the last, or 3rd last, MediaGroup available in the cache, or from the first one for 
+       which MediaGroupID  > N
+    - Request a specific MediaSession. Such a request enables subscribers to 
+       receive all future MediaStreams, MediaGroups and MediaObjects under a 
+       given MediaSession.
+       
+The MediSession ID to given to each subscriber via an out-of-band mechanism. The subscriber uses this 
+MediaSessionID to request the CATALOG, which in turn provides information to the client about the 
+MediaStreamIDs which it can use to construct its subscription requests.
```
afrind, line 352: Why out of band?

Also, better to use publisher/consumer rather than client/server terminology, since for ingest, the client is the publisher.

draft-lcurley-warp.md
```diff
@@ -175,19 +175,329 @@ x (b):
...
+pub: acme.tv/broadcasts/channel8/alice/hd                    `─────'                              
+pub: acme.tv/broadcasts/channel8/alice/sd                                                                   
+pub: acme.tv/broadcasts/channel8/alice/4k            sub: acme.tv/brodcasts/channel8/alice/sd  
+                                                     sub: acme.tv/brodcasts/channel8/alice/hd   
+
+~~~~                                                                                                                             
+  
+The relay does not intercept and parse the CATALOG messages, therefore it does not know the entireity of 
+the content being produced by Alice. It simply aggregates and forwards all subscription requests that 
+it receives.
```
afrind, line 420: I think there are features/performance that can be unlocked by a "media gateway" that does parse the CATALOG, such as the ability for a subscriber to ask the gateway to automatically switch between qualities within a latency bound, but perhaps such a function doesn't need to be considered a "relay".

afrind:
What's the disposition of this PR in light of the newer one?  Should we close it, or do we intend to iterate on both?

afrind:
Has this been overtaken by events, such as #121?  Can we close?

PR #86, opened 2023-02-01
Describe join points

vasilvv:
This is a very basic proposal that should allow us to solve the probem of "how do I join an existing media broadcast", and, partially, "how do I seek", without complexities involved in explicitly spelling out dependencies.

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -181,6 +181,8 @@ The basic element of Warp is *a media object*. A media object is a single addres
 
 *A media track* in Warp is a combination of *an init object* and a sequence of media objects. An init object is a format-specific self-contained description of the track that is required to decode any media object contained within the track, but can also be used as the metadata for track selection. If two media tracks carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
 
+*A join point* is an annotation attached to the media object that indicates that a subscriber to a specific track can start decoding the track from that point. It can optionally contain a set of metadata (like timestamps) that can be used for seeking. An object that is annotated as a join point MUST be decodable without having access to any earlier objects.
+
 *A Warp broadcast* is a collection of multiple media tracks produced by a single origin. When subscribing to a broadcast, a peer has an option of subscribing to one, many or all media tracks within the broadcast.
 
 A Warp broadcast is globally identifiable via a URI. Within the broadcast, every media track is identified via *a track ID* that is unique within the broadcast. Within a single media track, every media object is identified by an *object ID* that is unique within the track.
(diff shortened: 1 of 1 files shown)


kixelated (changes requested):
What is a "join point"? Do you mean independently decodable?

And it should use a sequence number rather than a boolean to handle packet loss. Basically group ID from QuicR.

suhasHere:
need to think about it . this as object can be useful on its own .. May be call it as StartGroupInfo to be more generic.

This along with groupId (instead of being a boolean) will address other use-cases that are not only time based ..

kixelated:
I'm fine adding a "group sequence" like QuicR. Monotonically incrementing so you can detect gaps when doing frame-based delivery.

I'm not sure if it actually needs to be in the object model though. I feel like it should be high level instead of trying to explain how every bit of functionality works.

suhasHere:
> I'm fine adding a "group sequence" like QuicR. Monotonically incrementing so you can detect gaps when doing frame-based delivery.
> 
> I'm not sure if it actually needs to be in the object model though. I don't think the object model should explain how every minor piece of functionality works.

Adding groupId is a minor addition that will be generic across use-cases and also avoids needing to specify all possible functionalities

kixelated:
```
OBJECT Message {
  Broadcast URI (b),
  Track ID (i),
  Object ID (i),
  Object Delivery Order (i),
->Group Sequence (i),
  Object Payload (b),
}
```

suhasHere:
> ```
> OBJECT Message {
>   Broadcast URI (b),
>   Track ID (i),
>   Object ID (i),
>   Object Delivery Order (i),
> ->Group Sequence (i),
>   Object Payload (b),
> }
> ```

that would work ... to be uniform across the names in there, would something like this be OK

```
OBJECT Message {
   Broadcast URI (b),
   Track ID (i),
   Object ID (i),
 ->Group ID (i), // add text to define its a sequence 
   Object Delivery Order (i),
   Object Payload (b),
 }
 ```

**Comment by kixelated** [comment]
It has to be monotonically increasing to detect gaps, so I think "sequence"
is more accurate than ID.

**Comment by kixelated** [comment]
To elaborate, to me an ID is just unique. There's no concept of ordering between IDs. It basically an opaque blob.

A sequence on the other hand, is a number that increases over time with each occurrence. It's still unique, but conveys creation order.

"Object ID" maaaay also need to be increasing now and renamed to "Object Sequence". That's because if I want Object 123 in Group 6, I need to download every Object < 123 in Group 6.

But I suppose downloading every object in Group 6 would be fine if delivery order is maintained. I would leave it as an ID for now.

**Comment by suhasHere** [comment]
ObjectId is a sequence too. Since being part of a group announces sort of dependency (one is delivery order) and also helps not keeping objectIDs flat.
While we are at it, we should go for something like

```
OBJECT Message {
   Broadcast URI (b),
   Track ID (i),
   Object Sequence (i),
   Group Sequence(i),
   Object Delivery Order (i),
   Object Payload (b),
 }
```

We can then tackle about the issues that were raised on Broadcast URI field later.

afrind:
Has this been overtaken by groups?

fluffy (commented):
I think a better way to indicate the start of a new join point is to have the group id and a new group ID indicates , well a new group, which is what we want

kixelated:
Yeah Cullen, that's the direction we went. A boolean is more frail during packet loss compared to a sequence number or ID.

PR #89, opened 2023-02-02
Add object/group sequence

suhasHere:


draft-lcurley-warp.md
```diff
@@ -514,8 +515,11 @@ The broadcast URI as declared in CATALOG ({{message-catalog}}).
 * Track ID:
 The track identifier as declared in CATALOG ({{message-catalog}}).
 
-* Object ID:
-A unique identifier for each object within the track.
+* Group Sequence :
+An integer always starts at 0 and increases sequentially at the original media publisher.
```
vasilvv, line 519: Please indicate that group sequence numbers are scoped to the track.

suhasHere:
Merging based on in-person discussions.

kixelated:
You merged a litttttle early. The wire format is good, but the accompanying text could use some tweaks. 

For example, it's not clear when the object sequence number should be incremented. It should be done on every independently decodable OBJECT.

There's also existing references to Object ID in the document that should be updated.

I'll make another PR.

PR #90, opened 2023-02-02
Add some clarification for group sequence.

kixelated:
We might need more text but I'm about to jump on a plane.


kixelated:
Actually this PR isn't quite ready yet but for some reason I frantically pushed it. I'll amend it when I get home in a few hours.

Group Sequence should definitely be scoped to a track and I need to add text for that.

Object Sequence can either be scoped to a track or group. They encode a similar amount of information with slight differences that really don't matter. I also need to add text to clarify since I removed Suhas' text for some reason.

I'm thinking we scope Object Sequence to the group mostly for consistentency. Objects are scoped to Groups, Groups are scoped to Tracks, and Tracks and scoped to broadcasts, creating a nice hierarchy.

draft-lcurley-warp.md
```diff
@@ -515,13 +515,14 @@ The broadcast URI as declared in CATALOG ({{message-catalog}}).
 * Track ID:
 The track identifier as declared in CATALOG ({{message-catalog}}).
 
-* Group Sequence :
-An integer always starts at 0 and increases sequentially at the original media publisher.
-Group sequences are scoped under a Track.
+* Group Sequence:
+An integer indicating the OBJECT is a member of the indicated group within a track.
+A group is indepdendently decodable (ex. a GoP) and can be used as an entry point for a decoder.
+The group sequence number MUST be incremented for each indepdendently decodable OBJECT within the track.
```
afrind, line 521: Does it need to be incremented **by one** or any increment is ok?

kixelated: I think any number is fine, so long as it increases. It gives the receiver less information about the number of GoPs (which it doesn't need to know) but it lets the sender insert discontinuities. I don't think we could enforce it increasing by one anyway since OBJECTs can be lost.

afrind: Is the MUST too strong here, would MAY suffice?  eg: even if I'm kicking off a new OBJECT and its independently decodable, can I keep the same group for my own reasons?

draft-lcurley-warp.md
```diff
@@ -515,13 +515,14 @@ The broadcast URI as declared in CATALOG ({{message-catalog}}).
...
 
 * Object Sequence:
-An integer always starts at 0 with in a Group and increases sequentially.
-Object Sequences are scoped to a Group.
+A monotonically increasing integer for each object indicating the decode order within a group.
```
afrind, line 524: Do you also need to say it starts with 0 at the beginning of the group?

fluffy: I don't think I get why this is changing - the original text seems fine.

kixelated: Yeah I'll loosen the text a bit.

This PR is aimed at giving group some properties. As it currently exists, being a member of a group means absolutely nothing, and the relay can't use this information for the intended purpose.

draft-lcurley-warp.md
```diff
@@ -179,6 +179,8 @@ x (b):
 
 The basic element of Warp is *a media object*. A media object is a single addressable cacheable unit that may either contain a sequence of media samples, or some media-specific metadata, and may have relay-related attributes such as TTL or delivery priority associated with it. A media object is a sequence of bytes with a finite length. A Warp media object is similar in function to what is often referred to as "segments" or "chunks" in other media protocols; however, they are different from the traditional notion of chunks. The first key distinction is that Warp media objects are not always expected to be fully available, and thus any relays have to be able to convey partial media objects. The second key distinction is that Warp media objects may not be fully decodable by themselves; an object will contain a description of the prerequisites if that is the case.
 
+A *group* is a collection of objects. An object can depend on another objects, but only within the same group and with a smaller sequence number. This guarantees that an object is decodable if all prior members of the group have been delivered.
```
kixelated, line 182: @suhasHere Feel free to overwrite this. I wanted to give an example of how I would phrase it.

draft-lcurley-warp.md
```diff
@@ -515,18 +517,20 @@ The broadcast URI as declared in CATALOG ({{message-catalog}}).
 * Track ID:
 The track identifier as declared in CATALOG ({{message-catalog}}).
 
-* Group Sequence :
-An integer always starts at 0 and increases sequentially at the original media publisher.
-Group sequences are scoped under a Track.
+* Group ID:
+Indicates the object is a member of a group within a track.
```
afrind, line 521: Since Group ID doesn't appear optional (all objects must be in a group), Group ID indicates the group this object belongs to.

draft-lcurley-warp.md
```diff
@@ -515,18 +517,20 @@ The broadcast URI as declared in CATALOG ({{message-catalog}}).
...
+Objects within a group MUST only depend on other objects within the same group.
+The group ID is scoped to the track; multiple tracks may use the same group ID.
```
afrind, line 523: s/scoped to/must be unique within/?

draft-lcurley-warp.md
```diff
@@ -515,18 +517,20 @@ The broadcast URI as declared in CATALOG ({{message-catalog}}).
...
-An integer always starts at 0 with in a Group and increases sequentially.
-Object Sequences are scoped to a Group.
+* Group Sequence:
+A monotonically increasing integer (starting at 0) for each object within a group.
+An object MUST only depend on other objects with a smaller sequence within the group.
+This means that sequence 0 MUST be independent, while other sequences MAY depend on prior sequences.
 
-* Object Delivery Order:
+* Delivery Order:
 An integer indicating the object delivery order ({{delivery-order}}).
```
afrind, line 531: Do we need explanation of why delivery order can be different that sequence?  Is delivery order scoped to a group, track, or a session?

draft-lcurley-warp.md
kpugin: > but only within the same group and with a smaller sequence number.

what about B-frames?

suhasHere:
@kixelated  I will send the review feedback in couple of days. Thanks

afrind:
Has this been overtaken by events?

suhasHere:
> Has this been overtaken by events?

I think so.

PR #91, opened 2023-02-02
Replace "Warp session" with "Warp connection"

vasilvv:
It turns out that "session" here is ambiguous, since in some contexts, it can be interpreted as an entity that exists across multiple hops, which is not the intention here.

This keeps "WebTransport session" as-is, since that is a terminology used by the WebTransport draft.

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -90,7 +90,7 @@ Bitstream:
 
 Client:
 
-: The party initiating a Warp session.
+: The party initiating a Warp connection.
 
 Codec:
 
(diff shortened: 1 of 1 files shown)


afrind:
I don't love having a "warp connection" inside a "webtransport session", since I think that may cause other confusion. However if other folks are comfortable with it, go ahead.

kixelated:
Yeah I agree with Alan, I actually think this is slightly more confusing in the hop-to-hop case and even more confusing in the end-to-end case. A Warp connection could outlive a QUIC connection..?

I would say we first settle on if a Warp session can outlive a WebTransport session. And if so, find a new name that is not connection or session.

vasilvv:
Abandon in favor of text in #135.

PR #94, opened 2023-02-15
Data model

huitema:
This is a PR to propose changes in the data model PR.

Largely work in progress, but I wanted to first propose an organization.

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -501,7 +501,8 @@ The format of the OBJECT message is as follows:
 OBJECT Message {
   Broadcast URI (b)
   Track ID (i),
-  Object ID (i),
+  Group Sequence (i),
+  Object Sequence (i),
   Object Delivery Order (i),
   Object Payload (b),
 }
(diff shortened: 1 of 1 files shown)


afrind:
The introductory text seems clear to me.  I have questions about what it means for an object to be 'atomic' - since it can necessarily be larger than an MTU it might not get delivered.  It can be repaired of course, but what are the consequences to other objects in the group or track if the object never completely arrives?  Are there restrictions on receivers attempting to use partial objects?  This reader is excited to read the next chapter :D

huitema:
@afrind The main consequence of "atomic" is that it is way better to send one complete object than two incomplete objects. That should very much guide scheduling of transmissions at relays. The other consequence is that if publisher can split a big blob of data into several objects that can be processed independently, they should. For example, instead of sending "the entire GOB as an object", they would get better results sending the GOB as a series of objects, with each object boundary becoming a proper "cut mark". And yes, I have to develop the text. I checked in a partial PR to Suhas' PR, mostly to test my editing chain...

suhasHere:
thanks Christian for detailed text. I will merge it into my branch and do an updated version based on this.

Issue #66, opened 2023-01-20
Relax the CATALOG definition to decouple the base protocol from the streaming format 


wilaw:
The success of HTTP streaming is partly due to the fact that it is built upon ubiquitous low level API, namely HTTP/S GET, POST of binary files.  The distribution architecture (ingest, caching nodes, edge nodes, routers etc) only has to be built to support this API. On top of that can be layered many streaming protocols - such as HLS, Smooth, HDS and DASH. These streaming protocols can innovate and co-exist and also evolve their features without having to make any changes to the distribution network. The distribution network is unaware of the internal workings for the streaming formats. This decoupling allows good economics (any distribution network can deliver any streaming format), scale (because distribution networks are not specific to a streaming format) and innovation ( networks and formats can be improved independently of one another, as long as the both preserve the API interface). 

The current WARP architecture is a relatively tight binding of streaming format to pub/sub architecture, with the CATALOG message having explicit descriptors for a f4mp4 solution with init segments. This is not broad enough to address all the [use-cases](https://datatracker.ietf.org/doc/draft-gruessing-moq-requirements/) that have been defined for MoQ to address.  

One option is to keep adding explicit fields to the CATALOG message as we encounter new use-cases. An alternative approach is that MoQ define a Common Base Protocol of core APIs. These can be very simple constructs such as AUTH,PUBLISH, SUBSCRIBE, ERROR commands and message structure, essentially much of what WARP draft defines. On top of this Base Protocol we can define and operate various streaming formats. Some might use fmp4 only with a 2s delay target, others might focus on real-time communications. But they can coexists and have quite different CATALOG contents. The benefits of economics, scale and flexibility will apply to this MoQ distribution infra as they have applied to HTTP. The WARP draft 03 actually meets many of the requirements for a base protocol definition. 

[...]

PR #98, opened 2023-03-03
Merge broadcast URN with the WebTransport URL.

kixelated:
There's quite a few benefits for such a minor change:

* broadcast selection is now explicitly authorized
* prioritization works at the WebTransport level
* OBJECT headers are much smaller.
* CATALOG messages are simpler (too simple).
* OBJECT messages no longer need to be rewritten.
* Only one control stream is needed. (TODO)
* Generic clients work via a single URL now.

See the mailing list for more in-depth rationale.

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -181,9 +181,15 @@ The basic element of Warp is *a media object*. A media object is a single addres
 
 *A media track* in Warp is a combination of *an init object* and a sequence of media objects. An init object is a format-specific self-contained description of the track that is required to decode any media object contained within the track, but can also be used as the metadata for track selection. If two media tracks carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
 
-*A Warp broadcast* is a collection of multiple media tracks produced by a single source. The sender has the ability to publish one or more tracks and the receiver has the ability to subscribe to zero or more tracks. Depending on the the application, the sender may decide to unilaterally push tracks without an explicit subscription to reduce latency.
+A *track bundle* is a collection of tracks intended to be delivered together.
+Objects within a track bundle may be prioritized relative to each other via the delivery order property.
+This allows objects to be prioritized within a track (ex. newer > older) and between tracks (ex. audio > video).
+The track bundle may contain a catalog indicating the available tracks.
 
-*A WebTransport session* is established for each Warp broadcast. The client issues a CONNECT request with a URL to the server. The server parses the URL based on the application, using it to identify the broadcast and authorize the user. Multiple broadcasts require multiple WebTransport sessions, which can be pooled over a single QUIC connection for efficiency.
+A WebTransport *session* is established for each track bundle.
+The client issues a CONNECT request with a URL which the server uses for identification and authentication.
+All control messages and prioritization occur within the context of a single WebTransport session, which means a single track bundle.
+Multiple WebTransport sessions may be pooled over a single QUIC connection for efficiency.
 
 
 # Motivation
@@ -328,15 +334,15 @@ WebTransport can currently operate via HTTP/3 and HTTP/2, using QUIC or TCP unde
 As mentioned in the motivation ({{motivation}}) section, TCP introduces head-of-line blocking and will result in a worse experience.
 It is RECOMMENDED to use WebTransport over HTTP/3.
 
-### CONNECT request
-The server uses the HTTP CONNECT request to identify client and the requested broadcast.
+### CONNECT
+The server uses the HTTP CONNECT request to identify client and the requested track bundle.
 The application dictates how this information is encoded into the request.
-For example, an authentication token and broadcast ID could be included in the path.
+For example, a broadcast ID and authentication token could be included in the path.
 
 The server MAY return an error status code for any reason. Examples include:
-* "400 Bad Request" when the broadcast and/or authorization cannot be parsed.
+* "400 Bad Request" when the request cannot be parsed.
 * "401 Unauthorized" when the client cannot be identified.
-* "403 Forbidden" when the client is unauthorized to access the specified broadcast.
+* "403 Forbidden" when the client is unauthorized to access the specified resource.
 * "404 Not Found" when the broadcast cannot be identified.
 * "307 Temporary Redirect" when the client should be served at a different url.
 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -344,7 +344,6 @@ The server MAY return an error status code for any reason. Examples include:
 * "401 Unauthorized" when the client cannot be identified.
 * "403 Forbidden" when the client is unauthorized to access the specified resource.
 * "404 Not Found" when the broadcast cannot be identified.
-* "307 Temporary Redirect" when the client should be served at a different url.
 
 Otherwise the server MUST respond with a "200 OK" to establish the WebTransport session.
 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -343,7 +343,7 @@ It is RECOMMENDED to use WebTransport over HTTP/3.
 ### CONNECT
 The server uses the HTTP CONNECT request for identification and authorization of a track bundle.
 The specific mechanism is left up to the application.
-For example, a broadcast ID and authentication token could be included in the path.
+For example, an identifier and authentication token could be included in the path.
 
 The server MAY return an error status code for any reason, for example a 403 when the client is forbidden.
 Otherwise the server MUST respond with a "200 OK" to establish the WebTransport session.
@@ -565,7 +565,7 @@ Track Descriptor {
 {: #warp-track-descriptor title="Warp Track Descriptor"}
 
 * Track ID:
-A unique identifier for the track
+A unique identifier for the track within the track bundle.
 
 * Container Format:
 The container format as defined in {{containers}}.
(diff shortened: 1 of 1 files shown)


draft-lcurley-warp.md
```diff
@@ -181,13 +181,9 @@ The basic element of Warp is *a media object*. A media object is a single addres
...
-
-Depending on the profile of the application using it, Warp supports both a mode of operation where the peer unilaterally sends a broadcast with the media tracks of its choice, and a mode where the peer has to explicitly subscribe to a broadcast and select media tracks it wishes to receive.
-
-As an example, consider a scenario where `example.org` hosts a simple live stream that anyone can subscribe to. That live stream would be a single Warp broadcast identified by the URL `https://example.org/livestream`. In the simplest implementation, it would provide only two media tracks, one with audio and one with video. In more complicated scenarios, it could provide multiple video formats of different levels of video quality; those tracks would be variants of each other. Note that the track IDs are opaque on the Warp level; if the player has not received the description of media tracks out of band in advance, it would have to request the broadcast description first.
```
kixelated, line 190: I can revert this change if requested.

I felt there was too much attention to a single, latency reduction feature in the object model, so I tried to summarize it into a single sentence.

draft-lcurley-warp.md
```diff
@@ -337,15 +333,14 @@ The application SHOULD use the WebTransport CONNECT request for authentication.
...
 An endpoint SHOULD terminate the connection with an error ({{termination}}) if the peer attempts an unauthorized action.
-For example, attempting to use a different role than pre-negotated ({{role}}) or using an invalid broadcast URI ({{message-catalog}}).
+For example, attempting to use a different role than pre-negotated ({{role}}).
 
 ## Streams
 Warp endpoints communicate over QUIC streams. Every stream is a sequence of messages, framed as described in {{messages}}.
 
 The first stream opened is a client-initiated bidirectional stream where the peers exchange SETUP messages ({{message-setup}}). The subsequent streams MAY be either unidirectional and bidirectional. For exchanging media, an application would typically send a unidirectional stream containing a single OBJECT message ({{message-object}}).
 
 Messages SHOULD be sent over the same stream if ordering is desired.
```
kixelated, line 343: @suhasHere We can now replace this with a single control stream for the entire session (the same one used for SETUP?).

draft-lcurley-warp.md
```diff
@@ -532,24 +523,20 @@ This is a media bitstream intended for the decoder and SHOULD NOT be processed b
 
 
 ## CATALOG {#message-catalog}
-The sender advertises an available broadcast and its tracks via the CATALOG message.
+The sender advertises tracks via the CATALOG message.
 
 The format of the CATALOG message is as follows:
 
 ~~~
 CATALOG Message {
```
kixelated, line 531: @wilaw The CATALOG is criminally simple now. Let's do your suggestion in #66.

vasilvv:
I agree that we shouldn't really need an extra layer of pooling if we can just use WebTransport pooling for this.

One minor concern I have is that we *are* chartered to specify MoQ that works both over raw QUIC and over WT, but I guess we can just say that if you use raw QUIC, you don't get pooling (given the motivations I can anticipate for using MoQ without WT, that sounds like a fair trade-off).

Will review the text in detail later.

afrind:
I like this concept, as it pushes any "pooling" complexity up one layer to WebTransport rather than reinventing (and re-arguing) all of those details here -- I just saw "flow control" mentioned on the list.

afrind:
I apologize for not being more careful about indicating the hat I'm wearing while reviewing PRs.  Please consider my comments on this PR as an individual.

draft-lcurley-warp.md
```diff
@@ -181,13 +181,9 @@ The basic element of Warp is *a media object*. A media object is a single addres
 
 *A media track* in Warp is a combination of *an init object* and a sequence of media objects. An init object is a format-specific self-contained description of the track that is required to decode any media object contained within the track, but can also be used as the metadata for track selection. If two media tracks carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
 
-*A Warp broadcast* is a collection of multiple media tracks produced by a single origin. When subscribing to a broadcast, a peer has an option of subscribing to one, many or all media tracks within the broadcast.
+*A Warp broadcast* is a collection of multiple media tracks produced by a single source. The sender has the ability to publish one or more tracks and the receiver has the ability to subscribe to zero or more tracks. Depending on the the application, the sender may decide to unilaterally push tracks without an explicit subscription to reduce latency.
```
wilaw, line 184: > Depending on the the application, the sender may decide to unilaterally push tracks without an explicit subscription to reduce latency.

How would this work with a relay? The relay would have no incoming subscription request to match the arriving objects against and would not know what to do with the traffic.

afrind:
As an individual: 

This may be a question for WebTransport -- how are streams across different WebTransport sessions prioritized?  The WebTransport CONNECT request can have an HTTP Priority header -- one logical interpretation is that all streams belonging to that session be prioritized "under" that top level prioritization.  eg: if I have two sessions at the same urgency and incremental on, session A has 3 streams and session B has 1, then I might expect the scheduler to round-robin at the session level (rather than the QUIC stream level), splitting the bandwidth 50/50 between A and B.  Is that correct?

draft-lcurley-warp.md
vasilvv, line 184: I think per discussion, we should replace "produced by a single origin" with "can be prioritized between each other", since this seems to be the most salient property here.

draft-lcurley-warp.md
```diff
@@ -332,20 +328,26 @@ WebTransport can currently operate via HTTP/3 and HTTP/2, using QUIC or TCP unde
...
+The server uses the HTTP CONNECT request to identify client and the requested broadcast.
+The application dictates how this information is encoded into the request.
+For example, an authentication token and broadcast ID could be included in the path.
+
+The server MAY return an error status code for any reason. Examples include:
+* "400 Bad Request" when the broadcast and/or authorization cannot be parsed.
+* "401 Unauthorized" when the client cannot be identified.
+* "403 Forbidden" when the client is unauthorized to access the specified broadcast.
+* "404 Not Found" when the broadcast cannot be identified.
+* "307 Temporary Redirect" when the client should be served at a different url.
```
vasilvv, line 341: We should probably remove this for now.  Recall that we don't actually expose status codes in the Web API, so we may need to add something like this to the SETUP message at some point.

vasilvv (approved):
LGTM as-is, but you may want to update the definition of the broadcast to explain why we choose to bundle those tracks.

afrind (commented):
flushing old review comments

draft-lcurley-warp.md
kixelated: Reverted the example.

vasilvv (approved):

draft-lcurley-warp.md
```diff
@@ -332,20 +340,20 @@ WebTransport can currently operate via HTTP/3 and HTTP/2, using QUIC or TCP unde
 As mentioned in the motivation ({{motivation}}) section, TCP introduces head-of-line blocking and will result in a worse experience.
 It is RECOMMENDED to use WebTransport over HTTP/3.
 
-## Authentication
-The application SHOULD use the WebTransport CONNECT request for authentication.
-For example, including an authentication token in the path.
+### CONNECT
+The server uses the HTTP CONNECT request for identification and authorization of a track bundle.
+The specific mechanism is left up to the application.
+For example, a broadcast ID and authentication token could be included in the path.
```
kpugin, line 346: > a broadcast ID and authentication token

nit: since we replacing "broadcast" with "track bundle", should we say just "an ID" instead of "broadcast ID"

draft-lcurley-warp.md
```diff
@@ -564,7 +565,7 @@ Track Descriptor {
 {: #warp-track-descriptor title="Warp Track Descriptor"}
 
 * Track ID:
-A unique identifier for the track within the broadcast.
+A unique identifier for the track
```
kpugin, line 568: unique globally or unique within track bundle?

kixelated: Track bundle; fixed.

draft-lcurley-warp.md
kixelated: seems gud

draft-lcurley-warp.md
```diff
@@ -217,14 +217,24 @@ DISCUSS: We need to determine what are the exact requirements we need to impose
...
-
-As an example, consider a scenario where `example.org` hosts a simple live stream that anyone can subscribe to. That live stream would be a single Warp broadcast identified by the URL `https://example.org/livestream`. In the simplest implementation, it would provide only two media tracks, one with audio and one with video. In more complicated scenarios, it could provide multiple video formats of different levels of video quality; those tracks would be variants of each other. Note that the track IDs are opaque on the Warp level; if the player has not received the description of media tracks out of band in advance, it would have to request the broadcast description first.
+## Track Bundle
+A track bundle is a collection of tracks intended to be delivered together.
+Objects within a track bundle may be prioritized relative to each other via the delivery order property.
+This allows objects to be prioritized within a track (ex. newer > older) and between tracks (ex. audio > video).
+The track bundle contains a catalog indicating the available tracks.
+
+## Session
+A WebTransport session is established for each track bundle.
```
fluffy, line 227: except when it is not. Can we figure out how to write this to work for both raw quic and webetransport

fluffy: My proposal would be once Victor sets up the PR on raw quick that we get the info in the URL from a connect message if we are not using web transport. I think that can keep the code about the same.

PR #99, opened 2023-03-12
Object Model V3 (post 2nd virtual interim)

suhasHere:
This is the 3rd attempt to capture some of the discussions out of the 2nd virtual interim

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -177,32 +177,40 @@ x (b):
 
 # Model
 
-## Object
+## Objects
 
-An Object is the smallest unit that makes sense to decode and may not be independently decodable.  An Object MUST belong to a group.
+The basic element of Warp is an *object*. An object is a single addressable
+cacheable unit whose payload is a sequence of bytes.  An object is not
+necessarily independently decodable. An object MUST belong to a group
+{{Groups}}. Objects carry associated metadata such as priority, TTL or other
+information useable by a relay, but relays MUST treat object payloads as opaque.
 
-Few examples include, for video media an object could be an H.264 P frame or could be just a single slice from inside the P Frame. For audio media, it could be a single audio frame.
+DISCUSS: Can an object be partially decodable by an endpoint?
 
-Objects are not partially decodable. The end to end encryption and
-authentication operations are performed across the whole object, thus
-rendering partial objects unusable. Objects MUST be uniquely identifiable within the MoQ delivery system. Objects carry associated header/metadata containining priority, time to live, and other information aiding the caching/forwarding decision at the Relays. Objects MAY be optionally cached at Relays, however its contents 
-are opaque to the Relays.
+Authors agree that an object is always partially *forwardable* by a relay but
+disagree on whether a partial object can be used by a receiving endpoint.
 
-## Object Group
+Option 1: Objects can be partially decodable
 
-Groups are composition of objects and they carry the necessary dependecy information needed to process the objects in the group. An Object MUST belong to a group. The first object in the group MUST have the necessary dependency information needed to processs the rest of the objects. However, certain grouping MAY allow other objects within the group to carry dependency information. In such cases, the objects that carry information required to resolve dependencies MUST be marked appropriately in their headers. 
+Example: sending an entire GOP as a single object.  A receiver can decode the
+GOP from the begining without having the entire object present, and the object's
+tail could be dropped.  Sending a GOP as a group of not-partially-decoable
+objects might incur additional overhead on the wire and reprocessing of existing
+HLS segments to find object boundaries.
 
-A group shall provide following utilities
+Partial decoability could be another property of an object.
 
-* A way for subscribers to specifiy the appropriate consumption
-  point for enabling joins, rewinds and replay the objects, for
-  certain media usecases.
+Option 2: Objects cannot be partially decodable
 
-* A way to specify refresh points within a group, serving as decode
-  points, points of switching between qualties for audio/video media
+Objects could be end-to-end encrypted and the receiver might not be able to
+decrypt or authenticate an object until it is fully present.  Allowing Objects
+to span more than one useable unit may create more than one viable application
+mapping from media to wire format, which could be confusing for protocol users.
 
-* Serve as checkpoint for relays to implement appropriate congestion
-  responses.
+## Groups
+
+Removed the Group section as Victor is going to summarize PR# 90 and the
+previous text in this PR.
 
 ## Track
 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -182,7 +182,7 @@ x (b):
 The basic element of Warp is an *object*. An object is a single addressable
 cacheable unit whose payload is a sequence of bytes.  An object MAY depend on other 
 objects to be decoded. An object MUST belong to a group {{groups}}. Objects carry 
-associated metadata such as priority, TTL or other information useable by a relay, 
+associated metadata such as priority, TTL or other information usable by a relay, 
 but relays MUST treat object payloads as opaque.
 
 DISCUSS: Can an object be partially decodable by an endpoint?
@@ -193,12 +193,12 @@ disagree on whether a partial object can be used by a receiving endpoint.
 Option 1: A receiver MAY start decoding an object before it has been completely received
 
 Example: sending an entire GOP as a single object.  A receiver can decode the
-GOP from the begining without having the entire object present, and the object's
-tail could be dropped.  Sending a GOP as a group of not-partially-decoable
+GOP from the beginning without having the entire object present, and the object's
+tail could be dropped.  Sending a GOP as a group of not-partially-decodable
 objects might incur additional overhead on the wire and/or additional processing of 
 video segments at a sender to find object boundaries.
 
-Partial decoability could be another property of an object.
+Partial decodability could be another property of an object.
 
 Option 2: A receiver MUST NOT start decoding an object before it has completely arrived
 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -181,7 +181,7 @@ x (b):
 
 The basic element of Warp is an *object*. An object is a single addressable
 cacheable unit whose payload is a sequence of bytes.  An object MAY depend on other 
-objects to be decoded. An object MUST belong to a group {{Groups}}. Objects carry 
+objects to be decoded. An object MUST belong to a group {{groups}}. Objects carry 
 associated metadata such as priority, TTL or other information useable by a relay, 
 but relays MUST treat object payloads as opaque.
 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -177,11 +177,39 @@ x (b):
 
 # Model
 
-The basic element of Warp is *a media object*. A media object is a single addressable cacheable unit that may either contain a sequence of media samples, or some media-specific metadata, and may have relay-related attributes such as TTL or delivery priority associated with it. A media object is a sequence of bytes with a finite length. A Warp media object is similar in function to what is often referred to as "segments" or "chunks" in other media protocols; however, they are different from the traditional notion of chunks. The first key distinction is that Warp media objects are not always expected to be fully available, and thus any relays have to be able to convey partial media objects. The second key distinction is that Warp media objects may not be fully decodable by themselves; an object will contain a description of the prerequisites if that is the case.
+## Object
 
-*A media track* in Warp is a combination of *an init object* and a sequence of media objects. An init object is a format-specific self-contained description of the track that is required to decode any media object contained within the track, but can also be used as the metadata for track selection. If two media tracks carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
+An Object is the smallest unit that makes sense to decode and may not be independently decodable.  An Object MUST belong to a group.
 
-*A Warp broadcast* is a collection of multiple media tracks produced by a single origin. When subscribing to a broadcast, a peer has an option of subscribing to one, many or all media tracks within the broadcast.
+Few examples include, for video media an object could be an H.264 P frame or could be just a single slice from inside the P Frame. For audio media, it could be a single audio frame.
+
+Objects are not partially decodable. The end to end encryption and
+authentication operations are performed across the whole object, thus
+rendering partial objects unusable. Objects MUST be uniquely identifiable within the MoQ delivery system. Objects carry associated header/metadata containining priority, time to live, and other information aiding the caching/forwarding decision at the Relays. Objects MAY be optionally cached at Relays, however its contents 
+are opaque to the Relays.
+
+## Object Group
+
+Groups are composition of objects and they carry the necessary dependecy information needed to process the objects in the group. An Object MUST belong to a group. The first object in the group MUST have the necessary dependency information needed to processs the rest of the objects. However, certain grouping MAY allow other objects within the group to carry dependency information. In such cases, the objects that carry information required to resolve dependencies MUST be marked appropriately in their headers. 
+
+A group shall provide following utilities
+
+* A way for subscribers to specifiy the appropriate consumption
+  point for enabling joins, rewinds and replay the objects, for
+  certain media usecases.
+
+* A way to specify refresh points within a group, serving as decode
+  points, points of switching between qualties for audio/video media
+
+* Serve as checkpoint for relays to implement appropriate congestion
+  responses.
+
+## Track
+
+A media track in Warp is a combination of *an init object* and a sequence of media objects. An init object is a format-specific self-contained description of the track that is required to decode any media object contained within the track, but can also be used as the metadata for track selection. If two media tracks carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
+
+## Broadcast
+A Warp broadcast is a collection of multiple media tracks produced by a single origin. When subscribing to a broadcast, a peer has an option of subscribing to one, many or all media tracks within the broadcast.
 
 A Warp broadcast is globally identifiable via a URI. Within the broadcast, every media track is identified via *a track ID* that is unique within the broadcast. Within a single media track, every media object is identified by an *object ID* that is unique within the track.
 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -195,8 +195,8 @@ Option 1: Objects can be partially decodable
 Example: sending an entire GOP as a single object.  A receiver can decode the
 GOP from the begining without having the entire object present, and the object's
 tail could be dropped.  Sending a GOP as a group of not-partially-decoable
-objects might incur additional overhead on the wire and reprocessing of existing
-HLS segments to find object boundaries.
+objects might incur additional overhead on the wire and/or additional processing of 
+video segments at a sender to find object boundaries.
 
 Partial decoability could be another property of an object.
 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -180,17 +180,17 @@ x (b):
 ## Objects
 
 The basic element of Warp is an *object*. An object is a single addressable
-cacheable unit whose payload is a sequence of bytes.  An object is not
-necessarily independently decodable. An object MUST belong to a group
-{{Groups}}. Objects carry associated metadata such as priority, TTL or other
-information useable by a relay, but relays MUST treat object payloads as opaque.
+cacheable unit whose payload is a sequence of bytes.  An object MAY depend on other 
+objects to be decoded. An object MUST belong to a group {{Groups}}. Objects carry 
+associated metadata such as priority, TTL or other information useable by a relay, 
+but relays MUST treat object payloads as opaque.
 
 DISCUSS: Can an object be partially decodable by an endpoint?
 
 Authors agree that an object is always partially *forwardable* by a relay but
 disagree on whether a partial object can be used by a receiving endpoint.
 
-Option 1: Objects can be partially decodable
+Option 1: A receiver MAY start decoding an object before it has been completely received
 
 Example: sending an entire GOP as a single object.  A receiver can decode the
 GOP from the begining without having the entire object present, and the object's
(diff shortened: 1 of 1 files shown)


draft-lcurley-warp.md
```diff
@@ -177,11 +177,39 @@ x (b):
...
+
+Objects are not partially decodable. The end to end encryption and
+authentication operations are performed across the whole object, thus
+rendering partial objects unusable. Objects MUST be uniquely identifiable within the MoQ delivery system. Objects carry associated header/metadata containining priority, time to live, and other information aiding the caching/forwarding decision at the Relays. Objects MAY be optionally cached at Relays, however its contents 
+are opaque to the Relays.
+
+## Object Group
+
+Groups are composition of objects and they carry the necessary dependecy information needed to process the objects in the group. An Object MUST belong to a group. The first object in the group MUST have the necessary dependency information needed to processs the rest of the objects. However, certain grouping MAY allow other objects within the group to carry dependency information. In such cases, the objects that carry information required to resolve dependencies MUST be marked appropriately in their headers.
```
vasilvv, line 193: > The first object in the group MUST have the necessary dependency information needed to processs the rest of the objects. However, certain grouping MAY allow other objects within the group to carry dependency information. 

I feel like this is a really weird way to phrase this when in reality we almost always expect the objects in the group to depend on other objects that are not the first object of the group.  Could you reword this?

kixelated: > In such cases, the objects that carry information required to resolve dependencies MUST be marked appropriately in their headers.

What does this mean? I don't understand who this MUST is targeting or how to mark these objects.

draft-lcurley-warp.md
```diff
@@ -177,11 +177,39 @@ x (b):
...
+Objects are not partially decodable. The end to end encryption and
+authentication operations are performed across the whole object, thus
+rendering partial objects unusable. Objects MUST be uniquely identifiable within the MoQ delivery system. Objects carry associated header/metadata containining priority, time to live, and other information aiding the caching/forwarding decision at the Relays. Objects MAY be optionally cached at Relays, however its contents 
+are opaque to the Relays.
+
+## Object Group
+
+Groups are composition of objects and they carry the necessary dependecy information needed to process the objects in the group. An Object MUST belong to a group. The first object in the group MUST have the necessary dependency information needed to processs the rest of the objects. However, certain grouping MAY allow other objects within the group to carry dependency information. In such cases, the objects that carry information required to resolve dependencies MUST be marked appropriately in their headers. 
+
+A group shall provide following utilities
```
kixelated, line 195: This section is awkward. These aren't even examples; these are requirements. 

My feedback for this entire section is that it the target audience is wrong. IMO, the goal of the object model is to summarize the protocol and the high level interactions for implementers. Instead, this text reads like requirements directed at draft authors.

draft-lcurley-warp.md
```diff
@@ -177,11 +177,39 @@ x (b):
...
+
+* Serve as checkpoint for relays to implement appropriate congestion
+  responses.
+
+## Track
+
+A media track in Warp is a combination of *an init object* and a sequence of media objects. An init object is a format-specific self-contained description of the track that is required to decode any media object contained within the track, but can also be used as the metadata for track selection. If two media tracks carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
+
+## Broadcast
+A Warp broadcast is a collection of multiple media tracks produced by a single origin. When subscribing to a broadcast, a peer has an option of subscribing to one, many or all media tracks within the broadcast.
```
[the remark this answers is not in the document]
kixelated: +1

kixelated (commented):
Remarkably better than V1 and V2. There's still a few rough areas, but it's headed in the right direction.

draft-lcurley-warp.md
```diff
@@ -177,11 +177,39 @@ x (b):
...
+## Object
 
-*A media track* in Warp is a combination of *an init object* and a sequence of media objects. An init object is a format-specific self-contained description of the track that is required to decode any media object contained within the track, but can also be used as the metadata for track selection. If two media tracks carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
+An Object is the smallest unit that makes sense to decode and may not be independently decodable.  An Object MUST belong to a group.
```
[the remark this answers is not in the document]
kixelated: hmm, maybe something like:

> An **object** is a fragment of a track. Each object is delivered and prioritized independently such that the track can be partially decoded in the event of congestion. An object may only depend on prior objects within the same group.

It's a little awkward because it mentions track. Might be a reason to list the object model top-down instead of bottom-up.

draft-lcurley-warp.md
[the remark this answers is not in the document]
suhasHere: this text is unchanged from the existing draft.

draft-lcurley-warp.md
```diff
@@ -177,11 +177,39 @@ x (b):
...
 
-*A Warp broadcast* is a collection of multiple media tracks produced by a single origin. When subscribing to a broadcast, a peer has an option of subscribing to one, many or all media tracks within the broadcast.
+Few examples include, for video media an object could be an H.264 P frame or could be just a single slice from inside the P Frame. For audio media, it could be a single audio frame.
+
+Objects are not partially decodable. The end to end encryption and
+authentication operations are performed across the whole object, thus
+rendering partial objects unusable. Objects MUST be uniquely identifiable within the MoQ delivery system. Objects carry associated header/metadata containining priority, time to live, and other information aiding the caching/forwarding decision at the Relays. Objects MAY be optionally cached at Relays, however its contents
```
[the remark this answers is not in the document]
suhasHere: I don't think transmitting object as a GOP is much useful and it creates unequal prioritization between objects .. Some objects can be frames and other objects can be GOPs .. Also reading the discussion on delivery order on the mailing list , it does seems not right to treat a GOP as an object but instead make each thing inside GOP as an object at the least.

draft-lcurley-warp.md
suhasHere: > > Groups are composition of objects and they carry the necessary dependecy information needed to process the objects in the group.
> 
> Yeah this sentence is just wrong. The first object in a group does not know about the existence of future objects, let along contain a dependency graph like the sentence implies.

All this is saying is group defines some form of dependency relationship . Yes for video each object depends on the previous ones .. But that might not be the case for a different media or different grouping.

draft-lcurley-warp.md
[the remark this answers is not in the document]
vasilvv: Ah, good point.  I'll send a separate PR for that.

draft-lcurley-warp.md
```diff
@@ -177,11 +177,39 @@ x (b):
...
+## Object
 
-*A media track* in Warp is a combination of *an init object* and a sequence of media objects. An init object is a format-specific self-contained description of the track that is required to decode any media object contained within the track, but can also be used as the metadata for track selection. If two media tracks carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
+An Object is the smallest unit that makes sense to decode and may not be independently decodable.  An Object MUST belong to a group.
 
-*A Warp broadcast* is a collection of multiple media tracks produced by a single origin. When subscribing to a broadcast, a peer has an option of subscribing to one, many or all media tracks within the broadcast.
+Few examples include, for video media an object could be an H.264 P frame or could be just a single slice from inside the P Frame. For audio media, it could be a single audio frame.
+
+Objects are not partially decodable. The end to end encryption and
```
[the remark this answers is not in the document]
vasilvv: Also, regardless of whether they are decodable, they should always be partially deliverable.  For instance, if a relay receives the first half of a large object (which would be common for I-frames in high quality video), it should start forwarding those bytes to the subscribers immediately, instead of waiting for the rest of the object.  Doing otherwise would cause extra latency.

draft-lcurley-warp.md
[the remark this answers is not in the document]
afrind: This seems to be the heart of the disagreement.

@suhasHere : do you have an example of specific negative consequences if the object property is relaxed from "Objects are not partially decodable" to "Objects may be partially decodable"?

Since the only one doing decoding is the ultimate receiver of the media, I imagine it might require extra logic in that endpoint.  If that is the concern, can it be mitigated by adding a property to the object header that indicates if it is partially decodable?

@kixelated @vasilvv : To transmit a GOP using the model in this PR, it sounds like you can -- make the GOP a single object, but it would need to be delivered in its entirety.  That seems counter to the goals, since tail dropping a GOP is the original use-case for Warp and seems like it should be supported.  The alternative is to break the GOP into two or more objects and assign send them in a group.  It sounds like the concern with this approach is 1) metadata overhead, 2) processing overhead to break up pre-existing segments to transmit them this way.  Is that correct?

draft-lcurley-warp.md
kpugin, line 186: > The end to end encryption and
authentication operations are performed across the whole object, thus
rendering partial objects unusable.

Hm... "whole object" means metadata as well? That's probably not what we want. Also, text should probably say "MAY be encrypted" - we shouldn't require e2e encryption

draft-lcurley-warp.md
```diff
@@ -177,11 +177,39 @@ x (b):
...
+
+* A way to specify refresh points within a group, serving as decode
+  points, points of switching between qualties for audio/video media
+
+* Serve as checkpoint for relays to implement appropriate congestion
+  responses.
+
+## Track
+
+A media track in Warp is a combination of *an init object* and a sequence of media objects. An init object is a format-specific self-contained description of the track that is required to decode any media object contained within the track, but can also be used as the metadata for track selection. If two media tracks carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
```
kpugin, line 209: hm... what about tracks that don't have or don't need init objects?

draft-lcurley-warp.md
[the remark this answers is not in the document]
suhasHere: Given we are coming short in time on draft deadline. I would suggest, we keep this as open issue and have both the options listed in the draft. This seems to be totally fine to have benefits from the larger WG discussion.
Thoughts ?

I would propose the same with TrackIds being globally unique and why it wouldn't work technically.

draft-lcurley-warp.md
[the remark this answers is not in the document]
suhasHere: > Objects may be partially decodable.

Implies there is something smaller than objects. What do we call them ? Sub-objects ?

Rather , i would lean towards objects are the smallest atomic unit that makes sense for the application. Anything larger that needs to be grouped , will be part of the group. 

This provides a very clear distinction of object and group.

draft-lcurley-warp.md
```diff
@@ -177,11 +177,39 @@ x (b):
...
+## Object
 
-*A media track* in Warp is a combination of *an init object* and a sequence of media objects. An init object is a format-specific self-contained description of the track that is required to decode any media object contained within the track, but can also be used as the metadata for track selection. If two media tracks carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
+An Object is the smallest unit that makes sense to decode and may not be independently decodable.  An Object MUST belong to a group.
 
-*A Warp broadcast* is a collection of multiple media tracks produced by a single origin. When subscribing to a broadcast, a peer has an option of subscribing to one, many or all media tracks within the broadcast.
+Few examples include, for video media an object could be an H.264 P frame or could be just a single slice from inside the P Frame. For audio media, it could be a single audio frame.
+
+Objects are not partially decodable. The end to end encryption and
+authentication operations are performed across the whole object, thus
```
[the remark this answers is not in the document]
suhasHere: I would strongly suggest us to consider ietF BCP on encryption schemes. I would rather follow what BCP says than keeping the option open for MAY BE.

draft-lcurley-warp.md
kpugin: nit: does text needs to say "media"? are captions media? What about some head tracking data for VR?

draft-lcurley-warp.md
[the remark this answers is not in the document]
suhasHere: may be reword it as

Groups are composition of objects, where the objects in a group carry the necessary dependecy information needed to process the objects in the group

draft-lcurley-warp.md
suhasHere: Shouldn't require vs when E2E encrypted to be useful are 2 different things. We need it to work for latter too

draft-lcurley-warp.md
suhasHere: this is the original text. I did raise concerns on it multiple times .. Agree with your question

draft-lcurley-warp.md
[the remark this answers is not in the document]
kixelated: > I don't think transmitting object as a GOP is much useful and also leaving object definition undefined is pretty hard for implementing any useful strategies

You're going to have to elaborate.

Why does a relay care if an object is partially decodable? Like Victor mentioned, a relay SHOULD forward incomplete objects anyway otherwise it adds latency at each hop. If the decoder could process incomplete objects... why disallow it?

> @kixelated @vasilvv : To transmit a GOP using the model in this PR, it sounds like you can -- make the GOP a single object, but it would need to be delivered in its entirety. That seems counter to the goals, since tail dropping a GOP is the original use-case for Warp and seems like it should be supported. The alternative is to break the GOP into two or more objects and assign send them in a group. It sounds like the concern with this approach is 1) metadata overhead, 2) processing overhead to break up pre-existing segments to transmit them this way. Is that correct?

Yeah exactly. 

I would claim that 99% of the time there's no reason to break segments into frames. This is especially true for high-latency use cases that just want to send CMAF segments as the OBJECT payload for both backwards compatibility and simplicity. But it's also true for real-time use-cases that can't afford the latency or quality hit of non-reference frames (WebRTC).

We should absolutely support the ability to send each frame as an object, but it's the niche case. It's only useful when using temporal scalability, which I would not recommend enabling on purpose given the trade-offs, demonstrated by the lack of usage in the industry despite being easy in RTP. The only exception is when a hardware encoder uses a fixed GoP structure, but even then it's likely not worth the hassle of detecting the splitting the frames.

For more generic use-cases, the usefulness of objects improves dramatically if they are not atomic. In fact, this is the core difference between SCTP messages and QUIC streams. I have split segments into frames and sent them over SCTP (WebRTC data channels) and it does not work *because* messages are atomic.

So please, loosen the restriction on objects; it does not improve the protocol in any way.

draft-lcurley-warp.md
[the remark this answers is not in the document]
vasilvv: > Implies there is something smaller than objects. What do we call them ? Sub-objects ?

You don't need to call them anything, they're just an object with only some bytes in the beginning available.

Imagine we're sending raw 16-bit PCM audio over MoQ.  If we require each atomically decodable unit to be an object, then our objects are now 2 bytes long.  We would then have to put a header on every two bytes, which does not sound like a desirable property.

draft-lcurley-warp.md
kixelated: Yeah, I agree Alan. In #90 I stated the same thing as:

> A object can depend on another objects, but only within the same group and with a smaller sequence number. This guarantees that an object is decodable if all prior members of the group have been delivered.

I think @huitema wording is better in some aspects and worse in others. Let's try to pin that down.

draft-lcurley-warp.md
vasilvv: Well, you presumably need something to tell you what's in that track.  "init object" is format-specific, and it's basically just a header for the track.

draft-lcurley-warp.md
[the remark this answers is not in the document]
vasilvv: I think the encryption scheme should be specific to the track format (i.e. the transport itself does not know if it's encrypted or not).

draft-lcurley-warp.md
kixelated: Yeah, right now we shove dependencies for all groups/objects into the catalog as the init payload, specified on a per format basis. 

We should either:
1. fully punt that to the catalog, so it's a property of the catalog format itself (ex. base64 encoded in HLS)
2. fully treat it as an object (ex. object 234 depends on object 0), which will make more sense when we properly support layered encoding

I don't like mentioning the init object at all in the object model.

draft-lcurley-warp.md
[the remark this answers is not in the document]
kixelated: I love that Victor commented on his own text. :P

draft-lcurley-warp.md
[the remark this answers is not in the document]
kixelated: Exactly, the relay is agnostic to any encoding or encryption.

draft-lcurley-warp.md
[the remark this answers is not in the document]
kixelated: Yeah, I haven't found the best way to word it. Something like GROUPs indicate crude dependencies, while the more detailed dependencies are encoded at a lower level (ex. the codec)

draft-lcurley-warp.md
```diff
@@ -177,32 +177,40 @@ x (b):
...
 
-## Object Group
+Option 1: Objects can be partially decodable
 
-Groups are composition of objects and they carry the necessary dependecy information needed to process the objects in the group. An Object MUST belong to a group. The first object in the group MUST have the necessary dependency information needed to processs the rest of the objects. However, certain grouping MAY allow other objects within the group to carry dependency information. In such cases, the objects that carry information required to resolve dependencies MUST be marked appropriately in their headers. 
+Example: sending an entire GOP as a single object.  A receiver can decode the
+GOP from the begining without having the entire object present, and the object's
+tail could be dropped.  Sending a GOP as a group of not-partially-decoable
+objects might incur additional overhead on the wire and reprocessing of existing
+HLS segments to find object boundaries.
```
suhasHere, line 199: the overhead claim is incorrect. Data model shouldn't even touch about transport mapping

suhasHere: I would like to remove HLS reference here. Since RTP reference will turn this entire thinking around

draft-lcurley-warp.md
```diff
@@ -177,32 +177,40 @@ x (b):
...
 
-## Object
+## Objects
 
-An Object is the smallest unit that makes sense to decode and may not be independently decodable.  An Object MUST belong to a group.
+The basic element of Warp is an *object*. An object is a single addressable
+cacheable unit whose payload is a sequence of bytes.  An object is not
+necessarily independently decodable. An object MUST belong to a group
+{{Groups}}. Objects carry associated metadata such as priority, TTL or other
+information useable by a relay, but relays MUST treat object payloads as opaque.
```
suhasHere, line 187: We don't have agreement on "An object is not
necessarily independently decodable" in this para

draft-lcurley-warp.md
```diff
@@ -177,32 +177,40 @@ x (b):
...
 
-Groups are composition of objects and they carry the necessary dependecy information needed to process the objects in the group. An Object MUST belong to a group. The first object in the group MUST have the necessary dependency information needed to processs the rest of the objects. However, certain grouping MAY allow other objects within the group to carry dependency information. In such cases, the objects that carry information required to resolve dependencies MUST be marked appropriately in their headers. 
+Example: sending an entire GOP as a single object.  A receiver can decode the
+GOP from the begining without having the entire object present, and the object's
+tail could be dropped.  Sending a GOP as a group of not-partially-decoable
+objects might incur additional overhead on the wire and reprocessing of existing
+HLS segments to find object boundaries.
 
-A group shall provide following utilities
+Partial decoability could be another property of an object.
```
suhasHere, line 202: Its a DISCUSS still.

draft-lcurley-warp.md
vasilvv: I thought this just means that an object can depend on other objects?  Otherwise you can't put an individual P-frame into an object.

afrind: @suhasHere I'm confused.  Do you mean that objects are necessarily independently decodable?  I meant to convey that an object might or might not be independently decodable -- you can't assume anything because it's an object.

Do you mean to say that metadata might indicate that it is independently decodable?  If so I don't think that's in conflict with this text.

draft-lcurley-warp.md
afrind: > the overhead claim is incorrect. Data model shouldn't even touch about transport mapping

This is what I heard from the other authors and it seems to me like adding OBJECT framing will add non-zero bytes so it seems legit.

> HLS

I was capturing the comment from @vasilvv here but I can remove the specific protocol reference and say something like existing segments.

draft-lcurley-warp.md
```diff
@@ -177,32 +177,40 @@ x (b):
...
 
-* A way for subscribers to specifiy the appropriate consumption
-  point for enabling joins, rewinds and replay the objects, for
-  certain media usecases.
+Option 2: Objects cannot be partially decodable
```
kixelated, line 203: s/cannot/must not

If objects are allowed to be partially decodable, that's option 1.

draft-lcurley-warp.md
kixelated: It really comes down to fragmentation. Is the application allowed to decide how to fragment the media into objects?

For example, fMP4 has a configurable fragmentation interval. You can set it as long as a fragment per frame, or as high as a fragment per GoP, and anywhere in between. It's meant to specify the overhead versus latency trade-off. DASH typically uses a fragment per GoP while LL-DASH typically uses a fragment per frame.

I see parallels with OBJECT, which is also why it was named FRAGMENT at one point in history. The number of fragments and their boundaries depends on the codec and application. It's the difference between processing digit fragments per second with shared metadata, or processing hundreds of fragments per second each with their own independent metadata and ordering.

Now technically this requires buffering the entire fragment because of how MP4 encodes sample metadata as a header, but that's a limitation with MP4 rather than a general rule. It's also why latency is part of the trade-off instead of just overhead.

kixelated: If an application is allowed to decide how to fragment media into objects, then it stands to reason that an application is allowed to decide on how to decode said objects.

We already agree that a relay must not buffer objects, so there's nothing preventing the application from partially decoding an object, just like an application may partially decode individual RTP payloads (split across packets).

kixelated (approved):
LGTM if Suhas confirms that option 2 is what he's asking for.

vasilvv (approved):
This should be a reasonable starting point for further discussion.

PR #100, opened 2023-03-13
Define object groups

vasilvv:


draft-lcurley-warp.md
```diff
@@ -179,7 +179,11 @@ x (b):
...
 
-*A media track* in Warp is a combination of *an init object* and a sequence of media objects. An init object is a format-specific self-contained description of the track that is required to decode any media object contained within the track, but can also be used as the metadata for track selection. If two media tracks carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
+An object group is a sequence of media objects. Beginning of an object group can be used as a point at which the receiver can start receiving a track without having any other object groups available.
```
kixelated, line 182: > Beginning of an object group can be used as a point at which the receiver can start receiving a track without having any other object groups available.

Eh the wording here is weird. What does the "receiver can start receiving" even mean?

Can we at least say that each group is independent? Thinking about it a bit, that's all we need; it's actually fine to drop the "smaller sequence" property because delivery order effectively encodes that anyway.

vasilvv: "start consuming" might be better to avoid tautology.

I can add "an object within a group cannot depend on an object within any other group" if @suhasHere agrees with that.

kixelated: I think that's the property you're trying to convey, and we should say it explicitly.

kixelated: and I would give an example of why this property is useful. ex. so the relay can decide to transmit at group boundaries (startup, ABR, etc).

afrind: s/Beginning/The beginning/

vasilvv: > I think that's the property you're trying to convey, and we should say it explicitly.

The property in the current PR text is a somewhat weaker property (e.g. the current wording technically allows you to have H.264 open GOPs, while the stricter wording would prevent that).  I am personally fine with stricter wording, though.

kixelated: > without having any other object groups available

That implies that groups are independent. If you want weaker text, then remove that last part of the sentence.

> The beginning of an object group can be used as a point at which the receiver can start consuming a track

draft-lcurley-warp.md
```diff
@@ -179,7 +179,11 @@ x (b):
...
 
-*A media track* in Warp is a combination of *an init object* and a sequence of media objects. An init object is a format-specific self-contained description of the track that is required to decode any media object contained within the track, but can also be used as the metadata for track selection. If two media tracks carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
+An object group is a sequence of media objects. Beginning of an object group can be used as a point at which the receiver can start consuming a track without having any other object groups available. Object groups have an ID that identifies them uniquely within a track.
```
afrind, line 182: Are groups orderedable by group ID?

draft-lcurley-warp.md
kixelated: It means nothing of course, but that's the point.

I would go a step further and say that senders choose objects to transmit based on the group. This makes ABR possible, not just startup. But of course that's a stronger statement and we don't have the time to discuss.

draft-lcurley-warp.md
kixelated: I actually don't think they need to be, since delivery order implies order.

kixelated: They will need to be (how do you determine the newest group?), but let's punt that for now.

draft-lcurley-warp.md
vasilvv: > That implies that groups are independent. If you want weaker text, then remove that last part of the sentence.

From what I understand, in H.264 with an open GOP support, your group can start with a bunch of B-frames that depend on the previous GOP (or end with B-frames referencing the future GOP?), and you're just supposed to skip those when seeking.  This works just fine (you still always have an I-frame), but none of your GOPs are ever fully independent.

kixelated (approved):

draft-lcurley-warp.md
```diff
@@ -209,12 +209,13 @@ mapping from media to wire format, which could be confusing for protocol users.
 
 ## Groups
 
-Removed the Group section as Victor is going to summarize PR# 90 and the
-previous text in this PR.
+An object group is a sequence of media objects. Beginning of an object group can be used as a point at which the receiver can start consuming a track without having any other object groups available. Object groups have an ID that identifies them uniquely within a track.
+
+DISCUSS: We need to determine what are the exact requirements we need to impose on how the media objects depend on each other. Such requirements would need to address the use case (a join point), while being flexible enough to accomodate scenarios like B-frames and temporal scaling.
```
kpugin, line 214: we also need to define how subscriber would know about what groups have already been produced

PR #106, opened 2023-03-23
Rename "delivery order" to "send order"

kixelated:
Matches the terminology used by W3C is generally much better.
This is the order at which objects are sent, but does not guarantee delivery in this order.
Doesn't preclude further conversations about prioritization.

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -63,7 +63,7 @@ informative:
 
 This document defines the core behavior for Warp, a live media transport protocol over QUIC.
 Media is split into objects based on the underlying media encoding and transmitted independently over QUIC streams.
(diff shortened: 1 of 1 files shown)


fluffy (approved):
I do think that send order make more sense than delivery order but the parts of this that have to do with the prioritization scheme should probably be moved to a new section of the draft where we talk about the algorithms to decide what to send next.

afrind:
This is just a rename so maybe merge it after fixing conflicts (you'll need another approval)?  Based on the discussion today we'll be updating the priority section to indicate other possibilities for expressing priority.

huitema (approved):
Looks good to me. I would like to see this merged soon, so we can make the proper references to "send order" in the additional PRs.

suhasHere (approved):

PR #121, opened 2023-04-05
IETF116 : Update track definition

suhasHere:
This PR updates the track definition to reflect consensus at IETF 116

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -223,7 +223,7 @@ A track is a transform of a uncompressed media or metadata using a specific enco
 
 ### Full Track Name {#track-fn}
 
-Tracks are identified by a globally unique identifier, called "Track Full Name" and defined as shown below:
+Tracks are identified by a globally unique identifier, called "Full Track Name" and defined as shown below:
 
 ~~~~~~~~~~~~~~~
 Full Track Name = Track Namespace  "/"  Track Name
@@ -236,12 +236,12 @@ This document does not define the exact mechanism of naming Track Namespaces. Ap
 Example: 1
 Track Namespace = videoconferencing.example.com
 Track Name = meeting123/audio
-Track URI = videoconferencing.example.com/meeting123/audio
+Full Track Name = videoconferencing.example.com/meeting123/audio
 
 Example: 2
 Track Namespace = livestream.example
 Track Name = uaCafDkl123/audio
-Track URI = livestream.example/uaCafDkl123/audio
+Full Track Name = livestream.example/uaCafDkl123/audio
 ~~~
 
 ### Connection URL
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -168,18 +168,6 @@ Variant:
 
 : A track with the same content but different encoding as another track. For example, a different bitrate, codec, language, etc.
 
-Track Name:
-
-: See {{track-name}}.
-
-Provider: 
-
-: Entity capable of hosting media application
-session based on the MOQ Transport protocol.  It is responsible for
-authorizing the publishers/subscribers, managing the names used
-for Tracks and is scoped under a domain for a specific application.
-In certain deployments, a provider is also responsible for establishing trust between clients and relays for delivering media.
-
 ## Notational Conventions
 
 This document uses the conventions detailed in Section 1.3 of {{!RFC9000}} when describing the binary encoding.
@@ -234,7 +222,7 @@ A Track is the central concept within the MoQ Transport protocol for delivering 
 
 A track is a transform of a uncompresss media or metadata using a specific encoding process, a set of parameters for that encoding, and possibly an encryption process. The MoQ Transport protocol is designed to transport tracks.
 
-### Identification {#track-name}
+### Track URI {#track-uri}
 
 Tracks are identified by a globally unique identifier, called "Track URI" with the scheme shown below:
 
@@ -242,8 +230,7 @@ Tracks are identified by a globally unique identifier, called "Track URI" with t
 Track URI = "moq" "://" Track Namespace  "/"  Track Name
 ~~~~~~~~~~~~~~~
 
-"Track Namespace" MUST identify a globaly unique identifier, such as domain name or IANA PEN or something equivalent. 
-This is followed by the application context specific "Track Name", encoded as an opaque string. 
+"Track Namespace" MUST identify a globaly unique identifier, such as domain name or something equivalent. This is followed by the application context specific "Track Name", encoded as an opaque string. 
 
 ~~~
 Example: 1
@@ -259,7 +246,7 @@ Track URI = moq://livestream.tv/uaCafDkl123/audio
 
 ### Connection URL
 
-Each track MAY have an associated hop-by-hop "Connection URL" that specifies the network host to setup the transport connection. The syntax of the `Connection URL` and connection setup procedures are specific to the underlying transport protocol usage {{transport-usages}}.
+Each track MAY have an associated hop-by-hop Connection URL that specifies the network host to setup the transport connection. The syntax of the Connection URL and the associated connection setup procedures are specific to the underlying transport protocol usage {{transport-usages}}.
 
 ## Session
 A WebTransport session is established for each track bundle.
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -234,30 +234,32 @@ A Track is the central concept within the MoQ Transport protocol for delivering 
 
 A track is a transform of a uncompresss media or metadata using a specific encoding process, a set of parameters for that encoding, and possibly an encryption process. The MoQ Transport protocol is designed to transport tracks.
 
-A Track is produced by an authorized MoQ Entity, such as an
-Emitter (section 2.1.1 {{MOQ-ARCH}}) or a Catalog Maker (section 2.1.5 {{MOQ-ARCH}}), under a single Provider.
-
-
 ### Identification {#track-name}
 
-Tracks are identified by a globally unique identifier, called "Track Name" which is made of 2 components called "Track Prefix" and "Track Suffix" respectively.
+Tracks are identified by a globally unique identifier, called "Track URI" with the scheme shown below:
 
-"Track Prefix" MUST identify the owning provider by a standardized identifier, such as domain name or equivalent, then followed by the application context specific "Track Suffix", encoded as an opaque string. The catalog for a given MoQ Application MUST define the specifics of Track Name for each track advertised.
+~~~~~~~~~~~~~~~
+Track URI = "moq" "://" Track Namespace  "/"  Track Name
+~~~~~~~~~~~~~~~
+
+"Track Namespace" MUST identify a globaly unique identifier, such as domain name or IANA PEN or something equivalent. 
+This is followed by the application context specific "Track Name", encoded as an opaque string. 
 
 ~~~
 Example: 1
-Track Prefix = https://www.example.org/livestream/stream123
-Track Suffix = audio
-Track Name = https://www.example.org/livestream/stream123/audio
+Track Namespace = acme.meetings.com
+Track Name = meeting123/audio
+Track URI = moq://meetings.com/meeting123/audio
+
+Example: 2
+Track Namespace = livestream.tv
+Track Name = uaCafDkl123/audio
+Track URI = moq://livestream.tv/uaCafDkl123/audio
 ~~~
 
-In the example above, the Track Name identifies an URL with the "Track Prefix" identifies the provider domain as "example.org" and "Track Suffix" captures individual tracks as defined by the applications under that domain.
+### Connection URL
 
-## Track Bundle
-A track bundle is a collection of tracks intended to be delivered together.
-Objects within a track bundle may be prioritized relative to each other via the delivery order property.
-This allows objects to be prioritized within a track (ex. newer > older) and between tracks (ex. audio > video).
-The track bundle contains a catalog indicating the available tracks.
+Each track MAY have an associated hop-by-hop "Connection URL" that specifies the network host to setup the transport connection. The syntax of the `Connection URL` and connection setup procedures are specific to the underlying transport protocol usage {{transport-usages}}.
 
 ## Session
 A WebTransport session is established for each track bundle.
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -221,31 +221,32 @@ A Track is the central concept within the MoQ Transport protocol for delivering 
 
 A track is a transform of a uncompressed media or metadata using a specific encoding process, a set of parameters for that encoding, and possibly an encryption process. The MoQ Transport protocol is designed to transport tracks.
 
-### Track URI {#track-uri}
+### Full Track Name {#track-fn}
 
-Tracks are identified by a globally unique identifier, called "Track URI" with the scheme shown below:
+Tracks are identified by a globally unique identifier, called "Track Full Name" and defined as shown below:
 
 ~~~~~~~~~~~~~~~
-Track URI = "moq" "://" Track Namespace  "/"  Track Name
+Full Track Name = Track Namespace  "/"  Track Name
 ~~~~~~~~~~~~~~~
 
-"Track Namespace" MUST identify a globaly unique identifier, such as domain name or something equivalent. This is followed by the application context specific "Track Name", encoded as an opaque string. 
+This document does not define the exact mechanism of naming Track Namespaces. Applications building on top of MoQ MUST ensure that the mechanism used guarantees global uniqueness; for instance, an application could use domain names as track namespaces. Track Namespace is followed by the application context specific Track Name, encoded as an opaque string. 
+
 
 ~~~
 Example: 1
-Track Namespace = acme.meetings.com
+Track Namespace = videoconferencing.example.com
 Track Name = meeting123/audio
-Track URI = moq://acme.meetings.com/meeting123/audio
+Track URI = videoconferencing.example.com/meeting123/audio
 
 Example: 2
-Track Namespace = livestream.tv
+Track Namespace = livestream.example
 Track Name = uaCafDkl123/audio
-Track URI = moq://livestream.tv/uaCafDkl123/audio
+Track URI = livestream.example/uaCafDkl123/audio
 ~~~
 
 ### Connection URL
 
-Each track MAY have an associated hop-by-hop Connection URL that specifies the network host to setup the transport connection. The syntax of the Connection URL and the associated connection setup procedures are specific to the underlying transport protocol usage {{transport-usages}}.
+Each track MAY have one or more associated connection URLs specifying network hosts through which a track may be accessed. The syntax of the Connection URL and the associated connection setup procedures are specific to the underlying transport protocol usage {{transport-usages}}.
 
 ## Session
 A WebTransport session is established for each track bundle.
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -238,12 +238,14 @@ Tracks have the following properties:
 
 * Tracks MUST have a single security configuration, when exists.
 
-* Tracks MAY contain *an init object*, a format-specific self-contained description of the track that is required to decode any object contained within the track, but can also be used as the metadata for track selection.
-
 
 ### Identification
 
 Tracks are identified by a globally unique identifier, called "Track Name" which is made of 2 components called "Track Prefix" and "Track Suffix" respectively.
+
+~~~
+Track Name := Track Prefix + "#" + Track Suffix
+~~~
 
 "Track Prefix" MUST identify the owning provider by a standardized identifier, such as domain name or equivalent, then followed by the application context specific "Track Suffix", encoded as an opaque string.
 
@@ -251,13 +253,13 @@ Tracks are identified by a globally unique identifier, called "Track Name" which
 Example: 1
 Track Prefix = https://www.example.org/livestream/stream123
 Track Suffix = audio
-Track Name = https://www.example.org/livestream/audio
+Track Name = https://www.example.org/livestream/stream123#audio
 
 
 Example: 2
-Track Prefix = https://www.example.org
-Track Suffix = meetings/meeting123/video
-Track Name = https://www.example.org/meetings/meeting123/video
+Track Prefix = https://www.example.org/meetings
+Track Suffix = meeting123/video
+Track Name = https://www.example.org/meetings#meeting123/video
 
 ~~~
 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -46,7 +46,6 @@ normative:
   QUIC-RECOVERY: RFC9002
   WebTransport: I-D.ietf-webtrans-http3
   URI: RFC3986
-  MOQ-ARCH: I-D.nandakumar-moq-arch-00
 
   ISOBMFF:
     title: "Information technology — Coding of audio-visual objects — Part 12: ISO Base Media File Format"
@@ -58,6 +57,7 @@ informative:
     date: 2020-03
   NewReno: RFC6582
   BBR: I-D.cardwell-iccrg-bbr-congestion-control-02
+  MOQ-ARCH: I-D.nandakumar-moq-arch-00
 
 
 --- abstract
@@ -165,6 +165,10 @@ Variant:
 
 : A track with the same content but different encoding as another track. For example, a different bitrate, codec, language, etc.
 
+Track Name:
+
+: See {{track-name}}.
+
 Provider: 
 
 : Entity capable of hosting media application
@@ -223,48 +227,28 @@ DISCUSS: We need to determine what are the exact requirements we need to impose 
 
 ## Track {#model-track}
 
-Tracks form the central concept within the MoQ Transport protocol for
-delivering media and is made up of sequence of objects ({{model-object}}) organized in the form of groups ({{model-group}}).
+A Track is the central concept within the MoQ Transport protocol for delivering media and is made up of sequence of objects ({{model-object}}) organized in the form of groups ({{model-group}}).
 
-A track is a transform of a uncompresss media using a specific encoding process, a set of parameters for that encoding, and possibly an encryption process. The MoQ Transport protocol is designed 
-to transport tracks.
+A track is a transform of a uncompresss media or metadata using a specific encoding process, a set of parameters for that encoding, and possibly an encryption process. The MoQ Transport protocol is designed to transport tracks.
 
-Tracks have the following properties:
-
-* A Track MUST be owned by a single authorized MoQ Entity, such as an
-  Emitter (section 2.1.1 {{MOQ-ARCH}}) or a Catalog Maker (section 2.1.5 {{MOQ-ARCH}}), under a single Provider.
-
-* Tracks MUST have a single encoding configuration.
-
-* Tracks MUST have a single security configuration, when exists.
+A Track is produced by an authorized MoQ Entity, such as an
+Emitter (section 2.1.1 {{MOQ-ARCH}}) or a Catalog Maker (section 2.1.5 {{MOQ-ARCH}}), under a single Provider.
 
 
-### Identification
+### Identification {#track-name}
 
 Tracks are identified by a globally unique identifier, called "Track Name" which is made of 2 components called "Track Prefix" and "Track Suffix" respectively.
 
-~~~
-Track Name := Track Prefix + "#" + Track Suffix
-~~~
-
-"Track Prefix" MUST identify the owning provider by a standardized identifier, such as domain name or equivalent, then followed by the application context specific "Track Suffix", encoded as an opaque string.
+"Track Prefix" MUST identify the owning provider by a standardized identifier, such as domain name or equivalent, then followed by the application context specific "Track Suffix", encoded as an opaque string. The catalog for a given MoQ Application MUST define the specifics of Track Name for each track advertised.
 
 ~~~
 Example: 1
 Track Prefix = https://www.example.org/livestream/stream123
 Track Suffix = audio
-Track Name = https://www.example.org/livestream/stream123#audio
-
-
-Example: 2
-Track Prefix = https://www.example.org/meetings
-Track Suffix = meeting123/video
-Track Name = https://www.example.org/meetings#meeting123/video
-
+Track Name = https://www.example.org/livestream/stream123/audio
 ~~~
 
-In both the examples above, the "Track Prefix" identifies the provider domain as "example.org" and "Track Suffix" captures individual tracks as defined by the applications under that domain.
-
+In the example above, the Track Name identifies an URL with the "Track Prefix" identifies the provider domain as "example.org" and "Track Suffix" captures individual tracks as defined by the applications under that domain.
 
 ## Track Bundle
 A track bundle is a collection of tracks intended to be delivered together.
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -46,6 +46,7 @@ normative:
   QUIC-RECOVERY: RFC9002
   WebTransport: I-D.ietf-webtrans-http3
   URI: RFC3986
+  MOQ-ARCH: I-D.nandakumar-moq-arch-00
 
   ISOBMFF:
     title: "Information technology — Coding of audio-visual objects — Part 12: ISO Base Media File Format"
@@ -164,6 +165,13 @@ Variant:
 
 : A track with the same content but different encoding as another track. For example, a different bitrate, codec, language, etc.
 
+Provider: 
+
+: Entity capable of hosting media application
+session based on the MOQ Transport protocol.  It is responsible for
+authorizing the publishers/subscribers, managing the names used
+for Tracks and is scoped under a domain for a specific application.
+In certain deployments, a provider is also responsible for establishing trust between clients and relays for delivering media.
 
 ## Notational Conventions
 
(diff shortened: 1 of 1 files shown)


draft-lcurley-warp.md
```diff
@@ -207,15 +215,54 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
+
+Tracks are identified by a globally unique identifier, called "Track Name" which is made of 2 components called "Track Prefix" and "Track Suffix" respectively.
+
+"Track Prefix" MUST identify the owning provider by a standardized identifier, such as domain name or equivalent, then followed by the application context specific "Track Suffix", encoded as an opaque string.
+
+~~~
+Example: 1
+Track Prefix = https://www.example.org/livestream/stream123
+Track Suffix = audio
+Track Name = https://www.example.org/livestream/audio
```
vasilvv, line 254: Is this meant to be `https://www.example.org/livestream/stream123/audio`?

draft-lcurley-warp.md
```diff
@@ -207,15 +215,54 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
+
+
+Example: 2
+Track Prefix = https://www.example.org
+Track Suffix = meetings/meeting123/video
+Track Name = https://www.example.org/meetings/meeting123/video
```
vasilvv, line 260: My issue with this syntax is that it's unclear which part is a prefix and which is a suffix.  How about `https://www.example.org/meetings/meeting123#video`?

draft-lcurley-warp.md
```diff
@@ -207,15 +215,54 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
+* Tracks MUST have a single security configuration, when exists.
+
+* Tracks MAY contain *an init object*, a format-specific self-contained description of the track that is required to decode any object contained within the track, but can also be used as the metadata for track selection.
+
+
+### Identification
+
+Tracks are identified by a globally unique identifier, called "Track Name" which is made of 2 components called "Track Prefix" and "Track Suffix" respectively.
+
+"Track Prefix" MUST identify the owning provider by a standardized identifier, such as domain name or equivalent, then followed by the application context specific "Track Suffix", encoded as an opaque string.
```
vasilvv, line 248: Can we just say that the track prefix identifies the MoQ server to which you have to connect to publish/subscribe a track?

draft-lcurley-warp.md
```diff
@@ -207,15 +215,54 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
+Tracks have the following properties:
+
+* A Track MUST be owned by a single authorized MoQ Entity, such as an
+  Emitter (section 2.1.1 {{MOQ-ARCH}}) or a Catalog Maker (section 2.1.5 {{MOQ-ARCH}}), under a single Provider.
+
+* Tracks MUST have a single encoding configuration.
+
+* Tracks MUST have a single security configuration, when exists.
+
+* Tracks MAY contain *an init object*, a format-specific self-contained description of the track that is required to decode any object contained within the track, but can also be used as the metadata for track selection.
```
vasilvv, line 241: If we decide that we're moving all the catalog stuff into the separate document, this can be just removed (since at least in the current draft, the init objects are in the catalog).

draft-lcurley-warp.md
```diff
@@ -207,15 +215,54 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
+to transport tracks.
+
+Tracks have the following properties:
+
+* A Track MUST be owned by a single authorized MoQ Entity, such as an
+  Emitter (section 2.1.1 {{MOQ-ARCH}}) or a Catalog Maker (section 2.1.5 {{MOQ-ARCH}}), under a single Provider.
+
+* Tracks MUST have a single encoding configuration.
+
+* Tracks MUST have a single security configuration, when exists.
```
vasilvv, line 239: Could you explain to me what are the consequences of those three requirements?  I.e. this section uses "MUST" in every paragraph, but I am not sure who is this a requirement on.

vasilvv (commented):
I like the idea of track prefix / track suffix, and I am excited that we are making progress on this.  That said, I do have some question about properties specifically.

draft-lcurley-warp.md
suhasHere: ah yes .. that's a typo

draft-lcurley-warp.md
suhasHere: good point, will update it

draft-lcurley-warp.md
suhasHere: The MUST is used to scope the definition of track within the MoQ system. 
For example, the first MUST says if there are 2 different encodings, then they should be treated as 2 different tracks and the same is with security configuration.

Without these, i am afraid the definition will be very loose and ambiguous.

draft-lcurley-warp.md
```diff
@@ -164,6 +165,13 @@ Variant:
 
 : A track with the same content but different encoding as another track. For example, a different bitrate, codec, language, etc.
 
+Provider:
```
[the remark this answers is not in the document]
suhasHere: Provider , as expained, is the authority domain along with the other things captured in the definition

MoQ Server, IIUC from the latest warp draft as "The party accepting an incoming Warp session." , can be any box that accepts WT connection ?

draft-lcurley-warp.md
suhasHere: I do see what you mean here ..

draft-lcurley-warp.md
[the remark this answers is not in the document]
suhasHere: May be i need your help to frame it appropriately - Can you explain what is MoQ Server from your understanding , like what are its roles and so on

draft-lcurley-warp.md
suhasHere: Agreed on the confusion .. Building on your proposal , should we form the track name as 

Track Name = Track Prefix + '#' + Track Suffix

draft-lcurley-warp.md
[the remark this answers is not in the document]
vasilvv: I think I made some unnecessary assumptions, so let me change the question a little bit.  Would it be fair to say that a provider is any entity that accepts publish/subscribe requests?

draft-lcurley-warp.md
```diff
@@ -207,15 +215,54 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
 DISCUSS: We need to determine what are the exact requirements we need to impose on how the media objects depend on each other. Such requirements would need to address the use case (a join point), while being flexible enough to accomodate scenarios like B-frames and temporal scaling.
 
-## Track
+## Track {#model-track}
+
+Tracks form the central concept within the MoQ Transport protocol for
+delivering media and is made up of sequence of objects ({{model-object}}) organized in the form of groups ({{model-group}}).
+
+A track is a transform of a uncompresss media using a specific encoding process, a set of parameters for that encoding, and possibly an encryption process. The MoQ Transport protocol is designed 
+to transport tracks.
```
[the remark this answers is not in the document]
vasilvv: I can have a track that just instructs the player to draw some overlay text at some point in time.  A catalog would be another example of a track that's not strictly speaking an encoding of some media.

draft-lcurley-warp.md
vasilvv: I see, let me go point-by-point then.

> A Track MUST be owned by a single authorized MoQ Entity, such as an Emitter (section 2.1.1 {{MOQ-ARCH}}) or a Catalog Maker (section 2.1.5 {{MOQ-ARCH}}), under a single Provider.

I'm not entirely sure what "owned" means.  Can we just say it's produced by a single entity?

> Tracks MUST have a single encoding configuration.

This might be unnecessarily restrictive, since it's not unreasonable for a track to change configuration in certain scenarios (e.g. in live ingestion, video dimensions can change mid-track in some scenarios).

> Tracks MUST have a single security configuration, when exists.

I understand what you're trying to say here, but "security configuration" is a bit vague.  Can we just say "a track is the largest entity to which a client can subscribe or publish"?   Since we perform authentication when we do subscribe/publish request, that would imply that the track is the unit at which we perform security.

draft-lcurley-warp.md
```diff
@@ -207,15 +215,56 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
+
+* Tracks MUST have a single security configuration, when exists.
+
+
+### Identification
+
+Tracks are identified by a globally unique identifier, called "Track Name" which is made of 2 components called "Track Prefix" and "Track Suffix" respectively.
+
+~~~
+Track Name := Track Prefix + "#" + Track Suffix
```
vasilvv, line 247: Since this is a URI, I think you might want to reference RFC 3986, and just say that track suffix is the fragment.

draft-lcurley-warp.md
vasilvv: I guess the question I have is, "once I have a track name, how do I actually subscribe (or publish) to it?".  I assume that by default, we would do the following:

1. Extract track prefix, and establish a QUIC connection (or a WebTransport session) to the host/port/path in it (unless we already have that connection from a previous track).
2. Send SUBSCRIBE_REQUEST / PUBLISH_REQUEST with the track suffix.

We can define alternative schemes, but without any other prior knowledge, I assume this is the expected approach.  Does this sound correct?

draft-lcurley-warp.md
fluffy: Have we agreed a track name needs to be a URL ? I 'm fine with that but thought some people did not want that.

draft-lcurley-warp.md
```diff
@@ -207,15 +215,56 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
+~~~
+
+"Track Prefix" MUST identify the owning provider by a standardized identifier, such as domain name or equivalent, then followed by the application context specific "Track Suffix", encoded as an opaque string.
+
+~~~
+Example: 1
+Track Prefix = https://www.example.org/livestream/stream123
+Track Suffix = audio
+Track Name = https://www.example.org/livestream/stream123#audio
```
fluffy, line 256: We can't use the #audio here - that would be mean something very differnt than /audio. I think we need to rethink this example.

draft-lcurley-warp.md
```diff
@@ -207,15 +215,56 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
+
+
+Example: 2
+Track Prefix = https://www.example.org/meetings
+Track Suffix = meeting123/video
+Track Name = https://www.example.org/meetings#meeting123/video
```
fluffy, line 262: again , I think use of # is wrong here.

fluffy (approved):

afrind (commented):
Individual review

draft-lcurley-warp.md
suhasHere: How does one setup a QUIC connection setup is not part of this PR  and subscribe/publish must carry the entire Track Name and we agreed it in IETF116

draft-lcurley-warp.md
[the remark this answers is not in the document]
vasilvv: I'm not entirely sure how you separate those.  If the track name includes a domain name, at some point, we would have to check that the server is allowed to speak on behalf of that domain.  Normally, that requires validating TLS certificate for that domain, which you typically get when you connect to it.  There are some ways to delegate this, of course, but this is what HTTP (and most applications of this kind) do, and what I assume is the base case here.

draft-lcurley-warp.md
vasilvv: Could you explain why `#` does not work here?

draft-lcurley-warp.md
vasilvv: I believe we agreed that the name has to be global, and we would need to rely on DNS for that.  I believe URL follows naturally for that.

kixelated: I don't think we should use URLs with this design.

For example, when the client connects to akamai.com, DNS/TLS identifies and authenticates the server based on the URL. But let's say the akamai.com server publishes a track with the prefix webex.com.

The client can't actually connect to the track URL. The very point of using a CDN in the first place is to shield the origin. 

We could use a URL if the hostname is used to verify the authenticity. But that would require that every object is signed using some key derived from a public certificate. It would still be weird to include a protocol though.

draft-lcurley-warp.md
kixelated: > Can we just say that the track prefix identifies the MoQ server to which you have to connect to publish/subscribe a track?

It's not though, because I'm pretty sure @suhasHere wants to serve track URLs from other domains. ex. akamai.com serving webex.com tracks.

suhasHere: I see 2 layers of things here . 
1. moq protocol that deals with pub/sub on track names
2. a connection protocol (WT or native QUIC) that is needed to connect based on some business logic

In some use-cases there is overlap between 1 and 2 and that's fine.  Keeping these 2 things apart will make the MoQ protocol be defined in terms of tracks and how things around tracks work to get the objects/groups delivered.

draft-lcurley-warp.md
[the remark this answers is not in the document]
suhasHere: yeah, that's right. Tracks are part of an authorization scope and how it is verified is very application dependent.

draft-lcurley-warp.md
suhasHere: I think we are mixing connection and moq tracks in the same space.

Agree when the connect happens to a relay , it happens to the akamai.com , but the track names identify media objects belonging to an application under a domain trusted by the relay.  The subscribes/publish should carry authz information proving that a given moq client is authorized for those tracks.

It is beneficial if we keep the relay discovery, connection separate from MoQ protocol concepts.

kixelated: Yeah, these are technically not URLs since they don't locate a resource. URI is probably the correct term here although warp-03 used URN.

vasilvv: > It is beneficial if we keep the relay discovery, connection separate from MoQ protocol concepts.

There are definitely a lot of use cases in which one would need to have a URL to tell the client the exact way it's supposed to fetch a specific track.  I'm not sure it's possible to keep discovery and naming separate, those are very tightly coupled problems.

> We could use a URL if the hostname is used to verify the authenticity. But that would require that every object is signed using some key derived from a public certificate. It would still be weird to include a protocol though.

Well, there are multiple ways to solve this problem, I can think of at least four (three of which have been discussed for HTTP before):
1. Require the CDN certificate be valid for the origin's hostname ([HTTP pooling](https://datatracker.ietf.org/doc/html/rfc9113#name-connection-reuse) / [Alt-Svc approach](https://datatracker.ietf.org/doc/html/rfc7838#section-2.1))
2. Authenticate the individual connection as valid for the origin ([secondary certificates approach](https://datatracker.ietf.org/doc/draft-ietf-httpbis-http2-secondary-certs/))
3. Have the origin sign off on every PUBLISH_REPLY for the origin it does not directly have a certificate for.
4. Sign individual objects ([signed exchanges approach](https://datatracker.ietf.org/doc/draft-yasskin-http-origin-signed-responses/))

suhasHere:
@vasilvv  @afrind  I have pushed a new commit that should address your feedback
Please let me know .. thanks

draft-lcurley-warp.md
```diff
@@ -207,15 +215,56 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
+  Emitter (section 2.1.1 {{MOQ-ARCH}}) or a Catalog Maker (section 2.1.5 {{MOQ-ARCH}}), under a single Provider.
+
+* Tracks MUST have a single encoding configuration.
+
+* Tracks MUST have a single security configuration, when exists.
+
+
+### Identification
+
+Tracks are identified by a globally unique identifier, called "Track Name" which is made of 2 components called "Track Prefix" and "Track Suffix" respectively.
```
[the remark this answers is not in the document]
suhasHere: I do see prefix, suffix and track name as part of the catalog. They all can be strings and it is left to application to define its length, encoding. A suffix need not be a compression technique as it is very application dependent naming.

The compression technique comes from TrackId that gets setup as part of pub/sub control exchange ( #122
)

draft-lcurley-warp.md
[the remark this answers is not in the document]
kixelated: If it's application dependent, why is it in this text? What would change if the relay was not aware of the concept of "provider"?

draft-lcurley-warp.md
kixelated: The word "provider" confuses me too, especially because I would consider a relay a provider (ie. "to supply or make available"). I'm not sure how a provider differs from a producer?

draft-lcurley-warp.md
kixelated: I think 1&2 require all tracks to be sourced from the same origin. The CDN would also have to know what resources the client is going to request during the TLS handshake.

I don't think 3 is quite good enough. The objects themselves could be modified without 4. A middle ground is using asymmetric crypto to sign a symmetric key in the PUBLISH / SUBSCRIBE_REPLY.

draft-lcurley-warp.md
```diff
@@ -207,15 +219,36 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
 DISCUSS: We need to determine what are the exact requirements we need to impose on how the media objects depend on each other. Such requirements would need to address the use case (a join point), while being flexible enough to accomodate scenarios like B-frames and temporal scaling.
 
-## Track
+## Track {#model-track}
+
+A Track is the central concept within the MoQ Transport protocol for delivering media and is made up of sequence of objects ({{model-object}}) organized in the form of groups ({{model-group}}).
+
+A track is a transform of a uncompresss media or metadata using a specific encoding process, a set of parameters for that encoding, and possibly an encryption process. The MoQ Transport protocol is designed to transport tracks.
+
+A Track is produced by an authorized MoQ Entity, such as an
```
kixelated, line 234: What is an "authorized MoQ Entity"? 

I would remove this line and the references to MOQ-ARCH.

draft-lcurley-warp.md
```diff
@@ -207,15 +219,36 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
 An object group is a sequence of media objects. Beginning of an object group can be used as a point at which the receiver can start consuming a track without having any other object groups available. Object groups have an ID that identifies them uniquely within a track.
 
 DISCUSS: We need to determine what are the exact requirements we need to impose on how the media objects depend on each other. Such requirements would need to address the use case (a join point), while being flexible enough to accomodate scenarios like B-frames and temporal scaling.
 
-## Track
+## Track {#model-track}
+
+A Track is the central concept within the MoQ Transport protocol for delivering media and is made up of sequence of objects ({{model-object}}) organized in the form of groups ({{model-group}}).
+
+A track is a transform of a uncompresss media or metadata using a specific encoding process, a set of parameters for that encoding, and possibly an encryption process. The MoQ Transport protocol is designed to transport tracks.
```
kixelated, line 232: > transform of a uncompresss media

typo

I would say "a track is an encoded bitstream".

draft-lcurley-warp.md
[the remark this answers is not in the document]
kixelated: Sure, let's punt that to the catalog format then.

draft-lcurley-warp.md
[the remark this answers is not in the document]
suhasHere: I don't understand the question here.  Relays need to know what providers they trust (via some business relationship or something)

draft-lcurley-warp.md
suhasHere: I don't see the current proposal prohibits from doing the single URL. It does duplicates few extra bytes. 
Relay discovery is outside of MoQ Transport protocol and there are multiple ways it can be done. There is no point in enforcing that in this spec.

draft-lcurley-warp.md
suhasHere: I can share the intent here and suggest me something that works

The provider is currently defined as
Entity capable of hosting media application session based on the MOQ Transport protocol. It is responsible for authorizing the publishers/subscribers, managing the names used for Tracks and is scoped under a domain for a specific application. In certain deployments, a provider is also responsible for establishing trust between clients and relays for delivering media.

It can be called origin, but it is overloaded .. Please suggest what should we call this thing

draft-lcurley-warp.md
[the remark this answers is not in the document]
suhasHere: yes we can punt it to the Catalog, but the transport spec must specify following properties about track names

Given a track name in the Subscribe/Publish the relay must be able to 
- verifying if the relay is authorized to serve the given provider
- validate the requesting endpoint is authorized for the track requested

This requires provider to be identified by some for a DNS name or equivalent  and latter can be done by carrying track authz info

draft-lcurley-warp.md
kixelated: It doesn't matter if the relay can decode the media; any entity should still be able to add/remove tracks. 

Examples:
- An edge relay could create an advertisement track (unique per viewer). It would contain signals when the user should request an advertisement and with what payload.
- A relay could insert a timed metadata, such chat messages on Twitch so they get rendered at the same time.
- An edge relay could create an estimated bitrate track (unique per hop). It would contain periodic bitrate estimates so the player can perform ABR.
- An edge relay could create a CMCD track. This is a superset of the estimated bitrate, and contains a lot of statistics about the session.
- The viewer could also create a CMCD track. In fact this is required if we want granular statistics about rebuffering.

There's a whole slew of possible use cases. I don't see any benefit to locking down the possible use-cases of tracks, which is what you're doing by requiring globally unique track URLs authorized by a reachable origin.

kixelated: @wilaw you might have some opinions here.

suhasHere: Let me ask few questions for clarification


> * An edge relay could create an advertisement track (unique per viewer). It would contain signals when the user should request an advertisement and with what payload.

Can i have an example of who is this track authorized to ? 

> * A relay could insert a timed metadata, such chat messages on Twitch so they get rendered at the same time.

is this inserting into media stream ? if so, relays can't do that ? If not, can you please eloborate of how it is done and how such a timed metadata identified ?



> * An edge relay could create an estimated bitrate track (unique per hop). It would contain periodic bitrate estimates so the player can perform better ABR.

Same question as earlier ? if the edge is controlled by akamai running twitch application, how would this track look like, who authorizes this track or the edge node ?



> * An edge relay could create a CMCD track. This is a superset of the estimated bitrate, and contains a lot of statistics about the session.

how is this track announced and what domain does this track belong to ? in the akamai + twich exmaple, is it part of akamai's auth scope or twitch's?

> There's a whole slew of possible use cases. I don't see any benefit to locking down the possible use-cases of tracks, which is what you're attempting to do by requiring globally unique track URLs authorized by an origin.

The  track name need to know under what domain it is served and where the authorization happens depends on how the business relationship is set between the domain and the CDN. Many of the cases , i suspect it happens at the edge. I don't see how it is blocking/locking possible track use-cases.

vasilvv:
I think I understand now better what is being proposed, which is good.  Unfortunately, now that I have a better understanding of the proposal, I am not sure I agree with it.

As far as I understand, the idea is:

1. Track prefix identifies the CDN customer to the relay, plus some extra context information.
2. Track suffix identifies the actual track within the prefix.
3. The information about where to connect to in order to publish or subscribe the track is not contained within the track name.

Observe, that as a general principle, when publishing or subscribing to a track, a client needs to know two things: where to connect to, and what kind of extra context to provide to the server in order to access the specific track.  Based on this observation, I have two objections to the proposed design:

1. Track names that don't have a host and a port are not useful, since they don't actually tell the client how to access them.
2. As far as I can tell, the distinction between the track prefix and the track suffix is an internal detail of relay/application -- a client can't actually do anything with the knowledge of whether the track belongs to a specific CDN customer.

I believe we should identify tracks by a tuple of `(network endpoint, endpoint-scoped track name)`.  An individual application can, of course, encode the customer name (what's referred to as _provider_ in this PR) in the track name component; alternatively, it can add a layer on top of MoQ transport that overrides the network endpoint name with an alternative network endpoints of choice (I assume that's what Suhas means when he says "relay discovery").

suhasHere:
> 2\. As far as I can tell, the distinction between the track prefix and the track suffix is an internal detail of relay/application -- a client can't actually do anything with the knowledge of whether the track belongs to a specific CDN customer.

The relay looks at the track name and it doesn't know where the prefix and suffix is. the track name is used to find out next steps since it contains the fully qualified resource name

vasilvv:
> I want to tease it a bit further. Why would the network end-point info be part of the track names. This mixes layers of abstraction.

Okay, maybe there's a bit of confusion of what "track name" means.  Let's say for the purpose of this comment that "track name" is the thing you send in a pub/sub request (and we'd have some other thing, "global track location", that is a tuple of network endpoint and track name).

> Track Names are application scoped (don't think they are end point scoped, since an end-point is working under an application scope to produce or consume media)

That's one way to structure an application, but there's somewhat of a problem here.  Application is not really a physical concept -- it's an agreement of several servers to do the same thing.  That is to say, you can have an application under domain `videoconferencing.example`, and you can structure it in a way that both `moq://cdn-x.example` and `moq://cdn-y.example` would serve the same data if you ask for a track named `videoconferencing.example/conference/person-a/video`.  However, the only things that happen on the wire are (1) the client connects to either `cdn-x.example` or `cdn-y.example`, and (2) it sends a subscribe request with a string `videoconferencing.example/conference/person-a/video`.  In that sense, even if we say that "the track is scoped to videoconferencing.example", the _de facto scope_ of any track name is the host you connect to, and any additional conditions you define (e.g. the track prefix being the name of the application) are naming conventions telling the server how to reply (note that in reality, the server can reply with anything).  If we say "all tracks must be prefix + suffix", we're adding an additional constraint, and I believe that this constraint reduces the flexibility we have (a little bit more on this below).

> The relay looks at the track name and it doesn't know where the prefix and suffix is. the track name is used to find out next steps since it contains the fully qualified resource name

I assume you mean that the relay would read the track name, which we would split into prefix `videoconferencing.example` and suffix `conference/person-a/video`, and use the rules for `videoconferencing.example` to figure out where to get `conference/person-a/video`.  That's one possible way to set things up (let's call this **Scenario A**), but there are also other approaches possible here, for instance:

**Scenario B**.  The CDN has two customers, `videoconferencing.example` and `live.example`.  Both point their A record (or even NS record) to `cdn.example`'s IP address.  `cdn.example` then identifies the traffic going to its IP address based on the SNI (if using WebTransport, `:authority` header can also be used).

**Scenario C**.  The CDN is owned entirely by one customer, let's say `vod.example`.  In that case, there is only one customer, so there's no need for a customer identifier -- you'd have a track name like `oHg5SJYRHA0/video`, and the VoD CDN would have some service-specific logic to find a backend where that track is stored.

Note that, while track prefix with application name is useful in Scenario A, it's not necessary in B or C (both are real ways people deploy applications today), and if your application relies on a CDN that uses approach in Scenario A, you hopefully already know about that and it's included in your track names in a way that a CDN would be able to parse, so I'm not sure we should always require it.

draft-lcurley-warp.md
```diff
@@ -46,6 +46,7 @@ normative:
   QUIC-RECOVERY: RFC9002
   WebTransport: I-D.ietf-webtrans-http3
   URI: RFC3986
+  MOQ-ARCH: I-D.nandakumar-moq-arch-00
```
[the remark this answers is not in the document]
afrind: @suhasHere : my personal philosophy is to have dependencies as minimal as possible, and I didn't see this dependency as being required at this point in the process.  That architecture document hasn't had as much scrutiny yet as this document, and with a normative dependency, would we need to adopt them both?  I'd prefer we pull the specific definitions we need from there into here for now if they are required.

draft-lcurley-warp.md
```diff
@@ -210,21 +209,43 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
+Track URI = "moq" "://" Track Namespace  "/"  Track Name
+~~~~~~~~~~~~~~~
+
+"Track Namespace" MUST identify a globaly unique identifier, such as domain name or something equivalent. This is followed by the application context specific "Track Name", encoded as an opaque string. 
+
+~~~
+Example: 1
+Track Namespace = acme.meetings.com
+Track Name = meeting123/audio
+Track URI = moq://meetings.com/meeting123/audio
```
afrind, line 238: I think you mean moq://acme.meetings.com/meeting123/audio

draft-lcurley-warp.md
afrind: There's still a typo uncompress -> uncompressed

draft-lcurley-warp.md
```diff
@@ -180,11 +179,11 @@ x (b):
 
 # Model
 
-## Objects
+## Objects {#model-object}
 
 The basic element of Warp is an *object*. An object is a single addressable
 cacheable unit whose payload is a sequence of bytes.  An object MAY depend on other 
-objects to be decoded. An object MUST belong to a group {{groups}}. Objects carry 
+objects to be decoded. An object MUST belong to a group {{model-group}}. Objects carry
```
afrind, line 186: I still think this MUST implies somehow that implementers have a choice (eg: somehow they could make an object that isn't in a group, and that would be an error).  Instead you can instead simply state "all objects belong to a group", and make Group a required parameter in the OBJECT header.

afrind (commented):
Individual Review

draft-lcurley-warp.md
suhasHere: ah ! you are right. I will fix them

draft-lcurley-warp.md
suhasHere: @afrind this PR doesn't change the text for object and groups. It just updated the section reference to the group and rest of the text is unchanged.

Regardless,I agree your suggestion makes sense and we should update it so.

suhasHere:
@afrind PR has been updated to reflect your most recent feedback. thanks

wilaw:
What utility is the URI definition serving? It is not used for the physical connection (WebTransport or QUIC). The 'moq://' protocol designator is not used elsewhere in the spec. Wouldn't we simply store the namespace and name as separate headers in every object?

```
Track Namespace = acme.meetings.com
Track Name = meeting123/audio
```

If we want to store them as a single item, then it could be stored as

`Track Name: acme.meetings.com/meeting123/audio`

with the rule that the first / separates the namespace from the name.

draft-lcurley-warp.md
```diff
@@ -210,21 +209,43 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
+
+### Track URI {#track-uri}
+
+Tracks are identified by a globally unique identifier, called "Track URI" with the scheme shown below:
+
+~~~~~~~~~~~~~~~
+Track URI = "moq" "://" Track Namespace  "/"  Track Name
+~~~~~~~~~~~~~~~
+
+"Track Namespace" MUST identify a globaly unique identifier, such as domain name or something equivalent. This is followed by the application context specific "Track Name", encoded as an opaque string.
```
vasilvv, line 232: I am a bit concerned about this MUST, since we don't actually enforce it on the protocol level.  We should also specify who is the target audience of the MUST.  How about something like:

"This document does not define the exact mechanism of naming track namespaces. Applications building on top of MoQ MUST ensure that the mechanism used guarantees global uniqueness; for instance, an application could use domain names as track namespaces."

draft-lcurley-warp.md
```diff
@@ -210,21 +209,43 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
+Example: 1
+Track Namespace = acme.meetings.com
+Track Name = meeting123/audio
+Track URI = moq://acme.meetings.com/meeting123/audio
+
+Example: 2
+Track Namespace = livestream.tv
+Track Name = uaCafDkl123/audio
+Track URI = moq://livestream.tv/uaCafDkl123/audio
+~~~
```
vasilvv, line 244: Please switch both examples to use example names from RFC 2606.

draft-lcurley-warp.md
```diff
@@ -210,21 +209,43 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
 
-A media track in Warp is a combination of *an init object* and a sequence of media object groups. An init object is a format-specific self-contained description of the track that is required to decode any media object contained within the track, but can also be used as the metadata for track selection. If two media tracks carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
+### Connection URL
 
-## Track Bundle
-A track bundle is a collection of tracks intended to be delivered together.
-Objects within a track bundle may be prioritized relative to each other via the delivery order property.
-This allows objects to be prioritized within a track (ex. newer > older) and between tracks (ex. audio > video).
-The track bundle contains a catalog indicating the available tracks.
+Each track MAY have an associated hop-by-hop Connection URL that specifies the network host to setup the transport connection. The syntax of the Connection URL and the associated connection setup procedures are specific to the underlying transport protocol usage {{transport-usages}}.
```
vasilvv, line 248: I believe we want to allow more than one connection URLs, for redundancy and multi-CDN support.  How about "have one or more associated connection URLs specifying network hosts through which a track may be accessed"?

draft-lcurley-warp.md
```diff
@@ -210,21 +209,43 @@ decrypt or authenticate an object until it is fully present.  Allowing Objects
...
+A Track is the central concept within the MoQ Transport protocol for delivering media and is made up of sequence of objects ({{model-object}}) organized in the form of groups ({{model-group}}).
+
+A track is a transform of a uncompressed media or metadata using a specific encoding process, a set of parameters for that encoding, and possibly an encryption process. The MoQ Transport protocol is designed to transport tracks.
+
+### Track URI {#track-uri}
+
+Tracks are identified by a globally unique identifier, called "Track URI" with the scheme shown below:
+
+~~~~~~~~~~~~~~~
+Track URI = "moq" "://" Track Namespace  "/"  Track Name
```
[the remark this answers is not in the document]
suhasHere: We agreed to make it track URI ( referring to Cullen's notes from the meeting) and decided to have a starting part of it to be globally unique namepsace and then track name.

i can remove moq:// if needed, but I strongly suggest we have tracks identified by globally unique identifiers.

suhasHere:
@vasilvv most recent commit must address your feedback

vasilvv (approved):

Issue #102, opened 2023-03-15
Refreshing Subscriptions


suhasHere:
We need to support subscription refreshes to ensure liveness and also signal explicit consent. The latter will be needed from security perspective and also protecting relay from resource exhaustions.
The protocol should also recommend what the Relays should do in such cases

vasilvv:
Consent freshness can be provided via QUIC itself, we just don't do that by default.  See w3c/webtransport#174 for some discussion we had regarding consent freshness in WebTransport (the short answer is that the server can send a PATH_CHALLENGE regularly).

Is there any specific reason you want this to be done at a subscription level?

suhasHere:
Thanks for the reference victor .. This is an explicit request from the client telling to Relay of the continuiing interest for a given set of subscriptions and to avoid Relays from needing to hold onto the subscription.

afrind:
It makes sense that a relay would need to reap upstream subscriptions when all the downstream subscribers are gone.  Is it better to automatically expire subscriptions and require the client to refresh, or automatically extend subscriptions and require the client to unsubscribe, and terminate all active subscriptions when the underlying WebTransport session or  QUIC connection closes?

What are the pros and cons of each approach?  Should the base protocol support both modes?

suhasHere:
> and terminate all active subscriptions when the underlying WebTransport session or QUIC connection closes?

The control stream PR proposes a solution for this. Subscriptions of one or more tracks, are tied to a control stream and when that streams closes, it closes all the linked media transfer state.

suhasHere:
> Is it better to automatically expire subscriptions and require the client to refresh

Relays always have a choice to terminate any subscription if the local policy decides. In those cases, if a client is still interested , it should have to renew/refresh the subscription if it has not already done. It does cause media disruptions at the clients to be torn down its media and having a mechanism to renew will be useful.

It is good security feature for relays to be told I am still interested in receiving media so to prevent DOS of relay's resources for maintaining the needed state.

afrind:
As an individual:

Continual subscription sounds expensive to me, so I'd prefer that not be the only mode.  I think there are tradeoffs to explicit cancellation client vs implicit cancellation.  If the publisher controls the implicit subscription timeout and they can set it to t = ~infinity it's kind of the same thing.

suhasHere:
> If the publisher controls the implicit subscription timeout and they can set it to t = ~infinity it's kind of the same thing.

[...]

PR #122, opened 2023-04-05
IETF 116:  Update Subscribe Messages

suhasHere:
This PR depends on #121 and has the following changes  to capture IETF 116 decisions 

1. Rename subscribe message to subscribe request 
2. Include track names and authorization in the subscribe request
3. Add  basic support for subscribe reply providing the results of subscription
4. Add support for short compression identifiers as Track Id (varint) to be used in the object header messages

Currently the subscribe granularity is per track and a future PR can be modified to support multiple tracks if needed.

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -169,11 +169,6 @@ Track Name:
 
 : See {{track-name}}.
 
-Track ID:
-
-: An hop-by-hop variable length integer for a Track Name that is carried in Object message header. Track ID is setup
-as part of Publish/Subscribe control message exchanges. See {{message-subscribe-reply}}
-
 Provider: 
 
 : Entity capable of hosting media application
@@ -678,9 +673,12 @@ The `TRACK INFO` structure is as defined below
 ~~~
 TRACK INFO {
   Track Name Length(i),
-  Track Name(...)...,
+  Track Name(...),
+  Track ID (i),
   Track Authorization Info Length(i),
-  Track Authorization Info (...)...
+  Track Authorization Info (...),
+  [Group Sequence (i)],
+  [Object Sequence (i)]
 }
 ~~~
 {: #warp-track-info-format title="Warp TRACK INFO Content"}
@@ -688,6 +686,16 @@ TRACK INFO {
 
 * Track Name:
 Identifies the fully qualified track name as defined in ({{model-track}}).
+
+* Track ID: 
+A proposal serving as short hop-by-hop identifier to be used in the OBJECT ({{message-object}}) message header for the requested tack. Given that the media corresponding to a track can potentially be delivered over multiple data streams, the Track ID provides the necessary mapping between the "Track Name" in the control message and the corresponding data streams. It also serves as a compression identifier for containing the size of object headers instead of carrying complete "Track Name" information in every object message. A client MUST choose an Track ID value that is unique within a QUIC Connection.
+
+* Group Sequence: 
+Identifies the group within the track to start delivering the objects. This field is optional and when omitted, the publisher MUST deliver the objects from the most recent group.
+
+* Object Sequence: 
+Identifies the object within the track to start the media delivery. The `Group Sequence` field MUST be set to identify the group under which to
+the deliver the objects. This field is optional and when omitted, the publisher MUST deliver starting from the beginning of the selected group.
 
 * Track Authorization Info: 
 Carries track authorization, typically in the form of a token, authorizing the client’s subscription to the track. The specifics of obtaining the authorization information is out of scope for this specification.
@@ -700,8 +708,6 @@ Subscriptions stay active until one of the following happens:
 
 On successful subscription, the publisher MUST deliver objects
 from the beginnig of the most recent group in the track, i.e, starting at Object Sequence of 0.
-
-A client MUST renew its subscriptions by sending a new `SUBSCRIBE` to keep them active. Such subscriptions MUST refresh the existing subscriptions and thus only the most recent SUBSCRIBE message is considered active. A renewal period of 5 seconds is RECOMMENDED.
 
 ## SUBSCRIBE REPLY {#message-subscribe-reply}
 
@@ -744,11 +750,12 @@ Identifies the track in the request message for which this
 response is provided.
 
 * Response:
-Provides result of the transaction, where a value of `Ok` indicates successful subscription. For failed response, the "Reason Phrase" is populated with appropriate reason code.
+Provides result of the transaction, where a value of `Ok` indicates successful subscription. For failed response, the "Reason Phrase" 
+is populated with appropriate reason.
+
 
 * Track ID:
-A short hop-by-hop identifier to be used in the OBJECT ({{message-object}}) message header for the requested tack. Given that the media corresponding to a track can potentially be delivered over multiple data streams, the `Track ID` provides the necessary mapping between the "Track Name" in the control message and the corresponding data streams. It also serves as a compression identifier for containing the size of object headers instead of carrying complete "Track Name" information in every object message. The `Track ID` chosen MUST be unique within the QUIC Connection and is populated for successful subscriptions.
-
+Identifies the hop-by-hop identifier mapping the given Track Name to be populated in the Object messages ({{message-object}}) and MUST only be present if the `Response` is `Ok`. This field is populated with either the `Track ID` value provided in the request or the one chosen by client processing the request. The Track ID field in the Object's header messages MUST be populated with the value in this field. 
 
 ## GOAWAY {#message-goaway}
 The `GOAWAY` message is sent by the server to force the client to reconnect.
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -741,7 +741,7 @@ SUBSCRIBE REQUEST Message {
 
 
 * Track URI:
-Identifies the track as defined in ({{model-track}}).
+Identifies the track as defined in ({{track-uri}}).
 
 * Track ID: 
 Session specific identifier that maps the Track URI to the Track ID in OBJECT ({{message-object}}) message headers for the advertised track. Peer processing the request message MAY end up choosing a different Track ID (see {{message-subscribe-ok}}). Track IDs are generally shorter than Track URIs and thus reduce the overhead in OBJECT messages.
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -805,8 +805,8 @@ enum Track Result
 }
 
 TRACK RESPONSE {
-  Track Name Length(i),
-  Track Name(...)...,
+  Track URI Length(i),
+  Track URI(...),
   Track Result Response,
   [Track ID(i)],
   [ Reason Phrase Length (i) ],
@@ -816,7 +816,7 @@ TRACK RESPONSE {
 {: #warp-track-response-format title="Warp TRACK RESPONSE Content"}
 
 
-* Track Name:
+* Track URI:
 Identifies the track in the request message for which this
 response is provided.
 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -57,8 +57,6 @@ informative:
     date: 2020-03
   NewReno: RFC6582
   BBR: I-D.cardwell-iccrg-bbr-congestion-control-02
-  MOQ-ARCH: I-D.nandakumar-moq-arch-00
-
 
 --- abstract
 
@@ -734,49 +732,44 @@ A future draft will add the ability to add/remove/update tracks.
 
 Entities that intend to receive media will do so via subscriptions to one or more tracks. The information for the tracks can be obtained via catalog or from incoming SUBSCRIBE REQUEST messages.
 
-The format of SUBSCRIBE is as follows:
+The format of SUBSCRIBE REQUEST is as follows:
 
 ~~~
 SUBSCRIBE REQUEST Message {
   TRACK INFO Track
 }
 ~~~
-{: #warp-subscribe-format title="Warp SUBSCRIBE Message"}
+{: #warp-subscribe-format title="Warp SUBSCRIBE REQUEST Message"}
 
 * Track:
 Identifies the track information as defined in `TRACK INFO`.
 
-The `TRACK INFO` structure is as defined below
+The `TRACK INFO` structure is defined as below:
 
 ~~~
+TrackInfo Parameter {
+  TrackInfo Parameter ID (i),
+  TrackInfo Parameter Length (i),
+  TrackInfo Parameter Value (..),
+}
+
 TRACK INFO {
-  Track Name Length(i),
-  Track Name(...),
+  Track URI Length(i),
+  Track URI(...),
   Track ID (i),
-  Track Authorization Info Length(i),
-  Track Authorization Info (...),
-  [Group Sequence (i)],
-  [Object Sequence (i)]
+  TrackInfo Parameters (..) ...
 }
 ~~~
 {: #warp-track-info-format title="Warp TRACK INFO Content"}
 
 
-* Track Name:
+* Track URI:
 Identifies the fully qualified track name as defined in ({{model-track}}).
 
 * Track ID: 
-A proposal serving as short hop-by-hop identifier to be used in the OBJECT ({{message-object}}) message header for the requested tack. Given that the media corresponding to a track can potentially be delivered over multiple data streams, the Track ID provides the necessary mapping between the "Track Name" in the control message and the corresponding data streams. It also serves as a compression identifier for containing the size of object headers instead of carrying complete "Track Name" information in every object message. A client MUST choose an Track ID value that is unique within a QUIC Connection.
+A short hop-by-hop identifier to be used in the OBJECT ({{message-object}}) message header for the requested track. Given that the media corresponding to a track can potentially be delivered over multiple data streams, the Track ID provides the necessary mapping between the "Track Name" in the control message and the corresponding data streams. It also serves as a compression identifier for containing the size of OBJECT message header instead of carrying complete "Track URI" information in every OBJECT message. A client MUST choose an Track ID value that is unique within a QUIC Connection.
 
-* Group Sequence: 
-Identifies the group within the track to start delivering the objects. This field is optional and when omitted, the publisher MUST deliver the objects from the most recent group.
-
-* Object Sequence: 
-Identifies the object within the track to start the media delivery. The `Group Sequence` field MUST be set to identify the group under which to
-the deliver the objects. This field is optional and when omitted, the publisher MUST deliver starting from the beginning of the selected group.
-
-* Track Authorization Info: 
-Carries track authorization, typically in the form of a token, authorizing the client’s subscription to the track. The specifics of obtaining the authorization information is out of scope for this specification.
+TrackInfo Parameters are defined in {{track-params}}.
 
 Subscriptions stay active until one of the following happens:
 
@@ -785,7 +778,7 @@ Subscriptions stay active until one of the following happens:
 - the underlying transport is disconnected.
 
 On successful subscription, the publisher MUST deliver objects
-from the beginnig of the most recent group in the track, i.e, starting at Object Sequence of 0.
+from the group sequence and object sequence as defined in {{track-params}}.
 
 ## SUBSCRIBE REPLY {#message-subscribe-reply}
 
@@ -829,11 +822,10 @@ response is provided.
 
 * Response:
 Provides result of the transaction, where a value of `Ok` indicates successful subscription. For failed response, the "Reason Phrase" 
-is populated with appropriate reason.
-
+is populated with appropriate reason and Reason Phrase Length carrying its length.
 
 * Track ID:
-Identifies the hop-by-hop identifier mapping the given Track Name to be populated in the Object messages ({{message-object}}) and MUST only be present if the `Response` is `Ok`. This field is populated with either the `Track ID` value provided in the request or the one chosen by client processing the request. The Track ID field in the Object's header messages MUST be populated with the value in this field. 
+Identifies the hop-by-hop identifier mapping the given Track Name to be populated in the OBJECT messages ({{message-object}}) and MUST only be present if the `Response` is `Ok`. This field is populated with either the `Track ID` value provided in the request or the one chosen by client processing the request. The Track ID field in the OBJECT's header messages MUST be populated with the value in this field. 
 
 ## GOAWAY {#message-goaway}
 The `GOAWAY` message is sent by the server to force the client to reconnect.
@@ -879,6 +871,20 @@ The ROLE parameter (key 0x00) allows the client to specify what roles it expects
 
 The client MUST send a ROLE parameter with one of the three values specified above. The server MUST close the connection if the ROLE parameter is missing, is not one of the three above-specified values, or it is different from what the server expects based on the application in question.
 
+# TrackInfo Parameters {#track-params}
+
+The TrackInfo parameters are described as below. Every parameter MUST appear at most once. The peers SHOULD verify that and close the connection if a parameter appears more than once.
+The Parameter Value Length field indicates the length of the Parameter Value.
+
+### Group Sequence
+
+The Group Sequence (key 0x00) identifies the group within the track to start delivering the objects. The publisher MUST deliver the objects from the most recent group, when this parameter is omitted.
+
+### Object Sequence 
+The Object Sequence (key 0x01) identifies the object within the track to start the media delivery. The `Group Sequence` parameter MUST be set to identify the group under which to the deliver the objects. The publisher MUST deliver starting from the beginning of the selected group when this parameter is omitted.
+
+### Track Authorization Info
+Track Authorization Info (key 0x02) identifies the mandatory parameter carrying track's authorization, typically in the form of a token, authorizing the client’s subscription to the track. The specifics of obtaining the authorization information is out of scope for this specification.
 
 # Containers
 The container format describes how the underlying codec bitstream is encoded.
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -908,6 +908,7 @@ TODO: fill out currently missing registries:
 * Warp version numbers
 * SETUP parameters
 * Track Request parameters
+* Subscribe Error codes
 * Track format numbers
 * Message types
 * Object headers
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -847,7 +847,7 @@ The client MUST send a ROLE parameter with one of the three values specified abo
 
 # Track Request Parameters {#track-req-params}
 
-The Track Request Parameters identify properties of the track requesting in either the PUBLISH REQUEST or SUSBCRIBE REQUEST control messages. Every parameter MUST appear at most once. The peers MUST close the connection if there are duplicates. The Parameter Value Length field indicates the length of the Parameter Value.
+The Track Request Parameters identify properties of the track requested in either the PUBLISH REQUEST or SUSBCRIBE REQUEST control messages. Every parameter MUST appear at most once. The peers MUST close the connection if there are duplicates. The Parameter Value Length field indicates the length of the Parameter Value.
 
 ### GROUP SEQUENCE Parameter
 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -605,7 +605,7 @@ OBJECT Message {
 {: #warp-object-format title="Warp OBJECT Message"}
 
 * Track ID:
-The track identifier as declared in CATALOG ({{message-catalog}}).
+The track identifier obtained as part of subscription and/or publish control message exchanges.
 
 * Group Sequence :
 An integer always starts at 0 and increases sequentially at the original media publisher.
@@ -711,7 +711,7 @@ the objects. This field is optional and when omitted, the
 publisher MUST deliver the objects from the most recent 
 group.
 
-* Object ID:
+* Object Sequence:
 Identifies the object within the track to start the media 
 delivery. The group MUST either match the `Group Sequence` if provided
 or MUST be the most recent group. This field is optional and 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -169,6 +169,11 @@ Track Name:
 
 : See {{track-name}}.
 
+Track ID:
+
+: An hop-by-hop variable length integer for a Track Name that is carried in Object message header. Track ID is setup
+as part of Publish/Subscribe control message exchanges. See {{message-subscribe-reply}}
+
 Provider: 
 
 : Entity capable of hosting media application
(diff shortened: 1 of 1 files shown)


draft-lcurley-warp.md
```diff
@@ -603,7 +605,7 @@ OBJECT Message {
 {: #warp-object-format title="Warp OBJECT Message"}
 
 * Track ID:
-The track identifier as declared in CATALOG ({{message-catalog}}).
+The track identifier obtained as part of subscription and/or publish control message exchanges.
```
afrind, line 608: I think it would be good to explain Track Name vs Track ID with some expository text where tracks are defined (maybe in #121).

Then here perhaps you can say "Track ID: unique identifier for this track within the current (session? connection?)"

draft-lcurley-warp.md
```diff
@@ -664,29 +666,116 @@ This contains base information required to decode OBJECT messages, such as codec
 An endpoint MUST NOT send multiple CATALOG messages.
 A future draft will add the ability to add/remove/update tracks.
 
-## SUBSCRIBE {#message-subscribe}
-After receiving a CATALOG message ({{message-catalog}}, the receiver sends a SUBSCRIBE message to indicate that it wishes to receive the indicated tracks.
+## SUBSCRIBE REQUEST {#message-subscribe-req}
+
+Entities that intend to receive media will do so via subscriptions to
+one or more tracks. The information for the tracks can be obtained via catalog or from incoming SUBSCRIBE messages.
```
afrind, line 672: Did you mean SUBSCRIBE_REQUEST?  SUBSCRIBE is no longer defined?

draft-lcurley-warp.md
```diff
@@ -664,29 +666,116 @@ This contains base information required to decode OBJECT messages, such as codec
...
+  Track Authorization Info (...)...
+  [Group Sequence (i)],
+  [Object Sequence (i)],
+}
+~~~
+{: #warp-track-info-format title="Warp TRACK INFO Content"}
+
+
+* Track Name:
+Identifies the fully qualified track name as defined in ({{model-track}}).
```
afrind, line 703: Does it always have to be fully qualified, or can that vary by "Streaming format" or another profile.  My understanding is that some use cases would like to only send the suffix here, and combine it with a prefix communicated elsewhere.  Is that possible?

draft-lcurley-warp.md
```diff
@@ -664,29 +666,116 @@ This contains base information required to decode OBJECT messages, such as codec
...
+
+* Track ID:
```
afrind, line 705: Is it better to have the publisher assign the track ID in the SUBSCRIBE_REPLY or the subscriber in the SUBSCRIBE_REQUEST?  Are there any tradeoffs?

fluffy (approved):
This seems about right.

draft-lcurley-warp.md
```diff
@@ -664,29 +666,116 @@ This contains base information required to decode OBJECT messages, such as codec
...
+A proposal serving as short hop-by-hop identifier to be used in the OBJECT ({{message-object}}) message header for the requested tack. Given that the media corresponding to a track can potentially be delivered over multiple data streams, the `Track ID` provides the necessary mapping between the "Track Name" in the control message and the corresponding data streams. It also serves as a compression identifier for containing the size of object headers instead of carrying complete "Track Name" information in every object message. A client MUST choose an `Track ID` value that is unique within a QUIC Connection.
+
+* Group Sequence:
+Identifies the group within the track to start delivering 
+the objects. This field is optional and when omitted, the
+publisher MUST deliver the objects from the most recent 
+group.
+
+* Object Sequence:
```
afrind, line 714: I don't think the layout can have both Group and Object sequence be optional -- if there's only one varint present in the message it's implicitly a Group Sequence, right?  

I read your intent to support these three cases:

1. No group/object info - send all groups/objects starting from beginning of most recent group, object 0
2. Group sequence - send from this specific group, starting from object 0
3. Group sequence and Object sequence - send only this specific object from this specific group

Is that right?

draft-lcurley-warp.md
```diff
@@ -664,29 +666,116 @@ This contains base information required to decode OBJECT messages, such as codec
...
+beginning of the selected group, i.e., starting at Object Sequence of 0.
+
+* Track Authorization Info: 
+Carries track authorization, typically in the form of a token, authorizing client’s subscription to the track. The specifics of obtaining the authorization information is out of scope for this specification.
+
+Subscriptions stay active until one of the following happens:
+
+- a client local policy dictates expiration of a subscription.
+- optionally, a server policy dicates subscription expiration.
+- the underlying transport is disconnected.
```
afrind, line 728: If we want automatic unsubscribe, then I think SETUP or SUBSCRIBE_REPLY needs to exchange information about the publisher's required refresh period.

I also think we want an explicit UNSUBSCRIBE message, in case the publishers refresh interval is much longer than the receiver wants to continue getting this track.

draft-lcurley-warp.md
```diff
@@ -664,29 +666,116 @@ This contains base information required to decode OBJECT messages, such as codec
...
+{
+  TRACK RESPONSE Track
+}
+~~~
+{: #warp-subscribe-reply format title="Warp SUBSCRIBE REPLY Message"}
+
+* Track:
+Captures the result of subscription as defined in `TRACK RESPONSE`
+
+The `TRACK RESPONSE` structure is as defined below
```
afrind, line 747: Is TRACK RESPONSE going to be reused elsewhere (maybe PUBLISH_REQUEST/RESPONSE)?  If not, just merge the fields there into SUBSCRIBE_REPLY?

draft-lcurley-warp.md
```diff
@@ -664,29 +666,116 @@ This contains base information required to decode OBJECT messages, such as codec
...
+
+~~~
+enum Track Result
+{
+  Ok(0),
+  Expired(1),
```
afrind, line 753: It's too early to bikeshed error codes, but I'm curious if you called out Expired here separately from Fail so that a publisher can send a SUBSCRIBE_REPLY once the track expires?  Eg, are you envisioning a sequence like:

Consumer -> Publisher: SUBSCRIBE_REQUEST
P -> C: SUBSCRIBE_REPLY
P -> C: OBJECT*
...
P -> C: SUBSCRIBE_REPLY (Expired)

draft-lcurley-warp.md
```diff
@@ -664,29 +666,116 @@ This contains base information required to decode OBJECT messages, such as codec
...
+
+* Track Name:
+Identifies the track in the request message for which this
+response is provided.
+
+* Response:
+Provides result of the transaction, where a value of `Ok` indicates successful subscription. For failed or expired responses, the "Reason Phrase" shall be populated with appropriate reason code.
+
+* Track ID:
+Identifies the hop-by-hop identifier mapping the given Track Name to be populated in the Object messages ({{message-object}}). This field is optional and is provided in cases where the end point is unable to use the `Track ID` proposed by the peer in the request message. If populated, the `Track ID` field in the Object's header messages MUST be populated with the value in this field otherwise the `Track ID` value MUST correspond to one in the request message for a given track.
```
afrind, line 777: If the publisher has full control over the ID space I think I would recommend against letting the client ask for particular IDs.  Do you have a use case in mind or would publisher-assigned work?

draft-lcurley-warp.md
```diff
@@ -664,29 +666,116 @@ This contains base information required to decode OBJECT messages, such as codec
...
+~~~
+{: #warp-track-response-format title="Warp TRACK RESPONSE Content"}
+
+
+* Track Name:
+Identifies the track in the request message for which this
+response is provided.
+
+* Response:
+Provides result of the transaction, where a value of `Ok` indicates successful subscription. For failed or expired responses, the "Reason Phrase" shall be populated with appropriate reason code.
```
afrind, line 774: I think shall is a normative verb, since it's optional, you can use MAY or can.  Also maybe say "subscription request" rather than "transaction".

afrind (commented):
Individual review

draft-lcurley-warp.md
suhasHere: good catch, will update

draft-lcurley-warp.md
suhasHere: We discussed this topic in length at the IETF116 in the context of WT and Native QUIC and why having full name will work for both cases uniformly and also for inter distribution network flows.

We concluded that  tracks are central component and they are self identifying and are carry that way in subscribes/publishes.

suhasHere:
@afrind 

Good points Alan on the comments. I will respond and update text to reflect these by early next week.

draft-lcurley-warp.md
afrind: @vasilvv @kixelated will this meet your use cases?  My understanding of Mo's proposal on the mailing list was specifically so that suffix only *could* be specified here -- both suffix only or fully qualified are OK.

draft-lcurley-warp.md
afrind: I'm not necessarily hung up on which PR defines it - only the name vs id concept deserves top level mention earlier in the document.

draft-lcurley-warp.md
suhasHere: yes it is used across Publish and Subscribe message .. same with Track Info too

draft-lcurley-warp.md
suhasHere: I think we did have a discussion at the mic on this aspect and we agreed to keep the behavior of subscribe and publish unifirm across all the nodes, it is best to keep the track as self identifying without tying it to the connection context.  Since this will allow 2 things
1.  Work in the same way across WT and native QUIC. For WT there is extra bytes of duplication which is not concerning
2. Nodes processing subscribe/publish can be uniformly defined without needing to explcitly specify the cases where one option would work vs the other. 
3. Tracks are authorized and one can define security properties in a common way across applications and distribution networks

draft-lcurley-warp.md
suhasHere: Good question @afrind 

In the current model, the publisher has the final say on the trackId to be used. 

The benefit of Subscriber Request proposing the TrackId is it will lead to a design where one can minimize the rewrites of the object headers per hop if there is no conflict. 

Again the publisher is the one that can decide the trackId and it will do so in the SUBSCRIBE_REPLY.

draft-lcurley-warp.md
suhasHere: @afrind  great points. I agree on both the points. I will add issues to cover those.

draft-lcurley-warp.md
suhasHere: +1 .. will update

draft-lcurley-warp.md
suhasHere: the only benefit is minimize object header rewrites. May be very are optimizing too early. i am fine with making it publisher controlled id space.

draft-lcurley-warp.md
suhasHere: #133 , #132  created these issues

draft-lcurley-warp.md
suhasHere: Removed expired for now. We can revisit later if its needed.

suhasHere:
@afrind pushed few changes to address your review feedback. please let me know thanks

draft-lcurley-warp.md
```diff
@@ -169,6 +169,11 @@ Track Name:
 
 : See {{track-name}}.
 
+Track ID:
```
kixelated, line 172: I don't think this should be in the definitions. I think Alan was asking for a section on how name/URL maps to ID.

draft-lcurley-warp.md
```diff
@@ -650,29 +657,97 @@ This contains base information required to decode OBJECT messages, such as codec
...
+Subscriptions stay active until one of the following happens:
+
+- a client local policy dictates expiration of a subscription.
+- optionally, a server policy dicates subscription expiration.
+- the underlying transport is disconnected.
+
+On successful subscription, the publisher MUST deliver objects
+from the beginnig of the most recent group in the track, i.e, starting at Object Sequence of 0.
+
+A client MUST renew its subscriptions by sending a new `SUBSCRIBE` to keep them active. Such subscriptions MUST refresh the existing subscriptions and thus only the most recent SUBSCRIBE message is considered active. A renewal period of 5 seconds is RECOMMENDED.
```
kixelated, line 704: This is a hard no from me as discussed in #102 .

I would be okay with the server indicating a subscription expiration in the SUBSCRIBE_REPLY message. That way the consumer (nit: not client) would resubscribe based on the policy set by the publisher. 

I would still want to talk about the use-case though; what problem are you solving?

draft-lcurley-warp.md
```diff
@@ -650,29 +657,97 @@ This contains base information required to decode OBJECT messages, such as codec
...
+* Track Authorization Info: 
+Carries track authorization, typically in the form of a token, authorizing the client’s subscription to the track. The specifics of obtaining the authorization information is out of scope for this specification.
+
+Subscriptions stay active until one of the following happens:
+
+- a client local policy dictates expiration of a subscription.
+- optionally, a server policy dicates subscription expiration.
+- the underlying transport is disconnected.
+
+On successful subscription, the publisher MUST deliver objects
```
kixelated, line 701: The group wouldn't start at 0 when joining a live stream. The relay should not rewrite object headers either, as it will break any form of history (ex. catalog that indicates group 45 = timestamp 12:34)

draft-lcurley-warp.md
kixelated: I would encode an "expires" in the SUBSCRIBE_REPLY. It would be a duration (not a timestamp) so the consumer would need to refresh the subscription a little early.

draft-lcurley-warp.md
```diff
@@ -650,29 +657,97 @@ This contains base information required to decode OBJECT messages, such as codec
...
+{
+  Ok(0),
+  Fail(1)
+}
+
+TRACK RESPONSE {
+  Track Name Length(i),
+  Track Name(...)...,
+  Track Result Response,
+  [Track ID(i)],
```
kixelated, line 734: I've mentioned this in a few PRs, but these optionals are not possible to encode.

draft-lcurley-warp.md
```diff
@@ -650,29 +657,97 @@ This contains base information required to decode OBJECT messages, such as codec
...
+enum Track Result
+{
+  Ok(0),
+  Fail(1)
+}
+
+TRACK RESPONSE {
+  Track Name Length(i),
+  Track Name(...)...,
+  Track Result Response,
```
kixelated, line 733: varint?

draft-lcurley-warp.md
kixelated: And why is this optional in the first place?

draft-lcurley-warp.md
```diff
@@ -650,29 +657,97 @@ This contains base information required to decode OBJECT messages, such as codec
...
+
+* Track Name:
+Identifies the track in the request message for which this
+response is provided.
+
+* Response:
+Provides result of the transaction, where a value of `Ok` indicates successful subscription. For failed response, the "Reason Phrase" is populated with appropriate reason code.
+
+* Track ID:
+A short hop-by-hop identifier to be used in the OBJECT ({{message-object}}) message header for the requested tack. Given that the media corresponding to a track can potentially be delivered over multiple data streams, the `Track ID` provides the necessary mapping between the "Track Name" in the control message and the corresponding data streams. It also serves as a compression identifier for containing the size of object headers instead of carrying complete "Track Name" information in every object message. The `Track ID` chosen MUST be unique within the QUIC Connection and is populated for successful subscriptions.
```
kixelated, line 750: I think the consumer should choose the track ID in the SUBSCRIBE message.

Otherwise, there's a race condition where the SUBSCRIBE_REPLY could be delayed, while the consumer receives objects with an unknown track ID. It's a whole lot simpler if the consumer chooses the ID in the SUBSCRIBE message.

draft-lcurley-warp.md
suhasHere: I see .. I did have it as part of the messages section. May be @afrind  can you see the message section changes and see if it still works.

draft-lcurley-warp.md
suhasHere: I don't think I mentioned group starting at 0 .. I meant object starting at 0 form the most recent group

draft-lcurley-warp.md
suhasHere: A failed subscription doesn't need a track Id .. Options are very common QUIC encoding. Can you explain what are the challenges you see ?

draft-lcurley-warp.md
```diff
@@ -650,29 +657,97 @@ This contains base information required to decode OBJECT messages, such as codec
...
+~~~
+{: #warp-track-response-format title="Warp TRACK RESPONSE Content"}
+
+
+* Track Name:
+Identifies the track in the request message for which this
+response is provided.
+
+* Response:
+Provides result of the transaction, where a value of `Ok` indicates successful subscription. For failed response, the "Reason Phrase" is populated with appropriate reason code.
```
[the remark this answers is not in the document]
suhasHere: +1 .. will upadte

draft-lcurley-warp.md
kixelated: I called this "Subscription ID" in a previous PR. It might be less confusing since it's decoupled from track.

draft-lcurley-warp.md
kixelated: I missed the `i.e` when reading this, which should be `ex.` anyway...

I think the example is fine just make up a number like 53.

draft-lcurley-warp.md
kixelated: So I think I understand what you're trying to do in this instance, but you need more text. There's no magic way to make a field optional in QUIC encoding.

How about:
> The track ID field MUST ONLY be present if the result is Ok. The reason and length MUST ONLY be present if the result is not Ok.

I would just make a separate messages honestly. SUBSCRIBE_OK and SUBSCRIBE_ERROR. If we want to consolidate them in the future, we can add a flag for the type (like how QUIC frames work).

draft-lcurley-warp.md
suhasHere: I did call it "Subscription ID" in #96  and there was some push backs on that term IIRC . This is just naming, I am fine with either of the names. We need to pick one and move on

kixelated: I don't see any push-back in that PR around the name. There was a LOT of other stuff in that PR. 

And yeah I called it "Subscribe ID" in #43. I think "Subscription ID" is better.

draft-lcurley-warp.md
suhasHere: i will remove it from this PR and we can discuss in #102

draft-lcurley-warp.md
suhasHere: i will revert some of these for a fresh discussion

suhasHere:
@fluffy @kixelated @vasilvv @afrind please give it another read and let me know . thanks

draft-lcurley-warp.md
afrind: I think this discussion is captured in #121 for now and we can update based on whatever the outcome is there.

draft-lcurley-warp.md
```diff
@@ -650,30 +652,110 @@ This contains base information required to decode OBJECT messages, such as codec
...
+}
+~~~
+{: #warp-track-info-format title="Warp TRACK INFO Content"}
+
+
+* Track Name:
+Identifies the fully qualified track name as defined in ({{model-track}}).
+
+* Track ID: 
+A proposal serving as short hop-by-hop identifier to be used in the OBJECT ({{message-object}}) message header for the requested tack. Given that the media corresponding to a track can potentially be delivered over multiple data streams, the Track ID provides the necessary mapping between the "Track Name" in the control message and the corresponding data streams. It also serves as a compression identifier for containing the size of object headers instead of carrying complete "Track Name" information in every object message. A client MUST choose an Track ID value that is unique within a QUIC Connection.
```
afrind, line 691: Editorial comments:

I don't think you need to say proposal:

> An session specific identifier used in OBJECT ... message headers for the requested track.

"provides the necessary mapping between" -> maps

> ... the Track ID maps the "Track Name" ...

Restating the compression bit more succinctly:

> Track IDs are generally shorter than Track Names and thus reduce the overhead in OBJECT messages.

afrind, line 691: Architectural Comment:

I'd prefer to just let the server pick the IDs rather than giving the client an option to ask for a specific one.  If we remove that functionality from this PR we can move this forward and have a separate PR to discuss the pros and cons of client hinting track IDs there.

draft-lcurley-warp.md
```diff
@@ -650,30 +652,110 @@ This contains base information required to decode OBJECT messages, such as codec
...
+
+* Group Sequence: 
+Identifies the group within the track to start delivering the objects. This field is optional and when omitted, the publisher MUST deliver the objects from the most recent group.
+
+* Object Sequence: 
+Identifies the object within the track to start the media delivery. The `Group Sequence` field MUST be set to identify the group under which to
+the deliver the objects. This field is optional and when omitted, the publisher MUST deliver starting from the beginning of the selected group.
+
+* Track Authorization Info: 
+Carries track authorization, typically in the form of a token, authorizing the client’s subscription to the track. The specifics of obtaining the authorization information is out of scope for this specification.
```
afrind, line 701: Maybe remove "typically in the form of a token", since "the specifics of ... authorization information is out of scope".

draft-lcurley-warp.md
afrind: I recommend you add "expires" in SUBSCRIBE_REPLY as @kixelated suggests, and rewrite this bit to say "the subscription is active until it expires, or the underlying transport is disconnected".  Once UNSUBSCRIBE is added, then you can add the bit here about "client policy".

draft-lcurley-warp.md
afrind: What do you want to varint here?

draft-lcurley-warp.md
afrind: The QUIC wg likes to use different message types for this kind of thing -- along the lines of Luke's second suggestion, but we can also move forward now with just some extra text explaining when the ID will be there and when it will not.

draft-lcurley-warp.md
```diff
@@ -650,30 +652,110 @@ This contains base information required to decode OBJECT messages, such as codec
...
+Identifies the track in the request message for which this
+response is provided.
+
+* Response:
+Provides result of the transaction, where a value of `Ok` indicates successful subscription. For failed response, the "Reason Phrase" 
+is populated with appropriate reason.
+
+
+* Track ID:
+Identifies the hop-by-hop identifier mapping the given Track Name to be populated in the Object messages ({{message-object}}) and MUST only be present if the `Response` is `Ok`. This field is populated with either the `Track ID` value provided in the request or the one chosen by client processing the request. The Track ID field in the Object's header messages MUST be populated with the value in this field.
```
afrind, line 758: > the `Track ID` value provided in the request or the one chosen by client processing the request

Do you mean "chosen by the *server* processing the request" ?

afrind (commented):
Individual Review

draft-lcurley-warp.md
```diff
@@ -650,30 +652,110 @@ This contains base information required to decode OBJECT messages, such as codec
...
+The `TRACK INFO` structure is as defined below
+
+~~~
+TRACK INFO {
+  Track Name Length(i),
+  Track Name(...),
+  Track ID (i),
+  Track Authorization Info Length(i),
+  Track Authorization Info (...),
+  [Group Sequence (i)],
```
kixelated, line 680: Add some text indicating how these are optional. In the current PR, it's only possible to tell if a field is present based on the message length. This is not how QUIC, HTTP/3, QPACK, TLS, etc encodes optional fields so it needs to be very explicit.

Or make them required and reserve 0. I think we should start with that.

draft-lcurley-warp.md
```diff
@@ -650,30 +652,110 @@ This contains base information required to decode OBJECT messages, such as codec
...
+
+* Track ID: 
+A proposal serving as short hop-by-hop identifier to be used in the OBJECT ({{message-object}}) message header for the requested tack. Given that the media corresponding to a track can potentially be delivered over multiple data streams, the Track ID provides the necessary mapping between the "Track Name" in the control message and the corresponding data streams. It also serves as a compression identifier for containing the size of object headers instead of carrying complete "Track Name" information in every object message. A client MUST choose an Track ID value that is unique within a QUIC Connection.
+
+* Group Sequence: 
+Identifies the group within the track to start delivering the objects. This field is optional and when omitted, the publisher MUST deliver the objects from the most recent group.
+
+* Object Sequence: 
+Identifies the object within the track to start the media delivery. The `Group Sequence` field MUST be set to identify the group under which to
+the deliver the objects. This field is optional and when omitted, the publisher MUST deliver starting from the beginning of the selected group.
```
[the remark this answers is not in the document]
kixelated: But can we remove this for now? I don't want to implement it, and I think the usefulness is nebulous at best.

draft-lcurley-warp.md
```diff
@@ -650,30 +652,110 @@ This contains base information required to decode OBJECT messages, such as codec
...
+~~~
+enum Track Result
+{
+  Ok(0),
+  Fail(1)
+}
+
+TRACK RESPONSE {
+  Track Name Length(i),
+  Track Name(...)...,
```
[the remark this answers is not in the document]
suhasHere: It correlates the track in the request, since there can be multiple in flight.
we can add further text for handling edge cases or error flows in the follow up PRs.

draft-lcurley-warp.md
```diff
@@ -715,29 +717,89 @@ This contains base information required to decode OBJECT messages, such as codec
...
 
-The format of SUBSCRIBE is as follows:
+The format of SUBSCRIBE REQUEST is as follows:
 
 ~~~
-SUBSCRIBE Message {
-  Track Count (i),
-  Track IDs (..),
+Track Parameter {
+  Track Parameter ID (i),
```
afrind, line 728: I see these IDs are defined below -- can you add a reference to that section.

draft-lcurley-warp.md
```diff
@@ -784,6 +846,19 @@ The ROLE parameter (key 0x00) allows the client to specify what roles it expects
 
 The client MUST send a ROLE parameter with one of the three values specified above. The server MUST close the connection if the ROLE parameter is missing, is not one of the three above-specified values, or it is different from what the server expects based on the application in question.
 
+# Track Parameters {#track-params}
+
+The Track parameters are described as below. Every parameter MUST appear at most once. The peers SHOULD verify that and close the connection if a parameter appears more than once. The Parameter Value Length field indicates the length of the Parameter Value.
+
+### GROUP SEQUENCE Parameter
+
+The GROUP SEQUENCE parameter (key 0x00) identifies the group within the track to start delivering the objects. The publisher MUST start delivering the objects from the most recent group, when this parameter is omitted.
```
afrind, line 855: In the case where the publisher is a cache and the media is no longer live, does this mean start from the last group?

draft-lcurley-warp.md
```diff
@@ -784,6 +846,19 @@ The ROLE parameter (key 0x00) allows the client to specify what roles it expects
...
+
+### OBJECT SEQUENCE Parameter
+The OBJECT SEQUENCE parameter (key 0x01) identifies the object with the track to start delivering the media. The `GROUP SEQUENCE` parameter MUST be set to identify the group under which to start the media delivery. The publisher MUST start delivering from the beginning of the selected group when this parameter is omitted.
+
+### AUTHORIZATION INFO Parameter
+AUTHORIZATION INFO parameter (key 0x02) identifies the mandatory parameter carrying track's authorization, authorizing the client’s subscription to the track. The specifics of obtaining the authorization information is out of scope for this specification.
```
afrind, line 861: If it's mandatory, do you want a MUST in here?  Aren't there applications where authorization isn't required at the track level (either public broadcast, or perhaps authorized at webtransport level for that mapping) - so maybe it shouldn't be mandatory?

draft-lcurley-warp.md
```diff
@@ -715,29 +717,89 @@ This contains base information required to decode OBJECT messages, such as codec
...
-The number of track IDs that follow.
-This MAY be zero to unsubscribe to all tracks.
 
-* Track IDs:
-A list of varint track IDs.
+* Track URI:
+Identifies the track as defined in ({{track-uri}}).
+
+* Track ID: 
+Session specific identifier that maps the Track URI to the Track ID in OBJECT ({{message-object}}) message headers for the advertised track. Peer processing the request message MAY end up choosing a different Track ID (see {{message-subscribe-ok}}). Track IDs are generally shorter than Track URIs and thus reduce the overhead in OBJECT messages.
```
afrind, line 747: I recommend removing Track ID from SUBSCRIBE REQUEST and leave it only in SUBSCRIBE OK.

afrind (commented):
Individual review

draft-lcurley-warp.md
afrind: Sounds good to me

draft-lcurley-warp.md
```diff
@@ -715,29 +717,89 @@ This contains base information required to decode OBJECT messages, such as codec
...
+
+A `SUBSCRIBE OK` control message is sent for successful subscriptions. 
+
+~~~
+SUBSCRIBE OK
+{
+  Track URI Length(i),
+  Track URI(...),
+  Track ID(i),
+  Expires (i)
```
vasilvv, line 764: Do we always need `Expires`?  Do we want to have key/value pairs in the OK message, making `Expires` one of those?

draft-lcurley-warp.md
```diff
@@ -715,29 +717,89 @@ This contains base information required to decode OBJECT messages, such as codec
...
+
+Subscriptions stay active until it expires or the underlying transport is disconnected.
+
+
+## SUBSCRIBE ERROR {#message-subscribe-error}
+
+A `SUBSCRIBE ERROR` control message is sent for unsuccessful subscriptions. 
+
+~~~
+SUBSCRIBE ERROR
```
vasilvv, line 787: Do we want to also add a machine-readable error code?

draft-lcurley-warp.md
```diff
@@ -715,29 +717,89 @@ This contains base information required to decode OBJECT messages, such as codec
...
+Entities that intend to receive media will do so via subscriptions to one or more tracks. The information for the tracks can be obtained via catalog or from incoming SUBSCRIBE REQUEST messages.
 
-The format of SUBSCRIBE is as follows:
+The format of SUBSCRIBE REQUEST is as follows:
 
 ~~~
-SUBSCRIBE Message {
-  Track Count (i),
-  Track IDs (..),
+Track Parameter {
```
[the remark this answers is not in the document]
suhasHere: @vasilvv since @afrind suggested we might want to reuse these parameters for Publish Request as well, how about I call it as 'Track Request Parameters'

suhasHere:
@vasilvv @afrind the last few commits should address your feedback. please let me know . Thanks

draft-lcurley-warp.md
```diff
@@ -833,6 +907,7 @@ The producer and consumer MUST cancel a stream, preferably the lowest priority,
 TODO: fill out currently missing registries:
 * Warp version numbers
 * SETUP parameters
+* Track Request parameters
```
afrind, line 910: +Subscribe Error codes

draft-lcurley-warp.md
afrind: I can imagine this tripping some people, who would expect that the default is the beginning and not the end.  I suppose it makes sense given the dominant use cases are live.

afrind (commented):
Thank you for the updates.

fluffy (approved):
LGTM

vasilvv (approved):
My issues:

1. We should have parameters in SUBSCRIBE_OK, since currently it has no extensibility.
2. Expiry needs more thought.

I'll file issues for those, but for now, LGTM.

PR #123, opened 2023-04-05
IETF116:  Update Publish messages

suhasHere:
This PR depends on #121 and #122  and captures discussions at the IETF 116 to carryout publishes on tracks

Specifically this PR adds support for the following
1. Publish Request message (similar to Subscribe Request) for track authorization
2. Publish Reply message (similar to Subscribe Reply) for providing the result of publish request.

The core data structures are shared between subscribe and the publish messages.

As with PR #122, the scope of a publish request message is one track and can be extended to support multiple tracks in a separate PR, if needed.

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -552,6 +552,10 @@ A length of 0 indicates the message is unbounded and continues until the end of 
 |------|----------------------------------------------|
 | 0x4  | SUBSCRIBE REPLY ({{message-subscribe-reply}})|
 |------|----------------------------------------------|
+| 0x5  | PUBLISH REQUEST ({{message-publish-req}})    |
+|------|----------------------------------------------|
+| 0x6  | PUBLISH REPLY ({{message-publish-reply}})    |
+|------|----------------------------------------------|
 | 0x10 | GOAWAY ({{message-goaway}})                  |
 |------|----------------------------------------------|
 
@@ -776,7 +780,38 @@ Provides result of the transaction, where a value of `Ok` indicates successful s
 * Track ID:
 Identifies the hop-by-hop identifier mapping the given Track Name to be populated in the Object messages ({{message-object}}). This field is optional and is provided in cases where the end point is unable to use the `Track ID` proposed by the peer in the request message. If populated, the `Track ID` field in the Object's header messages MUST be populated with the value in this field otherwise the `Track ID` value MUST correspond to one in the request message for a given track.
 
+## PUBLISH REQUEST {#message-publish-req}
 
+The `PUBLISH` control message sets up authorization for tracks that the publisher intends to publish media with.
+
+The format of PUBLISH is defined as below
+
+~~~
+PUBLISH REQUEST Message {
+  TRACK INFO Track
+}
+~~~
+{: #warp-publish-format title="Warp PUBLISH Message"}
+
+* Track:
+Identifies track information in `TRACK INFO`. The same rules as defined in {{message-subscribe-req}} will apply here as well.
+
+## PUBLISH REPLY {#message-publish-reply}
+
+The `PUBLISH REPLY` control message provides result of request to publish via the `PUBLISH REQUEST` message.
+
+The format of PUBLISH REPLY is defined as below
+
+~~~
+PUBLISH REPLY Message {
+ TRACK RESPONSE Track
+}
+~~~
+{: #warp-publish-reply-format title="Warp PUBLISH REPLY Message"}
+
+
+* Track:
+Captures the result of publish as by `TRACK RESPONSE` and the same rules apply as defined in {{message-subscribe-reply}}.
 
 ## GOAWAY {#message-goaway}
 The `GOAWAY` message is sent by the server to force the client to reconnect.
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -539,19 +539,21 @@ Warp Message {
 The Message Length field contains the length of the Message Payload field in bytes.
 A length of 0 indicates the message is unbounded and continues until the end of the stream.
 
-|------|-----------------------------------|
-| ID   | Messages                          |
-|-----:|:----------------------------------|
-| 0x0  | OBJECT ({{message-object}})       |
-|------|-----------------------------------|
-| 0x1  | SETUP ({{message-setup}})         |
-|------|-----------------------------------|
-| 0x2  | CATALOG ({{message-catalog}})     |
-|------|-----------------------------------|
-| 0x3  | SUBSCRIBE ({{message-subscribe}}) |
-|------|-----------------------------------|
-| 0x10 | GOAWAY ({{message-goaway}})       |
-|------|-----------------------------------|
+|------|----------------------------------------------|
+| ID   | Messages                                     |
+|-----:|:---------------------------------------------|
+| 0x0  | OBJECT ({{message-object}})                  |
+|------|----------------------------------------------|
+| 0x1  | SETUP ({{message-setup}})                    |
+|------|----------------------------------------------|
+| 0x2  | CATALOG ({{message-catalog}})                |
+|------|----------------------------------------------|
+| 0x3  | SUBSCRIBE REQUEST ({{message-subscribe-req}})|
+|------|----------------------------------------------|
+| 0x4  | SUBSCRIBE REPLY ({{message-subscribe-reply}})|
+|------|----------------------------------------------|
+| 0x10 | GOAWAY ({{message-goaway}})                  |
+|------|----------------------------------------------|
 
 ## SETUP {#message-setup}
 
@@ -664,29 +666,116 @@ This contains base information required to decode OBJECT messages, such as codec
 An endpoint MUST NOT send multiple CATALOG messages.
 A future draft will add the ability to add/remove/update tracks.
 
-## SUBSCRIBE {#message-subscribe}
-After receiving a CATALOG message ({{message-catalog}}, the receiver sends a SUBSCRIBE message to indicate that it wishes to receive the indicated tracks.
+## SUBSCRIBE REQUEST {#message-subscribe-req}
+
+Entities that intend to receive media will do so via subscriptions to
+one or more tracks. The information for the tracks can be obtained via catalog or from incoming SUBSCRIBE messages.
 
 The format of SUBSCRIBE is as follows:
 
 ~~~
 SUBSCRIBE Message {
-  Track Count (i),
-  Track IDs (..),
+  TRACK INFO Track
 }
 ~~~
 {: #warp-subscribe-format title="Warp SUBSCRIBE Message"}
 
-* Track Count:
-The number of track IDs that follow.
-This MAY be zero to unsubscribe to all tracks.
+* Track:
+Identifies the track information as defined in `TRACK INFO`.
 
-* Track IDs:
-A list of varint track IDs.
+The `TRACK INFO` structure is as defined below
+
+~~~
+TRACK INFO {
+  Track Name Length(i),
+  Track Name(...)...,
+  Track ID (i),
+  Track Authorization Info Length(i),
+  Track Authorization Info (...)...
+  [Group Sequence (i)],
+  [Object Sequence (i)],
+}
+~~~
+{: #warp-track-info-format title="Warp TRACK INFO Content"}
 
 
-Only the most recent SUBSCRIBE message is active.
-SUBSCRIBE messages MUST be sent on the same QUIC stream to preserve ordering.
+* Track Name:
+Identifies the fully qualified track name as defined in ({{model-track}}).
+
+* Track ID:
+A proposal serving as short hop-by-hop identifier to be used in the OBJECT ({{message-object}}) message header for the requested tack. Given that the media corresponding to a track can potentially be delivered over multiple data streams, the `Track ID` provides the necessary mapping between the "Track Name" in the control message and the corresponding data streams. It also serves as a compression identifier for containing the size of object headers instead of carrying complete "Track Name" information in every object message. A client MUST choose an `Track ID` value that is unique within a QUIC Connection.
+
+* Group Sequence:
+Identifies the group within the track to start delivering 
+the objects. This field is optional and when omitted, the
+publisher MUST deliver the objects from the most recent 
+group.
+
+* Object ID:
+Identifies the object within the track to start the media 
+delivery. The group MUST either match the `Group Sequence` if provided
+or MUST be the most recent group. This field is optional and 
+when omitted, the publisher MUST deliver starting from the 
+beginning of the selected group, i.e., starting at Object Sequence of 0.
+
+* Track Authorization Info: 
+Carries track authorization, typically in the form of a token, authorizing client’s subscription to the track. The specifics of obtaining the authorization information is out of scope for this specification.
+
+Subscriptions stay active until one of the following happens:
+
+- a client local policy dictates expiration of a subscription.
+- optionally, a server policy dicates subscription expiration.
+- the underlying transport is disconnected.
+
+A client MUST renew its subscriptions by sending a new `SUBSCRIBE` to keep them active. Such subscriptions MUST refresh the existing subscriptions and thus only the most recent SUBSCRIBE message is considered active. A renewal period of 5 seconds is RECOMMENDED.
+
+## SUBSCRIBE REPLY {#message-subscribe-reply}
+
+A `SUBSCRIBE REPLY` control message provides results of the subscription. 
+
+~~~
+SUBSCRIBE REPLY
+{
+  TRACK RESPONSE Track
+}
+~~~
+{: #warp-subscribe-reply format title="Warp SUBSCRIBE REPLY Message"}
+
+* Track:
+Captures the result of subscription as defined in `TRACK RESPONSE`
+
+The `TRACK RESPONSE` structure is as defined below
+
+~~~
+enum Track Result
+{
+  Ok(0),
+  Expired(1),
+  Fail(2)
+}
+
+TRACK RESPONSE {
+  Track Name Length(i),
+  Track Name(...)...,
+  Track Result Response,
+  [ Reason Phrase Length (i) ],
+  [ Reason Phrase (...) ],
+  [ Track ID(i) ]
+}
+~~~
+{: #warp-track-response-format title="Warp TRACK RESPONSE Content"}
+
+
+* Track Name:
+Identifies the track in the request message for which this
+response is provided.
+
+* Response:
+Provides result of the transaction, where a value of `Ok` indicates successful subscription. For failed or expired responses, the "Reason Phrase" shall be populated with appropriate reason code.
+
+* Track ID:
+Identifies the hop-by-hop identifier mapping the given Track Name to be populated in the Object messages ({{message-object}}). This field is optional and is provided in cases where the end point is unable to use the `Track ID` proposed by the peer in the request message. If populated, the `Track ID` field in the Object's header messages MUST be populated with the value in this field otherwise the `Track ID` value MUST correspond to one in the request message for a given track.
+
 
 
 ## GOAWAY {#message-goaway}
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -46,6 +46,7 @@ normative:
   QUIC-RECOVERY: RFC9002
   WebTransport: I-D.ietf-webtrans-http3
   URI: RFC3986
+  MOQ-ARCH: I-D.nandakumar-moq-arch-00
 
   ISOBMFF:
     title: "Information technology — Coding of audio-visual objects — Part 12: ISO Base Media File Format"
@@ -164,6 +165,13 @@ Variant:
 
 : A track with the same content but different encoding as another track. For example, a different bitrate, codec, language, etc.
 
+Provider: 
+
+: Entity capable of hosting media application
+session based on the MOQ Transport protocol.  It is responsible for
+authorizing the publishers/subscribers, managing the names used
+for Tracks and is scoped under a domain for a specific application.
+In certain deployments, a provider is also responsible for establishing trust between clients and relays for delivering media.
 
 ## Notational Conventions
 
@@ -177,11 +185,11 @@ x (b):
 
 # Model
 
-## Objects
+## Objects {#model-object}
 
 The basic element of Warp is an *object*. An object is a single addressable
 cacheable unit whose payload is a sequence of bytes.  An object MAY depend on other 
-objects to be decoded. An object MUST belong to a group {{groups}}. Objects carry 
+objects to be decoded. An object MUST belong to a group {{model-group}}. Objects carry 
 associated metadata such as priority, TTL or other information usable by a relay, 
 but relays MUST treat object payloads as opaque.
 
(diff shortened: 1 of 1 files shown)


fluffy (approved):

draft-lcurley-warp.md
```diff
@@ -776,8 +780,39 @@ Provides result of the transaction, where a value of `Ok` indicates successful s
 * Track ID:
 Identifies the hop-by-hop identifier mapping the given Track Name to be populated in the Object messages ({{message-object}}). This field is optional and is provided in cases where the end point is unable to use the `Track ID` proposed by the peer in the request message. If populated, the `Track ID` field in the Object's header messages MUST be populated with the value in this field otherwise the `Track ID` value MUST correspond to one in the request message for a given track.
 
+## PUBLISH REQUEST {#message-publish-req}
+
+The `PUBLISH` control message sets up authorization for tracks that the publisher intends to publish media with.
```
afrind, line 785: PUBLISH_REQUEST

suhasHere: Thanks for cathcing it . Will fix it

draft-lcurley-warp.md
```diff
@@ -776,8 +780,39 @@ Provides result of the transaction, where a value of `Ok` indicates successful s
...
+
+The format of PUBLISH is defined as below
+
+~~~
+PUBLISH REQUEST Message {
+  TRACK INFO Track
```
kixelated, line 791: This needs to contain a track ID.

kixelated: Actually, can you be explicit and list the fields here? I'd rather be explicit even if it means more duplication.

kixelated (changes requested):
Going to -1 this until we talk more about the publish use-case. This feels like PUSH_PROMISE.

The ability to publish is intended to remove a round-trip when the sender already knows the intent of the receiver. There has to be some form of coordination out of the scope of this specification, which automatically rules out generic clients.

If this coordination of intent already exists, then receiver could just SUBSCRIBE immediately. I don't actually think it introduces an RTT because of SETUP currently works, and now is not the time to be nit-picking over startup RTTs anyway.

The cost of PUBLISH is complexity and race conditions. OBJECTs arrive before they were requested, requiring some form of buffering and identification. We also need a way of saying "pls no" if the endpoints actually did not coordinate.

kixelated:
I talked to Alan on slack, and maybe there's a misunderstanding on my part. I think we've been simultaneously talking about two different PUBLISH messages, which I'm going to label 0-RTT and 1-RTT. The difference between the two is when OBJECT messages are sent.

**0-RTT PUBLISH:**
```
-> PUBLISH       track_id=5 track_name=/video
-> OBJECT        track_id=5 object_id=0
<- PUBLISH_REPLY track_id=5 status=ok
```

**1-RTT PUBLISH:**
```
-> PUBLISH       track_id=5 track_name=/video
<- PUBLISH_REPLY track_id=5 status=ok
-> OBJECT        track_id=5 object_id=0
```


The 0-RTT publish is problematic because there's a race. If the PUBLISH message is lost, then the receiver will receive (and have to buffer) objects for unknown track IDs. It's an optimization when both endpoints have pre-negotiated subscriptions, which is why I liken it to HTTP push. I would personally not implement it and likewise I suspect CDNs would not either.

The 1-RTT publish has a lot of overlap with the current CATALOG message since it announces available track URIs. However, the difference is that the track is described solely based on it's name instead of a detailed media descriptor. Additionally, sending a PUBLISH per track is racey, which is why I combined all tracks into a single message.

kixelated:
If the goal is 0-RTT PUBLISH, then it's a little early to optimize for RTTs. If both endpoints can pre-negotiate track names to push, then they can also just pre-negotiate track IDs to push and skip the PUBLISH altogether.

If the goal is 1-RTT PUBLISH, then I want to see a media catalog that describes the available tracks rather than an opaque URI. There should be some mechanism for the receiver to know the contents of the track before it agrees to it with a PUBLISH_REPLY.

suhasHere:
> There should be some mechanism for the receiver to know the properties of the track before it agrees to subscribe to it (via a PUBLISH REPLY).

Relays will not have access to catalog.  So a publish request going from client endpoint to a relay serves as authz enforcement for the track being advertised by the cliend endpoint as the publisher.

kixelated:
> > There should be some mechanism for the receiver to know the properties of the track before it agrees to subscribe to it (via a PUBLISH REPLY).
> 
> Relays will not have access to catalog. So a publish request going from client endpoint to a relay serves as authz enforcement for the track being advertised by the cliend endpoint as the publisher.



> > There should be some mechanism for the receiver to know the properties of the track before it agrees to subscribe to it (via a PUBLISH REPLY).
> 
> Relays will not have access to catalog. So a publish request going from client endpoint to a relay serves as authz enforcement for the track being advertised by the cliend endpoint as the publisher.

Okay, that makes sense. Is it fair to say that the purpose of the PUBLISH message just for authentication?

draft-lcurley-warp.md
```diff
@@ -801,6 +807,75 @@ response is provided.
...
+
+The format of PUBLISH is defined as below:
+
+~~~
+PUBLISH REQUEST Message {
+  Track URI Length(i),
+  Track URI(...),
+  Track ID (i),
+  Track Authorization Info Length(i),
+  Track Authorization Info (...),
```
afrind, line 822: Why are the track parameters generic for subscribe but specific for publish?  Can we make them the same?

draft-lcurley-warp.md
```diff
@@ -801,6 +807,75 @@ response is provided.
...
+}
+~~~
+{: #warp-publish-format title="Warp PUBLISH REQUEST Message"}
+
+* Track URI:
+Identifies the fully qualified track name as defined in ({{track-uri}}).
+
+* Track ID: 
+Session specific identifier that maps the Track URI to the Track ID in OBJECT ({{message-object}}) message headers for the advertised track. Peer processing the request message MAY end up choosing a different Track ID (see {{message-publish-ok}}). Track IDs are generally shorter than Track URIs and thus reduce the overhead in OBJECT messages.
```
afrind, line 831: Same comment as subscribe

afrind (commented):
Individual review

draft-lcurley-warp.md
suhasHere: @afrind I wasn't too sure if anything other than authz parameter made sense for publish. may be they do, I am not thinking clearly ?

afrind: @suhasHere OBJECT and GROUP may not make sense for PUBLISH_REQUEST, but these params provide a nice extension point for additional params that may be needed in the future.  The tradeoff is that params that only make sense for SUBSCRIBE or PUBLISH will need to state that explicitly.

Also, are there cases where publish need not be authorized at the track level, in which case this might be omitted?

afrind: @suhasHere : agree there are many cases where it's required -- only pointing out that because there are some that are not, we might not want to make it a mandatory field.  Any application that requires auth and detects the auth info is missing can return an error.

suhasHere:
@afrind  latest commits should address the feedback provided. Please let me know .. thanks

draft-lcurley-warp.md
```diff
@@ -800,6 +806,74 @@ Identifies an integer error code for subscription failure.
...
+  Track Request Parameters (..) ...
+}
+~~~
+{: #warp-publish-format title="Warp PUBLISH REQUEST Message"}
+
+* Full Track Name:
+Identifies the fully qualified track name as defined in ({{track-fn}}).
+
+* Track Request Parameters: 
+AUTHORIZATION INFO (see {{track-req-params}}) is be the only parameter applicable for the publish requests.
```
afrind, line 828: editorial: If OBJECT_SEQUENCE and GROUP_SEQUENCE are not valid in publish, I recommend making that clear in the param definitions in the Track Request Parameters {#track-req-params} section, then you can just reference that section here.  Alternately, subscribe can identify the parameters that are valid above.

afrind (commented):
Thanks for the updates.

draft-lcurley-warp.md
suhasHere: OBJECT_SEQUENCE and GROUP_SEQUENCE is  still valid. At this point in time, there is a clear use for Authz Info and thus the text reflected it. It may be good to deal on how group/object sequences can be used for publish in its own PR and thus keeping this PR focussed for the time being ?


I do see as we add more parameters, we will have some that are specific to one request vs the others.

suhasHere:
@afrind @kixelated @vasilvv @fluffy does the note added is good to proceed with this PR  ?

suhasHere:
@afrind @kixelated @vasilvv @fluffy does the note added is good to proceed with this PR ?

afrind:
Did you intend to remove the bit about "active discussion and will be revised" in d2c791b?

afrind:
I opened #143 to track the discussion

suhasHere:
@afrind that was not the intent . updated now. Thanks

afrind:
Thanks for the update.

fluffy (approved):
I think this looks like a fine starting point to #143 and we should merge it

fluffy:
So the the 0-RTT vs 1-RTT is not so much an issue with this PR but is a separate issue of if the relay gets to override the ID or not. I think we should deal with that as a separate issue.

wilaw:
Apologies for missing last weeks editors call where this was debated.  My concerns with this approach are several: 

1. The ANNOUNCE message, with its enumeration of track names, is an exact duplicate of data already contained in the proposed CATALOG. See https://wilaw.github.io/MoQ/wilaw-patch-1/draft-law-moq-warpmedia.html#name-catalog-objects. Why do we need two mechanisms for conveying the same data? We can easily add an AUTH field to the  track change descriptors if that is what is required to avoid duplication?
2. An ingest server that is authorizing what a connected publisher may publish is not acting as a relay. It is acting as an end-point and would therefore be able to parse the catalog. 
3. If we follow a publish-after-subscribe workflow (which I strongly believe we should), then an edge relay sitting on the contribution side will not have content coming to it that it didn't previously issue a subscribe request for. It will only issue subscriptions that come to it from upstream. Those subscriptions must have originated at an endpoint - either an origin or a end-client. That endpoint will have an access token to authorize the subscription. So the conferencing system can control the output of the publishing entities by controlling the access tokens that are used to pull content from them. 
4. For auth, we get to a situation where either the connection (WT or QUIC) specifies what can be subscribed to , or each object carries a token and the relay evaluates that token before routing. We would use an alias scheme for these tokens, like we do for object name, to reduce bits-on-the-wire and to reduce CPU effort in validating tokens. 

I'd love to whiteboard out the flow around subscription. Reconstructing a virtual workflow from PR comments and issue threads is difficult and leads to confusion.

suhasHere:
Auth is for the publisher and Announce will be needed for on path relays to trust the publisher and also to be able to find the publisher with in the relay network when a subscribe happens


I am still fine to call it as publish_request/response, if it helps as the semantics are more clearer

We need this functionality regardless

afrind (commented):
Individual/Editorial Review

fluffy:
I don't see this as the same as the catalog. The catalog has a bunch of information all the tracks. All this has is the authorization token that allows you to publish to a given track. I think I have outlined in other issues why we need this but I hope we can discuss in person today. We also have the issue of the relays can not read the catolog.

draft-lcurley-warp.md
```diff
@@ -908,6 +1009,7 @@ TODO: fill out currently missing registries:
 * Warp version numbers
 * SETUP parameters
 * Track Request parameters
+* Subscribe Error codes
```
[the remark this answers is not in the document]
suhasHere: good point Alan . I will update it

afrind:
Individual Clarification

> If we follow a publish-after-subscribe workflow

Announce isn't intended to change the publish-after-subscribe workflow.  It's a mechanism for authorizing individual tracks, dynamically building routing tables in relays that can't parse the catalog, and offers relays a chance to pre-subscribe to content if desired.

As I read it now, it's also optional: a receiver can issue a subscribe for a track without having seen the announce for it (publishers gotta handle subscribes for non-existent tracks anyways), and the receiver can reply ANNOUNCE_OK to any ANNOUNCE it sees without taking any further action.

fluffy (approved):

wilaw:
I would support merging this PR along with a DISCUSS statement similar to the following:

We need discuss and have consensus around how bi-directional routing is going to work with dynamic publishers. CDN's of today have a mechanism for handling ingest from multiple publishers. Can that mechanism be extended to MoQ? Or is a new mechanism, along the lines of ANNOUNCE, required to meet the pub/sub routing demands of applications such as real-time conferencing?

draft-lcurley-warp.md
```diff
@@ -800,6 +806,101 @@ Identifies an integer error code for subscription failure.
...
+}
+~~~
+{: #warp-track-info title="Warp Track Info"}
+
+* Full Track Name:
+Identifies the track in the request message for which this
+response is provided.
+
+* Track Request Parameters: 
+AUTHORIZATION INFO (see {{track-req-params}}) is the only parameter applicable for the announce messages.
```
kixelated, line 840: Do we need to authorize each track individually? I can't imagine a world where you would send a different authorization token for catalog/audio/video (same goes for subscribe). Duplicating that token for each track just seems wasteful.

draft-lcurley-warp.md
```diff
@@ -800,6 +806,101 @@ Identifies an integer error code for subscription failure.
 * Reason Phrase:
 Provides the reason for subscription error and `Reason Phrase Length` field carries its length.
 
+## ANNOUNCE {#message-announce}
+
+The publisher advertises the tracks via the `ANNOUNCE` control message. The `ANNOUNCE` message sets up authorization for tracks that the publisher intends to publish media with. The receiver can then SUBSCRIBE to the advertised tracks.
```
kixelated, line 811: > The `ANNOUNCE` message sets up authorization for tracks that the publisher intends to publish media with.

Authentication is a secondary purpose, especially since it's optional. ANNOUNCE is primarily for discovery, and possibly routing.

draft-lcurley-warp.md
```diff
@@ -800,6 +806,101 @@ Identifies an integer error code for subscription failure.
...
+{: #warp-track-info title="Warp Track Info"}
+
+* Full Track Name:
+Identifies the track in the request message for which this
+response is provided.
+
+* Track Request Parameters: 
+AUTHORIZATION INFO (see {{track-req-params}}) is the only parameter applicable for the announce messages.
+
+Additionally the `ANNOUNCE` message enables flows where publisher advertises the tracks to its peer (say Relays) and the peer lacks sufficient information on available tracks to issue subscriptions. This is typical of clients in a realtime conferencing publishing media through the relays or a broadcaster streaming to a server or publishing across distribution network boundaries.
```
kixelated, line 842: I don't think this is conferencing or even contribution specific. It's just a way of announcing what tracks are available, so a subscriber (viewer, relay, or ingest edge) could follow up with a subscription.

draft-lcurley-warp.md
```diff
@@ -800,6 +806,101 @@ Identifies an integer error code for subscription failure.
...
+* Track Name:
+List of tracks identified by their Full Track Name selected from the
+`Track Info` list in the `ANNOUNCE` message.
+
+## ANNOUNCE ERROR {#message-announce-error}
+
+A `ANNOUNCE ERROR` control message is sent for the tracks that failed their authorization at the peer.
+
+~~~
+ANNOUNCE ERROR
```
kixelated, line 870: Bleh, I don't like listing which tracks succeeded and which ones failed. Can we treat them as an all-or-nothing? Would a publisher even be able to recover if one or more tracks were rejected?

draft-lcurley-warp.md
suhasHere: One of the outcomes at IETF 116 was to define MoQ as delivering tracks, thus authz is tied around tracks.

I think we should make tge Authorization Info as mandatory parameter. However if a given AuthZ info applies to one track or array of tracks can be restructured in our message design 

would something like this work 
ANNOUNCE {
  <Array of track full names>,
  AuthzInfo
}

```

Comment (suhasHere) [reply draft-lcurley-warp.md:840]: Understood, i was suggesting something like this 
```
ANNOUNCE {
<track-names array>,
AuthzInfo
}
```

File: draft-lcurley-warp.md (hunk 8 above)
Comment (suhasHere) [reply draft-lcurley-warp.md:811]: Authorization is also primary purpose to ensure only authorized publishers can send media to the relay. I will update the text as

"The ANNOUNCE message provides for the relays to discover tracks being published, enable appropriate routing for serving subscriptions and set up authorization for tracks that the publisher intends to publish media with"

does this work ?

File: draft-lcurley-warp.md (hunk 10 above)
Comment (vasilvv) [reply draft-lcurley-warp.md:868]: If we allow individual tracks to succeed or fail, is there any particular reason to put multiple tracks into a single ANNOUNCE message?

Comment (suhasHere) [reply draft-lcurley-warp.md:868]: I understand the concern. How about something like this 

Announce Request {
  <array of track names>,
   Auth Info
}

Announce Reply {
  Response 
}

File: draft-lcurley-warp.md [hunk 11]
```diff
@@ -800,6 +806,99 @@ Identifies an integer error code for subscription failure.
...
 
+## ANNOUNCE {#message-announce}
+
+The publisher advertises the tracks via the `ANNOUNCE` control message. The `ANNOUNCE` message provides for the relays to discover tracks being published, enable appropriate routing for serving subscriptions and set up authorization for tracks that the publisher intends to publish media with. The receiver can then SUBSCRIBE to the advertised tracks.
+
+~~~
+ANNOUNCE Message {
+  Track Info Count (i),
+  Track Info (..) ...
+}
```
afrind, line 817: I'm wondering if we want a params struct here also to convey meta-data about the group-of-tracks being annouced.

draft-lcurley-warp.md
afrind: I prefer Announce / Announce Reply, where Reply has the success/failure of each track in the Announce.

draft-lcurley-warp.md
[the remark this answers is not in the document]
afrind: Add Announce error codes to this list somewhere?

afrind (commented):
Individual review

draft-lcurley-warp.md
suhasHere: @wilaw  that would work and would require defining further about track naming the draft.

IF we have full track name defined as 
Full Track Name = Track Namespace + "/" + Emitter Id + "/" + Track Name
, then we can define announce as

Annouce {
   Track Reference,
   Params
}
where Track Reference is
Track Reference = Track Namespace + "/" + Emitter Id


Does this work ?

suhasHere: Created #154 to discuss this further

wilaw: > Does this work ?

LGTM. We should DISCUSS the details and consequences, but I think it provides an efficient mechanism for publisher registration and discovery in highly distributed networks and is worth exploring further under #154.

fluffy (approved):
I put thumbs up on several comments I agree with. This definitely looks close enough to merge to me and we can refine some of the details after that.

draft-lcurley-warp.md
vasilvv: Thinking more about this, I think a sender should be able to announce a bunch of tracks atomically (meaning all get accepted or rejected).

Two reasons for this:

1. If a client announces three tracks, let's say a catalog, an audio, and a video, and the server accepts the catalog, but denies an audio or a video track, this leaves the system in a state where there is a catalog for a bunch of unavailable tracks.
2. If the client announces a catalog before video or audio, there is a race condition in which someone could subscribe to the catalog and subscribe to the video before the server processes the ANNOUNCE for the video.

You can get around number 2 (and partially number 1, if you're willing to wait an extra RTT) by announcing the video track before the catalog, but this feels like a thing that's really easy to get wrong.

Both announcing multiple tracks, and announcing a prefix, should be fine, but we should probably pick one.

suhasHere: @vasilvv agree with the thinking there.  I want to be able to support use-cases where either
  - applications where catalog is not needed, say, security cameras broadcasting a static hardcoded set of track qualities
  - applications where catalog is exchanged out of band

Catalog as a track is a good option to have in the transport, but I don't think we can mandate there will be catalog tracks for all classes of applications.

Having said that, I am also open to allowing announce with multiple tracks  as a scoped set which is either completely accepted or rejected. I would prefer prefix approach better though when there are large cardinality of tracks.

I think https://github.com/kixelated/warp-draft/pull/123#discussion_r1190264132 address both classes of applications.

Also #154 supports the same idea for Subscriptions too.

draft-lcurley-warp.md
```diff
@@ -800,6 +806,107 @@ Identifies an integer error code for subscription failure.
...
+  Track Request Parameters (..) ...
+}
+~~~
+{: #warp-track-info title="Warp Track Info"}
+
+* Full Track Name:
+Identifies the track as defined in ({{track-fn}}).
+
+* Track Request Parameters: 
+AUTHORIZATION INFO (see {{track-req-params}}) is the only parameter applicable for the announce messages.
```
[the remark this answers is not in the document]
suhasHere: @vasilvv good point, will update in the next set of commits

draft-lcurley-warp.md
afrind: We may need to clarify:

Is either ANNOUNCE or transmitting a catalog in band mandatory in all media formats?  If a publisher and receiver have out-of-band mechanisms for exchanging track names, can a receiver do a "blind" SUBSCRIBE?

wilaw: IMO the catalog should be mandatory, for interop. Clients should always be able to depend upon it. We may want to use it to signal EOS and/or advertising etc. However, if a receiver has out of band info, they are welcome to do an "accelerated" subscription before they receive the catalog. The two are not exclusive and can co-exist.

suhasHere: Having a assigned catalog track needs to be specified but that it always needs to be populated cannot be done at this level. 

I agree WMF can mandate it further as well

suhasHere: Also agree that both can coexist. Since at the base transport we can't say one way or the other

draft-lcurley-warp.md
```diff
@@ -800,6 +813,71 @@ Identifies an integer error code for subscription failure.
...
+  Reason Phrase (...),  
+}
+~~~
+{: #warp-announce-error format title="Warp ANNOUNCE ERROR Message"}
+
+* Track Namespace:
+Identifies the track namespace in the ANNOUNCE message for which this response is provided.
+
+* Error Code:
+Identifies an integer error code for subscription failure.
```
vasilvv, line 876: s/subscription/announcement/

draft-lcurley-warp.md
```diff
@@ -800,6 +813,71 @@ Identifies an integer error code for subscription failure.
...
+
+* Reason Phrase:
+Provides the reason for subscription error and `Reason Phrase Length` field carries its length.
```
vasilvv, line 879: s/subscription/announcement/

vasilvv (approved):

draft-lcurley-warp.md
```diff
@@ -228,19 +228,26 @@ Tracks are identified by a globally unique identifier, called "Full Track Name"
...
 
 Example: 2
-Track Namespace = livestream.example
+Track Namespace = livestream.example.com
 Track Name = uaCafDkl123/audio
-Full Track Name = livestream.example/uaCafDkl123/audio
+Full Track Name = livestream.example.com/uaCafDkl123/audio
+
+Example: 3
+Track Namespace = security-camera.example.com/camera1/
```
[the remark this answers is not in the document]
suhasHere: that was my typo. I will fix it .

draft-lcurley-warp.md
```diff
@@ -826,6 +838,71 @@ Identifies an integer error code for subscription failure.
...
+
+* Track Namespace:
+Identifies a track's namespace as defined in ({{track-fn}})
+
+* Track Request Parameters: 
+As defined in {{track-req-params}}.
+
+The `ANNOUNCE` message allows for relays to:
+
+- Discover tracks being published by a publisher and enable setting up appropriate routing for serving subscriptions for tracks whose track full name shares common prefix with the announced namepace.
```
afrind, line 862: > serving subscriptions for tracks whose track full name shares common prefix with the announced namepace

I'm concerned about the `common prefix` bit, since longest prefix match algorithms might be problematic: can one publisher announce `example.org/steven' and another announce 'example.org/steve'?  Also there's a typo in namespace.

Can you change it to something like:

> serving subscriptions for tracks whose track full name matches the announced namespace

We can refine what it means to match in a later PR if needed, or possibly delegate to implementations?

afrind (commented):
Individual/editorial review

draft-lcurley-warp.md
suhasHere: It is indeed a prefix , since track namespace is only first part of the full track name.  For the typo scenario above, the authz will fail for the announce.

afrind: @suhasHere : Sorry I think I confused things.  

1) Your PR just misspelled the word "namespace" as "namepace". 

2) I meant to show that example.org/steve and example.org/steven are both valid and authorized namespaces in my application.

As I mentioned in the other PR, how about: 

> serving subscriptions for tracks whose Track Namespace matches the announced namespace

Does that work for you?

draft-lcurley-warp.md
```diff
@@ -826,6 +838,71 @@ Identifies an integer error code for subscription failure.
...
+~~~
+{: #warp-announce-format title="Warp ANNOUNCE Message"}
+
+* Track Namespace:
+Identifies a track's namespace as defined in ({{track-fn}})
+
+* Track Request Parameters: 
+As defined in {{track-req-params}}.
+
+The `ANNOUNCE` message allows for relays to:
```
kixelated, line 860: Not relay specific, and should probably go into a separate section anyway.

draft-lcurley-warp.md
```diff
@@ -883,17 +960,17 @@ if `query` is present, the client MUST concatenate `?`, followed by the `query`
 
 # Track Request Parameters {#track-req-params}
 
-The Track Request Parameters identify properties of the track requested in either the PUBLISH REQUEST or SUSBCRIBE REQUEST control messages. Every parameter MUST appear at most once. The peers MUST close the connection if there are duplicates. The Parameter Value Length field indicates the length of the Parameter Value.
+The Track Request Parameters identify properties of the track requested in either the ANNOUNCE or SUSBCRIBE REQUEST control messages. Every parameter MUST appear at most once. The peers MUST close the connection if there are duplicates. The Parameter Value Length field indicates the length of the Parameter Value.
 
 ### GROUP SEQUENCE Parameter
```
kixelated, line 965: Why add `This parameter is applicable in SUBSCRIBE REQUEST message.` if it applies to all track request parameters?

draft-lcurley-warp.md
```diff
@@ -883,17 +960,17 @@ if `query` is present, the client MUST concatenate `?`, followed by the `query`
 
 # Track Request Parameters {#track-req-params}
```
kixelated, line 961: Editorial: We should remove this section. It's confusing to lump ANNOUNCE and SUBSCRIBE parameters and then add a sentence at the end of each paragraph indicating the parameter only applies to one of them.

I would be verbose with how messages are defined even if it involves some copy-pasting. We can consolidate when the parameters are more stable.

kixelated (approved):
Let's merge this.

I do want to add my piece that ANNOUNCE is half-baked. It's not useful when track identification and authentication is pre-negotiated, and it's not comprehensive enough otherwise. We need to have more discussions as a working group.

PR #125, opened 2023-04-06
IETF116 - Relay behavior Section Template

suhasHere:
This PR introduces Relays section but leaves most of the subsections under it as TODo. The idea is keep the PR sizes small and focus on one change at a time.

If we can agree on the general shape of the relay section,  I will work on individual PRs for each of the TODOs

PR #126, opened 2023-04-06
IETF116: Fill in Relay Subscribe Interaction Sub-Section

suhasHere:
This PR depends on #122 and #125 

The scope of this PR is to add Relay Subscribe interactions. This is expected to be a starter text and we can add further clarification and details in a separate PR as we develop further.

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -497,7 +497,17 @@ Relays serves as policy enforcement points by validating subscribe
 and publish requests to the tracks.
 
 ## Subscriber Interactions
-TODO: This section shall cover relay handling of subscriptions.
+
+Subscribers interact with the Relays by sending a "SUBSCRIBE REQEUST"  ({{message-subscribe-req}}) control message for the tracks of interest. Relays MUST be willing to act on behalf of the subscriptions before they can forward the media, which implies that the subscriptions MUST be authorized and it is done as follows:
+
+1. Provider for the tracks MUST be authorized. This requires identifying the Provider of the requested track and verifying that the Relay is authorized to serve that Provider. "Track Prefix" component of the "Track Name" MUST identify the Provider. Specifics of Provider authorization depends on the way the relay is managed and is typically based on prior business agreement with the Provider.
+
+2. Verify that the subscriber is authorized to access the specified content. Subscriptions MUST carrying enough authorization information proving the subscriber has access to the requested track. 
+
+In all the scenarios, the end-point client making the subscribe
+request is notified of the result of the subscription, via "SUBSCRIBE REPLY" ({{message-subscribe-reply}}) control message.
+
+For cases where the subscriptions are successfully validated, Relay proceeed to save the subscription information by maintaining the mapping from the track information to the list of subscribers. This will enable Relays to forward on-going publishes (live or from cache) to the subscribers, if available, and also forward all the future publishes, until the subscriptions cases to exist. Relays make such forwarding and/or caching decisions, based on match of the identfiers associated in the object's header against the list of subscribers.
 
 ## Publisher Interactions
 TODO: This section shall cover relay handling of publishes.


draft-lcurley-warp.md
```diff
@@ -497,7 +497,17 @@ Relays serves as policy enforcement points by validating subscribe
 and publish requests to the tracks.
 
 ## Subscriber Interactions
-TODO: This section shall cover relay handling of subscriptions.
+
+Subscribers interact with the Relays by sending a "SUBSCRIBE REQEUST"  ({{message-subscribe-req}}) control message for the tracks of interest. Relays MUST be willing to act on behalf of the subscriptions before they can forward the media, which implies that the subscriptions MUST be authorized and it is done as follows:
+
+1. Provider for the tracks MUST be authorized. This requires identifying the Provider of the requested track and verifying that the Relay is authorized to serve that Provider. "Track Prefix" component of the "Track Name" MUST identify the Provider. Specifics of Provider authorization depends on the way the relay is managed and is typically based on prior business agreement with the Provider.
+
```
fluffy, line 504: Think we might want to add a bit more detail here

draft-lcurley-warp.md
```diff
@@ -497,7 +497,17 @@ Relays serves as policy enforcement points by validating subscribe
...
+2. Verify that the subscriber is authorized to access the specified content. Subscriptions MUST carrying enough authorization information proving the subscriber has access to the requested track. 
+
+In all the scenarios, the end-point client making the subscribe
+request is notified of the result of the subscription, via "SUBSCRIBE REPLY" ({{message-subscribe-reply}}) control message.
+
+For cases where the subscriptions are successfully validated, Relay proceeed to save the subscription information by maintaining the mapping from the track information to the list of subscribers. This will enable Relays to forward on-going publishes (live or from cache) to the subscribers, if available, and also forward all the future publishes, until the subscriptions cases to exist. Relays make such forwarding and/or caching decisions, based on match of the identfiers associated in the object's header against the list of subscribers.
```
fluffy, line 511: Should we discuss when the subscribe goes away ? or how state is cleaned up if connection closes ?

fluffy (approved):

afrind (commented):
Is it possible to rebase this PR on main so that it's only showing the deltas from main?  It's currently showing all the deltas in main that are unrelated and it makes it difficult to see what the proposed change is here.

draft-lcurley-warp.md
```diff
@@ -477,27 +578,36 @@ The endpoint breached an agreement, which MAY have been pre-negotiated by the ap
...
+
+- Verify that the subscriber is authorized to access the specified content associated with the "Full Track Name". Subscriptions MUST carry enough authorization information proving the subscriber has access to the requested track. Specifics of the authorization process depends on the way the relay is managed and is typically based on prior business agreement with the Origin, for example.
+
+In all the scenarios, the end-point client making the subscribe
+request is notified of the result of the subscription, via "SUBSCRIBE OK" ({{message-subscribe-ok}}) or the "SUBSCRIBE ERROR" {{message-subscribe-error}} control message.
+
+For successful subscriptions, Relay proceeds to save the subscription information by maintaining mapping from the track information to the list of subscribers. This will enable Relays to forward on-going publishes (live or from cache) to the subscribers, if available, and also forward all the future publishes, until the subscriptions cases to exist. A given susbcription ceases to exist because its expired or the publisher of the track stops producing media.
```
afrind, line 610: > forward all the future publishes

future OBJECTs related to the subscription?

afrind, line 610: > the end-point client making the subscribe request is notified of the result of the subscription

> Relay proceeds to save the subscription information by maintaining mapping from the track information to the list of subscribers

> A given susbcription ceases to exist because its expired or the publisher of the track stops producing media.

These bits seems the same for all publishers, not just relays, right? I think the group agreed to focus the relay section on behaviors that are specific/different for relays.

draft-lcurley-warp.md
```diff
@@ -335,60 +367,114 @@ A media object:
...
+~~~~~~~~~~~~~~~
+
+The `authority` portion MUST NOT contain a non-empty `userinfo` portion.
+The `moq` URI scheme supports the `/.well-known/` path prefix defined in {{!RFC8615}}.
+
+This protocol does not specify any semantics on the `path-abempty` and `query` portions of the URI.
+The contents of those is left up to the application.
+
+The client can establish a connection to a MoQ server identified by a given URI
+by setting up a QUIC connection to the host and port identified by the `authority` section of the URI.
```
fluffy, line 471: I would rather not have how we find the relays in this PR and instead just make a sperate section of the draft about a client decides what to connect to

afrind (commented):
Individual review

draft-lcurley-warp.md
```diff
@@ -335,60 +367,114 @@ A media object:
...
+The `path-abempty` and `query` portions of the URI are communicated to the server using
+the PATH parameter ({{path}}).
```
fluffy, line 473: where is the path paramter ? Is that in the SUB message ? If thinking the connect message then seems like it should not be here.

draft-lcurley-warp.md
```diff
@@ -335,60 +367,114 @@ A media object:
...
+The ALPN value {{!RFC7301}} used by the protocol is `moq-00`.
```
fluffy, line 474: again, this seems like it should be the stuff stuff where the client forms a connection to the server ( regadless of it the server is a relay or not ) and not in this PR

draft-lcurley-warp.md
```diff
@@ -426,8 +523,8 @@ This can be useful when the sender does not support stream prioritization.
 ## Relays
 Warp encodes the delivery information for a stream via OBJECT headers ({{message-object}}).
 
-A relay SHOULD prioritize streams ({{prioritization}}) based on the delivery order.
-A relay MAY change the delivery order, in which case it SHOULD update the value on the wire for future hops.
+A relay SHOULD prioritize streams ({{prioritization}}) based on the send order.
```
fluffy, line 526: I think thisw should just say do it the same way as client as specified in that section. Also not clear this should be in sub PR

draft-lcurley-warp.md
```diff
@@ -717,6 +894,30 @@ The ROLE parameter (key 0x00) allows the client to specify what roles it expects
 
 The client MUST send a ROLE parameter with one of the three values specified above. The server MUST close the connection if the ROLE parameter is missing, is not one of the three above-specified values, or it is different from what the server expects based on the application in question.
 
+## PATH parameter {#path}
+
+The PATH parameter (key 0x01) allows the client to specify the path of the MoQ URI when using native QUIC ({{native-quic}}).
```
fluffy, line 899: I'd rather just seem this be the same regardless of what the transport is.

fluffy (approved):
Sorry - prev review had included stuff from PR not rebased to latest.

draft-lcurley-warp.md
```diff
@@ -599,7 +599,15 @@ Relays serves as policy enforcement points by validating subscribe
 and publish requests to the tracks.
 
 ## Subscriber Interactions
-TODO: This section shall cover relay handling of subscriptions.
+
+Subscribers interact with the Relays by sending a "SUBSCRIBE REQUEST"  ({{message-subscribe-req}}) control message for the tracks of interest. Relays MUST be willing to act on behalf of the subscriptions before they can forward the media, which implies that the subscriptions MUST be authorized and it is done as follows:
```
kixelated, line 603: The MUST's here are confusing and just wrong.

A relay will receive a SUBSCRIBE and forward it upstream. If it performs authentication at each hop is optional. In fact the relay could forward any authentication token to the upstream, rather than validating it.

The one time a relay MUST authenticate a subscription is if it plans to coalesce subscriptions. This occurs when N downstream subscriptions turn into 1 upstream subscription. Authentication is still out of scope though so there's not much to really say.

draft-lcurley-warp.md
```diff
@@ -599,7 +599,15 @@ Relays serves as policy enforcement points by validating subscribe
...
+
+- Verify that the subscriber is authorized to access the specified content associated with the "Full Track Name". Subscriptions MUST carry enough authorization information proving the subscriber has access to the requested track. Specifics of the authorization process depends on the way the relay is managed and is typically based on prior business agreement with the Origin, for example.
```
kixelated, line 605: > Subscriptions MUST carry enough authorization information proving the subscriber has access to the requested track.

The session itself may be authorized, or the track may not require authentication. An auth token in a SUBSCRIBE is optional.

draft-lcurley-warp.md
```diff
@@ -599,7 +599,15 @@ Relays serves as policy enforcement points by validating subscribe
...
+
+In all the scenarios, the end-point client making the subscribe
```
kixelated, line 607: This is true for generic endpoints.

draft-lcurley-warp.md
kixelated: And this entire paragraph applies to generic endpoints, not a relay specifically.

kixelated (changes requested):
1. I'd like to see some text on how subscriptions are coalesced/cached, since it is actually specific to relays.
2. Authentication is currently out of scope, so some of the MUSTs surrounding authentication don't make any sense.
3. Most of this applies to generic endpoints. You could say `subscriptions MAY require authorization` and leave it at that.

draft-lcurley-warp.md
kixelated: Yeah, I think this text is still useful, but it should explain how the pub/sub messages work in general (not specific to relays).

suhasHere:
@vasilvv @kixelated @afrind @fluffy could you please review the latest updates. Thanks

draft-lcurley-warp.md
```diff
@@ -599,7 +599,14 @@ Relays serves as policy enforcement points by validating subscribe
 and publish requests to the tracks.
 
 ## Subscriber Interactions
-TODO: This section shall cover relay handling of subscriptions.
+
+Subscribers interact with the Relays by sending a "SUBSCRIBE REQUEST"  ({{message-subscribe-req}}) control message for the tracks of interest. Relays MUST ensure subscribers are authorized for the tracks. This is done by
+
+- Verifying that the subscriber is authorized to access the content associated with the "Full Track Name". The authorization information can be part of subscriptions themselves or part of the encompassing session. Specifics of the authorization process depends on the way the relay is managed and is typically based on prior business agreement with the Origin, for example.
```
vasilvv, line 605: Is this necessarily a relay-specific property?  I imagine any entity receiving a subscribe request would want to check if the sender is authorized to subscribe.

draft-lcurley-warp.md
```diff
@@ -599,7 +599,14 @@ Relays serves as policy enforcement points by validating subscribe
...
+
+In all the scenarios, the end-point client making the subscribe request is notified of the result of the subscription, via "SUBSCRIBE OK" ({{message-subscribe-ok}}) or the "SUBSCRIBE ERROR" {{message-subscribe-error}} control message.
+
+For successful subscriptions, relays proceed to save the subscription information by maintaining mapping from the track information to the list of subscriber. This will enable relays to forward on-going publishes (live or from cache) to the subscribers, if available, and also forward all the future publishes, until the subscriptions cases to exist. A given susbcription ceases to exist because its expired or the publisher of the track stops producing media. Relays MAY perform subscription aggregations, wherein only the unique subscriptions per "Full Track Name" are forwarded upstream for further processing.
```
vasilvv, line 609: > Relays MAY perform subscription aggregations, wherein only the unique subscriptions per "Full Track Name" are forwarded upstream for further processing.

I am not entirely sure what this means -- could you elaborate on what "aggregation" means here?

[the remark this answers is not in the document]
suhasHere: > Here's what I would say: A relay will have a list of frontend subscriptions (aka downstream) and a list of backend subscriptions (aka upstream). The relay maintains a map from each frontend subscription to a backend subscription, which it uses to forward received OBJECT messages from the backend.

I don't think we need to go in that detail. The current statement indicates there will be subscribers list per track requested in Subscribe Request.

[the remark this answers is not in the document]
suhasHere: > I don't understand the difference between "on-going publishes" and "future publishes", or what a "publish" is in the first place.

on-going publishes will happen when a subscriber asks for a track for which there is already media being published.

future-publishes indicates, as long as subscription is active and sender is producing media for the track of interest, the subscribers will receive the media

draft-lcurley-warp.md
suhasHere: yes and no. Agree that every entity should verify if the sender of subscribe is allowed to access the content. I do feel, how it is done is more nuanced at relays than endpoint client

draft-lcurley-warp.md
```diff
@@ -585,7 +585,17 @@ Relays serves as policy enforcement points by validating subscribe
...
+
+In all the scenarios, the end-point making the subscribe request is notified of the result of the subscription, via "SUBSCRIBE OK" ({{message-subscribe-ok}}) or the "SUBSCRIBE ERROR" {{message-subscribe-error}} control message.
+
+For successful subscriptions, relays proceed to save the subscription information by maintaining mapping from the track information to the list of subscribers. This will enable relays to forward any on-going matching publishes (live or from cache), if available and also forward all the future publishes for the requested track. Subscriptions stay active until it is expired or the publisher of the track stops producing media or other reasons that result in error (see {{message-subscribe-error}}).
```
kixelated, line 595: > Subscriptions stay active until it is expired or the publisher of the track stops producing media or other reasons that result in error (see {{message-subscribe-error}}).

I think the examples are fine, but a relay shouldn't care why the subscription ended. I would add a note that the relay SHOULD forward any received SUBSCRIBE ERROR message to downstream subscribers.

suhasHere:
@kixelated @vasilvv thoughts on the current text ?

draft-lcurley-warp.md
```diff
@@ -585,7 +585,17 @@ Relays serves as policy enforcement points by validating subscribe
...
+
+Relays MAY aggregate subscriptions for a given track when multiple subscribers request for the same track. Subscriptions aggregation allows relays to forward only the unique subscriptions per track for futher processing, say to setup routing for delivering media, rather than forwarding all the subscriptions received. When the authorization information is carried in the subscribes, the relay needs to verify the authorization information in the subscribe request, in order to deduplicate the subscriptions.
```
kixelated, line 597: > Subscriptions aggregation allows relays to forward only the unique subscriptions per track for futher processing, say to setup routing for delivering media, rather than forwarding all the subscriptions received.

I don't understand. Subscription aggregation is done to share a cache.

> When the authorization information is carried in the subscribes, the relay needs to verify the authorization information in the subscribe request, in order to deduplicate the subscriptions.

A relay MUST authorize a subscription prior to serving it from cache.

kixelated (approved):
it's fine as a first draft

draft-lcurley-warp.md
suhasHere: this is already said in the previous paragraph

PR #127, opened 2023-04-06
IETF116: Fill in Relay Publish Interaction Sub-Section

suhasHere:
This PR depends on #123 and #125 

The scope of this PR is to add Relay Publish interactions. This is expected to be a starter text and we can add further clarification and details in a separate PR as we develop further.

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -599,15 +599,15 @@ Relays MAY aggregate subscriptions for a given track when multiple subscribers r
 
 ## Publisher Interactions
 
-Publishing through the relay starts with publisher sending "ANNOUNCE" control message for the set of tracks, identified via their Full Track Names ({{model-track}}). Relays MUST ensure that publishers are authorized by:
+Publishing through the relay starts with publisher sending "ANNOUNCE" control message with a `Track Namespace` ({{model-track}}).
 
-- Verifying that the publisher is authorized to publish the content associated with the "Full Track Name". The authorization information can be part of annoucement themselves or part of the encompassing session. Specifics of the authorization process depends on the way the relay is managed and is typically based on prior business agreement with the Origin, for example.
+Relays MUST ensure that publishers are authorized by:
+
+- Verifying that the publisher is authorized to publish the content associated with the set of tracks whose Full Track Name share a common prefix with the announced namespace. Specifics of where the authorization happens, either at the relays or forwarded for further processing, depends on the way the relay is managed and is application specific (typically based on prior business agreement). If forwarded, the authorization information from the original annouce control message MUST be identical.
 
 Relays respond with "ANNOUNCE OK" and/or "ANNONCE ERROR" control messages providing the results of announcement.
 
-OBJECT message header carry short hop-by-hop `Track Id` that maps to the "Full Track Name". Relays MUST use the `Track Id` in the incoming OBJECT messages to find active subscribers with track information matching the `Track Id`. Relays MUST NOT depend on OBJECT payload content for making forwarding decisions and MUST only depend on the fields, such as priority order and other metadata properties in the OBJECT message header. Unless determined by congestion response, Relays MUST forward the OBJECT message to the matching subscribers. 
-
-__Note to authors: This above send behavior is common across all senders and once we have a sufficient text defined elsewhere, we can just refer to the appropriate section.__
+OBJECT message header carry short hop-by-hop Track Id that maps to the Full Track Name (see {{message-subscribe-ok}}). Relays use the Track ID of an incoming OBJECT message to identify its track and find the active subscribers for that track. Relays MUST NOT depend on OBJECT payload content for making forwarding decisions and MUST only depend on the fields, such as priority order and other metadata properties in the OBJECT message header. Unless determined by congestion response, Relays MUST forward the OBJECT message to the matching subscribers. 
 
 ## Relay Discovery and Failover
 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -615,11 +615,18 @@ Relays make such forwarding and/or caching decisions, based on match of the iden
 
 
 ## Publisher Interactions
-Publishers MAY be configured to publish the objects to a Relays based on the application configuration and topology.  Publishing set of tracks through the Relay starts with a "PUBLISH" control messages that identifies the tracks via their Track Names ({{model-track}}). 
 
-As specified with subscriber interactions, Relays MUST be authorized to serve a given track's Provider and the publisher MUST be authorized to publish content on the tracks advertised in the "PUBLISH" message.
+Publishers MAY be configured to publish the objects to a Relays based on the application configuration and topology. Publishing set of tracks through the Relay starts with a "ANNOUNCE" control messages that identifies the tracks via their Full Track Names ({{model-track}}). 
 
-Relays makes use of priority order and other metadata properties from the published objects to make forward or drop decisions when reacting to congestion as indicated by the underlying QUIC stack.  The same can be used to make caching decisions.
+The "ANNOUNCE" message advertises set of tracks and their authorization information. For each track in the "ANNOUNCE" message,: 
+
+1. For cases where the "Track Namespace" component of the "Full Track Name" is the Origin domain, Relays MUST ensure the Origin is authorized. 
+
+2. Verify that the publisher is authorized to produce on the specified track by validating the authorization information.
+
+Relays respond with "ANNOUNCE OK" and/or "ANNONCE ERROR" control messages reflecting the authorization status.
+
+Relays match the identifiers for the tracks in the OBJECT message published against the list of active subscribers for making forwarding decisions. Relays makes use of priority order and other metadata properties from the published objects to make forward or drop decisions when reacting to congestion as indicated by the underlying QUIC stack.  The same can be used to make caching decisions.
 
 ## Relay Discovery and Failover
 TODO: This section shall cover aspects of relay failover and protocol interactions
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -610,7 +610,7 @@ Publishing through the relay starts with publisher sending "ANNOUNCE" control me
 
 Relays MUST ensure that publishers are authorized by:
 
-- Verifying that the publisher is authorized to publish the content associated with the set of tracks whose Full Track Name share a common prefix with the announced namespace. Specifics of where the authorization happens, either at the relays or forwarded for further processing, depends on the way the relay is managed and is application specific (typically based on prior business agreement). If forwarded, the authorization information from the original annouce control message MUST be identical.
+- Verifying that the publisher is authorized to publish the content associated with the set of tracks whose Full Track Name share a common prefix with the announced namespace. Specifics of where the authorization happens, either at the relays or forwarded for further processing, depends on the way the relay is managed and is application specific (typically based on prior business agreement).
 
 Relays respond with "ANNOUNCE OK" and/or "ANNONCE ERROR" control messages providing the results of announcement.
 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -500,7 +500,11 @@ and publish requests to the tracks.
 TODO: This section shall cover relay handling of subscriptions.
 
 ## Publisher Interactions
-TODO: This section shall cover relay handling of publishes.
+Publishers MAY be configured to publish the objects to a Relays based on the application configuration and topology.  Publishing set of tracks through the Relay starts with a "PUBLISH" control messages that identifies the tracks via their Track Names ({{model-track}}). 
+
+As specified with subscriber interactions, Relays MUST be authorized to serve a given track's Provider and the publisher MUST be authorized to publish content on the tracks advertised in the "PUBLISH" message.
+
+Relays makes use of priority order and other metadata properties from the published objects to make forward or drop decisions when reacting to congestion as indicated by the underlying QUIC stack.  The same can be used to make caching decisions.
 
 ## Relay Discovery and Failover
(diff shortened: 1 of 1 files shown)


draft-lcurley-warp.md
```diff
@@ -500,7 +500,11 @@ and publish requests to the tracks.
 TODO: This section shall cover relay handling of subscriptions.
 
 ## Publisher Interactions
-TODO: This section shall cover relay handling of publishes.
+Publishers MAY be configured to publish the objects to a Relays based on the application configuration and topology.  Publishing set of tracks through the Relay starts with a "PUBLISH" control messages that identifies the tracks via their Track Names ({{model-track}}). 
+
+As specified with subscriber interactions, Relays MUST be authorized to serve a given track's Provider and the publisher MUST be authorized to publish content on the tracks advertised in the "PUBLISH" message.
+
+Relays makes use of priority order and other metadata properties from the published objects to make forward or drop decisions when reacting to congestion as indicated by the underlying QUIC stack.  The same can be used to make caching decisions.
```
fluffy, line 507: I would say they act like clients in deciding what to send next and reference the priority section. ( which we need to add )

draft-lcurley-warp.md
```diff
@@ -500,7 +500,11 @@ and publish requests to the tracks.
 TODO: This section shall cover relay handling of subscriptions.
 
 ## Publisher Interactions
-TODO: This section shall cover relay handling of publishes.
+Publishers MAY be configured to publish the objects to a Relays based on the application configuration and topology.  Publishing set of tracks through the Relay starts with a "PUBLISH" control messages that identifies the tracks via their Track Names ({{model-track}}). 
+
+As specified with subscriber interactions, Relays MUST be authorized to serve a given track's Provider and the publisher MUST be authorized to publish content on the tracks advertised in the "PUBLISH" message.
+
```
fluffy, line 506: Need more detail on how they are authorized

draft-lcurley-warp.md
```diff
@@ -610,7 +610,18 @@ request is notified of the result of the subscription, via "SUBSCRIBE OK" ({{mes
...
+The "ANNOUNCE" message advertises set of tracks, identified via their Full Track Names ({{model-track}}) and carry necessary authorization information. Relays MUST 
+
+- Verify that the publisher is authorized to produce media for a given "Full Track Name" by validating the authorization information. Specifics of the authorization process depends on the way the relay is managed and is typically based on prior business agreement with the Origin, for example.
+
+Relays respond with "ANNOUNCE OK" and/or "ANNONCE ERROR" control messages reflecting the authorization status.
+
+Relays MUST match the track id in the OBJECT message header against the active subscribers for a given track. Relays MUST NOT depend on OBJECT payload content for making forwarding decisions and MUST only depend on the fields, such as priority order and other metadata properties in the OBJECT message header. Unless determined by congestion response, Relays MUST forward the OBJECT message to the matching subscribers.
```
fluffy, line 622: I think we could get clearer text on "Relays MUST match the track id in the OBJECT message header against the active subscribers for a given track." part. Perhaps "Relays match the trackID in the object header agains the  trackID created in the Annouce transaction to know which track the object is for."

fluffy (changes requested):

fluffy:
I think we still need to deal with the issue of who creates the TrackID but seems that can be a separate PR.

draft-lcurley-warp.md
kixelated, line 622: > Relays MUST match the track id in the OBJECT message header against the active subscribers for a given track.

`track ID` is a actually hop-by-hop property (should be renamed to `subscription ID`). Each downstream subscription may use different subscription IDs, which may not match the upstream subscription ID either, so it's not quite as simple as matching them. 

And again these are weird MUSTs. What happens if a relay... doesn't do this? In fact I would want relays to be able to rewrite track names, so it's not clear how that would even work with this assertive yet unenforceable statement.

draft-lcurley-warp.md
```diff
@@ -610,7 +610,18 @@ request is notified of the result of the subscription, via "SUBSCRIBE OK" ({{mes
 For successful subscriptions, Relay proceeds to save the subscription information by maintaining mapping from the track information to the list of subscribers. This will enable Relays to forward on-going publishes (live or from cache) to the subscribers, if available, and also forward all the future publishes, until the subscriptions cases to exist. A given susbcription ceases to exist because its expired or the publisher of the track stops producing media. 
 
 ## Publisher Interactions
-TODO: This section shall cover relay handling of publishes.
+
+Publishers MAY be configured to publish the objects to a Relays based on the application configuration and topology. Publishing set of tracks through the Relay starts with a "ANNOUNCE" control message
+
+The "ANNOUNCE" message advertises set of tracks, identified via their Full Track Names ({{model-track}}) and carry necessary authorization information. Relays MUST
```
kixelated, line 616: Are multiple ANNOUNCE messages allowed? What happens when authentication needs to be renewed? Can an endpoint UNANNOUNCE tracks...? What happens if another endpoint ANNOUNCEs the same track (ex. reconnection)?

kixelated: How does the receiver identify tracks? 

In order to support ingestion from generic clients (like OBS), we need some way of identifying the catalog; it can't be negotiated out-of-band. I included the track format in my proposal to this end. 

@vasilvv suggested ending the track name with a `/` if it's a catalog, which is kind of neat since a catalog is like a directory listing. I think ending with `/catalog_v0` is more future proof, although I would like to see a registry rather than encoding behavior into the track name,.

kixelated (changes requested):

draft-lcurley-warp.md
```diff
@@ -610,7 +610,18 @@ request is notified of the result of the subscription, via "SUBSCRIBE OK" ({{mes
...
 ## Publisher Interactions
-TODO: This section shall cover relay handling of publishes.
+
+Publishers MAY be configured to publish the objects to a Relays based on the application configuration and topology. Publishing set of tracks through the Relay starts with a "ANNOUNCE" control message
+
+The "ANNOUNCE" message advertises set of tracks, identified via their Full Track Names ({{model-track}}) and carry necessary authorization information. Relays MUST 
+
+- Verify that the publisher is authorized to produce media for a given "Full Track Name" by validating the authorization information. Specifics of the authorization process depends on the way the relay is managed and is typically based on prior business agreement with the Origin, for example.
+
+Relays respond with "ANNOUNCE OK" and/or "ANNONCE ERROR" control messages reflecting the authorization status.
```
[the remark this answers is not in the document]
kixelated: edit: I do think we need these messages primarily to add an `expires` field in ANNOUNCE OK similar to SUBSCRIBE OK. I would really like to combine authentication in both though.

draft-lcurley-warp.md
```diff
@@ -609,15 +609,27 @@ In all the scenarios, the end-point client making the subscribe request is notif
...
+
+- Verifying that the publisher is authorized to publish the content associated with the "Full Track Name". The authorization information can be part of annoucement themselves or part of the encompassing session. Specifics of the authorization process depends on the way the relay is managed and is typically based on prior business agreement with the Origin, for example.
+
+Relays respond with "ANNOUNCE OK" and/or "ANNONCE ERROR" control messages providing the results of announcement.
+
+OBJECT message header carry short hop-by-hop `Track Id` that maps to the "Full Track Name". Relays MUST use the `Track Id` in the incoming OBJECT messages to find active subscribers with track information matching the `Track Id`. Relays MUST NOT depend on OBJECT payload content for making forwarding decisions and MUST only depend on the fields, such as priority order and other metadata properties in the OBJECT message header. Unless determined by congestion response, Relays MUST forward the OBJECT message to the matching subscribers.
```
afrind, line 619: Individual/Editorial comment

> Relays MUST use the `Track Id` in the incoming OBJECT messages to find active subscribers with track information matching the `Track Id`.

This MUST reads a little prescriptive to me.  Maybe rewrite the sentence without the MUST and make it informative rather than normative, like:

`Relays use the `Track ID` of an incoming OBJECT message to identify its track and find the active subscribers for that track`

suhasHere: Without that MUST, I am not sure what's the source of truth in object message to find the track information

draft-lcurley-warp.md
suhasHere: i do think we need to answer/clarify on some of these , but it's not in the scope of this PR.

draft-lcurley-warp.md
```diff
@@ -598,15 +598,27 @@ Relays MAY aggregate subscriptions for a given track when multiple subscribers r
 
 
 ## Publisher Interactions
-TODO: This section shall cover relay handling of publishes.
+
+Publishing through the relay starts with publisher sending "ANNOUNCE" control message with a `Track Namespace` ({{model-track}}).
+
+Relays MUST ensure that publishers are authorized by:
+
+- Verifying that the publisher is authorized to publish the content associated with the set of tracks whose Full Track Name share a common prefix with the announced namespace. Specifics of where the authorization happens, either at the relays or forwarded for further processing, depends on the way the relay is managed and is application specific (typically based on prior business agreement). If forwarded, the authorization information from the original annouce control message MUST be identical.
```
afrind, line 606: Same concern about common prefix -- can we say "Full Track Name matches the announced namespace"?

afrind (commented):
Individual review

draft-lcurley-warp.md
suhasHere: Full Track Name includes both Track Namespace and Track Name, so here the match should happen only on the namespace part. I wonder if the suggestion is saying namespace is the full track name

afrind: Hmm, yeah I agree that could be a reading of it which I didn't intend. Maybe:

> set of tracks whose Track Namespace matches the announced namespace

kixelated, line 606: This text is duplicated in both the ANNOUNCE and SUBSCRIBE section... but is slightly different. Can you move it to it's own "authorization" sub-second?

draft-lcurley-warp.md
```diff
@@ -598,15 +598,27 @@ Relays MAY aggregate subscriptions for a given track when multiple subscribers r
...
+
+Relays respond with "ANNOUNCE OK" and/or "ANNONCE ERROR" control messages providing the results of announcement.
+
+OBJECT message header carry short hop-by-hop Track Id that maps to the Full Track Name (see {{message-subscribe-ok}}). Relays use the Track ID of an incoming OBJECT message to identify its track and find the active subscribers for that track. Relays MUST NOT depend on OBJECT payload content for making forwarding decisions and MUST only depend on the fields, such as priority order and other metadata properties in the OBJECT message header. Unless determined by congestion response, Relays MUST forward the OBJECT message to the matching subscribers.
```
kixelated, line 610: > OBJECT message header carry short hop-by-hop Track Id that maps to the Full Track Name (see {{message-subscribe-ok}}). Relays use the Track ID of an incoming OBJECT message to identify its track and find the active subscribers for that track.

What does this have to do with publishing? This whole paragraph should be in the subscription section... it even mentions subscribers.

> Unless determined by congestion response, Relays MUST forward the OBJECT message to the matching subscribers.

If it's not enforceable, then it's a SHOULD and not a MUST. We should also have a discussion about if a transmuxer / transcoder is a "relay".

afrind: > We should also have a discussion about if a transmuxer / transcoder is a "relay".

I think we have had this discussion, right?  A relay doesn't have access to the raw media / e2e encryption keys so a transmuxer / transcoder is not a relay, it's a subscriber and publisher glued together.

suhasHere: i was defining transcoder as https://datatracker.ietf.org/doc/html/draft-nandakumar-moq-arch-00#section-2.2

it has following security roles publisher, subscriber, emitter and receiver

suhasHere: > What does this have to do with publishing? This whole paragraph should be in the subscription section... it even mentions subscribers.

this is needed to tell what is relay doing when a publish happens

draft-lcurley-warp.md
suhasHere: once we refactor to have common sender/receiver section , it would be more clearer to move things around inside the relay section. I know its duplicate, but would prefer to fix in a separate PR though

afrind: @suhasHere : do you agree with matching namespaces rather than matching prefixes?

draft-lcurley-warp.md
```diff
@@ -609,15 +609,27 @@ In all the scenarios, the end-point client making the subscribe request is notif
...
 
 ## Publisher Interactions
-TODO: This section shall cover relay handling of publishes.
+
+Publishing through the relay starts with publisher sending "ANNOUNCE" control message for the set of tracks, identified via their Full Track Names ({{model-track}}). Relays MUST ensure that publishers are authorized by:
+
+- Verifying that the publisher is authorized to publish the content associated with the "Full Track Name". The authorization information can be part of annoucement themselves or part of the encompassing session. Specifics of the authorization process depends on the way the relay is managed and is typically based on prior business agreement with the Origin, for example.
+
+Relays respond with "ANNOUNCE OK" and/or "ANNONCE ERROR" control messages providing the results of announcement.
```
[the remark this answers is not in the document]
suhasHere: This section is about Relays and using relays goes along with previous page. I think the above change can be part of #153

suhasHere:
@kixelated  I did try to respond to your comments. I do agree in general with refactoring suggestions, however, I feel it can done as part of #153

draft-lcurley-warp.md
suhasHere: yes @afrind

kixelated (approved):
it's fine, we can do an editorial pass later to consolidate the publisher/subscriber/authentication/relay text

PR #129, opened 2023-04-07
IETF116 : Add a placeholder section for dealing with congestion response

suhasHere:
This PR adds a placeholder section for inserting details on various congestion response and prioritization proposals being discussed.  At IETF116 , it was very clear that there are couple of ways this can be address and having its own section would help along as we go further.


fluffy (approved):
I think merging something like this now will make it much easier to have PR that don't all depend on each other and and allow the priority stuff to get discussed in one section.

vasilvv (approved):

PR #130, opened 2023-04-07
IETF116: Move Transport Usages to its own section

suhasHere:
This PR adds a starter text for moving WT and Native QUIC usage to its own section.

Follow up PRs should focus on how to concentrate transport specific things to this section and transport independent usages (control stream, data streams, data model mapping to the streams) should be covered in its own PR

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -73,6 +73,7 @@ Warp is a live media transport protocol that utilizes the QUIC network protocol 
 
 * {{motivation}} covers the background and rationale behind Warp.
 * {{objects}} covers how media is fragmented into objects.
+* {{transport-usages}} covers aspects of setting up a MoQ transport connection.
 * {{quic}} covers how QUIC is used to transfer media.
 * {{messages}} covers how messages are encoded on the wire.
 * {{containers}} covers how media tracks are packaged.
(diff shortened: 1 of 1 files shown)


draft-lcurley-warp.md
```diff
@@ -387,6 +393,23 @@ For example, an identifier and authentication token could be included in the pat
...
+  n00".
+
+*  Bilateral and unilateral streams are mapped directly to equivalent QUIC streams.
+
+TODO: Add futher details on specifcs of each transport in setting up the session and catalog retrieval.
+
+# QUIC
+
+## Connection 
+As defined in {{transport-usages}}, either WebTransport or Native QUIC can be used to setup underlying QUIC primitives for carrying out the protocol defined in this specification.
```
fluffy, line 411: Might add that nearly all parts of the protocol are the same regardless of if MoQ is running on WebTrasnport or raw QUIC.

draft-lcurley-warp.md
```diff
@@ -686,6 +709,20 @@ The ROLE parameter (key 0x00) allows the client to specify what roles it expects
 
 The client MUST send a ROLE parameter with one of the three values specified above. The server MUST close the connection if the ROLE parameter is missing, is not one of the three above-specified values, or it is different from what the server expects based on the application in question.
 
+
```
fluffy, line 712: Did you intent to move this section ? Not sure I following reason for move here.

fluffy (approved):

draft-lcurley-warp.md
suhasHere: good point, will do

draft-lcurley-warp.md
suhasHere: Hmm .. that was no the intent. Let me fix it in the next update. Thanks for catching it.

vasilvv (approved):

PR #134, opened 2023-04-10
Generic transport draft.

kixelated:
The plan is to split the draft into two parts: one for generic transport and one for media the media mapping (catalog).

There's still a lot of references to media that need to be cleaned up. I tried to limit the changes by primarily deleting sections, only changing the text if there was a reference that no longer worked.

A follow-up PR will address any lingering media references. I'll deal with the inevitable merge conflicts.


suhasHere:
i wonder if we should rename the protocol as well to avoid confusions with Warp media format work that @wilaw started?

kixelated:
> i wonder if we should rename the protocol as well to avoid confusions with Warp media format work that @wilaw started?

Yep yep, will do. The plan is `warp-transport` and `warp-media`.

suhasHere:
> Yep yep, will do. The plan is `warp-transport` and `warp-media`. I just didn't want to rename the file in this PR so it's easy to review.

my 2 cents would be to call the base transport protocol as just `moq-transport` or `moq-transport-protocol` or something generic.. warp-media name seems fine to the media spec though

draft-lcurley-warp.md
```diff
@@ -214,8 +152,7 @@ An object group is a sequence of media objects. Beginning of an object group can
 DISCUSS: We need to determine what are the exact requirements we need to impose on how the media objects depend on each other. Such requirements would need to address the use case (a join point), while being flexible enough to accomodate scenarios like B-frames and temporal scaling.
 
 ## Track
-
-A media track in Warp is a combination of *an init object* and a sequence of media object groups. An init object is a format-specific self-contained description of the track that is required to decode any media object contained within the track, but can also be used as the metadata for track selection. If two media tracks carry semantically equivalent but differently encoded media, they are referred to as *variants* of each other.
+A track in Warp is a combination of *an init object* and a sequence of object groups. An init object is a format-specific self-contained description of the track that is required to decode any object contained within the track, but can also be used as the metadata for track selection.
```
vasilvv, line 155: I thought we no longer have init objects, since those were in the catalog, and we're moving catalog out of transport?

kixelated: Yeh let me do another pass and remove more media stuff.

kixelated:
@suhasHere We can replace the warp prefix with moq as part of adoption. I just don't feel comfortable claiming that an individual draft represents moq until the working group agrees.

kixelated:
Actually let's close this until the media half is fully ready. 

Let's merge/close the existing PRs to avoid a nightmare of merge conflicts. @suhasHere no new PRs, okay?

Issue #107, opened 2023-03-28
Stabilizing the handshake format


vasilvv:
If we are going to ship experimental versions of MoQ (which we should), those probably should be interoperable on a handshake level (meaning that a server can accept two clients speaking different iterations of MoQ).  We currently have the following format in the draft:

```
SETUP Parameter {
  Parameter Key (i),
  Parameter Value Length (i),
  Parameter Value (..),
}

Client SETUP Message Payload {
  Number of Supported Versions (i),
  Supported Version (i) ...,
  SETUP Parameters (..) ...,
}

Server SETUP Message Payload {
  Selected Version (i),
  SETUP Parameters (..) ...,
}
```

Astute reader will notice that this negotiation mechanism is similar to [a certain other IETF protocol](https://tlswg.org/tls13-spec/draft-ietf-tls-rfc8446bis.html#name-supported-versions).

The question I want to explicitly ask here is: are we okay with this format?  We obviously can change it later down the line, but that would be tricky since we'd create an incompatibility (and we also don't have the Alt-Svc escape hatch that QUIC used).

fluffy:
Seems like a down the road issue, but sure , this would work to for now and we can make sure we  rethink it as we get closer to done. I had assumed we would have a MoQ ALPN too. Would that be part of this ?

afrind:
>  we would have a MoQ ALPN too.

This doesn't work for the WebTransport case, which will use "h3" ALPN.  Perhaps there could be the ALPN equivalent in the CONNECT headers?

vasilvv:
> Seems like a down the road issue, but sure , this would work to for now and we can make sure we rethink it as we get closer to done. I had assumed we would have a MoQ ALPN too. Would that be part of this ?

No, I was mostly concerned about this for MoQ over WebTransport.  For MoQ over raw QUIC, it's actually entirely possible to just use ALPN for version negotiation (which is what H/3 does).

----------

I am closing this, since it appears we did have a very rough agreement at 116 (or at least no one seemed to object).

PR #135, opened 2023-04-12
Define the protocol over raw QUIC

vasilvv:


draft-lcurley-warp.md
```diff
@@ -364,55 +369,38 @@ Objects MUST synchronize frames within and between tracks using presentation tim
...
+A Warp server that is accessible via raw QUIC can be identified by a URI with a "moq" scheme.
+The "moq" URI scheme is defined as follows, using definitions from {{!RFC3986}}:
 
-* Connection setup corresponds to the establishment of a QUIC
-  connection, in which the ALPN value indicates use of MoQ.  For
-  versions implementing this draft, the ALPN value is set to "moq-
-  n00".
+~~~~~~~~~~~~~~~
+moq-URI = "moq" "://" authority path-abempty [ "?" query ]
+~~~~~~~~~~~~~~~
```
suhasHere, line 391: for my better understanding on why we need a elaborate URI scheme
- why do we need path ? and how is it set ?
- what is the usecase for query ?

draft-lcurley-warp.md
```diff
@@ -364,55 +369,38 @@ Objects MUST synchronize frames within and between tracks using presentation tim
...
-The server uses the HTTP CONNECT request for identification and authorization of a track bundle.
-The specific mechanism is left up to the application.
-For example, an identifier and authentication token could be included in the path.
+A Warp server that is accessible via WebTransport can be identified using an HTTPS URI ({{!RFC9110, Section 4.2.2}}).
+A Warp transport session can be established by sending an extended CONNECT request to the host and the path indicated by the URI,
+as described in {{WebTransport, Section 3}}.
 
-The server MAY return an error status code for any reason, for example a 403 when the client is forbidden.
-Otherwise the server MUST respond with a "200 OK" to establish the WebTransport session.
+## Raw QUIC
```
suhasHere, line 385: Native QUIC would be my inclination to call it out though :-)

draft-lcurley-warp.md
```diff
@@ -711,6 +701,15 @@ The ROLE parameter (key 0x00) allows the client to specify what roles it expects
...
 
+## PATH parameter {#path}
+
+The PATH parameter (key 0x01) allows the client to specify the path of the MoQ URI when using raw QUIC ({{raw-quic}}).
+It MUST NOT be used by the server, or when WebTransport is used.
+If the peer receives a PATH parameter from the server, or when WebTransport is used, it MUST close the connection.
+
+When connecting to a server using a URI with the "moq" scheme,
+the client MUST set the PATH parameter to the `path-abempty` portion of the URI;
+if `query` is present, the client MUST concatenate `?`, followed by the `query` portion of the URI to the parameter.
```
huitema, line 713: There is no "path" parameter in raw QUIC. If we do want one, we will have to define some kind of "SETTINGS" transaction to specialize the QUIC/MoQ connection, but it may be a bit early to define that. I would replace this paragraph by a place-holder for now, something like: "TODO: define whether the `path` parameter is actually needed for raw QUIC connections, and if it is define how to carry it. Waiting for that, assume that the `path` parameter is empty when using raw QUIC."

huitema (changes requested):
The definition of raw QUIC suggests using a "path" parameter, but there is no such thing in QUIC itself. The use of the path parameter is largely linked to the discussion of bundling, and that discussion has not yet concluded. Waiting for that, I believed we have agreed to not consider bundling in the version-00 of the transport draft, and to simply let the client subscribe to a bunch of tracks, each with their own URI. Maybe mark the whole "path" discussion as "for further study"?

draft-lcurley-warp.md
vasilvv: Renamed.

draft-lcurley-warp.md
vasilvv: It isn't really elaborate -- I just made it equivalent to what HTTP supports.  We could potentially drop path/query, though that can be a useful routing lever.

draft-lcurley-warp.md
vasilvv: > If we do want one, we will have to define some kind of "SETTINGS" transaction to specialize the QUIC/MoQ connection, but it may be a bit early to define that.

This document already defines one (and we in fact discussed in #107).

draft-lcurley-warp.md
```diff
@@ -364,55 +369,38 @@ Objects MUST synchronize frames within and between tracks using presentation tim
...
+~~~~~~~~~~~~~~~
 
+The `authority` portion MUST NOT contain a non-empty `userinfo` portion.
+The `moq` URI scheme supports the `/.well-known/` path prefix defined in {{!RFC8615}}.
 
-# QUIC
+The client can establish a connection to a MoQ server identified by a given URI
+by setting up a QUIC connection to the host and port identified by the `authority` section of the URI.
+The `path-abempty` and `query` portions of the URI are communicated to the server using
+the PATH parameter ({{path}}).
```
kixelated, line 399: Needs some further explanation.

> ...in the SETUP message after establishing the QUIC connection.

draft-lcurley-warp.md
kixelated: Yeah, this `SETUP` message is that `SETTINGS` transaction you're talking about @huitema .

kixelated (approved):
LGTM

draft-lcurley-warp.md
huitema: Yes, that would work. I am a bit worried about the setting mechanism itself -- "the first bidir stream from the client" seems like a bit underspecified, given that the order in which the stream opening order is not guaranteed the same as client and server. But assuming we fix that, yes, sending "role", "path", and probably "version" on a setup stream makes sense. In raw QUIC, I would assume that the setup could be sent in 0-RTT.

suhasHere: just want to make sure that we are moving the path and query parameter to SETTINGS ? since atleast for now, I am not sure how it applies to native QUIC ..

suhasHere (changes requested):
just want to make sure that we are moving the path and query parameter to SETTINGS ? since atleast for now, I am not sure how it applies to native QUIC ..

kixelated:
> just want to make sure that we are moving the path and query parameter to SETTINGS ? since atleast for now, I am not sure how it applies to native QUIC ..

No Suhas, the path is in the SETUP message. There's no SETTINGS message for native QUIC. 

I was alluding to how the SETUP message performs a similar role to the HTTP/3 SETTINGS message. Both endpoints negotiate functionality, but in the case of native QUIC, also exchange a connection URL.

draft-lcurley-warp.md
huitema: I really wonder what the point of QUERY parameters would be, if we have an explicit SETUP process. That would mean two option for specifying parameters: as SETUP type/value, or as query items. Leading to potential conflicts, etc.

We had discussions in previous meetings of using the Path parameter to specify some kind of "bundle", maybe a common origin for the tracks. But that too would be better handled by a specific SETUP type/value.

That means we could also simplify the HTTP specification, and explain that the only purpose of the path-abempty would be to distinguish MoQ from other webtransport services implemented on the server -- pretty much the equivalent of the ALPN.

suhasHere:
> > just want to make sure that we are moving the path and query parameter to SETTINGS ? since atleast for now, I am not sure how it applies to native QUIC ..
> 
> The path is in the SETUP message. There's no SETTINGS message for native QUIC and it's transparent anyway to MoQ over WebTransport.
> 
> I was alluding to how the SETUP message performs a similar role to the HTTP/3 SETTINGS message. Both endpoints negotiate functionality, but in the case of native QUIC, also exchange a connection URL.


Sorry for the confusion , yes I meant SETUP message. The PR currently specifies PATH and query in the connection URL for native QUIC, which may be reworded under a TODO for now. Other than that, we should be fine ?

suhasHere (approved):

PR #147, opened 2023-04-26
Resolve group DISCUSS in object model

kixelated:
It was quite difficult to get the wording correct and I'm open to suggestions.

The main difficult is that groups are not truly independent, particularly for audio decoding. In Yokohama I learned I was mistaken in the assumption that audio consists of independent "frames"; technically it self-corrects any errors. This was extra apparent in my sample code as reseting the decoder at each group caused noticeable blips (we can remove this anecdote from the text if it's not useful). Video works the same way when using intra-refresh so we need to clearly spell it out.

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -214,20 +214,14 @@ to span more than one useable unit may create more than one viable application
 mapping from media to wire format, which could be confusing for protocol users.
 
 ## Groups {#model-group}
-A *group* is sequence of objects.
-Each group is independent and behaves as a join point for new subscribers.
+A *group* is collection of objects, part of a larger track ({{model-track}}).
 
-An object MUST be decodable after the delivery of all prior objects within the same group.
-This implies that the first object within a group MUST NOT depend on other objects.
-The application MAY use the sequence number to perform ordering and detect gaps within a group.
+A group behaves as a join point for subscriptions.
+A new subscriber may not want to receive the entire track, and will instead opt to receive only the latest group(s).
+The sender then selectively transmits objects based on their group membership.
 
-A sender SHOULD selectively transmit objects based on group membership.
-For example, the sender could transmit only the newest group to new subscriber, as the objects are then guaranteed to be decodable.
-However, the sender MUST still transmit objects based on the priority {{send-order}}, not based on the group sequence number.
-
-A decoder MAY use objects from prior groups when using a self-correcting encoding scheme.
-For example, an audio decoder should not be reinitialized at group boundaries, otherwise it causes a noticeable blip.
-Likewise, an video decoder may temporarily reference a prior group when using intra-refresh encoding.
+The application is responsible for how objects are placed into groups.
+In general, objects within a group SHOULD NOT depend on objects in other groups.
 
 
 ## Track {#model-track}
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -214,8 +214,8 @@ to span more than one useable unit may create more than one viable application
 mapping from media to wire format, which could be confusing for protocol users.
 
 ## Groups {#model-group}
-A *group* is an independent sequence of objects.
-Each group is a join point, such that a new subscriber can begin receiving a track without prior progress.
+A *group* is sequence of objects.
+Each group is independent and behaves as a join point for new subscribers.
 
 An object MUST be decodable after the delivery of all prior objects within the same group.
 This implies that the first object within a group MUST NOT depend on other objects.
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -659,7 +659,6 @@ The track identifier obtained as part of subscription and/or publish control mes
 
 * Group Sequence :
 The object is a member of the indicated group {{model-group}} within the track.
-The sequence starts at 0, increasing sequentially for each group within the track.
 
 * Object Sequence:
 The order of the object within the group.
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -214,10 +214,21 @@ to span more than one useable unit may create more than one viable application
 mapping from media to wire format, which could be confusing for protocol users.
 
 ## Groups {#model-group}
+A *group* is an independent sequence of objects.
+Each group is a join point, such that a new subscriber can begin receiving a track without prior progress.
 
-An object group is a sequence of media objects. Beginning of an object group can be used as a point at which the receiver can start consuming a track without having any other object groups available. Object groups have an ID that identifies them uniquely within a track.
+An object MUST be decodable after the delivery of all prior objects within the same group.
+This implies that the first object within a group MUST NOT depend on other objects.
+The application MAY use the sequence number to perform ordering and detect gaps within a group.
 
-DISCUSS: We need to determine what are the exact requirements we need to impose on how the media objects depend on each other. Such requirements would need to address the use case (a join point), while being flexible enough to accomodate scenarios like B-frames and temporal scaling.
+A sender SHOULD selectively transmit objects based on group membership.
+For example, the sender could transmit only the newest group to new subscriber, as the objects are then guaranteed to be decodable.
+However, the sender MUST still transmit objects based on the priority {{send-order}}, not based on the group sequence number.
+
+A decoder MAY use objects from prior groups when using a self-correcting encoding scheme.
+For example, an audio decoder should not be reinitialized at group boundaries, otherwise it causes a noticeable blip.
+Likewise, an video decoder may temporarily reference a prior group when using intra-refresh encoding.
+
 
 ## Track {#model-track}
 
@@ -653,12 +664,12 @@ OBJECT Message {
 The track identifier obtained as part of subscription and/or publish control message exchanges.
 
 * Group Sequence :
-An integer always starts at 0 and increases sequentially at the original media publisher.
-Group sequences are scoped under a Track.
+The object is a member of the indicated group {{model-group}} within the track.
+The sequence starts at 0, increasing sequentially for each group within the track.
 
 * Object Sequence:
-An integer always starts at 0 with in a Group and increases sequentially.
-Object Sequences are scoped to a Group.
+The order of the object within the group.
+The sequence starts at 0, increasing sequentially for each object within the group.
 
 * Object Send Order:
 An integer indicating the object send order ({{send-order}}).


kixelated:
Replaces #90.

draft-lcurley-warp.md
```diff
@@ -214,10 +214,21 @@ to span more than one useable unit may create more than one viable application
 mapping from media to wire format, which could be confusing for protocol users.
 
 ## Groups {#model-group}
+A *group* is an independent sequence of objects.
+Each group is a join point, such that a new subscriber can begin receiving a track without prior progress.
```
afrind, line 218: 'receiving' doesn't strike me as the right verb.  'consuming' as in the previous text was probably ok, or using/decoding/processing?

afrind, line 218: 'prior progress' reads awkwardly to me.  Something similar to the previous text might work "without access to any other groups"

draft-lcurley-warp.md
```diff
@@ -214,10 +214,21 @@ to span more than one useable unit may create more than one viable application
...
 
-An object group is a sequence of media objects. Beginning of an object group can be used as a point at which the receiver can start consuming a track without having any other object groups available. Object groups have an ID that identifies them uniquely within a track.
+An object MUST be decodable after the delivery of all prior objects within the same group.
+This implies that the first object within a group MUST NOT depend on other objects.
```
afrind, line 221: I don't think you need "This implies that".

draft-lcurley-warp.md
```diff
@@ -214,10 +214,21 @@ to span more than one useable unit may create more than one viable application
...
+The application MAY use the sequence number to perform ordering and detect gaps within a group.
+
+A sender SHOULD selectively transmit objects based on group membership.
+For example, the sender could transmit only the newest group to new subscriber, as the objects are then guaranteed to be decodable.
+However, the sender MUST still transmit objects based on the priority {{send-order}}, not based on the group sequence number.
```
afrind, line 226: I'm a bit confused by this part.  Maybe leave it out for now since it doesn't seem purely object model related - the priority stuff can go in the priority section?

draft-lcurley-warp.md
```diff
@@ -214,10 +214,21 @@ to span more than one useable unit may create more than one viable application
...
+
+A decoder MAY use objects from prior groups when using a self-correcting encoding scheme.
```
afrind, line 228: I'm trying to reconcile "MUST NOT depend" with "MAY use".  It sounds like there's some encoding schemes that have "weak" rather than strict dependencies.  I would maybe put this MAY up near line 220/221 at least.

afrind (commented):
Individual review, mostly editorial comments.

draft-lcurley-warp.md
suhasHere: +1 to @afrind  suggestion. This doesn't seem to fit in the data model section and the SHOULD here is probably  making it a suggestion.

draft-lcurley-warp.md
suhasHere: I don't see how this helps with the data model and also may be we should replace the decoder here with "application" to be generic

draft-lcurley-warp.md
```diff
@@ -653,12 +664,12 @@ OBJECT Message {
 The track identifier obtained as part of subscription and/or publish control message exchanges.
 
 * Group Sequence :
-An integer always starts at 0 and increases sequentially at the original media publisher.
-Group sequences are scoped under a Track.
+The object is a member of the indicated group {{model-group}} within the track.
+The sequence starts at 0, increasing sequentially for each group within the track.
```
suhasHere, line 669: I think we still need " is starts at 0 at the original media publisher".  A subscriber joining mid-way through a call will be delivered object message with group_sequence > 0.

draft-lcurley-warp.md
```diff
@@ -214,10 +214,21 @@ to span more than one useable unit may create more than one viable application
 mapping from media to wire format, which could be confusing for protocol users.
 
 ## Groups {#model-group}
+A *group* is an independent sequence of objects.
+Each group is a join point, such that a new subscriber can begin receiving a track without prior progress.
 
-An object group is a sequence of media objects. Beginning of an object group can be used as a point at which the receiver can start consuming a track without having any other object groups available. Object groups have an ID that identifies them uniquely within a track.
+An object MUST be decodable after the delivery of all prior objects within the same group.
```
suhasHere, line 220: I don't think "MUST be decodable after the delivery of all prior objects" is true. Decoders can decode a p-frame even if its proir p-frame is missing, but the experience for the user will not be great. 

May be somethign on these lines ?

1) Objects within a group express dependency relationship and in certain application settings an object MAY NOT be decodable until all the prior objects are delivered.

2) "Also, the first object within a group MUST NOT depend on other objects within a group"

suhasHere (changes requested):
Thanks for the PR and few high level comments

At a higher level, it might be useful to capture group properties as
- Groups are consumption points for enabling functions like joins, rewinds and replays
- Groups are checkpoints for nodes implementing relay functionality to implement appropriate congestion responses
- Some text about object dependencies within a group

afrind (commented):
Individual/Editorial comments

draft-lcurley-warp.md
suhasHere: my concern is the requirement of delivery .. MAY NOT was suggested to say that it is true in some cases not all the cases.

afrind: I think we're all saying the same thing.  Maybe it's easier to read this normative requirement if it were in the active rather than passive voice:

An endpoint MUST NOT add an object to a group if that object strictly depends on an object in another group, or another object in the same group with a higher sequence number.

?

suhasHere: that is not true entirely .. when there are B frames, there will be dependency in both directions .. 

These details are very specific to media encoding and these might totally change when a new codec come.

I would rather keep it at an higher level and suggest something on the lines of " objects within a group express dependency relationship and the first object must be the independent object."

We can then add few example for video coding , but those are just examples and not normative requirements.

afrind: Good point about B frames - let me make sure I understand:

The dependency graph is like this:  P1 <- B -> P2 (and also maybe P1 <- P2?)

And an app might want to put B and P2 in the same group, but transmit B first because it's available first, right?

I was thinking that making stricter normative guarantees about group properties would make publisher, relay and receiver operation easier, but with graphs like these maybe there isn't anything we can definitively say.

When an encoder spits out a B frame, does it know how many subsequent frames it's going to depend on?  If so, the publisher could send P1(seq=0), B(seq=2), P2(seq=1), but maybe that opens up another can of worms.

suhasHere: An application knows exactly based on the encoding scheme its dependency structure and it can use techniques to order the objects that satisfy the relationship. I feel we should probably not define more about these specifics in this spec

kixelated: @suhasHere media should always* be transmitted in decode order (DTS), not presentation order (PTS).

The object sequence is basically the same thing as DTS but sequential. A B-frame would have a higher object sequence than any frames it references.

I think the sequential nature is useful because a relay can detect gaps. A relay can guarantee that an object is decodable if all prior sequences have been delivered, without access to the underlying frame references.

draft-lcurley-warp.md
kixelated: I think this mostly points out a flaw with the concept of groups. They can't actually provide hard dependency guarantees.

Personally, I think we should position groups as the addressable units, while objects are merely chunks within a group.

draft-lcurley-warp.md
kixelated: I see what you mean. Is there a better way to phrase "at the original media publisher"? Could we just remove this "starts at zero" altogether?

kixelated:
> * Groups are consumption points for enabling functions like joins, rewinds and replays
> * Groups are checkpoints for nodes implementing relay functionality to implement appropriate congestion responses

I'm having trouble explaining WHY groups can be used as sync points or checkpoints. What are the concrete properties of groups that enable these use-cases?

draft-lcurley-warp.md
suhasHere: Objects are sequenced when inserted into the group, the delivery can have gaps (due to drops) or out of sequence .. But the application knows to put them in order. 

May be we need to say "Nodes SHOULD always attempt to deliver objects within the group as defined by its object sequence, wherever possible"

suhasHere:
> > * Groups are consumption points for enabling functions like joins, rewinds and replays
> > * Groups are checkpoints for nodes implementing relay functionality to implement appropriate congestion responses
> 
> I'm having trouble explaining WHY groups can be used as sync points or checkpoints. What are the concrete properties of groups that enable these use-cases?

Consumption points - subscriber joins, rewind and replays
Checkpoints - For relays to apply congestion response at group boundaries

draft-lcurley-warp.md
suhasHere: may be say it as "group sequence start at 0 however subscribers should be able to consume objects starting at any group sequence" ??

draft-lcurley-warp.md
kixelated: > I don't think "MUST be decodable after the delivery of all prior objects" is true. Decoders can decode a p-frame even if its proir p-frame is missing, but the experience for the user will not be great.

If all prior frames have been delivered, then yeah, the P-frame MUST be decodable. But like Alan said, I was careful to not say anything about the opposite case. I agree that we can't say that "an object MUST NOT be decodable without all prior frames".

kpugin: > The object sequence is basically the same thing as DTS but sequential. A B-frame would have a higher object sequence than any frames it references.

hm... DTS of B-frame would be lower than DTS of P-frame it depends on - am I messing up something? In that case how does sequence number of B frame can be higher than all frames it depends on?

draft-lcurley-warp.md
fluffy: I think we should just remove lines 225 to 226 and move to the section on how senders decide what to send next.

draft-lcurley-warp.md
fluffy: I propose we resolve this by removing the word "prior"

draft-lcurley-warp.md
suhasHere: i think the latest commit didn't fix this yet ?

draft-lcurley-warp.md
```diff
@@ -214,10 +214,15 @@ to span more than one useable unit may create more than one viable application
...
 ## Groups {#model-group}
+A *group* is collection of objects, part of a larger track ({{model-track}}).
 
-An object group is a sequence of media objects. Beginning of an object group can be used as a point at which the receiver can start consuming a track without having any other object groups available. Object groups have an ID that identifies them uniquely within a track.
+A group behaves as a join point for subscriptions.
+A new subscriber may not want to receive the entire track, and will instead opt to receive only the latest group(s).
+The sender then selectively transmits objects based on their group membership.
+
+The application is responsible for how objects are placed into groups.
+In general, objects within a group SHOULD NOT depend on objects in other groups.
```
suhasHere, line 224: I am inclining that we just skip this sentence altogether ?

draft-lcurley-warp.md
```diff
@@ -214,10 +214,15 @@ to span more than one useable unit may create more than one viable application
 mapping from media to wire format, which could be confusing for protocol users.
 
 ## Groups {#model-group}
+A *group* is collection of objects, part of a larger track ({{model-track}}).
 
-An object group is a sequence of media objects. Beginning of an object group can be used as a point at which the receiver can start consuming a track without having any other object groups available. Object groups have an ID that identifies them uniquely within a track.
+A group behaves as a join point for subscriptions.
```
suhasHere, line 219: May be also add a point for relays to make congestion response decisions.

draft-lcurley-warp.md
fluffy: I would be happy to see this merged either way.

fluffy (approved):

draft-lcurley-warp.md
kixelated: Oh yeah I forgot about this section.

vasilvv (approved):

suhasHere (approved):

PR #148, opened 2023-04-26
Resolve object DISCUSS in object model

kixelated:
Removes a discuss in the object model, going with option 1.

I do think we should add an atomic/incremental flag to OBJECT, so the relay can know when dropping an object early will cause an object to be full undecodable. But that's a separate discussion.

### DIFF
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -185,21 +185,17 @@ x (b):
 
 ## Objects {#model-object}
 
-The basic element of Warp is an *object*. An object is a single addressable
-unit whose payload is a sequence of bytes.  An object MAY depend on other
-objects: such objects are members of the same group {{model-group}}. Objects carry
-associated metadata such as priority, TTL or other information usable by a relay,
-but relays MUST treat object payloads as opaque.
+The basic element of Warp is an *object*.
+An object is an addressable unit whose payload is a sequence of bytes.
+All objects belong to a group, indicating ordering and potential dependencies. {{model-group}}
+Objects carry associated metadata such as priority, TTL, or other information usable by a relay, but relays MUST treat the object payload as opaque.
 
 The application is solely responsible for the contents of objects.
 This includes the underlying encoding, compression, any end-to-end encryption, or authentication.
 A relay MUST NOT combine, split, or otherwise modify object payloads.
 
-The application MAY be able to partially decode an object.
-A sender SHOULD transmit incomplete objects to avoid incurring additional latency.
-A sender MAY drop or deprioritize the remainder of an object during congestion,
-potentially rendering the object undecodable in favor of a higher priority object.
-
+An endpoint SHOULD transmit objects based on their priority.
+An object MAY arrive out of the intended order or dropped altogether.
 
 ## Groups {#model-group}
 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -187,7 +187,7 @@ x (b):
 
 The basic element of Warp is an *object*. An object is a single addressable
 unit whose payload is a sequence of bytes.  An object MAY depend on other
-objects to be decoded: such objects are members of the same group {{model-group}}. Objects carry
+objects: such objects are members of the same group {{model-group}}. Objects carry
 associated metadata such as priority, TTL or other information usable by a relay,
 but relays MUST treat object payloads as opaque.
 
--- a/draft-lcurley-warp.md
+++ b/draft-lcurley-warp.md
@@ -186,32 +186,20 @@ x (b):
 ## Objects {#model-object}
 
 The basic element of Warp is an *object*. An object is a single addressable
-cacheable unit whose payload is a sequence of bytes.  An object MAY depend on other 
-objects to be decoded. All objects belong to a group {{model-group}}. Objects carry 
-associated metadata such as priority, TTL or other information usable by a relay, 
+unit whose payload is a sequence of bytes.  An object MAY depend on other
+objects to be decoded: such objects are members of the same group {{model-group}}. Objects carry
+associated metadata such as priority, TTL or other information usable by a relay,
 but relays MUST treat object payloads as opaque.
 
-DISCUSS: Can an object be partially decodable by an endpoint?
+The application is solely responsible for the contents of objects.
+This includes the underlying encoding, compression, any end-to-end encryption, or authentication.
+A relay MUST NOT combine, split, or otherwise modify object payloads.
 
-Authors agree that an object is always partially *forwardable* by a relay but
-disagree on whether a partial object can be used by a receiving endpoint.
+The application MAY be able to partially decode an object.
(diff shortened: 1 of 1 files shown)


draft-lcurley-warp.md
```diff
@@ -186,32 +186,20 @@ x (b):
...
 but relays MUST treat object payloads as opaque.
 
-DISCUSS: Can an object be partially decodable by an endpoint?
+The application is solely responsible for the contents of objects.
+This includes the underlying encoding, compression, any end-to-end encryption, or authentication.
+A relay MUST NOT combine, split, or otherwise modify object payloads.
 
-Authors agree that an object is always partially *forwardable* by a relay but
-disagree on whether a partial object can be used by a receiving endpoint.
+The application MAY be able to partially decode an object.
```
afrind, line 198: Maybe 'decode a partially received object' ?

draft-lcurley-warp.md
```diff
@@ -186,32 +186,20 @@ x (b):
...
+A sender SHOULD transmit incomplete objects to avoid incurring additional latency.
```
afrind, line 199: A sender SHOULD transmit partial object data as soon as possible to avoid incurring additional latency.

draft-lcurley-warp.md
```diff
@@ -186,32 +186,20 @@ x (b):
...
+A sender MAY drop or deprioritize the remainder of an object during congestion,
```
afrind, line 200: The meaning of 'remainder' is unclear - remainder after what?  Maybe:

A sender MAY drop or deprioritize object data during congestion, potentially rendering an object undecodable in favor of a higher priority object.

afrind (commented):
Editorial suggestions

draft-lcurley-warp.md
```diff
@@ -186,32 +186,20 @@ x (b):
 ## Objects {#model-object}
 
 The basic element of Warp is an *object*. An object is a single addressable
-cacheable unit whose payload is a sequence of bytes.  An object MAY depend on other 
-objects to be decoded. All objects belong to a group {{model-group}}. Objects carry 
-associated metadata such as priority, TTL or other information usable by a relay, 
+unit whose payload is a sequence of bytes.  An object MAY depend on other
+objects: such objects are members of the same group {{model-group}}. Objects carry
```
afrind, line 190: We lost the statement 'All objects belong to a group' - which I think was providing some value.  Maybe put it first like:

All objects belong to a group.  An object MAY depend on other objects in the same group.

suhasHere: +1 to @afrind  suggestion on ``` All objects belong to a group. An object MAY depend on other objects in the same group.```

Also , I think we still need it to be cacheable unit too,  since it will give Relays unit of caching as objects and can support object queries to be served from cache

draft-lcurley-warp.md
```diff
@@ -186,32 +186,20 @@ x (b):
...
+associated metadata such as priority, TTL or other information usable by a relay,
 but relays MUST treat object payloads as opaque.
 
-DISCUSS: Can an object be partially decodable by an endpoint?
+The application is solely responsible for the contents of objects.
+This includes the underlying encoding, compression, any end-to-end encryption, or authentication.
+A relay MUST NOT combine, split, or otherwise modify object payloads.
```
suhasHere, line 196: +1 on "MUST NOT combine, split, or otherwise modify object payloads"

draft-lcurley-warp.md
suhasHere: per our 1:1 chat at IETF116, we agreed to not mention anything about partially decodable property. I would prefer removing this line altogether. I think it is application way to define object granularity ( GOP vs single frame, for example) and they know if they are partially decodable or not.

draft-lcurley-warp.md
suhasHere: Again, I will remove this sentence and not useful for the data model section and also "partial" is not well defined.  

How about we have something on these lines 
"Latency sensitive applications using moq-transport MAY choose not to wait for the object to be fully constructed before transmitting. The mechanisms of how the parts of the objects are identified at the receiver is defined by the application"

draft-lcurley-warp.md
suhasHere: I am not sure this level of detail is needed in data model section

suhasHere (changes requested):

afrind (commented):
Individual review/editorial

draft-lcurley-warp.md
suhasHere: in my intent above, i was thinking of just the end-points and hence used the word Applications.  I agree we can add a section on relays above too.

kixelated: > This would work for me too, although the statement is largely aimed at relays, rather than endpoints. I'm guessing Luke was trying to be generic with "sender" but maybe spelling out the relay bit is also useful, like

I want to make sure that MoQ libraries support streaming objects too. The intention is to signal that generic libraries and endpoints should not buffer objects.

bad: `send_object(id: int, payload: bytes)`
good: `start_object(id: int); write(chunk: bytes); write(chunk: bytes)`

> in my intent above, i was thinking of just the end-points and hence used the word Applications.

I only think of the "application" as the first and last hop. A generic relay has no idea that an application is "latency sensitive", and can't change its behavior in response.


> I will remove this sentence and not useful for the data model section and also "partial" is not well defined.

I want OBJECTs to behave like QUIC streams, where they could be used for both incremental (partially buffered) and atomic (fully buffered) purposes. I think it's important to clarify that somehow.

draft-lcurley-warp.md
kixelated: I think it's important to mention that an object might not be delivered, without going into detail. Although we can remove stuff about partial objects because those fall into the same "might not be (fully) delivered" bucket.

draft-lcurley-warp.md
kixelated: > Also , I think we still need it to be cacheable unit too, since it will give Relays unit of caching as objects and can support object queries to be served from cache

I just don't know what "cacheable" means. I think it's fair to say that an object is addressable and that TTL is an example property. But otherwise, I don't think anything is being added by including that word.

draft-lcurley-warp.md
```diff
@@ -185,33 +185,17 @@ x (b):
...
-Partial decodability could be another property of an object.
-
-Option 2: A receiver MUST NOT start decoding an object before it has completely arrived
-
-Objects could be end-to-end encrypted and the receiver might not be able to
-decrypt or authenticate an object until it is fully present.  Allowing Objects
-to span more than one useable unit may create more than one viable application
-mapping from media to wire format, which could be confusing for protocol users.
+An endpoint SHOULD transmit objects based on their priority.
+An object MAY arrive out of order or dropped altogether.
```
afrind, line 198: or *be* dropped

suhasHere (approved):

fluffy (approved):

PR #162, opened 2023-05-23
Update the section on track naming

### LINKED ISSUE #160: Proposal: clarifying the semantics of the track names
(background: #70, probably more)

At the end of the last meeting, I've volunteered to write up a proposal that would let us make progress on the track name semantics.  I propose the following (in the form of individual numbered points, so that they are easier to comment on):

1. (Observation) In order to subscribe to a track, the client in the current design needs at least three distinct pieces of information: _a connect URL_, _a track namespace_ and _a track name_.
1. (Should be in the spec) The _connect URL_ tells the client how to establish a network connection.
1. (Already in the spec) The _track name_ and the _track namespace_ are the two strings sent in the SUBSCRIBE request; together they are referred to as the _full track name_.
1. (New definition) A single _MoQ server_ is an entity that is uniquely identified by a connect URL (that is to say, if two entities have different URLs, we consider them different servers).
1. (New definition) Let us introduce a new term, _MoQ scope_.  A MoQ scope is defined to be a set of servers for which requesting the same full track name will result in the client subscribing to the same track with the same data on it.
1. (Already in the spec, but clarified) It is up to the application building on top of MoQ to define how large the scope has to be.  An application that deals with connections between devices on a local network may limit the scope to a single connection.  An application that uses multiple CDNs to serve media may require the scope to be all of those CDNs.
1. (New rule) The full track name is the only piece of information that a specific MoQ server uses to identify the track. It does not use any sort of header-based negotiation like HTTP provides with Vary/Accept/etc.  The connect URL is not used directly, but it does imply the scope for the name.
1. (Corollary) A full track name can be used as a complete cache key (for a given scope).

(Open problem: while a full track name might identify the track at a specific instance of time, at a later point in time that name could refer to a completely different track.  We should decide how we want to deal with that, but probably in a different issue)

@fluffy @suhasHere @kixelated What do you think about this proposal? Does it address your concerns?

vasilvv:
Introduce the notion of MoQ scope.
Require that the full track name is the sole name for the track.

Closes #160

### DIFF
--- a/draft-lcurley-moq-transport.md
+++ b/draft-lcurley-moq-transport.md
@@ -147,15 +147,28 @@ In general, objects within a group SHOULD NOT depend on objects in other groups.
 A *track* is a sequence of objects ({{model-object}}) organized into groups ({{model-group}}).
 A subscriber can request individual tracks at group boundaries, including any new objects while the track is active.
 
-### Full Track Name {#track-fn}
+### Track Naming and Scopes {#track-fn}
 
-Tracks are identified by a globally unique identifier, called "Full Track Name" and defined as shown below:
+In MoQTransport, every track has a *track name* and a *track namespace* associated with it.
+If two or more different tracks belong to the same track namespace, they have to be announced atomically
+(see {{message-announce}}). A track name identifies an individual track within the namespace.
+
+A tuple of a track name and a track namespace together is known as *a full track name*:
 
 ~~~~~~~~~~~~~~~
 Full Track Name = Track Namespace  "/"  Track Name
 ~~~~~~~~~~~~~~~
 
-This document does not define the exact mechanism of naming Track Namespaces. Applications building on top of MoQ MUST ensure that the mechanism used guarantees global uniqueness; for instance, an application could use domain names as part of track namespaces. Track Namespace is followed by the application context specific Track Name, encoded as an opaque string.
+A *MoQ scope* is a set of MoQ servers (as identified by their connection URIs) for which
+subscribing to the same full track name would result in the subscriber receiving the same track with the same data on it.
+It is up to the application building on top of MoQ to define how broad or narrow the scope has to be.
+An application that deals with connections between devices on a local network may limit the scope to a single connection;
+by contrast, an application that uses multiple CDNs to serve media may require the scope to include all of those CDNs.
+
+The full track name is the only piece of information that is used to identify the track within a given scope.
+MoQTransport does not provide any in-band content negotiation methods similar to the ones defined by HTTP
+({{?RFC9110, Section 10}}); if, at a given moment in time, two tracks within the same scope contain different data,
+they have to have different full track names. This makes full track name suitable for use a key for caching.
 
 ~~~
 Example: 1


draft-lcurley-moq-transport.md
```diff
@@ -147,15 +147,28 @@ In general, objects within a group SHOULD NOT depend on objects in other groups.
...
+(see {{message-announce}}). A track name identifies an individual track within the namespace.
+
+A tuple of a track name and a track namespace together is known as *a full track name*:
 
 ~~~~~~~~~~~~~~~
 Full Track Name = Track Namespace  "/"  Track Name
 ~~~~~~~~~~~~~~~
 
-This document does not define the exact mechanism of naming Track Namespaces. Applications building on top of MoQ MUST ensure that the mechanism used guarantees global uniqueness; for instance, an application could use domain names as part of track namespaces. Track Namespace is followed by the application context specific Track Name, encoded as an opaque string.
+A *MoQ scope* is a set of MoQ servers (as identified by their connection URIs) for which
```
afrind, line 162: To clarify, all MoQ servers with the same connection URI share a single scope.  Some MoQ servers with different connection URIs MAY share the same scope. Is that right?  Can a MoQ server with a single connection URI belong to more than one scope?

draft-lcurley-moq-transport.md
```diff
@@ -147,15 +147,28 @@ In general, objects within a group SHOULD NOT depend on objects in other groups.
...
+subscribing to the same full track name would result in the subscriber receiving the same track with the same data on it.
+It is up to the application building on top of MoQ to define how broad or narrow the scope has to be.
+An application that deals with connections between devices on a local network may limit the scope to a single connection;
+by contrast, an application that uses multiple CDNs to serve media may require the scope to include all of those CDNs.
+
+The full track name is the only piece of information that is used to identify the track within a given scope.
+MoQTransport does not provide any in-band content negotiation methods similar to the ones defined by HTTP
+({{?RFC9110, Section 10}}); if, at a given moment in time, two tracks within the same scope contain different data,
+they have to have different full track names. This makes full track name suitable for use a key for caching.
```
afrind, line 171: have to have -> MUST have ?

draft-lcurley-moq-transport.md
```diff
@@ -147,15 +147,28 @@ In general, objects within a group SHOULD NOT depend on objects in other groups.
 A *track* is a sequence of objects ({{model-object}}) organized into groups ({{model-group}}).
 A subscriber can request individual tracks at group boundaries, including any new objects while the track is active.
 
-### Full Track Name {#track-fn}
+### Track Naming and Scopes {#track-fn}
 
-Tracks are identified by a globally unique identifier, called "Full Track Name" and defined as shown below:
+In MoQTransport, every track has a *track name* and a *track namespace* associated with it.
+If two or more different tracks belong to the same track namespace, they have to be announced atomically
```
afrind, line 153: ANNOUNCE only announces a namespace (though is that a problem?  Can you announce a single track?), therefore I don't understand what it means that tracks to have to be announced atomically -- that's the only way to do it, right?

afrind (commented):
Please coordinate with Will to minimize merge hassle with the forthcoming editorial PR(s).

draft-lcurley-moq-transport.md
suhasHere: I don't think we even need to say the above.  How about

The full track name is the only piece of information that is used to identify the track within a given scope and  used as  a key for caching.

draft-lcurley-moq-transport.md
suhasHere: i don't think we need to mention about ANNOUNCE in the object model section ?

draft-lcurley-moq-transport.md
suhasHere: s/connection URIs/connection URLs/ ?

suhasHere: my interpretation of moq-scope was   **"Set of application servers and relays for which there is uniqueness guarantee for a  given full track name."**

Then, the following also holds (has some slight modifications from the PR )

"Within a moq-scoppe, subscribing to the same full track name would result in the subscriber receiving media for the same track. It is up to the application building on top of MoQ to define how broad or narrow the scope has to be. An application that deals with connections between devices on a local network may limit the scope to a single connection (identified via the connection url); by contrast, an application that uses multiple CDNs to serve media may require the scope to include all of those CDNs.

suhasHere (changes requested):
Thanks Victor for the PR. I have few suggestions and happy to discuss live if needed. please let me know .

draft-lcurley-moq-transport.md
suhasHere: editoria , s/scope/moq-scope

draft-lcurley-moq-transport.md
```diff
@@ -147,15 +147,28 @@ In general, objects within a group SHOULD NOT depend on objects in other groups.
...
+
+A tuple of a track name and a track namespace together is known as *a full track name*:
 
 ~~~~~~~~~~~~~~~
 Full Track Name = Track Namespace  "/"  Track Name
 ~~~~~~~~~~~~~~~
 
-This document does not define the exact mechanism of naming Track Namespaces. Applications building on top of MoQ MUST ensure that the mechanism used guarantees global uniqueness; for instance, an application could use domain names as part of track namespaces. Track Namespace is followed by the application context specific Track Name, encoded as an opaque string.
+A *MoQ scope* is a set of MoQ servers (as identified by their connection URIs) for which
+subscribing to the same full track name would result in the subscriber receiving the same track with the same data on it.
```
kpugin, line 163: 'full track name' here means namespace + track name?

draft-lcurley-moq-transport.md
```diff
@@ -147,15 +147,28 @@ In general, objects within a group SHOULD NOT depend on objects in other groups.
...
 
-Tracks are identified by a globally unique identifier, called "Full Track Name" and defined as shown below:
+In MoQTransport, every track has a *track name* and a *track namespace* associated with it.
+If two or more different tracks belong to the same track namespace, they have to be announced atomically
+(see {{message-announce}}). A track name identifies an individual track within the namespace.
+
+A tuple of a track name and a track namespace together is known as *a full track name*:
 
 ~~~~~~~~~~~~~~~
 Full Track Name = Track Namespace  "/"  Track Name
```
fluffy, line 159: I would rather remove the "/" and just have Track Namespace and Track Name be binary blogs that are concatenated.

draft-lcurley-moq-transport.md
vasilvv: Changed sentence to discuss uniqueness guarantee.

draft-lcurley-moq-transport.md
vasilvv: Removed this sentence.

draft-lcurley-moq-transport.md
vasilvv: Yes, see the sentence above ("A tuple of a track name and a track namespace together is known as *a full track name*")

draft-lcurley-moq-transport.md
vasilvv: I agree that we don't need it, but I feel like this might be easier to understand what this requirement means if we explain it in terms of what we are not going to do.

kixelated (approved):
looks gud

suhasHere (approved):

fluffy (approved):

suhasHere (approved):

PR #170, opened 2023-05-25
rewrite motivation

fluffy:


draft-lcurley-moq-transport.md
```diff
@@ -74,49 +74,67 @@ It was originally developed for live media, but has been generalized for similar
...
+receiver feedback, reassembly, and more. One goal of MoQTransport is to
+achieve the best of both these worlds: leverage the features of QUIC to
+create a simple yet flexible low latency protocol that can rapidly
+detect and respond to congestion.
+
+#### Leveraging QUIC
+
+Applying {{QUIC}} to HAS via HTTP/3 does not yield generalized
+improvements in throughput. One reason for this is that sending segments
+down a single QUIC stream still allows head-of-line blocking to
```
suhasHere, line 105: This is not entirely true , isn't it. Each GET/POST is its own QUIC Stream or am I understanding it incorrectly ?

suhasHere (commented):
Have a clarification question on HTTP/3 , but rest looks fine

draft-lcurley-moq-transport.md
```diff
@@ -74,49 +74,67 @@ It was originally developed for live media, but has been generalized for similar
...
-Other similar live content would then be distributed other yet additional protocols: for example updates, chat, metadata, etc.
+#### Latency
 
-This draft attempts to build a unified base transport protocol for media and similar use-cases.
-Any live content can be fragmented into objects and annotated to achieve the intended behavior.
-The goal is not to reinvent how content is encoded, just delivered.
+HTTP Adaptive Streaming (HAS) has been successful at achieving scale
+although often at the cost of latency. Latency is necessary to correct
+for variable network throughput. Ideally live content is consumed at the
+same rate it is produced. End-to-end latency would be fixed and only
```
fluffy, line 86: s/rate/time/

fluffy (commented):
I really like the new text

draft-lcurley-moq-transport.md
```diff
@@ -74,49 +74,67 @@ It was originally developed for live media, but has been generalized for similar
...
+Applying {{QUIC}} to HAS via HTTP/3 does not yield generalized
+improvements in throughput. One reason for this is that sending segments
+down a single QUIC stream still allows head-of-line blocking to
+occur. Only by leveraging the parallel nature of QUIC streams can
+improved throughput be achieved in the face of loss. A goal of MoQT is
+to design a streaming protocol to leverage the transmission benefits
+afforded by parallel QUIC streams as well exercising options for
+flexible loss recovery.
+
+#### Workflow efficiency
```
kixelated, line 112: I like "Universal" more. There's a multitude of benefits for consolidating standards, workflow efficiency being one of them.

draft-lcurley-moq-transport.md
kixelated: The implicit assumption is that HLS/DASH do not issue concurrent HTTP requests, so they can't fully benefit from QUIC.

draft-lcurley-moq-transport.md
```diff
@@ -74,49 +74,67 @@ It was originally developed for live media, but has been generalized for similar
...
+overall latency. TCP-based protocols are simple but are slow to detect
+congestion and suffer from head-of-line blocking. UDP-based protocols
+can avoid queuing, but the application is now responsible for the
+complexity of fragmentation, congestion control, retransmissions,
+receiver feedback, reassembly, and more. One goal of MoQTransport is to
+achieve the best of both these worlds: leverage the features of QUIC to
+create a simple yet flexible low latency protocol that can rapidly
+detect and respond to congestion.
+
+#### Leveraging QUIC
```
kixelated, line 101: I understand the point trying to be made, but it's a bit convoluted and not particularly compelling.

I would focus on _why_ we want to use QUIC, instead of focusing on how we use QUIC better than HLS/DASH. We can build a thin layer on top of the QUIC ecosystem and API, leveraging the full feature set.

kixelated (commented):
I'm biased but I thought this motivation section was pretty good.

draft-lcurley-moq-transport.md
suhasHere: that makes sense. But that might change in the future  and adding it here is  bit misleading.

May be , how about we say

"As it stands today, applying {{QUIC}} to HAS via HTTP/3 does not yield generalized ....."

draft-lcurley-moq-transport.md
suhasHere: I think I am with Luke on this .. Universal or Unified works for me

draft-lcurley-moq-transport.md
afrind: I think rate was intended here - there's an acknowledgement that e2e latency and encoding delays are baked in.

draft-lcurley-moq-transport.md
afrind: s/does not yield/has not yet yielded/ ?

afrind (commented):
lgtm with the small tweaks to H3 and section heading requested by Luke/Suhas

suhasHere (approved):

PR #171, opened 2023-05-25
add new terminology

fluffy:


draft-lcurley-moq-transport.md
```diff
@@ -140,6 +140,18 @@ Consumer:
 
 : A QUIC endpoint receiving media over the network. This could be a consumer or a relay.
 
+Endpoint:
+
+: The original producer or the final consumer in a transmission chain.
```
suhasHere, line 145: Suggest split this into 2 terms

Producer : The original producer of the media.
Consumer: the final consumer of the media.

draft-lcurley-moq-transport.md
```diff
@@ -140,6 +140,18 @@ Consumer:
...
 : A QUIC endpoint receiving media over the network. This could be a consumer or a relay.
 
+Endpoint:
+
+: The original producer or the final consumer in a transmission chain. 
+
+Group:
+
+: A temporal sequence of objects. A group represents a subscription point in a track. The media data within the group must not depend on the data in any other groups. 
+
```
suhasHere, line 150: remove last sentence and add reference to the group ?

fluffy: I thought we had come to "SHOULD NOT" depend.

draft-lcurley-moq-transport.md
```diff
@@ -140,6 +140,18 @@ Consumer:
 
 : A QUIC endpoint receiving media over the network. This could be a consumer or a relay.
 
+Endpoint:
+
+: The original producer or the final consumer in a transmission chain. 
+
+Group:
+
+: A temporal sequence of objects. A group represents a subscription point in a track. The media data within the group must not depend on the data in any other groups.
```
fluffy, line 149: I have a slight preference for "join point" instead of "subscription point"

draft-lcurley-moq-transport.md
kixelated: This is does not match my understand of "endpoint". A QUIC endpoint is a client or a server.

draft-lcurley-moq-transport.md
kixelated: IMO keep this definition simple and link to the section in the object model. No need to describe all properties of a group up here.

> The media data within the group must not depend on the data in any other groups.

We agreed to remove any mention of dependencies when updating the object/group model. Technically, audio groups depend on each other, as do video groups when intra-refresh is used.

kixelated (changes requested):

draft-lcurley-moq-transport.md
suhasHere: I think at one point, we decided to not update terminology section for -05 and update it once we discuss the entities and roles.

could we just revert these changes? @kixelated @fluffy would that work ?

suhasHere: Added an OPEN ISSUE on the terminology

suhasHere:
@kixelated does the updated text with OPEN ISSUE looks about right to you. I personally feel we should revisit all the terms and get to a common understaning across the documents. For -05 an OPEN ISSUE seems like OK. thoughts ?

draft-lcurley-moq-transport.md
afrind: I had the same comment as Luke on 163.  Endpoint appears 10 times in the doc though and all the references match Luke's definition (either the QUIC client or the QUIC server for a session) as I read it.

The point of the terminology is to help understand the document, so this definition is not accomplishing that.  I'd say either update the definition to match the usage, or remove the definition.  Adding this definition plus a discuss doesn't seem optimal to me.

kixelated (approved):

PR #172, opened 2023-05-25
rewrite model section

fluffy:


draft-lcurley-moq-transport.md
```diff
@@ -167,36 +167,32 @@ x (b):
...
-The basic element of MoQTransport is an *object*.
+The basic data element of MoQTransport is an *object*.
 An object is an addressable unit whose payload is a sequence of bytes.
 All objects belong to a group, indicating ordering and potential dependencies. {{model-group}}
-Objects carry associated metadata such as priority, TTL, or other information usable by a relay, but relays MUST treat the object payload as opaque.
-
-The application is solely responsible for the contents of objects.
-This includes the underlying encoding, compression, any end-to-end encryption, or authentication.
-A relay MUST NOT combine, split, or otherwise modify object payloads.
+Objects are comprised of two parts: metadata and a payload.  The metadata is visible to relays but the payload is only visible to the producer and consumer. The application is solely responsible for the content of the object payload.
```
fluffy, line 177: I would rather not loose priority and TTL in metadata here.

draft-lcurley-moq-transport.md
```diff
@@ -167,36 +167,32 @@ x (b):
...
 ## Objects {#model-object}
 
-The basic element of MoQTransport is an *object*.
+The basic data element of MoQTransport is an *object*.
 An object is an addressable unit whose payload is a sequence of bytes.
 All objects belong to a group, indicating ordering and potential dependencies. {{model-group}}
-Objects carry associated metadata such as priority, TTL, or other information usable by a relay, but relays MUST treat the object payload as opaque.
-
-The application is solely responsible for the contents of objects.
-This includes the underlying encoding, compression, any end-to-end encryption, or authentication.
```
kixelated, line 180: Could we mention that the application is responsible for any further sub divisions?

draft-lcurley-moq-transport.md
```diff
@@ -167,36 +167,32 @@ x (b):
 : Indicates that x consists of a variable length integer, followed by that many bytes of binary data.
 
 
-# Model
+# Object Model {#model}
```
kixelated, line 170: What about working top-down right than bottom up?

I think it makes more sense to describe tracks first, and then groups, and then objects, and then anything smaller (application).

kixelated (commented):
We spent a while debating the object/group text so I'd prefer smaller changes.

draft-lcurley-moq-transport.md
suhasHere: totally agree .. This was my suggestion on google docs too . Its more natural that way.

afrind: I also had the same suggestion, though I think reading bottom up is still getting the job done.  If we're pressed for time today for -05, I would leave this and revisit it later.

kixelated (approved):

PR #173, opened 2023-05-25
add model intro paragraph

fluffy:


draft-lcurley-moq-transport.md
```diff
@@ -169,6 +169,13 @@ x (b):
 
 # Model
 
+MoQT is a transport that moves entitites called messages
```
suhasHere, line 172: s/MoQT/moqt ?

suhasHere (approved):
editorial nit if it can be addressed, otherwise LGTM

draft-lcurley-moq-transport.md
kixelated: I like "MoQ Transport". If we're going to use initials, it should probably be all uppercase (ex. MOQT)

draft-lcurley-moq-transport.md
```diff
@@ -169,6 +169,13 @@ x (b):
...
+{{message}}. Messages are divided into two classes: control messages and
```
kixelated, line 173: In the current draft, OBJECTs are messages. There's no distinction between control streams/messages and data streams/messages.

IMO the object model is meant to explain the properties of the protocol, while messages are the wire encoding of those properties. We should cover messages in another section and leave this section high level.

kixelated (changes requested):
Let's introduce the concept of control streams and data streams in another PR. But those should have nothing to do with the object model.

fluffy (commented):
+1 other things people want to fix but LGTM

suhasHere:
@kixelated  please check the latest commit . thanks

afrind:
Something is wrong, this PR appears blank

kixelated (approved):

PR #163, opened 2023-05-24
Editorial changes

wilaw:
The purpose of this PR is to improve the readability, language and structure of the current draft, and to incorporate editorial  changes suggested by https://docs.google.com/document/d/1MAWoSJq7YM7AZ6y6TJqNGRcKwlhWlVOOfHxt8R5jwzQ .

PR #174, opened 2023-05-25
add examples

fluffy:


draft-lcurley-moq-transport.md
```diff
@@ -234,6 +234,21 @@ Track Namespace = security-camera.example.com/camera1
...
+Example: 5
+Track Namespace = "" (empty string)
+Track Name = audio
+Full Track Name = audio
+
+Example: 6
+Track Namespace = 12345
+Track Name = 6789
+Full Track Name = 123456789
+
```
suhasHere, line 251: With the new change to remove "/", i think these examples can now go away

suhasHere (changes requested):

draft-lcurley-moq-transport.md
kixelated: Yeah I would remove all of these examples. It's not a difficult concept that the full track name is split into two parts.

suhasHere:
Reflecting on the authors call and latest changes in #163  , I will go ahead and close this

PR #175, opened 2023-05-25
add new catalog track section

fluffy:


fluffy (commented):
I don't think we should merge this yet

hardie:
Chair hat:
Given the desire to get a document out, what will it take to get a
disposition on this?  Can this be merged if a "To be discussed" is added?

Not chair chat:

This text:

All producers of content MUST produce a catalog track which describes
the availability of the tracks which they are capable of publishing.

seems like an odd way to deploy the MUST, because "capable of publishing"
seems like it could be divorced from the timing of the act of publication.
I think you mean "Producers MUST provide a catalog of the tracks available
when they publish the tracks."

Producers MUST update the catalog tracks when new tracks are added or
existing tracks deleted.  The payload of the catalog track objects is
defined by the MoQStreaming Format being used by the publisher, with the
exception that the first varint of the payload MUST be unencrypted and
MUST hold the type of the MoQStreaming Format, as registered in the IANA
registration table {{iana}} for MoQtransport Streaming Formats.

I also think the last sentence is quite hard to read. The intent appears to
be something like "The catalog track must begin with an unencrypted
indication of which MoQStreaming Format is in use; all subsequent objects
are provided in that format.  The IANA registrations of streaming formats
{{iana}} will associate a value with each registered format which may be
used to provide the indicator above."

PR #176, opened 2023-05-25
remove connection URL section

fluffy:


suhasHere:
We agreed to put this section back in the Author's call today. May be we don't need this change

suhasHere (changes requested):
Close this PR, per authors call outcome

PR #177, opened 2023-05-25
remove session and objects

fluffy:


suhasHere (approved):

fluffy (commented):
I do think we should remove all this for now but wanted to at least flag it for reiview.

draft-lcurley-moq-transport.md
```diff
@@ -241,17 +241,7 @@ Full Track Name = security-camera.example.com/camera1/hd-video
...
-A transport session is established for each track bundle.
-The client issues a CONNECT request with a URL which the server uses for identification and authentication.
-All control messages and prioritization occur within the context of a single transport session, which means a single track bundle.
-When WebTransport is used, multiple transport sessions may be pooled over a single QUIC connection for efficiency.
-
-
-# Objects
-MoQTransport works by transferring objects over QUIC streams.
-The application determines how live content is fragmented into tracks, groups, and objects.
+# Sessions {#session}
```
kixelated, line 244: What is a session? Why do we start with a DISCUSS object priorities? What does this even have to do with sessions...?

kixelated (changes requested):
I don't understand this change.

draft-lcurley-moq-transport.md
fluffy: Two things happened here. First was we kept the Session and moved it to a top level session ( the ## to # ).  And 2nd things was to remove all the stuff on the left. I think the further text that was left here on Order Priorities and such get moved/removed in a different PR about priorities. It confusing to see here but it will be killed later

afrind (commented):
This PR by itself doesn't make a lot of sense but I'm ok rolling with it if it's part of the grand plan

kixelated (approved):

PR #178, opened 2023-05-25
remove groups

fluffy:


suhasHere (approved):

kixelated (approved):

fluffy (commented):
LGTM

huitema (commented):
Yes, we can remove this. We will certainly revisit the issue later, because this is a thorny subject. For example, we assume that "objects MUST not depend from objects in the group that have a higher priority code or a higher object ID" -- object at P2 does not depend on another object at P3, object number 1000 cannot depend on object number 2001. That much has consensus, but it is a negative statement.

Issue #138, opened 2023-04-19
Proposal: Control and data streams


kixelated:
### Current
The client creates a bidirectional stream and sends the SETUP message. The server replies with its own SETUP message on the same stream. Afterwards, both endpoints can create any number of unidirectional streams containing any non-SETUP messages.

### Problems
- #105 Messages with different priorities should not be mixed in the same stream, which means control messages should never be mixed with data messages.
- #136 Messages can't be sent or have to be buffered until the receipt of the SETUP message.
- #137 Control messages over separate streams arrive in an unknown order.

### Proposed
- The client creates a single bidirectional control stream. The control stream MUST start with a SETUP message, followed by any number of control messages (ex. SUBSCRIBE).
- The sender creates unidirectional data streams. The data streams MUST only contain OBJECT messages.

**Pros**:
- Partially fixes #105. Control messages are no longer intermingled with with data messages, and the control stream MUST always be the highest priority.
- Fixes #136. Endpoints no longer have to buffer messages, as QUIC will automatically make sure SETUP arrives before any other control messages. Data streams MUST NOT be created until receipt of SETUP.
- Fixes #137. Control messages arrive in a specified, deterministic order.

**Cons**:
- Head-of-line blocking between control messages, which sometimes may not be desired.

afrind:
> The sender creates unidirectional data streams

While I was reading the requirements draft, I found some use cases mentioning bidirectionality.  Is it expected/supported that there can be data flowing in both directions (eg: a p2p call)?

kixelated:
> > The sender creates unidirectional data streams
> 
> While I was reading the requirements draft, I found some use cases mentioning bidirectionality. Is it expected/supported that there can be data flowing in both directions (eg: a p2p call)?

Yeah, my intent is that there's a "sender" and a "receiver" per track. The ROLE param in the SETUP message indicates if an endpoint (client/server) intends to send tracks, receive tracks, or both.

In the bidirectional case, both endpoints would say ROLE=BOTH. Then they could independently PUBLISH or SUBSCRIBE to tracks.

Alternatively, we could say that tracks may be bidirectional... but I don't think that makes any sense.

wilaw:
We can simply say that data objects MUST NOT be sent over control streams, and remove any reference to whether data streams should be unidirectional or bidirectional. It doesn't matter and we can leave it to the end-points discretion.

suhasHere:
The proposal above aligns pretty well with the idea in #96 

[...]

PR #179, opened 2023-05-25
rewrite stream initialization

fluffy:


suhasHere (approved):

kixelated (approved):
Looks gud. I would like to see #138 eventually because it's unclear when an endpoint is allowed to create "subsequent streams".

fluffy (commented):
LGTM

PR #197, opened 2023-05-26
Rename to Media Over QUIC Transport (MOQT)

suhasHere:


draft-lcurley-moq-transport.md
```diff
@@ -856,9 +853,9 @@ TODO: fill out currently missing registries:
 
 TODO: register the URI scheme and the ALPN
 
-TODO: the MoQTransport spec should establish the IANA registration table
-for MoQtransport Streaming Formats. Each MoQTransport streaming format
-can then register its type in that table. The MoQT Streaming Format type
+TODO: the MOQT spec should establish the IANA registration table
+for MOQT Streaming Formats. Each MOQT streaming format
```
wilaw, line 857: The table name should not use the abbreviation. It should be `for MoQTransport Streaming Formats. Each MoQTransport Streaming Format can then register its type in that table. The MoQTransprt Streaming Format type` ..

draft-lcurley-moq-transport.md
```diff
@@ -43,17 +43,14 @@ author:
...
 
 informative:
-  NewReno: RFC6582
-  BBR: I-D.cardwell-iccrg-bbr-congestion-control-02
 
 --- abstract
 
-This document defines the core behavior for MoQTransport, a media
-transport protocol over QUIC. MoQTransport allows a producer of media
+This document defines the core behavior for Media Over QUIC Transport(MOQT), a media
```
wilaw, line 52: Needs a space between Transport and (

draft-lcurley-moq-transport.md
afrind: These are called "MoQ Streaming Formats" when referenced earlier, and I'm inclined to keep that in both places.

suhasHere: we decided to go with Media Over QUIC Transport or MOQT. I can go with the latter for the above.

kixelated (commented):
I think `MOQT` is an ugly acronym.

My recommendation:
long: `Media over QUIC Transport`
short: `MoQ Transport`

It's fine if the short name is more verbose.

kixelated:
But I also don't care; we're going to rename this thing anyway.

draft-lcurley-moq-transport.md
```diff
@@ -241,7 +238,7 @@ active.
 
 ### Track Naming and Scopes {#track-name}
 
-In MoQTransport, every track has a *track name* and a *track namespace* associated with it.
+In MOQT, every track has a *track name* and a *track namespace* associated with it.
```
afrind, line 241: The below paragraph has 5 uses of "MoQ" by itself:

MoQ scope (3x) -> MOQT Scope
MoQ servers -> just "servers"? Or "MOQT servers"?
"on top of MoQ" -> on top of MOQT / using MOQT

kixelated: I'd be fine calling it `MoQ Transport` and leaving stuff like `MoQ Scope` and `MoQ Servers`

suhasHere:
@afrind  the latest commit should address your feedback

kixelated (approved):

kixelated:
I don't like MOQT but we can continue to argue about the name forever.

afrind:
This LGTM for implementing what we discussed on the call.  Let's consider it a consensus placeholder while we search for that perfect name.

fluffy (approved):
LGTM

PR #199, opened 2023-05-26
fix up whitespace newlines for some tool and add a period

fluffy:


suhasHere (approved):
LGTM, as it seems to just whitespaces only change.

fluffy:
No real changes here (other than adding a missing period) but changes newlines for checker tools

PR #200, opened 2023-05-26
remove italics which generates invalid txt files with kramdown

fluffy:
The stuff we had to italicize the stuff in html is inserting underscores in the .txt file so I just want to remove this until for now until we figure this out. Did not change anything other than making the words not italicized


suhasHere:
Thanks for catching this , the rendered TXT file looks different for these things. Merging it for now and we can work on figuring out the right markdown syntax.

suhasHere (approved):

PR #201, opened 2023-05-26
reflow document

fluffy:
I have no idea if people want to merge this or not. This does not change a single character of the draft but does reflow all the markdown. Some people like this, some people hate this. 

I like it because github only really does line by line diffs and this minimizes the stuff on a single line thus makes diffs smaller and easier to deal with and less likely to get a merge conflict..  I think it works well with a github work flow as long as wen you edit stuff, you do not reflow and try and just only change the line you need to change. Over time things get out of wack and eventually you need to have a reflow PR like this. 

If we are going to do a reflow, now is an ideal time because we do not have a bunch of outstanding PRs.

### DIFF
--- a/draft-lcurley-moq-transport.md
+++ b/draft-lcurley-moq-transport.md
@@ -49,34 +49,37 @@ informative:
 
 --- abstract
 
-This document defines the core behavior for Media over QUIC Transport (MOQT), a media
-transport protocol over QUIC. MOQT allows a producer of media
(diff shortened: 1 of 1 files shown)


suhasHere (approved):
This seems fine to me.  The changes look daunting but they are just moving the text around for wrapping

afrind:
In general I'm a fan of this - what is the max line length you used?

kixelated (approved):

PR #232, opened 2023-08-14
Specify draft versions.

### LINKED ISSUE #214: Draft version numbers
It's time to specify some version numbers so we can attempt interop.

We should do something like QUIC, where draft versions are prefixed with 0xff (or something?).

kixelated:
Based on QUIC: https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport-27#name-versions

Fixes #214

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -680,8 +680,13 @@ Server SETUP Message Payload {
 ~~~
 {: #moq-transport-setup-format title="MOQT SETUP Message"}
 
-The Parameter Value Length field indicates the length of the Parameter
-Value.
+The available versions and SETUP parameters are detailed in the next sections.
+
+### Versions {#setup-versions}
+
+MoQ Transport versions are a 32-bit unsigned integer, encoded as a varint.
+This version of the specification is identified by the number 0x00000001.
+Versions with the most significant 16 bits of the version number cleared are reserved for use in future IETF consensus documents.
 
 The client offers the list of the protocol versions it supports; the
 server MUST reply with one of the versions offered by the client. If the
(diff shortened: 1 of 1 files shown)


ianswett (approved):

fluffy (approved):
Looks fine to me. Sure we might do more complicated stuff later but this seems like good enough for now.

suhasHere (approved):
I would want to consider using alpn at some point in time. 
Want to keep it open for major.minor version at some point too.

This PR is fine to continue the experimentation.

fluffy:
Based on call today, sounds like when running over QUIC we should also use an ALPN of "moq-13" for draft 13

ianswett (changes requested):
Per conversation, the intent is to use ALPN in raw QUIC to negotiate protocol and version.

ie: moq-<draftnum> as the ALPN string.

The intent is to work with the WebTransport working group on how to achieve similar functionality in WebTransport.

kixelated:
The conclusion was that we should move version negotiation out of the SETUP message and into an ALPN for native QUIC and an unspecified mechanism for WebTransport.

However, I can't think of any replacement for WebTransport that would work today. Query parameters would work for passing the client supported versions but there's no mechanism for replying with the server selected version. The javascript API provides no information about connection establishment (ex. any returned headers), only that it completed. If we moved version negotiation out of SETUP, then it would not be possible to interop until new functionality was specifying in WebTransport and deployed.

Could we go ahead with this PR until there's an available alternative? I like the direction of utilizing more of the QUIC/WebTransport handshake, but I don't think we should block interop awaiting this marginally better design.

ianswett:
Can we add in the ALPN text as part of this PR and for now say something like "When QUIC is used, the negotiated alpn MUST match the value in the SETUP message."?

PR #234, opened 2023-08-16
Fix Track Request Params section ordering and missing info in announce_ok

suhasHere:
This PR moves track request params to its own section. This was messed up during huge PR merge as part of preparing for IETF117 work. 

Also it add track namespace length field to Announce_OK message which was omitted.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -882,37 +882,6 @@ ANNOUNCE Message {
 * Track Request Parameters: The parameters are defined in
 {{track-req-params}}.
 
-### Track Request Parameters {#track-req-params}
-
-The Track Request Parameters identify properties of the track requested
-in either the ANNOUNCE or SUSBCRIBE REQUEST control messages. The peers
-MUST close the connection if there are duplicates. The Parameter Value
-Length field indicates the length of the Parameter Value.
-
-
-#### GROUP SEQUENCE Parameter
-
-The GROUP SEQUENCE parameter (key 0x00) identifies the group within the
-track to start delivering objects. The publisher MUST start delivering
-the objects from the most recent group, when this parameter is
-omitted. This parameter is applicable in SUBSCRIBE REQUEST message.
-
-#### OBJECT SEQUENCE Parameter
-
-The OBJECT SEQUENCE parameter (key 0x01) identifies the object with the
-track to start delivering objects. The `GROUP SEQUENCE` parameter MUST
-be set to identify the group under which to start delivery. The
-publisher MUST start delivering from the beginning of the selected group
-when this parameter is omitted. This parameter is applicable in
-SUBSCRIBE REQUEST message.
-
-#### AUTHORIZATION INFO Parameter
-
-AUTHORIZATION INFO parameter (key 0x02) identifies track's authorization
-information. This parameter is populated for cases where the
-authorization is required at the track level. This parameter is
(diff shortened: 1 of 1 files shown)


vasilvv (approved):
You may also want to move the actual definition of Track Request Parameter, it's currently in the SUBSCRIBE REQUEST section.

suhasHere:
@vasilvv good catch. updated the commit to address your comment. Thanks

draft-ietf-moq-transport.md
```diff
@@ -982,6 +946,48 @@ The client:
...
+~~~
+{: #moq-track-request-param format title="MOQT Track Request Parameter"}
+
+
+### GROUP SEQUENCE Parameter
+
+The GROUP SEQUENCE parameter (key 0x00) identifies the group within the
+track to start delivering objects. The publisher MUST start delivering
+the objects from the most recent group, when this parameter is
+omitted. This parameter is applicable in SUBSCRIBE REQUEST message.
```
kixelated, line 973: > This parameter is applicable in SUBSCRIBE REQUEST message.

It's quite difficult to cross-reference which parameters are valid for which messages. We only have 3 parameters now and 2 messages so it's not a big deal now, but even just scanning these parameter descriptions to determine which parameters are valid in which messages is time consuming. It also produces a poor narrative, as the behavior of SUBSCRIBE REQUEST is described in this section. 

There should be a single way to encode optional parameters specified in the encoding section which would also be used by SETUP. Then have each message declare the available parameters and their functionality instead of using this shared section.

draft-ietf-moq-transport.md
```diff
@@ -921,6 +884,7 @@ successful authorization and acceptance of an ANNOUNCE message.
 ~~~
 ANNOUNCE OK
 {
+  Track Namespace Length(i),
   Track Namespace
```
kixelated, line 888: `(..)`

Also, I added the `(b)` encoding to avoid the need explicitly put the size in front of each byte slice. We should use it or remove it.

```
Track Namespace (b)
```

Is equivalent to:
```
Track Namespace Length (i)
Track Namespace (..)
```

kixelated (approved):
It's an improvement so we should merge, but we do need to clean up how parameters are encoded.

draft-ietf-moq-transport.md
```diff
@@ -982,6 +946,48 @@ The client:
   processing objects from both in parallel.
 
 
+## Track Request Parameters {#track-req-params}
+
+The Track Request Parameters identify properties of the track requested
+in either the ANNOUNCE or SUSBCRIBE REQUEST control messages. The peers
+MUST close the connection if there are duplicates. The Parameter Value
```
ianswett, line 953: Should this be close the session?

draft-ietf-moq-transport.md
```diff
@@ -982,6 +946,48 @@ The client:
...
+Length field indicates the length of the Parameter Value.
+
+The format of `Track Request Parameter` is as follows:
+
+~~~
+Track Request Parameter {
+  Track Request Parameter Key (i),
```
ianswett, line 960: Can we use 'Type' instead of key, since that's more standard?

draft-ietf-moq-transport.md
suhasHere: IIRC, Someone from the quic community thought b(...) was confusing and QUIC's defined way is clear enough .. Hence most of this was reverted at some point ..

I am open to either approach. Let's pick one

Issue #145, opened 2023-04-24
Receiving unknown OBJECT messages (who creates the track-id)


kixelated:
#141 adds a `Track ID` field. This is used to compress a long string (in SUBSCRIBE) into a short varint (in OBJECT).

The problem is that the mapping from string to varint is indicated in the SUBSCRIBE_OK message, which is sent in parallel with OBJECT messages. This introduces a race condition, where packet loss will cause SUBSCRIBE_OK to arrive late. The receiver will be unable to associate OBJECT messages with tracks until the arrival of the SUBSCRIBE_OK.

A correct implementation must buffer any OBJECT messages with unknown track IDs for an indefinite amount of time. Any other implementation will error or skip media due to decode errors. This also introduce a risk of resource exhaustion, for example if the stream containing SUBSCRIBE_OK is reset.

kixelated:
My proposal is to move the `Track ID` to the SUBSCRIBE message. The OBJECT sender MUST use a `Track ID` in a received SUBSCRIBE message. This means the receiver will never receive an unknown or unrequested OBJECT.

kixelated:
I know @wilaw wants the sender to be able to choose the `Track ID`, but that makes this race condition unavoidable. The only way to prevent it is to operate like QPACK; a literal track name is used until the SUBSCRIBE_OK is somehow acknowledged.

The argument is that the sender should choose the Track ID so it could use the same ID across multiple connections. There's a claim that this would improve edge efficiency, but I am extremely doubtful. My time spent profiling QUIC stacks makes me believe that per-hop properties (ex. QUIC connection ID, QUIC steam ID, HTTP/3 QPACK, WebTransport session ID) contribute an insignificant amount to CPU usage.

suhasHere:
At a high level, all the nodes needs to be able to handle unknown objects - its true regardless of trackId issue.

A typical way is to we define the following in the text
- send full track-names + trackId multiple times for the initial X time units over a separate message
- drop objects after Y time units 

We need some text on handling such cases as recommendations from security perspective too.

afrind:
As an Individual:

In many ways this ID is to moq what the QUIC stream ID is to HTTP/3 - and for those, there is a simple, unilateral, and deterministic way to choose.  Maybe that is a good model to follow?  Do we want to allow re-use of Track IDs for different tracks over the life of a connection?

 > A correct implementation must buffer any OBJECT messages with unknown track IDs for an indefinite amount of time. 
 
[...]

PR #235, opened 2023-08-16
Add support for Unsubscribe message

suhasHere:
This PR adds support for subscribers to issue unsubscribe requests to express its interest in not receiving media for a given track. This reflects discussions from #203 

The high level call flow is

```mermaid
sequenceDiagram
  Bob->>Relay: SUBSCRIBE( webex.com/alice/hd-video)
  Bob->>Relay: SUBSCRIBE( webex.com/alice/audio)
  loop media
        Relay-->Bob: Media Exchange (audio, hd-video)
   end
   Note left of Bob: Not interested in hd-video anymore
   Bob->>Relay: UNSUBSCRIBE(webex.com/alice/hd-video)
   Relay->>Bob: UNSUBSCRIBE OK (webex.com/alice/hd-video)
   loop media
        Relay-->Bob: Media Exchange continues unaltered for audio
   end
```

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -639,19 +639,15 @@ continues until the end of the stream.
 |-------|--------------------------------------------------|
 | 0x5   | SUBSCRIBE ERROR ({{message-subscribe-error}})    |
 |-------|--------------------------------------------------|
-| 0x6   | UNSUBSCRIBE REQUEST ({{message-unsubscribe-req}})|
+| 0x6   | ANNOUNCE  ({{message-announce}})                 |
 |-------|--------------------------------------------------|
-| 0x7   | UNSUBSCRIBE OK ({{message-unsubscribe-ok}})      |
+| 0x7   | ANNOUNCE OK ({{message-announce-ok}})            |
 |-------|--------------------------------------------------|
-| 0x8   | UNSUBSCRIBE ERROR ({{message-unsubscribe-error}})|
+| 0x8   | ANNOUNCE ERROR ({{message-announce-error}})      |
 |-------|--------------------------------------------------|
-| 0x9   | ANNOUNCE  ({{message-announce}})                 |
+| 0x9   | UNSUBSCRIBE ({{message-unsubscribe}})            |
 |-------|--------------------------------------------------|
-| 0x10  | ANNOUNCE OK ({{message-announce-ok}})            |
-|-------|--------------------------------------------------|
-| 0x11  | ANNOUNCE ERROR ({{message-announce-error}})      |
-|-------|--------------------------------------------------|
-| 0x12  | GOAWAY ({{message-goaway}})                      |
+| 0x10  | GOAWAY ({{message-goaway}})                      |
 |-------|--------------------------------------------------|
 
 ## SETUP {#message-setup}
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -626,27 +626,33 @@ The Message Length field contains the length of the Message Payload
...
+|-------|--------------------------------------------------|
+| 0x1   | SETUP ({{message-setup}})                        |
+|-------|--------------------------------------------------|
+| 0x3   | SUBSCRIBE REQUEST ({{message-subscribe-req}})    |
+|-------|--------------------------------------------------|
+| 0x4   | SUBSCRIBE OK ({{message-subscribe-ok}})          |
+|-------|--------------------------------------------------|
+| 0x5   | SUBSCRIBE ERROR ({{message-subscribe-error}})    |
+|-------|--------------------------------------------------|
+| 0x6   | UNSUBSCRIBE REQUEST ({{message-unsubscribe-req}})|
```
afrind, line 642: Renumbering will cause interop problems, but I agree that by the final rev we should have the messages in some reasonable ordering.  We could

1) Keep it like this, and assume there's not much interop happening yet, so this is a freebie.
2) keep all the numbers the same for now, add UNSUBSCRIBE and friends at the end, then renumber the messages later when version negotiation is better settled

draft-ietf-moq-transport.md
```diff
@@ -860,6 +866,72 @@ this response is provided.
...
+## UNSUBSCRIBE REQUEST {#message-unsubscribe-req}
+
+A subscriber issues a `UNSUBSCRIBE REQUEST` message to a publisher indicating it is no longer interested in receiving media for the specified track.
+
+The format of `UNSUBSCRIBE REQUEST` is as follows:
+
+~~~
+UNSUBSCRIBE REQUEST Message {
+  Full Track Name Length (i),
+  Full Track Name (...),
```
afrind, line 878: You can also do this with the track ID, since you can only unsubscribe things that had an ID.  I might leave FTN in UNSUBSCRIBE_OK/ERROR though.

afrind (commented):
Individual review

draft-ietf-moq-transport.md
suhasHere: @afrind  you are right. The same thought went through my mind too. Then it felt , if we are changing it is good to change before our interop meeting :-) 

I am fine with either options too

draft-ietf-moq-transport.md
```diff
@@ -626,27 +626,33 @@ The Message Length field contains the length of the Message Payload
...
+|-------|--------------------------------------------------|
+| 0x4   | SUBSCRIBE OK ({{message-subscribe-ok}})          |
+|-------|--------------------------------------------------|
+| 0x5   | SUBSCRIBE ERROR ({{message-subscribe-error}})    |
+|-------|--------------------------------------------------|
+| 0x6   | UNSUBSCRIBE REQUEST ({{message-unsubscribe-req}})|
+|-------|--------------------------------------------------|
+| 0x7   | UNSUBSCRIBE OK ({{message-unsubscribe-ok}})      |
+|-------|--------------------------------------------------|
+| 0x8   | UNSUBSCRIBE ERROR ({{message-unsubscribe-error}})|
```
kixelated, line 646: I don't think we need this message. Why would an UNSUBSCRIBE fail?

### The producer already terminated the subscription
This means the producer already sent a SUBSCRIBE_ERROR/SUBSCRIBE_RESET, so sending an UNSUBSCRIBE_ERROR is just redundant.

### The subscription does not exist
Which means the corresponding SUBSCRIBE was not received first. I think this is a protocol error and the connection can be closed instead. 

### Because I said so?
I don't want to support a producer to say: "no, I'm still going to keep sending this junk". An UNSUBSCRIBE should always succeed.

draft-ietf-moq-transport.md
```diff
@@ -626,27 +626,33 @@ The Message Length field contains the length of the Message Payload
...
+|-------|--------------------------------------------------|
+| 0x3   | SUBSCRIBE REQUEST ({{message-subscribe-req}})    |
+|-------|--------------------------------------------------|
+| 0x4   | SUBSCRIBE OK ({{message-subscribe-ok}})          |
+|-------|--------------------------------------------------|
+| 0x5   | SUBSCRIBE ERROR ({{message-subscribe-error}})    |
+|-------|--------------------------------------------------|
+| 0x6   | UNSUBSCRIBE REQUEST ({{message-unsubscribe-req}})|
+|-------|--------------------------------------------------|
+| 0x7   | UNSUBSCRIBE OK ({{message-unsubscribe-ok}})      |
```
kixelated, line 644: I don't think we need this message. I would have the producer reply with a SUBSCRIBE_RESET instead, mirroring how QUIC streams are reset. 

**SUBSCRIBE_RESET**: The subscription has ended and no further objects will be transmitted above the provided group/object sequence. Mirrors RESET_STREAM.

**UNSUBSCRIBE**: Please terminate the subscription (at a given group/object sequence?). Implicitly acknowledged by the receipt of a SUBSCRIBE_RESET. Mirrors STOP_SENDING.

Note: I think SUBSCRIBE_ERROR and SUBSCRIBE_RESET are the same message, but I'm splitting them here for sake of argument.

draft-ietf-moq-transport.md
kixelated: Using the track ID isn't great in the current spec because you can't UNSUBSCRIBE prior to receiving a SUBSCRIBE_OK. 

However, that's fixed once the subscriber chooses the ID (#145), in which case all SUBSCRIBE messages should reference that ID instead of the full track name. There's no reason to use the full track name unless you're sending these control messages over separate streams, which will introduce a boat-load of race conditions.

draft-ietf-moq-transport.md
```diff
@@ -911,7 +983,7 @@ SUBSCRIBE REQUEST message.
 AUTHORIZATION INFO parameter (key 0x02) identifies track's authorization
 information. This parameter is populated for cases where the
 authorization is required at the track level. This parameter is
-applicable in SUBSCRIBE REQUEST and ANNOUNCE messages.
+applicable in SUBSCRIBE REQUEST, UNSUBSCRIBE and ANNOUNCE messages.
```
kixelated, line 986: Why do we need auth information in an UNSUBSCRIBE message? This seems like an invalid use-case.

kixelated (changes requested):

vasilvv (approved):
We'll need to revisit using track name vs track ID for unsubscribe once we have an answer to #145, but for now, this LGTM.

kixelated (approved):
It might be useful to have an error code, but I'm not sure. LGTM.

kixelated:
Thinking about it some more, an error code doesn't seem necessary. 

publisher -> subscriber should have an error code.
subscriber -> publisher doesn't need an error code.

Even with an error code, it would get lost in the deduplication process.

Issue #248, opened 2023-09-11
Zero-length messages are foot guns


martinduke:
A zero-length message length indicates that the message runs to the end of the stream. It is also possible to have actual zero length messages (in the spec, currently just GOAWAY).

Zero-length messages can only exist at the end of the stream; otherwise, their encoding is ambiguous. (This is fine for GOAWAY, but not in general).

There are several ways to design out of this, which are pretty obvious. We ought to pick one.

kixelated:
Absolutely, I agree.

I want to send each GoP over a QUIC stream. The problem is that the group/object abstraction doesn't let you do that (#244). The closest you can get is to send the entire GoP as an unbounded object, because then a relay is unable to coalesce multiple (unbounded) objects into the same stream.

I think we choose:
- A stream per OBJECT with an unspecified length. QUIC will report the final size via FIN.
- A stream per OBJECT, broken into sized DATA frames like HTTP/3. This allows intermixing messages on the same stream but is transparent to the application.
- A stream per GROUP, broken into sized OBJECT frames. Object boundaries are exposed to the application.

I'm not a fan of the third option. I want to take my 2s media segment and send it over MoQ without parsing/fragmenting at frame boundaries.

martinduke:
My parser would be MUCH simpler if the "end of stream" magic number was UINT62_MAX instead of zero. Of course, this is all inter-related with both @kixelated's concerns and the question as to if we need message length at all except for OBJECT frames.

kixelated:
UINT62_MAX is also a footgun, because you really should throw UnexpectedEOF if a stream FIN truncates a message.

Control messages:
- Don't need a length if we depend on versioning/extensions. ex. QUIC frames.
- Need a length if you want to ignore unknown messages. ex. HTTP/3 frames

Data messages:
- Don't need a length if they run until the end of the stream. ex. one OBJECT per QUIC stream
- Need a length if there's multiple messages per stream. ex. HTTP/3 DATA frames

I'm fine with an explicit length for all messages, but only if group/object abstraction has a tighter coupling with QUIC streams. The only way to transmit a byte stream right now using MoQ is to make a single OBJECT with size=0.

kixelated:
Although I want to avoid implicit sizes like #240. If the last field in a message is a variable size (ex. string), then it still needs to encode its size instead of running until the end of the message. It wastes bytes but that's how TLS works, and it means you don't need separate encoders/decoders for implicit sizes.

[...]

PR #255, opened 2023-09-28
Disallow unknown messages

### LINKED ISSUE #241: Actions on unknown Message Type
Fail, or pass over?

martinduke:
Fixes #241 and #248.

if messages must be known, explicit message lengths are redundanct except for OBJECT and messages with parameters.

This also eliminates a whole class of potential errors where the message length does not match the actual parsing of the message. Not checking this -- and not have any message type be potentially zero-length -- greatly simplifies the parser.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -613,18 +613,16 @@ limits.  See section 2.2 in {{QUIC}}.
 Both unidirectional and bidirectional QUIC streams contain sequences of
 length-delimited messages.
 
+An endpoint that receives an unknown message type MUST close the connection.
+
+
 ~~~
 MOQT Message {
   Message Type (i),
-  Message Length (i),
   Message Payload (..),
 }
 ~~~
 {: #moq-transport-message-format title="MOQT Message"}
-
-The Message Length field contains the length of the Message Payload
-field in bytes.  A length of 0 indicates the message is unbounded and
-continues until the end of the stream.
 
 |-------|--------------------------------------------------|
 | ID    | Messages                                         |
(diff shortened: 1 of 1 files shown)


kixelated (approved):
I love it.

It also means you can implement streaming encoders. Prefixing each message with the length required a temporary buffer or 2nd pass to compute the length.

kixelated:
And trying to use custom messages without negotiating an extension/version just won't work for MoQ. We're designing a protocol with multiple hops so you do need relay cooperation, or at the very least some information on how to forward unknown messages.

kixelated:
Also I love the fact that you're using "parameter count" instead of "parameter size (bytes)". It makes it much easier to build an encoder/decoder.

fluffy:
I thought we agreed to allow extensibility by allowing new message types to be created.

fluffy:
Like to see us agree on a extensibility strategy before deciding what to do on this. 

We also need to decide if messages can have optional fields

kixelated:
> I thought we agreed to allow extensibility by allowing new message types to be created.

We have version negotiation, and presumably we'll have extension negotiation too. What's the use-case for adding new messages without negotiating them? 

The problem is that we can't relay unknown messages, so all it takes is a single uncooperative relay on the chain to shallow them. I think we need explicit negotiation as a result.

fluffy:
If we have extension negotiation, I agree.  But if our extension negotiation mechanism relies on not getting an error on unknown messages .... 

PR is wrong place to discuss this but I think what I would like to see is a good way to do extension negotiation, and then like you have here, error on any message that is not part of a negotiated extension.  I think I would lean that way if we had an extension negotiation mechanism.

Note this is the opposite direction of HTTP which is often our model for things. HTTP just skips past unknown headers.

suhasHere (approved):

ianswett (approved):
LG.

The agreement at the interim was to follow up with separate frame types for object with and without length, but do that in a separate PR.

fluffy (approved):
LGTM

Issue #251, opened 2023-09-26
What's the point of mandatory SETUP parameters?


martinduke:
Parameters are a useful tool for optional fields in a message.

But AFAICT, none of the SETUP parameters are actually optional. Clients MUST send ROLE.  #151 and #250 discuss the server ROLE situation, but ISTM the server ROLE will be either required or forbidden.

Meanwhile, PATH will be required for raw QUIC clients and forbidden for all other cases.

So why not make these part of the SETUP message and save ourselves all the parameter encoding overhead? It's true that it would require different formats for Webtrans client, Raw QUIC client, and server, but we've already split SETUP into client and server variants.

I would not be opposed to leaving parameters in SETUP with an empty registry, as it would be a good extension point for capability negotiation.

kixelated:
Yeah, it's quite frustrating right now. Especially the server SETUP, because it supports generic parameters but receipt of any registered kinds should be an error.

However the messages do need to be generic so they can be used to perform version negotiation. I wish we could punt version negotiation to another layer like ALPN but no equivalent exists in WebTransport.

martinduke:
Hmm, interesting point about VN. I did a PR, so the group can decide.

vasilvv:
> So why not make these part of the SETUP message and save ourselves all the parameter encoding overhead?

TLS used to do that, and as a result, like half of TLS client hello in 1.3 are unused legacy fields with fixed values.  Though I guess if we defer version negotiation to ALPN, this is less of an issue.

martinduke:
It might be worthwhile to sketch out exactly what VN looks like. Is the idea here that the client will include parameters for every version for which it advertises support? Adn then the server responds with the subset of parameters for the chosen version?

martinduke:
People who know me from QUIC will know that I am not a fan of negotiation via ALPN.

ianswett:
We now have some text on negotiation.  I think if we saw a PR for this, it could be landed, especially now that SETUP has been split into two in #297

martinduke:
The VN text answers the question of what SETUP parameters are for. Closing.

Issue #250, opened 2023-09-12
server-sent ROLE parameters


martinduke:
Sec 6.1.1.1 specifies client sending or ROLE, and says nothing of server-sent.

I guess server-sent ROLE is not needed, but maybe we should fail if it is sent?

kixelated:
I filed #151 about the role parameter. The client doesn't actually know what the server intends to do. The server should be able to reply with its functionality, which might be a subset of what the client can support.

Issue #151, opened 2023-05-04
ROLE doesn't support negotiation


kixelated:
The SETUP message contains a ROLE parameter. In the current draft, the client specifies the value, indicating if the client will send data (0x1), if the server will send data (0x2), or if both endpoints will send data (0x3).

A conferencing client supports the ability to send and receive media, and would always send role=0x3 because it supports both functionalities. However, it's unaware of the server functionality, and the server has no way of informing the client if it intends to send media. This matters because a generic client needs to know if it shows a black screen and/or should send a SUBSCRIBE message (ex. for /catalog).

I propose that both endpoints send the ROLE parameter: 0x1 = publisher, 0x2 = consumer, 0x3 = both. This gives the server the ability to specify its own functionality.

suhasHere:
I am not very sure why do we even need Role in the first place .. can it be changed mid-session ?

Say the use-case
- Participant in media conference joins as subscriber only since he is driving and want to listen in
- Then at some point mid session , he wants to contribute and he now becomes as publisher and subscriber

what is the expected flow in this case ?

kpugin:
Current draft doesn't really explain what ROLE parameter is for, only that client needs to send it - what was original idea? Why not just always allow everyone to send and receive data?

kixelated:
I think ROLE is good for negotiating half-functionality.

- If the remote endpoint cannot send media (ex. player), then sending a SUBSCRIBE is invalid.
- If the remote endpoint cannot receive media (ex. OBS), then sending a ANNOUNCE/OBJECT is invalid.

@suhasHere brings up a good point. If an endpoint _could_ switch roles, then it should advertise full functionality. That would include publishing an empty catalog so it can append it later if it wishes to actually publish tracks.

fluffy:
I don't think I seem much use for the Role - it seems like something the application would know at a higher layer. But I have no objection to including it if others do see a need for it. It's not like it makes life harder for applications that don't use it.

VMatrix1900:
If the ROLE is used to enforce the half-functionality, then it needs to be clear what should be done when invalid message is received. Sending a GOAWAY? Sending a generic error message? Does the connection need to be terminated? I think most of the time, application will just choose the full functionality in case of role switching on the fly and avoid the invalid response code(just ignore the invalid message).

kixelated:
Oh hey, I found a good use-case for ROLE.

[...]

PR #256, opened 2023-09-28
Parameter Reform

### LINKED ISSUE #247: MUST close connection on duplicate parameters
Does this apply for duplicate unknown parameters?

If so, it makes for annoying bookkeeping. One has to keep an unbounded data structure of received parameter types,  instead of a handful of bools for the known parameter types.

martinduke:
Fixes #247, #251

- Removes "mandatory parameters" from the SETUP message and just puts them in the main part of the message (retaining the notion of parameters as an extension point)

- Unifies the two parameter namespaces into one (previously known as "Track Request Parameters")

- Resolves the inconsistency about receiver reaction to duplicate parameters as a "SHOULD" close, not MUST. (Tracking the appearance of unknown parameter types is relatively resource-intensive)

- Specifies that parameters that arrive in messages where they don't apply are ignored, without error.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -880,7 +880,7 @@ SUBSCRIBE REQUEST Message {
 
 * Full Track Name: Identifies the track as defined in ({{track-name}}).
 
-* Parameters: As defined in {{version-specific params}}. All are optional
+* Parameters: As defined in {{version-specific-params}}. All are optional
 but valid for this message.
 
 On successful subscription, the publisher SHOULD start delivering
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -674,12 +674,15 @@ SETUP Parameter {
 Client SETUP Message Payload {
   Number of Supported Versions (i),
   Supported Version (i) ...,
-  SETUP Parameters (..) ...,
+  Role (i),
+  [Length of Path (i)],
+  [Path (..) ....],
+  Parameters (..) ...,
 }
 
 Server SETUP Message Payload {
   Selected Version (i),
-  SETUP Parameters (..) ...,
+  Parameters (..) ...,
 }
 ~~~
 {: #moq-transport-setup-format title="MOQT SETUP Message"}
@@ -704,18 +707,9 @@ publication.]]
 The version number for the final version of this specification (0x00000001), is reserved for the version of the protocol that is published as an RFC.
 Version numbers used to identify IETF drafts are created by adding the draft number to 0xff000000. For example, draft-ietf-moq-transport-13 would be identified as 0xff00000D.
 
-### SETUP Parameters {#setup-parameters}
+### Role {#role}
 
-Every parameter MUST appear at most once within the SETUP message. The
-peers SHOULD verify that and close the connection if a parameter appears
-more than once.
-
-The ROLE parameter is mandatory for the client. All of the other
-parameters are optional.
-
-#### ROLE parameter {#role}
-
-The ROLE parameter (key 0x00) allows the client to specify what roles it
+The Role field allows the client to specify what roles it
 expects the parties to have in the MOQT connection. It has three
 possible values:
 
@@ -733,23 +727,25 @@ possible values:
 
 : Both the client and the server are expected to send objects.
 
-The client MUST send a ROLE parameter with one of the three values
-specified above. The server MUST close the connection if the ROLE
-parameter is missing, is not one of the three above-specified values, or
-it is different from what the server expects based on the application.
+The server MUST close the connection if the Role field is not one of the three
+three above-specified values, or it is different from what the server expects
+based on the application.
 
-#### PATH parameter {#path}
+### Path {#path}
 
-The PATH parameter (key 0x01) allows the client to specify the path of
-the MoQ URI when using native QUIC ({{QUIC}}).  It MUST NOT be used by
-the server, or when WebTransport is used.  If the peer receives a PATH
-parameter from the server, or when WebTransport is used, it MUST close
-the connection.
+The Length of Path and Path fields are only present for connections over native
+QUIC. The Path field allows the client to specify the path of the MoQ URI when
+using native QUIC ({{QUIC}}). 
 
 When connecting to a server using a URI with the "moq" scheme, the
 client MUST set the PATH parameter to the `path-abempty` portion of the
 URI; if `query` is present, the client MUST concatenate `?`, followed by
 the `query` portion of the URI to the parameter.
+
+###  Parameters {#setup-parameters}
+
+The Parameters field offers an extension point for future capabilities
+negotiation. Its semantics are described in {{params}}.
 
 ## OBJECT {#message-object}
 
@@ -797,7 +793,7 @@ The format of SUBSCRIBE REQUEST is as follows:
 SUBSCRIBE REQUEST Message {
   Full Track Name Length (i),
   Full Track Name (...),
-  Track Request Parameters (..) ...
+  Parameters (..) ...
 }
 ~~~
 {: #moq-transport-subscribe-format title="MOQT SUBSCRIBE REQUEST Message"}
@@ -805,7 +801,7 @@ SUBSCRIBE REQUEST Message {
 
 * Full Track Name: Identifies the track as defined in ({{track-name}}).
 
-* Track Request Parameters: As defined in {{track-req-params}}.
+* Parameters: As defined in {{params}}.
 
 On successful subscription, the publisher SHOULD start delivering
 objects from the group sequence and object sequence as defined in the
@@ -894,7 +890,7 @@ publish tracks under this namespace.
 ANNOUNCE Message {
   Track Namespace Length(i),
   Track Namespace(..),
-  Track Request Parameters (..) ...,
+  Parameters (..) ...,
 }
 ~~~
 {: #moq-transport-announce-format title="MOQT ANNOUNCE Message"}
@@ -903,7 +899,7 @@ ANNOUNCE Message {
 ({{track-name}})
 
 * Track Request Parameters: The parameters are defined in
-{{track-req-params}}.
+{{params}}.
 
 ## ANNOUNCE OK {#message-announce-ok}
 
@@ -993,31 +989,38 @@ The client:
   processing objects from both in parallel.
 
 
-## Track Request Parameters {#track-req-params}
+## Parameters {#params}
 
-The Track Request Parameters identify properties of the track requested
-in either the ANNOUNCE or SUSBCRIBE REQUEST control messages. The peers
-MUST close the connection if there are duplicates. The Parameter Value
-Length field indicates the length of the Parameter Value.
+Some messages include a Parameters field that contain a series of Type-
+Length-Value encoded optional parameters.
 
-The format of `Track Request Parameter` is as follows:
+Senders MUST NOT repeat the same parameter type in a message. Receivers
+SHOULD check that there are no duplicate parameters and close the connection
+if found.
+
+Receivers do not process unrecognized parameter types, using the Parameter
+Length field to skip to the next parameter.
+
+The format of `Parameter` is as follows:
 
 ~~~
-Track Request Parameter {
-  Track Request Parameter Key (i),
-  Track Request Parameter Length (i),
-  Track Request Parameter Value (..),
+Parameter {
+  Parameter Key (i),
+  Parameter Length (i),
+  Parameter Value (..),
 }
 ~~~
 {: #moq-track-request-param format title="MOQT Track Request Parameter"}
 
+The Parameter Length field encodes the length of the Parameter Value field.
 
 ### GROUP SEQUENCE Parameter
 
 The GROUP SEQUENCE parameter (key 0x00) identifies the group within the
 track to start delivering objects. The publisher MUST start delivering
 the objects from the most recent group, when this parameter is
-omitted. This parameter is applicable in SUBSCRIBE REQUEST message.
+omitted. This parameter is applicable in SUBSCRIBE REQUEST message and will be
+ignored otherwise.
 
 ### OBJECT SEQUENCE Parameter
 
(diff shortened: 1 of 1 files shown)


kixelated (changes requested):
SETUP needs to support arbitrary parameters for version negotiation. Version 00 might require role/path, but other versions might not. These fields would be deprecated pretty quickly as set to 0 much like some TLS headers before the switch was made to generic fields.

However if we used ALPN or equivalent in WebTransport for version negotiation, then I would love to used a fixed encoding. It makes life sooo much easier. I would like to use a fixed encoding in SUBSCRIBE/ANNOUNCE since the version has already been negotiated though.

suhasHere:
I generally like the direction of this PR. But would like to discuss this as part of or post interop.

fluffy:
Don't see this as big deal one way another but looks fine to me.

huitema:
Check the [transport parameter encoding in QUIC](https://www.rfc-editor.org/rfc/rfc9000.html#name-transport-parameter-encodin):
~~~
Transport Parameter {
  Transport Parameter ID (i),
  Transport Parameter Length (i),
  Transport Parameter Value (..),
}
~~~
I think this is solving the exact same problem as Setup parameters: transport parameters with unknown ID can just be skipped.

martinduke:
Major changes from Boston:
- There are SETUP parameters (which are constant across all versions) and version-specific parameters (which apply to all messages except SETUP-- SUBSCRIBE_REQUEST and ANNOUNCE).
- The LSB of the type indicates if the length is varint-encoded or the length is itself a varint.
- Editorial rearranging.
- Keeps the new rules about duplicate parameters and ignoring inapplicable parameters.

draft-ietf-moq-transport.md
```diff
@@ -652,6 +652,119 @@ continues until the end of the stream.
...
+parameter. SETUP message parameters use a namespace constant across all MoQ
+Transport versions. All other messages use a version-specific namespace. For
+example, the integer '1' can refer to different parameters for SETUP messages
+and other message types.
+
+The Parameter Length field of the String Parameter encodes the length
+of the Parameter Value field in bytes.
+
+Boolean parameters (e.g., to advertise support for a capability) can use the
+String Parameter format with a zero length.
```
afrind, line 700: Seems like they could also be represented as an Int param with value=0 or 1.  Is that allowed?

draft-ietf-moq-transport.md
```diff
@@ -652,6 +652,119 @@ continues until the end of the stream.
...
+URI; if `query` is present, the client MUST concatenate `?`, followed by
+the `query` portion of the URI to the parameter.
+
+### Version-specific parameters {#version-specific-params}
+
+These parameters only apply to version 1 and messages other than SETUP.
+
+#### GROUP SEQUENCE Parameter
+
+The GROUP SEQUENCE parameter (type 0x01) identifies the group within the
```
afrind, line 749: nit: Would GROUP START SEQUENCE or just GROUP START be a more apt name?  Particularly if we add GROUP END later.

draft-ietf-moq-transport.md
```diff
@@ -652,6 +652,119 @@ continues until the end of the stream.
...
+track to start delivering objects. The publisher MUST start delivering
+the objects from the most recent group, when this parameter is
+omitted.
```
afrind, line 752: Editorial: with the `, when this param` reads awkwardly to me.  I think it's better without the comma, or start the sentence with the when clause?

draft-ietf-moq-transport.md
```diff
@@ -652,6 +652,119 @@ continues until the end of the stream.
...
+
+#### AUTHORIZATION INFO Parameter
+
+AUTHORIZATION INFO parameter (key 0x02) identifies track's authorization
+information. This parameter is populated for cases where the
+authorization is required at the track level.
+
```
afrind, line 759: Did we put AUTH before OBJECT SEQUENCE because we're listing them numerically?  I think keeping GROUP and OBJECT next to each other will read better.

draft-ietf-moq-transport.md
```diff
@@ -652,6 +652,119 @@ continues until the end of the stream.
...
+#### OBJECT SEQUENCE Parameter
+
+The OBJECT SEQUENCE parameter (key 0x03) identifies the object with the
```
afrind, line 762: nit: same name comment as with GROUP SEQUENCE

draft-ietf-moq-transport.md
```diff
@@ -903,7 +978,8 @@ ANNOUNCE Message {
 ({{track-name}})
 
 * Track Request Parameters: The parameters are defined in
-{{track-req-params}}.
+{{version-specific-params}}. The Authorization Info is optional but is
+applicable to this message. All other parameters are ignored.
```
afrind, line 982: Maybe:

Authorization Info is applicable to this message, but is not mandatory.
.

afrind, line 982: Maybe:

Authorization Info is applicable to this message, but is not mandatory.
.

draft-ietf-moq-transport.md
```diff
@@ -797,15 +871,16 @@ The format of SUBSCRIBE REQUEST is as follows:
...
 }
 ~~~
 {: #moq-transport-subscribe-format title="MOQT SUBSCRIBE REQUEST Message"}
 
 
 * Full Track Name: Identifies the track as defined in ({{track-name}}).
 
-* Track Request Parameters: As defined in {{track-req-params}}.
+* Parameters: As defined in {{version-specific params}}. All are optional
+but valid for this message.
```
afrind, line 883: Maybe this is best captured with a small table like:

```
Param     | Mandatory / Optional / Invalid
-----------------
GROUP SEQ | Optional
OBJ SEQ   | Optional
AUTH INFO | Optional
```
 Similar for other messages with params

draft-ietf-moq-transport.md
afrind, line 982: See above comment about a table.

afrind (commented):
Individual/Editorial Review

draft-ietf-moq-transport.md
```diff
@@ -652,6 +652,119 @@ continues until the end of the stream.
...
+
+This document specifies the following SETUP Parameters:
+
+#### Role {#role}
+
+The Role parameter has a type of 0x01 is therefore an integer type. It allows
+the client to specify what roles it expects the parties to have in the MOQT
+connection. It has three possible values:
+
+0x01:
```
fluffy, line 716: I would prefer to see these have an named value or label then an IANA table that mapped that name or label to the value 0x01

fluffy (commented):
I'd like to review this in more detail but initial reaction is mostly looks fine to me.

huitema (commented):
The proposed syntax looks like early optimization, trying to save a few bytes of encoding by having two classes of identifiers. I would much rather have something like the QUIC transport parameters:

~~~
Parameter {
   Parameter Type (i),
   Parameter Length (i),
   Parameter Value (...)
}
~~~

My proposal would assume that parameter types are standardized, or at least registered, and that the registration specifies the encoding. That encoding will in many cases be "integer" or "string", but it does not have to be forever.

draft-ietf-moq-transport.md
martinduke: Well the issue with that is that you to have to extract the length if you don't know the parameter type.

kixelated:
Yeah, maybe we just copy QUIC... which is basically the current encoding. I didn't realize that QUIC transport parameters include a redundant size for varints.

I do like Cullen's suggestion that the encoding is specific to the ID. It means we can have other encodings other than string/int, kinda like `preferred_address` in QUIC. The cost is that we include the redundant size for varints, but whatever.

martinduke:
> The proposed syntax looks like early optimization, trying to save a few bytes of encoding by having two classes of identifiers. I would much rather have something like the QUIC transport parameters:
> 
> ```
> Parameter {
>    Parameter Type (i),
>    Parameter Length (i),
>    Parameter Value (...)
> }
> ```
> 
> My proposal would assume that parameter types are standardized, or at least registered, and that the registration specifies the encoding. That encoding will in many cases be "integer" or "string", but it does not have to be forever.

I agree that QUIC does this. It does create the error case where the varint length is different from the parameter length, and it's nice when encodings don't allow these kinds of errors

suhasHere:
Leaving a note for myself. I would love to review this post interim.

draft-ietf-moq-transport.md
```diff
@@ -652,6 +652,119 @@ continues until the end of the stream.
...
+Some messages include a Parameters field that encode optional message
+elements. There are two formats that encode a type, length, and value in
+some form.
+
+Senders MUST NOT repeat the same parameter type in a message. Receivers
+SHOULD check that there are no duplicate parameters and close the connection
+if found.
+
+Receivers do not process unrecognized parameter types, using the encoded length
+to skip to the next parameter.
```
suhasHere, line 666: I would rather just say, "Receiver ignore unrecognized parameters".

draft-ietf-moq-transport.md
```diff
@@ -652,6 +652,119 @@ continues until the end of the stream.
...
+  Parameter Type (i),
+  Parameter Length (i),
+  Parameter Value (..),
+}
+~~~
+{: #moq-string-param format title="MOQT String Parameter"}
+
+Parameter Type is an integer that indicates the semantic meaning of the
+parameter. SETUP message parameters use a namespace constant across all MoQ
+Transport versions. All other messages use a version-specific namespace. For
```
suhasHere, line 692: what is a namespace constant?

draft-ietf-moq-transport.md
```diff
@@ -652,6 +652,119 @@ continues until the end of the stream.
...
+example, the integer '1' can refer to different parameters for SETUP messages
+and other message types.
```
suhasHere, line 694: but unclear on what it means different to other message types. May be more clarification is needed ?

draft-ietf-moq-transport.md
```diff
@@ -652,6 +652,119 @@ continues until the end of the stream.
...
+
+The Parameter Length field of the String Parameter encodes the length
+of the Parameter Value field in bytes.
```
suhasHere, line 697: May be this comment is not for this PR. The way it is defined today, will limit the lenghth of the string to 7 bytes. are we ok with that ?

draft-ietf-moq-transport.md
suhasHere: I would rather define a Bool parameter type than overloading the meaning of the value.

draft-ietf-moq-transport.md
```diff
@@ -652,6 +652,119 @@ continues until the end of the stream.
...
+
+0x02:
+
+: Only the server is expected to send objects on the connection. This is
+  commonly referred to as the delivery case.
+
+0x03:
+
+: Both the client and the server are expected to send objects.
+
```
suhasHere, line 729: the role parameter needs to be optional or there should be a default value if not provided.

draft-ietf-moq-transport.md
suhasHere: Please add reference to to section defining groups

draft-ietf-moq-transport.md
suhasHere: @afrind  this comes due to integer/string encoding ;)

draft-ietf-moq-transport.md
suhasHere: please add reference to section defining object sequence

draft-ietf-moq-transport.md
kixelated: I'm not sure I follow. The parameter length is a VarInt with a maximum size of 2^62

suhasHere (changes requested):
Thanks for the work Martin. I do have few nits and clarification questions. Overall i feel this is looking good.

draft-ietf-moq-transport.md
kixelated: I think role is very useful, but kind of annoying as it is currently defined. The application has to know and specify the intention of the server, which also means there's no possible default. I would like to see the role describe the local endpoint's capability, not the server's intended behavior. See #151 

But this is out of scope of the PR.

draft-ietf-moq-transport.md
kixelated: +1

draft-ietf-moq-transport.md
kixelated: Yeah my concern is that we're going to have a full compatibility matrix when we add more params to more messages. I'd rather just define the parameters for each messages individually, instead of sharing one big parameter namespace.

draft-ietf-moq-transport.md
kixelated: Maybe remove this line? I think using both a `string` and `int` are valid ways to encode a boolean, and it's up to the parameter to specify the encoding.

I'm not sure you want to use a string though. Do you just put arbitrary data if it's a truthy value, zero bytes if it's falsy, and missing if it's undefined?

martinduke: The model I had in mind was for extension negotiation, where presence of the parameter was an implied True. I think there's value of setting a norm for how this works. But since this seems to be creating a lot of trouble I'll just delete it.

draft-ietf-moq-transport.md
martinduke: Added the reference. I don't mind the name change but I took this from the current draft and would prefer a separate issue/PR for parameter renaming.

draft-ietf-moq-transport.md
martinduke: a namespace THAT IS constant across...

draft-ietf-moq-transport.md
martinduke: and for all other message types...

draft-ietf-moq-transport.md
martinduke: yes, the limit is 2^62.

draft-ietf-moq-transport.md
martinduke: I agree that this is out of scope.  See #250

martinduke:
There are a couple of sub-issues where I'm not quite ready to surrender yet: length encoding, and the number of parameter registries/namespaces we're going to have. I sent an email to the list to discuss further.

draft-ietf-moq-transport.md
fluffy: no I did not mean that. I mean that in the text of the draft we in the role section we do not put 0x01, we put a name for that constant like SETUP_PAR_ROLE then later in the document we have an IANA table that defined SETUP_PAR_ROLE  as 0x01. This is only a change in the way the doc is organized, not anything on the wire. I really don't care about this very much but I find it an easy way to make sure that two different places in the doc did not accidentally use 0x01 for when they mean to have different code points.  It also helps with library API because then people end up definiing the constants in the API to match the spec and it makes it easier to figure out what the API does. Again, all way secondary benefits and we can do this way later in the spec development. Do not need to figure this out to merge this PR.

martinduke: Ah, i see now that I massively misinterpreted your comment. I added a TODO since the IANA considerations section is not ready for this yet.

kixelated:
My ideal solution would be to keep flexible SETUP parameters for version/extension negotiation, but use fixed encodings for other messages. Or move optional parameters like `auth` to dedicated messages.

afrind:
Are we planning to rebase this on main for the remaining issues, or should we close it?

martinduke:
I think this is rapidly becoming OBE, closing.

PR #258, opened 2023-09-29
Allow the Track Alias to be selected by the publisher

kixelated:
Currently SUBSCRIBE_OK and OBJECT will race. If OBJECT wins (>50% chance for multi-threaded QUIC libraries), then the subscriber will receive an unknown OBJECT that needs to be buffered. Moving Track ID to SUBSCRIBE fixes this race, removes the need to buffer OBJECTs in the MoqT library, and simplifies the flow. See #145


draft-ietf-moq-transport.md
```diff
@@ -818,21 +821,13 @@ A `SUBSCRIBE OK` control message is sent for successful subscriptions.
...
 {: #moq-transport-subscribe-ok format title="MOQT SUBSCRIBE OK Message"}
 
-* Full Track Name: Identifies the track for which this response is
-provided.
-
-* Track ID: Session specific identifier that is used as an alias for the
-Full Track Name in the Track ID field of the OBJECT ({{message-object}})
-message headers of the requested track. Track IDs are generally shorter
-than Full Track Names and thus reduce the overhead in OBJECT messages.
+* Track ID: The track identifier as specified in the SUBSCRIBE REQUEST {{message-subscribe-req}}.
```
suhasHere, line 830: This is fine as long as we have a way to let server propose a trackId in the SubscribeError or replace the one in here.

suhasHere (changes requested):

draft-ietf-moq-transport.md
vasilvv: I've still not seen documentation of a use case in which server-proposed ID is useful.  Can we have an example that we can work through?

kixelated: Yeah, I just don't understand the use-case. 

You could still use the same ID within the CDN (ex. hash of the name). The only difference is that the ID is rewritten on egress instead of ingress. There's a tangible difference between the two because of fanout, but we're talking about encoding a VarInt; it will never show up on profiles.

Could you file an issue and really dive into the problem?

draft-ietf-moq-transport.md
```diff
@@ -795,13 +794,17 @@ The format of SUBSCRIBE REQUEST is as follows:
 
 ~~~
 SUBSCRIBE REQUEST Message {
+  Track ID (i),
```
fluffy, line 797: I would propose add a subscribe ID

draft-ietf-moq-transport.md
```diff
@@ -818,21 +821,13 @@ A `SUBSCRIBE OK` control message is sent for successful subscriptions.
 ~~~
 SUBSCRIBE OK
 {
-  Full Track Name Length(i),
-  Full Track Name(...),
-  Track ID(i),
+  Track ID (i),
```
fluffy, line 824: change to subscribe ID

draft-ietf-moq-transport.md
```diff
@@ -847,17 +842,15 @@ failed SUBSCRIBE REQUEST.
 ~~~
 SUBSCRIBE ERROR
 {
-  Full Track Name Length(i),
-  Full Track Name(...),
+  Track ID (i),
```
fluffy, line 845: change to subscribe ID

draft-ietf-moq-transport.md
```diff
@@ -847,17 +842,15 @@ failed SUBSCRIBE REQUEST.
...
   Error Code (i),
   Reason Phrase Length (i),
   Reason Phrase (...),
```
fluffy, line 848: Add an error code for, use differnt track-id and add a field of proposed track-id

draft-ietf-moq-transport.md
```diff
@@ -875,13 +868,12 @@ The format of `UNSUBSCRIBE` is as follows:
 
 ~~~
 UNSUBSCRIBE Message {
-  Full Track Name Length (i),
-  Full Track Name (...),
+  Track ID (i),
```
fluffy, line 871: change to subscribe id

draft-ietf-moq-transport.md
fluffy: sure simple use case that is given over and over again, eliminate the need to keep track of per hop rewrite information and do the rewrites.

fluffy (changes requested):
Proposed somethings that I think would allow us to get to consensus on this one.

kixelated:
@fluffy @suhasHere I think #305 should immediately follow and split Track ID into Subscription ID and Track Alias.

draft-ietf-moq-transport.md
```diff
@@ -1073,25 +1067,24 @@ failed SUBSCRIBE.
...
-* Track Name: Identifies the track name as defined in ({{track-name}}).
+* Track ID: The track identifier as specified in the SUBSCRIBE REQUEST {{message-subscribe-req}}.
 
 * Error Code: Identifies an integer error code for subscription failure.
 
-* Reason Phrase Length: The length in bytes of the reason phrase.
-
 * Reason Phrase: Provides the reason for subscription error.
 
+* New Track ID: When not equal to Track ID, the subscriber SHOULD re-issue the SUBSCRIBE with this Track ID instead if possible.
```
suhasHere, line 1084: nit, if possible is redundant

suhasHere (approved):

fluffy (approved):
This looks good and I really hope we can agree on this compromise. It's not perfect for everyone but I think it is the best we can get to now.

draft-ietf-moq-transport.md
```diff
@@ -1073,25 +1072,24 @@ failed SUBSCRIBE.
...
+* Track ID: The track identifier as specified in the SUBSCRIBE REQUEST {{message-subscribe-req}}.
 
 * Error Code: Identifies an integer error code for subscription failure.
 
-* Reason Phrase Length: The length in bytes of the reason phrase.
-
 * Reason Phrase: Provides the reason for subscription error.
 
+* New Track ID: When not equal to Track ID, the subscriber SHOULD re-issue the SUBSCRIBE with this Track ID instead.
+  If this Track ID is already in use, the receiver MUST close the connection with a Duplicate Track ID error ({{session-termination}}).
```
kixelated, line 1090: @fluffy @suhasHere is this okay with you? There could be a race where:

subscriber:
```
send SUBSCRIBE id=69 name=foo
send SUBSCRIBE id=420 name=bar
```

publisher:
```
recv SUBSCRIBE id=69 name=foo
send SUBSCRIBE_ERROR id=69 new_id=420 # publisher remap from 69 -> 420
recv SUBSCRIBE id=420 name=bar
send SUBSCRIBE_ERROR id=420 new_id=1234 # publisher rename from 420 -> 1234
```

subscriber:
```
recv SUBSCRIBE_ERROR id=69 new_id=420
# invalid!: SUBSCRIBE id=420 name=foo
```

The subscriber could wait until it receives the anticipated `SUBSCRIBE_ERROR id=420`. However, I doubt anybody would ever implement this, and you would still need throw an error if you ever receive `SUBSCRIBE_OK id=420`. 

You can avoid this race by sharding publisher/subscriber IDs. Or the publisher can use an unguessable track ID (ex. hash) and it's virtually impossible to ever encounter this, which I think is gud enough.

draft-ietf-moq-transport.md
```diff
@@ -952,6 +956,11 @@ SUBSCRIBE REQUEST Message {
 ~~~
 {: #moq-transport-subscribe-format title="MOQT SUBSCRIBE Message"}
 
+* Track ID: A session specific identifier for the track.
+  Messages that reference a track, such as OBJECT ({{message-object}}),
+  reference this Track ID instead of the Full Track Name to reduce overhead.
+  If the Track ID is already in use, the receiver MUST close the session with a Duplicate Track ID error ({{session-termination}}).
```
kixelated, line 962: I added this line. We actually didn't throw an error if you received a SUBSCRIBE for two different tracks using the same ID. I think we can share the error code instead of making a new one.

vasilvv (approved):
LGTM on the basis of this being an improvement over the current text in the draft.

draft-ietf-moq-transport.md
fluffy: It seems like once we have separate the subscribe-id and track-id, this problem goes away ? Perhaps I misunderstand but agree on if a publisher ever suggests reusing a track-id that is already in use for a different track, that is totally borked and the connection should be closed with an error.

kixelated: I think the problem still exists when subscribe-id and track-id are split.

The problem is that there's an RTT chasm between the publisher and subscriber. The publisher may send `SUBSCRIBE_ERROR id=1 new_alias=420`, but there was a whole RTT where the subscriber MAY have decided to `SUBSCRIBE id=2 alias=420` for some completely different track (collision). The subscriber can't send `SUBSCRIBE id=2 alias=420` like the publisher is requesting until the existing `SUBSCRIBE id=1 alias=420` in flight has been rejected. 

I think maybe the subscriber could `UNSUBSCRIBE id=1` and then immediately `SUBSCRIBE id=2 alias=420` though thanks to the split. It would still be absolutely fatal if the subscriber receives `SUBSCRIBE_OK id=1`, even after the unsubscribe was sent, because it would be impossible to disambiguate `OBJECT alias=420`.

draft-ietf-moq-transport.md
fluffy: I'd like to nuance this.  I think a common implementation will be to hash the full track name to get the track id and what we need to say here is that it is only an error if the track-id is reused *and* it does not refer to the same full track name. So if I had two differnt subscribes both to the same track, they could use the same track-id

draft-ietf-moq-transport.md
```diff
@@ -961,18 +967,26 @@ SUBSCRIBE Message {
...
 `Subscribe ID` and it is included in the corresponding SUBSCRIBE_OK or
 SUBSCRIBE_ERROR messages.
 
+* Track Alias: A session specific identifier for the track.
+Messages that reference a track, such as OBJECT ({{message-object}}),
+reference this Track Alias instead of the Track Name and Track Namespace to
+reduce overhead. If the Track Alias is already in use, the publisher MUST
+close the session with a Duplicate Track Alias error ({{session-termination}}).
+Unless the publisher responsds with a different Track Alias in SUBSCRIBE_ERROR,
+the subscriber SHOULD specify a Track Alias that is equal to the Subscribe ID.
```
ianswett, line 983:
```suggestion
the subscriber MUST specify a Track Alias that is equal to the Subscribe ID.
```

draft-ietf-moq-transport.md
```diff
@@ -1105,10 +1112,14 @@ SUBSCRIBE_ERROR
...
-* Reason Phrase Length: The length in bytes of the reason phrase.
-
 * Reason Phrase: Provides the reason for subscription error.
 
+* Track Alias: When not equal to Subscribe ID, the subscriber SHOULD re-issue the
+  SUBSCRIBE with this Track Alias instead. If this Track Alias is already in use,
+  the receiver MUST close the connection with a Duplicate Track Alias error
+  ({{session-termination}}).
+  TODO: Add a registry for subscribe error codes and make this field conditional.
+
```
afrind, line 1122: I think you need to change the condition here.  

C->S: Subscribe (sub-id 1, track alias 1)
S->C: Subscribe Error (sub-id 1, track alias 19, error code=try new id) 
C->S: Subscribe (sub-id 2, track alias 19)
S->C: Subscribe Error (sub-id 2, track alias 19, error code=bad auth)  # sub-id != track alias, but server isn't asking for a resubscribe with new id

draft-ietf-moq-transport.md
```diff
@@ -881,8 +886,8 @@ OBJECT Message {
 ~~~
 {: #moq-transport-object-format title="MOQT OBJECT Message"}
 
-* Track Alias : The compressed full track name obtained as part of
-subscription and/or publish control message exchanges.
+* Track Alias :The track identifier as specified in the
+  SUBSCRIBE request {{message-subscribe-req}}.
```
suhasHere, line 891: "The Track Alias as specified in the SUBSCRIBE message {{message-subscribe-req}}"

draft-ietf-moq-transport.md
```diff
@@ -961,18 +967,26 @@ SUBSCRIBE Message {
...
 the publishers to identify a given subscription. Subscribers specify the
 `Subscribe ID` and it is included in the corresponding SUBSCRIBE_OK or
 SUBSCRIBE_ERROR messages.
 
+* Track Alias: A session specific identifier for the track.
+Messages that reference a track, such as OBJECT ({{message-object}}),
+reference this Track Alias instead of the Track Name and Track Namespace to
+reduce overhead. If the Track Alias is already in use, the publisher MUST
+close the session with a Duplicate Track Alias error ({{session-termination}}).
+Unless the publisher responsds with a different Track Alias in SUBSCRIBE_ERROR,
```
suhasHere, line 982: s/responsds/responds

draft-ietf-moq-transport.md
```diff
@@ -961,18 +967,26 @@ SUBSCRIBE Message {
...
+the subscriber MUST specify a Track Alias that is equal to the Subscribe ID.
+
```
suhasHere, line 984: I don't see the need for ``` subscriber MUST specify a Track Alias that is equal to the Subscribe ID ```.

suhasHere: May be all we need to say is .. 

"Unless the publisher responds with a different Track Alias in SUBSCRIBE_ERROR for the subscriber to retry the SUBSCRIBE with publisher chosen Track Alias"

draft-ietf-moq-transport.md
suhasHere: I am not understanding why a track alias value must match subscribe id.

suhasHere (changes requested):

draft-ietf-moq-transport.md
```diff
@@ -418,6 +420,9 @@ code, as defined below:
 * Protocol Violation: The remote endpoint performed an action that was
   disallowed by the specification.
 
+* Duplicate Track Alias: The endpoint attempted to use a Track Alias
+that was already in use.
```
ianswett, line 424:
```suggestion
  that was already in use.
```

draft-ietf-moq-transport.md
```diff
@@ -936,18 +942,26 @@ SUBSCRIBE Message {
...
 the publishers to identify a given subscription. Subscribers specify the
 `Subscribe ID` and it is included in the corresponding SUBSCRIBE_OK or
 SUBSCRIBE_ERROR messages.
 
+* Track Alias: A session specific identifier for the track.
+Messages that reference a track, such as OBJECT ({{message-object}}),
+reference this Track Alias instead of the Track Name and Track Namespace to
+reduce overhead. If the Track Alias is already in use, the publisher MUST
+close the session with a Duplicate Track Alias error ({{session-termination}}).
+Unless the publisher responsds with a different Track Alias in SUBSCRIBE_ERROR,
```
ianswett, line 957:
```suggestion
Unless the publisher responds with a different Track Alias in SUBSCRIBE_ERROR,
```

draft-ietf-moq-transport.md
```diff
@@ -936,18 +942,26 @@ SUBSCRIBE Message {
...
 `Subscribe ID` and it is included in the corresponding SUBSCRIBE_OK or
 SUBSCRIBE_ERROR messages.
 
+* Track Alias: A session specific identifier for the track.
+Messages that reference a track, such as OBJECT ({{message-object}}),
+reference this Track Alias instead of the Track Name and Track Namespace to
+reduce overhead. If the Track Alias is already in use, the publisher MUST
+close the session with a Duplicate Track Alias error ({{session-termination}}).
+Unless the publisher responds with a different Track Alias in SUBSCRIBE_ERROR,
+the subscriber MUST specify a Track Alias that is equal to the Subscribe ID.
```
ianswett, line 958:
```suggestion
the subscriber can specify a Track Alias that is equal to the Subscribe ID.
```

draft-ietf-moq-transport.md
ianswett: I changed it to a 'can' from a MUST

draft-ietf-moq-transport.md
ianswett: The idea would be to prevent some sort of loop where the server keeps coming back with a new subscribe ID, but the text also assumes the MUST I just removed.

draft-ietf-moq-transport.md
suhasHere: SubscribeId and Track Alias are semantically different concepts. Former is per transaction (subscription) and latter is a short name for Full Track Name. we should remove the last sentence above.

draft-ietf-moq-transport.md
suhasHere: > S->C: Subscribe Error (sub-id 2, track alias 19, error code=bad auth) # sub-id != track alias, but server isn't asking for a resubscribe with new id

I don't see why this is an error. 

I expected the flow to be something like:

```
C->S: Subscribe (sub-id 1, track alias 1)
S->C: Subscribe Error (sub-id 1, track alias 19, error code=try new track alias)
C->S: Subscribe (sub-id 2, track alias 19)
S->C: Sub Ok 
```

draft-ietf-moq-transport.md
```diff
@@ -856,8 +861,8 @@ OBJECT Message {
 ~~~
 {: #moq-transport-object-format title="MOQT OBJECT Message"}
 
-* Track Alias : The compressed full track name obtained as part of
-subscription and/or publish control message exchanges.
+* Track Alias :The track identifier as specified in the
+  SUBSCRIBE request {{message-subscribe-req}}.
```
ianswett, line 865:
```suggestion
  SUBSCRIBE message {{message-subscribe-req}}.
```

draft-ietf-moq-transport.md
```diff
@@ -936,18 +942,26 @@ SUBSCRIBE Message {
...
 SUBSCRIBE_ERROR messages.
 
+* Track Alias: A session specific identifier for the track.
+Messages that reference a track, such as OBJECT ({{message-object}}),
+reference this Track Alias instead of the Track Name and Track Namespace to
+reduce overhead. If the Track Alias is already in use, the publisher MUST
+close the session with a Duplicate Track Alias error ({{session-termination}}).
+Unless the publisher responds with a different Track Alias in SUBSCRIBE_ERROR,
+the subscriber can specify a Track Alias that is equal to the Subscribe ID.
+
```
ianswett, line 959:
```suggestion

```

suhasHere (approved):

Issue #227, opened 2023-07-24
Announces with same name in reconnect


fluffy:
When the relays received an ANNOUNCE with exact same name as previous ANNOUNCE, what happens? 

Options seem to be replace previous ANNOUNCE data or return an error. 

This is related to https://github.com/moq-wg/moq-transport/issues/225 and https://github.com/moq-wg/moq-transport/issues/204

Discussion to follow in thread

fluffy:
One of use cases we should think about is a client that crashes and reconnects.

afrind:
As an individual, 

The crash (or disconnect) use case is compelling for "the most recent one wins".  Are there applications where that policy would break something?

fluffy:
I agree with Alan that "most recent" overwrites any earlier ones is best path forward.

suhasHere:
+1 on most recent overwrites. this is what we do atleast today and it works fine to the amount we have tested.

kixelated:
Going with the most recent ANNOUNCE works for me, so long as the relay MUST send an `ANNOUNCE_ERROR` and `UNSUBSCRIBE` for any matching subscriptions to the original session. Otherwise we'll introduce zombie announcements and subscriptions.

fluffy:
WFM

suhasHere:
> Going with the most recent `ANNOUNCE` works for me, so long as the relay MUST send an `ANNOUNCE_ERROR` and `UNSUBSCRIBE` for any matching subscriptions to the original session. Otherwise we'll introduce zombie announcements and subscriptions.

I Dont think existing subscriptions needs to be terminated. 

Let say the case where, A publisher disconnects due to getting into an elevator for a moment and sends announce on reconnect, sending out protocol messages to cancel the subscriptions on the second announce is too noisy.

afrind:
As an individual: 

I don't think we need to force unsubscribe everyone if the relay successfully resubscribes to the new ANNOUNCE.

kixelated:
> As an individual:
> 
> I don't think we need to force unsubscribe everyone if the relay successfully resubscribes to the new ANNOUNCE.
> 
> I think Luke is suggesting that the relay unsubscribe on the session that issued the old ANNOUNCE?

Yeah exactly. The upstream subscription is reissued to the new connection, but any downstream subscriptions can be maintained by updating a lookup table. I can elaborate what I mean if this isn't clear.

ianswett:
Closed by #525

PR #264, opened 2023-10-06
Session migration and GOAWAY encoding

### LINKED ISSUE #32: GOAWAY seems overly restrictive
The general motivation here isn't bad but mandating client behaviours in response to these types of messages seems a bit like overreach. The requirements also more restrictive than H2 or H3 GOAWAY, which are on theory also available as a capability in the same connection. 

Rephrasing the text can probably hit the notes you want with respect to preserving service continuity, without painting people into corners.

### LINKED ISSUE #213: No encoding for GOAWAY
The draft doesn't specify how to encode GOAWAY. [See the draft](https://moq-wg.github.io/moq-transport/draft-lcurley-moq-transport.html#name-goaway). At a minimum we need a new connect URL.

kixelated:
Fixes #213 
Fixes #32

Is this too much? Do we need to prescribe an algorithm or can we rely on clients to figure it out?

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -436,13 +436,14 @@ the current URI is reused. The server SHOULD terminate the session with GOAWAY
 Timeout after a sufficient timeout if a stubborn client is preventing a drain.
 
 The GOAWAY message does not impact subscription state. A subscriber SHOULD
-terminate existing subscriptions while a publisher MAY reject new subscriptions
-while in this draining state. Note that server may be a subscriber, in which
-case it should send a GOAWAY message followed by any UNSUBSCRIBE messages.
+individually UNSUBSCRIBE for each existing subscription, while a publisher MAY
+reject new SUBSCRIBEs while in this draining state. Note that server may be a
+subscriber, in which case it should send a GOAWAY message followed by any
+UNSUBSCRIBE messages.
 
 The client receives a GOAWAY message and determines when to terminate the
-session. It's RECOMMENDED that the client waits until all active subscriptions
-have terminated and then closes the session with NO\_ERROR. This should be
+session. It's RECOMMENDED that the client waits until there are no more active
+subscriptions and then closes the session with NO\_ERROR. This should be
 performed transparently to the application which involves establishing the new
 session in the background and migrating all active subscriptions/announcements
 from the old session. The client may choose to delay closing the session if
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -437,8 +437,8 @@ Timeout after a sufficient timeout if a stubborn client is preventing a drain.
 
 The GOAWAY message does not impact subscription state. A subscriber SHOULD
 individually UNSUBSCRIBE for each existing subscription, while a publisher MAY
-reject new SUBSCRIBEs while in this draining state. Note that server may be a
-subscriber, in which case it should send a GOAWAY message followed by any
+reject new SUBSCRIBEs while in this draining state. Note that the server may be
+a subscriber, in which case it should send a GOAWAY message followed by any
 UNSUBSCRIBE messages.
 
 The client receives a GOAWAY message and determines when to terminate the
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -426,27 +426,28 @@ pre-negotiated by the application.
 MoqTransport requires a long-lived and stateful session. However, a service
 provider needs the ability to shutdown/restart a server without waiting for all
 sessions to drain naturally, as that can take days for long-form media.
+MoqTransport avoids this via the GOAWAY message ({{message-goaway}}).
 
-MoqTransport accomplishes this via the GOAWAY message ({{message-goaway}}). The
-server sends a GOAWAY message which signals the client to establish a new
-session. The GOAWAY message MAY contain a URI for the new session, useful in
-some load balancing schemes, otherwise the current URI is reused.
-The server SHOULD terminate the session with GOAWAY Failure after
-a sufficient timeout, otherwise a misbehaving client could prevent a drain.
+The server sends a GOAWAY message, signaling that the client should establish a
+new session and migrate any active subscriptions. The GOAWAY message may contain
+a new URI for the new session, useful in some load balancing schemes, otherwise
+the current URI is reused. The server SHOULD terminate the session with GOAWAY
+Timeout after a sufficient timeout if a stubborn client is preventing a drain.
 
-It's RECOMMENDED that session migration is performed transparently to the
-application. This involves establishing the new session in the background and
-migrating all active subscriptions/announcements from the old session. The
-client terminates the old session with NO\_ERROR once there are no more active
-subscriptions. The client MAY choose to delay if OBJECTs are queued on the
-network, but needs to be prepared to receive a GOAWAY Failure from the server if
-it waits too long.
+The GOAWAY message does not impact subscription state. A subscriber SHOULD
+terminate existing subscriptions while a publisher MAY reject new subscriptions
+while in this draining state. Note that server may be a subscriber, in which
+case it should send a GOAWAY message followed by any UNSUBSCRIBE messages.
 
-The GOAWAY message does not change the state of active subscriptions.
-It is RECOMMENDED that the subscriber close existing subscriptions with an
-UNSUBSCRIBE message and the client close the connection once all subscriptions
-have terminated. An endpoint MAY reject new subscriptions while in the draining
-state.
+The client receives a GOAWAY message and determines when to terminate the
+session. It's RECOMMENDED that the client waits until all active subscriptions
+have terminated and then closes the session with NO\_ERROR. This should be
+performed transparently to the application which involves establishing the new
+session in the background and migrating all active subscriptions/announcements
+from the old session. The client may choose to delay closing the session if
+OBJECTs are queued or inflight, but needs to be prepared to receive a GOAWAY
+Timeout from the server if it takes too long.
+
 
 # Prioritization and Congestion Response {#priority-congestion}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -441,12 +441,12 @@ MAY reject new SUBSCRIBEs while in the draining state. When the server is
 a subscriber, it SHOULD send a GOAWAY message prior to any UNSUBSCRIBE messages.
 
 After the client receives a GOAWAY, it's RECOMMENDED that the client waits until
-there are no more active subscriptions before closing the session with NO\_ERROR.
-Ideally this is transparent to the application, which involves establishing a new
-session in the background and migrating active subscriptions and announcements.
-The client can choose to delay closing the session if it expects more OBJECTs to
-be delivered. The server can send a 'GOAWAY Timeout' if the client doesn't close
-the session quickly enough.
+there are no more active subscriptions before closing the session with NO_ERROR.
+Ideally this is transparent to the application using MOQT, which involves
+establishing a new session in the background and migrating active subscriptions
+and announcements. The client can choose to delay closing the session if it
+expects more OBJECTs to be delivered. The server closes the session with a
+'GOAWAY Timeout' if the client doesn't close the session quickly enough.
 
 
 # Prioritization and Congestion Response {#priority-congestion}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -402,7 +402,9 @@ code, as defined below:
 |------|--------------------|
 | 0x2  | Unauthorized       |
 |------|--------------------|
-| 0x10 | GOAWAY             |
+| 0x3  | Protocol Violation |
+|------|--------------------|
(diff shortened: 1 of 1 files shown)


kixelated:
IMO I would remove the `### Graceful` section. I think we should have this recommended algorithm somewhere so people can build gud clients, but I'm not sure it needs to be in the draft. Let me know what you think.

draft-ietf-moq-transport.md
```diff
@@ -413,8 +415,87 @@ terminate the session.
...
+suggested algorithm.
+
+If an endpoint does not support graceful migration, it SHOULD terminate the
+session immediately with a GOAWAY Failure error code ({{session-termination}}).
+Additionally, the server SHOULD terminate the session with GOAWAY Failure after
+a sufficient timeout, otherwise a misbehaving client could prevent the drain.
+The GOAWAY Failure error SHOULD be exposed to the application so it has the
+option to perform retry logic.
+
+### Graceful {#session-migration-graceful}
```
kixelated, line 448: IMO I would remove this entire section and the publisher/subscriber subsections. I think we should have this recommended algorithm _somewhere_ so people can build gud clients, but I'm not sure it needs to be in the draft.

What do you think?

draft-ietf-moq-transport.md
```diff
@@ -413,8 +415,87 @@ terminate the session.
...
+The client terminates the old session once there are no more active
+subscriptions or announcements. The client MAY choose to delay if OBJECTs are
+still being sent/received as queues are drained, but should be prepared to
+receive a GOAWAY Failure from the server if it waits too long.
+
+#### Publisher
+A publisher needs to migrate all active announcements to the new session. For
+each active announcement:
+
+1. The publisher sends an ANNOUNCE on the new session.
```
afrind, line 474: Part of this depends on #227, how will the relay (network) handle a duplicate announcements.

draft-ietf-moq-transport.md
```diff
@@ -967,30 +1048,24 @@ UNANNOUNCE Message {
...
+({{session-migration}}) with an optional URI.
+
+The server MUST terminate the session with a Protocol Violation
+({{session-termination}}) if it receives a GOAWAY message. The client MUST
+terminate the session with a Protocol Violation ({{session-termination}} if it
+receives multiple GOAWAY messages.
+
+~~~
+GOAWAY Message {
+  Session URI (b)
```
afrind, line 1061: Maybe `New Session URI`

draft-ietf-moq-transport.md
```diff
@@ -413,8 +415,87 @@ terminate the session.
...
+   the maximum received object sequence.
+2. The client sends an UNSUBSCRIBE message on the old session.
+3. The client receives a SUBSCRIBE\_OK on the new session.
+4. The client receives a SUBSCRIBE\_ERROR on the old session, acknowledging the
+   UNSUBSCRIBE.
+
+The SUBSCRIBE and UNSUBSCRIBE are sent in parallel to avoid underutilizing the
+network, however this results in overutilizing the network for at least an RTT.
+The endpoint may receive duplicate OBJECT messages across both sessions which
+MUST be discarded.
```
afrind, line 498: I'm not sure you need this MUST, probably `the endpoint 'can' discard any redundant objects` is sufficient.

afrind (commented):
Individual review.

Some high level questions:

1. Should we make it clear that OBJECTs on existing subscriptions continue flowing until an UNSUBSCRIBE, the subscription expires, or the session closes (eg: GOAWAY does not change any subscription states from either peer)?
2. If the session closes and the peer is left with a partial OBJECT - do they need to discard it?
3. When can the sender of GOAWAY close the connection gracefully - as soon as all subscriptions and announces have been terminated?
4. Do we need client initiated GOAWAY - to ask the server to UNSUBSCRIBE gracefully, or should the client send the soon to be "SUBSCRIPTION END" with a code like "going away" instead?
5. Do we need to make explicit that SUBSCRIBE/ANNOUNCE can race with a peer GOAWAY, so the GOAWAY sender should SUBSCRIBE ERROR/ANNOUNCE ERROR with a code like "going away"?  There's no implicit cancellation like in HTTP/[2,3].

draft-ietf-moq-transport.md
```diff
@@ -413,8 +415,87 @@ terminate the session.
...
+session and migrate any active subscriptions. The GOAWAY message MAY contain a
+URI for the new session, useful in some load balancing schemes, otherwise the
+current URI is reused.
+
+It's RECOMMENDED that session migration is performed transparently to the
+application. See the following section ({{session-migration-graceful}}) for a
+suggested algorithm.
+
+If an endpoint does not support graceful migration, it SHOULD terminate the
+session immediately with a GOAWAY Failure error code ({{session-termination}}).
```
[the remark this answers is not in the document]
kixelated: As a server operator, it would be have to have some insight into how many clients failed due to not supporting GOAWAY. I imagine it will be more common than HTTP because it's more difficult and there will be more client variety (as opposed to 2 browser implementations).

Upon receiving a GOAWAY, a client would either:
1. Gracefully migrate subscriptions and close the session with error code 0.
2. Ignore the GOAWAY, waiting for the server to terminate the session.
3. Terminate immediately with some error code

That 3rd case worries me a little bit. We could eyeball the number of session terminations around when the GOAWAY was sent. However it would be nice if the draft said either don't do that, or use a specific error code if you did.

draft-ietf-moq-transport.md
```diff
@@ -413,8 +415,87 @@ terminate the session.
...
+
+MoqTransport accomplishes this via the GOAWAY message ({{message-goaway}}). The
+server sends a GOAWAY message which signals the client to establish a new
+session and migrate any active subscriptions. The GOAWAY message MAY contain a
+URI for the new session, useful in some load balancing schemes, otherwise the
+current URI is reused.
+
+It's RECOMMENDED that session migration is performed transparently to the
+application. See the following section ({{session-migration-graceful}}) for a
+suggested algorithm.
```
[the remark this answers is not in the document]
kixelated: Yeah I think we move the recommended algorithm into a Github issue. I mostly wanted to write it down to make sure the mechanisms in MoqT allow for graceful drains (almost), even if we don't document the algorithm itself in the draft.

draft-ietf-moq-transport.md
```diff
@@ -396,25 +396,51 @@ code, as defined below:
...
+| 0x0  | No Error           |
 |------|--------------------|
 | 0x1  | Generic Error      |
 |------|--------------------|
 | 0x2  | Unauthorized       |
 |------|--------------------|
-| 0x10 | GOAWAY             |
+| 0x3  | Protocol Violation |
+|------|--------------------|
+| 0x10 | GOAWAY Failure     |
```
afrind, line 407: Is the only case for this "GOAWAY Drain Timeout" - eg I told you to unsubscribe everything and you didn't do it?  If so, maybe rename.

draft-ietf-moq-transport.md
```diff
@@ -967,30 +993,25 @@ UNANNOUNCE Message {
 
 
 ## GOAWAY {#message-goaway}
+The server sends a `GOAWAY` message to initiate session migration
+({{session-migration}}) with an optional URI.
+
+The server MUST terminate the session with a Protocol Violation
+({{session-termination}}) if it receives a GOAWAY message. The client MUST
```
afrind, line 1000: Another use case for client initiated GOAWAY is a moq intermediary that acts as both a server and a client.  When it wants to go down for maintenance, it may want to ask its upstream to unsubscribe to cleanly terminate.

If you're not sold, maybe we should open an issue to discuss the directionality of GOAWAY.

afrind (commented):
Individual review

I think we want to say something like you MUST NOT send any new SUBSCRIBE/ANNOUNCE messages after receiving a GOAWAY.  If you receive a SUBSCRIBE/ANNOUNCE after sending GOAWAY, reply with an error like "DRAINING" or "REFUSED".

We should clarify if GOAWAY changes the state of any active subscriptions issued by the sender - eg: does GOAWAY implicitly UNSUBSCRIBE or not?

draft-ietf-moq-transport.md
```diff
@@ -396,25 +396,51 @@ code, as defined below:
...
+  migration ({{session-migration}}).
+
+## Migration {#session-migration}
+
+MoqTransport requires a long-lived and stateful session. However, a service
+provider needs the ability to shutdown/restart a server without waiting for all
+sessions to drain naturally, as that can take days for long-form media.
+
+MoqTransport accomplishes this via the GOAWAY message ({{message-goaway}}). The
+server sends a GOAWAY message which signals the client to establish a new
```
ianswett, line 431: I think GOAWAY really means to unsubscribe from all subscriptions on this server and if the client wants to establish a new session, it does so.

ianswett (commented):
Editorial nits, but this LG overall.  I'm inclined to merge this sooner rather than later.

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -396,25 +396,57 @@ code, as defined below:
...
+session. The GOAWAY message MAY contain a URI for the new session, useful in
+some load balancing schemes, otherwise the current URI is reused.
+The server SHOULD terminate the session with GOAWAY Failure after
```
afrind, line 434: GOAWAY Timeout

draft-ietf-moq-transport.md
afrind: Agree - GOAWAY is a signal to UNANNOUNCE/UNSUBSCRIBE.

draft-ietf-moq-transport.md
```diff
@@ -396,25 +396,57 @@ code, as defined below:
...
+a sufficient timeout, otherwise a misbehaving client could prevent a drain.
+
+It's RECOMMENDED that session migration is performed transparently to the
+application. This involves establishing the new session in the background and
+migrating all active subscriptions/announcements from the old session. The
+client terminates the old session with NO\_ERROR once there are no more active
+subscriptions. The client MAY choose to delay if OBJECTs are queued on the
+network, but needs to be prepared to receive a GOAWAY Failure from the server if
```
afrind, line 442: s/Failure/Timeout/

draft-ietf-moq-transport.md
```diff
@@ -396,25 +396,57 @@ code, as defined below:
...
+it waits too long.
+
+The GOAWAY message does not change the state of active subscriptions.
+It is RECOMMENDED that the subscriber close existing subscriptions with an
```
afrind, line 446: Add a sentence about the sender of the GOAWAY also sending UNSUBSCRIBEs?

afrind (commented):
Individual Review

draft-ietf-moq-transport.md
ianswett: I'd like this fixed before merging.

draft-ietf-moq-transport.md
```diff
@@ -393,25 +393,58 @@ code, as defined below:
...
+sessions to drain naturally, as that can take days for long-form media.
+MoqTransport avoids this via the GOAWAY message ({{message-goaway}}).
+
+The server sends a GOAWAY message, signaling that the client should establish a
+new session and migrate any active subscriptions. The GOAWAY message may contain
+a new URI for the new session, useful in some load balancing schemes, otherwise
+the current URI is reused. The server SHOULD terminate the session with GOAWAY
+Timeout after a sufficient timeout if a stubborn client is preventing a drain.
+
+The GOAWAY message does not impact subscription state. A subscriber SHOULD
```
kixelated, line 434: @ianswett @afrind Does this text address your comments?

I decided that the subscriber should close subscriptions (via `UNSUBSCRIBE`). This is tricky because the server might be a subscriber, and yet the client is the one who chooses when to terminate the connection. That's why I think a normative SHOULD is needed, otherwise the client might be waiting indefinitely for the server to terminate subscriptions.

Alternatively, we should switch it such that the client terminates subscriptions. This means it sends `UNSUBSCRIBE` and `SUBSCRIBE_RESET` in response to a `GOAWAY`. I don't think this will be as graceful in the contribution scenario; I would rather have the server `UNSUBSCRIBE` than a sudden `RESET`.

ianswett: I think this works for me, and a SHOULD is fine for now.  One can never enforce GOAWAY behavior anyway.

draft-ietf-moq-transport.md
```diff
@@ -436,13 +436,14 @@ the current URI is reused. The server SHOULD terminate the session with GOAWAY
 Timeout after a sufficient timeout if a stubborn client is preventing a drain.
 
 The GOAWAY message does not impact subscription state. A subscriber SHOULD
-terminate existing subscriptions while a publisher MAY reject new subscriptions
-while in this draining state. Note that server may be a subscriber, in which
-case it should send a GOAWAY message followed by any UNSUBSCRIBE messages.
+individually UNSUBSCRIBE for each existing subscription, while a publisher MAY
+reject new SUBSCRIBEs while in this draining state. Note that server may be a
```
ianswett, line 440:
```suggestion
reject new SUBSCRIBEs while in this draining state. Note that the server may be a
```

PR #271, opened 2023-10-09
Payload to end of stream is a different message type

### LINKED ISSUE #206: Length optimization on message syntax
Since message (6.  Messages) will be the main piece of info of MOQ I think we should try to minimize its size. Assuming most of the messages send to the wire will be objects inside a quic stream, the length should be NOT needed. Could we use 1 bit in "type" to indicate if length is present, we could save some bytes without overcomplicating the protocol

martinduke:
Fixes #206 and #248.

There was consensus in Cambridge that removing the special meaning of zero length gets rid of difficult code, removes potental foot guns, saves a byte, and enables actual zero length payloads, which supposedly has use cases.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -628,9 +628,9 @@ MOQT Message {
 |------:|:---------------------------------------------------|
 | 0x0   | OBJECT without payload length ({{message-object}}) |
 |-------|----------------------------------------------------|
-| 0x1   | OBJECT with payload length ({{message-object}})    |
+| 0x1   | SETUP ({{message-setup}})                          |
 |-------|----------------------------------------------------|
-| 0x2   | SETUP ({{message-setup}})                          |
+| 0x2   | OBJECT with payload length ({{message-object}})    |
 |-------|----------------------------------------------------|
 | 0x3   | SUBSCRIBE REQUEST ({{message-subscribe-req}})      |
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -623,31 +623,33 @@ MOQT Message {
...
-| 0x9   | UNANNOUNCE  ({{message-unannounce}})             |
-|-------|--------------------------------------------------|
-| 0x10  | GOAWAY ({{message-goaway}})                      |
-|-------|--------------------------------------------------|
-| 0xA   | UNSUBSCRIBE ({{message-unsubscribe}})            |
-|-------|--------------------------------------------------|
+|-------|----------------------------------------------------|
+| ID    | Messages                                           |
+|------:|:---------------------------------------------------|
+| 0x0   | OBJECT without payload length ({{message-object}}) |
```
ianswett, line 629: We've been trying to avoid renumbering existing messages, so can you give this the next available codepoint?

ianswett (approved):
LG, but I think we'd like to avoid changing the existing codepoints.

draft-ietf-moq-transport.md
ianswett, line 629: In the interest in not making changes to existing types, can this be OBJECT with payload length, as it is today?

ianswett (approved):

PR #277, opened 2023-10-16
Explicitly specify the namespace

### LINKED ISSUE #204: SUBSCRIBE REQUEST ambiguity
Assume a setup with two publishers, a single relay and one subscriber.

P1 ANNOUNCE's track namespace fooba
P2 ANNOUNCE's track namespace foo

S1 sends a SUBSCRIBE REQUEST for Full Track Name `foobar`

How does the relay know whether `foobar` is Track Namespace: `foo` and Track Name: `bar`, or Track Namespace: `fooba` and Track Name `r`, and hence which publisher to route the subscribe to? 

I see a few possibilities:

1) The relay should not allow this scenario (via namespace construction and/or authorization).  If so the draft should include text explaining that this is disallowed (normatively?) or "bad things(tm)" happen.

2) Change SUBSCRIBE REQUEST to be able to disambiguate the Track Namespace from the Track Name.  This could be done by sending name and namespace in separate message fields, or having stricter rules for the structure of Full Track Name such that a relay can extract the constituent parts.

3) Explicitly state that Full Track Names are matched to namespaces via "longest prefix match".

ianswett:
Fixes #204 by explicitly specifying the namespace.

Also consistently changes to the (b) notation to indicate a variable length sequence of bytes.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -802,7 +802,7 @@ The format of SUBSCRIBE REQUEST is as follows:
 
 ~~~
 SUBSCRIBE REQUEST Message {
-  Namespace (b),
+  Track Namespace (b),
   Track Name (b),
   Number of Parameters (i),
   Track Request Parameters (..) ...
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -802,16 +802,17 @@ The format of SUBSCRIBE REQUEST is as follows:
 
 ~~~
 SUBSCRIBE REQUEST Message {
-  Full Track Name Length (i),
-  Full Track Name (...),
+  Namespace (b),
```
afrind, line 805: Track Namespace?

kixelated (approved):

afrind:
This PR got lost when merging 289.

PR #278, opened 2023-10-16
Start sending incomplete objects

### LINKED ISSUE #205: Relay send while receiving
Should we mention that a relay should start sending object as soon as the object can be identified (after the header)?

Since the draft points out latency as a big deal, it seems this is a must have if we want to cover ULL cases

ianswett:
Fixes #205

Also, can someone explain what the following paragraph is saying, because I'm having a difficult time parsing it.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -600,13 +600,15 @@ combine, split, or otherwise modify object payloads.  A relay SHOULD
 prioritize streams ({{priority-congestion}}) based on the send
 order/priority.
 
+A sender SHOULD begin sending incomplete objects when available to
+avoid incurring additional latency.
+
 A relay that reads from a stream and writes to stream in order will
 introduce head-of-line blocking.  Packet loss will cause stream data to
 be buffered in the QUIC library, awaiting in order delivery, which will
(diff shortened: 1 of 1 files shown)


afrind:
The next paragraph is trying to say that if you receive HoL blocked stream data at a relay you should try to forward what you have.  The goal is to mitigate accumulating latency at every relay hop.

I think this makes sense in two cases:

You receive data that you know to be within an object you have already seen the header for.  Eg: the object length is 100 and you miss bytes 0-49 but get 50-99, you can forward 50-99, as long as your QUIC library supports out of order reads and writes to a stream (not many presently do).  This also applies when the object has no length and occupies the entire stream.

You are missing some object data for object N on a stream, but you receive the header for object N + 1.  This is (potentially) less problematic since the relay might be able to open a new stream to forward it, so out-of-order stream write APIs are not required.

In any case, the text probably needs clarification, and perhaps moving from SHOULD (but we know you won't) to MAY is warranted?

fluffy (approved):
sooner is better

I do think this might be interpreted by some people in an over zealous sort of way ( like turning off nagle type things ). I do wonder about just using MAY. But happy either way. SHOULD and MAY mean the same thing.

PR #279, opened 2023-10-16
Resetting Streams doesn't affect MoQ

### LINKED ISSUE #202: Section 3.3: Cancellation
Section 3.3 currently says:

"3.3. Cancellation

   A QUIC stream MAY be canceled at any point with an error code.  The
   producer does this via a RESET_STREAM frame while the consumer
   requests cancellation with a STOP_SENDING frame.

   When using order, lower priority streams will be starved during
   congestion, perhaps indefinitely.  These streams will consume
   resources and flow control until they are canceled.  When nearing
   resource limits, an endpoint SHOULD cancel the lowest priority stream
   with error code 0.

   The sender MAY cancel streams in response to congestion.  This can be
   useful when the sender does not support stream prioritization.

   TODO: this section actually describes stream cancellation, not
   session cancellation.  Is this section required, or can it be
   deleted, or added to a new "workflow" section."

[BA]  While the meaning of STOP_SENDING is clear for frame/stream transport
(e.g. stop sending the frame whose transmission is currently in progress),
it is less clear when multiple frames are transported on the same QUIC
stream.

In the case where multiple frames are being transmitted on a stream, the
problem is that STOP_SENDING does not indicate *which* frame the receiver
has given up on.  So how is the sender to interpret it?  For example, after
resetting the stream, does the sender open a new stream?  If so, what
frame(s) are to be transmitted on that new stream?

To address this ambiguity, an application-layer message(s) could be
defined, indicating exactly what the receiver wants the sender to do:

* Stop (re-)transmission of a particular (named) frame.  The existing
stream is reset and a new stream is opened for transmission of subsequent
frames.

* Stop transmission of layer(s). The receiver wishes to the sender to
adjust the operating point, by stopping the transmission of one or more
temporal or spatial layers.  Only the layers that remain will be
transmitted on new stream(s).

Mailing list thread: https://mailarchive.ietf.org/arch/msg/moq/1Zi-IO12grZpEkrYw5c8aXyK9-M/

ianswett:
Fixes #202

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -374,10 +374,10 @@ results in the session being closed as a 'Protocol Violation'.
 
 ## Stream Cancellation
 
-A QUIC stream MAY be canceled due to congestion or other reasons by either
-the sender or receiver. Early termination of a QUIC stream does not
-affect the MoQ application state, and therefore has no effect on outstanding
-subscriptions.
+QUIC streams aside from the control stream MAY be canceled due to congestion
+or other reasons by either the sender or receiver. Early termination of a
+QUIC stream does not affect the MoQ application state, and therefore has no
+effect on outstanding subscriptions.
 
 ## Termination  {#session-termination}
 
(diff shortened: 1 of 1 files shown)


fluffy (approved):
Might want to make it clear this does not apply to control streams

fluffy (approved):
Like the change.

PR #280, opened 2023-10-16
Define a control stream

### LINKED ISSUE #31: Control streams
As mentioned in #29, there's currently no distinction between control streams and media streams.

Should we have dedicated control streams? When should these be cached or forwarded by relays?

For example, GOAWAY must NOT be forwarded by a relay, but something like TRACK (does not exist yet) should be forwarded.

### LINKED ISSUE #137: Control message ordering
The current draft specifies that control messages may be sent over separate streams. This means that control messages will arrive in an unknown order.

This can create divergent behavior if messages depend on any form of ordering. The simplest example if sending a SUBSCRIBE message followed by a later UNSUBSCRIBE message on a separate stream. If the UNSUBSCRIBE arrives first for whatever reason (packet loss or poor prioritization), then it has a completely different meaning.

The current draft adds some advisory text that control messages should be on the same stream if ordering is required. I think we should put all control messages on the same stream, so ordering is enforced and the behavior is deterministic.

### LINKED ISSUE #138: Proposal: Control and data streams

### LINKED ISSUE #230: Resetting control streams
There's an assumption in the current draft that some messages are partially reliable (OBJECT messages) while some are fully reliable (everything else, aka control messages). The ability to drop a message depend on the ability to reset the QUIC stream.

However, reseting a stream containing control messages wrecks havok on the subscription state machine. There's no way within QUIC determine if a messages was received or processed prior to a reset. For example, if I send a SUBSCRIBE on stream and then a receive a STOP_SENDING for that stream, then the subscription enters a limbo state:

1. The SUBSCRIBE was processed and a SUBSCRIBE_OK or SUBSCRIBE_ERROR is forthcoming.
2. The SUBSCRIBE was never processed and a SUBSCRIBE_OK or SUBSCRIBE_ERROR will never be received.

The only option currently is to set a timer after a reset is received. After a few RTTs or so, we assume the SUBSCRIBE was lost and issue a new SUBSCRIBE on a new stream, introducing latency in the process. Though things get extra weird if we receive a response after this timeout expires...

---

I think we either need to state:
1. Control streams MUST NOT be reset (like HTTP/3)
2. Reseting a control stream also resets all prior messages (ie. all prior SUBSCRIBEs).
3. Reseting a control stream resets all unacknowledged messages (ie. SUBSCRIBEs without SUBSCRIBE_OK).

Option 3 is problematic, since the subscriber knows that the OK was received prior to the reset but the publisher doesn't. Personally, I think option 1 is the best is there's a single control stream, while option 2 is the best if there's multiple (scoped) control streams. 

---

Regardless of the approach, data messages and control messages should be split (#31) because they have different reliability. A receiver should know if it can issue STOP_SENDING for an OBJECT without impacting control messages, especially if we take approach 1.

My application uses a bidirectional stream for control messages and unidirectional streams for data messages. It makes it very easy for the application to know if a stream is resettable just based on the ID.

ianswett:
Fixes #230
Fixes #138
Fixes #137
Fixes #31

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -364,6 +364,11 @@ a control message on a different stream closes the session as a
 peer receiving an Object on the control stream closes the session as a
 'Protocol Violation'.
 
+This draft only specifies a single use of bidirectional streams. Objects are
+sent on unidirectional streams.  Because there are no other uses of
+bidirectional streams, a peer MAY currently close the connection if it
+receives a second bidirectional stream.
+
 The control stream MUST NOT be abruptly closed at the QUIC layer.  Doing so
 results in the session being closed as a 'Protocol Violation'.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -352,17 +352,20 @@ PATH parameter ({{path}}) which is sent in the SETUP message at the
 start of the session.  The ALPN value {{!RFC7301}} used by the protocol
 is `moq-00`.
 
+
 ## Session initialization {#session-init}
 
-The first stream opened is a client-initiated bidirectional stream where
-the peers exchange SETUP messages ({{message-setup}}). The subsequent
-streams MAY be either unidirectional or bidirectional. For exchanging
-content, an application would typically send a unidirectional stream
-containing a single OBJECT message ({{message-object}}), as putting more
-than one object into one stream may create head-of-line blocking delays.
-However, if one object has a hard dependency on another object, putting
-them on the same stream could be a valid choice.
+The first stream opened is a client-initiated bidirectional control stream
+where the peers exchange SETUP messages ({{message-setup}}).  All messages
+defined in this draft are sent on the control stream after the SETUP message.
+Control messages MUST NOT be sent on any other stream, and a peer receiving
+a control message on a different stream closes the session as a
+'Protocol Violation'. Objects MUST NOT be sent on the control stream, and a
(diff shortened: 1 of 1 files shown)


kixelated (approved):
I would add a note that data streams are sent on unidirectional streams, but looks gud.

fluffy (approved):
LGTM

PR #283, opened 2023-10-16
MoQ has no TCP/etc fallback

### LINKED ISSUE #88: Failover strategy when QUIC is unavailable
Issue raised at interim on Feb 1, 2023:

Does the protocol need to define a failover if QUIC is not available on a specific path.  If so, is it the same failover in all cases?

ianswett:
Fixes #88

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -319,7 +319,7 @@ semantics (see {{?I-D.ietf-webtrans-overview, Section 4}}); thus, the
 main difference lies in how the servers are identified and how the
 connection is established.  There is no definition of the protocol
 over other transports, such as TCP, and applicaitons using MoQ might
-need to fallback to another protocol when QUIC and WebTransport aren't
+need to fallback to another protocol when QUIC or WebTransport aren't
 available.
 
 ### WebTransport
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -317,7 +317,10 @@ over a QUIC connection directly [QUIC], and over WebTransport
 [WebTransport].  Both provide streams and datagrams with similar
 semantics (see {{?I-D.ietf-webtrans-overview, Section 4}}); thus, the
 main difference lies in how the servers are identified and how the
-connection is established.
+connection is established.  There is no definition of the protocol
+over other transports, such as TCP, and applicaitons using MoQ might
+need to fallback to another protocol when QUIC and WebTransport aren't
```
kixelated, line 322: > when QUIC and WebTransport

should be:

> when QUIC or WebTransport

kixelated (approved):

vasilvv (approved):

draft-ietf-moq-transport.md
ianswett, line 322:
```suggestion
need to fallback to another protocol when QUIC or WebTransport aren't
```

ianswett: SG, feel free to leave a suggestion I can just commit it for editorial suggestions like this.

PR #285, opened 2023-10-16
Change message names to use _

### LINKED ISSUE #275: "SUBSCRIBE OK" and "SUBSCRIBE ERROR" or SUBSCRIBE_OK and SUBSCRIBE_ERROR
I'd make this change in a single PR, but I prefer frame names have underscores to having to quote them when referring to them when they are two words.

Observed when reviewing #274

ianswett:
Fixes #275

Also renames SUBSCRIBE to SUBSCRIBE_REQUEST, since it was already being referred to as SUBSCRIBE in multiple cases and it's parallel to ANNOUNCE.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -562,10 +562,10 @@ Relays MUST ensure that publishers are authorized by:
   the way the relay is managed and is application specific (typically
   based on prior business agreement).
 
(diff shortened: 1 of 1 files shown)


kixelated (approved):
+1 to SUBSCRIBE instead of SUBSCRIBE_REQUEST

PR #287, opened 2023-10-16
Protocols directly over UDP

### LINKED ISSUE #73: QUIC is over UDP
In Sec 3.1 it says "This is possible with the emergence of QUIC, designed to fix the shortcomings of TCP" in the context of TCP and UDP-based protocols.

Pedants will note that QUIC is over UDP. It is probably worth to add something to the effect of:

"While QUIC is also over UDP, it frees the media protocol from handling functions that might otherwise be handled by TCP."

ianswett:
Fixes #73

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -104,8 +104,8 @@ deliver content encoded at a higher bitrate than the network can support
 causes queuing along the path from producer to consumer. The speed at
 which a protocol can detect and respond to queuing determines the
 overall latency. TCP-based protocols are simple but are slow to detect
-congestion and suffer from head-of-line blocking. UDP-based protocols
-can avoid queuing, but the application is now responsible for the
+congestion and suffer from head-of-line blocking. Protocols directly over
+UDP can avoid queuing, but the application is now responsible for the
 complexity of fragmentation, congestion control, retransmissions,
 receiver feedback, reassembly, and more. One goal of MOQT is to achieve
 the best of both these worlds: leverage the features of QUIC to create a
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -104,8 +104,8 @@ deliver content encoded at a higher bitrate than the network can support
 causes queuing along the path from producer to consumer. The speed at
 which a protocol can detect and respond to queuing determines the
 overall latency. TCP-based protocols are simple but are slow to detect
-congestion and suffer from head-of-line blocking. UDP-based protocols
-can avoid queuing, but the application is now responsible for the
+congestion and suffer from head-of-line blocking. Protocols directly over
```
kixelated, line 107: Is there a better way to phrase this?

martinduke: There are many ways to say this. The language in this PR is fine, but some alternatives...

"UDP-based protocols aside from QUIC..."

"Applications that use UDP but not QUIC avoid queuing, but are now responsible..."

Whatever people like is fine.

martinduke (approved):

draft-ietf-moq-transport.md
ianswett: I don't like UDP-based protocols aside from QUIC, because one can run SCTP or another transport over UDP, and those have their own properties.

kixelated (approved):
"Protocols directly utilizing UDP" would be my vote but anything works.

martinduke:
I am fine with any phrasing that makes it clear that an intervening full-featured transport over UDP negates the issue.

PR #246, opened 2023-09-11
Experimental Extensions - Transport Delivery Mode and Subscription Hints

suhasHere:
This is a placeholder PR to support the issues #244 and #245 .  

All the discussions have to be done on those above issues listed and the suggestions/outcoms from the discussions are to be reflected in this PR.

PR #289, opened 2023-10-17
Add subscribe hints

### LINKED ISSUE #111: Objects can be individually cached and served (A way to GET one or more objects)
Some use-case like  catchup , go back to a given object for layer refresh or server the game state from a given object , and similar scenarios ,  would need a name to ask for the object. 
It would be good for the protocol to be able to subscribe to not just track and group, but also object.

### LINKED ISSUE #245: Subscription Hints
Subscribers needs a way to hint the publishers on when do they have the want the objects to delivered to them within the requested track. This is needed in case due to the fact that different media types can be handled differently by the application ( audio can be delivered to the application immediately,  video needs some form of key-frame to have the best experience or something more specific that the application wants )

In that effect , i want to propose an **experimental** extension to the Track Request Parameters called "SubscriptionHint" with the following variants.

* latest (0):
The "latest" subscription hint identifies delivery of the objects for the
requested track to begin from the most recent group and object sequence.

* catchup (1):
The "catchup" subscription hint identifies delivery of the objects for the
requested track from the beginning of the group provided in the GroupSequence track request param. 

* waitup(2): 
The "waitup" subscription hint identifies delivery of the objects for the
requested track  from the beginning of the new group.

### LINKED ISSUE #260: Subscription requests and buffers
Currently, we have GROUP SEQUENCE and OBJECT SEQUENCE parameters that allow subscribing to a track at a specific offset.  As far as I can tell, we do not actually describe what happens when those are not present.  I imagine what most subscribers would want is to receive most recent N groups when they subscribe (for some subscriber-specified N that is determined by how large they want their buffer to be), and we should add a parameter for the said N.

suhasHere:
This is a initial PR to get some basic definition of subscription hints. I would like to get us to an agreement for the basic proposal and then discuss further on error codes, edge cases and impacts/interactions with other messages (Sub Fin, RST, UPD)

Fixes #111 
Fixes #245 
Fixes #260

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1030,7 +1030,7 @@ The  `Payload` is specific to the chosen `HintType` as descirbed in the followin
 
 ### StartPoint Hint
 
-The `StartPoint` subscription hint identifies a starting point within the track for delivering objects. The requested start point specified is relative to publisher's current view of the state within the track. A publisher's current view of the track is defined as the most recent group and object for the track received (or in the cache), if available, at the time of request.
+The `StartPoint` subscription hint identifies a starting point within a track for delivering objects. The requested start point specified is relative to publisher's view  of the ongoing state of the track. A publisher's current state of the track is defined by the most recent group and object sequence received (or active in the cache), if available, at the time of request.
 
 The payload for StartPoint hint has the following structure:
 
@@ -1043,13 +1043,13 @@ StartPoint Payload {
 
 A subscriber can request the starting point to be one of the following values, as idenitfied by the `Mode` parameter:
 
-* Current: Has value of `0x1` and specifies delivering objects from the beginning of the current group of objects. The optional parameter `GroupCount`, if specified, MUST be ignored when processing this hint. 
+* PREVIOUS: Start point for object delivery is from an earlier group relative to the current group. The optional parameter `GroupCount` MUST be specified and it specifies the number of groups to go back from the current group to determine the start point. It has value of `0x1`.
 
-* Previous: Has value of `0x2` and specifies delivering objects from earlier group(s) relative to the current group. The optional parameter `GroupCount` MUST be specified and it specifies the number of groups to go back from the current group to determine the start point.
+* NEXT: Start point for object delivery is set to a future group relative to the current group. The optional parameter `GroupCount` MUST be specified and it specifies the number of groups to wait on before delivering the objects. It has value of `0x2`.
 
-* Next: Has value of `0x3` and specifies delivering objects from future group(s). The optional parameter `GroupCount` MUST be specified and it specifies the number of groups, from the current group, to wait on before delivering the objects.
+* CURRENT: Start point for object delivery is set to beginning of the current group. The optional parameter `GroupCount`, if specified, MUST be ignored when processing this hint. It has value of `0x3`.
 
-* Now: Has value of `0x4` and specifies start delivering from the current group and the object sequence. The optional parameter `GroupCount`, if specified, MUST be ignored when processing the hint.
+* NOW: Start point for object delivery is set to most recent obect from the current group. The optional parameter `GroupCount`, if specified, MUST be ignored when processing the hint. It has value of `0x4`
 
 
 Publishers processing the StartPoint hint needs to take following into consideration:
@@ -1066,7 +1066,7 @@ TODO: Add a note in the security consideration section.
 
 ### Interval Hint
 
-The `Interval` subscription hint allows subscribers to request for  range of objects by specifying values pertaining to start group/object and end group/object sequences. 
+The `Interval` subscription hint allows subscribers to request for range of objects by specifying values pertaining to start and end group/object sequences. 
 
 The payload for Interval hint has the following structure:
 
@@ -1080,7 +1080,7 @@ Interval Payload {
 }
 ~~~
 
-The `StartGroupSequence` and `StartObjectSequence` parameters specifies the starting group and object sequence from where the objects within a tracks needs to be delivered. The `EndGroupSequence` and `EndObjectSequence` parameters identify the ending group and object sequence values respectively. Such requests are "closed/bounded interval" requests and subscription is terminated when the objects defined by the requested range are published. if the requested range cannot be satisfied, publishers MUST send SUBSCRIBE ERROR message with Error Code set to TODO.
+The `StartGroupSequence` and `StartObjectSequence` parameters specify the start group and object sequence of the interval and   `EndGroupSequence` and `EndObjectSequence` parameters capture the ending group and object sequence values respectively. Such requests are "closed/bounded interval" requests with the subscription terminated when the objects defined by the requested range are published. If the requested range cannot be satisfied, publishers MUST send SUBSCRIBE ERROR message with Error Code set to TODO.
 
 Additionally subscribers can include only the `StartGroupSequence` and `StartObjectSequence` parameters to indicate "open/unbounded interval" requests, wherin only the absolute start point for object delivery is specified. If either of the  `StartGroupSequence` and `StartObjectSequence` cannot be satisfied, publishers MUST respond with SUBSCRIBE ERROR message with Error Code set to TODO.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1073,10 +1073,10 @@ The payload for Interval hint has the following structure:
 ~~~
 
 Interval Payload {
-  StartGroupSequence,
-  StartObjectSequence,
-  [EndGroupSequence],
-  [EndObjectSequence]
+  StartGroupSequence(i),
+  StartObjectSequence(i),
+  [EndGroupSequence(i)],
+  [EndObjectSequence(i)]
 }
 ~~~
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -797,6 +797,7 @@ The format of SUBSCRIBE REQUEST is as follows:
 SUBSCRIBE REQUEST Message {
   Full Track Name Length (i),
   Full Track Name (...),
+  Number of Hints (i),
   Subscription Hints (..) ...
   Track Request Parameters (..) ...
 }
@@ -1005,37 +1006,40 @@ Subscription Hints have the following structure:
 
 ~~~
 SUBSCRIPTION HINT {
-  HintType (8),
-  Payload (b)
+  HintType (i),
+  Value (b)
 }
 ~~~
 {: #moq-transport-subscription-hint format title="MOQT Subscription HInt"}
 
 The `HintType` parameter identifies one of the following values:
 
-
 |------------|-------------------------|
 | Hint Type  | Hint                    |                               
 |-----------:|:------------------------|
-| 0x0        | Reserved                |
+| 0x0        | RelativeStartPoint      |
 |------------|-------------------------|
-| 0x1        | StartPoint              |
+| 0x1        | AbsoluteStartPoint      |
 |------------|-------------------------|
 | 0x2        | Interval                |
 |------------|-------------------------|
 | 0x3 - 0xFF | Reserved for future use |
 |------------|-------------------------|
 
-The  `Payload` is specific to the chosen `HintType` as descirbed in the following subsections.
+The `Value` is specific to the chosen `HintType` as descirbed in the following subsections.
 
-### StartPoint Hint
+TODO: Define registry for specifying the hints.
 
-The `StartPoint` subscription hint identifies a starting point within a track for delivering objects. The requested start point specified is relative to publisher's view  of the ongoing state of the track. A publisher's current state of the track is defined by the most recent group and object sequence received (or active in the cache), if available, at the time of request.
+TODO: Define experimental range.
 
-The payload for StartPoint hint has the following structure:
+### RelativeStartPoint Hint {#relative}
+
+The `RelativeStartPoint` subscription hint identifies a starting point relative to publisher's view of the ongoing state of the track for delivering objects. A publisher's current state of the track is defined by the most recent group and object sequence received (or active in the cache), if available, at the time of request.
+
+The structure for the  RelativeStartPoint hint value is as follows:
 
 ~~~
-StartPoint Payload {
+RelativeStartPoint Value {
   Mode (i),
   [GroupCount (i)]
 }
@@ -1043,46 +1047,57 @@ StartPoint Payload {
 
 A subscriber can request the starting point to be one of the following values, as idenitfied by the `Mode` parameter:
 
-* PREVIOUS: Start point for object delivery is from an earlier group relative to the current group. The optional parameter `GroupCount` MUST be specified and it specifies the number of groups to go back from the current group to determine the start point. It has value of `0x1`.
+* PREVIOUS: Start point for object delivery is from an earlier group relative to the current group. The optional parameter `GroupCount` MUST be specified and it specifies the number of groups to go back from the current group to determine the start point. The mode `PREVIOUS` has value of `0x1`.
 
-* NEXT: Start point for object delivery is set to a future group relative to the current group. The optional parameter `GroupCount` MUST be specified and it specifies the number of groups to wait on before delivering the objects. It has value of `0x2`.
+* NEXT: Start point for object delivery is set to a future group relative to the current group. The optional parameter `GroupCount` MUST be specified and it specifies the number of groups to wait on before delivering the objects. The mode `NEXT` has value of `0x2`.
 
-* CURRENT: Start point for object delivery is set to beginning of the current group. The optional parameter `GroupCount`, if specified, MUST be ignored when processing this hint. It has value of `0x3`.
+* CURRENT: Start point for object delivery is set to beginning of the current group. The optional parameter `GroupCount` MUST NOT be present. The mode `CURRENT` has value of `0x3`.
 
-* NOW: Start point for object delivery is set to most recent obect from the current group. The optional parameter `GroupCount`, if specified, MUST be ignored when processing the hint. It has value of `0x4`
+* NOW: Start point for object delivery is set to most recent obect from the current group. The optional parameter `GroupCount` MUST NOT be present. The mode `NOW` has value of `0x4`
 
 
-Publishers processing the StartPoint hint needs to take following into consideration:
-
-- If the StartPoint mode is not understood, the publisher MUST send SUBSCRIBE ERROR message with an appropriate error code (TODO: need to define this).
-
-- If the GroupCount cannot be satisfied, the publisher MUST send SUBSCRIBE ERROR message with an appropriate error code (TODO: need to define this).
-
-- If the GroupCount can be satisfied, the publisher SHOULD respond with SUBSCRIBE OK and start delivering the objects.
-
+If the RelativeStartPoint mode is not understood, the publisher MUST send SUBSCRIBE ERROR message with an appropriate error code (TODO: need to define this).
 
 TODO: Add a note in the security consideration section.
 
+
+### AbsoluteStartPoint Hint {#absolute}
+
+The `AbsoluteStartPoint` subscription hint allows subscribers to specify an absolute point in the track to start delivering objects, as indicated by the `TrackOffset`.
+
+The structure for AbsoluteStartPoint hint value is as follows:
+
+~~~
+AbsoluteStartPoint Value {
+  TrackOffset start
+}
+~~~
+
+* TrackOffset:  Identifies the group and optionally, the object sequence value within the track as the start point for the delivery. TrackOffset is defined as below.
+
+~~~
+TrackOffset {
+  Type (i),  // 0 = Group, 1 = Group and Object
+  GroupSequence (i),
+  [ObjectSequence (i)]
+}
+~~~
+
+ The `Type` controls if both group and object sequences are provided. A value of 0x0 indicates only the group sequence is provided and a value of 0x1 indicates both the group and object 
+ sequence values are provided.
 
 ### Interval Hint
 
 The `Interval` subscription hint allows subscribers to request for range of objects by specifying values pertaining to start and end group/object sequences. 
 
-The payload for Interval hint has the following structure:
+The value for Interval hint has the following structure:
 
 ~~~
-
 Interval Payload {
-  StartGroupSequence(i),
-  StartObjectSequence(i),
-  [EndGroupSequence(i)],
-  [EndObjectSequence(i)]
+  TrackOffset start,
+  TrackOffset end
 }
 ~~~
-
-The `StartGroupSequence` and `StartObjectSequence` parameters specify the start group and object sequence of the interval and   `EndGroupSequence` and `EndObjectSequence` parameters capture the ending group and object sequence values respectively. Such requests are "closed/bounded interval" requests with the subscription terminated when the objects defined by the requested range are published. If the requested range cannot be satisfied, publishers MUST send SUBSCRIBE ERROR message with Error Code set to TODO.
-
-Additionally subscribers can include only the `StartGroupSequence` and `StartObjectSequence` parameters to indicate "open/unbounded interval" requests, wherin only the absolute start point for object delivery is specified. If either of the  `StartGroupSequence` and `StartObjectSequence` cannot be satisfied, publishers MUST respond with SUBSCRIBE ERROR message with Error Code set to TODO.
 
 ## Track Request Parameters {#track-req-params}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -810,6 +810,10 @@ SUBSCRIBE REQUEST Message {
 On successful subscription, the publisher SHOULD start delivering
 objects from the group sequence and object sequence described below.
 
+====================
+
+PROPOSAL 1: 6 Track Request Parameters as discussed in the editors call
+
 Several track request parameters to SUBSCRIBE_REQUEST control the start and
 optional end of the subscription within the track.  For a live track, the
 largest group sequence is called the `Current Group` and the largest object
@@ -917,6 +921,46 @@ Parameters:
 ~~~
 
 TODO: Security Considerations related to these hints
+
+====================
+
+PROPOSAL 2: Simpler, trying to follow discussion on the PR through the day.
+
+SUBSCRIBE_REQUEST defines a Track Request Parameter START_POINT (0x0) whose
+value is a varint indicating where in the track to start the subscription.  For
+a live track, the largest group sequence is called the `Current Group` and the
+largest object sequence in that group is the `Current Object`.
+
+|-----------|-------|-------------|--------------|
+| Name      | Value | Start Group | Start Object |
+|----------:|:------|-------------|--------------|
+| Current   | 0x0   | Current     | 0            |
+| Now       | 0x1   | Current     | Current + 1  |
+| Previous  | 0x2   | Current - 1 | 0            |
+| Next      | 0x3   | Curernt + 1 | 0            |
+|------------------------------------------------|
+
+SUBSCRIBE_REQUEST also defines a Track Request Parameter RANGE (0x2) for
+requesting a closed range of objects.  The value is as follows:
+
+~~~
+Range Payload {
+  Start Group Sequence(i),
+  Start Object Sequence(i),
+  End Group Sequence(i),
+  End Object Sequence(i)
+}
+~~~
+
+Exactly one of START_POINT and RANGE MUST be specified.  If neither or both are
+present it is a Protocol Violation.  All Group and Object Sequence values are
+absolute values within the requested track.  The subscription is for all objects
+starting from the start object up to but not including the end object.
+
+If a publisher cannot satisfy the requested start or end for the subscription it
+MAY send a SUBSCRIBE_ERROR with code TBD.
+
+====================
 
 TODO: Issues related to more than one concurrent subscribe to the same track
 
@@ -1172,6 +1216,8 @@ leading varint in catalog track objects.
 --- back
 
 # Pseudo-code for Interpreting Subscribe Track Request Parameters
+
+From PROPOSAL 1:
 
 ~~~
 def get_start_object_and_group(params):
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -212,8 +212,7 @@ This document also defines an additional field type for binary data:
 
 x (b):
 : Indicates that x consists of a variable length integer, followed by
-  that many bytes of binary data. For empty values, the lenght 
-  is set to 0.
+  that many bytes of binary data.
 
 
 # Object Model {#model}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -797,6 +797,7 @@ The format of SUBSCRIBE REQUEST is as follows:
 SUBSCRIBE REQUEST Message {
   Full Track Name Length (i),
   Full Track Name (...),
+  Subscription Hints (..) ...
   Track Request Parameters (..) ...
 }
 ~~~
@@ -807,9 +808,12 @@ SUBSCRIBE REQUEST Message {
 
 * Track Request Parameters: As defined in {{track-req-params}}.
 
+* Subscription Hints: As defined in {{sub-hints}}.
+
 On successful subscription, the publisher SHOULD start delivering
 objects from the group sequence and object sequence as defined in the
 `Track Request Parameters`.
+
 
 ## SUBSCRIBE OK {#message-subscribe-ok}
 
@@ -992,6 +996,93 @@ The client:
 * SHOULD remain connected on both connections for a short period,
   processing objects from both in parallel.
 
+
+## Subscription Hints {#sub-hints}
+
+Subscription hints provide a way for a subscriber to indicate, to the publisher, its preferences for receiving objects from a given track.
+
+Subscription Hints have the following structure:
+
+~~~
+SUBSCRIPTION HINT {
+  HintType (8),
+  Payload (b)
+}
+~~~
+{: #moq-transport-subscription-hint format title="MOQT Subscription HInt"}
+
+The `HintType` parameter identifies one of the following values:
+
+
+|------------|-------------------------|
+| Hint Type  | Hint                    |                               
+|-----------:|:------------------------|
+| 0x0        | Reserved                |
+|------------|-------------------------|
+| 0x1        | StartPoint              |
+|------------|-------------------------|
+| 0x2        | Interval                |
+|------------|-------------------------|
+| 0x3 - 0xFF | Reserved for future use |
+|------------|-------------------------|
+
+The  `Payload` is specific to the chosen `HintType` as descirbed in the following subsections.
+
+### StartPoint Hint
+
+The `StartPoint` subscription hint identifies a starting point within the track for delivering objects. The requested start point specified is relative to publisher's current view of the state within the track. A publisher's current view of the track is defined as the most recent group and object for the track received (or in the cache), if available, at the time of request.
+
+The payload for StartPoint hint has the following structure:
+
+~~~
+StartPoint Payload {
+  Mode (i),
+  [GroupCount (i)]
+}
+~~~
+
+A subscriber can request the starting point to be one of the following values, as idenitfied by the `Mode` parameter:
+
+* Current: Has value of `0x1` and specifies delivering objects from the beginning of the current group of objects. The optional parameter `GroupCount`, if specified, MUST be ignored when processing the hint. 
+
+* Previous: Has value of `0x2` and specifies delivering objects from earlier group(s). The optional parameter `GroupCount` MUST be specified and it specifies the number of groups to go back from the current group.
+
+* Next: Has value of `0x3` and specifies delivering objects from future group(s). The optional parameter `GroupCount` MUST be specified and it specifies the number of groups, from the current group,  to wait for before delivering the objects.
+
+* Now: Has value of `0x4` and specifies delivering object from the current group and the object sequence. The optional parameter `GroupCount`, if specified, MUST be ignored when processing the hint.
+
+
+Publishers processing the StartPoint hint needs to take following into consideration:
+
+- If the StartPoint mode is not understood, the publisher MUST send SUBSCRIBE ERROR message with an appropriate error code (TODO: need to define this).
+
+- If the GroupCount cannot be satisfied, the publisher MUST send SUBSCRIBE ERROR message with an appropriate error code (TODO: need to define this).
+
+- If the GroupCount can be partially satisfied, the publisher SHOULD respond with SUBSCRIBE OK and start delivering the objects from the publisher chosen 
+
+
+TODO: Add a note in the security consideration section.
+
+
+### Interval Hint
+
+The `Interval` subscription hint allows subscribers to request for the range of objects by specifying values pertaining to start group/object sequence and end group/object sequence. 
+
+The payload for Interval hint has the following structure:
+
+~~~
+
+Interval Payload {
+  StartGroupSequence,
+  StartObjectSequence,
+  [EndGroupSequence],
+  [EndObjectSequence]
+}
+~~~
+
+The `StartGroupSequence` and `StartObjectSequence` parameters specifies the starting group and object sequence from where the objects within a tracks needs to be delivered. The `EndGroupSequence` and `EndObjectSequence` parameters identify the ending group and object sequence values respectively. Such requests are "closed/bounded interval" requests and subscription is terminated when the objects defined by the range are delivered. For cases where the requested range cannot be satisfied, publishers MUST respond with SUBSCRIBE ERROR message with Error Code set to TODO.
+
+Additionally subscribers can include only the `StartGroupSequence` and `StartObjectSequence` parameters to indicate  "open/unbounded interval" requests, wherin only the absolute start point for object delivery is specified. If either of the  `StartGroupSequence` and `StartObjectSequence` cannot be satisfied, publishers MUST respond with SUBSCRIBE ERROR message with Error Code set to TODO.
 
 ## Track Request Parameters {#track-req-params}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -842,7 +842,7 @@ index can be used.
 Groups:                  0    1    2    3    4   [5]  [6] ...
                                              ^
                                       Current Group
-RelativePrevious Group:  3    2    1    0
+RelativePrevious Group:  4    3    2    1    0
 RelativeNext Group:                               0    1  ...
 ~~~
 {: title="Relative Group Indexing"}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -995,7 +995,8 @@ Mode is None for an open-ended subscription.
 EndObject's Mode MUST be None if EndGroup's Mode is None.  EndObject's Mode MUST
 NOT be None if EndGroup's Mode is not None.
 
-* Track Request Parameters: As defined in {{track-req-params}}.
+* Track Request Parameters: The parameters are defined in
+{{version-specific-params}}
 
 On successful subscription, the publisher SHOULD start delivering
 objects from the group sequence and object sequence described above.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1058,8 +1058,6 @@ Start = Group 3, Object 0
 End = Group 3, Object <last>
 ~~~
 
-TODO: Security Considerations related to these hints
-
 
 ## SUBSCRIBE_OK {#message-subscribe-ok}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1010,7 +1010,7 @@ SUBSCRIPTION HINT {
   Value (b)
 }
 ~~~
-{: #moq-transport-subscription-hint format title="MOQT Subscription HInt"}
+{: #moq-transport-subscription-hint format title="MOQT Subscription Hint"}
 
 The `HintType` parameter identifies one of the following values:
 
@@ -1021,7 +1021,7 @@ The `HintType` parameter identifies one of the following values:
 |------------|-------------------------|
 | 0x1        | AbsoluteStartPoint      |
 |------------|-------------------------|
-| 0x2        | Interval                |
+| 0x2        | Range                   |
 |------------|-------------------------|
 | 0x3 - 0xFF | Reserved for future use |
 |------------|-------------------------|
@@ -1044,6 +1044,8 @@ RelativeStartPoint Value {
   [GroupCount (i)]
 }
 ~~~
+{: #moq-transport-relative-start-point-hint format title="MOQT RelativeStartPoint Hint"}
+
 
 A subscriber can request the starting point to be one of the following values, as idenitfied by the `Mode` parameter:
 
@@ -1072,6 +1074,8 @@ AbsoluteStartPoint Value {
   TrackOffset start
 }
 ~~~
+{: #moq-transport-absolute-start-point-hint format title="MOQT AbsoluteStartPoint Hint"}
+
 
 * TrackOffset:  Identifies the group and optionally, the object sequence value within the track as the start point for the delivery. TrackOffset is defined as below.
 
@@ -1082,22 +1086,26 @@ TrackOffset {
   [ObjectSequence (i)]
 }
 ~~~
+{: #moq-transport-track-offset format title="MOQT TrackOffset"}
+
 
  The `Type` controls if both group and object sequences are provided. A value of 0x0 indicates only the group sequence is provided and a value of 0x1 indicates both the group and object 
  sequence values are provided.
 
-### Interval Hint
+### Range Hint
 
-The `Interval` subscription hint allows subscribers to request for range of objects by specifying values pertaining to start and end group/object sequences. 
+The `Range` subscription hint allows subscribers to request for range of objects by specifying values pertaining to start and end group/object sequences. 
 
 The value for Interval hint has the following structure:
 
 ~~~
-Interval Payload {
+Range Payload {
   TrackOffset start,
   TrackOffset end
 }
 ~~~
+{: #moq-transport-range-hint format title="MOQT Range Hint"}
+
 
 ## Track Request Parameters {#track-req-params}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1043,13 +1043,13 @@ StartPoint Payload {
 
 A subscriber can request the starting point to be one of the following values, as idenitfied by the `Mode` parameter:
 
-* Current: Has value of `0x1` and specifies delivering objects from the beginning of the current group of objects. The optional parameter `GroupCount`, if specified, MUST be ignored when processing the hint. 
+* Current: Has value of `0x1` and specifies delivering objects from the beginning of the current group of objects. The optional parameter `GroupCount`, if specified, MUST be ignored when processing this hint. 
 
-* Previous: Has value of `0x2` and specifies delivering objects from earlier group(s). The optional parameter `GroupCount` MUST be specified and it specifies the number of groups to go back from the current group.
+* Previous: Has value of `0x2` and specifies delivering objects from earlier group(s) relative to the current group. The optional parameter `GroupCount` MUST be specified and it specifies the number of groups to go back from the current group to determine the start point.
 
-* Next: Has value of `0x3` and specifies delivering objects from future group(s). The optional parameter `GroupCount` MUST be specified and it specifies the number of groups, from the current group,  to wait for before delivering the objects.
+* Next: Has value of `0x3` and specifies delivering objects from future group(s). The optional parameter `GroupCount` MUST be specified and it specifies the number of groups, from the current group, to wait on before delivering the objects.
 
-* Now: Has value of `0x4` and specifies delivering object from the current group and the object sequence. The optional parameter `GroupCount`, if specified, MUST be ignored when processing the hint.
+* Now: Has value of `0x4` and specifies start delivering from the current group and the object sequence. The optional parameter `GroupCount`, if specified, MUST be ignored when processing the hint.
 
 
 Publishers processing the StartPoint hint needs to take following into consideration:
@@ -1058,7 +1058,7 @@ Publishers processing the StartPoint hint needs to take following into considera
 
 - If the GroupCount cannot be satisfied, the publisher MUST send SUBSCRIBE ERROR message with an appropriate error code (TODO: need to define this).
 
-- If the GroupCount can be partially satisfied, the publisher SHOULD respond with SUBSCRIBE OK and start delivering the objects from the publisher chosen 
+- If the GroupCount can be satisfied, the publisher SHOULD respond with SUBSCRIBE OK and start delivering the objects.
 
 
 TODO: Add a note in the security consideration section.
@@ -1066,7 +1066,7 @@ TODO: Add a note in the security consideration section.
 
 ### Interval Hint
 
-The `Interval` subscription hint allows subscribers to request for the range of objects by specifying values pertaining to start group/object sequence and end group/object sequence. 
+The `Interval` subscription hint allows subscribers to request for  range of objects by specifying values pertaining to start group/object and end group/object sequences. 
 
 The payload for Interval hint has the following structure:
 
@@ -1080,9 +1080,9 @@ Interval Payload {
 }
 ~~~
 
-The `StartGroupSequence` and `StartObjectSequence` parameters specifies the starting group and object sequence from where the objects within a tracks needs to be delivered. The `EndGroupSequence` and `EndObjectSequence` parameters identify the ending group and object sequence values respectively. Such requests are "closed/bounded interval" requests and subscription is terminated when the objects defined by the range are delivered. For cases where the requested range cannot be satisfied, publishers MUST respond with SUBSCRIBE ERROR message with Error Code set to TODO.
+The `StartGroupSequence` and `StartObjectSequence` parameters specifies the starting group and object sequence from where the objects within a tracks needs to be delivered. The `EndGroupSequence` and `EndObjectSequence` parameters identify the ending group and object sequence values respectively. Such requests are "closed/bounded interval" requests and subscription is terminated when the objects defined by the requested range are published. if the requested range cannot be satisfied, publishers MUST send SUBSCRIBE ERROR message with Error Code set to TODO.
 
-Additionally subscribers can include only the `StartGroupSequence` and `StartObjectSequence` parameters to indicate  "open/unbounded interval" requests, wherin only the absolute start point for object delivery is specified. If either of the  `StartGroupSequence` and `StartObjectSequence` cannot be satisfied, publishers MUST respond with SUBSCRIBE ERROR message with Error Code set to TODO.
+Additionally subscribers can include only the `StartGroupSequence` and `StartObjectSequence` parameters to indicate "open/unbounded interval" requests, wherin only the absolute start point for object delivery is specified. If either of the  `StartGroupSequence` and `StartObjectSequence` cannot be satisfied, publishers MUST respond with SUBSCRIBE ERROR message with Error Code set to TODO.
 
 ## Track Request Parameters {#track-req-params}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3,12 +3,12 @@ title: "Media over QUIC Transport"
 abbrev: moq-transport
 docname: draft-ietf-moq-transport-00
 date: {DATE}
-category: std 
+category: std
 
 ipr: trust200902
 area: Applications and Real-Time
 submissionType: IETF
-workgroup: MOQ 
+workgroup: MOQ
 keyword: Internet-Draft
 
 stand_alone: yes
@@ -172,7 +172,7 @@ Consumer:
 
 Endpoint:
 
-: A QUIC Client or a QUIC Server. 
+: A QUIC Client or a QUIC Server.
 
 Group:
 
@@ -797,8 +797,6 @@ The format of SUBSCRIBE REQUEST is as follows:
 SUBSCRIBE REQUEST Message {
   Full Track Name Length (i),
   Full Track Name (...),
-  Number of Hints (i),
-  Subscription Hints (..) ...
   Track Request Parameters (..) ...
 }
 ~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -797,6 +797,7 @@ The format of SUBSCRIBE REQUEST is as follows:
 SUBSCRIBE REQUEST Message {
   Full Track Name Length (i),
   Full Track Name (...),
+  Subscription Hints (..) ...
```
afrind, line 800: Should it be a separate field or track request parameter?  Track Request Params has group/object sequence already which can maybe be collapsed here one way or another.

draft-ietf-moq-transport.md
```diff
@@ -993,6 +997,93 @@ The client:
...
 
+## Subscription Hints {#sub-hints}
+
+Subscription hints provide a way for a subscriber to indicate, to the publisher, its preferences for receiving objects from a given track.
+
+Subscription Hints have the following structure:
+
+~~~
+SUBSCRIPTION HINT {
+  HintType (8),
```
afrind, line 1008: Someone will say "varint", I just wanted to be first.

draft-ietf-moq-transport.md
```diff
@@ -993,6 +997,93 @@ The client:
...
+
+A subscriber can request the starting point to be one of the following values, as idenitfied by the `Mode` parameter:
+
+* PREVIOUS: Start point for object delivery is from an earlier group relative to the current group. The optional parameter `GroupCount` MUST be specified and it specifies the number of groups to go back from the current group to determine the start point. It has value of `0x1`.
+
+* NEXT: Start point for object delivery is set to a future group relative to the current group. The optional parameter `GroupCount` MUST be specified and it specifies the number of groups to wait on before delivering the objects. It has value of `0x2`.
+
+* CURRENT: Start point for object delivery is set to beginning of the current group. The optional parameter `GroupCount`, if specified, MUST be ignored when processing this hint. It has value of `0x3`.
+
+* NOW: Start point for object delivery is set to most recent obect from the current group. The optional parameter `GroupCount`, if specified, MUST be ignored when processing the hint. It has value of `0x4`
```
afrind, line 1052: I don't think you can have it be optional and ignored, since there's no flag indicating it's presence or absence other than `mode`.  PREVIOUS and NEXT mean there is a GroupCount and CURRENT and NOW mean there is not.

afrind, line 1052: s/obect/object/

draft-ietf-moq-transport.md
```diff
@@ -993,6 +997,93 @@ The client:
...
+StartPoint Payload {
+  Mode (i),
+  [GroupCount (i)]
+}
+~~~
+
+A subscriber can request the starting point to be one of the following values, as idenitfied by the `Mode` parameter:
+
+* PREVIOUS: Start point for object delivery is from an earlier group relative to the current group. The optional parameter `GroupCount` MUST be specified and it specifies the number of groups to go back from the current group to determine the start point. It has value of `0x1`.
+
```
huitema, line 1047: What has value 0x1? The groupcount parameter?

draft-ietf-moq-transport.md
```diff
@@ -993,6 +997,93 @@ The client:
...
+* NEXT: Start point for object delivery is set to a future group relative to the current group. The optional parameter `GroupCount` MUST be specified and it specifies the number of groups to wait on before delivering the objects. It has value of `0x2`.
+
```
huitema, line 1049: What has value 0x2? The groupcount parameter?

draft-ietf-moq-transport.md
```diff
@@ -993,6 +997,93 @@ The client:
...
+* CURRENT: Start point for object delivery is set to beginning of the current group. The optional parameter `GroupCount`, if specified, MUST be ignored when processing this hint. It has value of `0x3`.
+
```
huitema, line 1051: What has value 0x3? The groupcount parameter?

draft-ietf-moq-transport.md
```diff
@@ -993,6 +997,93 @@ The client:
...
+Subscription Hints have the following structure:
+
+~~~
+SUBSCRIPTION HINT {
+  HintType (8),
+  Payload (b)
+}
+~~~
+{: #moq-transport-subscription-hint format title="MOQT Subscription HInt"}
+
```
huitema, line 1013: Why a fixed 8 bits field? we are using varint everywhere.

draft-ietf-moq-transport.md
```diff
@@ -993,6 +997,93 @@ The client:
...
+  StartGroupSequence(i),
+  StartObjectSequence(i),
+  [EndGroupSequence(i)],
+  [EndObjectSequence(i)]
+}
+~~~
+
+The `StartGroupSequence` and `StartObjectSequence` parameters specify the start group and object sequence of the interval and   `EndGroupSequence` and `EndObjectSequence` parameters capture the ending group and object sequence values respectively. Such requests are "closed/bounded interval" requests with the subscription terminated when the objects defined by the requested range are published. If the requested range cannot be satisfied, publishers MUST send SUBSCRIBE ERROR message with Error Code set to TODO.
+
+Additionally subscribers can include only the `StartGroupSequence` and `StartObjectSequence` parameters to indicate "open/unbounded interval" requests, wherin only the absolute start point for object delivery is specified. If either of the  `StartGroupSequence` and `StartObjectSequence` cannot be satisfied, publishers MUST respond with SUBSCRIBE ERROR message with Error Code set to TODO.
```
afrind, line 1085: I think this is not parseable as written - I don't know when there is or isn't an EndGroupSequence or EndGroupObject. 

Maybe something like the following

```
TrackOffset {
  Type (i),  // 0 = Group, 1 = Group and Object
  Group (i),
  [Object (i)]
}

Hint Types = RelativeStart, AbsoluteStart, AbsoluteInterval
RelativeStartPayload {
  Mode (i),
  [GroupCount(i)]
}
Modes = Previous, Next, Current, Now
// Previous and Next have a GroupCount, Current and Now do not have GroupCount.

AbsoluteStartPayload {
  TrackOffset start
}
This would replace the track request params for GROUP SEQUENCE / OBJECT SEQUENCE

AbsoluteIntervalPayload {
  TrackOffset start,
  TrackOffset end
}
```

huitema (commented):
Overall, that's a complicated syntax. You have optional elements, but the TLS1.3 syntax that we use limits how we can specify optional elements. For example, when you write:
~~~
StartPoint Payload {
  Mode (i),
  [GroupCount (i)]
}
~~~
the decoder has an issue. If there are bytes in the message following the encoding of encoding of "mode", do those bytes encode the `GroupCount`, or do they belong to the encoding of the next element in the subscribe command? Your description of "now" says that "The optional parameter `GroupCount`, if specified, MUST be ignored" -- but there the decoder is really in a bind.

This is a problem throughout. You allow extension points, but the syntax is not designed to "skip over" unknown extensions. To be on the safe side, I would change the syntax of the subscription hint to be:
~~~
SUBSCRIPTION HINT {
  HintType (i),
  HintLength(i),
  Payload (...)
}
~~~
If you do that, then you can have optional elements for each payload type. (I would rename "payload" to "hint parameters" or "hintValue".)

Regarding the `HintType`, how is this defined? Suppose that I want to define a new hint type, what do I do? Will there be an IANA registry? Will there be a range of numbers reserved for experimentation?

afrind (commented):
Individual Review

Thanks for this PR, I think it's directionally right but needs a few tweaks for parseability.

afrind:
Can you mark this as Fixes: #245 so it shows as having a PR attached (also possibly 111 and 260)?  Also recommend closing #246 as this replaces it.

draft-ietf-moq-transport.md
afrind: You can also spell it a little differently, like:

```
Hint Types = Current, Now, RelativeStartPrevious, RelativeStartNext, AbsoluteStart, AbsoluteInterval

CurrentPayload = Empty

NowPayload = Empty

RelativeStartPreviousPayload {
  GroupCount(i)
}

RelativeStartNextPayload {
  GroupCount(i)
}

AbsoluteStartPayload {
  TrackOffset start
}

AbsoluteIntervalPayload {
  TrackOffset start,
  TrackOffset end
}
```

kixelated (commented):
It's really a complicated encoding and yet not flexible. I don't like modes that can't be generalized; we'll end up with spaghetti.

Here's my proposal: https://github.com/moq-wg/moq-transport/issues/245#issuecomment-1767492853

draft-ietf-moq-transport.md
kixelated: +1 a registry/table would be clearer.

draft-ietf-moq-transport.md
kixelated: Good catch.

kixelated (changes requested):

suhasHere:
> Here's my proposal: [#245 (comment)](https://github.com/moq-wg/moq-transport/issues/245#issuecomment-1767492853)

I feel #245 proposal is more ambiguous though.  I agree we can make the proposed ending in this PR to make it easier to encode/decode.

I have to get higher level idea agreed upon

1. There are subscription hints. 2 hints are defined in this PR  a) StartPoint b) Interval
2. There may be hints we might want to add in future and hence a Hint structure identifies a type and value
3. StartPoint hint can support 4 modes. These modes are only applicable when the HintType is StartPoint 
4. Interval hint is totally separate hintType that defines ranges of groups/objects

draft-ietf-moq-transport.md
```diff
@@ -993,6 +997,93 @@ The client:
...
+
+### StartPoint Hint
+
+The `StartPoint` subscription hint identifies a starting point within a track for delivering objects. The requested start point specified is relative to publisher's view  of the ongoing state of the track. A publisher's current state of the track is defined by the most recent group and object sequence received (or active in the cache), if available, at the time of request.
+
+The payload for StartPoint hint has the following structure:
+
+~~~
+StartPoint Payload {
+  Mode (i),
```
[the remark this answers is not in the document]
suhasHere: this is a not a mode in a mode. Its mode in a given hintType called StartPoint

draft-ietf-moq-transport.md
suhasHere: 0x1 correspond to Mode and not the groupCount. I can clarify the same.

draft-ietf-moq-transport.md
suhasHere: 0x2 correspond to Mode and not the groupCount. I can clarify the same.

draft-ietf-moq-transport.md
suhasHere: 0x3 correspond to Mode and not the groupCount. I can clarify the same.

draft-ietf-moq-transport.md
suhasHere: I am fine with varint. I didn't see it being needed in this case.

draft-ietf-moq-transport.md
suhasHere: yes , once we get the hints agreed upon, i will do a follow up PR to change Track Params. Trying to keep the PR focussed.

draft-ietf-moq-transport.md
suhasHere: Mode is the flag that defined the presence and absence. The text is trying to say that, if for mode "Now", some one encodes GroupCount, don't process that param than treating it as error.

I am not sure its ambiguous, but happy to take suggestions on how to make it clear.

draft-ietf-moq-transport.md
```diff
@@ -993,6 +997,93 @@ The client:
...
+* CURRENT: Start point for object delivery is set to beginning of the current group. The optional parameter `GroupCount`, if specified, MUST be ignored when processing this hint. It has value of `0x3`.
+
+* NOW: Start point for object delivery is set to most recent obect from the current group. The optional parameter `GroupCount`, if specified, MUST be ignored when processing the hint. It has value of `0x4`
+
+
+Publishers processing the StartPoint hint needs to take following into consideration:
+
+- If the StartPoint mode is not understood, the publisher MUST send SUBSCRIBE ERROR message with an appropriate error code (TODO: need to define this).
+
+- If the GroupCount cannot be satisfied, the publisher MUST send SUBSCRIBE ERROR message with an appropriate error code (TODO: need to define this).
```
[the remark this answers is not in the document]
suhasHere: yes , agree with the analysis. I did had it written originally aligning so, but not able to remember why i decided to go otherway.

But, i will update it.

suhasHere:
> Regarding the `HintType`, how is this defined? Suppose that I want to define a new hint type, what do I do? Will there be an IANA registry? Will there be a range of numbers reserved for experimentation?

Having a registry is the idea and I was thinking once we get to agreement of Hint strucutres , then we can follow with defining error codes, registries for types and  cleaning up track request params

suhasHere:
> SUBSCRIPTION HINT {
>   HintType (i),
>   HintLength(i),
>   Payload (...)
> }

Payload(b) is defined to have len and value .  Here is the definition as defined in the draft today 
**x (b): : Indicates that x consists of a variable length integer, followed by that many bytes of binary data.**

draft-ietf-moq-transport.md
kixelated: It's the same encoding for values < 2^6 , so you might as well specify a varint.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
kixelated: What I'm suggesting is that you merge `Mode` and `HintType`. So there would be 5 possible values, instead of 4+1. The parameters for each would be based on the type; there would be a `group` for PREVIOUS/NEXT and a start/end pair for INTERVAL.

draft-ietf-moq-transport.md
kixelated: What Alan is trying to say is that there's no way to specify "GroupCount" when mode=CURRENT or NOW. The field is optional and the mode determines if the field is present, or at least that's our interpretation. If the field is always present and MUST be ignored like this text specifies, then `[GroupCount (i)]` shouldn't be wrapped in brackets.

draft-ietf-moq-transport.md
kixelated: Yeah, the bracket notation that QUIC uses doesn't encode if a field is present.

```
[EndGroupSequence(i)]
```

Needs to be written as something like:

```
EndGroupPresent (i) // 1 or 0
[EndGroupSequence(i)]
```

This is quite wasteful, which is why you generally have flags or modes to specify if multiple fields are present. Or a special meaning for 0.

kixelated (changes requested):

draft-ietf-moq-transport.md
```diff
@@ -993,6 +997,93 @@ The client:
...
+
+
+|------------|-------------------------|
+| Hint Type  | Hint                    |                               
+|-----------:|:------------------------|
+| 0x0        | Reserved                |
+|------------|-------------------------|
+| 0x1        | StartPoint              |
+|------------|-------------------------|
+| 0x2        | Interval                |
```
kixelated, line 1024: nit: `Range` instead of `Interval`

suhasHere:
So pushed few more commits cleaning up somethings and considering suggestion from Alan on TrackOffset and removing text on error.  

Would love to hear feedback on the whole text.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
afrind: Yeah, this comment I left from the previous draft shows one possible way to merge Mode and HintType - https://github.com/moq-wg/moq-transport/pull/289#discussion_r1362945991.  I think I like that slightly better because it has only one extension axis, and it's more compact, but I can live with this too.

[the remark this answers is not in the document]
huitema: I was going to make the same suggestion. We can certainly collapse these two layers. We would end up with Hints being a choice of "Now, Current, Previous, Next, Absolute, Range".

draft-ietf-moq-transport.md
```diff
@@ -993,6 +998,115 @@ The client:
...
+The `Range` subscription hint allows subscribers to request for range of objects by specifying values pertaining to start and end group/object sequences. 
+
+The value for Interval hint has the following structure:
+
+~~~
+Range Payload {
+  TrackOffset start,
+  TrackOffset end
+}
+~~~
```
huitema, line 1106: The syntax seems uselessly complicated. I would remove the type, just define:
~~~
TrackOffset {
  GroupSequence (i),
  ObjectSequence (i)
}
~~~
And set Object to 0 if the range shall include the whole group.

You also need to specify whether the end is inclusive or exclusive. I like exclusive, because it allows to set the end to "Group=N+1, Object=0", which would mean "until the end of Group N" without knowing in advance how many objects there are in group N.

huitema (commented):
This is progressing. Apart from the comments below, I would take a hard look at the extension points. Do we need them, or can we just say that "a future version of this spec may define more hint types"?

This would be in line with only having extensions by version number, negotiated during the connection.

draft-ietf-moq-transport.md
afrind: > And set Object to 0 if the range shall include the whole group.

Thanks @huitema for this suggestion, it makes much more sense.

`group=7, object=0` vs `type=0, group=7`
`group=9, object=17` vs `type=1, group=9, object=17`

draft-ietf-moq-transport.md
```diff
@@ -993,6 +998,115 @@ The client:
...
+
+The `Value` is specific to the chosen `HintType` as descirbed in the following subsections.
+
+TODO: Define registry for specifying the hints.
+
+TODO: Define experimental range.
+
+### RelativeStartPoint Hint {#relative}
+
+The `RelativeStartPoint` subscription hint identifies a starting point relative to publisher's view of the ongoing state of the track for delivering objects. A publisher's current state of the track is defined by the most recent group and object sequence received (or active in the cache), if available, at the time of request.
```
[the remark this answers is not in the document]
suhasHere: I wonder there is any way a subscriber can know this info. Whatever publisher says in the OBJECT message is considered to be the start for all relative requests. isn't it ?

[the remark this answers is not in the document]
afrind: My thought was that the first object received might not be the first object sent by the publisher (due to out of order delivery, congestion, etc), hence it might be good for the publisher to indicate to the subscriber what they are going to get.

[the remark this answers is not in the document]
suhasHere: the thing is , when you send subscribe , there might not be any publisher started publishing .. so info in OK is not sufficient .. 
May be we need a bit in the object header, that if set means its the first object in the track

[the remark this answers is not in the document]
afrind: > the thing is , when you send subscribe , there might not be any publisher started publishing .. so info in OK is not sufficient ..

This makes sense too. Optional flag in SUBSCRIBE OK?   I could live with punting this feature also for now.

draft-ietf-moq-transport.md
```diff
@@ -993,6 +999,121 @@ The client:
...
+~~~
+TrackOffset {
+  GroupSequence (i),
+  ObjectSequence (i)
+}
+~~~
+{: #moq-transport-track-offset format title="MOQT TrackOffset"}
+
+`ObjectSequence` set to 0x0 implies until the end of the group identified in `GroupSequence`.
+
```
huitema, line 1099: I would think that `ObjectSequence` set to 0x0 implies ~until the end~ *from the beginning* of the group identified in `GroupSequence`.

huitema (approved):
Just one minor comment. I think this is pretty much ready.

draft-ietf-moq-transport.md
```diff
@@ -797,6 +797,8 @@ The format of SUBSCRIBE REQUEST is as follows:
 SUBSCRIBE REQUEST Message {
   Full Track Name Length (i),
   Full Track Name (...),
+  Number of Hints (i),
```
vasilvv, line 800: Could you explain what does it mean to have multiple hints?  Like, semantically, if I supply both a relative and an absolute offset, what should happen?

draft-ietf-moq-transport.md
```diff
@@ -993,6 +999,121 @@ The client:
...
+### RelativeStartNext Hint
+
+`RelativeStartNext` subscription hint specifies the start point for object delivery to a future group relative to the current group. The `GroupCount` hint attribute specifies the number of groups to wait on before delivering the objects.
+
+~~~
+RelativeStartNext HintValue {
+  GroupCount(i)
+}
+~~~
+{: #moq-transport-relative-start-next-hint format title="MOQT RelativeStartNext Hint"}
```
vasilvv, line 1072: What is the relationship between Current, RelativeStartPrevious with count 0, and RelativeStartNext with count 0?  Are they identical? Is count 0 disallowed by the protocol?

draft-ietf-moq-transport.md
```diff
@@ -993,6 +999,121 @@ The client:
...
+
+### AbsoluteStart Hint
+
+The `AbsoluteStart` subscription hint allows subscribers to specify an absolute point in the track to start delivering objects, as indicated by the `TrackOffset` hint attribute.
```
vasilvv, line 1076: Can `TrackOffset` be in the future? If so, please document the expected behavior.

draft-ietf-moq-transport.md
```diff
@@ -993,6 +998,115 @@ The client:
...
+|------------|-------------------------|
+| Hint Type  | Hint                    |                               
+|-----------:|:------------------------|
+| 0x0        | RelativeStartPoint      |
+|------------|-------------------------|
+| 0x1        | AbsoluteStartPoint      |
+|------------|-------------------------|
+| 0x2        | Range                   |
+|------------|-------------------------|
+| 0x3 - 0xFF | Reserved for future use |
```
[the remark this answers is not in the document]
vasilvv: (ignore this, somehow submitted outdated comment)

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
suhasHere: yes, let's open an new issue to add  a flag to the OBJECT and also have some info on OK message

draft-ietf-moq-transport.md
suhasHere: i was thinking if we define a new hint type that is not related to start point in future, allowing it to be useful.  OTOH i wonder if we can make a request that if there exists both relative and absolute start point, have absolute take preference always ?

I don't know why would one define, but we need to define what happens when one does so

draft-ietf-moq-transport.md
suhasHere: my thinking is count of 0 for RelativeStartPrevious gives the current group. Do you see if there is any issues with that ? We can have GroupCount > 0 always too ..

draft-ietf-moq-transport.md
suhasHere: yes, it can be .. What were you thinking we should document ?

draft-ietf-moq-transport.md
```diff
@@ -807,10 +809,14 @@ SUBSCRIBE REQUEST Message {
 
 * Track Request Parameters: As defined in {{track-req-params}}.
 
+* Subscription Hints: As defined in {{sub-hints}}.
+
 On successful subscription, the publisher SHOULD start delivering
 objects from the group sequence and object sequence as defined in the
 `Track Request Parameters`.
 
+OPEN QUESTION: Should we disallow multiple hints ?
```
afrind, line 818: My preference is for only a single hint.  The current hint set can be mixed in non-sensical ways.

draft-ietf-moq-transport.md
```diff
@@ -993,6 +999,121 @@ The client:
...
+## Subscription Hints {#sub-hints}
+
+Subscription hints provide a way for a subscriber to indicate, to the publisher, its preferences for receiving objects from a given track. 
+
+Subscription Hints have the following structure:
+
+~~~
+SUBSCRIPTION HINT {
+  HintType (i),
+  HintValueLength (i),
```
afrind, line 1011: As currently defined, the length is not needed because the length of all hint types is known.

draft-ietf-moq-transport.md
```diff
@@ -993,6 +999,121 @@ The client:
...
+
+~~~
+AbsoluteInterval HintValue {
+  TrackOffset start,
+  TrackOffset end
+}
+~~~
+{: #moq-transport-absolute-interval-hint format title="MOQT AbsoluteInterval Hint"}
+
+The end track offset of the range is exclusive.
```
afrind, line 1114: So that's:

[0, 0], to [2, 0] = all of groups 0 and 1
[0, 0], to [2, 1] = all of groups 0 and 1, and group 2, object 0, 

etc?

draft-ietf-moq-transport.md
afrind: Maybe - 1) state that it is intended that subscribers can ask for a future offset (similar to RelativeStartNext with a positive number), and then specify if the publisher needs to wait for those objects to be available to issue SUBSCRIBE OK, or it issues a SUBSCRIBE OK immediately?  What if those objects never materialize?

afrind (commented):
Individual Review

draft-ietf-moq-transport.md
kixelated: I really don't want multiple hints. It's not clear how overlapping hints are supposed to interact, which makes it implementation specific.

draft-ietf-moq-transport.md
kixelated: Yeah, you only need the length if you want to support unknown HintTypes. However, I think it should just be a protocol violation if you use an unknown type, so you don't need the length.

draft-ietf-moq-transport.md
```diff
@@ -993,6 +999,121 @@ The client:
...
+In the sections below, a publisher's current state of the track is defined by the most recent group and object sequence received (or active in the cache), if available, at the time of request.
+
+### Current Hint
+
+`Current` subscription hint specifies the start point for object delivery from the beginning of the current group. Current hint type defines no further hint attributes.
+
+
+### Now Hint
+
+`Now` subscription hint specifies the start point for object delivery from the most recent object of the current group.  Now hint type defines no further hint attributes.
```
kixelated, line 1048: Does this include the most recent object? If so, the latest object could have been generated hours ago, so "now" is a little misleading.

I would rename:

| old | new |
|-|-|
| Current | CurrentGroup |
| Now | CurrentObject |

draft-ietf-moq-transport.md
```diff
@@ -993,6 +999,121 @@ The client:
...
+
+
+### RelativeStartPrevious Hint
+
+`RelativeStartPrevious` subscription hint specifies the start point for object delivery from an earlier group relative to the current group. The  `GroupCount` hint attribute specifies the number of groups to go back from the current group to determine the start point.
```
kixelated, line 1053: Would `RelativeStartPrevious GroupCount = 0` be identical to `Current`?

Can you clarify if this is the case, and give an example?

draft-ietf-moq-transport.md
```diff
@@ -993,6 +999,121 @@ The client:
...
+RelativeStartPrevious HintValue {
+  GroupCount(i)
+}
+~~~
+{: #moq-transport-relative-start-previous-hint format title="MOQT RelativeStartPrevious Hint"}
+
+
+### RelativeStartNext Hint
+
+`RelativeStartNext` subscription hint specifies the start point for object delivery to a future group relative to the current group. The `GroupCount` hint attribute specifies the number of groups to wait on before delivering the objects.
```
kixelated, line 1065: Would `RelativeStartNext GroupCount = 0` be identical to `Current`?

Can you clarify if this is the case, and give an example?

draft-ietf-moq-transport.md
kixelated: It's just confusing and you need to clarify at least. This is a classic off-by-one bug waiting to happen. It also might make caching more difficult because there's 3 different ways to request the latest group.

IMO
`RelativeStartPrevious` is relative to `current`
`RelativeStartNext` is relative to `current+1`
`Current` is removed in favor of `RelativeStartPrevious count=0`

draft-ietf-moq-transport.md
kixelated: +100

How are you supposed to request the first object from a group otherwise? What is even the use-case of requesting the max object from an old group?

kixelated (changes requested):
I wanted to +1 Alan's comment but I think it got buried in changes. We will eventually need the selected start/end group/object echoed in SUBSCRIBE_OK. Otherwise I don't think you can actually use these relative hints; there's a bunch of edge cases and races.

draft-ietf-moq-transport.md
suhasHere: I don't see it causes implementation confusion or off-by-one bug. The semantics are clear and yes, more text can be added, if needed.  

```
                          5  4  3  2  1                           5  4  3  2  1
<--------Previous -------> CURRENT (0) <-------- Next ---------->

```

draft-ietf-moq-transport.md
suhasHere: The existing text does say the count needs to be happen from the current group as anchor. That seems pretty clear though.

draft-ietf-moq-transport.md
suhasHere: I tend to like Now better than CurrentObject since it feels more suited when used with subscribe. CurrentObject makes me wonder if it is more befitting to Fetch/Get semantics (#111 )

draft-ietf-moq-transport.md
huitema: Yes, seems right. The problem is that if you want inclusive, you have to write:

[0, 0], to [1, 7563] = all of groups 0 and 1

But you can only do that if you know in advance that group 1 has exactly 7563 objects -- and that's no obvious.

suhasHere: that's true ..  as a mental model, it helps me build nice breakpoints between end of one range and start of new range in a easier way ..

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
suhasHere: TO DISCUSS LIVE

draft-ietf-moq-transport.md
suhasHere: DISCUSS LIVE

draft-ietf-moq-transport.md
suhasHere: DISCUSS LIVE

draft-ietf-moq-transport.md
suhasHere: DISCUSS LIVE

draft-ietf-moq-transport.md
suhasHere: DISCUSS LIVE

draft-ietf-moq-transport.md
suhasHere: DISCUSS LIVE

afrind (commented):
Trying to summarize what we need to discuss live:

1. Do the use cases mentioned by Luke in 245 need to be covered now?
2. One hint or multiple hints?
3. Is RelativeStart(Next/Prev), Group=0 the same as Current?
3a. If so, should those be disallowed?
4. How should a publisher respond to a hint in the future?
5. Does TrackOffset ObjectSequence=0 mean "from the beginning" or "til the end" of the group?

Anything else?

suhasHere:
> 1. Do the use cases mentioned by Luke in 245 need to be covered now?

This is very vague given the length of 245. Could you please summarize the use-cases to help focus the discussion.

In the order of small to big, my suggestion of the order would be "3, 4, 5, 3,  2, 1 "

afrind:
> Given the length of 245. Could you please summarize the use-cases to help focus the discussion.

There are 4 examples right at the end of 245, I copied them in this review comment:

https://github.com/moq-wg/moq-transport/pull/289#pullrequestreview-1685348042

We can discuss tomorrow.

fluffy:
Alan proposed we use 6 parameters.

Start Object
Start Group
Start relative / absolute flag 

End Object
End Group 
End relative / absolute flag 

Preference for not two ways to say same thing. Use 0 to indicate current. Offset encoding. 

You can subscribe to the track twice with two different hints if that is what you want. 

Seems good consensus on call for this.

kixelated:
My proposal is 4 parameters, combining the flag with the value:

- Start Group + Relative/Absolute Flag
- Start Object + Relative/Absolute Flag
- End Group + Relative/Absolute Flag
- End Object + Relative/Absolute Flag

I'm using the lower bits of the VarInt to save some bytes and make parsing easier. Otherwise you have to deal with missing the corresponding sequence/mode parameter.

I think it's even easier if these 4 VarInts are just encoded as part of SUBSCRIBE instead of being parameters. There would be a None flag if you don't want an end group and/or object.

suhasHere:
> My proposal is 4 parameters, combining the flag with the value:
> 
> * Start Group + Relative/Absolute Flag
> * Start Object + Relative/Absolute Flag
> * End Group + Relative/Absolute Flag
> * End Object + Relative/Absolute Flag
> 
> I'm using the lower bits of the VarInt to save some bytes and make parsing easier. Otherwise you have to deal with missing the corresponding sequence/mode parameter.
> 

Even with flags embedded, the logic is exactly same. Either you need to handle missing parameter or a flag bits set to a certain value. May be we can add this optimization later too ?

> I think it's even easier if these 4 VarInts are just encoded as part of SUBSCRIBE instead of being parameters. There would be a None flag if you don't want an end group and/or object.

+1 on making it part of Subscribe.

kixelated:
Regardless of the approach, I think these are the use-cases we MUST support:

| start group | start object | end group | end object | notes |
|-|-|-|-|-|
| absolute | 0 | fin | fin | start at timestamp in catalog |
| absolute | absolute | fin | fin | reconnect from max object |
| absolute | 0 | absolute | fin | repopulate group cache (unknown end) |
| absolute | absolute | absolute | absolute | repopulate object cache (known end) |
| latest-N | 0 | fin | fin | live with initial buffer of at most N groups |
| latest | latest-N | fin | fin | live with initial buffer of at most N objects  |
| latest+1 | 0 | fin | fin | wait for next sync point |

I think these are use-cases worth considering:

| start group | start object | end group | end object | notes |
|-|-|-|-|-|
| latest-N | 0 | latest-M | fin | fetch the latest M-N groups (ex. clip) |
| latest+N | 0 | fin | fin | start at a group further in the future |
| latest-N | 0 | latest+M | fin | subscribe for M-N groups pending a resubscribe (keep-alive) |
| absolute | absolute | latest | latest | fetch the track as it currently exists (live unsupported)  |

And not mentioned in this PR, but I do think we need UNSUBSCRIBE hints.

| end group | end object | notes |
|-|-|-|
| latest | latest | end immediately |
| latest | fin | end after current group |
| absolute | fin | end after target group (session migration) |
| latest+M | fin | end after future group (session migration) |
| start group | start object | end before next SUBSCRIBE starts (deduplication?) |

kixelated:
New proposal:

```
SUBSCRIBE {
  Start Group + Absolute/Relative/Future (i)
  Start Object + Absolute/Relative/Future (i)
}

UNSUBSCRIBE {
  End Group + Absolute/Relative/Future (i)
  End Object + Absolute/Relative/Future (i)
}
```

We'll need to handle multiple subscribes and unsubscribes though, probably with a subscription ID. ex. You might want to cancel a future UNSUBSCRIBE, replacing it with a later future UNSUBSCRIBE or perhaps an immediate one.

huitema:
In addition to what is summarized above, we had a discussion about the way to express relative ranges. In the case of the start hint, the well understood scenarios that I heard are:

* start at the current point (current group, current object)
* start at the beginning of the current group (e.g., because the receiver needs all the frames from the beginning of the current group to synchronize its decoder)
* start at the beginning of the next group (e.g., because the receiver does not want to process old frames, and would rather start clean on a group start)

We can make it more complicated, and we have proposals such as:

* start N (groups, objects) before the current one (in the past)
* start N (groups, objects) after the current one (in the future).

I think that these relative offsets bring a lot of complexity, and a lot of corner cases. They are somewhat undefined, because the subscriber does not know what the current group or object is. If it knew, it could just ask for a specific start point. The combination of groups and objects is a bit weird, because the subscriber does not know how many objects there are in the past groups. The syntax allows for silly requests, like "start 1,000,000 groups in the past", or requests that are latent DOS attack like "start 1,000,000 groups in the future", which means "keep state in memory forever". There is a potential simplification there...

suhasHere:
I want to keep the scope simple and focussed. When i presented on Subscription hints, there were few real use-cases that had a need for hints 
   - User joining a mid cal
   - User reconnecting and joining at a known point
   - User want to buffer a few seconds from past to keep the player go smoothly
   - User asking for range of frames and be done with it as catch up
   
Can we define a system that works for all possible combinations ever - Oh YEAH .. But do we really need it - I DOUBT IT

I am waiting on Alan's proposal for the promised simplification. @afrind  no pressure :-)

afrind:
I was pretty busy today but have been written up what we discussed during the call - using 6 track request parameters offering the full range of options that are available in this PR as is.  By using defaults when a param is omitted, there was only one invalid combination, which I like.  The logic to determine what group and object to actually start serving is about 20 lines of pseudo code, which I added as an appendix.

I'm not sure what to make of the discussion on this PR.  I'll write a second PR that is simpler and just includes the 4 basic modes, and maybe something for range.  Stay tuned.

Would it be better to submit both as separate PR's, and leave this proposal as is, or should I push what we discussed this morning here?

suhasHere:
> I'm not sure what to make of the discussion on this PR. I'll write a second PR that is simpler and just includes the 4 basic modes, and maybe something for range. Stay tuned.
> 
> Would it be better to submit both as separate PR's, and leave this proposal as is, or should I push what we discussed this morning here?

Can't both the merged into 1 commit ?

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,164 @@ SUBSCRIBE REQUEST Message {
...
+
+The following Subscribe Hint Modes are defined for determining the requested
+group:
+
+Absolute (0x0): The specified group is requested
+
+RelativePrevious (0x1): The specified group is an index relative to the `Current
+Group`. 0 indicates the current group, 1 indicates the previous group, and so on.
+
+RelativeNext (0x2): The specified group is an index relative to the `Current
```
ianswett, line 834: I thought we decided it wasn't clear we needed this, except maybe there was interest in the 0 case when you want to start at the next group.

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,164 @@ SUBSCRIBE REQUEST Message {
...
+
+ Start = Group 3, Object 0
+ End = Group 3, Object <last>
+~~~
+
+TODO: Security Considerations related to these hints
+
+====================
+
+PROPOSAL 2: Simpler, trying to follow discussion on the PR through the day.
```
ianswett, line 927: Overall, I prefer this, because I think it's easier to reason about and there are fewer edge cases.

ianswett (commented):
I prefer Proposal 2, but I can live with Proposal 1 as well.

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,164 @@ SUBSCRIBE REQUEST Message {
...
+|------------------------------------------------|
+
+SUBSCRIBE_REQUEST also defines a Track Request Parameter RANGE (0x2) for
+requesting a closed range of objects.  The value is as follows:
+
+~~~
+Range Payload {
+  Start Group Sequence(i),
+  Start Object Sequence(i),
+  End Group Sequence(i),
```
ianswett, line 950: If one makes both of these the max value, does that indicate an open ended subscription?

fluffy:
Commenting on Luke's table of use cases:

Love this way of enumerating and sorting the use cases. Thanks. All the use case I think are critical are covered in the first table and nothing that I did not care about seemed like it was going to complicated implementations so my first reaction was YES. But looking at it, we were getting close to covering 100% of all the cases and since relays are likely to upgrade slower than clients, I just came down to we should just support all the possible combinations even if it is not clear  there is s use case for it today - that will actually make implementations simpler. I think the proposed solutions are starting to look like they do all the cases anyways. By all the cases, I mean a subscribe can 

Start at absolute or relative for both group and object. Relative can have delta to current.  

Start at absolute or relative or fin for both group and object. Relative can have delta to current.

fluffy:
On the topic of do we need unsubscribe hints .... 

How I think of it is subscribe installs some state on the relay or publisher telling it what to do. It can say what the first and last object is that should be forwarded. However, I think as unsubscribe as the message that removes all that state. 

If we start making unsubscribe be "forward until you see object X" then remove all the state, we get into questions like what happens if object X-1 arrives after X. 

That said, I do agree with the use case that Luke is getting at of things like. Client was subscribed to something forever, but now I only want to be subscribe up to the end of the current group so that the client can do a clean switch over to some different subscription at the group boundary

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,164 @@ SUBSCRIBE REQUEST Message {
...
+and object is always determined from the parameters.  An end group is only
+specified when END_MODE is present.  Specifying an END_GROUP or END_OBJECT
+without specifying END_MODE is a Protocol Violation.
+
+The start/end group is determined by examining the START/END_MODE and
+START/END_GROUP (or their defaults) and using the logic above (see
+{{subscribe-group}}).  Whenever START/END_OBJECT is specified, it is an
+absolute object sequence in the determined group. If START/END_MODE is
+RelativePrevious or omitted and START/END_GROUP is 0 or omitted, then
+START/END_OBJECT defaults to `Current Object` + 1 if not specified.  In all
```
fluffy, line 881: I think the "+1" on current object here is a problem. For audio where you don't send packets, you often want the last packet even so that you can see things are in VAD ( Voice Activity Detection ) mode where you are not sending audio. It's only a small tweek to proposal but given there no is way to specify the relative object, I would want therelative object to be the current, not the current+1.

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,164 @@ SUBSCRIBE REQUEST Message {
...
+SUBSCRIBE_REQUEST defines a Track Request Parameter START_POINT (0x0) whose
+value is a varint indicating where in the track to start the subscription.  For
+a live track, the largest group sequence is called the `Current Group` and the
+largest object sequence in that group is the `Current Object`.
+
+|-----------|-------|-------------|--------------|
+| Name      | Value | Start Group | Start Object |
+|----------:|:------|-------------|--------------|
+| Current   | 0x0   | Current     | 0            |
+| Now       | 0x1   | Current     | Current + 1  |
```
fluffy, line 938: Again, I think I would want current  , not current +1 in object. We could add another enum value to have both 

Recent | 0xXXX| Current | Current 
and 
Now | 0xXXX | Current | Current + 1

draft-ietf-moq-transport.md
```diff
@@ -1077,6 +1213,63 @@ Streaming Formats. Each MoQ streaming format can then register its type
 in that table. The MoQ Streaming Format type MUST be carried as the
 leading varint in catalog track objects.
 
+--- back
+
+# Pseudo-code for Interpreting Subscribe Track Request Parameters
+
```
fluffy, line 1219: I'm not keen on adding code like to the draft. If the design is so hard to explain in text that we need code like this, then the design is too complicated. 

A big part of why I don't like code like this is inevitably over time it gets slightly out of sync with the normative code in the draft but lots of people use it and then you have huge conflicts on what the normative specification is and that makes interoperability worse.

draft-ietf-moq-transport.md
fluffy: One not huge change to fix this would be to say that if start/end_object was zero or a positive number, then it was the the absolute object number. If it was negative, then it is a delta applied to the current object +1. ( So a value of -1 gives you the current object, -2 the object before that, and so on ). 

This all feels like it is getting too clever and too complicated.  Not sure I like it.

fluffy (changes requested):
Both of these proposals don't meet the requirement Luke had in his MUST table of starting at latest group and latest-N object.

fluffy:
I was thinking about use case for things like text media some of the stuff with the game moves draft that are not audio and video.  It is pretty common to  want to be able to get the last N things in a group first then later go get the earlier things in a group. The usual rest style pagination type thing where. Imagine you had a group where each object was a message in a chat, you might only want the most recent 10 messages but then when user scrolled back, to get the rest. 

The more I thought about this, I think we have a requirement for 

start group | start object | end group | end object | notes
-- | -- | -- | -- | --
absolute | latest-N | absolute | latest | Get last N object in a given group

kixelated:
> On the topic of do we need unsubscribe hints ....

Yeah, I'm thinking we should use RESUBSCRIBE instead of adding hints to UNSUBSCRIBE. UNSUBSCRIBE would always be an immediate and cleanup any state, while RESUBSCRIBE would be used to update the start/end bounds.

fluffy:
I took the liberty of adding a 3rd proposal to the PR but did not change anything else in the PR.

fluffy:
> > On the topic of do we need unsubscribe hints ....
> 
> Yeah, I'm thinking we should use RESUBSCRIBE instead of adding hints to UNSUBSCRIBE. UNSUBSCRIBE would always be an immediate and cleanup any state, while RESUBSCRIBE would be used to update the start/end bounds.

+1 - that does seem like it will be a very good way to deal with it

draft-ietf-moq-transport.md
afrind: I added the +1 because `Current Object` implies that every hop needs to cache the latest object, or refetch it on subscribe, and I wasn't sure if that was the intention. So "Now" was intended to mean - I'm subscribed and don't need any old objects, just give me the new ones.

> If it was negative

varints are non-negative

> This all feels like it is getting too clever and too complicated. Not sure I like it.

That's also where I ended up when I finished the PR.

draft-ietf-moq-transport.md
afrind: I think that's the net effect, but it's preferred to not use RANGE to issue open ended subscriptions.

kixelated:
@fluffy I wrote this up on my phone before noticing your commit. It's basically your proposal 3, but not using optional parameters.

It would be encoded as:

```
SUBSCRIBE {
  ...
  start_group_hint: (i)
  start_object_hint: (i)
  end_group_hint (i)
  end_object_hint (i)
  ...
}
```

I really like that Alan wrote some pseudo-code, although I agree that it shouldn't be in the final draft. The devil is in the details with this encoding, and I think this scheme is both the simplest and the most powerful.

```python
# Each hint is a VarInt, where the lower two bits specify the mode
def resolve_hint(hint, latest):
  match hint & 0x3:
    case 0x0: # absolute
      return hint >> 2
    case 0x1: # relative
      return max(latest - hint >> 2, 0)
    case 0x2: # future
      return latest + hint >> 2 + 1
    case 0x3: # none
      return Infinity
    
# Convert subscribe hints on the wire into absolute subscribes
def subscribe_hint(start_group, start_object, end_group, end_object): 
  # latest_group: returns max group in cache
  # latest_object: returns max object in cache for given group
  # both will be async when relay has no active subscription 
  start_group = resolve(start_group, latest_group())
  start_object = resolve(start_object, latest_object(start_group))
  end_group = resolve(end_group, latest_group())
  end_object = resolve(end_object, latest_object(end_group))

  # now they can be used with should_serve
  return ( start_group, start_object, end_group, end_object )

# Called for each object (same for all proposals)
def should_serve(obj, start_group, start_object, end_group, end_object)
  if obj.group < start_group or obj.group > end_group:
    return false
  elif obj.group == start_group and obj.sequence < start_object:
    return false
  elif obj.group == end_group and obj.sequence > end_object:
    return false
  else:
    return true 

```

draft-ietf-moq-transport.md
afrind: I wrote the code before I wrote the PR, so I thought I would include it to help illustrate.  It doesn't have to stay.  The design might be too complicated.

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,205 @@ SUBSCRIBE REQUEST Message {
...
+There are 4 track request parameters that indicate group and object for
+the subscription to start and end at. Each of these has a flag that
+indicates if the value is the absolute value of the group or object or
+if it is a relative offset form to the currently largest group or
+currently largest object in the specified group.
+
+|--------------|-------|----------|--------------------|
+|Parameter     | Value | Required | Default            |
+|-------------:|:------|----------|--------------------|
+| START_GROUP_MODE   | 0x1   | No       | Relative   |
```
afrind, line 974: Looks like you are using just two modes (relative and absolute), with a signed delta.

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,205 @@ SUBSCRIBE REQUEST Message {
...
+|--------------|-------|----------|--------------------|
+| END_OBJECT   | 0x8   | No       |  Inf |
+|--------------|-------|----------|--------------------|
+
+The MODE values can be relative or absolute. 
+
+If the mode for a given GROUP_DELTA is absolute, the group to start or
+end at is the absolute value in the DELTA while if the mode is relative,
+then the group is the current group plus the signed value found in the
+delta.
```
afrind, line 996: We need to express a signed integer encoding since varints don't have one, but maybe this is easier.

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,205 @@ SUBSCRIBE REQUEST Message {
...
+|--------------|-------|----------|--------------------|
+| START_GROUP_DELTA  | 0x2   | No       | 0                  |
+|--------------|-------|----------|--------------------|
+| START_OBJECT_MODE   | 0x3   | No       | Relative   |
+|--------------|-------|----------|--------------------|
+| START_OBJECT_DELTA | 0x4   | No       | 0 |
+|--------------|-------|----------|--------------------|
+| END_GROUP_MODE     | 0x5   | No       | Relative               |
+|--------------|-------|----------|--------------------|
+| END_GROUP    | 0x6   | No       | Inf                 |
```
afrind, line 984: I assume you intended the start and end group parameters to be consistently named (either START_GROUP/END_GROUP or START_GROUP_DELAT/END_GROUP_DELTA)?

kixelated:
And I thought it would be a bug to specify start_object=none with my encoding, but it actually seems like a valid use-case.

start group | start object | end group | end object | notes
-- | -- | -- | -- | --
none | 0 | none | none | Check if a track exists

You would get a SUBSCRIBE_OK if the track exists, but the subscription won't actually do anything. It kinda seems useful for specifying intent to, like for pre-warming the cache.

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,205 @@ SUBSCRIBE REQUEST Message {
...
+|--------------|-------|----------|--------------------|
+|Parameter     | Value | Required | Default            |
+|-------------:|:------|----------|--------------------|
+| START_GROUP_MODE   | 0x1   | No       | Relative   |
+|--------------|-------|----------|--------------------|
+| START_GROUP_DELTA  | 0x2   | No       | 0                  |
+|--------------|-------|----------|--------------------|
+| START_OBJECT_MODE   | 0x3   | No       | Relative   |
+|--------------|-------|----------|--------------------|
+| START_OBJECT_DELTA | 0x4   | No       | 0 |
```
afrind, line 980: I had gone on the assumption that I tried to validate in the meeting, which is that relative objects only make sense in the current group, relative to the current object.  I'm a bit surprised to see we need "Nth from end" of a group, given that at least one of the functions of group is a join point.

This encoding also allows some invalid things to be specified, like starting at future objects in old groups: 

START_MODE=Relative, START_GROUP_DELTA=-100, START_OBJECT_MODE=Relative, START_OBJECT_DELTA=100  ==> Start 100 objects after the end of a group that ended a long time ago.

afrind (commented):
Thanks for the feedback and the additional proposal. I think there's a few details that need to be filled in if that's the direction folks want to take.

suhasHere:
> Yeah, I'm thinking we should use RESUBSCRIBE instead of adding hints to UNSUBSCRIBE. UNSUBSCRIBE would always be an immediate and cleanup any state, while RESUBSCRIBE would be used to update the start/end bounds.

My proposal in one of the authors call was Subscribe Update message.

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,205 @@ SUBSCRIBE REQUEST Message {
...
+
+5. Range, All of group 3
+
+Parameters:
+ Start Mode = Absolute, Start Group = 3
+ End Mode = Absolute, End Group = 4
+
+ Start = Group 3, Object 0
+ End = Group 3, Object <last>
+~~~
```
suhasHere, line 921: My least favorite right now is proposal -1 .. This basically looks like original proposal is (which is one varint encdoding 6 values) expanded into 6 different parameters.

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,205 @@ SUBSCRIBE REQUEST Message {
...
+
+Exactly one of START_POINT and RANGE MUST be specified.  If neither or both are
+present it is a Protocol Violation.  All Group and Object Sequence values are
+absolute values within the requested track.  The subscription is for all objects
+starting from the start object up to but not including the end object.
+
+If a publisher cannot satisfy the requested start or end for the subscription it
+MAY send a SUBSCRIBE_ERROR with code TBD.
+
+PROPOSAL 3: More explicit
```
suhasHere, line 963: thing i like about this proposal is having just 2 dimensions to define the combinations of things. This also seems pretty flexible.

I feel we don't need signed value but rather a  direction  indicator instead to compute the resultign object/group -  Increasing/Decreasing Delta 

Having thought about it a bit more , I seem to agree with @fluffy  on Relay update cycles and thus supporting flexibility is a good way to move.

draft-ietf-moq-transport.md
suhasHere: START_MODE=Relative, START_GROUP_DELTA=-100, START_OBJECT_MODE=Relative, START_OBJECT_DELTA=100

The way I read it as
Go back 100 groups from the current group, then start at the 100th   from the beginning of the group

draft-ietf-moq-transport.md
suhasHere: One clarification needed in the text is ,

One should compute target group index first  and then compute the object index. 
If the start_group mode is relative, then either use +ve or -ve delta from the current group to compute the target group
if the start_group mode is absolute, then target group is the group in the request

Same thing is one for target object index.


if

draft-ietf-moq-transport.md
huitema: I think that the real "relative" cases for video are:

- current group, current object (now)
- current group, first object of current group (previous)
- next group, first object (next)

There might be a specific problem for audio, in which receiving N packets before the current one allows for fast re-synchronization of the audio codec. Whether that requires asking for "group N-1" is debatable, but the syntax would be "N objects before current", and treated really like an optional hint.

The option "125 groups in the future" feels like a foot gun, not sure that we need it.

draft-ietf-moq-transport.md
fluffy: True but we can make  a SignedVarInt if we need one. Just about everywhere but QUIC VarInts are signed and UnsignedVarInts are not :-)

draft-ietf-moq-transport.md
fluffy: It did help me understand the definition of how it works.

fluffy:
So @kixelated proposal in the thread above that is similar to mine just seems better than what I proposed.  It covers all the cases, it is easy to describe and validate work and does not have a bunch of uses cases. 

I would put in a very weak argument for instead of taking the bottom two bits to encode the (absolute, relative, future, none ) , instead just put that into it's own VarInt. I would not argue very much for that because this is all going to get put in 5 lines of code I will never look at again and it's does not turn up on the wire in something that has to be small. Probably the worst part of is is someone looking at network dump might be confused but that just means the wireshark dissector needs to know about how to the >>2 works. I'd be totally fine with proposal as is. 

Some strong points I see for this:
1) we could merge it by Monday 
2) It covers all the uses case we can imagine now plus has a high chance of  covering the future cases we can not imagine now
3) It is simple to understand, think about, and test

draft-ietf-moq-transport.md
fluffy: it had not really occurred to me we would not end up having a signed but yes, make sense how you had the mode help express the direction.  I see Luke's proposal does the same which I like better than option 3

draft-ietf-moq-transport.md
fluffy: Oh, crap, yes, mean to have that all be name the same way. SOrry

draft-ietf-moq-transport.md
fluffy: In the meeting I was just thinking about real time interactive audio and video - and I don't see where it would need it. But later I started thinking about a text chat client where each "chat room" was a track, and each thread was a group, and each message was in an object. MIMI would be better with MoQ.  Regardless of if this use case is bogus or not, it does make you think there might be future uses cases were flexibility was good as long as it did not make the current implementations way more complicated.

afrind: > Go back 100 groups from the current group, then start at the 100th from the beginning of the group

I think that would be 

```
START_GROUP_MODE=relative
START_GROUP=-100
START_OBJECT_MODE=*absolute*
START_OBJECT=100
```

> But later I started thinking about a text chat client where each "chat room" was a track, and each thread was a group, and each message was in an object.

This implies a track with multiple active groups.  I guess the current draft doesn't prohibit that, but I'm not sure we're thought through the implications completely.  

I guess I can see "give me the N previous objects from Now (current group)", but struggling to see why I would want N previous objects from the largest one you have in some previous group.  That said, the simplicity of the Luke's proposal is alluring, even if it allows you do request some bizarre things.

afrind (commented):
Per Luke's proposal:

> It would be encoded as:
>
>SUBSCRIBE {
>  ...
>  start_group_hint: (i)
>  start_object_hint: (i)
>  end_group_hint (i)
>  end_object_hint (i)
>  ...
>}

I think I would also use a separate varint for the mode for each param rather that the LSBs.  Keep in mind that taking 2 bits from a varint means only 0-15 can be encoded in a single byte, and we shouldn't be trying to squeeze bytes out of this control message.

```
Hint {
  Mode (i),
  [Value (i)]
}
Modes = None, Relative, Next, Absolute
Value is not present when Mode is None

SUBSCRIBE {
  ...
  start_group_hint: Hint
  start_object_hint: Hint
  end_group_hint: Hint
  end_object_hint: Hint
  ...
}
```
Start Group and Start Object cannot be None.  End Group and Object must either both be None (Closed Range), or Neither (Open Subscribe).

Also bikeshed on whether this is a Hint or a Parameter (not to be confused with Track Request Parameter).  Hint implies that it's ok if the publisher ignores it.

kixelated:
> I think I would also use a separate varint for the mode for each param rather that the LSBs. Keep in mind that taking 2 bits from a varint means only 0-15 can be encoded in a single byte, and we shouldn't be trying to squeeze bytes out of this control message.

A 0-16 range would work for the majority of relative/future hints, so it would definitely save that byte. But I'm fine keeping it verbose for now; the important part is the reusable `Hint` encoding.

> Start Group and Start Object cannot be None. End Group and Object must either both be None (Closed Range), or Neither (Open Subscribe).

Start group = none would be useful for checking track existence. But I'm okay ruling it out for now because it would mean a subscription that matches zero objects.

End object is None to specify the fin of a group. ex. end at the last object of group 69.

> Also bikeshed on whether this is a Hint or a Parameter (not to be confused with Track Request Parameter). Hint implies that it's ok if the publisher ignores it.

I think it's a hint, since the publisher can choose to drop objects/groups on a whim. Even if you ask to start at group 3, you might never get it for congestion control reasons.

But the bigger issue is that a subscriber doesn't know the upstream's expiration time. A publisher may only cache the last 3 groups, so a subscriber can't just ask for the the last 8 groups and expect to receive them all. I think the publisher should signal the (absolute) start/end range in SUBSCRIBE_OK, which may be a subset of the requested range.

draft-ietf-moq-transport.md
fluffy: I'm also totally fine taking the signed relative and using unhinged and replacing relative with two modes of  future and past.

draft-ietf-moq-transport.md
fluffy: Active groups from way back might not happen, but getting the last objects of group N slightly after the first objects of group N+1 seems like it will happen often.  Luke's proposal relies on that working for prioritization

fluffy:
I like the direction that Alan has take Lukes stuff. 

To bikeshed the Hint. I think "Location" would be a better name than "Hint". I think this is not really a hint but more a filter that relay will not send me objects on this subscription before the Start location or after the end Location. So if you want my bikeshedy 0.000002 cents, I would define 

Location {
  GroupMode (i),
  GroupValue (i),
  ObjectMode (i),
  ObjectValue (i)
}
ObjectModes = None, Absolute, Past, Future 
GroupModes = None, Absolute, Past, Future 

The the subscribe would have a Start and End  Location.

Location become a clean way to specify a specific a specific object in a track and may turn out to be useful in other parts of the spec.

fluffy:
<rant> 
I'd like to take a moment to reflect that the notation we are using for is both shitty and bad. 

Take
Hint {
  Mode (i),
  [Value (i)]
}
Modes = None, Relative, Next, Absolute

By shitty, I mean super non intuitive. What does (i) mean? Is [X] optional or an array? But no, that would be ... Can I have more than one X?  If I had Mode(2) followed by [Value(i)], would it be bit packed or would the Mode be padded out to a byte? And what type is Mode. Or if Mode is a type what is then encoding of all that crap. Most schema languages are fairly intuitive to read and guess what they might mean. This is not. 

Then there is the bad. Say I want to define a type called Mode then have two fields in a record called ObjectMode and GroupMode both of type Mode. What do I do?

Location {
GroupMode (i),
ObjectMode (i)
}
ObjectModes = None, Absolute, Past, Future
GroupModes = None, Absolute, Past, Future

This can not possibly be the right way to do it. I mean now that Mode is not reusable and repeats itself. 

If only we had a well defined language for expressing this. Oh wait we do https://datatracker.ietf.org/doc/html/rfc5234

ianswett (commented):
Option 3 looks good, but I'm unsure why we might want different modes for the group and object.  I guess it depends upon how you define how the group modes affects the object?

draft-ietf-moq-transport.md
ianswett: Do we really need future?  Except for 'next group', I thought we didn't have a clear use case, so that'd just be a matter of defining 0 to be the next group.

I also think it'd be useful to have a 3rd mode for 'Head/Latest Group'.  If that was present, then I don't think we'd need a mode for Group?  It could always be absolute.

kixelated:
> Option 3 looks good, but I'm unsure why we might want different modes for the group and object. I guess it depends upon how you define how the group modes affects the object?

I went ahead and enumerated the permutations

group | object | useful | notes
-|-|-|-
absolute | absolute | yes | range request 
absolute | relative | audio* | fetch tail of group 
absolute | future | no | 
absolute | none | yes | receive full group 
relative | absolute | yes | fetch head of group
relative | relative | audio* | fetch tail of live group 
relative | future | audio* | fetch next audio sample(s) 
relative | none | yes | receive full group 
future | absolute | yes | fetch head of group
future | relative | no |  
future | future | no |
future | none | yes | fetch future group 
none | * | empty subscribe

The ones labeled with audio* are only when you have partially independent samples.  For example, each audio GROUP is 2s long and each audio OBJECT is 21ms. 

```
SUBSCRIBE audio start_group=relative/0 start_object=relative/5
```

This would get you the last 105ms of audio (target initial buffer) most of the time. However, you could get unlucky and a new group could have started with less than that available, but it doesn't really matter if the target buffer is small because the player will just wait.

Personally, I want to remove this use-case. If objects can be independent, then they should be separate groups IMO. So each audio frame would be a group of size 21ms. You could then combine the group/object mode since object boundaries can either be absolute, or none (continue until FIN).

afrind:
I pushed a new commit with PROPOSAL 4, which we have been discussing, pulling the "best of" text and examples from the other proposals.  I deleted the PROPOSAL 1 sample code and didn't add code for 4.

There's one DISCUSS, whether we should allow Start/EndObject to be None and have defaults, or require them to be specified.

🤞

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,347 @@ SUBSCRIBE REQUEST Message {
...
+
+DISCUSS: StartObject and EndObject could be optional, with a default of
+Absolute/0, or we can require the subscriber to be explicit.
+
+* EndGroup: The last Group requested in the subscription, inclusive.  EndGroup's
+Mode MAY be None if it is an open-ended subscription.
+
+* EndObject: The last Object requested in the subscription, exclusive.
+EndObject's Mode MUST be None if EndGroup's Mode is None.  EndObject's Mode MUST
+NOT be None if EndGroup's Mode is NOT None.
```
kixelated, line 1082: > EndObject's Mode MUST NOT be None if EndGroup's Mode is NOT None.

I don't think this is true. `end_object=none` means you want the entire group. 

For example, this would request the subscription end once all objects in group 69 have been transmitted:
```
end_group=absolute/69
end_object=none
```

In my head and likely in many implementations, `none` means infinity. `end_object=infinity` means there's no end object.

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,347 @@ SUBSCRIBE REQUEST Message {
...
+On successful subscription, the publisher SHOULD start delivering
+objects from the group sequence and object sequence described below.
+
+If a publisher cannot satisfy the requested start or end for the subscription it
+MAY send a SUBSCRIBE_ERROR with code TBD.
+
+### Examples
+
+~~~
+1. Now
```
kixelated, line 1095: This is technically `Now+1` because the object doesn't exist yet. But yeah I understand it's meant to align with the other proposals.

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,347 @@ SUBSCRIBE REQUEST Message {
...
+====================
+
+PROPOSAL 4
+
+A receiver issues a SUBSCRIBE REQUEST to a publisher to request a track.
+
+### Susbscribe Locations {#susbscribe-locations}
+
+The receiver specifies a start and optional end `Location` for the subscription.
+A location value may be an absolute group or object sequence, or it may be a
```
ianswett, line 1013: This is pre-existing nomenclature, but the use of the word sequence is awkward to me.  Is it shorthand for sequence number?

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,347 @@ SUBSCRIBE REQUEST Message {
...
+None.
+
+* StartObject: The Location of the requested object.  StartObject's Mode MUST NOT
+be None.
+
+DISCUSS: StartObject and EndObject could be optional, with a default of
+Absolute/0, or we can require the subscriber to be explicit.
+
+* EndGroup: The last Group requested in the subscription, inclusive.  EndGroup's
+Mode MAY be None if it is an open-ended subscription.
```
ianswett, line 1078: SHOULD be None for an open-ended subscription?

Or just "is None for an open-ended subscription."

draft-ietf-moq-transport.md
ianswett: I would agree.

draft-ietf-moq-transport.md
```diff
@@ -805,11 +805,347 @@ SUBSCRIBE REQUEST Message {
...
+~~~
+{: #moq-transport-subscribe-format-4 title="MOQT SUBSCRIBE REQUEST Message"}
+
+
+* Full Track Name: Identifies the track as defined in ({{track-name}}).
+
+* StartGroup: The Location of the requested group.  StartGroup's Mode MUST NOT be
+None.
+
+* StartObject: The Location of the requested object.  StartObject's Mode MUST NOT
```
ianswett, line 1071: I think None means you want the whole group, but I guess that's identical to absolute/0 and I'm ok with being explicit.

If I specify a StartGroup of absolute or RelativePrevious/1, then the entire group is available.  What does RelativePrevious or RelativeNext for StartObject in those cases?

ianswett (commented):
4 looks good, though I think there are a lot of cases RelativeNext and RelativePrevious don't make much sense for StartGroup and EndGroup.

fluffy (approved):
Merge it :-)

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -939,11 +982,83 @@ SUBSCRIBE Message {
...
+
+* Track Request Parameters: As defined in {{track-req-params}}.
 
 On successful subscription, the publisher SHOULD start delivering
-objects from the group sequence and object sequence as defined in the
-Parameters.
+objects from the group sequence and object sequence described above.
+
+If a publisher cannot satisfy the requested start or end for the subscription it
+MAY send a SUBSCRIBE_ERROR with code TBD.
```
ianswett, line 1004:
```suggestion
MAY send a SUBSCRIBE_ERROR with code TBD. A publisher MUST NOT send objects
from outside the requested start and end.
```

draft-ietf-moq-transport.md
```diff
@@ -939,11 +982,84 @@ SUBSCRIBE Message {
...
+Start Group: Mode=Absolute, Value=3
+Start Object: Mode=Absolute, Value=0
+End Group: Mode=Absolute, Value=4
+End Object: Mode=Absolute, Value=0
+
+Start = Group 3, Object 0
+End = Group 3, Object <last>
+~~~
+
+TODO: Security Considerations related to these hints
```
ianswett, line 1061:
```suggestion

```

draft-ietf-moq-transport.md
```diff
@@ -1184,6 +1299,7 @@ GOAWAY Message {
   New Session URI (b)
 }
 ~~~
+
```
ianswett, line 1302:
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -1234,7 +1349,6 @@ Streaming Formats. Each MoQ streaming format can then register its type
 in that table. The MoQ Streaming Format type MUST be carried as the
 leading varint in catalog track objects.
```
ianswett, line 1351:
```suggestion


```

PR #294, opened 2023-10-19
Smaller Parameter Reform

### LINKED ISSUE #247: MUST close connection on duplicate parameters

martinduke:
Fixes #247.

People asked for a less controversial version of #256. Here it is, defaulting to the status quo where there is disagreement.

There is some renaming and moving around of sections.

Concrete changes:
- Receivers SHOULD kill the connection on duplicate parameters, instead of MUST, to ease bookkeeping.
- Clarify that receivers ignore unknown parameters
- Clarify what to do when messages contain parameters they shouldn't (ignore)
- Made all integers varints, but retained the redundant length field. Clarify what to do when they don't match (ignore)
- Explicitly state the type of each parameter value

I would like to eliminate the parameter length in varints as in #256, but that does not have consensus.

@kixelated would like to have a separate parameter namespace for each message type, but this also does not have consensus. So I've defaulted to the current arrangement: one for SETUP, and one for SUBSCRIBE/ANNOUNCE, though I renamed the latter.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -730,10 +730,10 @@ parameter using the value in the Parameter Length field.
 
 ### Version Specific Parameters {#version-specific-params}
 
-If a version-specific parameter appears in a message for which its meaning is
-not defined below, it MUST be ignored. Note that since SETUP parameters use a
-separate namespace, it is impossible for these parameters to appear in SETUP
-messages.
+Each version-specific parameter definition indicates the message types in which
+it can appear. If it appears in some other type of message, it MUST be ignored.
+Note that since SETUP parameters use a separate namespace, it is impossible for
+these parameters to appear in SETUP messages.
 
 #### GROUP SEQUENCE Parameter {#group-sequence}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -689,6 +689,74 @@ MOQT Message {
...
+
+SETUP message parameter types are defined in {{setup-params}}. Version-
+specific parameter types are defined in {{version-specific-params}}.
+
+The Parameter Length field of the String Parameter encodes the length
+of the Parameter Value field in bytes.
+
+Each parameter description will indicate the data type in the Parameter Value
+field. If the parameter value is a varint, but the self-encoded length of that
+varint does not match the Parameter Length field, the receiver MUST ignore the
```
ianswett, line 728: I'd argue for closing the connection, but I guess this is ok.

ianswett (approved):
This looks like a good start

draft-ietf-moq-transport.md
martinduke: If the receiver didn't know the parameter it wouldn't know it was a varint and couldn't apply this.

It is totally consistent to say "if you know the parameter is malformed kill the connection" but this seemed easier to write.

ianswett: That makes sense, let's ship it.

PR #297, opened 2023-10-23
Split SETUP messages

### LINKED ISSUE #212: SETUP message type
There's a single SETUP message, but the contents change based on if the endpoint is a server or a client. [See the draft](https://moq-wg.github.io/moq-transport/draft-lcurley-moq-transport.html#name-setup).

This is difficult to implement in type safe languages. I would implement `ClientSetup` and `ServerSetup`, sharing the same ID. However, it's then difficult to implement a generic `decode` method as it needs to know the expected role to distinguish between the two.

I propose a separate `ClientSetup` and `ServerSetup` message akin to how TLS has a separate `ClientHello` and `ServerHello`. Alternatively, implement #138 to reclassify SETUP and OBJECT as control and data stream headers respectively.

vasilvv:
Fixes #212

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -346,7 +346,7 @@ The client can establish a connection to a MoQ server identified by a
 given URI by setting up a QUIC connection to the host and port
 identified by the `authority` section of the URI.  The `path-abempty`
 and `query` portions of the URI are communicated to the server using the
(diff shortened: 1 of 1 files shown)


vasilvv:
Feel free to change the numbers to any ones you prefer.

fluffy (approved):
LGTM

kixelated (approved):

ianswett (approved):

PR #298, opened 2023-10-23
Reinitialize repository

DavidSchinazi:
This PR reinitializes the repository using the [internet-draft-template repo](https://github.com/martinthomson/internet-draft-template) to allow the usual actions to work


afrind (approved):
We may be losing some .editorconfig customizations but we can fix them later

PR #305, opened 2023-10-27
Add Subscribe Id support

suhasHere:
This PR is expected to be base PR to support multi-subscribe like flows that will follow up.

It adds support for SubscribeID and also renames the trackId to `Full Track Name Alias` to avoid the confusion and make the scope clear.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -622,8 +622,8 @@ providing the result of announcement. The entity receiving the
 ANNOUNCE MUST send only a single response to a given ANNOUNCE of
 either ANNOUNCE_OK or ANNOUNCE_ERROR.
 
-OBJECT message header carry short hop-by-hop `Track ID` that maps to the
-Full Track Name. Relays use the `Track ID`
+OBJECT message header carry short hop-by-hop Track ID that maps to the
+Full Track Name (see {{message-subscribe-ok}}). Relays use the Track ID
 of an incoming OBJECT message to identify its track and find the active
 subscribers for that track. Relays MUST NOT depend on OBJECT payload
 content for making forwarding decisions and MUST only depend on the
@@ -1121,7 +1121,7 @@ SUBSCRIBE_ERROR
 ~~~
 {: #moq-transport-subscribe-error format title="MOQT SUBSCRIBE_ERROR Message"}
 
-* SubscribeID: Subscription Identifer as defined in {{message-subscribe-req}}.
+* Subscribe ID: Subscription Identifer as defined in {{message-subscribe-req}}.
 
 * Error Code: Identifies an integer error code for subscription failure.
 
@@ -1145,7 +1145,7 @@ UNSUBSCRIBE Message {
 ~~~
 {: #moq-transport-unsubscribe-format title="MOQT UNSUBSCRIBE Message"}
 
-* SubscribeID: Subscription Identifer as defined in {{message-subscribe-req}}.
+* Subscribe ID: Subscription Identifer as defined in {{message-subscribe-req}}.
 
 ## SUBSCRIBE_FIN {#message-subscribe-fin}
 
@@ -1190,7 +1190,7 @@ SUBSCRIBE_RST Message {
 ~~~
 {: #moq-transport-subscribe-rst format title="MOQT SUBSCRIBE RST Message"}
 
-* SubscribeI D: Subscription Identifier as defined in {{message-subscribe-req}}.
+* Subscribe ID: Subscription Identifier as defined in {{message-subscribe-req}}.
 
 * Error Code: Identifies an integer error code for subscription failure.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1094,7 +1094,7 @@ SUBSCRIBE_OK
 
 * Track ID: Session specific identifier that is used as an alias for the
 Full Track Name in the Track ID field of the OBJECT ({{message-object}})
-message headers of the requested track. Track IDs are generally shorter than
+message headers of the requested track. Track IDs are shorter than
 Full Track Names and thus reduce the overhead in OBJECT messages
 
 * Subscribe ID: Subscription Identifer from the incoming subscription request
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -987,8 +987,8 @@ SUBSCRIBE Message {
 is monotonically increasing variable length integer and it MUST uniquely
 identify a subscription within a session. `Subscribe ID` is used by
 subscribers and the publishers to identify a given subscription. Subscribers
-generate the `Subscribe ID` and it MUST be copied by the publisher when
-responding to the subscription requests.
+specify the `Subscribe ID` and it is included in the corresponding SUBSCRIBE_OK
+or SUBSCRIBE_ERROR.
 
 * StartGroup: The Location of the requested group.  StartGroup's Mode MUST NOT be
 None.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -622,8 +622,8 @@ providing the result of announcement. The entity receiving the
 ANNOUNCE MUST send only a single response to a given ANNOUNCE of
 either ANNOUNCE_OK or ANNOUNCE_ERROR.
 
-OBJECT message header carry short hop-by-hop Track Id that maps to the
-Full Track Name (see {{message-subscribe-ok}}). Relays use the Track ID
+OBJECT message header carry short hop-by-hop `FullTrackName Alias` that maps to the
+Full Track Name (see {{message-subscribe-ok}}). Relays use the `FullTrackName Alias`
 of an incoming OBJECT message to identify its track and find the active
 subscribers for that track. Relays MUST NOT depend on OBJECT payload
 content for making forwarding decisions and MUST only depend on the
@@ -882,23 +882,23 @@ A OBJECT message contains a range of contiguous bytes from from the
 specified track, as well as associated metadata required to deliver,
 cache, and forward it. There are two subtypes of this message. When the
 message type is 0x00, the optional Object Payload Length field is
-present. When the message type ix 0x02, the field is not present.
+present. When the message type is 0x02, the field is not present.
 
 The format of the OBJECT message is as follows:
 
 ~~~
 OBJECT Message {
-  Track ID (i),
+  FullTrackName Alias (i),
   Group Sequence (i),
   Object Sequence (i),
   Object Send Order (i),
-  [Object Payload Length (i),]
+  [Object Payload Length (i)],
   Object Payload (b),
 }
 ~~~
 {: #moq-transport-object-format title="MOQT OBJECT Message"}
 
-* Track ID: The track identifier obtained as part of subscription and/or
+* FullTrackName Alias : The compressed full track name obtained as part of subscription and/or
 publish control message exchanges.
 
 * Group Sequence : The object is a member of the indicated group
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -998,13 +1001,18 @@ NOT be None if EndGroup's Mode is not None.
...
 
+
 On successful subscription, the publisher SHOULD start delivering
 objects from the group sequence and object sequence described above.
 
 If a publisher cannot satisfy the requested start or end for the subscription it
 MAY send a SUBSCRIBE_ERROR with code TBD. A publisher MUST NOT send objects
 from outside the requested start and end.
 
+`SUBSCRIBE` with an existing `SubscribeID` updates the existing subscription and MUST NOT
```
afrind, line 1012: Not sure you can enforce this MUST NOT - 'SHOULD NOT' or 'need not'?

draft-ietf-moq-transport.md
```diff
@@ -982,6 +982,9 @@ SUBSCRIBE REQUEST Message {
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
+* SubscribeID: Session unique identifier for the subscription. `SubscribeID` MUST uniquely identify a subscription within a session and is used by subscribers and publishers to identify a given subscription. Subscribers generate the `SubscribeID` and MUST be
```
afrind, line 985: My recommendation is to be more prescriptive on subscribe ID assignment rather than just uniqueness - even/odd for client/server, monotonically increasing, cannot re-use IDs even after they have expired/closed.

afrind (commented):
Individual Review:

Thanks for this PR.  Can you link it to the corresponding issues?

draft-ietf-moq-transport.md
```diff
@@ -882,24 +882,23 @@ A OBJECT message contains a range of contiguous bytes from from the
...
-  [Object Payload Length (i),]
+  [Object Payload Length (i)],
   Object Payload (b),
 }
 ~~~
 {: #moq-transport-object-format title="MOQT OBJECT Message"}
 
-* Track ID: The track identifier obtained as part of subscription and/or
-publish control message exchanges.
+* Full Track Name Alias : The compressed full track name obtained as part of subscription and/or publish control message exchanges.
```
ianswett, line 901: I'd prefer to keep this as Track ID for now, because it's so much shorter.  Or we could call it 'Track Alias' if that's slightly clearer?  Either way, let's discuss some other names on a PR and then we can do the rename in a separate PR from adding a Subscribe ID.

ianswett (commented):
Thanks for writing this.  I agree we need a Subscribe ID and this looks like it's heading in the right direction.

Can we remove the rename of Track ID from this PR and discuss it separately, since it's purely editorial?

draft-ietf-moq-transport.md
kixelated: I think we prohibit reusing `Subscribe IDs` until we flesh out resubscriptions in a separate PR.

draft-ietf-moq-transport.md
kixelated, line 985: I don't think we need even/odd. Each publisher would have their own number space as they do currently.

But +1 to it being +1 for each SUBSCRIBE. It would be required for a cumulative MAX_SUBSCRIBES if go that direction.

nit: Can you add a space so its `Subscribe ID` instead of `SubscribeID`?

draft-ietf-moq-transport.md
kixelated: +1 to `Track Alias`

draft-ietf-moq-transport.md
```diff
@@ -1067,23 +1075,17 @@ A SUBSCRIBE_OK control message is sent for successful subscriptions.
...
-({{track-name}}).
+* SubscribeID: Subscription Identifer from the incoming subscription request for which this message is the response.
 
-* Track Name: Identifies the track name as defined in ({{track-name}}).
 
-* Track ID: Session specific identifier that is used as an alias for the
-Full Track Name in the Track ID field of the OBJECT ({{message-object}})
-message headers of the requested track. Track IDs are generally shorter
-than Full Track Names and thus reduce the overhead in OBJECT messages.
+* Full Track Name Alias: Session specific identifier that is used as an alias for the Full Track Name in the `Full Track Name Alias` field of the OBJECT ({{message-object}}) message headers of the requested track. `Full Track Name Alias` is generally shorter than Full Track Names and thus reduce the overhead in OBJECT messages.
```
kixelated, line 1088: Can you configure your editor to obey `.editorconfig` and word wrap at 80 characters? It's difficult to determine what changed.

draft-ietf-moq-transport.md
```diff
@@ -1149,18 +1144,14 @@ The format of `SUBSCRIBE_FIN` is as follows:
...
   Final Object (i),
 }
 ~~~
 {: #moq-transport-subscribe-fin-format title="MOQT SUBSCRIBE_FIN Message"}
 
-* Track Namespace: Identifies the namespace of the track as defined in
-({{track-name}}).
-
-* Track Name: Identifies the track name as defined in ({{track-name}}).
+* SubscribeID: Identifies the subscription.
```
kixelated, line 1154: This definition is useless. Can you at least include a link back to the `SUBSCRIBE` section?

kixelated (commented):
The approach is good. A lot of editorial nits.

draft-ietf-moq-transport.md
suhasHere: good point, i will remove this line and add it in the follow up PR i am working on

draft-ietf-moq-transport.md
suhasHere: +1 , i will update to

draft-ietf-moq-transport.md
```diff
@@ -882,24 +882,23 @@ A OBJECT message contains a range of contiguous bytes from from the
...
 message type is 0x00, the optional Object Payload Length field is
-present. When the message type ix 0x02, the field is not present.
+present. When the message type is 0x02, the field is not present.
 
 The format of the OBJECT message is as follows:
 
 ~~~
 OBJECT Message {
-  Track ID (i),
+  Full Track Name Alias (i),
```
afrind, line 891: Individual Comment:

An alternate design is to put the Subscribe ID here, effectively merging it with Full Track Name Alias (I think this was Will's suggestion too).  Subscribe ID becomes an alias both for the name and the other subscribe params.  That simplifies a number of things, but the cost comes if we allow overlapping subscriptions to the same track and want to deduplicate objects among them.

Deduplicating overlapping subscriptions is a pain to the subscriber, who needs to track the ranges of their open subscriptions to know which callers want a given object, and for the server to deduplicate the objects before sending.

fluffy (approved):
LGTM. 

I like how this makes it clear what is the transaction identifier for subscribes and the compression alias for objects full track names.

suhasHere:
@kixelated  @ianswett @afrind wondering if the latest commits address the feedback. thanks

afrind (commented):
The normative bits look fine.  PR needs some cleanup.  

Recommend splitting ID -> Alias to another PR, so this subscribe ID doesn't get caught up with bikeshedding the field name.

vasilvv (approved):
Having reviewed some code that does subscription matching with the current draft, I strongly endorse splitting those.

draft-ietf-moq-transport.md
```diff
@@ -960,14 +960,15 @@ RelativeNext Value:                               0    1  ...
...
 
-The format of SUBSCRIBE REQUEST is as follows:
+The format of SUBSCRIBE is as follows:
 
 ~~~
-SUBSCRIBE REQUEST Message {
+SUBSCRIBE Message {
   Full Track Name Length (i),
   Full Track Name (...),
+  Subscribe ID (i),
```
kixelated, line 971: IMO this should be the first field.

draft-ietf-moq-transport.md
```diff
@@ -1070,6 +1081,7 @@ SUBSCRIBE_OK
   Track Namespace (b),
   Track Name (b),
   Track ID (i),
+  Subscribe ID (i),
```
kixelated, line 1084: IMO this should be the first field.

kixelated (approved):
LGTM but I would like Subscribe ID to be the first field in every `SUBSCRIBE_*` message for consistency.

draft-ietf-moq-transport.md
```diff
@@ -982,6 +983,13 @@ SUBSCRIBE REQUEST Message {
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
+* Subscribe ID: Session unique identifier for the subscription. `Subscribe ID`
+is monotonically increasing variable length integer and it MUST uniquely
+identify a subscription within a session. `Subscribe ID` is used by
+subscribers and the publishers to identify a given subscription. Subscribers
+generate the `Subscribe ID` and it MUST be copied by the publisher when
+responding to the subscription requests.
```
ianswett, line 991:
```suggestion
specify the `Subscribe ID` and it is included in the corresponding SUBSCRIBE_OK
or SUBSCRIBE_ERROR.
```

draft-ietf-moq-transport.md
```diff
@@ -1082,8 +1094,11 @@ SUBSCRIBE_OK
 
 * Track ID: Session specific identifier that is used as an alias for the
 Full Track Name in the Track ID field of the OBJECT ({{message-object}})
-message headers of the requested track. Track IDs are generally shorter
-than Full Track Names and thus reduce the overhead in OBJECT messages.
+message headers of the requested track. Track IDs are generally shorter than
```
ianswett, line 1097:
```suggestion
message headers of the requested track. Track IDs are shorter than
```

draft-ietf-moq-transport.md
```diff
@@ -998,13 +1006,17 @@ NOT be None if EndGroup's Mode is not None.
 * Track Request Parameters: The parameters are defined in
 {{version-specific-params}}
 
+
```
ianswett, line 1009:
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -1082,8 +1093,11 @@ SUBSCRIBE_OK
 
 * Track ID: Session specific identifier that is used as an alias for the
 Full Track Name in the Track ID field of the OBJECT ({{message-object}})
-message headers of the requested track. Track IDs are generally shorter
-than Full Track Names and thus reduce the overhead in OBJECT messages.
+message headers of the requested track. Track IDs are shorter than
+Full Track Names and thus reduce the overhead in OBJECT messages
+
+* Subscribe ID: Subscription Identifer from the incoming subscription request
+for which this message is the response.
```
ianswett, line 1100:
```suggestion
* Subscribe ID: Subscription Identifier specified in the corresponding SUBSCRIBE message.
```

draft-ietf-moq-transport.md
```diff
@@ -1082,8 +1093,10 @@ SUBSCRIBE_OK
 
 * Track ID: Session specific identifier that is used as an alias for the
 Full Track Name in the Track ID field of the OBJECT ({{message-object}})
-message headers of the requested track. Track IDs are generally shorter
-than Full Track Names and thus reduce the overhead in OBJECT messages.
+message headers of the requested track. Track IDs are shorter than
+Full Track Names and thus reduce the overhead in OBJECT messages
+
+* Subscribe ID: Subscription Identifier specified in the corresponding SUBSCRIBE message.
```
ianswett, line 1099:
```suggestion
* Subscribe ID: Subscription Identifer as defined in {{message-subscribe-req}}.
```

ianswett (approved):

PR #307, opened 2023-10-28
Use session more and specify error codes

### LINKED ISSUE #238: Ensure use of connection/session are clear and specify error codes when closing a session.
Part of PR #234 has text "MUST close the connection if there are duplicates"

It seems like we should be closing the session, and we should specify an error code.

ianswett:
Fixes #238 

I also wonder if Connection URL should be renamed to Session URL?

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -388,8 +388,8 @@ peer receiving an Object on the control stream closes the session as a
 
 This draft only specifies a single use of bidirectional streams. Objects are
 sent on unidirectional streams.  Because there are no other uses of
-bidirectional streams, a peer MAY currently close the connection if it
-receives a second bidirectional stream.
+bidirectional streams, a peer MAY currently close the session as a
+'Protocol Violation' if it receives a second bidirectional stream.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -810,7 +810,7 @@ The client offers the list of the protocol versions it supports; the
 server MUST reply with one of the versions offered by the client. If the
 server does not support any of the versions offered by the client, or
 the client receives a server version that it did not offer, the
-corresponding peer MUST close the connection.
+corresponding peer MUST close the session.
```
ianswett, line 813: Should we add a 'VN Failed' error code?

kixelated: I think so.

kixelated (approved):
I would prefer:

> close the session with a 'Protocol Violation'

instead of:

> close the session as a 'Protocol Violation'

fluffy (approved):
LGTM

draft-ietf-moq-transport.md
ianswett: SG, will do in a different PR.

PR #308, opened 2023-10-28
Remove TODOs

ianswett:
I believe some of this is covered by the recently updated GOAWAY section and I'm not sure what else to add here.


kixelated:
I don't think these TODOs have been covered yet. That being said, I would rather we file issues for these so there can be a conversation, rather than litter the draft with TODOs.

kixelated (approved):

ianswett:
I'm unsure what's unresolved, but if you can file new issues as appropriate and link them to this PR, that would be helpful.

Issue #15, opened 2022-10-21
Discontinuities


kixelated:
There's currently no way to reset the timescale. This is very useful when stitching in content, as the new media could start at PTS 0 but still appear continuous. Otherwise the server would need to rewrite the media timestamps. Discontinuities are also used to reset the decoder, although I'm not sure what else it impacts other than the timestamp.

A discontinuity flag for each layer wouldn't work since layers arrive out of order. In order to support discontinuities, you need to know the last timestamp+duration, which may not have arrived yet.

A possible solution is to include `baseTimestamp` in each layer, indicating the decoder should add this value to each timestamp. The only problem is that it would be a little annoying to repeat the base timestamp for each layer. Maybe the base timestamp could be inherited by dependency layer(s)? Not sure.

vasilvv:
I suggest that we have an "epoch" counter.  You start with epoch = 0, and then increment it by one every time you want to reset the timestamps.

We may also want to tie init segments (#21) to the epoch?  E.g. if you want to switch the resolution mid-stream, you would probably want to reset the decoder and add new video format metadata anyways.

kixelated:
Yeah that's my thought as well.

fluffy:
Hmm - I'd think it would be a good idea to sketch out a few design on this and get better idea of all the issue. I lean towards something that is mappable to real ntp time so that things can be debugged as well as synchronized. (note "mapped to" leave open a wide range of possible solutions including probably what you were talking about above )

kixelated:
I definitely want some way to map PTS to wall clock timestamp as well.

kixelated:
Still needs discussion, but most definitely part of the streaming format draft.

PR #316, opened 2023-10-30
Describe how to use datagrams

### LINKED ISSUE #35: QUIC Streams and Datagrams support
Creating this issue to reflect WG discussions. Not sure what's the right next steps, but would be good as a placeholder to further discussions

### LINKED ISSUE #38: Relay latency caused by streams
This was brought up by @suhasHere, and also at IETF 115 by @huitema.

Let's assume a sender wants to transmit a stream with 3k bytes passed through a relay. The relay will read from the source stream and write any data to the destination stream. The sender fragments the stream into 3 packets of 1k bytes each.

1. The 1st packet arrives. The QUIC library will flush the contents to the application, which then writes to the destination stream.
2. The 2nd packet is dropped. If will be retransmitted later when a gap is detected.
3. The 3rd packet arrives. The QUIC library will NOT flush the contents to the application due to the gap.
4. The 2nd packet is retransmitted and arrives. The QUIC library will flush the stream contents of the 2nd and 3rd packet to the application, which then writes to the destination stream.

This relay setup means we do not immediately transmit stream contents as they arrive. Each additional relay exacerbates the issue.

I will note that is not an issue specific to MoQ or even QUIC, as TCP relays suffer the same problem. However it's a larger issue for live media as latency is critical.

ianswett:
Fixes #35

Also fixes #38 because there is existing text about streams:
```
"A relay that reads from a stream and writes to stream in order will
introduce head-of-line blocking.  Packet loss will cause stream data to
be buffered in the library, awaiting in order delivery, which will
increase latency over additional hops.  To mitigate this, a relay SHOULD
read and write stream data out of order subject to flow control
limits.  See section 2.2 in {{QUIC}}."
```

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -313,7 +313,7 @@ over a QUIC connection directly [QUIC], and over WebTransport
 semantics (see {{?I-D.ietf-webtrans-overview, Section 4}}); thus, the
 main difference lies in how the servers are identified and how the
 connection is established.  There is no definition of the protocol
-over other transports, such as TCP, and applicaitons using MoQ might
+over other transports, such as TCP, and applications using MoQ might
 need to fallback to another protocol when QUIC or WebTransport aren't
 available.
 
@@ -669,9 +669,6 @@ limits.  See section 2.2 in {{QUIC}}.
 
 # Messages {#message}
 
-Both unidirectional and bidirectional streams contain sequences of
-length-delimited messages.
-
 An endpoint that receives an unknown message type MUST close the session.
 
 ~~~
@@ -868,6 +865,14 @@ specified track, as well as associated metadata required to deliver,
 cache, and forward it. There are two subtypes of this message. When the
 message type is 0x00, the optional Object Payload Length field is
 present. When the message type ix 0x02, the field is not present.
+
+OBJECT messages are transmitted over unidirectional streams and can
+also be transmitted over datagrams if the object or objects fit into
+a single datagram.  For larger objects, a stream needs to be used.
+Streams are by default reliable and datagrams are by default unreliable,
+but implementations can provide equivalent levels of reliability,
+including the support of a time-to-live parameter that indicates
+when to stop retransmitting lost data from the stream or datagram.
 
 The format of the OBJECT message is as follows:
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -874,7 +874,8 @@ but implementations can provide equivalent levels of reliability,
 including the support of a time-to-live parameter that indicates
 when to stop retransmitting lost data from the stream or datagram.
 The choice of whether to use a datagram or stream is a hop-by-hop
-decision, because MTUs may differ.
+decision, because MTUs may differ.  Datagrams do not have flow control
+limits, but streams do, which may affect the decision of which to use.
 
 The format of the OBJECT message is as follows:
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -873,6 +873,8 @@ Streams are by default reliable and datagrams are by default unreliable,
 but implementations can provide equivalent levels of reliability,
 including the support of a time-to-live parameter that indicates
 when to stop retransmitting lost data from the stream or datagram.
+The choice of whether to use a datagram or stream is a hop-by-hop
+decision, because MTUs may differ.
 
 The format of the OBJECT message is as follows:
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -869,6 +866,14 @@ cache, and forward it. There are two subtypes of this message. When the
 message type is 0x00, the optional Object Payload Length field is
 present. When the message type ix 0x02, the field is not present.
 
+OBJECT messages are transmitted over unidirectional streams and can
+also be transmitted over datagrams if the object or objects fit into
+a single datagram.  For larger objects, a stream needs to be used.
```
afrind, line 871: I think we need more specificity of how the maximum size is determined.  QUIC datagrams can have different maximum sizes (including 0 = disabled) at different hops.

afrind, line 871: `a stream needs to ` or `a stream MUST` ?

draft-ietf-moq-transport.md
```diff
@@ -869,6 +866,14 @@ cache, and forward it. There are two subtypes of this message. When the
...
+Streams are by default reliable and datagrams are by default unreliable,
```
afrind, line 872: Should we mention that streams are subject to limits and datagrams are not?

afrind (commented):
Individual Review: 

Should we provide more guidance for publishers about the tradeoffs for sending an object on a stream vs a datagram?

draft-ietf-moq-transport.md
ianswett: If it doesn't fit, what might the alternatives be?  We can use a MUST if you feel strongly.

ianswett (commented):
Thanks @afrind PTAL

kixelated:
+1 to providing more guidance. Using datagrams instead of streams is the wrong approach unless you really know what you're doing. I'm worried about the ecosystem treating them as interchangeable.

I think we need to be more prescriptive. Streams are the default approach while datagrams are an extension, viable only in very specific circumstances:

1. An OBJECT is below the MTU.
2. Both endpoints have negotiated datagram support in SETUP.
3. Some form of congestion is anticipated, otherwise QUIC datagrams are kind of pointless.
4. A relay that received a stream needs to provide comparable reliability when sending as datagram instead.
5. Datagrams are prioritized like streams, otherwise sending audio as datagrams (and video as streams) might produce the opposite effect.

Personally, I think vanilla QUIC datagrams are always the wrong approach:

- QUIC datagrams with a negotiated FEC/reliability scheme make sense.
- UDP datagrams within a CDN make sense. (lower overhead or multicast?)
- TCP streams within a CDN make sense. (cost savings)
 
But I don't quite understand why you would use a QUIC `DATAGRAM` frame as opposed to a QUIC `STREAM` frame. They're almost identical on the wire except for an incrementing (stream) ID the aforementioned lack of flow control. What's your envisioned use-case @ianswett ?

kixelated:
Also another gotcha: DATAGRAM frames MAY be coalesced together depending on the QUIC implementation. Any application that uses datagrams with the expectation that datagram 1 has independent reliability from datagram 2 will be incorrect without tight QUIC bindings.

I think this is why any sort of FEC and custom reliability should happen at the QUIC layer. If you want to send OBJECTs over datagrams for custom reliability, then it really needs to be done at the QUIC frame level, and you could just use STREAM_FIN instead of DATAGRAM.

afrind:
Maybe move this discussion to an issue rather than a PR so we avoid another unicorn PR.

fluffy (approved):
LGTM. 

I think this will make a significant difference for applications doing interactive audio on networks with significant amount of loss.

suhasHere:
this PR is right starting point. We will refine the text once we have more experience with implementation.

suhasHere (approved):

kixelated (changes requested):
Forgot to leave a -1. I think we need `datagram in == datagram out`.

This PR allows arbitrarily converting between streams and datagrams, but they have different reliability characteristics and it's not clear why you would do this. The text "implementations can provide equivalent levels of reliability" doesn't hold true because of [#15](https://github.com/quicwg/datagram/issues/15); datagram acknowledgements can't quuiiiite be used to provide reliability.

The bigger issue is that an application that sends datagrams to a relay expects to receive datagrams on the other side. This depends on the stream mapping discussion of course, but clearly the current abtraction of "send objects via any mechanism" is not a useful one.

Of course a service is allowed to do whatever it wants inside of their network (including not using QUIC), but we don't need to specify that.

Issue #136, opened 2023-04-19
Messages during handshake


kixelated:
It's unclear what messages are allowed to be sent or received during the SETUP handshake. Here's my intuition:

Until receiving the SETUP, the **client**:
1. MUST NOT send messages
2. MUST buffer any received messages.

Until receiving the SETUP, the **server**:
1. MUST NOT send messages.
2. MUST NOT receive messages.

The client needs to buffer messages because the server is unaware of when the client receives the SETUP. There's no equivalent to HANDSHAKE_DONE, and adding something like that would add an RTT anyway.

kixelated:
A potential optimization exists if the message encoding if the same for all supported versions (invariants). For example, if the client only advertises support for versions 1,2,3 and those versions share the same SUBSCRIBE encoding and meaning. This seems unlikely, given a separate version was created in the first place, and seems very brittle.

If this was the case, then the client could process any invariant messages immediately and buffer the others. However, the client still couldn't send messages unless the server buffered any received messages.

vasilvv:
The most flexible optimization would be the one where you could encode the SUBSCRIBE_REQUEST for every version you offer as a field in SETUP parameters.  TLS does not do exactly that, but there are some things in ClinetHello that are only applicable for certain versions.

PR #324, opened 2023-11-03
Clarify allowed messages during SETUP

afrind:
I think the intent of the draft was clear but this makes it explicit not to open a uni stream before setup exchange is done.

Fixes: #136

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -390,8 +390,6 @@ This draft only specifies a single use of bidirectional streams. Objects are
 sent on unidirectional streams.  Because there are no other uses of
 bidirectional streams, a peer MAY currently close the session as a
 'Protocol Violation' if it receives a second bidirectional stream.
-An endpoint MUST NOT open a unidirectional stream until after it has completed
-the exchange of SETUP messages is complete.
 
 The control stream MUST NOT be abruptly closed at the underlying transport
 layer.  Doing so results in the session being closed as a 'Protocol Violation'.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -377,19 +377,21 @@ separate SETUP parameters for that information in each version.
 
 ## Session initialization {#session-init}
 
-The first stream opened is a client-initiated bidirectional control stream
-where the peers exchange SETUP messages ({{message-setup}}).  All messages
-defined in this draft are sent on the control stream after the SETUP message.
-Control messages MUST NOT be sent on any other stream, and a peer receiving
-a control message on a different stream closes the session as a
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -377,19 +377,21 @@ separate SETUP parameters for that information in each version.
...
+and a peer receiving a control message on a different stream closes the session
+as a 'Protocol Violation'. Objects MUST NOT be sent on the control stream, and a
+peer receiving an Object on the control stream closes the session as a 'Protocol
+Violation'.
 
 This draft only specifies a single use of bidirectional streams. Objects are
 sent on unidirectional streams.  Because there are no other uses of
 bidirectional streams, a peer MAY currently close the session as a
 'Protocol Violation' if it receives a second bidirectional stream.
+An endpoint MUST NOT open a unidirectional stream until after it has completed
```
kixelated, line 393: I think this implies that the server has to wait until SERVER SETUP is acknowledged? It can really only know that when it receives the first non-setup control message.

Really the text should say "until after the receipt of a SUBSCRIBE message". It's used as both an acknowledgement of the SETUP handshake being completed and It's just a protocol violation to send OBJECTs without a SUBSCRIBE anyway. But I'm also fine with "until the receipt of the first non-SETUP message" just to keep things generic for the future.

fluffy (approved):

draft-ietf-moq-transport.md
afrind: Never trust transport acks at the app layer!  Definitely don't want to imply that.  I can update it to:

`An endpoint MUST not open a unidirectional stream or send an OBJECT* message until after receiving a SUBSCRIBE and sending a SUBSCRIBE OK. `

That transitively means you can't do it during setup, since you can't send or receive a subscrbie* until after setup is complete.  Would that work?

kixelated: > That transitively means you can't do it during setup, since you can't send or receive a subscrbie* until after setup is complete.

Mostly true. Now that control messages are sent on the same stream as SETUP, you can abuse head-of-line blocking. If client only supports versions/extensions with the same SUBSCRIBE encoding, then the client can send the SUBSCRIBE immediately:

```
--> SETUP (stream 1)
--> SUBSCRIBE (stream 1)
<-- SETUP (stream 1)
<-- SUBSCRIBE_OK (stream 1)
<-- OBJECT (stream 2)
```

Your proposed text now allows this so that's gud. I think it can be simplified even further:

> An endpoint MUST not send an OBJECT before receiving a corresponding SUBSCRIBE message.

This rule out publisher push, independent of the handshake itself.

ianswett (approved):

PR #329, opened 2023-11-07
Specify comparison for Track Name / Namespace

### LINKED ISSUE #284: Are there restrictions on track names and namespaces?
In HTTP, there are a number of restrictions on what characters are allowed in a path, headers, etc.

I don't see any text restricting track names and track namespaces, but the examples are all ASCII, implying the namespace and name are not intended to be completely opaque and are instead intended to be human readable.

There's nothing in the spec that prohibits them being treated as opaque identifiers, but I'm unclear if that's the intent and I think it'd be good to explicitly say what we intend.

fluffy:
Fixes #284

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -294,6 +294,14 @@ Full Track Name = security-camera.example.com/camera1/hd-video
 
 ~~~
 
+In this specification, both the Track Namespace and the Track Name are
+not constrained to a specific type of encoding. They carry a range of
+bytes and any comparison between two Track Namespaces/Track Names is
+done simply by exact compare of the bytes. Higher level specifications
+may constrain the information in these fields to UTF-8 or portions of
+particular types or URLs. Any specification that does that needs to
+specify the canonicalization into the bytes in the Track Namespace or
+Track Name such that comparison works correctly.
 
 ### Connection URL
 


draft-ietf-moq-transport.md
```diff
@@ -294,6 +294,14 @@ Full Track Name = security-camera.example.com/camera1/hd-video
 
 ~~~
 
+In this specification, both the Track Namespace and the Track Name are
+not constrained to a specific type of encoding. They carry a range of
+bytes and any comparison between two Track Namespaces/Track Names is
+done simply by exact compare of the bytes. Higher level specifications
+may constrain the information in these fields to UTF-8 or portions of
+particular types or URLs. Any specification that does that needs to
```
ianswett, line 302:
```suggestion
may constrain the information in these fields, for example by restricting
them to UTF-8. Any specification that does that needs to
```

ianswett (commented):
Some small suggestions

suhasHere (approved):
Looks fine to me. Agree with most of the nits suggested by Ian

draft-ietf-moq-transport.md
```diff
@@ -294,6 +294,14 @@ Full Track Name = security-camera.example.com/camera1/hd-video
...
 ~~~
 
+In this specification, both the Track Namespace and the Track Name are
+not constrained to a specific encoding. They carry a sequence of
+bytes and comparison between two Track Namespaces or Track Names is
+done by exact comparison of the bytes. Specifications that use MoQ Transport
+may constrain the information in these fields, for example by restricting
+them to UTF-8. Any specification that does that needs to
+specify the canonicalization into the bytes in the Track Namespace or
+Track Name such that comparison works correctly.
```
ianswett, line 304:
```suggestion
Track Name such that exact comparison works.
```

ianswett (approved):

PR #331, opened 2023-11-07
Remove Full Track Name

### LINKED ISSUE #328: Space in names are resulting in definitions that make no sense.
Take the line from your drafts that says 

```
Full Track Name = Track Namespace Track Name
```

There is no valid way to parse the right hand of this. We have things like "Full Track Name Length" and "Full Track Name". 

I do not think we should have ambiguous syntax for things that require precision.  I think it would also be good to have things that can be directly mapped to common programming languages.  This moves beyound simply editorial, or specification is underspecified and ambiguous until we fix this.

ianswett:
It's only used a few times and now that Namespace and Name are an opaque sequence of bytes, it's even less compelling.

Fixes #328

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -252,47 +252,23 @@ In MOQT, every track has a track name and a track namespace associated
 with it.  A track name identifies an individual track within the
 namespace.
 
-A tuple of a track name and a track namespace together is known as a
-full track name:
-
-~~~~~~~~~~~~~~~
-Full Track Name = Track Namespace Track Name
-~~~~~~~~~~~~~~~
-
 A MOQT scope is a set of servers (as identified by their connection
-URIs) for which full track names are guaranteed to be unique.  This
-implies that within a single MOQT scope, subscribing to the same full
-track name would result in the subscriber receiving the data for the
-same track.  It is up to the application using MOQT to define how broad
-or narrow the scope has to be.  An application that deals with
-connections between devices on a local network may limit the scope to a
-single connection; by contrast, an application that uses multiple CDNs
-to serve media may require the scope to include all of those CDNs.
+URIs) for which the tuple of Track Name and Track Namespace are
+guaranteed to be unique. Within a single MOQT scope, subscribing to the
+same Track Name and Track Namespace results in the subscriber receiving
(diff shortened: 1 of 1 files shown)


kixelated (approved):
love it

draft-ietf-moq-transport.md
```diff
@@ -252,47 +252,23 @@ In MOQT, every track has a track name and a track namespace associated
...
-or narrow the scope has to be.  An application that deals with
-connections between devices on a local network may limit the scope to a
-single connection; by contrast, an application that uses multiple CDNs
-to serve media may require the scope to include all of those CDNs.
-
-The full track name is the only piece of information that is used to
-identify the track within a given MOQT scope and is used as cache key.
+URIs) for which the tuple of Track Name and Track Namespace are
+guaranteed to be unique. Within a single MOQT scope, subscribing to the
+same Track Name and Track Namespace results in the subscriber receiving
```
suhasHere, line 258: s/  subscribing to the same Track Name and Track Namespace are guaranteed/  

  ==> subscribing to the same tuple of Track Name and its Namespace results in ...

It is important to know the combination of these 2 (as a tuple) is what identifies a track and just track name, since they are no unique even within a moq scope

suhasHere (changes requested):

draft-ietf-moq-transport.md
```diff
@@ -252,47 +252,21 @@ In MOQT, every track has a track name and a track namespace associated
...
+URIs) for which the tuple of Track Name and Track Namespace are
+guaranteed to be unique and identify a specific track. It is up to
+the application using MOQT to define how broad or narrow the scope is.
+An application that deals with connections between devices
+on a local network may limit the scope to a single connection; by
+contrast, an application that uses multiple CDNs to serve media may
+require the scope to include all of those CDNs.
+
+The tuple of Track Namespace and Track Name identify the track
+within a given MOQT scope and is used as cache key.
```
ianswett, line 265:
```suggestion
Because the tuple of Track Namespace and Track Name are unique within an
MOQT scope, they can be used as a cache key.
```

draft-ietf-moq-transport.md
ianswett: Thanks Suhas, I clarified this and shortened the text.

PR #333, opened 2023-11-08
Object Model to Transport Mapping

### LINKED ISSUE #244: Object Model mapping to QUIC
Recent discussions around the points highlighted in https://datatracker.ietf.org/doc/html/draft-jennings-moq-usages-00 has made it clear that the core transport protocol needs to provide advices on how to moqt object model maps to the QUIC.

Given that different use-cases have slightly varying requirements, there is also need for applications to specify their preference on how to use QUIC streams to transport objects and groups within a track from the publisher to the subscriber. 

Here is the  proposal for  `Transport Delivery Mode` enumeration as part of `SUBSCRIBE OK` message as an **experimental** addition to the protocol for publishers to express their interest .

Following delivery modes are proposed

* StreamPerObject(0): 
In this mode, the publisher intends to use one QUIC Stream per MOQT Object.

* Stream Per Group(1):
In this mode, the publisher intends to use one QUIC Stream per MOQT Group. All the objects from a given Group share the same QUIC stream.

* Stream Per Priority(2):
In this mode, the publisher intends to use one QUIC Stream per priority 
as determined by the MOQT object priority value. 

* Stream Per Track(3):
In this mode, the publisher intends to use one QUIC Stream for a given track.

The core idea behind such experimental extension is to help under the application, deployment and interoperability requirements before the final path is chosen for the core transport

suhasHere:
This PR brings in currently active proposals for supporting explicit indicators of mapping object model to quic transport.

Fixes #244

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -593,7 +593,7 @@ terminates the track with a SUBSCRIBE_FIN
 (see {{message-subscribe-rst}}).
 
 A relay MUST not reorder or drop objects received on a multi-object stream when
-forwarding to subscribers.
+forwarding to subscribers, unless it has application specific information.
 
 Relays MAY aggregate authorized subscriptions for a given track when
 multiple subscribers request the same track. Subscription aggregation
@@ -932,7 +932,8 @@ TODO: figure out how a relay closes these streams
 
 When a stream begins with `STREAM_HEADER_TRACK`, all objects on the stream
 belong to the track requested in the Subscribe message identified by `Subscribe
-ID`.  All objects on the stream have the same `Object Send Order`.
+ID`.  All objects on the stream have the `Object Send Order` specified in the
+stream header.
 
 
 ~~~
@@ -964,7 +965,7 @@ stream that is associated with the subscription, or open a new one and send the
 When a stream begins with `STREAM_HEADER_GROUP`, all objects on the stream
 belong to the track requested in the Subscribe message identified by `Subscribe
 ID` and the group indicated by `Group ID`.  All objects on the stream
-have the same `Object Send Order`.
+have the `Object Send Order` specified in the stream header.
 
 ~~~
 STREAM_HEADER_GROUP Message {
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -897,38 +897,44 @@ NOT be processed by a relay.
 
 The publisher specifies its preference for mapping the elements of the
 object model - Tracks, Groups and Objects, to the underlying transport
-mechanisms such as QUIC Streams and QUIC Datagrams. A publisher chooses it preference driven by the application needs.
+mechanisms such as QUIC Streams and QUIC Datagrams. A publisher chooses it
+preference driven by the application needs.
 
 #### Proposal 1 - Stream Placement Indicator (Victor's Proposal and Text)
 
-`Stream Placement Indicator` specifies as variable length integer indicating
-the publisher's intention when choosing the QUIC stream to transport a given
-object. The details of the same is provided below:
+`Stream Placement Indicator` specifies as variable length integer informing
+the stream to choose for sending a given object.
+
+Each publisher follows the procedure defined below :
 
 Let `N` be the object sequence of the current object and `D` be the
-stream placement indicator, and `S`` be a stack of stream that is initially
+stream placement indicator, and `S` be a stack of streams that is initially
 empty.
 
-For every object to be transported:
+For each object to be sent:
 
-1. If `D` is zero, open a new stream, push it at the top of the stack.
+1. If `D` is zero, open a new stream and push it at the top of the stack.
 Send the object on the stream at the top of the stack.
 
 2. If `D` is non-zero, check if the object sequence of the most recent object
-that sent on the stream at the top of the stack has a value of  `N - D`.
+that was sent on the stream at the top of the stack has a value of  `N - D`.
 
-  2.1 If it is,  continue sending the object on the stream at the top of the
-      stack.
+  -  If it is, continue sending the current object on the stream at the top
+     of the stack.
 
-  2.2 If it is not, close the stream at the top of the stack, pop the stream
-     off the stack and repeat the Step 2. If the stack is empty, close
+
+  - If it is not, close the stream at the top of the stack, pop the stream
+     off the stack and repeat from the Step 2. If the stack is empty, close
      the connection with appropriate error code.
 
+TODO: More details are needed to determine when the stream can be closed.
+      Refer to disucssions on issue 318 on explicit markers for signaling
+      stream completion.
 
 #### Proposal 2 - Object Placement Mode
 
 `Object Placement Mode` is a enumeration of possible mappings of the object
-model to the underlying transport mechanism for delivering media.
+model to the underlying transport mechanisms.
 
 
 ##### Mode: StreamPerGroup
@@ -944,7 +950,7 @@ over a single QUIC stream. `StreamPerObject` has a value of 0x1.
 ##### Mode: StreamPerTrack
 
 The mode `StreamPerTrack` specifies that all the groups and objects for a
-given track is sent over a single QUIC stream. `StreamPerTrack` has a
+given track are sent over a single QUIC stream. `StreamPerTrack` has a
 value of 0x2.
 
 ##### Mode: StreamPerPriority
@@ -957,6 +963,10 @@ value of 0x3.
 
 The mode `Datagram` specifies that objects are sent using QUIC datagrams and
 has a value of 0x4.
+
+Note: Certain considerations apply when using Datagram. See the
+section on datagram that follows from discussions in the PR316.s
+under the PR 316
 
 ## SUBSCRIBE {#message-subscribe-req}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -862,6 +862,7 @@ OBJECT Message {
   Track Alias (i),
   Group ID (i),
   Object ID (i),
+  Object Delivery Preference (i),
   Object Send Order (i),
   [Object Payload Length (i)],
   Object Payload (b),
@@ -879,6 +880,9 @@ OBJECT Message {
 IDs starts at 0, increasing sequentially for each object within the
 group.
 
+* Object Delivery Preference: Publisher intended delivery preference,
+See {{delivery-prefs}}.
+
 * Object Send Order: An integer indicating the object send order
 {{send-order}} or priority {{ordering-by-priorities}} value.
 
@@ -888,6 +892,71 @@ field is absent, the object payload continues to the end of the stream.
 * Object Payload: An opaque payload intended for the consumer and SHOULD
 NOT be processed by a relay.
 
+
+### Object Delivery Preferences {#delivery-prefs}
+
+The publisher specifies its preference for mapping the elements of the
+object model - Tracks, Groups and Objects, to the underlying transport
+mechanisms such as QUIC Streams and QUIC Datagrams. A publisher chooses it preference driven by the application needs.
+
+#### Proposal 1 - Stream Placement Indicator (Victor's Proposal and Text)
+
+`Stream Placement Indicator` specifies as variable length integer indicating
+the publisher's intention when choosing the QUIC stream to transport a given
+object. The details of the same is provided below:
+
+Let `N` be the object sequence of the current object and `D` be the
+stream placement indicator, and `S`` be a stack of stream that is initially
+empty.
+
+For every object to be transported:
+
+1. If `D` is zero, open a new stream, push it at the top of the stack.
+Send the object on the stream at the top of the stack.
+
+2. If `D` is non-zero, check if the object sequence of the most recent object
+that sent on the stream at the top of the stack has a value of  `N - D`.
+
+  2.1 If it is,  continue sending the object on the stream at the top of the
+      stack.
+
+  2.2 If it is not, close the stream at the top of the stack, pop the stream
+     off the stack and repeat the Step 2. If the stack is empty, close
+     the connection with appropriate error code.
+
+
+#### Proposal 2 - Object Placement Mode
+
+`Object Placement Mode` is a enumeration of possible mappings of the object
+model to the underlying transport mechanism for delivering media.
+
+
+##### Mode: StreamPerGroup
+
+The mode `StreamPerGroup` specifies that all the objects for a given group
+are sent over a single QUIC stream. `StreamPerGroup` has a value of 0x0.
+
+##### Mode: StreamPerObject
+
+The mode `StreamPerObject` specifies that each object is sent
+over a single QUIC stream. `StreamPerObject` has a value of 0x1.
+
+##### Mode: StreamPerTrack
+
+The mode `StreamPerTrack` specifies that all the groups and objects for a
+given track is sent over a single QUIC stream. `StreamPerTrack` has a
+value of 0x2.
+
+##### Mode: StreamPerPriority
+
+The mode `StreamPerPriority` specifies that objects marked with a
+given priorirty shared the same underlying QUIC stream. `StreamPerTrack` has a
+value of 0x3.
+
+##### Mode: Datagram
+
+The mode `Datagram` specifies that objects are sent using QUIC datagrams and
+has a value of 0x4.
 
 ## SUBSCRIBE {#message-subscribe-req}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -852,127 +852,9 @@ the `query` portion of the URI to the parameter.
 
 ## OBJECT {#message-object}
 
-A OBJECT message contains a range of contiguous bytes from from the
+An OBJECT message contains a range of contiguous bytes from from the
 specified track, as well as associated metadata required to deliver,
-cache, and forward it. There are two subtypes of this message. When the
-message type is 0x00, the optional Object Payload Length field is
-present. When the message type is 0x02, the field is not present.
-
-The format of the OBJECT message is as follows:
-
-~~~
-OBJECT Message {
-  Track Alias (i),
-  Group ID (i),
-  Object ID (i),
-  Object Delivery Preference (i),
-  Object Send Order (i),
-  [Object Payload Length (i)],
-  Object Payload (...),
-}
-~~~
-{: #moq-transport-object-format title="MOQT OBJECT Message"}
-
-* Track Alias :The track identifier as specified in the
-  SUBSCRIBE message {{message-subscribe-req}}.
-
-* Group ID : The object is a member of the indicated group ID
-{{model-group}} within the track.
-
-* Object ID: The order of the object within the group.  The
-IDs starts at 0, increasing sequentially for each object within the
-group.
-
-* Object Delivery Preference: Publisher intended delivery preference,
-See {{delivery-prefs}}.
-
-* Object Send Order: An integer indicating the object send order
-{{send-order}} or priority {{ordering-by-priorities}} value.
-
-* Object Payload Length: The length of the following Object Payload. If this
-field is absent, the object payload continues to the end of the stream.
-
-* Object Payload: An opaque payload intended for the consumer and SHOULD
-NOT be processed by a relay.
-
-
-### Object Delivery Preferences {#delivery-prefs}
-
-The publisher specifies its preference for mapping the elements of the
-object model - Tracks, Groups and Objects, to the underlying transport
-mechanisms such as QUIC Streams and QUIC Datagrams. A publisher chooses it
-preference driven by the application needs.
-
-#### Proposal 1 - Stream Placement Indicator (Victor's Proposal and Text)
-
-`Stream Placement Indicator` specifies as variable length integer informing
-the stream to choose for sending a given object.
-
-Each publisher follows the procedure defined below :
-
-Let `N` be the object sequence of the current object and `D` be the
-stream placement indicator, and `S` be a stack of streams that is initially
-empty.
-
-For each object to be sent:
-
-1. If `D` is zero, open a new stream and push it at the top of the stack.
-Send the object on the stream at the top of the stack.
-
-2. If `D` is non-zero, check if the object sequence of the most recent object
-that was sent on the stream at the top of the stack has a value of  `N - D`.
-
-  -  If it is, continue sending the current object on the stream at the top
-     of the stack.
-
-
-  - If it is not, close the stream at the top of the stack, pop the stream
-     off the stack and repeat from the Step 2. If the stack is empty, close
-     the connection with appropriate error code.
-
-
-#### Proposal 2 - Object Placement Mode
-
-`Object Placement Mode` is a enumeration of possible mappings of the object
-model to the underlying transport mechanisms.
-
-
-##### Mode: StreamPerGroup
-
-The mode `StreamPerGroup` specifies that all the objects for a given group
-are sent over a single QUIC stream. `StreamPerGroup` has a value of 0x0.
-
-##### Mode: StreamPerObject
-
-The mode `StreamPerObject` specifies that each object is sent
-over a single QUIC stream. `StreamPerObject` has a value of 0x1.
-
-##### Mode: StreamPerTrack
-
-The mode `StreamPerTrack` specifies that all the groups and objects for a
-given track are sent over a single QUIC stream. `StreamPerTrack` has a
-value of 0x2.
-
-##### Mode: StreamPerPriority
-
-The mode `StreamPerPriority` specifies that objects marked with a
-given priorirty shared the same underlying QUIC stream. `StreamPerTrack` has a
-value of 0x3.
-
-##### Mode: Datagram
-
-The mode `Datagram` specifies that objects are sent using QUIC datagrams and
-has a value of 0x4.
-
-Note: Certain considerations apply when using Datagram. See the
-section on datagram that follows from discussions in the PR316.s
-under the PR 316
-
-TODO: More details are needed to determine when a stream can be closed.
-      Refer to disucssions on issue 318 on explicit markers for signaling
-      stream completion.
-
-## OBJECT (Proposal 3)
+cache, and forward it.
 
 ### Canonical Object Fields
 
@@ -980,11 +862,11 @@ A canonical MoQ Object has the following information:
 
 * Track Namespace and Track Name: The track this object belongs to.
 
-* Group Sequence: The object is a member of the indicated group
+* Group ID: The object is a member of the indicated group ID
 {{model-group}} within the track.
 
-* Object Sequence: The order of the object within the group.  The
-sequence starts at 0, increasing sequentially for each object within the
+* Object ID: The order of the object within the group.  The
+IDs starts at 0, increasing sequentially for each object within the
 group.
 
 * Object Send Order: An integer indicating the object send order
@@ -1013,13 +895,25 @@ serialize object fields below, and terminate the stream.
 ~~~
 OBJECT_STREAM Message {
   Subscribe ID (i),
-  Group Sequence (i),
-  Object Sequence (i),
+  Track Alias (i),
+  Group ID (i),
+  Object ID (i),
   Object Send Order (i),
   Object Payload (...),
 }
 ~~~
 {: #moq-transport-object-stream-format title="MOQT OBJECT_STREAM Message"}
+
+* Subscribe ID: Subscription Identifer as defined in {{message-subscribe-req}}.
+
+* Track Alias: Identifies the Track Namespace and Track Name as defined in
+{{message-subscribe-req}}.
+
+The Track Namespace and Track Name identified by the Track Alias is different
+from the one identified by Subscribe ID, the receiver MUST close the session
+with a Protocol Violation.
+
+* Other fields: As described in {{canonical-object-fields}}.
 
 ### Multi-Object Streams
 
@@ -1056,8 +950,8 @@ stream that is associated with the subscription, or open a new one and send the
 
 ~~~
 {
-  Group Sequence (i),
-  Object Sequence (i),
+  Group ID (i),
+  Object ID (i),
   Object Payload Length (i),
   Object Payload (...),
 }
@@ -1066,13 +960,13 @@ stream that is associated with the subscription, or open a new one and send the
 
 When a stream begins with `STREAM_HEADER_GROUP`, all objects on the stream
 belong to the track requested in the Subscribe message identified by `Subscribe
-ID` and the group indicated by `Group Sequence`.  All objects on the stream
+ID` and the group indicated by `Group ID`.  All objects on the stream
 have the same `Object Send Order`.
 
 ~~~
 STREAM_HEADER_GROUP Message {
   Subscribe ID (i),
-  Group Sequence (i)
+  Group ID (i)
   Object Send Order (i)
 }
 ~~~
@@ -1082,13 +976,13 @@ All Objects received on a stream opened with `STREAM_HEADER_GROUP` have an
 `Object Forwarding Preference` = `Group`.
 
 To send an Object with `Object Forwarding Preference` = `Group`, find the open
-stream that is associated with the subscription, `Group Sequence` and `Object
+stream that is associated with the subscription, `Group ID` and `Object
 Send Order`, or open a new one and send the `STREAM_HEADER_GROUP` if needed,
 then serialize the following fields.
 
 ~~~
 {
-  Object Sequence (i),
+  Object ID (i),
   Object Payload Length (i),
   Object Payload (...),
 }
@@ -1102,11 +996,12 @@ Sending a track on one stream:
 ~~~
 STREAM_HEADER_TRACK {
   Subscribe ID = 1
+  Track Alias = 1
   Object Send Order = 0
 }
 {
-  Group Sequence = 0
-  Object Sequence = 0
+  Group ID = 0
+  Object ID = 0
   Object Payload Length = 4
   Payload = "abcd"
 }
@@ -1126,6 +1021,7 @@ Stream = 2
 
 STREAM_HEADER_GROUP {
   Subscribe ID = 2
+  Track Alias = 2
   Group Sequence = 0
   Object Send Order = 0
 }
@@ -1144,6 +1040,7 @@ Stream = 6
 
 OBJECT_STREAM {
   Subscribe ID = 2
+  Track Alias = 2
   Group Sequence = 0
   Object Sequence = 1
   Payload = "moqrocks"
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -865,7 +865,7 @@ OBJECT Message {
   Object Delivery Preference (i),
   Object Send Order (i),
   [Object Payload Length (i)],
-  Object Payload (b),
+  Object Payload (...),
 }
 ~~~
 {: #moq-transport-object-format title="MOQT OBJECT Message"}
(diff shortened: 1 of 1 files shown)


afrind:
Please keep discussion of the concepts on the issue #244

fluffy:
If there are any other proposed solutions to this, would be great to have them documented as additional proposal in this PR.

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,82 @@ field is absent, the object payload continues to the end of the stream.
...
+  -  If it is, continue sending the current object on the stream at the top
+     of the stack.
+
+
+  - If it is not, close the stream at the top of the stack, pop the stream
+     off the stack and repeat from the Step 2. If the stack is empty, close
+     the connection with appropriate error code.
+
+
+#### Proposal 2 - Object Placement Mode
```
kixelated, line 935: It's so verbose to put this mode in every OBJECT. There's also a plethora of collisions that make it almost impossible to use multiple modes.

What about a single mode per track in SUBSCRIBE_OK instead?

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,82 @@ field is absent, the object payload continues to the end of the stream.
...
+`Object Placement Mode` is a enumeration of possible mappings of the object
+model to the underlying transport mechanisms.
+
+
+##### Mode: StreamPerGroup
+
+The mode `StreamPerGroup` specifies that all the objects for a given group
+are sent over a single QUIC stream. `StreamPerGroup` has a value of 0x0.
+
+##### Mode: StreamPerObject
```
kixelated, line 946: If I set this mode for an OBJECT that would then prohibit:

- any OBJECT in the group from using StreamPerGroup
- any OBJECT in the track from using StreamPerTrack
- any OBJECT in the priority from using StreamPerPriority

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,82 @@ field is absent, the object payload continues to the end of the stream.
...
+
+#### Proposal 2 - Object Placement Mode
+
+`Object Placement Mode` is a enumeration of possible mappings of the object
+model to the underlying transport mechanisms.
+
+
+##### Mode: StreamPerGroup
+
+The mode `StreamPerGroup` specifies that all the objects for a given group
```
kixelated, line 943: All OBJECTs in the specific group MUST also use mode StreamPerGroup.

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,82 @@ field is absent, the object payload continues to the end of the stream.
...
+are sent over a single QUIC stream. `StreamPerGroup` has a value of 0x0.
+
+##### Mode: StreamPerObject
+
+The mode `StreamPerObject` specifies that each object is sent
+over a single QUIC stream. `StreamPerObject` has a value of 0x1.
+
+##### Mode: StreamPerTrack
+
+The mode `StreamPerTrack` specifies that all the groups and objects for a
```
kixelated, line 953: All OBJECTs in the specified track MUST also use mode StreamPerTrack

kixelated (changes requested):

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,82 @@ field is absent, the object payload continues to the end of the stream.
...
+     the connection with appropriate error code.
+
+
+#### Proposal 2 - Object Placement Mode
+
+`Object Placement Mode` is a enumeration of possible mappings of the object
+model to the underlying transport mechanisms.
+
+
+##### Mode: StreamPerGroup
```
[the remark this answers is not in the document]
suhasHere: Stream per group, every object that belongs to Group 1, send it on the stream that is assigned for Group 1.  So where is the out of order scenario coming from ? 

Gaps has nothing to do with the transport mode mapping. A gap can be created due to object being dropped by a node on the path and if there exists an object for a group, it gets sent on the stream associated with the group

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,82 @@ field is absent, the object payload continues to the end of the stream.
...
+
+##### Mode: StreamPerTrack
+
+The mode `StreamPerTrack` specifies that all the groups and objects for a
+given track are sent over a single QUIC stream. `StreamPerTrack` has a
+value of 0x2.
+
+##### Mode: StreamPerPriority
+
+The mode `StreamPerPriority` specifies that objects marked with a
```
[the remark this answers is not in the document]
suhasHere: that's right

draft-ietf-moq-transport.md
suhasHere: If you want send all the objects for a track over single stream , you set the mode as StreamPerTrack,
if you want to send all the objects in a group together, you set StreamPerGroup,
if you want to send objects on independent streams, you mark them as StreamPerObject.

draft-ietf-moq-transport.md
suhasHere: that's the intention with the existing statement

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,82 @@ field is absent, the object payload continues to the end of the stream.
...
 
 
+### Object Delivery Preferences {#delivery-prefs}
+
+The publisher specifies its preference for mapping the elements of the
+object model - Tracks, Groups and Objects, to the underlying transport
+mechanisms such as QUIC Streams and QUIC Datagrams. A publisher chooses it
+preference driven by the application needs.
+
+#### Proposal 1 - Stream Placement Indicator (Victor's Proposal and Text)
```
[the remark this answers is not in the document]
suhasHere: Could you please clarify the major problems that you see ? also what do you see the concerns with different hops having different mappings ( not that I am proposing we allow that , but curious to understand more)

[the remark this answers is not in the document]
kixelated: As Cullen said at the mic, it's just a bug if the relay doesn't forward streams as they were intended.

Packet loss will delay the delivery of a stream relative to others. This algorithm is quite stateful and works based on the most recently received object (regardless of stream) so it will race like crazy and produce wildly unexpected results.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
kixelated: > Stream per group, every object that belongs to Group 1, send it on the stream that is assigned for Group 1. So where is the out of order scenario coming from ?

If objects arrive on different streams then they can arrive out of intended order. If we force implicit == explicit then this isn't an issue, but the text should mention that.

> Gaps has nothing to do with the transport mode mapping. A gap can be created due to object being dropped by a node on the path and if there exists an object for a group, it gets sent on the stream associated with the group

That's fair, but then we need another mode with no gaps. An application that uses delta updates cannot tolerate an arbitrary hole in the middle.

draft-ietf-moq-transport.md
kixelated: Yeah sorry I was suggesting that you add it.

draft-ietf-moq-transport.md
kixelated: To clarify, I'm talking about when OBJECTs could have different modes. If you want ALL objects to use the same mode then that shouldn't be encoded per-OBJECT, it could be in SUBSCRIBE_OK instead.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
kixelated: Sorry, is this a stream...

- per priority per track?
- per priority per namespace?
- per priority (per session)?

The last one is my interpretation of the text and it's impossible because some other namespace/track will definitely want to use another mode.

afrind:
My recommendation is to pause a bit on this PR. We need more normative language, explanatory prose, and examples to help readers understand and evaluate the proposals without breaking github with a long chain of questions and comments. 

The outcome from Monday's meeting was that the chairs were going to designate a small group to put the proposal forward.  I'll circle with Ian and Ted.

fluffy:
> The outcome from Monday's meeting was that the chairs were going to designate a small group to put the proposal forward. I'll circle with Ian and Ted.

OK - will wait for that

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+object. The preferences are Track, Group, Object, Priority and Datagram.  An
+Object MUST be sent according to its `Object Forwarding Preference`, described
+below.
+
+* Object Payload: An opaque payload intended for the consumer and SHOULD
+NOT be processed by a relay.
+
+### Object Message Formats
+
+An `OBJECT_STREAM` message carries a single object on a stream.  There is no
```
suhasHere, line 1005: s/on a stream/on a QUIC stream

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+explicit length of the payload; it is determined by the end of the stream.  An
+`OBJECT_STREAM` message MUST be the first message on a unidirectional stream.
```
suhasHere, line 1007: I think we should also say , there MUST be only occurrence of `OBJECT_STREAM` message else its an protocol error.

draft-ietf-moq-transport.md
afrind: I left out QUIC to account for WebTransport or QUIC.  I think this matches elsewhere now?  In an earlier draft I said "transport stream".

draft-ietf-moq-transport.md
afrind: Is that not implied by saying that the payload length is determined by the length of the stream?

kixcord:
This is all just so confused. The number of messages, modes, and permutations are exploding. Some ideas are in the right direction (ex. stream header) but the hodgepodge of modes/messages is not.

I'd like to explicitly add proposal 4: a relay forwards data streams.

- when a relay receives an upstream stream, it creates a downstream stream (for each subscriber)
- when a relay reads an OBJECT from an upstream, it writes it to the corresponding downstream if it matches the SUBSCRIBE criteria.
- when the upstream is closed, the corresponding downstreams are closed with the same error code or FIN.
- when a relay receives an upstream DATAGRAM, it sends it downstream if it matches the SUBSCRIBE criteria.

No wire changes. I would like a stream header eventually but it's more for compression and to prevent foot guns (ex. prioritization).

This unlocks experimentation as the application can perform any stream mapping. The relay doesn't care and remains simple. The important part is the guarantee that the relay (middlebox) won't rearrange or reorder shit.

afrind:
> I'd like to explicitly add proposal 4: a relay forwards data streams.
> ...
> No wire changes. I would like a stream header eventually but it's more for compression and to prevent foot guns (ex. prioritization).

The consensus in Prague was to move forward with a proposal that uses explicit signals for now, get more implementation experience and data, and revisit. Proposal 4 only uses implicit signals from the QUIC layer rather than explicit signals from the MoQ layer.  I think you can achieve similar outcomes with proposal 3 by adding the appropriate STREAM_HEADER at the beginning of each stream?

suhasHere:
+1 to @afrind's proposal. Let's clean it up and get it ready to merge.

draft-ietf-moq-transport.md
ianswett, line 1007:
```suggestion
`OBJECT_STREAM` message MUST be the first and only message on a unidirectional stream.
```

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+
+An Object received in an `OBJECT_STREAM` message has an `Object Forwarding
+Preference` = `Object`.
+
+To send an Object with `Object Forwarding Preference` = `Object`, open a stream,
+serialize object fields below, and terminate the stream.
+
+~~~
+OBJECT_STREAM Message {
```
ianswett, line 1016: Don't we need a type value so if we receive a stream, we know what type of stream it is?

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+An `OBJECT_DATAGRAM_PREFERRED` message carries a single object in a datagram or
+a stream. There is no explicit length of the payload; it is determined by the
+length of the datagram or stream.  If this message appears on a stream, it MUST
+be the first message on a unidirectional stream.
+
+An Object received in an `OBJECT_DATAGRAM_PREFERRED` message has an `Object
+Forwarding Preference` = `Datagram`.
+
+To send an Object with `Object Forwarding Preference` = `Datagram`, serialize
+the fields and payload, and compare the serialized size with the maximum
```
ianswett, line 1035:
```suggestion
To send an Object with `Object Forwarding Preference` = `Datagram`, determine
the length of the fields and payload, and compare the length with the maximum
```

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+datagram size of the session.  If the object size is less than or equal maximum
+datagram size, send the serialized data as a datagram.  Otherwise, open a
+stream, send the serialized data and terminate the stream.  An implementation
```
ianswett, line 1038:
```suggestion
datagram size of the session.  If the object size is less than or equal maximum
datagram size, send the serialized data as a datagram.  Otherwise, the object on a
unidirectional stream as you would an OBJECT_STREAM message.  An implementation
```

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+  Object Payload (...),
+}
+~~~
+{: #object-datagram-format title="MOQT OBJECT_DATAGRAM_PREFERRED"}
+
+### Multi-Object Streams
+
+When multiple objects are sent on a stream, the stream begins with a stream
+header message and is followed by one or more sets of serialized object fields.
+If a stream ends in the middle of a serialized Object, terminate the session
```
ianswett, line 1057: What if the stream is reset?

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+belong to track indicated by `Track Alias`.
+
+~~~
+STREAM_HEADER_TRACK Message {
+  Track Alias (i)
+}
+~~~
+{: #stream-header-track-format title="MOQT STREAM_HEADER_TRACK Message"}
+
+All Objects received on a stream opened with STREAM_HEADER_TRACK have an `Object
```
ianswett, line 1072: TRACK_STREAM or just TRACK would be simpler to me.

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+  Group Sequence (i),
+  Object Sequence (i),
+  Object Send Order (i),
+  Object Payload Length (i),
+  Object Payload (...),
+}
+~~~
+{: #object-track-format title="MOQT Track Stream Object Fields"}
+
+When a stream begins with `STREAM_HEADER_GROUP`, all objects on the stream
```
ianswett, line 1090: How about GROUP?  Also, it'd be more natural to put this above TRACK in terms of ordering.

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+}
+~~~
+{: #stream-header-track-format title="MOQT STREAM_HEADER_TRACK Message"}
+
+All Objects received on a stream opened with STREAM_HEADER_TRACK have an `Object
+Forwarding Preference` = `Track`.
+
+To send an Object with `Object Forwarding Preference` = `Track`, find the open
+stream that is associated with the `Track Alias`, or open a new one and send the
+`STREAM_HEADER_TRACK` if needed, then serialize the the following object fields.
```
ianswett, line 1077: How do I know if they're needed?

Do we have one Object message at the beginning of the stream after the 'type' is serialized and then omit the duplicate fields on subsequent Object messages?

Or do we have a 'message' that starts the stream that could be only the type, followed by the shared values, and then a sequence of objects?  They're fairly equivalent, but we should pick one.

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+  Object Payload (...),
+}
+~~~
+{: #object-group-format title="MOQT Group Stream Object Fields"}
+
+When a stream begins with `STREAM_HEADER_PRIORITY`, all objects on the stream
+belong have the same `Track Alias`, `Group Sequence` and `Object Send Order`.
+
+~~~
+STREAM_HEADER_PRIORITY Message {
```
ianswett, line 1124: I don't think we need this.  I'd rather have every stream have a single priority.  If we want to do that later we can, but we all agree that mixing priorities on a single stream is problematic at best.

ianswett (commented):
3 LGTM overall, but I had some comments on the details.

draft-ietf-moq-transport.md
afrind: > How do I know if they're needed?

If you have to open a stream for this, then you need to send a header.

> Or do we have a 'message' that starts the stream that could be only the type, followed by the shared values, and then a sequence of objects?

This one.

draft-ietf-moq-transport.md
afrind: Ok

draft-ietf-moq-transport.md
afrind: > I'd rather have every stream have a single priority. 

If that's the case, I think you want this message and remove the GROUP one that allows a different send order per object.

kixelated: +1 priority should definitely be hoisted to the stream header.

There's an argument that a relay could perform application-level prioritization by reordering objects on a stream, but that suffers from head-of-line blocking and contradicts the ordering/reliability that QUIC streams are supposed to provide.

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+Object MUST be sent according to its `Object Forwarding Preference`, described
+below.
+
+* Object Payload: An opaque payload intended for the consumer and SHOULD
+NOT be processed by a relay.
+
+### Object Message Formats
+
+An `OBJECT_STREAM` message carries a single object on a stream.  There is no
+explicit length of the payload; it is determined by the end of the stream.  An
```
kixcord, line 1006: I would actually be in favor of removing this message type.

I'm using objects with an unbounded length to get around the lack of stream mapping in the current draft. The idea is that an OBJECT_STREAM can't be moved or combined with other streams, so it's a way to actually enforce stream mapping across relays.

However we don't need it any longer if there's explicit stream mapping and ordering/reliability. I would be fine with each OBJECT having a specified length (like HTTP/3 DATA frames) provided the relay can't rearrange them. It does mean that we have to buffer each frame to determine the length but that's often unavoidable.

draft-ietf-moq-transport.md
kixcord, line 1038: We do need to send RESET_STREAM for flow control reasons. What error code do we use? Do we need to indicate why the datagram was dropped (ex. flow control)?

Does the receiver also reply with a STOP_SENDING?

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+To send an Object with `Object Forwarding Preference` = `Datagram`, serialize
+the fields and payload, and compare the serialized size with the maximum
+datagram size of the session.  If the object size is less than or equal maximum
+datagram size, send the serialized data as a datagram.  Otherwise, open a
+stream, send the serialized data and terminate the stream.  An implementation
+MUST NOT send an Object with `Object Forwarding Preference` = `Datagram` on a
+stream if it is possible to send it as a datagram.
+
+~~~
+OBJECT_DATAGRAM_PREFERRED Message {
```
kixcord, line 1043: I think this is a footgun. It has the same downsides as IP fragmentation.

The application want specific behavior from datagrams, namely that they are independent. If a relay converts a datagram a stream, it will be split across IP packets and may even share packets with other datagram fragments. This makes implementing something like FEC basically impossible.

In my opinion, we should require the datagram extension and that all OBJECT_DATAGRAM messages are smaller than 1.2kB.

kixcord: Also, is a relay expected to cache datagrams or does it require timely delivery? I feel like the intent of datagrams is for transient data, but currently it only disables retransmissions. That feels like the wrong behavior, especially with datagrams that span multiple MTU because it only serves to amplify loss.

kixcord (commented):
Sorry for the earlier outburst. Alan's proposal is definitely in the right direction and is starting to align with my vision. I think I just need an outlet to write that vision down because I'm getting overwhelmed by leaving piecewise opinions on Github.

I think we should:
- delete `OBJECT_STREAM`, unless somebody wants it.
- delete `STREAM_HEADER_PRIORITY`, moving priority to the stream header.
- add `OBJECT_DATAGRAM`, but we should discuss datagrams more in general.

kixcord:
Oh and we should be some text:

> A relay MUST NOT reorder or drop OBJECTs within a stream.

Otherwise we don't fix the core issue, which is that relays can arbitrarily change how OBJECTs are delivered. Stream mapping is useless if it's not enforced.

vasilvv:
3 is a nice refactor of approach in 1. I like it, expect for two things:
- I don't think we need to support track-per-stream, and I don't remember if we ever actually got consensus for supporting it (we might need to discuss this separately).
- Priority/send order should be a stream-global, since that's the level at which those are determined in QUIC.

The text is not entirely clear to me: is it still legal to have two open streams for the same group with STREAM_HEADER_GROUP (with understanding that they will get merged at the relay)?

I think we could make approach 4 that does to 2 the same things that 3 does to 1. Canonical objects would have two fields:
- 'Object Placement Preference' that says that object is either (1) to be placed on a new stream, or (2) to be placed on the same stream as object N.
- 'Object Placement Close Bit' indicates whether there will be objects after this one on the stream or not (i.e. whether there is a FIN or not)

There's one OBJECT_STREAM frame with Track Alias, Group Sequence, Priority / Send Order as headers. You can trivially figure out how to backfill those fields (you only need to remember the most recent object for each stream); the placement when retrieving from cache is similarly easy (you do need a map for "most recent object -> stream").

I think this fits in with the "implicit-explicit" approach that Luke has mentioned before. It also has a nice property that there are no cases in which what's on the wire can mismatch with what's in the canonical object (unless streams get randomly reset).

wilaw:
> I don't think we need to support track-per-stream, and I don't remember if we ever actually got consensus for supporting it (we might need to discuss this separately).

Track-per-stream would be the preferred delivery mode for VOD files. I'd very much like to ensure that we keep support for that use case open, even if most current implementations are focused on live.

kixelated:
> The text is not entirely clear to me: is it still legal to have two open streams for the same group with STREAM_HEADER_GROUP (with understanding that they will get merged at the relay)?

I think it should be illegal as it breaks `implicit=explicit`. The broadcaster uses stream mapping A but wants downstream relays to use stream mapping B, assuming there even is a relay involved. I don't think there's a use-case for this special treatment of the first-hop. 

The only use-case that I've heard is a broadcast client that is smart enough to detect group/object boundaries (ie. frames or key frames) to annotate each OBJECT with the correct stream mapping, but doesn't want to open multiple streams? That's a very poor reason, especially if the first hop may be congested, and the broadcast client should just use the indicated stream mapping.

IMO a relay should not split or merge streams. It increases complexity and hurts performance. There's no need for your proposal 4 then.

kixelated:
> > I don't think we need to support track-per-stream, and I don't remember if we ever actually got consensus for supporting it (we might need to discuss this separately).
> 
> Track-per-stream would be the preferred delivery mode for VOD files. I'd very much like to ensure that we keep support for that use case open, even if most current implementations are focused on live.

Maybe.

The assumption is a track-per-stream would allow you to write a simple VOD-only client. It wouldn't need to reorder or reassemble objects because they're all delivered in order on a single stream. That's optimistic, as it will still need to support potentially unordered/unreliable objects (like a live client) due to seeking and gaps in the recording.

But I think we need to explore VOD and DVR more. I definitely don't want to require the server to modify the stream mapping at rest in order to serve VOD traffic, and it's not clear how that would even work for DVR.

At the very least we should first support VOD with the original stream mapping. Much like how HLS can use the same segments for live, DVR, and VOD.

wilaw:
> That's optimistic, as it will still need to support potentially unordered/unreliable objects (like a live client) due to seeking and gaps in the recording.

This assumes LIVE-TO-VOD workflow. When I speak of VOD, its for pure VOD, which has no gaps, and was never live. The track-per-stream would would mean the client would never have to deal with out-of-order or lost objects, as the single stream would deliver them reliably and in order. Like you, I really like the appeal of the simple client interface for this. All I need is more space in my cache and I have a new market segment to address.

vasilvv:
> IMO a relay should not split or merge streams. It increases complexity and hurts performance. There's no need for your proposal 4 then.

I'm confused. Proposal 4 does not allow merging or splitting streams.

> Track-per-stream would be the preferred delivery mode for VOD files. I'd very much like to ensure that we keep support for that use case open, even if most current implementations are focused on live.

That is true, but VOD requires different delivery mechanism from what live does (e.g. live cares about catching up to the live state, but for VOD, you usually just want to show everything in order).

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+* Object Sequence: The order of the object within the group.  The
+sequence starts at 0, increasing sequentially for each object within the
+group.
+
+* Object Send Order: An integer indicating the object send order
+{{send-order}} or priority {{ordering-by-priorities}} value.
+
+* Object Forwarding Preference: An enumeration indicating how a sender sends an
+object. The preferences are Track, Group, Object, Priority and Datagram.  An
+Object MUST be sent according to its `Object Forwarding Preference`, described
```
jordicenzano, line 997: If the "Object *MUST* be sent according to its `Object Forwarding Preference`", should we remove the word "Preference". It seems contradictory with MUST

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+{{model-group}} within the track.
+
+* Object Sequence: The order of the object within the group.  The
+sequence starts at 0, increasing sequentially for each object within the
+group.
+
+* Object Send Order: An integer indicating the object send order
+{{send-order}} or priority {{ordering-by-priorities}} value.
+
+* Object Forwarding Preference: An enumeration indicating how a sender sends an
```
jordicenzano, line 995: We need a way to tell the publisher how to map QUIC streams and this seems like a good proposal to me. Perhaps we should add something about relay implementation, perhaps: The relay would need remember how those objects were received on the subscriber side and replicate that in the publisher side

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+If a stream ends in the middle of a serialized Object, terminate the session
+with a Protocol Violation.
+
+TODO: figure out how a relay closes these streams
+
+When a stream begins with `STREAM_HEADER_TRACK`, all objects on the stream
+belong to track indicated by `Track Alias`.
+
+~~~
+STREAM_HEADER_TRACK Message {
```
jordicenzano, line 1066: I assume we can send 1 track (ex: video track) in mutiple Streams with STREAM_HEADER_TRACK, can I or that would be a protocol violation?

jordicenzano (commented):
Overall I like the proposal the explicitly map objects to quic streams. I left few comments

draft-ietf-moq-transport.md
afrind: I tried to capture this in the sections below describing how to determine the preference when receiving an object and how to send an object with a given preference.  I'm also going to add a normative bit to the relay section.

draft-ietf-moq-transport.md
afrind: Do we like "Object forwarding mode" better than preference?  I'm neutral.

draft-ietf-moq-transport.md
afrind: Jordi is right -- it's the same.

draft-ietf-moq-transport.md
afrind: I'm going to remove datagram for now.

draft-ietf-moq-transport.md
afrind: same as above.

draft-ietf-moq-transport.md
```diff
@@ -893,6 +897,312 @@ field is absent, the object payload continues to the end of the stream.
...
+stream, send the serialized data and terminate the stream.  An implementation
+MUST NOT send an Object with `Object Forwarding Preference` = `Datagram` on a
+stream if it is possible to send it as a datagram.
+
+~~~
+OBJECT_DATAGRAM_PREFERRED Message {
+  Track Alias (i),
+  Group Sequence (i),
+  Object Sequence (i),
+  Object Send Order (i),
```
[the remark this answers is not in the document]
afrind: This field is to convey the priority signal to the next hop, which might be using an API that allows setting priority on datagrams.

draft-ietf-moq-transport.md
afrind: You can terminate it cleanly on an object boundary, or reset it, but you can't fin in the middle of an object.

draft-ietf-moq-transport.md
afrind: I'm going to add that a sender SHOULD only have one stream open with the exact same STREAM_HEADER.  It's not all that enforceable at the receiver though because of reordering.

afrind (commented):
Individual Review

draft-ietf-moq-transport.md
wilaw: Binding the transport mode to the subscription seems like a reasonable constraint. Once a publisher is producing a track, I can't find an obvious use-case as to why they would need to alter the object-to-transport mapping. If there was one in the future, the publisher could work around it by terminating the track and advertising a new one with the new mapping. Binding the mapping to the subscription means, as kixelated points out, that the mapping info can be sent just once in SUSBCRIBE_OK. For compactness, anything that doesn't change over the life of a subscription (such as track alias, subscribe ID and now mapping) should be communicated via SUBSCRIBE_OK.

draft-ietf-moq-transport.md
```diff
@@ -893,6 +900,260 @@ field is absent, the object payload continues to the end of the stream.
...
+{: #object-track-format title="MOQT Track Stream Object Fields"}
+
+When a stream begins with `STREAM_HEADER_GROUP`, all objects on the stream
+belong to the track requested in the Subscribe message identified by `Subscribe
+ID` and the group indicated by `Group Sequence`.  All objects on the stream
+have the same `Object Send Order`.
+
+~~~
+STREAM_HEADER_GROUP Message {
+  Subscribe ID (i),
```
fluffy, line 1078: I think this should be track alias not subscribe ID

draft-ietf-moq-transport.md
```diff
@@ -882,12 +872,180 @@ group.
...
+}
+~~~
+{: #stream-header-group-format title="MOQT STREAM_HEADER_GROUP Message"}
+
+All Objects received on a stream opened with `STREAM_HEADER_GROUP` have an
+`Object Forwarding Preference` = `Group`.
+
+To send an Object with `Object Forwarding Preference` = `Group`, find the open
+stream that is associated with the subscription, `Group ID` and `Object
+Send Order`, or open a new one and send the `STREAM_HEADER_GROUP` if needed,
```
suhasHere, line 980: I really don't see a need to have a match on the SendOrder here. StreamPerGroup sends all the objects with the same groupId. if the application wants to send each priority in its own group, then it arranges that way.

draft-ietf-moq-transport.md
```diff
@@ -882,12 +872,180 @@ group.
...
+}
+{
+  Group ID = 0
+  Object ID = 0
+  Object Payload Length = 4
+  Payload = "abcd"
+}
+{
+  Group Sequence = 1
+  Object Sequence = 0
```
suhasHere, line 1010: sequence to Id

draft-ietf-moq-transport.md
```diff
@@ -882,12 +872,180 @@ group.
...
+  Object Payload Length = 4
+  Payload = "efgh"
+}
+~~~
```
suhasHere, line 1014: May be we should have this example to extended as 
{Stream_Header_Track}{g1, o0}{g1, o1} ... {g2,o0}, {g2,o1}, ....

draft-ietf-moq-transport.md
```diff
@@ -882,12 +872,180 @@ group.
...
+Sending a group on one stream, with a unordered object in the group appearing
+on its own stream.
+
+~~~
+Stream = 2
+
+STREAM_HEADER_GROUP {
+  Subscribe ID = 2
+  Track Alias = 2
+  Group Sequence = 0
```
suhasHere, line 1025: sequence to Id

draft-ietf-moq-transport.md
```diff
@@ -882,12 +872,180 @@ group.
...
+  Group Sequence = 1
+  Object Sequence = 0
+  Object Payload Length = 4
+  Payload = "efgh"
+}
+~~~
+
+Sending a group on one stream, with a unordered object in the group appearing
+on its own stream.
+
```
suhasHere, line 1018: the unordered object example is confusing here and I prefer not to include and create a new example for OBJECT_STREAM

suhasHere (commented):
Thanks @afrind  .. this is coming pretty close. I left few comments.

draft-ietf-moq-transport.md
```diff
@@ -592,6 +592,9 @@ terminates the track with a SUBSCRIBE_FIN
 (see {{message-subscribe-fin}}) or a SUBSCRIBE_RST
 (see {{message-subscribe-rst}}).
 
+A relay MUST not reorder or drop objects received on a multi-object stream when
+forwarding to subscribers.
```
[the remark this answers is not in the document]
afrind: I think we said in a previous author's call that we would relax this when we formally add concepts like TTL or drop preference, which give relay instructions about which objects are droppable.  I can update to "A relay MUST not reorder or drop objects *without application specific information*" - or something like that?

draft-ietf-moq-transport.md
```diff
@@ -882,12 +872,180 @@ group.
...
+  Object Payload Length (i),
+  Object Payload (...),
+}
+~~~
+{: #object-track-format title="MOQT Track Stream Object Fields"}
+
+When a stream begins with `STREAM_HEADER_GROUP`, all objects on the stream
+belong to the track requested in the Subscribe message identified by `Subscribe
+ID` and the group indicated by `Group ID`.  All objects on the stream
+have the same `Object Send Order`.
```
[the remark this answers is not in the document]
afrind: The intent is to tell the receiver how to populate the canonical 'Object Send Order'.  I can say "have the `Object Send Order` specified in the stream header" ?

draft-ietf-moq-transport.md
afrind: The intent was to allow more flexibility than strictly stream-per-group.  With this design, the publisher can split a group into multiple streams with different priorities.  The original proposal 3 had both kinds of streams and the feedback was to combine them.

afrind (commented):
Thanks for the quick review. I have addressed some, left some responses/suggestions for the others.

For examples, let's wait til all the feedback is in and we can make another pass.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
suhasHere: works for me. Thanks

draft-ietf-moq-transport.md
```diff
@@ -882,12 +874,183 @@ group.
...
+  Object Send Order (i),
+}
+~~~
+{: #stream-header-track-format title="MOQT STREAM_HEADER_TRACK Message"}
+
+All Objects received on a stream opened with STREAM_HEADER_TRACK have an `Object
+Forwarding Preference` = `Track`.
+
+To send an Object with `Object Forwarding Preference` = `Track`, find the open
+stream that is associated with the subscription, or open a new one and send the
```
kixelated, line 952: I'm confused by this sentence. This seems like an implementation detail depending on the API?

Personally, I wouldn't lazy initialize the stream because then the MoQ library won't know when to close it. I think you can safely remove this sentence and just say that multiple objects follow.

draft-ietf-moq-transport.md
```diff
@@ -882,12 +874,183 @@ group.
...
+`STREAM_HEADER_TRACK` if needed, then serialize the the following object fields.
+
+~~~
+{
+  Group ID (i),
+  Object ID (i),
```
kixelated, line 958: Can we add the constraint that both of these must be increasing over the life of the stream? Otherwise dropping and subsequent gap detection becomes very difficult.

draft-ietf-moq-transport.md
```diff
@@ -882,12 +874,183 @@ group.
...
+`Object Forwarding Preference` = `Group`.
+
+To send an Object with `Object Forwarding Preference` = `Group`, find the open
+stream that is associated with the subscription, `Group ID` and `Object
+Send Order`, or open a new one and send the `STREAM_HEADER_GROUP` if needed,
+then serialize the following fields.
+
+~~~
+{
+  Object ID (i),
```
kixelated, line 990: Same thing, can we make this increasing only?

kixelated (changes requested):
It's almost there. This is exponentially easier for a relay and receiver than prior proposals.

draft-ietf-moq-transport.md
suhasHere: i think this needs to caoture the intent as " all the objects for a given track is sent on the same stream" .

draft-ietf-moq-transport.md
suhasHere: groupId and object Id are monotonically increasing integers.

draft-ietf-moq-transport.md
```diff
@@ -882,12 +874,183 @@ group.
...
+When multiple objects are sent on a stream, the stream begins with a stream
+header message and is followed by one or more sets of serialized object fields.
+If a stream ends gracefully in the middle of a serialized Object, terminate the
+session with a Protocol Violation.
+
+A sender SHOULD NOT open more than one multi-object stream at a time with the
+same stream header message type and fields.
+
+
+TODO: figure out how a relay closes these streams
```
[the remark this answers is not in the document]
suhasHere: agree we need to get to this , but I feel this can be done in a separate PR>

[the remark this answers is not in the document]
afrind: > I don't think a relay can implement the draft until this is specified.

I agree.

> agree we need to get to this , but I feel this can be done in a separate PR

I'm inclined to take the consensus we have, merge it and build on it, rather than keeping this PR open.

I don't think we want to take it from the QUIC layer and we do need an MoQ layer signal.  Is the stream carrying the track data even sufficient or could it be reset?  The stream termination messages may need to go on the control stream?

draft-ietf-moq-transport.md
afrind: The intent was to explain how relays send outgoing objects, but make it generic for publishers too.  Similar language was used for other multi-object stream formats.

I don't think it conflicts with your plan to not-lazy-initialize, that would fit in "find the open stream...", but I'm open to different language too.

draft-ietf-moq-transport.md
```diff
@@ -882,12 +874,183 @@ group.
...
+  Group ID = 0
+  Object Send Order = 0
+}
+{
+  Object ID = 0
+  Object Payload Length = 4
+  Payload = "abcd"
+}
+{
+  Object ID = 1
```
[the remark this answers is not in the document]
afrind: Yes, I think I meant for this to be Object ID=2.

draft-ietf-moq-transport.md
afrind: I'm ok adding an explicit restriction here.  Does anyone object?

draft-ietf-moq-transport.md
afrind: I'm ok adding an explicit restriction here.  Does anyone object?

draft-ietf-moq-transport.md
afrind: > A relay could still convert single object streams into multi object streams that by buffering until EOF.

I don't think so.  The rules for sending objects say

```
To send an Object with `Object Forwarding Preference` = `Object`, open a stream,
serialize object fields below, and terminate the stream.
```

afrind (commented):
Individual comments

ianswett (approved):

draft-ietf-moq-transport.md
kixelated: > groupId and object Id are monotonically increasing integers.

Are they monotonically increasing? I thought the use-case was two streams with objects interlaced, in which case they wouldn't be.

kixelated (approved):

draft-ietf-moq-transport.md
```diff
@@ -895,12 +888,190 @@ group.
...
+}
+~~~
+{: #moq-transport-object-stream-format title="MOQT OBJECT_STREAM Message"}
+
+* Subscribe ID: Subscription Identifer as defined in {{message-subscribe-req}}.
+
+* Track Alias: Identifies the Track Namespace and Track Name as defined in
+{{message-subscribe-req}}.
+
+The Track Namespace and Track Name identified by the Track Alias is different
```
ianswett, line 928:
```suggestion
If the Track Namespace and Track Name identified by the Track Alias is different
```

ianswett: Also, this applies to all types of mappings, so maybe move it up to the end of the previous section?

ianswett (approved):

Issue #318, opened 2023-11-01
Impossible to know group FIN


kixelated:
The current approach of allowing any OBJECTs to arrive on any stream makes it impossible for the receiver to know when a group ends.

There's a few conditions when you need to know the end of a group:

1. A relay being told to terminate the subscription after group N (subscription hints).
2. A relay knowing when to close a QUIC stream (if it writes a group per stream).
3. A decoder deciding when to start decoding/rendering the next group (ex. GoP).

Likewise there's a similar problem of knowing when there's an gap within a group. There's no way to signal that OBJECT X will never arrive, so the receiver should not wait for it.

kixelated:
I strongly advocate a 1:1 mapping for group to stream. Objects within a group are in order, but there may be gaps. Gaps can be detected via sequence numbers and the end is detected via stream FIN.

kixelated:
Alternatively, we could have `OBJECT_DROP`, `GROUP_DROP` and `GROUP_FIN` control messages. I think this is a mess, and the receiver would need a corresponding `STOP_GROUP` in response to an unrecoverable gap caused by `OBJECT_DROP`. 

I don't think we can annotate OBJECTs with a FIN flag because they're lossy. You need some reliable mechanism to signal the end.

kixelated:
Maybe this a duplicate of #315 ?

afrind:
The issue mentioning Last Object Previous Group is in #224

fluffy:
I think the best solutions to this is a header on the last object in the group that indicates it is the last object in the group.

afrind:
Individual Comment:

> I think the best solutions to this is a header on the last object in the group that indicates it is the last object in the group.

I share Luke's concern above about OBJECT being somewhat unreliable.  I think a GROUP_FIN message with stronger reliability guarantees would look similar on the wire but work better.

eg:

Instead of 

```
OBJECT(G1,O1)  OBJECT(G1,O2)  OBJECT(G1,O3,last)
```

Have


```
OBJECT(G1,O1)  OBJECT(G1,O2)  OBJECT(G1,O3), GROUP_FIN(G1, O3)
```

This works better than implicit mapping from groups to streams.  A relay might need to close a connection with an open stream, but wouldn't want to reset it and wouldn't want the stream FIN to be interpreted as a group FIN.  I don't want to put GROUP_FIN on the control stream, but that would be the safest.

I think the other messages Luke mentioned related to DROP are probably an orthogonal issue.

[...]

PR #334, opened 2023-11-08
Add support for end message.

suhasHere:
This PR attempts to support the requirements needed for #318 and #244

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -701,6 +701,8 @@ MOQT Message {
 |-------|----------------------------------------------------|
 | 0x10  | GOAWAY ({{message-goaway}})                        |
 |-------|----------------------------------------------------|
+| 0x11  | END ({{message-end}})                              |
+|-------|----------------------------------------------------|
 | 0x40  | CLIENT_SETUP ({{message-setup}})                   |
 |-------|----------------------------------------------------|
 | 0x41  | SERVER_SETUP ({{message-setup}})                   |
@@ -1268,11 +1270,43 @@ GOAWAY Message {
 ~~~
 {: #moq-transport-goaway-format title="MOQT GOAWAY Message"}
 
-* New Session URI: The client MUST use this URI for the new session if provded.
+* New Session URI: The client MUST use this URI for the new session if provided.
   If the URI is zero bytes long, the current URI is reused instead. The new
   session URI SHOULD use the same scheme as the current URL to ensure
   compatibility.
 
+
+## END {#message-end}
+
+A publisher sends a `END` message to indicate the termination condition when
+either a track, group or object ends.
+
+The format of `END` message is as shown below
+
+~~~
+END Message {
+  Type(i),
+  Track Alias (i)
+  Group Sequence (i),
+  Object Sequence (i)
+}
+~~~
+
+There are 3 possible values for the `Type` field as defined below.
+
+* Track (0x0) : The `END` message corresponds to termination of a track.
+
+* Group (0x1) : The `END` message corresponds to termination of a group.
+
+* Object (0x2) : The `END` message corresponds to termination of an object.
+
+In all the cases, the fields `Track Alias`, `Group Sequence` and
+`Object Sequence` identify the track, group and object respectively for
+which the termination condition applies.
+
+Also to note, when using object messages that is length aware, the receiver
+can compute the object competion by counting the length number of
+bytes received.
 
 # Security Considerations {#security}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1278,30 +1278,37 @@ GOAWAY Message {
 
 ## END {#message-end}
 
-A publisher sends `END` message to indicate the completion of either the track
-or group or an object. The `END` message MUST NOT be sent on the control stream.
+A publisher can optionally send `END` message to indicate the completion of
+either the track or group or an object. The `END` message MUST NOT be
+sent on the control stream.
 
 The format of `END` message is as shown below
 
 ~~~
 END Message {
   Type (i),
-  Track Alias (i)
+  Track Alias (i),
   Group (i),
-  Object (i)
+  [Object (i)]
 }
 ~~~
 
+* Track Alias: Identifies the track corresponding to the `END` message.
+
 There are 3 possible values for the `Type` field as defined below.
 
-* Track (0x0) : The `END` message corresponds to completion of a track.
+* Track (0x0) : The `END` message corresponds to completion of a track. The
+`Group` and `Object` field identify the final group and object sequences
+for the track.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1268,12 +1270,52 @@ GOAWAY Message {
...
   If the URI is zero bytes long, the current URI is reused instead. The new
   session URI SHOULD use the same scheme as the current URL to ensure
   compatibility.
 
 
+## END {#message-end}
+
+A publisher can optionally send `END` message to indicate the completion of
+either the track or group or an object. The `END` message MUST NOT be
+sent on the control stream.
```
afrind, line 1283: I'm somewhat concerned about including END on OBJECT streams.  In the case where the peer sends a STOP_SENDING, the original sender may not know if the receiver has received the END message.  Instead, I'm inclined to say that END must appear on a non-resettable stream.  This could be a track-specific control stream, or the sender can just open a stream, send END and close it.

draft-ietf-moq-transport.md
```diff
@@ -1268,12 +1270,52 @@ GOAWAY Message {
...
+  Group (i),
+  [Object (i)]
+}
+~~~
+
+* Track Alias: Identifies the track corresponding to the `END` message.
+
+There are 3 possible values for the `Type` field as defined below.
+
+* Track (0x0) : The `END` message corresponds to completion of a track. The
```
afrind, line 1300: How is this different from SUBSCRIBE_FIN?

draft-ietf-moq-transport.md
```diff
@@ -1268,12 +1270,52 @@ GOAWAY Message {
...
+`Group` and `Object` field identify the final group and object sequences
+for the track.
+
+* Group (0x1) : The `END` message corresponds to completion of a group
+identified in the `Group` field. The `Object` field identifies the last
+object of the completed group.
+
+* Object (0x2) : The `END` message corresponds to completion of an object,
```
afrind, line 1308: I'm unsure when I would send this.  If I'm using OBJECT_WITH_LENGTH, the end is well known - it's over when you receive the terminal byte of payload.  If I'm using OBJECT_WITHOUT_LENGTH, there isn't a place to put END in the same stream, and presumably the end of the stream marks the end.

Do you have another use-case in mind?

afrind (commented):
Individual Review:

If SUBSCRIBE_FIN works for END(Track), then the main missing use-case is group.  Maybe we can morph this to END_GROUP { Track Alias, Group, Object } for now?

draft-ietf-moq-transport.md
```diff
@@ -1268,12 +1270,52 @@ GOAWAY Message {
...
+
+* Track Alias: Identifies the track corresponding to the `END` message.
+
+There are 3 possible values for the `Type` field as defined below.
+
+* Track (0x0) : The `END` message corresponds to completion of a track. The
+`Group` and `Object` field identify the final group and object sequences
+for the track.
+
+* Group (0x1) : The `END` message corresponds to completion of a group
```
kixelated, line 1304: This use-case is useful without stream mapping. I still want a stream per group, where the stream FIN signals the end of a group, instead of this (noisy) message on the control stream.

draft-ietf-moq-transport.md
kixelated: +1 There's no use-case here.

draft-ietf-moq-transport.md
```diff
@@ -1268,12 +1270,52 @@ GOAWAY Message {
...
+## END {#message-end}
+
+A publisher can optionally send `END` message to indicate the completion of
+either the track or group or an object. The `END` message MUST NOT be
+sent on the control stream.
+
+The format of `END` message is as shown below
+
+~~~
+END Message {
```
kixelated, line 1288: You also need to to include the last group/object like `SUBSCRIBE_RST/FIN`. Sending a boolean `END` message without them is extremely racey.

draft-ietf-moq-transport.md
```diff
@@ -1268,12 +1270,52 @@ GOAWAY Message {
...
+* Group (0x1) : The `END` message corresponds to completion of a group
+identified in the `Group` field. The `Object` field identifies the last
+object of the completed group.
+
+* Object (0x2) : The `END` message corresponds to completion of an object,
+identified in the `Object` field whose group is indicated in the `Group`
+field.
+
+
+Sending `END` message is optional for cases where the completion
```
kixelated, line 1313: We should remove the object mode because like this paragraph outlines, it provides duplicate functionality.

However the `END` message MUST/SHOULD be sent for the end of each track/group, otherwise a receiver can't rely on it.

kixelated (changes requested):
As the protocol currently stands, we need a way to mark the end of a group, but we also need a way to mark when individual objects have been dropped. Otherwise, a subscriber waiting for an END object as signaled by this message may never unblock. I don't think this end group is useful otherwise, and perhaps misleading.

draft-ietf-moq-transport.md
fluffy: This is in the data plane and does not change the subscription. That allows it to indicate the end of each group as well as the end of track. But even when using it for end of track, imagine a care where things are out of order. Lets say group 100 is the last one, group 99 may still arriver after the END object. The END object just tells you what the publisher is saying is the last object in the group or track. We could do this with flags in the object but that has other problems.

draft-ietf-moq-transport.md
```diff
@@ -1276,12 +1278,52 @@ GOAWAY Message {
...
-* New Session URI: The client MUST use this URI for the new session if provded.
+* New Session URI: The client MUST use this URI for the new session if provided.
   If the URI is zero bytes long, the current URI is reused instead. The new
   session URI SHOULD use the same scheme as the current URL to ensure
   compatibility.
 
 
+## END {#message-end}
+
+A publisher can optionally send `END` message to indicate the completion of
```
fluffy, line 1289: This needs more information to explain how this is differnt than what is happening with SUBSCRIBES and that relays need to cache this information and deliver it. I feel like it might be better done as as a special type of object.

draft-ietf-moq-transport.md
fluffy: I agree with Luke on this should not be optional

fluffy (changes requested):

draft-ietf-moq-transport.md
wilaw: I think we should signal the end of publication of a track, but it should not apply at the GROUP level otherwise we have a very chatty. 

Furthermore, END can be scoped at multiple different layers. It think this should apply specifically to tracks and that it should look like this:

TRACK_END
{
   Track Namespace (b),
   Track Name (b),
   LastGroup (Location),
   LastObject (Location),
}

fluffy:
As a wY of making progress on this, lets first focus on how we know about the end of group, then jump into the end of track.

PR #338, opened 2023-11-13
Fail session on malformed parameter

### LINKED ISSUE #314: VarInt parameters length mismatch
Currently, VarInts have two lengths:

1. In the parameter length.
2. In the first two bits of the parameter value.

The decision in #294 was that the two MUST match, which is a good solution without a full #256 reform.

However:

> If the parameter value is a varint, but the self-encoded length of that varint does not match the Parameter Length field, the receiver MUST ignore the parameter using the value in the Parameter Length field.

Ignoring seems like the wrong play. For example, we know the role parameter (0x0) is a VarInt in the current draft. However this text says I MUST ignore any errors when parsing this varint and pretend like the parameter was never sent, which is both unenforceable and a major foot-gun (silently ignoring parameters).

I think we close the connection with an error if there's an error parsing a known parameter.

martinduke:
Fixes #314

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -749,9 +749,9 @@ The Parameter Length field of the String Parameter encodes the length
 of the Parameter Value field in bytes.
 
 Each parameter description will indicate the data type in the Parameter Value
-field. If the parameter value is a varint, but the self-encoded length of that
-varint does not match the Parameter Length field, the receiver MUST ignore the
-parameter using the value in the Parameter Length field.
+field. If a receiver understands a parameter type, and  the parameter value is a
+varint, but the self-encoded length of that varint does not match the Parameter
+Length field, the receiver MUST terminate the session.
 
 ### Version Specific Parameters {#version-specific-params}
 


draft-ietf-moq-transport.md
```diff
@@ -749,9 +749,9 @@ The Parameter Length field of the String Parameter encodes the length
 of the Parameter Value field in bytes.
 
 Each parameter description will indicate the data type in the Parameter Value
-field. If the parameter value is a varint, but the self-encoded length of that
-varint does not match the Parameter Length field, the receiver MUST ignore the
-parameter using the value in the Parameter Length field.
+field. If a receiver understands a parameter type, and  the parameter value is a
+varint, but the self-encoded length of that varint does not match the Parameter
```
kixelated, line 753: Can we change this so it's not VarInt specific? If the type doesn't match the length (buffer under-run or over-run), then it's a error.

draft-ietf-moq-transport.md
```diff
@@ -749,9 +749,9 @@ The Parameter Length field of the String Parameter encodes the length
...
+Length field, the receiver MUST terminate the session.
```
kixelated, line 754: With what error code?

draft-ietf-moq-transport.md
martinduke: done

draft-ietf-moq-transport.md
martinduke: I made it error code TBD because there aren't any error codes

ianswett (approved):
LG, though we can bikeshed about the error code name.

PR #339, opened 2023-12-11
Change Sequence to ID for Objects and Groups

ianswett:
Sequence typically indicates a number of things.


suhasHere (approved):
This is a good change.

Issue #337, opened 2023-11-10
Extend SUBSCRIBE_OK message to include information about the live edge of the track


wilaw:
Consider an edge relay, receiving and forwarding relative SUBSCRIBE requests for -5, -3, and -1 for the same track. The first object it receives back is group 37. It cannot tell where to place this object in the sequence and must therefore hold it until it knows more information. If the next object is Group 39, it still cannot forward it, because the head sate is undefined.

We can solve this problem simply by having the SUBSCRIBE_OK response indicate the latest group and object number:

SUBSCRIBE_OK
{
  Track Namespace (b),
  Track Name (b),
  Track Alias (i),
  Expires (i),
  LastestGroup(i),
  LastestObject(i)
}

One consequence of adding these fields is that an edge, which has no active subscription for the track,  cannot send the SUBSCRIBE_OK response until it itself receives a SUBSCRIBE_OK after forwarding the SUBSCRIBE. This makes subscription an async process in which the acknowledgment of the subscription may take in some cases a RTT back to the origin. I think this is necessary for stable state management.

kixelated:
+1 this is required to build a relay and a non-realtime player.

> One consequence of adding these fields is that an edge, which has no active subscription for the track, cannot send the SUBSCRIBE_OK response until it itself receives a SUBSCRIBE_OK after forwarding the SUBSCRIBE. This makes subscription an async process in which the acknowledgment of the subscription may take in some cases a RTT back to the origin.

This was required anyway, otherwise the relay would swallow any application SUBSCRIBE_ERROR codes. My mental model is that SUBSCRIBE_OK == HTTP response headers.

suhasHere:
> One consequence of adding these fields is that an edge, which has no active subscription for the track, cannot send the SUBSCRIBE_OK response until it itself receives a SUBSCRIBE_OK after forwarding the SUBSCRIBE

Moqt is a pub/sub protocol and is hop-by-hop protocol. A subscribe ok means the susbcribe request was validated at the relay. A OK isn't tied to getting a OK from the original publisher. Its sub/sub-ok is a async operation ( like any typical pub/sub systems out there). An ok for subscribe means a promise that if there is a matching data ( either in cache or coming in from a live publisher at some point), it will be forwarded .

suhasHere:
[...]

PR #343, opened 2024-01-22
Add largest Group ID and Object ID to SUBSCRIBE_OK

wilaw:
Fixes https://github.com/moq-wg/moq-transport/issues/337

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1293,11 +1293,23 @@ SUBSCRIBE_OK
 longer valid. A value of 0 indicates that the subscription stays active
 until it is explicitly unsubscribed.
 
-* ContentExists: 1 if an object has been published on this track, 0 if not. If 0, then the Largest Group ID and Largest Object ID fields will not be present. 
+* ContentExists: 1 if an object has been published on this track, 0 if not.
+* If 0, then the Largest Group ID and Largest Object ID fields will not be
+* present. 
 
-* Largest Group ID: the largest Group ID available for this track. This Group ID corresponds to the Group that would be returned with a {{subscribe-locations}} RelativePrevious value of 0. It represents the latest Group boundary available for this track. Note that the Group may not be complete when this value is returned. This field is only present if ContentExists has a value of 1.
+* Largest Group ID: the largest Group ID available for this track. This
+* Group ID corresponds to the Group that would be returned with a
+* {{subscribe-locations}} RelativePrevious value of 0. It represents the latest
+* Group boundary available for this track. Note that the Group may not be
+* complete when this value is returned. This field is only present if
+* ContentExists has a value of 1.
 
-* Largest Object ID: the largest Object ID available within the largest Group ID for this track. This Object ID corresponds to the Object that would be returned with a {{subscribe-locations}} RelativePrevious value of 0. It represents the latest Object boundary available for this track. Note that the Object may not be complete when this value is returned. This field is only present if ContentExists has a value of 1.
+* Largest Object ID: the largest Object ID available within the largest Group ID
+* for this track. This Object ID corresponds to the Object that would be
+* returned with a {{subscribe-locations}} RelativePrevious value of 0. It
+* represents the latest Object boundary available for this track. Note that the
+* Object may not be complete when this value is returned. This field is only
+* present if ContentExists has a value of 1.
 
 
 ## SUBSCRIBE_ERROR {#message-subscribe-error}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1299,17 +1299,13 @@ until it is explicitly unsubscribed.
 
 * Largest Group ID: the largest Group ID available for this track. This
 * Group ID corresponds to the Group that would be returned with a
-* {{subscribe-locations}} RelativePrevious value of 0. It represents the latest
-* Group boundary available for this track. Note that the Group may not be
-* complete when this value is returned. This field is only present if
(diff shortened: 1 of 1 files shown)


suhasHere (changes requested):

draft-ietf-moq-transport.md
```diff
@@ -1256,6 +1258,10 @@ SUBSCRIBE_OK
 longer valid. A value of 0 indicates that the subscription stays active
 until it is explicitly unsubscribed.
 
+* Largest Group ID: the largest Group ID available for this track. This Group ID corresponds to the Group that would be returned with a {{subscribe-locations}} RelativePrevious value of 0. It represents the latest Group boundary available for this track. Note that the Group may not be complete when this value is returned. 
+
+* Largest Object ID: the largest Object ID available within the largest Group ID for this track. This Object ID corresponds to the Object that would be returned with a {{subscribe-locations}} RelativePrevious value of 0. It represents the latest Object boundary available for this track. Note that the Object may not be complete when this value is returned.
+
```
suhasHere, line 1264: same comments as above applies here i think

ianswett (commented):
I was thinking we would add the start group and Object ID for the requested subscription.  Even for an absolute subscription, I can imagine cases the requested start values can't be served, so newer ones have to be served instead?

It'd also be nice to make these optional, but I think that's a detail that depends upon what fields we end up wanting here.

fluffy (changes requested):
This seems like it will have many race conditions. It seems like it would be better to transfer this in the STREAM_HEADER_TRACK and equivalent.  Thoughts on that ?

I also worry about how long a subscribe would wait to find out this information - that seems like it could be problematic for the client knowing when to give up on a subscribe. 

Hopefully we can discuss in Denver

draft-ietf-moq-transport.md
```diff
@@ -1256,6 +1258,10 @@ SUBSCRIBE_OK
 longer valid. A value of 0 indicates that the subscription stays active
 until it is explicitly unsubscribed.
 
+* Largest Group ID: the largest Group ID available for this track. This Group ID corresponds to the Group that would be returned with a {{subscribe-locations}} RelativePrevious value of 0. It represents the latest Group boundary available for this track. Note that the Group may not be complete when this value is returned. 
+
```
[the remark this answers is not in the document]
wilaw: This has now been resolved with the addition of the ContentExists flag.

draft-ietf-moq-transport.md
wilaw: Same resolution.

afrind:
nit: can you wrap the lines at 80 chars?

draft-ietf-moq-transport.md
```diff
@@ -1290,6 +1293,12 @@ SUBSCRIBE_OK
 longer valid. A value of 0 indicates that the subscription stays active
 until it is explicitly unsubscribed.
 
+* ContentExists: 1 if an object has been published on this track, 0 if not. If 0, then the Largest Group ID and Largest Object ID fields will not be present. 
+
+* Largest Group ID: the largest Group ID available for this track. This Group ID corresponds to the Group that would be returned with a {{subscribe-locations}} RelativePrevious value of 0. It represents the latest Group boundary available for this track. Note that the Group may not be complete when this value is returned. This field is only present if ContentExists has a value of 1.
```
suhasHere, line 1298: I don't see a need to include text on locations here.

suhasHere: Also , this seems uneccesary "Note that the Group may not be complete when this value is returned.".

draft-ietf-moq-transport.md
suhasHere: same comments as above, Let's just indicate largest object Id os the largest known object under the group indicated by "LargestGroupId".

suhasHere (changes requested):
Small nits. Thanks for the change.s

draft-ietf-moq-transport.md
wilaw: > I don't see a need to include text on locations here.

I think we need to cleanly define what "largest" means in the context of groups and how this maps precisely to the Location description. Simply saying its the "largest group" doesn't tie in to how an subscriber would actually request that object. 

> Also , this seems unnecessary "Note that the Group may not be complete when this value is returned.".

This was added because there was some ambiguity as to whether this was the largest fully complete group available for the track, or just the largest group for which at least one object has started publishing. It's the latter, but thats not obvious if the language above is omitted.

suhasHere: Largest is the largest know groupId and objectId for a given track seen so far or available. I think we are making it less clear by adding location information. Locations are filters and the answer one gets doesn't have to match it since the state of publisher may have very well moved out. 

I sitll feel if the group being full or not doesn't need to be specified here.

draft-ietf-moq-transport.md
afrind: Individual Comment: 

I recommend we keep the text here very simple, and reference the Locations section for a definition of what Largest Group and Largest Object mean, since they are used throughout that section.  Right now, isn't a good definition of Largest Group and Largest Object there, but we can add that in a follow-up PR?

ianswett (approved):

draft-ietf-moq-transport.md
ianswett: I shortened the descriptions some, though we could probably make them even shorter.

ianswett:
In the call last Wednesday, we agreed this design was the correct approach, so merging even though there are some reviews outstanding requesting changes.

PR #342, opened 2024-01-11
Add Forwarding Preference=Datagram

afrind:
This was previously part of #333. It allows an object to be sent in a datagram f it fits, and put on a stream if it doesn't.

Closes #316

PR #344, opened 2024-01-22
Land Datagram Preferred (PR #342)

ianswett:
Update main to reflect the tx-mode branch after #342 was merged.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -593,7 +593,7 @@ terminates the track with a SUBSCRIBE_FIN
 (see {{message-subscribe-rst}}).
 
 A relay MUST not reorder or drop objects received on a multi-object stream when
-forwarding to subscribers.
+forwarding to subscribers, unless it has application specific information.
 
 Relays MAY aggregate authorized subscriptions for a given track when
 multiple subscribers request the same track. Subscription aggregation
@@ -932,7 +932,8 @@ TODO: figure out how a relay closes these streams
 
 When a stream begins with `STREAM_HEADER_TRACK`, all objects on the stream
 belong to the track requested in the Subscribe message identified by `Subscribe
-ID`.  All objects on the stream have the same `Object Send Order`.
+ID`.  All objects on the stream have the `Object Send Order` specified in the
+stream header.
 
 
 ~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -931,6 +935,36 @@ the receiver MUST close the session with a Protocol Violation.
...
+be the first message on a unidirectional stream.
+
+An Object received in an `OBJECT_PREFER_DATAGRAM` message has an `Object
+Forwarding Preference` = `Datagram`.
+
+To send an Object with `Object Forwarding Preference` = `Datagram`, determine
+the length of the fields and payload, and compare the length with the maximum
+datagram size of the session.  If the object size is less than or equal maximum
+datagram size, send the serialized data as a datagram.  Otherwise, open a
+stream, send the serialized data and terminate the stream.  An implementation
```
suhasHere, line 952: i think we need to preserve the fact that this is preferred as datagram when we put it in a stream . Wonder where should that go ? 

Should we extend OBJECT_STREAM message with a bool saying "preferred_datagram". This is needed for receviers handling and forwarding

suhasHere (commented):
Looks fine . Have one comment though

draft-ietf-moq-transport.md
suhasHere: this can be added in a separate PR too

suhasHere (approved):

draft-ietf-moq-transport.md
afrind: Preserving the preference was the intent.  When datagram is sent on a stream, it starts with OBJECT_PREFER_DATAGRAM (0x1) to communicate this.

draft-ietf-moq-transport.md
```diff
@@ -931,6 +935,36 @@ the receiver MUST close the session with a Protocol Violation.
 
 * Other fields: As described in {{canonical-object-fields}}.
 
+**Object Prefer Datagram Message**
+
+An `OBJECT_PREFER_DATAGRAM` message carries a single object in a datagram or
+a stream. There is no explicit length of the payload; it is determined by the
+length of the datagram or stream.  If this message appears on a stream, it MUST
+be the first message on a unidirectional stream.
```
afrind, line 943: s/first/only/  based on a comment in 342.

afrind (commented):
Individual Review: 

There was some feedback on previous iterations about this PR lacking guidance when encountering flow control limits (eg: send as datagram, relay converts to stream, but is out of streams or flow control)

draft-ietf-moq-transport.md
ianswett:
```suggestion
be the only message on a unidirectional stream.
```

martinduke (commented):
Whoa, I've missed some PRs, but LGTM

ianswett:
> Individual Review:
> 
> There was some feedback on previous iterations about this PR lacking guidance when encountering flow control limits (eg: send as datagram, relay converts to stream, but is out of streams or flow control)

Agreed, but I'm not sure there's any normative language there.  Is this a case of writing a paragraph cautioning that such limits could cause issues?

PR #354, opened 2024-02-04
clean up details on binary encoding syntax

fluffy:
This PR cleans up some nits to make all the places we use the RFC9000 encoding syntax valid, updates our extensions to it, and add a non normative summary of parts of the syntax we use. 

I don't think this is making any changes to anything on the wire or normative changes to spec other than the statement SHOULD use smallest number of bytes for variable length integers which RFC9000 only requires for some fields.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -206,14 +206,51 @@ Track:
 
 ## Notational Conventions
 
-This document uses the conventions detailed in ({{?RFC9000, Section 1.3}})
-when describing the binary encoding.
+This document uses the conventions detailed in ({{?RFC9000, Section
+1.3}}) when describing the binary encoding.
 
-This document also defines an additional field type for binary data:
+As a quick reference, the following list provides a non normative
+summary of the parts of RFC9000 syntax that are used in this
+specification.
+
+x (L):
+
+: Indicates that x is L bits long
+
+x (i):
+
+: Indicates that x holds an integer value using the variable-length
+  encoding as described in ({{?RFC9000, Section 16}})
+
+x (..):
+
+: Indicates that x can be any length including zero bits long.  Values
+ in this format always end on a byte boundary.
+
+[x (L)]:
+
+: Indicates that x is optional and has a length of L
+
+x (L) ...:
+
+: Indicates that x is repeated zero or more times and that each instance
+  has a length of L
+
+This document extends the RFC9000 syntax and with the additional field types:
 
 x (b):
-: Indicates that x consists of a variable length integer (i), followed by
-  that many bytes of binary data (..).
+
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -206,14 +206,51 @@ Track:
...
+  described in ({{?RFC9000, Section 16}}), followed by that many bytes
+  of binary data
+
+x (Location):
+
+: Indicates x is a Location structure defined in {{subscribe-locations}}.
+
+To reduce unnecessary use of bandwidth, variable length integers SHOULD
+be encoded using the least number of bytes possible to represent the
+required value.
```
fluffy, line 253: The above sentence is the only part of this PR that seems non editorial to me. If I should put it in a separate PR, glad to do that ...

martinduke (approved):
I am not in love with the ABNF idea, but this is non-controversial.

suhasHere (approved):

draft-ietf-moq-transport.md
kixelated: `RECOMMENDED` maybe, but I would just remove this sentence IMO.

I actually used a 2-byte varint for packet sizes in QUIC so I could avoid a copy once the final size is known. I imagine you could do something similar in MoQ, or use larger varints to make rewriting OBJECT headers easier.

Either way we can't say MUST and I'm not sure what we gain with a SHOULD. A decoder has to be prepared for anything.

afrind: Consider an RFC 6919 "OUGHT TO" https://datatracker.ietf.org/doc/html/rfc6919#section-4

fluffy: Agree that is a real change and happy to refactor it into a separate PR.

kixelated: > OK - updated this. I will note that from https://datatracker.ietf.org/doc/html/rfc2119#section-3 SHOULD just means "RECOMMENDED" not "REQUIRED"

Ah okay I missed that memo.

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -209,11 +209,47 @@ Track:
...
-: Indicates that x consists of a variable length integer (i), followed by
-  that many bytes of binary data (..).
+
+: Indicates that x consists of a variable length integer encoding as
+  described in ({{?RFC9000, Section 16}}), followed by that many bytes
+  of binary data
+
+x (Location):
+
+: Indicates x is a Location structure defined in {{subscribe-locations}}.
```
ianswett, line 248: I find this forward reference a bit odd, and it's only used for one frame, which is likely to change soon, so I'm going to remove it to avoid it becoming out of sync.

ianswett (approved):

Issue #359, opened 2024-02-07
Update PREFER_DATAGRAM to require DATAGRAM only delivery


suhasHere:
This came out of interim that falling back to stream for Datagram delivery is not an great idea

afrind:
This was fixed by #360

PR #360, opened 2024-02-07
Make Datagram mode as REQUIRED

suhasHere:
Addresses #359

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -942,9 +942,7 @@ There is no explicit length of the payload; it is determined by the
 length of the datagram.
 
 An Object received in an `OBJECT_DATAGRAM` message has an `Object
-Forwarding Preference` = `Datagram`. To send an Object with `Object Forwarding Preference` = `Datagram`, determine the length of the fields and payload and send the Object as datagram. In certain scenarios where the object size can be larger than maximum datagram size of the session, it is upto the implementation to
-either drop the object or apply any applicable transport level fragmentation,
-necessary for forwarding.
+Forwarding Preference` = `Datagram`. To send an Object with `Object Forwarding Preference` = `Datagram`, determine the length of the fields and payload and send the Object as datagram. In certain scenarios where the object size can be larger than maximum datagram size for the session, the Object will be dropped. 
 
 ~~~
 OBJECT_DATAGRAM Message {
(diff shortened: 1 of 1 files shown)


fluffy (approved):
LGTM

kixcord:
We should have some text on what to do if the datagram extension is not negotiated.

afrind:
> We should have some text on what to do if the datagram extension is not negotiated.

Do you want a behavior other than "all datagrams will be dropped"?  We could have the publisher reply with SUBSCRIBE ERROR.

suhasHere:
Datagram is object preference like the other one. I don't understand the meaning of not being negotiated. The same rule applies to all the object preferences and if we want to say something then we need to add it in general section.

afrind:
> I don't understand the meaning of not being negotiated.

I think Luke meant if the QUIC Datagram extension is not negotiated.  I believe it's required for WebTransport, but for raw QUIC the peer may not advertise support.  We could also require it for MoQ also just simply fail the entire MoQ connection without datagrams.

suhasHere:
> > I don't understand the meaning of not being negotiated.
> 
> I think Luke meant if the QUIC Datagram extension is not negotiated. I believe it's required for WebTransport, but for raw QUIC the peer may not advertise support. We could also require it for MoQ also just simply fail the entire MoQ connection without datagrams.

Ah thanks for the clarification.  Makes sense , but I also wonder if that would be in its own PR as it is not specific to object preference. Thoughts ?

kixelated:
> > > I don't understand the meaning of not being negotiated.
> > 
> > 
> > I think Luke meant if the QUIC Datagram extension is not negotiated. I believe it's required for WebTransport, but for raw QUIC the peer may not advertise support. We could also require it for MoQ also just simply fail the entire MoQ connection without datagrams.
> 
> Ah thanks for the clarification. Makes sense , but I also wonder if that would be in its own PR as it is not specific to object preference. Thoughts ?

I think we should make the datagram extension required as part of this PR.

The PREFER_DATAGRAM message over streams was a fallback when datagrams are not supported. By removing that fallback, we're in a state where servers and clients might silently drop all datagrams which is no bueno.

@ianswett we should have a section in connection establishment section that requires datagram support.

afrind:
Here's the text from WebTransport if we want to copy it:

```
WebTransport over HTTP/3 also requires support for QUIC datagrams. To indicate support, both the client and the server MUST send a max_datagram_frame_size transport parameter with a value greater than 0 (see Section 3 of [QUIC-DATAGRAM]).
```

ianswett:
I wrote #384 to require DATAGRAM support when using QUIC directly.

ianswett (approved):

Issue #350, opened 2024-01-26
When to use Track Alias or Subscribe ID?


martinduke:
Sec 6.4.2: "If the Track Alias is already in	
 	      use, the publisher MUST close the session with a Duplicate Track	
 	      Alias error (Section 3.5)."

This makes sense if it's a different track name. But if it's the same track for two subscriptions, they should have the same track_alias, right? What's the purpose of track alias and subscribe ID always going together in objects, anyway?

martinduke:
Subscribe OK has just the ID. Subscribe_error has DI and alias. What?

martinduke:
Having read through this now, I think we should just eliminate Subscribe_id from the Object message types.

If there are overlapping subscriptions, senders should only send it once, and it's ambiguous what subscribe ID to use.

afrind:
Individual Comment: 

> If there are overlapping subscriptions, senders should only send it once, and it's ambiguous what subscribe ID to use.

I think the idea is that you have to send the object multiple times if it matches more than one subscription.  Without subscription ID, there's no way for a receiver to know if a received object matches the relative subscribe locations it sent.  That said, we could come up with a format that allows for more than one subscription ID, which is almost certainly more compact. 

There was also concern about mixing objects from different subscriptions in the same multi-object stream.

martinduke:
I confess to not having thought about relays as much as most, but I'm not sure how this model works for relays.

1) Clearly, the relay cannot just forward each subscription, as that totally negate the scalability benefit.

2) If the relay is aggrating subscriptions which come in over time, that is going to result in duplicate objects en masse? Or are we planning to have relays routinely SUBSCRIBE_FIN when a new subscription comes in, so it can send a new consolidated SUBSCRIBE?

3) Or are we counting on the vast majority of client subscriptions being "give me the next one", so as not to require a new relay subscription to produce?

martinduke:
also: how does sending multiple copies relate to forwarding preference? i.e. if I send the same object twice, is it on the same stream or on different streams?

hardie:
On Mon, Jan 29, 2024 at 5:47 PM Martin Duke ***@***.***>
wrote:

> also: how does sending multiple copies relate to forwarding preference?
> i.e. if I send the same object twice, is it on the same stream or on
> different streams?
>
The current theory, at least as I understand it, is that it depends on how
many subscriptions there are.  If the two objects are being sent because
there are two active subscriptions, they would be on different streams even
[...]

PR #375, opened 2024-02-09
One forwarding preference per track

### LINKED ISSUE #351: Is the scope of Forwarding Preference really Objects?
Draft-02 strongly implies the forwarding preference is a per-object decision. thus the following sequences would be legal:
[sequences are written as (group, object)]

Stream 2: (0, 1), (0, 2) (0, 3) (0, 5) -- type is StreamHeaderGroup
Stream 6: (0, 4) -- type is ObjectStream

or

Stream 2: (0, 0), (1, 0), (2, 0), (3, 0) -- type is StreamHeaderTrack
Stream 6: (1,1) -- type is Object Stream

etc.

If someone can imagine a use case for this, that's great. If not, it would be easier to implement with different assumptions, e.g.

(1) Forwarding preference is a track-level property. Thus the preference could be provided in the SUBSCRIBE_OK and we wouldn't evens strictly need four different object codepoints (though you would want the distinct formats, and might want to retain the codepoints for that reason)

(2) Forwarding preference is group-level -- some groups could be aggregated under StreamHeaderTrack, but others could be under StreamHeaderGroup. If there is a StreamHeaderGroup for a group, there MUST NOT be object-level streams for that group.

Again, I have little sense of the use cases out there, but I think we should explicitly decide what we are doing and make it clear in  the doc.

martinduke:
Fixes #351.

There could stand to be (1) a section on a subscriptions, and (2) rearranging the sections here to pull objects out of the message section. But I think it would be easier to review if cut-and-paste happens in a separate PR.

There are other issues (e.g. explicit payload length all the time) that are not addressed in this PR.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1255,7 +1255,10 @@ exactly one stream.
 WebTransport datagram, if possible. If not possible, it is delivered on exactly
 one stream.
 
-Any other value terminates the session with error PROTOCOL_VIOLATION.
+Any other value terminates the session with error PROTOCOL_VIOLATION. Senders
+MUST NOT send SUBSCRIBE_OKs with different forwarding preferences for the same
+track. Receivers SHOULD terminate the session with error PROTOCOL_VIOLATION if
+they can detect this condition.
 
 * Expires: Time in milliseconds after which the subscription is no
 longer valid. A value of 0 indicates that the subscription stays active
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -886,19 +886,19 @@ group.
 * Object Payload: An opaque payload intended for the consumer and SHOULD
 NOT be processed by a relay.
 
-### Object Message Formats
+### Object Header Formats
 
 Objects are delivered in unidirectional streams. Each object is preceded by a
 header that contains metadata about the object. The format of this header
-depednds on the forwarding preference for the track, as communicated by the
-SUBSCRIBE_OK message {{message-subscribe-ok}}.
+depends on the forwarding preference for the track, as communicated by the
+SUBSCRIBE_OK message ({{message-subscribe-ok}}).
 
 The first bytes of every unidirectional stream encode the subscribe ID, which
-refers to a SUBSCRIBE_OK that contains parsing instructions. Note that if
-a streams arrives for a subscribe_id for which a SUBSCRIBE_OK has not arrived,
-the receiver ought to buffer objects for that stream until it receives the
-SUBSCRIBE_OK. Due to resource constraints, the receiver could opt to send
-STOP_SENDING for the stream instead.
+refers to a SUBSCRIBE_OK that contains parsing instructions. If a stream arrives
+for a subscribe_id for which a SUBSCRIBE_OK has not arrived, the receiver SHOULD
+buffer objects for that stream until it receives the SUBSCRIBE_OK. Due to
+resource constraints, the receiver could opt to send STOP_SENDING for the stream
+instead.
 
 **Object or Prefer Datagram Forwarding Preference**
 
@@ -966,8 +966,8 @@ their corresponding object payloads.
 If a stream ends gracefully in the middle of a serialized Object, terminate the
 session with a Protocol Violation.
 
-A sender MUST NOT open more than stream for a given group and the same
-subscription.
+A sender MUST NOT open more than one stream for a given tuple of subscribe_id,
+track_alias, and group_id.
 
 TODO: figure out how a relay closes these streams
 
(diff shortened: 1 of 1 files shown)


martinduke:
It's subscription-level rather than track-level because

(1) the only convenient place to communicate forwarding preference is SUBSCRIBE_OK
(2) as it currently stands, objects are sent once per subscription
(3) I am not a Use Case Haver, but I can see that an archival subscribe might want a single track stream while the live edge probably would not, even for the same track.

suhasHere (changes requested):
This changes the direction of the original intent with the forwarding preference. Originally it was one preferred producer's intent that applies uniformly to all the subscribers. With the current proposal, it seems like it can vary across subscribers IIUC.

Am i understanding the proposal properly ?

martinduke:
@suhasHere yes, see the comment immediately before your question. If this is really objectionable, we can change it, but it would require an error condition check if two SUBSCRIBE_OKs had different preferences for the same track_alias.

I have no strong opinion on this, but this is way easier. When I write a server, it will probably always use one preference for a track.

If we change objects to be sent once per track, instead of once per subscription, this would have to change.

fluffy (changes requested):
Lets talk about this but I don't think this reflect what was said in the meeting and I don't think it will work. We said we could move to one forwarding preference per track, not per subscription. 

As this is not, I a very much against merging it.

ianswett:
I expect this will be a straightforward PR once we figure out the FETCH/SUBSCRIBE work, but in the meantime it's caught up in #350

martinduke:
No one seemed to like subscription scope, so I changed it.

I think we can merge this before the FETCH/SUBSCRIBE stuff. The bulk of the changes are in the Object message.

draft-ietf-moq-transport.md
```diff
@@ -1279,13 +1232,34 @@ A SUBSCRIBE_OK control message is sent for successful subscriptions.
 SUBSCRIBE_OK
 {
   Subscribe ID (i),
+  Forwarding Preference (8),
```
suhasHere, line 1235: i personally feel that all the track level information can be learnt from catalog.  I wonder if we should have anything in Subscribe OK ?

wilaw: But relays need to know the forwarding preference and they can't read the catalog, so SUBSCRIBE_OK is a good location to carry subscription-scoped metadata.

suhasHere: Out encoding has all the necessary information to know the object forwarding preference. my implementation uses that info and sets up the needed forwarding preference today.

martinduke: The PR eliminates that encoding to save a byte on objects, since it's extraneous.

kixelated (changes requested):
I think we just need some text to say it's a protocol violation (connection closed) if you do dumb like have two streams cover the same object/group/track twice. This PR prevents some instances of that, but at the cost of making decoding more annoying.

Frankly, I don't even want the relay to know about forwarding preference. It should just receive a stream, apply the subscription as a filter, send the stream. The stream headers are only meant to remove redundant encoding.

afrind:
> Frankly, I don't even want the relay to know about forwarding preference. It should just receive a stream, apply the subscription as a filter, send the stream. The stream headers are only meant to remove redundant encoding.

Chair Comment:

The WG agreed to have an explicit encoding of the forwarding preference for now and revisit that based on experimental data at a later date.  The current signal serves dual purpose with object property compression.

Individual Comment:

The publisher should be able to communicate immutable Track properties in Subscribe OK without those being properties of the Subscribe.  Using HTTP as an example, response headers contain immutable properties of the requested resource (Content-Length) and variable properties (Date, Server).

suhasHere:
> The publisher should be able to communicate immutable Track properties in Subscribe OK without those being properties of the Subscribe. Using HTTP as an example, response headers contain immutable properties of the requested resource (Content-Length) and variable properties (Date, Server).

In HTTP the request is between the client and the Origin server, but in MOQT the subscribe req/resp is a hop-by-hop property.  Also we have catalog that lists track level immutable properties anyways.  

There are 3  kinds of information that would be good to learn and we need to separate architectural constraints for the same


- Static track level properties like track info, preferences -> this can go in catalog and hence served over control plane/slow path.

- Dynamic properties reflecting the current state of objects like live/expiry edge, last object/group info --> this would be best served on data plane/fast path along with data. 

- Data that relay doesn't usually have access to but it needs to make decision on per object basis. this includes  object properties like priority, its preferences, alias (some of which can be inherited from track level settings and set by the original producer endpoint when publishing) --> this would be best served on the fast path as well.

ianswett:
After discussion at today's meeting, the conclusion was that we'd like to do the minimal change of saying you can't have different forwarding preferences for a given track and if a subscriber receives different ones, close the session with an error.

PR #376, opened 2024-02-11
Objects are identical

### LINKED ISSUE #361: Objects with identical names should be bit identical
This came as part of interim at Denver when discussing multiple announce for same track.
There needs to text that mandates, if 2 objects comes with same name, they need to be bit identical.

Also needs to clarify on how long the name needs to be unique

ianswett:
Regardless of where they're retrieved from and over time.

Fixes #361

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -226,8 +226,12 @@ groups and tracks.
 The basic data element of MOQT is an object.  An object is an
 addressable unit whose payload is a sequence of bytes.  All objects
 belong to a group, indicating ordering and potential
-dependencies. {{model-group}} Objects are comprised of two parts:
-metadata and a payload.  The metadata is never encrypted and is always
+dependencies. {{model-group}}  An object is uniquely identified by
+its namespace, name, group ID and object ID, and must be an identical
+sequence of bytes regardless of how or where it is retrieved.
+
+Objects are comprised of two parts: metadata and a payload.
+The metadata is never encrypted and is always
 visible to relays. The payload portion may be encrypted, in which case
 it is only visible to the producer and consumer. The application is
 solely responsible for the content of the object payload. This includes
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -226,8 +226,12 @@ groups and tracks.
 The basic data element of MOQT is an object.  An object is an
 addressable unit whose payload is a sequence of bytes.  All objects
 belong to a group, indicating ordering and potential
-dependencies. {{model-group}} Objects are comprised of two parts:
-metadata and a payload.  The metadata is never encrypted and is always
+dependencies. {{model-group}}  An object is uniquely identified by
+its namespace, name, group ID and object ID, and must be an identical
```
fluffy, line 230: perhaps "track namespace" and "track name".

draft-ietf-moq-transport.md
```diff
@@ -226,8 +226,12 @@ groups and tracks.
...
+sequence of bytes regardless of how or where it is retrieved.
```
fluffy, line 231: So if it has expired, or gets an error, you don't get the bytes. I wonder if it is better to say something more like they contents of an object can never be changed and if it is retried, will always have the same bit wise exact data in it.

fluffy (changes requested):

fluffy:
I put a two very small nit comments but don't care either way. They do not change the substance of the PR and +1 on the overall thing this is saying.

draft-ietf-moq-transport.md
ianswett: I tried to capture the idea that it doesn't change over time, but I couldn't figure out a good rephrase for the rest, so I left it as is for now.  Feel free to write an editorial PR to improve the text after it is merged if you'd like.

Issue #372, opened 2024-02-08
Create an ability to pre-warm a track


wilaw:
At the interim there was interest in creating the ability for a receiver to ask a sender to subscribe to a track but not to forward it. The purpose of this is to minimize the delay when the receiver SUBSCRIBES to the track in the future. 

Design considerations
1. This would need to be converted to a regular SUBSCRIBE for forward requests. So any relay receiving a PREWARM request MUST never forward it.
2. Is there a matching UNPREWARM, or would a regular UNSUBSCRIBE suffice?
3. Do we need a new method PREWARM or can it be a flag on the existing SUBSCRIBE?
4. This is a powerful DDOS attack vector. How do we mitigate that threat in the design?

ianswett:
I find this concerning from a DDoS/memory exhaustion perspective.

Is the motivation for this to minimize the latency of switching resolutions for video playback?  If so, couldn't you just issue a SUBSCRIBE at a lower priority and if it starts coming quickly, then start switching relatively soon, and if it takes a while, wait to switch.  Or is there a different usecase?

wilaw:
> Is the motivation for this to minimize the latency of switching resolutions for video playback? 

Yes. 

> If so, couldn't you just issue a SUBSCRIBE at a lower priority and if it starts coming quickly, then start switching relatively soon, and if it takes a while, wait to switch

That would work for switching up, where you have the luxury of time (your switch can always occur later).  However the primary use case is in switching down, where you don't want to incur in the extra delay in the relay having to go back to origin to retrieve the new track. It also is an optimization when you have only 1 or a few clients subscribed to that relay for that presentation. Given a sufficiently large cohort of clients, the relay would be naturally warmed. 

> I find this concerning from a DDoS/memory exhaustion perspective.

I agree on the DDOS concern, as the OP states. However, from a memory exhaustion, this is really no different from having  a multiplicity of clients each pulling a different track from the relay. 1 client pulling 3 tracks is the same relay memory requirement as 100 clients asking for those same 3 tracks.

gwendalsimon:
> I find this concerning from a DDoS/memory exhaustion perspective.

+1 on the concern. The risk of overloading the CDN backbone unnecessarily is significant. Furthermore, any just-in-time processing (transcoding and packaging) would be useless although these are not only money but also energy reduction

ianswett:
[...]

PR #381, opened 2024-02-16
Subscriber indicator on track readiness

suhasHere:
Attempts to address some of the usecases discussed in #372 

Love to hear feedback

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -588,7 +588,7 @@ subscribers for each track. For subscriptions with `DeliveryPreference`
 of `0x1 Ready` , each new OBJECT belonging to the
 track within the subscription range is forwarded to all the matching
 subscribers, dependent on the congestion response. For subscriptions with
-`DeliveryPreference` value of `0x0 Park`, a relay implementation
+`DeliveryPreference` value of `0x0 Freeze`, a relay implementation
 may choose to subscribe upstream to receive objects in anticipation of
 having media readily available. A subscription
 remains active until it expires, until the publisher of the track
@@ -1197,7 +1197,7 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
-* DeliveryPreference: The subscriber's preference to indicate its readiness to accept the published media. A DeliverPreference value of `Park (0x0)`  specifies
+* DeliveryPreference: The subscriber's preference to indicate its readiness to accept the published media. A DeliverPreference value of `Freeze (0x0)`  specifies
 media delivery MUST be paused by the publisher. A DeliverPreference value of `Ready (0x1)` specifies that the publisher MUST start delivering the objects for the track indicated by the `SUBSCRIBE` as and when they are
 available. A subscriber can update the DeliveryPreference for a track at
 any point by issuing a new `SUBSCRIBE` message with `Subscribe ID` matching the original `SUBSCRIBE` message.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1168,7 +1168,7 @@ SUBSCRIBE Message {
   Track Alias (i),
   Track Namespace (b),
   Track Name (b),
-  DeliveryPreference (i)
+  DeliveryPreference (i),
   StartGroup (Location),
   StartObject (Location),
   EndGroup (Location),
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1199,7 +1199,7 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 
 * DeliveryPreference: The subscriber's preference to indicate its readiness to accept the published media. A DeliverPreference value of `Park (0x0)`  specifies
 media delivery MUST be paused by the publisher. A DeliverPreference value of `Ready (0x1)` specifies that the publisher MUST start delivering the objects for the track indicated by the `SUBSCRIBE` as and when they are
-available. A Subscriber can update the DeliveryPreference for a track at
+available. A subscriber can update the DeliveryPreference for a track at
 any point by issuing a new `SUBSCRIBE` message with `Subscribe ID` matching the original `SUBSCRIBE` message.
 
 * StartGroup: The Location of the requested group.  StartGroup's Mode MUST NOT be
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -626,7 +626,7 @@ a given SUBSCRIBE of either SUBSCRIBE_OK or SUBSCRIBE_ERROR.
 
 For successful subscriptions, the publisher maintains a list of
 subscribers for each track. For subscriptions with `DeliveryPreference`
-of `0x1 Ready` , each new OBJECT belonging to the
+of `0x1 Ready`, each new OBJECT belonging to the
 track within the subscription range is forwarded to all the matching
 subscribers, dependent on the congestion response. For subscriptions with
 `DeliveryPreference` value of `0x0 Freeze`, a relay implementation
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -319,7 +319,7 @@ URI with a "moq" scheme.  The "moq" URI scheme is defined as follows,
 using definitions from {{!RFC3986}}:
 
 ~~~~~~~~~~~~~~~
-moq-URI = "moqt" "://" authority path-abempty [ "?" query ]
+moq-URI = "moq" "://" authority path-abempty [ "?" query ]
 ~~~~~~~~~~~~~~~
```
suhasHere, line 323: i will revert this change. Looks like merge issue

draft-ietf-moq-transport.md
```diff
@@ -584,9 +584,13 @@ The entity receiving the SUBSCRIBE MUST send only a single response to
...
 
 For successful subscriptions, the publisher maintains a list of
-subscribers for each track. Each new OBJECT belonging to the
-track within the subscription range is forwarded to each active
-subscriber, dependent on the congestion response. A subscription
+subscribers for each track. For subscriptions with `DeliveryPreference`
+of `0x1 Ready` , each new OBJECT belonging to the
+track within the subscription range is forwarded to all the matching
+subscribers, dependent on the congestion response. For subscriptions with
+`DeliveryPreference` value of `0x0 Park`, a relay implementation
```
afrind, line 591: Individual Comment:

How do you unpark?  Do you send another subscribe with `Ready`, or maybe you were thinking subscribe update would allow you to do that?  Can a subscriber go from Ready back to Parked?

suhasHere: Sending subscribe with same subscribe_id is used to update an existing subscription. Yes it is possible to move between park and ready states multiple times.

fluffy: For what it is worth, I prefer calling these frozen instead of parked but I don't care much

kixelated (commented):
This approach is reasonable, but I dunno I'd rather punt pre-warming. It's an optional, pre-mature optimization that I don't imagine anyone will implement (on either side).

draft-ietf-moq-transport.md
zafergurel: What would be the response of the relay if DeliveryPreference is `Park` in the Subscribe message and relay does not support/prefer parking?

suhasHere: The relay returns Subscribe Error in that case

fluffy: I don't think this should be something that is optional for Relays.

fluffy (approved):
A small nit, in calling , park and pickup are commonly used to mean something different than this that has to do with moving media from one device to another so I would prefer a name other than park. That said, I can live with park and I think we should merge this.

suhasHere:
> A small nit, in calling , park and pickup are commonly used to mean something different than this that has to do with moving media from one device to another so I would prefer a name other than park. That said, I can live with park and I think we should merge this.

Updated to rename park to freeze.

draft-ietf-moq-transport.md
```diff
@@ -625,9 +625,13 @@ The entity receiving the SUBSCRIBE MUST send only a single response to
 a given SUBSCRIBE of either SUBSCRIBE_OK or SUBSCRIBE_ERROR.
 
 For successful subscriptions, the publisher maintains a list of
-subscribers for each track. Each new OBJECT belonging to the
-track within the subscription range is forwarded to each active
-subscriber, dependent on the congestion response. A subscription
+subscribers for each track. For subscriptions with `DeliveryPreference`
+of `0x1 Ready` , each new OBJECT belonging to the
```
afrind, line 629: nit: extra whitespace

draft-ietf-moq-transport.md
```diff
@@ -625,9 +625,13 @@ The entity receiving the SUBSCRIBE MUST send only a single response to
...
+track within the subscription range is forwarded to all the matching
+subscribers, dependent on the congestion response. For subscriptions with
+`DeliveryPreference` value of `0x0 Freeze`, a relay implementation
+may choose to subscribe upstream to receive objects in anticipation of
```
afrind, line 633: lowercase `may` - did you want MAY or something else?

My read of this text is that it's ok for a relay to send a subscribe OK and not issue an upstream subscription, or issue an upstream subscription with pref=Freeze.  Is that your intent?

draft-ietf-moq-transport.md
```diff
@@ -831,6 +835,7 @@ information in a SUBSCRIBE or ANNOUNCE message. This parameter is populated for
 cases where the authorization is required at the track level. The value is an
 ASCII string.
 
+
```
afrind, line 838: nit: extra newline?

draft-ietf-moq-transport.md
```diff
@@ -1248,6 +1254,11 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
+* DeliveryPreference: The subscriber's preference to indicate its readiness to accept the published media. A DeliverPreference value of `Freeze (0x0)`  specifies
+media delivery MUST be paused by the publisher. A DeliverPreference value of `Ready (0x1)` specifies that the publisher MUST start delivering the objects for the track indicated by the `SUBSCRIBE` as and when they are
+available. A subscriber can update the DeliveryPreference for a track at
+any point by issuing a new `SUBSCRIBE` message with `Subscribe ID` matching the original `SUBSCRIBE` message.
```
afrind, line 1260: I'm a little uneasy about defining the semantics of subscribe update in this PR, which has a quite different thesis, and also without explaining the implications of changing location preferences.  If others are ok to do that in a follow up I can be convinced.

afrind (commented):
Individual Review

draft-ietf-moq-transport.md
suhasHere: This part is defining subscirbe update in the context of DeliveryPreference. I think we should add text to define the same in terms of location too. I do see these coming in 2 different PRs thouhg.

draft-ietf-moq-transport.md
suhasHere: Since every hop is a new subscription, Relays can make either of the choices.  In that context, your interpretation is correct.

ianswett:
TBH, I'm still not convinced we need this.  I feel like it's a fix for a problem we may have in the future, but we're unsure, because we haven't resolved a number of other details about how priorities, unreliability, etc work.

fluffy:
In the past, I had asked if this was actually significantly better than just unsubscribe then re subscribe and I was talking to someone that had implemented this and got an interesting insight that I wanted to add here. In a multithreaded relay, creating a subscription will typically require a global lock to create the upstream subscription to deal with case of two clients trying to subscribe to same thing at the same time. However, this subscribe update would only need to lock the one downstream subscription and if there was only one thread handle that subscription, there may be no need of a lock. 

That seemed like a pretty compelling argument to be able to do this a subscribe update that changed the state of the subscribe vs constantly deleting and recreating the whole subscribe.

afrind:
Individual Comment: 

I have an alternate proposal that will solve this use case and I think a couple others, which is adding a form of flow control or rate control to subscriptions. Subscribe would have a param like "max group", and the publisher could only send up to the max group before the subscription would pause.  The subscriber would unblock by sending a larger max group.  So you could implement "parked" by sending max group = 0.

The additional use case is related to VoD and live rewind, where the subscriber is consuming the track at playback speed but behind the live head.  Today that's accomplished with the client making a series of range subscribes.

suhasHere:
I would like to see some form of this discussed as this is helping a real functionality . @afrind @martinduke wondering any thoughts on how should we proceed ? Thanks

draft-ietf-moq-transport.md
fluffy: Probably best to rephrase not to use lower case may just to avoid any confusion about what it means

fluffy (approved):

ianswett:
In an effort to get the number of open PRs down, I'd like to close this for now, and we can revisit it later.

PR #382, opened 2024-02-18
Forwarding preference can't change for a track

### LINKED ISSUE #351: Is the scope of Forwarding Preference really Objects?

ianswett:
Fixes #351 

Closes #375

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -901,7 +901,7 @@ NOT be processed by a relay.
 
 Every Track has a single 'Object Forwarding Preference' and publishers
 MUST NOT mix different forwarding preferences within a single track.
-If a subscriber receives different forward prefernces for a track, it
+If a subscriber receives different forwarding preferences for a track, it
 SHOULD close the session with an error of 'Protocol Violation'.
 
 **Object Stream Message**
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -899,6 +899,11 @@ NOT be processed by a relay.
 
 ### Object Message Formats
 
+Every Track has a single 'Object Forwarding Preference' and publishers
+MUST NOT mix different forwarding preferences within a single track.
+If a subscriber receives different forward prefernces for a track, it
+SHOULD close the session with an error of 'Protocol Violation'.
+
 **Object Stream Message**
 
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -899,6 +899,11 @@ NOT be processed by a relay.
 
 ### Object Message Formats
 
+Every Track has a single 'Object Forwarding Preference' and publishers
+MUST NOT mix different forwarding preferences within a single track.
+If a subscriber receives different forward prefernces for a track, it
```
suhasHere, line 904: type on prefernces

martinduke (approved):
This is an improvement, though I would still like to get rid of all the wasted bits

vasilvv (approved):

draft-ietf-moq-transport.md
ianswett, line 904:
```suggestion
If a subscriber receives different forwarding preferences for a track, it
```

kixelated:
> This is an improvement, though I would still like to get rid of all the wasted bits

They won't be wasted until we have 64 message types valid on unidirectional streams. We could save the message type byte if OBJECT is the only message allowed on unidirectional streams, but that's not forward-compatible.

PR #383, opened 2024-02-18
Remove 'Generic Error'

ianswett:
'Generic Error' is never used in the draft and given "Protocol Violation" is also available, I don't think it's useful.

I didn't renumber the error codes, but I could.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -405,8 +405,6 @@ code, as defined below:
 |-----:|:--------------------------|
 | 0x0  | No Error                  |
 |------|---------------------------|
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

vasilvv:
We use this quite a bit in our code similar to how we use "internal error" in QUIC.  Maybe rename it to just that?  We definitely need some kind of "syscall failed unexpectedly and now I just have to tear down your connection" code.

ianswett:
Thanks Victor, so is the intent that this is similar to a 500 in HTTP?  If so, renaming it to 'Internal Error' seems appropriate.

PR #384, opened 2024-02-18
Require DATAGRAM support when using QUIC directly

ianswett:
From discussion on #360

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -338,9 +338,9 @@ over a QUIC connection directly [QUIC], and over WebTransport
 [WebTransport].  Both provide streams and datagrams with similar
 semantics (see {{?I-D.ietf-webtrans-overview, Section 4}}); thus, the
 main difference lies in how the servers are identified and how the
-connection is established.  QUIC datagrams MUST be supported via the
-[QUIC-DATAGRAM] extension, which is already a requirement for
-WebTransport over HTTP/3.
+connection is established.  When using QUIC, QUIC datagrams MUST be
+supported via the [QUIC-DATAGRAM] extension, which is already a
+requirement for WebTransport over HTTP/3.
 
 There is no definition of the protocol over other transports,
 such as TCP, and applications using MoQ might need to fallback to
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -338,7 +338,7 @@ over a QUIC connection directly [QUIC], and over WebTransport
 [WebTransport].  Both provide streams and datagrams with similar
 semantics (see {{?I-D.ietf-webtrans-overview, Section 4}}); thus, the
 main difference lies in how the servers are identified and how the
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -338,10 +338,15 @@ over a QUIC connection directly [QUIC], and over WebTransport
 [WebTransport].  Both provide streams and datagrams with similar
 semantics (see {{?I-D.ietf-webtrans-overview, Section 4}}); thus, the
 main difference lies in how the servers are identified and how the
-connection is established.  There is no definition of the protocol
-over other transports, such as TCP, and applications using MoQ might
-need to fallback to another protocol when QUIC or WebTransport aren't
-available.
+connection is established.  To ensure consistent semantics, use of
+QUIC also requires support for QUIC datagrams. To indicate support,
```
kixelated, line 342: This phrase is a bit awkward since "use of QUIC" includes WebTransport. It's not immediately clear to the reader that WebTransport already requires support.

I would say:
> QUIC datagrams MUST be supported via the [QUIC-DATAGRAM] extension, which is already a requirement for WebTransport over HTTP/3.

Or just say datagrams are required and punt the details, including the `max_dataframe_frame_size` parameter, to the QUIC section below.

draft-ietf-moq-transport.md
```diff
@@ -338,10 +338,15 @@ over a QUIC connection directly [QUIC], and over WebTransport
...
+both the client and the server MUST send a max_datagram_frame_size
+transport parameter with a value greater than 0
```
kixelated, line 344: Do we want a minimum/maximum size? QUIC can get away with negotiating the size but we can't really because of relays.

I would have a datagram section with something like:

> To facilitate relays, datagrams SHOULD NOT be larger than 1150 bytes and `max_datafram_frame_size` SHOULD be greater than 1150 bytes.

I'm also just fine leaving it zero. Or punting this to its own issue.

draft-ietf-moq-transport.md
ianswett: Thanks, suggestion applied.

draft-ietf-moq-transport.md
```diff
@@ -338,10 +338,13 @@ over a QUIC connection directly [QUIC], and over WebTransport
 [WebTransport].  Both provide streams and datagrams with similar
 semantics (see {{?I-D.ietf-webtrans-overview, Section 4}}); thus, the
 main difference lies in how the servers are identified and how the
-connection is established.  There is no definition of the protocol
-over other transports, such as TCP, and applications using MoQ might
-need to fallback to another protocol when QUIC or WebTransport aren't
-available.
+connection is established.  When using QUIC, QUIC datagrams MUST be
```
ianswett, line 341:
```suggestion
connection is established.  When using QUIC, datagrams MUST be
```

kixelated (approved):

suhasHere (approved):

PR #385, opened 2024-02-19
Add SUBSCRIBE_UPDATE

### LINKED ISSUE #269: Multiple subscriptions and updating a subscription

### LINKED ISSUE #363: UNSUBSCRIBE is underspecified
UNSUBSCRIBE has some items I'd like clarified.

How do you know the last Group and Object that will be delivered?  Do you get a SUBSCRIBE_RST or SUBSCRIBE_FIN?  Or maybe we want to allow the UNSUBSCRIBE to specify that last Object it wants delivered, somewhat similar to a GOAWAY.  We could also allow the subscriber to RST the SUBSCRIBE?

Or maybe we don't care about knowing the final Object?

ianswett:
Fixes #363, Fixes #269

This attempts to provide the minimum functionality I expect we'll need, regardless of how SUBSCRIBE is changed in the near future.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1380,9 +1380,9 @@ SUBSCRIBE_ERROR
 
 ## SUBSCRIBE_UPDATE {#message-subscribe-update}
 
-A subscriber issues a `SUBSCRIBE_UPDATE` message to a publisher indicating it wants to
-modify the end of an existing subscription.  One common type of modification is
-unsubscription.
+A subscriber issues a `SUBSCRIBE_UPDATE` message to a publisher requesting a
+modification to the end of an existing subscription.  One common type of
+modification is unsubscription.
 
 If an update cannot be completed by the relay, it replies with a 'SUBSCRIBE_ERROR'
 with error code 'Update Failed'.  A SUBSCRIBE_UPDATE MUST NOT fail if the specified
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1380,7 +1380,7 @@ SUBSCRIBE_ERROR
 
 ## SUBSCRIBE_UPDATE {#message-subscribe-update}
 
-A subscriber issues a `SUBSCRIBE_UPDATE` message to a publisher requesting a
+A subscriber issues a `SUBSCRIBE_UPDATE` message to a publisher to request a
 modification to the end of an existing subscription.  One common type of
 modification is unsubscription.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1386,8 +1386,8 @@ modification is unsubscription.
 
 If an update cannot be completed by the relay, it replies with a SUBSCRIBE_ERROR
 with error code 'Update Failed' and the subscription is unchanged.
-A SUBSCRIBE_UPDATE MUST NOT fail if the specified EndGroup and EndObject have
-already been received.
+A SUBSCRIBE_UPDATE MUST NOT fail if the specified EndGroup and EndObject are
+earlier than the current values.
 
 The subscription is not fully terminated until a SUBSCRIBE_RST or SUBSCRIBE_FIN
 are received.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1385,8 +1385,12 @@ modification to the end of an existing subscription.  One common type of
 modification is unsubscription.
 
 If an update cannot be completed by the relay, it replies with a SUBSCRIBE_ERROR
-with error code 'Update Failed'.  A SUBSCRIBE_UPDATE MUST NOT fail if the specified
-EndGroup and EndObject have already been received.
+with error code 'Update Failed' and the subscription is unchanged.
+A SUBSCRIBE_UPDATE MUST NOT fail if the specified EndGroup and EndObject have
+already been received.
+
+The subscription is not fully terminated until a SUBSCRIBE_RST or SUBSCRIBE_FIN
+are received.
 
 The format of `SUBSCRIBE_UPDATE` is as follows:
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -650,7 +650,8 @@ as defined below:
 |------|---------------------------|
 | 0x2  | Retry Track Alias         |
 |------|---------------------------|
-
+| 0x3  | Update Failed             |
+|------|---------------------------|
 
 ## Publisher Interactions
 
@@ -1382,6 +1383,10 @@ SUBSCRIBE_ERROR
 A subscriber issues a `SUBSCRIBE_UPDATE` message to a publisher indicating it wants to
 modify the end of an existing subscription.  One common type of modification is
 unsubscription.
+
+If an update cannot be completed by the relay, it replies with a 'SUBSCRIBE_ERROR'
+with error code 'Update Failed'.  A SUBSCRIBE_UPDATE MUST NOT fail if the specified
+EndGroup and EndObject have already been received.
 
 The format of `SUBSCRIBE_UPDATE` is as follows:
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -487,10 +487,9 @@ SHOULD terminate the session with 'GOAWAY Timeout' after a sufficient timeout if
 there are still open subscriptions on a connection.
 
 The GOAWAY message does not immediately impact subscription state. A subscriber
-SHOULD individually UNSUBSCRIBE for each existing subscription, while a
+SHOULD individually unsubscribe for each existing subscription, while a
 publisher MAY reject new SUBSCRIBEs while in the draining state. When the server
-is a subscriber, it SHOULD send a GOAWAY message prior to any UNSUBSCRIBE
-messages.
+is a subscriber, it SHOULD send a GOAWAY message prior to unsubscribing.
 
 After the client receives a GOAWAY, it's RECOMMENDED that the client waits until
 there are no more active subscriptions before closing the session with NO_ERROR.
@@ -744,7 +743,7 @@ MOQT Message {
 |-------|-----------------------------------------------------|
 | 0x9   | UNANNOUNCE  ({{message-unannounce}})                |
 |-------|-----------------------------------------------------|
-| 0xA   | UNSUBSCRIBE ({{message-unsubscribe}})               |
+| 0xA   | SUBSCRIBE_UPDATE ({{message-subscribe-update}})     |
 |-------|-----------------------------------------------------|
 | 0xB   | SUBSCRIBE_FIN ({{message-subscribe-fin}})           |
 |-------|-----------------------------------------------------|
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1378,22 +1378,36 @@ SUBSCRIBE_ERROR
   the receiver MUST close the connection with a Duplicate Track Alias error
   ({{session-termination}}).
 
-## UNSUBSCRIBE {#message-unsubscribe}
+## SUBSCRIBE_UPDATE {#message-subscribe-update}
 
-A subscriber issues a `UNSUBSCRIBE` message to a publisher indicating it is no
-longer interested in receiving media for the specified track.
+A subscriber issues a `SUBSCRIBE_UPDATE` message to a publisher requesting a
```
ianswett, line 1383:
```suggestion
A subscriber issues a `SUBSCRIBE_UPDATE` message to a publisher to request a
```

draft-ietf-moq-transport.md
```diff
@@ -1378,22 +1378,36 @@ SUBSCRIBE_ERROR
...
+## SUBSCRIBE_UPDATE {#message-subscribe-update}
 
-A subscriber issues a `UNSUBSCRIBE` message to a publisher indicating it is no
-longer interested in receiving media for the specified track.
+A subscriber issues a `SUBSCRIBE_UPDATE` message to a publisher to request a
+modification to the end of an existing subscription.  One common type of
+modification is unsubscription.
 
-The format of `UNSUBSCRIBE` is as follows:
+If an update cannot be completed by the relay, it replies with a 'SUBSCRIBE_ERROR'
```
ianswett, line 1387:
```suggestion
If an update cannot be completed by the relay, it replies with a SUBSCRIBE_ERROR
```

draft-ietf-moq-transport.md
```diff
@@ -1378,22 +1378,40 @@ SUBSCRIBE_ERROR
...
+## SUBSCRIBE_UPDATE {#message-subscribe-update}
+
+A subscriber issues a `SUBSCRIBE_UPDATE` message to a publisher to request a
+modification to the end of an existing subscription.  One common type of
+modification is unsubscription.
+
+If an update cannot be completed by the relay, it replies with a SUBSCRIBE_ERROR
+with error code 'Update Failed' and the subscription is unchanged.
+A SUBSCRIBE_UPDATE MUST NOT fail if the specified EndGroup and EndObject have
+already been received.
```
ianswett, line 1390:
```suggestion
A SUBSCRIBE_UPDATE MUST NOT fail if the specified EndGroup and EndObject are
earlier than the current values.
```

draft-ietf-moq-transport.md
```diff
@@ -1378,22 +1378,40 @@ SUBSCRIBE_ERROR
...
+## SUBSCRIBE_UPDATE {#message-subscribe-update}
+
+A subscriber issues a `SUBSCRIBE_UPDATE` message to a publisher to request a
+modification to the end of an existing subscription.  One common type of
+modification is unsubscription.
+
+If an update cannot be completed by the relay, it replies with a SUBSCRIBE_ERROR
+with error code 'Update Failed' and the subscription is unchanged.
+A SUBSCRIBE_UPDATE MUST NOT fail if the specified EndGroup and EndObject are
+earlier than the current values.
```
ianswett, line 1390:
```suggestion
A SUBSCRIBE_UPDATE MUST NOT fail if the specified EndGroup and EndObject are
earlier than the current values for the subscription.
```

draft-ietf-moq-transport.md
```diff
@@ -1373,22 +1373,40 @@ SUBSCRIBE_ERROR
...
-The format of `UNSUBSCRIBE` is as follows:
+The format of `SUBSCRIBE_UPDATE` is as follows:
 
 ~~~
-UNSUBSCRIBE Message {
-  Subscribe ID (i)
+SUBSCRIBE_UPDATE Message {
+  Subscribe ID (i),
+  EndGroup (Location),
+  EndObject (Location),
```
suhasHere, line 1396: This needs to be optional and only set if the subscription needs to be ended.

draft-ietf-moq-transport.md
```diff
@@ -744,7 +744,7 @@ MOQT Message {
 |-------|-----------------------------------------------------|
 | 0x9   | UNANNOUNCE  ({{message-unannounce}})                |
 |-------|-----------------------------------------------------|
-| 0xA   | UNSUBSCRIBE ({{message-unsubscribe}})               |
+| 0xA   | SUBSCRIBE_UPDATE ({{message-subscribe-update}})     |
 |-------|-----------------------------------------------------|
```
suhasHere, line 748: I feel Unsubscribe makes more sense if one wants to end the subscription.
We don;t need a new message to update an existing subscription. A subscriber can update an existing subscription by sending `SUBSCRIBE` message with SubscribeId that matches an existing subscription.

suhasHere (changes requested):

draft-ietf-moq-transport.md
```diff
@@ -1373,22 +1373,40 @@ SUBSCRIBE_ERROR
...
   ({{session-termination}}).
 
-## UNSUBSCRIBE {#message-unsubscribe}
+## SUBSCRIBE_UPDATE {#message-subscribe-update}
+
+A subscriber issues a `SUBSCRIBE_UPDATE` message to a publisher to request a
+modification to the end of an existing subscription.  One common type of
+modification is unsubscription.
+
+If an update cannot be completed by the relay, it replies with a SUBSCRIBE_ERROR
```
kixelated, line 1382: But what about SUBSCRIBE_OK? We should send it too, especially if the end groups use relative offsets.

draft-ietf-moq-transport.md
```diff
@@ -1373,22 +1373,40 @@ SUBSCRIBE_ERROR
...
+with error code 'Update Failed' and the subscription is unchanged.
```
kixelated, line 1383: I don't like the subscription being unchanged. It causes a partial failure which is just really difficult to support.

draft-ietf-moq-transport.md
kixelated: Also what if there are multiple SUBSCRIBE_UPDATES? Do the SUBSCRIBE_ERRORs have to be returned in order? With no OK, it's impossible to differentiate if the 1st or 2nd UPDATE failed.

draft-ietf-moq-transport.md
```diff
@@ -1373,22 +1373,40 @@ SUBSCRIBE_ERROR
...
+The subscription is not fully terminated until a SUBSCRIBE_RST or SUBSCRIBE_FIN
+are received.
 
-The format of `UNSUBSCRIBE` is as follows:
+The format of `SUBSCRIBE_UPDATE` is as follows:
 
 ~~~
-UNSUBSCRIBE Message {
-  Subscribe ID (i)
+SUBSCRIBE_UPDATE Message {
```
kixelated, line 1393: Can we also update the start group/object? It would be super useful for relays to deal with:

viewer1: SUBSCRIBE start=-2
(pause)
viewer2: SUBSCRIBE start=-3

Currently, the relay would issue to origin:
SUBSCRIBE start=-2
(wait for SUBSCRIBE_OK -2=65)
SUBSCRIBE start=64 end=65

Which is complicated and incurs delay. But if we allow updating start ranges:
SUBSCRIBE start=-2
SUBSCRIBE_UPDATE start=-3

kixelated (changes requested):

draft-ietf-moq-transport.md
afrind: Updating the start point can mess with object mapping to streams, especially moving backwards.  In your example, if the track is stream-per-track, the publisher may open a stream and being publishing group=65.  In the current design, where subscription ID appears in objects, the second subscribe for group=64 will come on its own new stream.  If we allow updating the subscription start point backwards to 64, what should the publisher do? Should it close the open stream and open a new one, beginning at group=64, and resend objects from group=65?

draft-ietf-moq-transport.md
afrind: In the context of this PR and the current draft, where the only thing you can change is the end, this message is OK.  However, if we add more parameters to subscribe (eg #381), then we need a way to update some of the parameters and not others.

We should also clarify what happens if you issue the SUBSCRIBE twice for the same ID - I expect in this PR as is, we would make that an error?

afrind (commented):
Individual Review:

We should have a way to update the endpoint of a subscription (like subscribe update), and also a way to unsubscribe immediately when you don't know or care about the end.  I suppose we can do that by setting the new end-point to absolute 0/0.

draft-ietf-moq-transport.md
ianswett: I do think updating the start is potentially interesting and useful.

Updates of the start should not be able to move the start point of the subscription earlier.  As you said, it's just too complex.  For similar reasons, I think updating the end to be farther in the future is error prone as well, though there may be use cases for that.

draft-ietf-moq-transport.md
```diff
@@ -1387,22 +1387,40 @@ SUBSCRIBE_ERROR
   the receiver MUST close the connection with a Duplicate Track Alias error
   ({{session-termination}}).
 
-## UNSUBSCRIBE {#message-unsubscribe}
+## SUBSCRIBE_UPDATE {#message-subscribe-update}
+
+A subscriber issues a `SUBSCRIBE_UPDATE` message to a publisher to request a
+modification to the end of an existing subscription.  One common type of
```
fluffy, line 1393: I think the "end of existing" should say the "end of the subscription range".

draft-ietf-moq-transport.md
```diff
@@ -1387,22 +1387,40 @@ SUBSCRIBE_ERROR
...
+modification is unsubscription.
```
fluffy, line 1394: I don't think this is right. The update should just be to change the metadata about the subscribe, not change the state of the subscribe. We should use some other message to cause the SUSBCRIBE to end and remove all the state about the SUBSCRIBE.

draft-ietf-moq-transport.md
```diff
@@ -761,7 +761,7 @@ MOQT Message {
 |-------|-----------------------------------------------------|
 | 0x9   | UNANNOUNCE  ({{message-unannounce}})                |
 |-------|-----------------------------------------------------|
-| 0xA   | UNSUBSCRIBE ({{message-unsubscribe}})               |
+| 0xA   | SUBSCRIBE_UPDATE ({{message-subscribe-update}})     |
 |-------|-----------------------------------------------------|
```
fluffy, line 765: I think we need both the UNSUBSCRIBE and UPDATE ( or whatever names we decide to call them ).

draft-ietf-moq-transport.md
```diff
@@ -1387,22 +1387,40 @@ SUBSCRIBE_ERROR
...
   ({{session-termination}}).
 
-## UNSUBSCRIBE {#message-unsubscribe}
+## SUBSCRIBE_UPDATE {#message-subscribe-update}
+
+A subscriber issues a `SUBSCRIBE_UPDATE` message to a publisher to request a
+modification to the end of an existing subscription.  One common type of
+modification is unsubscription.
+
+If an update cannot be completed by the relay, it replies with a SUBSCRIBE_ERROR
```
fluffy, line 1396: I think this should be an UPDATE_OK/ERROR. Don't care if it is two message or not. But should be clear that if some part of the UPDATE can not be applied, then no part of the update is applied, subscribe stays with same meta data as before, and error returned. So an update is all or nothing.

draft-ietf-moq-transport.md
```diff
@@ -1387,22 +1387,40 @@ SUBSCRIBE_ERROR
...
+with error code 'Update Failed' and the subscription is unchanged.
+A SUBSCRIBE_UPDATE MUST NOT fail if the specified EndGroup and EndObject are
+earlier than the current values for the subscription.
 
-A subscriber issues a `UNSUBSCRIBE` message to a publisher indicating it is no
-longer interested in receiving media for the specified track.
+The subscription is not fully terminated until a SUBSCRIBE_RST or SUBSCRIBE_FIN
+are received.
```
fluffy, line 1402: Uh, I think this needs to be UNSUBSCRIBE or whatever the  SUBSCRIBE CLOSE/STOP/FIN/RST/END is called

draft-ietf-moq-transport.md
fluffy: +1 on start can only be moved forward, end can only be moved backwards

fluffy: As a side note where I don't really care ... I'm not sure the use case to move the start forward

draft-ietf-moq-transport.md
```diff
@@ -1387,22 +1387,40 @@ SUBSCRIBE_ERROR
...
 
-The format of `UNSUBSCRIBE` is as follows:
+The format of `SUBSCRIBE_UPDATE` is as follows:
 
 ~~~
-UNSUBSCRIBE Message {
-  Subscribe ID (i)
+SUBSCRIBE_UPDATE Message {
+  Subscribe ID (i),
+  EndGroup (Location),
```
fluffy, line 1409: This is not a location this shoudl be an optional var int

draft-ietf-moq-transport.md
```diff
@@ -1387,22 +1387,40 @@ SUBSCRIBE_ERROR
...
+  EndObject (Location),
```
fluffy, line 1410: This should be var int not Location

draft-ietf-moq-transport.md
```diff
@@ -1387,22 +1387,40 @@ SUBSCRIBE_ERROR
...
 }
 ~~~
-{: #moq-transport-unsubscribe-format title="MOQT UNSUBSCRIBE Message"}
+{: #moq-transport-subscribe-update-format title="MOQT SUBSCRIBE_UPDATE Message"}
 
 * Subscribe ID: Subscription Identifer as defined in {{message-subscribe-req}}.
 
+* EndGroup: The last Group requested in the subscription, inclusive.  EndGroup's
+Mode is None for an open-ended subscription.
```
fluffy, line 1418: I would prefer to make the end group / object optional in the update instead of having a None value. I think it makes a better patter for how to add future things to the update.

fluffy (changes requested):
My main comment is I don't think we shoudl try and use UPDATE as the way we do UNSUSCRIBE

afrind:
Is this overtaken by #435 ?

PR #386, opened 2024-02-19
Rename SUBSCRIBE_RST to SUBSCRIBE_RESET

ianswett:
QUIC uses RESET, we should too.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -627,8 +627,8 @@ track within the subscription range is forwarded to each active
 subscriber, dependent on the congestion response. A subscription
 remains active until it expires, until the publisher of the track
 terminates the track with a SUBSCRIBE_FIN
-(see {{message-subscribe-fin}}) or a SUBSCRIBE_RST
-(see {{message-subscribe-rst}}).
+(see {{message-subscribe-fin}}) or a SUBSCRIBE_RESET
+(see {{message-subscribe-reset}}).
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -751,7 +751,7 @@ MOQT Message {
 |-------|-----------------------------------------------------|
 | 0xB   | SUBSCRIBE_FIN ({{message-subscribe-fin}})           |
 |-------|-----------------------------------------------------|
-| 0xC   | SUBSCRIBE_RST ({{message-subscribe-rst}})           |
+| 0xC   | SUBSCRIBE_RESET ({{message-subscribe-reset}})       |
```
suhasHere, line 754: I prefer `SUBSCRIBE_CANCEL` to indicate an existing subscription is canceled now. I do understand RESET goes with QUIC naming, but Cancel seems more nature given the nomenclature of subscription vs that of a transport stream construct.

kixelated (approved):
Teeechnically, QUIC uses STOP for receiver-initiated closes. But RESET is better than RST.

draft-ietf-moq-transport.md
ianswett: I used CANCEL in the ANNOUNCE PR, and happy to consider different names, and we should make an effort to be somewhat consistent between SUBSCRIBE and ANNOUNCE to avoid confusion.

PR #387, opened 2024-02-19
Don't send Objects for failed subscriptions

### LINKED ISSUE #357: Objects before SUBSCRIBE_OK
Since they're on different streams, an object can arrive before its SUBSCRIBE_OK. This is easy to process because the Subscribe ID and Track Alias must have been the ones that the subscriber chose.

What if the publisher then sends a SUBSCRIBE_ERROR? Is this valid behavior or an error?

(If it weren't for the Expires field, SUBSCRIBE_OK would be entirely redundant and we could get rid of it.)

ianswett:
Fixes #357

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -630,6 +630,10 @@ terminates the track with a SUBSCRIBE_FIN
 (see {{message-subscribe-fin}}) or a SUBSCRIBE_RST
 (see {{message-subscribe-rst}}).
 
+Objects MUST NOT be sent for unsuccessful subscriptions, and if a subscriber
+receives a SUBSCRIBE_ERROR after receiveing objects, it MUST close the session
+with a 'Protocol Violation'.
+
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -633,6 +633,10 @@ terminates the track with a SUBSCRIBE_FIN
 (see {{message-subscribe-fin}}) or a SUBSCRIBE_RST
 (see {{message-subscribe-rst}}).
 
+Objects MUST NOT be sent for unsuccessful subscriptions, and if a subscriber
+receives a SUBSCRIBE_ERROR after receiveing objects, it MUST close the session
```
martinduke, line 637:
```suggestion
receives a SUBSCRIBE_ERROR after receiving objects, it MUST close the session
```

martinduke (approved):

PR #392, opened 2024-02-19
Add ANNOUNCE_CANCEL

### LINKED ISSUE #319: Subscriber rejecting ANNOUNCE
> Relays respond with an ANNOUNCE_OK or ANNOUNCE_ERROR control message providing the result of announcement. The entity receiving the ANNOUNCE MUST send only a single response to a given ANNOUNCE of either ANNOUNCE_OK or ANNOUNCE_ERROR.

The problem is not possible for the subscriber to reject an ANNOUNCE afterwards. This would be very useful, for example to inform a publisher that no more subscriptions will be routed to them for whatever reason. Examples:

1. The auth parameter in ANNOUNCE has expired.
2. Another publisher has sent an overriding ANNOUNCE (reconnect?)
3. The broadcast has been closed remotely, ex. some moderation tool.

Without this message, the only recourse for the subscriber is to close the connection. This won't work if multiple ANNOUNCE messages are pooled over the same connection, nor is it clean.

ianswett:
To allow subscribers to indicate that no new subscriptions will be routed to the publisher for a given namespace.

Happy to bikeshed on the name and whether we really need another frame for this.

Fixes #319

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -670,10 +670,10 @@ Relays MUST ensure that publishers are authorized by:
 Relays respond with an ANNOUNCE_OK or ANNOUNCE_ERROR control message
 providing the result of announcement. The entity receiving the
 ANNOUNCE MUST send only a single response to a given ANNOUNCE of
-either ANNOUNCE_OK or ANNOUNCE_ERROR.  When a publisher wants to stop new
-subscriptions for an announced namespace it uses UNANNOUNCE. When a
-subscriber wants to indicate it will no longer route subscriptions for a
-namespace it previously responded ANNOUNCE_OK to, it sends an
+either ANNOUNCE_OK or ANNOUNCE_ERROR.  When a publisher wants to stop
+new subscriptions for an announced namespace it sends an UNANNOUNCE.
+A subscriber indicates it will no longer route subscriptions for a
+namespace it previously responded ANNOUNCE_OK to by sending an
 ANNOUNCE_CANCEL.
 
 OBJECT message headers carry a short hop-by-hop `Track Alias` that maps to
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -670,7 +670,11 @@ Relays MUST ensure that publishers are authorized by:
 Relays respond with an ANNOUNCE_OK or ANNOUNCE_ERROR control message
 providing the result of announcement. The entity receiving the
 ANNOUNCE MUST send only a single response to a given ANNOUNCE of
-either ANNOUNCE_OK or ANNOUNCE_ERROR.
+either ANNOUNCE_OK or ANNOUNCE_ERROR.  When a publisher wants to stop new
+subscriptions for an announced namespace it uses UNANNOUNCE. When a
+subscriber wants to indicate it will no longer route subscriptions for a
+namespace it previously responded ANNOUNCE_OK to, it send an
```
ianswett, line 676:
```suggestion
namespace it previously responded ANNOUNCE_OK to, it sends an
```

draft-ietf-moq-transport.md
```diff
@@ -1522,6 +1526,25 @@ UNANNOUNCE Message {
 * Track Namespace: Identifies a track's namespace as defined in
 ({{track-name}}).
 
+## ANNOUNCE_CANCEL {#message-announce-cancel}
```
kixelated, line 1529: Bikeshed: `ANNOUNCE_RESET` so it matches `SUBSCRIBE_RESET`.

kixelated (commented):
If we're going to combine `SUBSCRIBE_ERROR` and `SUBSCRIBE_RESET`, then we should just combine `ANNOUNCE_ERROR` and `ANNOUNCE_CANCEL` now.

My desire is to have four messages:
publisher create: ANNOUNCE
publisher close: UNANNOUNCE
subscriber accept: ANNOUNCE_OK
subscriber close: ANNOUNCE_STOP

The same goes for subscriptions:
subscriber create: SUBSCRIBE
subscriber close: UNSUBSCRIBE
publisher accept: SUBSCRIBE_OK
publisher close: SUBSCRIBE_STOP

fluffy (approved):

PR #394, opened 2024-02-19
Track Namespace is an exact match

### LINKED ISSUE #252: Does relay matching behavior need to be negotiated in-band?
This came out of discussion on #204

When a relay receives an SUBSCRIBE it may need to issue its own SUBSCRIBE to a publisher that has previously ANNOUNCE'd a namespace matching that track.  The current draft doesn't specify how this match is done.  A relay could implement an exact match on Track Namespace (requires a tweak to SUBSCRIBE), or could do a longest prefix match on track name vs all ANNOUNCE'd namespaces, or possibly something else.  Is it ok to leave this up to implementations or should the draft specify this behavior more exactly?  If more than one matching type is allowed, does it need to be negotiated, and by whom (publisher, subscriber, or both)?

ianswett:
Fixes #252

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -675,6 +675,12 @@ providing the result of announcement. The entity receiving the
 ANNOUNCE MUST send only a single response to a given ANNOUNCE of
 either ANNOUNCE_OK or ANNOUNCE_ERROR.
 
+A relay manages sessions from multiple publishers and subscribers,
+connecting them based on the track namespace. This MUST use an exact
+match on track namespace unless otherwise negotiated by the application.
+For example, a SUBSCRIBE namespace=foobar message will be forwarded to
+the session that sent ANNOUNCE namespace=foobar.
+
 Relays that respond with an ANNOUNCE_OK for a 'Track Namespace'
 send subscriptions with a 'Track Namespace' that exactly match to
 that publisher.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -681,10 +681,6 @@ match on track namespace unless otherwise negotiated by the application.
 For example, a SUBSCRIBE namespace=foobar message will be forwarded to
 the session that sent ANNOUNCE namespace=foobar.
 
-Relays that respond with an ANNOUNCE_OK for a 'Track Namespace'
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -675,6 +675,10 @@ providing the result of announcement. The entity receiving the
 ANNOUNCE MUST send only a single response to a given ANNOUNCE of
 either ANNOUNCE_OK or ANNOUNCE_ERROR.
 
+Relays that respond with an ANNOUNCE_OK for a 'Track Namespace'
+send subscriptions with a 'Track Namespace' that exactly match to
```
vasilvv, line 679: "Relays send" feels ambiguous to me, as in, I don't understand what is the actual normative requirement that we are trying to communicate here.

kixelated: Same, I don't understand this.

kixelated (changes requested):

draft-ietf-moq-transport.md
ianswett: Agreed, I was having a difficult time determining what text to add and where.  Suggestions or alternate PRs are welcome.

fluffy (approved):
I'm not sure how to wordsmith this but +1 to what it is saying.

draft-ietf-moq-transport.md
kixelated: I think the text is trying to be too technical. Here's my suggestion:

A relay manages sessions from multiple publishers and subscribers, connecting them based on the track namespace. This SHOULD use a full string match unless otherwise negotiated by the application. For example, a `SUBSCRIBE namespace=foobar` message will be forwarded to the session that sent `ANNOUNCE namespace=foobar`.

I put a SHOULD instead of a MUST so we leave door open. Relays are going to rewriting the namespace, or using a portion of it based on the application, and it would be impossible to enforce anyway.

ianswett: Given you specify the unless, I'd prefer MUST unless, but otherwise LG.

ianswett:
PR updated, PTAL

kixelated (approved):

suhasHere (approved):
Thanks for the update

PR #395, opened 2024-02-19
Clarify what ROLE means

### LINKED ISSUE #151: ROLE doesn't support negotiation

### LINKED ISSUE #250: server-sent ROLE parameters

ianswett:
Add some normative text and requires both client and server to send their role.

I can also see the argument for removing this entirely, but it's in the current draft, so if we're going to keep it, I'd like to define it better.

Fixes #151, Fixes #250

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -879,7 +879,7 @@ which are of type varint:
 
 0x01:
 
-: The endpoint process subscriptions and can send objects, but not subscribe.
+: The endpoint can process subscriptions and send objects, but not subscribe.
   The endpoint MUST NOT send a SUBSCRIBE message and an ANNOUNCE MUST NOT be
   sent to it.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -885,7 +885,7 @@ which are of type varint:
 
 0x02:
 
-: The endpoing can send subscriptions and receive objects, but not publish.
+: The endpoint can send subscriptions and receive objects, but not publish.
   The endpoint MUST NOT send an ANNOUNCE message and a SUBSCRIBE MUST NOT be
   sent to it.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -881,19 +881,19 @@ The ROLE parameter (key 0x00) allows each endpoint to independently specify what
 funnctionality they support for the session. It has three possible values,
 which are of type varint:
 
-0x01:
+0x01: Publisher
 
 : The endpoint can process subscriptions and send objects, but not subscribe.
   The endpoint MUST NOT send a SUBSCRIBE message and an ANNOUNCE MUST NOT be
   sent to it.
 
-0x02:
+0x02: Subscriber
 
 : The endpoint can send subscriptions and receive objects, but not publish.
   The endpoint MUST NOT send an ANNOUNCE message and a SUBSCRIBE MUST NOT be
   sent to it.
 
-0x03:
+0x03: PubSub
 
 : The endpoint can act as a publisher or subscriber, and can send or process
   any message type.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -873,28 +873,30 @@ identified as 0xff00000D.
 
 #### ROLE parameter {#role}
 
-The ROLE parameter (key 0x00) allows the client to specify what roles it
-expects the parties to have in the MOQT connection. It has three
-possible values, which are of type varint:
+The ROLE parameter (key 0x00) allows each endpoint to independently specify what
+funnctionality they support for the session. It has three possible values,
+which are of type varint:
 
 0x01:
 
-: Only the client is expected to send objects on the connection. This is
-  commonly referred to as the ingestion case.
+: The endpoint process subscriptions and can send objects, but not subscribe.
+  The endpoint MUST NOT send a SUBSCRIBE message and an ANNOUNCE MUST NOT be
+  sent to it.
 
 0x02:
 
-: Only the server is expected to send objects on the connection. This is
-  commonly referred to as the delivery case.
+: The endpoing can send subscriptions and receive objects, but not publish.
+  The endpoint MUST NOT send an ANNOUNCE message and a SUBSCRIBE MUST NOT be
+  sent to it.
 
 0x03:
 
-: Both the client and the server are expected to send objects.
+: The endpoint can act as a publisher or subscriber, and can send or process
+  any message type.
 
-The client MUST send a ROLE parameter with one of the three values
-specified above. The server MUST close the session if the ROLE
-parameter is missing, is not one of the three above-specified values, or
-it is different from what the server expects based on the application.
+Both endpoints MUST send a ROLE parameter with one of the three values
+specified above. Both endpoints MUST close the session if the ROLE
+parameter is missing, is not one of the three above-specified values.
 
 #### PATH parameter {#path}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -896,7 +896,7 @@ which are of type varint:
 
 Both endpoints MUST send a ROLE parameter with one of the three values
 specified above. Both endpoints MUST close the session if the ROLE
-parameter is missing, is not one of the three above-specified values.
+parameter is missing or is not one of the three above-specified values.
 
 #### PATH parameter {#path}
 


draft-ietf-moq-transport.md
```diff
@@ -873,28 +873,30 @@ identified as 0xff00000D.
 
 #### ROLE parameter {#role}
 
-The ROLE parameter (key 0x00) allows the client to specify what roles it
-expects the parties to have in the MOQT connection. It has three
-possible values, which are of type varint:
+The ROLE parameter (key 0x00) allows each endpoint to independently specify what
+funnctionality they support for the session. It has three possible values,
+which are of type varint:
```
suhasHere, line 879: I still don't see a need for the ROLE parameter. I wonder what's the value add here and be fine if we don't need it.

The implementation can decipher the role based on arrival of subscribe/publish/both messages.

However, happy to hear what problem are we addressing to solve with the Role.

kixelated (approved):
Love it.

@suhasHere I definitely want ROLE in the protocol because there will be many one-sided implementations. A broadcaster like OBS won't support ROLE=subscriber, likewise a viewer like VLC won't support ROLE=publisher. This explicit flag in SETUP is a lot better than crashing if an unsupported message is sent by the peer.

draft-ietf-moq-transport.md
```diff
@@ -873,28 +873,30 @@ identified as 0xff00000D.
...
+: The endpoint can act as a publisher or subscriber, and can send or process
+  any message type.
 
-The client MUST send a ROLE parameter with one of the three values
-specified above. The server MUST close the session if the ROLE
-parameter is missing, is not one of the three above-specified values, or
-it is different from what the server expects based on the application.
+Both endpoints MUST send a ROLE parameter with one of the three values
+specified above. Both endpoints MUST close the session if the ROLE
+parameter is missing, is not one of the three above-specified values.
```
kixelated, line 899: `is missing or is not`

suhasHere:
> This explicit flag in SETUP is a lot better than crashing if an unsupported message is sent by the peer.

Having a flag in SETUP doesn't stop from someone sending unsupported messages. A faulty implementation or intentionally malicious endpoint can send unsupported messages. A receiver should be prepared to handle unsupported messages regardless.

kixelated:
> > This explicit flag in SETUP is a lot better than crashing if an unsupported message is sent by the peer.
> 
> Having a flag in SETUP doesn't stop from someone sending unsupported messages. A faulty implementation or intentionally malicious endpoint can send unsupported messages. A receiver should be prepared to handle unsupported messages regardless.

Right, an implementation is still going to crash if it receives an unsupported message, but there's a difference between signaling that upfront and yolo.

A one-sided implementation will look something like this:
```
switch (message_type) {
case SUBSCRIBE:
  // ...
case UNSUBSCRIBE:
  // ..
default:
  throw PROTOCOL_ERROR
}
```

The purpose of the ROLE parameter is to avoid accidentally triggering that default, as it will close the connection.

For example, I absolutely want to have a way to get per-viewer feedback: a `feedback` track produced by each viewer. Without this ROLE field, a viewer might cause a crash if it sends an `ANNOUNCE feedback` to the server, or the server might cause a crash if it sends a `SUBSCRIBE feedback` to the viewer. 

The only alternative to the ROLE field is to implement the entire protocol, but that's unnecessarily restrictive. Take HTTP for example, it's almost taken for granted that an implementation is for a client or a server, but rarely both. I imagine that will also be the case with MoQ with the sole exception of relays.

Even in my MoQ.js library that does support both modes, I ran into a deadlock because of ROLE=both. My publisher was not consuming ANNOUNCE messages because it was not expecting them. Explicitly signaling ROLE=publisher would have prevented this, making it a protocol violation to send half of the messages in the draft.

suhasHere:
A resilient implementation state machine will be something like

```
eceive(message) {
  switch (message.type) {
  case SUBSCRIBE:
    // ...
  case UNSUBSCRIBE:
    // ..
  default:
    // drop the message
  }
}
```
Also, Many of the conferencing examples will mostly always be send and receive by default. Our client for example does pub and subscribe always.

I personally feel the implementation constraints are resolvable and I feel still unconvinced for a ROLE parameter.

suhasHere (changes requested):
I wonder if it helps to discuss the role of role parameter during authors call for a short bit @ianswett

kixelated:
> A resilient implementation state machine will be something like
> 
> ```
> eceive(message) {
>   switch (message.type) {
>   case SUBSCRIBE:
>     // ...
>   case UNSUBSCRIBE:
>     // ..
>   default:
>     // drop the message
>   }
> }
> ```

Heh, please don't do that. This is exactly why we have version/extension negotiation; not supporting a message after negotiating it is a protocol violation.

Otherwise, my endpoint has to assume that your endpoint could ignore messages, and the MoQ draft in general, on a whim. Just tell me that you don't support ANNOUNCE via ROLE. Otherwise I would have to send an ANNOUNCE, start a timer, and guess that no OK/ERROR means no support? Probing for support is miserable.

> Since for unsupported messages, you don't want a protocol action as it will be DOS surface, for example.

You close the connection on an unsupported message.

> Also, Many of the conferencing examples will mostly always be send and receive by default. Our client for example does pub and subscribe always.
> 
> I personally feel the implementation constraints are resolvable and I feel still unconvinced for a ROLE parameter.

Just send ROLE=both then. Not every conferencing is bidirectional though, for example clubhouse-style. At Discord we also have a dedicated connection for screen sharing (don't ask) and a very explicit role=publisher vs role=subscriber flag.

kixelated:
> **Every client has to be built to be resilient to arbitrary incoming messages**. It can't break because the other side said their role was subscriber but then they send you an ANNOUCE. 

I want publisher and subscriber support to be optional. That way it's only necessary to implement half of the draft on many cases. A CDN distribution edge would be one such situation. In fact, it's a hardening measure to ensure the edge cannot receive a ANNOUNCE which might otherwise cause a vulnerability.

And generally speaking, if something is optional, then it should be negotiated. Otherwise you end up with Schroeder's extension. Silently ignoring unknown/unsupported messages is just terrible protocol design and it's s far better to close the connection with a protocol violation instead.

That's the point of the ROLE parameter. But yeah let's chat about it on Wed.

afrind:
Individual Comment:

Every implementation has to have that switch statement and some not-crash action when it receives a message it doesn't support.  Some logical options are:

1. Ignore the message
2. Send the corresponding error message (eg SUBSCRIBE_ERROR, ANNOUNCE_ERROR) with an error code like "not-implemented"
3. Close the entire connection

I think the value of ROLE is that it gives a consistent behavior early in the life of the session.  If I want to send subscribes and the peer says they are subscriber only, I close the connection.  The MUST in this PR can be improved by explaining what happens when it's violated.

suhasHere:
> If I want to send subscribes and the peer says they are subscriber only, I close the connectio

In that case the peer will close the connection since it knows that it is subscribe only and the role in the setup is supreflous

fluffy:
This is the type of things where I don't think we need it but it does not break anything, is not a big deal to implement, and does not hit performance. Given other people feel strongly this is needed, I can easily live with this. If this helps us get to consensus, I'm fine with adding it. I think the rewrite makes things clearer than they were before.

draft-ietf-moq-transport.md
```diff
@@ -877,28 +877,30 @@ identified as 0xff00000D.
...
 #### ROLE parameter {#role}
 
-The ROLE parameter (key 0x00) allows the client to specify what roles it
-expects the parties to have in the MOQT connection. It has three
-possible values, which are of type varint:
+The ROLE parameter (key 0x00) allows each endpoint to independently specify what
+funnctionality they support for the session. It has three possible values,
+which are of type varint:
 
 0x01:
```
fluffy, line 884: Small nit. I think these three different values should each have a name because it will make them easier to talk about.

ianswett:
Discussion concluded that supporting this was not that difficult and it's a clear signal of what's supported, to so AI is to fix up the PR to clarify behavior.

draft-ietf-moq-transport.md
ianswett: Agreed, I named the third type 'PubSub', but I'm open to a better name.

draft-ietf-moq-transport.md
```diff
@@ -877,28 +877,30 @@ identified as 0xff00000D.
...
+which are of type varint:
 
-0x01:
+0x01: Publisher
 
-: Only the client is expected to send objects on the connection. This is
-  commonly referred to as the ingestion case.
+: The endpoint can process subscriptions and send objects, but not subscribe.
+  The endpoint MUST NOT send a SUBSCRIBE message and an ANNOUNCE MUST NOT be
+  sent to it.
```
afrind, line 888: If we reorganized the document into "publisher messages" and "subscriber messages", then the text here (and below) could be improved.  We could say:

The endpoint MUST NOT send any subscriber messages.  If the endpoint receives any publisher messages, it MUST close the connection with "error code TBD".

kixelated: +1 I would love to see this.

kixelated: A simple way of doing it would be to add a "role" column to the table of messages and IDs.

fluffy (approved):
LGTM me. Ship it!

draft-ietf-moq-transport.md
ianswett: This SGTM, but seems like it could be a reasonably large change to the doc organization, so I'd prefer to do it in a separate PR.

PR #401, opened 2024-02-28
Merge SUBSCRIBE_RESET and SUBSCRIBE_FIN 

### LINKED ISSUE #310: Consolidating SUBSCRIBE errors
As mentioned in #309, there are 3 different ways for a publisher to close a subscriber, each with minor differences.

- `SUBSCRIBE_ERROR`: I never sent you an object.
- `SUBSCRIBE_FIN`: There are no more objects to send after group/object X/Y.
- `SUBSCRIBE_RESET`: There are more objects after group/object X/Y... but I won't send them.

A few things:

### 1. `SUBSCRIBE_ERROR` naming
I think it should be renamed since it's not the only ERROR message. Maybe `SUBSCRIBE_REJECT`?

### 2. `SUBSCRIBE_FIN` and `SUBSCRIBE_RESET` are virtually identical
The only difference is that one has an error code. Otherwise they both include a final group/object and it's up to the application/library to decide if they wait for the final object(s). I'd like to only implement this once but right now I have to copy/paste any code to handle these almost identical messages.

What if we merged them and use error code 0 to signal a FIN. The reason phrase should be removed as part of #273.

### 3. It's difficult to determine which messages are protocol violations

There really needs to be a state machine like #237 with this many ways to terminate a subscription.

I think these are all protocol violations:
- `SUBSCRIBE_OK` then `SUBSCRIBE_ERROR`
- `SUBSCRIBE_ERROR` then `SUBSCRIBE_*` 
- `SUBSCRIBE_RESET/FIN` then `SUBSCRIBE_*`

My interpretation of the draft is that a `SUBSCRIBE_FIN` followed by a `SUBSCRIBE_RESET` is invalid, which is not obvious and unlike QUIC.
> A subscription remains active until it expires, until the publisher of the track terminates the track with a SUBSCRIBE_FIN (see [Section 6.8](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#message-subscribe-fin)) or a SUBSCRIBE_RST (see [Section 6.9](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#message-subscribe-rst)).

### 4. Do these error distinctions actually matter to the transport?

I don't think so. The library/relay behavior doesn't change based on `SUBSCRIBE_RESET/FIN`. The only thing that miiiight differentiate between the two is the application, which can use errors codes for this purpose. At the very least I would merge these two messages.

But I would also like to explore a single way for the publisher to close a subscription. The application can signal the nuanced differences between errors (ex. error code 404 vs 410). The library/relay doesn't care, it just forwards the error code.

### LINKED ISSUE #356: SUBSCRIBE_FIN/RST for old subscriptions
Sections 6.8 and 6.9 strongly imply that SUBSCRIBE_FIN/RST are sent at the true end of the track. So are they not sent at the conclusion of a SUBSCRIBE that does not cover the end of the track?

For example, if the last object of a track is (10, 12), and there's one subscribe for (4,0) to (5, 100)

Is there a SUBSCRIBE_FIN?

If not, how does the subscriber know that in fact (5, 60) is the last object in that subscription?

If so, does Final Group and Final Object refer to the last object in the subscription?

### LINKED ISSUE #398: Current naming is hard to keep track of <BIKESHED>
Propose names from my understanding of discussion call 

Announce (send by publisher)  , Annouce_OK, Annouce_Error
UnAnnounce (send by publisher)  UnAnnounce_OK, UnAnnounce_Error 
Annouce_End ( sent by thing that received Annouce) , AnnouceEnd_OK, AnnouceEnd_OK

Subscribe (send by subscriber)  , Subscribe_OK,  Subscribe_Error
UnSubscribe (send by subscriber)  , Subscribe_OK,  Subscribe_Error
Subscribe_End (send by thing that received subscriber), Subscribe_OK,  Subscribe_Error

Mostly trying to get rid of FIN / RST and have different names for the messages that flow in opposite directions.

ianswett:
Into SUBSCRIBE_DONE.

I didn't remove the SUBSCRIBE_ERROR message in favor of SUBSCRIBE_CLOSED as @kixelated suggested in this PR.

Fixes #398 
Fixes #356
Addresses some comments in #310, but possibly not all

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -629,8 +629,8 @@ subscribers for each track. Each new OBJECT belonging to the
 track within the subscription range is forwarded to each active
 subscriber, dependent on the congestion response. A subscription
 remains active until it expires, until the publisher of the track
-terminates the track with a SUBSCRIBE_CLOSED
-(see {{message-subscribe-closed}}).
+terminates the track with a SUBSCRIBE_DONE
+(see {{message-subscribe-done}}).
 
 Objects MUST NOT be sent for unsuccessful subscriptions, and if a subscriber
 receives a SUBSCRIBE_ERROR after receiving objects, it MUST close the session
@@ -659,7 +659,7 @@ as defined below:
 |------|---------------------------|
 
 The applicaiton SHOULD use a relevant status code in
-SUBSCRIBE_CLOSE, as defined below:
+SUBSCRIBE_DONE, as defined below:
 
 |------|---------------------------|
 | Code | Reason                    |
@@ -781,7 +781,7 @@ MOQT Message {
 |-------|-----------------------------------------------------|
 | 0xA   | UNSUBSCRIBE ({{message-unsubscribe}})               |
 |-------|-----------------------------------------------------|
-| 0xB   | SUBSCRIBE_CLOSED ({{message-subscribe-closed}})     |
+| 0xB   | SUBSCRIBE_DONE ({{message-subscribe-done}})         |
 |-------|-----------------------------------------------------|
 | 0x10  | GOAWAY ({{message-goaway}})                         |
 |-------|-----------------------------------------------------|
@@ -1408,7 +1408,7 @@ SUBSCRIBE_ERROR
 A subscriber issues a `UNSUBSCRIBE` message to a publisher indicating it is no
 longer interested in receiving media for the specified track and Objects
 should stop being sent as soon as possible.  The publisher sends a
-SUBSCRIBE_CLOSED to acknowledge the unsubscribe was successful and indicate
+SUBSCRIBE_DONE to acknowledge the unsubscribe was successful and indicate
 the final Object.
 
 The format of `UNSUBSCRIBE` is as follows:
@@ -1422,16 +1422,16 @@ UNSUBSCRIBE Message {
 
 * Subscribe ID: Subscription Identifer as defined in {{message-subscribe-req}}.
 
-## SUBSCRIBE_CLOSED {#message-subscribe-closed}
+## SUBSCRIBE_DONE {#message-subscribe-done}
 
-A publisher issues a `SUBSCRIBE_CLOSED` message to indicate it
+A publisher issues a `SUBSCRIBE_DONE` message to indicate it
 is done publishing Objects for that subscription.  The Status Code indicates why
 the subscription ended, and whether it was an error.
 
-The format of `SUBSCRIBE_CLOSED` is as follows:
+The format of `SUBSCRIBE_DONE` is as follows:
 
 ~~~
-SUBSCRIBE_CLOSED Message {
+SUBSCRIBE_DONE Message {
   Subscribe ID (i),
   Status Code (i),
   Reason Phrase (b),
@@ -1440,7 +1440,7 @@ SUBSCRIBE_CLOSED Message {
   [Final Object (i)],
 }
 ~~~
-{: #moq-transport-subscribe-fin-format title="MOQT SUBSCRIBE_CLOSED Message"}
+{: #moq-transport-subscribe-fin-format title="MOQT SUBSCRIBE_DONE Message"}
 
 * Subscribe ID: Subscription identifier as defined in {{message-subscribe-req}}.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1436,7 +1436,6 @@ message in this track.
 * Final Object: The largest Object ID sent by the publisher in an OBJECT
 message in the `Final Group` for this track.
 
-
 ## ANNOUNCE {#message-announce}
 
 The publisher sends the ANNOUNCE control message to advertise where the
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -490,10 +490,10 @@ SHOULD terminate the session with 'GOAWAY Timeout' after a sufficient timeout if
 there are still open subscriptions on a connection.
 
 The GOAWAY message does not immediately impact subscription state. A subscriber
-SHOULD individually UNSUBSCRIBE for each existing subscription, while a
-publisher MAY reject new SUBSCRIBEs while in the draining state. When the server
-is a subscriber, it SHOULD send a GOAWAY message prior to any UNSUBSCRIBE
-messages.
+SHOULD individually unsubscribe for each existing subscription using
+SUBSCRIBE_CANCEL.  A publisher MAY reject new SUBSCRIBEs while in the draining
+state. When the server is a subscriber, it SHOULD send a GOAWAY message prior to
+any SUBSCRIBE_CANCEL messages.
 
 After the client receives a GOAWAY, it's RECOMMENDED that the client waits until
 there are no more active subscriptions before closing the session with NO_ERROR.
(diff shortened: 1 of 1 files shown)


kixelated (approved):
I like it.

I would also like to merge ERROR with CLOSED, but we'll chat about that.

draft-ietf-moq-transport.md
```diff
@@ -1448,11 +1418,11 @@ SUBSCRIBE_RESET Message {
...
 }
 ~~~
-{: #moq-transport-subscribe-reset format title="MOQT SUBSCRIBE RESET Message"}
+{: #moq-transport-subscribe-fin-format title="MOQT SUBSCRIBE_CLOSED Message"}
 
-* Subscribe ID: Subscription Identifier as defined in {{message-subscribe-req}}.
+* Subscribe ID: Subscription identifier as defined in {{message-subscribe-req}}.
 
-* Error Code: Identifies an integer error code for subscription failure.
+* Error Code: Identifies an integer error code for subscription end.
```
martinduke, line 1425: So what are these error codes? It's important to identify what they are, and I don't think they're like the other error codes (because some of them are not really "errors" in the classic sense.

In fact, I'd prefer if we called them a "status code". As I commented in #356, I think there are numerous states the sender might wish to communicate:
1) You got all of the objects in the range;
2) You got almost all of the objects, but some were dropped for latency/bandwidth reasons
3) The track was truncated for some reason
and possibly, a relay might say:
4) You got everything I have but i'm not sure if that's everything

martinduke (approved):

draft-ietf-moq-transport.md
```diff
@@ -1387,59 +1384,32 @@ SUBSCRIBE_ERROR
   the receiver MUST close the connection with a Duplicate Track Alias error
   ({{session-termination}}).
 
-## UNSUBSCRIBE {#message-unsubscribe}
+## SUBSCRIBE_CANCEL {#message-subscribe-cancel}
```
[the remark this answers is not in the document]
wilaw: I agree with Martin on this. UNSUBSCRIBE is clearly the semantic negation of SUBSCRIBE, while SUBSCRIBE_CANCEL leaves me with no intuitive understanding as to what is happening.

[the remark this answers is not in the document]
suhasHere: I am not pushing against the PR ,but slight preference on the names. But I can live with it 

My preference would be

Subscribe --> Unsubscribe (both sent by the subscriber)
SubscribeCancel -> Sent by the publisher after sending an OK to subscribe

fluffy:
So my notes from this discussion. CLOSE is confusing with QUIC layer.  People seemed to prefer UNSUBSCRIBE over SUBSCRIBE_CLOSE. I love the move aware from RST and FIN as naming.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
ianswett: I renamed these back to UNSUBSCRIBE and UNANNOUNCE.

draft-ietf-moq-transport.md
```diff
@@ -1390,7 +1406,10 @@ SUBSCRIBE_ERROR
 ## UNSUBSCRIBE {#message-unsubscribe}
 
 A subscriber issues a `UNSUBSCRIBE` message to a publisher indicating it is no
-longer interested in receiving media for the specified track.
+longer interested in receiving media for the specified track and Objects
+should stop being sent as soon as possible.  The publisher sends a
+SUBSCRIBE_CLOSED to acknowledge the unsubscribe was successful and indicate
+the final Object.
```
suhasHere, line 1412: I don't think we need a response for Unsubscribe . The data path should carry the information on the final group/object.

draft-ietf-moq-transport.md
```diff
@@ -1399,60 +1418,33 @@ UNSUBSCRIBE Message {
   Subscribe ID (i)
 }
 ~~~
-{: #moq-transport-unsubscribe-format title="MOQT UNSUBSCRIBE Message"}
+{: #moq-transport-subscribe-cancel-format title="MOQT UNSUBSCRIBE Message"}
```
suhasHere, line 1421: I think this should be renamed to unsubscribe.

suhasHere (changes requested):
Thanks Ian .. Overall it looks fine except for using SubscribeClosed as a response for Unsubsubscrib.
 
Also another nit, I preferred SubscribeEnded better than closed since it seems more easy to think when you read the name

draft-ietf-moq-transport.md
```diff
@@ -1399,60 +1418,33 @@ UNSUBSCRIBE Message {
...
+is done publishing Objects for that subscription.  The Status Code indicates why
+the subscription ended, and whether it was an error.
 
-The format of `SUBSCRIBE_FIN` is as follows:
+The format of `SUBSCRIBE_CLOSED` is as follows:
 
 ~~~
-SUBSCRIBE_FIN Message {
+SUBSCRIBE_CLOSED Message {
   Subscribe ID (i),
```
suhasHere, line 1435: s/SUBSCRIBE_CLOSED/SUBSCRIBE_END to have naming not to conflate with underlying transport usages

draft-ietf-moq-transport.md
ianswett: Putting it in the data path makes sense if it's going to be cached and it's "Part of the track", but for state that's specific to a subscription, like the last Object sent, I think its makes more sense for it to be in subscription specific messages.

draft-ietf-moq-transport.md
ianswett, line 1421:
```suggestion
{: #moq-transport-unsubscribe-format title="MOQT UNSUBSCRIBE Message"}
```

ianswett: Good catch, fixed.

draft-ietf-moq-transport.md
```diff
@@ -659,6 +658,25 @@ as defined below:
...
+|-----:|:--------------------------|
+| 0x0  | Unsubscribed              |
+|------|---------------------------|
+| 0x1  | Internal Error            |
+|------|---------------------------|
+| 0x2  | Unauthorized              |
+|------|---------------------------|
+| 0x3  | Track Ended               |
+|------|---------------------------|
+| 0x4  | Subscription Ended        |
```
fluffy, line 675: I don't know what causes this one on a relay.

draft-ietf-moq-transport.md
```diff
@@ -659,6 +658,25 @@ as defined below:
...
+|------|---------------------------|
+| Code | Reason                    |
+|-----:|:--------------------------|
+| 0x0  | Unsubscribed              |
+|------|---------------------------|
+| 0x1  | Internal Error            |
+|------|---------------------------|
+| 0x2  | Unauthorized              |
+|------|---------------------------|
+| 0x3  | Track Ended               |
```
fluffy, line 673: Just becasue a track has ended does not mean all the data has been delivered. I don't know how a relay would know when it could send this. I think it is better for the subscriber to just unsubscribe once it has the data. To put this more specifically, I don't know how a relay would be able to know when to send this.

fluffy (changes requested):

draft-ietf-moq-transport.md
afrind: > I don't know how a relay would be able to know when to send this.

A relay could send it when it receives a SUBSCRIBE_CLOSE from the publisher with code=Track Ended?

draft-ietf-moq-transport.md
ianswett: You delivered all the requested data up to the end Object and Group.

draft-ietf-moq-transport.md
ianswett: I have mixed feelings about this one, so I'd be happy to remove it if we think the track end should be explicit in the Object model.  If so, we could just use Subscription Ended when the track ends.

PR #405, opened 2024-03-04
bit align ContentExists. Fixes #404

### LINKED ISSUE #404: ContentExists makes messages hard to parse
#343 and #389 introduced a `ContentExists` field to SUBSCRIBE_OK, SUBSCRIBE_FIN, and SUBSCRIBE_RESET. If I read the notation correctly, it has a size of 1 bit, which makes it hard to parse the rest of the message. Could we encode that information in the message type by using two types per message instead of a 1-bit field?

fluffy:


fluffy:
@ianswett - I updated, have another read and see if it looks OK. Thanks

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -243,6 +243,15 @@ x (b):
   described in ({{?RFC9000, Section 16}}), followed by that many bytes
   of binary data
 
+x (f):
```
ianswett, line 246: f makes me think float.  Admittedly, we don't use floats, so maybe this is fine.

ianswett (approved):

draft-ietf-moq-transport.md
fluffy: As I wrote this, I had the same concern, and failed to come up with better idea but, uh, yah, not thrilled about the conflict with float.

ianswett: I thought about this for a while and couldn't come up with anything better, so I'm fine with f for flag.

ianswett (approved):

Issue #379, opened 2024-02-14
Single Message ID space is unnecessary given messages have context


kixelated:
Currently, there's one message section and number space. It's confusing to read and it's not clear what messages can be encoded on a control stream, data stream, or datagram.

### Proposal 1: Annotate Table
| id | name | context |
|-|-|-|
| 0x0 | OBJECT_STREAM | data |
| 0x1 | OBJECT_DATAGRAM | datagram |
| 0x3 | SUBSCRIBE | control |
| 0x4 | SUBSCRIBE_OK | control |
| 0x5 | SUBSCRIBE_ERROR | control |
| 0x6 | ANNOUNCE | control |
| 0x7 | ANNOUNCE_OK | control |
| 0x8 | ANNOUNCE_ERROR | control |
| 0x9 | UNANNOUNCE | control |
| 0xa | UNSUBSCRIBE | control |
| 0xb | SUBSCRIBE_FIN | control |
| 0xc| SUBSCRIBE_RST | control |
| 0x10 | GOAWAY | control |
| 0x40 | CLIENT_SETUP | control |
| 0x41 | SERVER_SETUP | control |
| 0x50 | STREAM_HEADER_TRACK | data |
| 0x51 | STREAM_HEADER_GROUP | data |

### Proposal 2: Split Numberspace
**control messages (bidirectional streams)**
| id | name |
|-|-|
| 0x0 | CLIENT_SETUP |
| 0x1 | SERVER_SETUP |
| 0x3 | SUBSCRIBE |
| 0x4 | SUBSCRIBE_OK |
| 0x5 | SUBSCRIBE_ERROR |
| 0x6 | ANNOUNCE |
| 0x7 | ANNOUNCE_OK |
| 0x8 | ANNOUNCE_ERROR |
| 0x9 | UNANNOUNCE |
| 0xa | UNSUBSCRIBE |
| 0xb | SUBSCRIBE_FIN |
| 0xc| SUBSCRIBE_RST |
| 0x10 | GOAWAY |

**data messages (unidirectional streams)**
| id | name |
|-|-|
| 0x0 | OBJECT_STREAM |
| 0x1 | OBJECT_TRACK |
| 0x2 | OBJECT_GROUP |

**datagram messages:**
| id | name |
|-|-|
| 0x0 | OBJECT_DATAGRAM |

suhasHere:
[...]

PR #407, opened 2024-03-04
Split messages by stream type.

kixelated:
- Control messages
- Data messages
- Datagram messages

Tried to keep things completely editorial. I would like to do more cleanup but I would absolutely get carried away.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -404,31 +404,25 @@ If a given parameter carries the same information in multiple versions,
 but might have different optimal values in those versions, there SHOULD be
 separate Setup parameters for that information in each version.
 
-## Session initialization {#session-init}
+## Control Stream
+The first stream opened is a client-initiated bidirectional stream where
+the peers exchange Setup Messages ({{message-setup}}).
+All control messages ({{message-control}}) are sent over this control stream.
 
-The first stream opened is a client-initiated bidirectional control stream where
-the peers exchange Setup messages ({{message-setup}}).  All messages defined in
-this draft except OBJECT and OBJECT_WITH_LENGTH are sent on the control stream
-after the Setup message. Control messages MUST NOT be sent on any other stream,
-and a peer receiving a control message on a different stream closes the session
-as a 'Protocol Violation'. Objects MUST NOT be sent on the control stream, and a
-peer receiving an Object on the control stream closes the session as a 'Protocol
-Violation'.
-
-This draft only specifies a single use of bidirectional streams. Objects are
-sent on unidirectional streams.  Because there are no other uses of
-bidirectional streams, a peer MAY currently close the session as a
-'Protocol Violation' if it receives a second bidirectional stream.
+This draft only specifies a single use of bidirectional streams.
+Because there are no other uses of bidirectional streams, a peer MAY currently close the session as a 'Protocol Violation' if it receives a second bidirectional stream.
 
 The control stream MUST NOT be abruptly closed at the underlying transport
-layer.  Doing so results in the session being closed as a 'Protocol Violation'.
+layer. Doing so results in the session being closed as a 'Protocol Violation'.
 
-## Stream Cancellation
+## Data Streams
+A publisher will create unidirectional streams containing one or more data messages ({{message-data}}).
 
-Streams aside from the control stream MAY be canceled due to congestion
-or other reasons by either the sender or receiver. Early termination of a
-stream does not affect the MoQ application state, and therefore has no
-effect on outstanding subscriptions.
+Data streams MAY be canceled due to congestion or other reasons by either the sender or receiver. Early termination of a stream does not affect the MoQ application state, and therefore has no effect on outstanding subscriptions.
+
+## Datagrams
+A publisher may transmit datagrams containing a datagram messages ({{message-datagram}}).
+This is done to avoid retransmissions and is useful for transient objects.
 
 ## Termination  {#session-termination}
 
@@ -500,7 +494,7 @@ there are no more active subscriptions before closing the session with NO_ERROR.
 Ideally this is transparent to the application using MOQT, which involves
 establishing a new session in the background and migrating active subscriptions
 and announcements. The client can choose to delay closing the session if it
-expects more OBJECTs to be delivered. The server closes the session with a
+expects more objects to be delivered. The server closes the session with a
 'GOAWAY Timeout' if the client doesn't close the session quickly enough.
 
 
@@ -625,7 +619,7 @@ The entity receiving the SUBSCRIBE MUST send only a single response to
 a given SUBSCRIBE of either SUBSCRIBE_OK or SUBSCRIBE_ERROR.
 
 For successful subscriptions, the publisher maintains a list of
-subscribers for each track. Each new OBJECT belonging to the
+subscribers for each track. Each new object belonging to the
 track within the subscription range is forwarded to each active
 subscriber, dependent on the congestion response. A subscription
 remains active until it expires, until the publisher of the track
@@ -707,14 +701,14 @@ match on track namespace unless otherwise negotiated by the application.
 For example, a SUBSCRIBE namespace=foobar message will be forwarded to
 the session that sent ANNOUNCE namespace=foobar.
 
-OBJECT message headers carry a short hop-by-hop `Track Alias` that maps to
+Data messages {{message-data}} carry a short hop-by-hop `Track Alias` that maps to
 the Full Track Name (see {{message-subscribe-ok}}). Relays use the
-`Track Alias` of an incoming OBJECT message to identify its track and find
-the active subscribers for that track. Relays MUST NOT depend on OBJECT
-payload content for making forwarding decisions and MUST only depend on the
+`Track Alias` of an incoming message to identify its track and find
+the active subscribers for that track. Relays MUST NOT depend on
+payload contents for making forwarding decisions and MUST only depend on the
 fields, such as priority order and other metadata properties in the
-OBJECT message header. Unless determined by congestion response, Relays
-MUST forward the OBJECT message to the matching subscribers.
+header. Unless determined by congestion response, Relays
+MUST forward data messages {{message-data}} to the matching subscribers.
 
 ## Congestion Response at Relays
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -404,31 +404,25 @@ If a given parameter carries the same information in multiple versions,
...
+
+This draft only specifies a single use of bidirectional streams.
+Because there are no other uses of bidirectional streams, a peer MAY currently close the session as a 'Protocol Violation' if it receives a second bidirectional stream.
 
 The control stream MUST NOT be abruptly closed at the underlying transport
-layer.  Doing so results in the session being closed as a 'Protocol Violation'.
+layer. Doing so results in the session being closed as a 'Protocol Violation'.
+
+## Data Streams
+A publisher will create unidirectional streams containing one or more data messages ({{message-data}}).
```
ianswett, line 419:
```suggestion
A publisher creates unidirectional streams containing one or more data messages ({{message-data}}).
```

draft-ietf-moq-transport.md
```diff
@@ -404,31 +404,25 @@ If a given parameter carries the same information in multiple versions,
...
-## Stream Cancellation
+Data streams MAY be canceled due to congestion or other reasons by either the sender or receiver. Early termination of a stream does not affect the MoQ application state, and therefore has no effect on outstanding subscriptions.
 
-Streams aside from the control stream MAY be canceled due to congestion
-or other reasons by either the sender or receiver. Early termination of a
-stream does not affect the MoQ application state, and therefore has no
-effect on outstanding subscriptions.
+## Datagrams
+A publisher may transmit datagrams containing a datagram messages ({{message-datagram}}).
+This is done to avoid retransmissions and is useful for transient objects.
```
ianswett, line 425:
```suggestion
This avoids retransmissions, minimizes framing and flow control overhead, and is useful for transient objects.
```

draft-ietf-moq-transport.md
```diff
@@ -745,26 +738,33 @@ limits.  See section 2.2 in {{QUIC}}.
...
+- Data Messages ({{message-data}}): unidirectional streams
+- Datagram Messages ({{message-datagram}}): datagrams
+
+Each category has its own registry of Message ID which may overlap.
+For example, the meaning of Message ID = 3 changes depend on the stream type.
+
+An endpoint that receives an unknown message type MUST close the session with a 'Protocol Violation'.
+
+## Control Messages {#message-control}
+Control messages are transmitted over bidirectional streams.
```
ianswett, line 763:
```suggestion
Control messages are transmitted over the bidirectional stream used to establish the session.
```

draft-ietf-moq-transport.md
```diff
@@ -1575,6 +1329,267 @@ GOAWAY Message {
   compatibility.
 
 
+## Data Messages {#message-data}
```
ianswett, line 1332: The distinction between data messages and those sent over datagrams is a bit odd to me, since it's the same type of data.

ianswett (commented):
This is going in the right direction, but it creates a large diff, so it would be useful to have more reviewers.

draft-ietf-moq-transport.md
```diff
@@ -404,31 +404,25 @@ If a given parameter carries the same information in multiple versions,
...
 
 The control stream MUST NOT be abruptly closed at the underlying transport
-layer.  Doing so results in the session being closed as a 'Protocol Violation'.
+layer. Doing so results in the session being closed as a 'Protocol Violation'.
+
+## Data Streams
+A publisher will create unidirectional streams containing one or more data messages ({{message-data}}).
 
-## Stream Cancellation
+Data streams MAY be canceled due to congestion or other reasons by either the sender or receiver. Early termination of a stream does not affect the MoQ application state, and therefore has no effect on outstanding subscriptions.
```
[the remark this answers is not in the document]
ianswett: I think this text is largely existing.

draft-ietf-moq-transport.md
```diff
@@ -745,26 +738,33 @@ limits.  See section 2.2 in {{QUIC}}.
...
-Both unidirectional and bidirectional streams contain sequences of
-length-delimited messages.
-
-An endpoint that receives an unknown message type MUST close the session.
+MoQT endpoints exchange sequentially encoded messages over QUIC streams and datagrams.
 
 ~~~
 MOQT Message {
-  Message Type (i),
+  Message ID (i),
```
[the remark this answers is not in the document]
ianswett: Good point, I'd prefer to stick with Type.  ID does imply something that increments, like a Subscribe ID or Stream ID to me.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
kixelated: Yeah this is copy-paste.

draft-ietf-moq-transport.md
```diff
@@ -404,31 +404,25 @@ If a given parameter carries the same information in multiple versions,
...
-sent on unidirectional streams.  Because there are no other uses of
-bidirectional streams, a peer MAY currently close the session as a
-'Protocol Violation' if it receives a second bidirectional stream.
+## Control Stream
+The first stream opened is a client-initiated bidirectional stream where
+the peers exchange Setup Messages ({{message-setup}}).
+All control messages ({{message-control}}) are sent over this control stream.
+
+This draft only specifies a single use of bidirectional streams.
+Because there are no other uses of bidirectional streams, a peer MAY currently close the session as a 'Protocol Violation' if it receives a second bidirectional stream.
```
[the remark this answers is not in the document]
kixelated: We should definitely do a pass and create more specific error codes. "Protocol Violation" is really a "TODO".

In this instance, I think we'll remove this error case eventually. I'd like to support multiple control streams, ideally one per session/broadcast, to avoid head-of-line blocking between unrelated tracks.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
kixelated: Agreed, the table below used `ID` so I wanted to be consistent. I can change the table instead.

draft-ietf-moq-transport.md
```diff
@@ -745,26 +738,33 @@ limits.  See section 2.2 in {{QUIC}}.
...
 ~~~
 {: #moq-transport-message-format title="MOQT Message"}
 
+There are three categories of messages, identified by the underlying transport type.
+
+- Control Messages ({{message-control}}): bidirectional streams
+- Data Messages ({{message-data}}): unidirectional streams
+- Datagram Messages ({{message-datagram}}): datagrams
+
+Each category has its own registry of Message ID which may overlap.
```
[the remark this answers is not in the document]
kixelated: No, everything is wire identical to today and we're not using extra bytes.

I'm trying to specify proposal 2 in #379 but without reflowing the types yet so this PR is editorial only. I'm just trying to say that the meaning of message type = 3 depends on if the message was received over a bidirectional or unidirectional stream (or datagram).

The benefit is that we can reuse IDs (saving bits) and avoid constantly checking the context of a message. It's no longer possible to receive a SUBSCRIBE message over a data stream (unidirectional) or datagram because that type is only registered for control streams (bidirectional).

draft-ietf-moq-transport.md
```diff
@@ -1575,6 +1329,267 @@ GOAWAY Message {
...
+  Subscribe ID (i),
+  Track Alias (i),
+  Group ID (i),
+  Object ID (i),
+  Object Send Order (i),
+  Object Payload (..),
+}
+~~~
+{: #object-datagram-format title="MOQT OBJECT_DATAGRAM Message"}
+
```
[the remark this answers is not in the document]
kixelated: I can make it easier to review by splitting it into two commits. The first one changes the order of messages while the second one reflows some text, specifically about data messages.

draft-ietf-moq-transport.md
kixelated: It's just to specify that only a subset of messages are allowed in datagram contexts. There might be a better way of doing that though.

Also I thought the plan is to add a length to OBJECT_STREAM, but we definitely shouldn't do that for OBJECT_DATAGRAM. Maybe we also add datagram-only messages in the future and it's just more future-compatible to have a separate object type registry for datagrams.

draft-ietf-moq-transport.md
kixelated: I don't want to upsell datagrams. They save 2-4 bytes (just the stream ID) but that's not the reason why you should use them.

I'll write another blurb that focuses on the timeliness angle.

suhasHere: +1 on Ian's suggestions.

draft-ietf-moq-transport.md
```diff
@@ -404,31 +404,27 @@ If a given parameter carries the same information in multiple versions,
...
+Each data stream contains one or more data messages ({{message-data}}).
+Each stream MAY be canceled due to congestion or other reasons by either the sender or receiver. Early termination of a stream does not affect the MoQ application state, and therefore has no effect on outstanding subscriptions.
 
-Streams aside from the control stream MAY be canceled due to congestion
-or other reasons by either the sender or receiver. Early termination of a
-stream does not affect the MoQ application state, and therefore has no
-effect on outstanding subscriptions.
+Each datagram contains a single datagram message ({{message-datagram}}).
+Datagrams are transient, intended for objects with a tight deadline that do not need retransmissions or flow control.
+A datagram MAY be dropped for any reason, such as congestion, and SHOULD NOT be queued or cached.
```
suhasHere, line 427: I don't think queuing or caching statement is appropriate here. Its relay implementation detail and they are free to cache objects or not reagrdless of where they come from (stream/datagram)

kixelated: > I don't think queuing or caching statement is appropriate here. Its relay implementation detail and they are free to cache objects or not reagrdless of where they come from (stream/datagram)

Sure, I didn't realize this statement would be controversial and we can discuss in #408 instead.

draft-ietf-moq-transport.md
ianswett: I agree the byte savings is small, but it's also easy to understand.  I think flow control is the most compelling reason to use datagrams instead of object per stream.  If a connection only allows 100 streams at once, it's extremely easy for stream limits to be a constraint.

One doesn't have to retransmit stream data, but I agree that is the assumed default behavior.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
ianswett: That would be very helpful.  I think this is a good change, but I'm hesitant to merge such a large single change because it is difficult to know exactly what changed.

ianswett (commented):
I think this is good, but splitting it into two PRs to make it easier to see the changes would be very helpful.

kixelated:
Yeah I'll split it into two PRs.

kixelated:
I'll make another pass after splitting this PR into two.

PR #413, opened 2024-03-11
Fix the bit encoding of ContentExists

martinduke:
This is a one-byte field, not a one-bit field


vasilvv:
This is already being fixed in #405

ianswett:
This issue was fixed by #405

PR #414, opened 2024-03-13
Add a message type for ANNOUNCE_CANCEL

### LINKED ISSUE #409: ANNOUNCE_CANCEL does not have a message type
ANNOUNCE_CANCEL is missing in [Table 4](https://datatracker.ietf.org/doc/html/draft-ietf-moq-transport-03#table-4).

kixelated:
Fixes #409


ianswett (approved):
Good catch, thanks for fixing this.

PR #417, opened 2024-03-18
Clarify Send Order

ianswett:
Merges the two send order approaches because it appears they're very similar from the perspective of MoQ transport and removes text that is not relevant to MoQ Transport.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -525,25 +525,18 @@ This section is expected to cover details on:
 - Considerations for merging multiple streams across domains onto single
   connection and interactions with specific prioritization schemes.
 
-## Object Send Order {#send-order}
+## Object Priority {#object-priority}
 
-Goals of Send Order include:
+Goals of Object Priority include:
 * Enable objects to be delivered in the order intended by the publisher
 * Allow nodes and relays to delay or skip some objects under congestion
 * Create more reliable behavior of relays
 
-The publisher's send order SHOULD be followed when possible. The contents
-of each object are still delivered in order.
+The publisher's Object Priority SHOULD be followed when possible. The
+contents of each object are still delivered in order.
 
-A receiver MUST NOT assume that objects will be received in send order,
-for reasons including:
-* Newly encoded objects can have a smaller send order than outstanding
-  objects.
-* Packet loss or flow control can delay sending individual streams.
-* The sender might not support stream prioritization.
-
-When the publisher chooses which object to send next via `send_order`, they
-apply the following rules:
+When the publisher chooses which object to send next, they apply the
+following rules:
 
 * Within the same group, objects with a lower priority number (e.g. 1)
   are always sent before objects with a numerically greater priority
@@ -922,8 +915,8 @@ A canonical MoQ Object has the following information:
 IDs starts at 0, increasing sequentially for each object within the
 group.
 
-* Object Send Order: An integer indicating the object send order
-{{send-order}}.
+* Object Priority: An integer indicating the publisher's preferred
+transmission order {{object-priority}}.
 
 * Object Forwarding Preference: An enumeration indicating how a sender sends an
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
@@ -958,7 +951,7 @@ OBJECT_STREAM Message {
   Track Alias (i),
   Group ID (i),
   Object ID (i),
-  Object Send Order (i),
+  Object Priority (i),
   Object Payload (..),
 }
 ~~~
@@ -994,7 +987,7 @@ OBJECT_DATAGRAM Message {
   Track Alias (i),
   Group ID (i),
   Object ID (i),
-  Object Send Order (i),
+  Object Priority (i),
   Object Payload (..),
 }
 ~~~
@@ -1017,7 +1010,7 @@ TODO: figure out how a relay closes these streams
 
 When a stream begins with `STREAM_HEADER_TRACK`, all objects on the stream
 belong to the track requested in the Subscribe message identified by `Subscribe
-ID`.  All objects on the stream have the `Object Send Order` specified in the
+ID`.  All objects on the stream have the `Object Priority` specified in the
 stream header.
 
 
@@ -1025,7 +1018,7 @@ stream header.
 STREAM_HEADER_TRACK Message {
   Subscribe ID (i)
   Track Alias (i),
-  Object Send Order (i),
+  Object Priority (i),
 }
 ~~~
 {: #stream-header-track-format title="MOQT STREAM_HEADER_TRACK Message"}
@@ -1056,14 +1049,14 @@ equal to a previously sent Object ID within a given group on that stream.
 When a stream begins with `STREAM_HEADER_GROUP`, all objects on the stream
 belong to the track requested in the Subscribe message identified by `Subscribe
 ID` and the group indicated by `Group ID`.  All objects on the stream
-have the `Object Send Order` specified in the stream header.
+have the `Object Priority` specified in the stream header.
 
 ~~~
 STREAM_HEADER_GROUP Message {
   Subscribe ID (i),
   Track Alias (i),
   Group ID (i)
-  Object Send Order (i)
+  Object Priority (i)
 }
 ~~~
 {: #stream-header-group-format title="MOQT STREAM_HEADER_GROUP Message"}
@@ -1073,7 +1066,7 @@ All Objects received on a stream opened with `STREAM_HEADER_GROUP` have an
 
 To send an Object with `Object Forwarding Preference` = `Group`, find the open
 stream that is associated with the subscription, `Group ID` and `Object
-Send Order`, or open a new one and send the `STREAM_HEADER_GROUP` if needed,
+Priority`, or open a new one and send the `STREAM_HEADER_GROUP` if needed,
 then serialize the following fields.
 
 ~~~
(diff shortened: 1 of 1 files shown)


fluffy:
On my todo list to review but need to think about carefully on this one.

draft-ietf-moq-transport.md
```diff
@@ -972,7 +923,7 @@ IDs starts at 0, increasing sequentially for each object within the
 group.
 
 * Object Send Order: An integer indicating the object send order
-{{send-order}} or priority {{ordering-by-priorities}} value.
+{{send-order}}.
```
fluffy, line 926: So the algorithm above refers to "priority". I agree we only need one integer but need to align names.

draft-ietf-moq-transport.md
```diff
@@ -522,91 +522,42 @@ This section is expected to cover details on:
...
-
-A sender MUST send each object over a dedicated stream.  The library
-should support prioritization ({{priority-congestion}}) such that
-streams are transmitted in send order.
+## Object Send Order {#send-order}
 
-A receiver MUST NOT assume that objects will be received in send order,
-for the following reasons:
+Goals of Send Order include:
+* Enable objects to be delivered in the order intended by the publisher
```
fluffy, line 531: This sounds right for subscribe, I think it would be reasonable for Fetch to add an preference to deliver in order, or in reverse order.

fluffy (approved):
Few small comments on the text.  

At some level, I don't think this is enough. It does not deal with deciding what track to send next when there are objects from multiple tracks waiting to be sent to the same client.  But that said, I think we should merge this, start using it, then fix it and extend it as we learn more.

suhasHere:
+1 on the simplification and keep it focussed.

suhasHere (approved):
This PR is needed and is focussed.

draft-ietf-moq-transport.md
```diff
@@ -522,91 +522,42 @@ This section is expected to cover details on:
...
-Media is produced as a set of layers, such as for example low definition
-and high definition, or low frame rate and high frame rate. Each object
-belonging to a track and a group has two attributes: the object-id, and
-the priority (or layer).
-
 When nodes or relays have to choose which object to send next, they
 apply the following rules:
 
-* within the same group, objects with a lower priority number (e.g. P1)
+* Within the same group, objects with a lower priority number (e.g. 1)
```
kixelated, line 548: Priorities are applied to streams so this gets super confusing based on the stream mode. The object_id is not always known after the stream header (track/group mode); am I supposed to change the stream priority in the middle of a stream?

draft-ietf-moq-transport.md
```diff
@@ -522,91 +522,42 @@ This section is expected to cover details on:
...
-eventually reach consensus and define meta data that controls this
-behavior.
-
-There have been proposals to allow emitters to coordinate the allocation
-of layer priorities across multiple coordinated tracks. At this point,
-these proposals have not reached consensus.
+  number (e.g., 2)
+* Within the same group and priority, objects with a lower object-id
+  are always sent before objects with a higher object-id.
+* Objects from later groups are sent before objects of previous groups.
```
kixelated, line 553: What happens when a newer group has a higher send order?

I think it's a great recommendation to send newer groups first... unless you want head-of-line blocking for reliable live or VOD. This text makes it seem illegal?

draft-ietf-moq-transport.md
```diff
@@ -522,91 +522,42 @@ This section is expected to cover details on:
...
-further details on various proposals.
-
-### Proposal - Ordering by Priorities {#ordering-by-priorities}
-
-Media is produced as a set of layers, such as for example low definition
-and high definition, or low frame rate and high frame rate. Each object
-belonging to a track and a group has two attributes: the object-id, and
-the priority (or layer).
-
 When nodes or relays have to choose which object to send next, they
```
kixelated, line 545: I think this text could be salvaged by recommending how `send_order` is computed. Otherwise it just conflicts with the above paragraphs that just tell you to use send order.

```suggestion
When the publisher chooses which object to send next via `send_order`, they
```

draft-ietf-moq-transport.md
```diff
@@ -522,91 +522,42 @@ This section is expected to cover details on:
...
 * Newly encoded objects can have a smaller send order than outstanding
   objects.
-* Packet loss or flow control can delay the send of individual streams.
+* Packet loss or flow control can delay sending individual streams.
 * The sender might not support stream prioritization.
 
-TODO: Refer to Congestion Response and Prioritization Section for
-further details on various proposals.
-
-### Proposal - Ordering by Priorities {#ordering-by-priorities}
```
kixelated, line 580: I feel like this proposal has been deprecated since implementing stream mapping, since the priority is in the stream header now. Relays are expected to respect it and forward streams as they were received.

draft-ietf-moq-transport.md
```diff
@@ -522,91 +522,42 @@ This section is expected to cover details on:
...
-an ongoing group results in better user experience than strict adherence
-to the freshness rule. We expect that that the working group will
-eventually reach consensus and define meta data that controls this
-behavior.
-
-There have been proposals to allow emitters to coordinate the allocation
-of layer priorities across multiple coordinated tracks. At this point,
-these proposals have not reached consensus.
+  number (e.g., 2)
+* Within the same group and priority, objects with a lower object-id
```
kixelated, line 551: Regardless of send order?

kixelated (changes requested):
I very much like the cleanup but there's some conflicts, hence being two separate sections in the first place.

draft-ietf-moq-transport.md
suhasHere: @kixelated can you please elaborate on what needs clarification here ?   Streaming modes naturally fit with priorities and we have implemented the same.

If stream per group is selected , all the objects with in a group have same priority.  So objects are delivered in the receive order at a relay within a stream.
if stream per object is selected and if the application marks all the objects with same priority within a group, then objects end up on the same priority queue level and are sent in the receive order.  If application marks each object with a different priorities, they end up on different levels in the queue and are sent according to the priority order.

The above text seems to cover all the cases if am reading it right.

draft-ietf-moq-transport.md
ianswett: Good point, personally I think I like "Object Priority" based on how it's discussed and used.  But suggestions are welcome.

draft-ietf-moq-transport.md
kixelated: So in general this text needs some clarification on if it's normative or non-normative. I interpret it as normative, outlining the algorithm that each publisher uses to compute `send_order`.

If so, it breaks some use-cases, like using send order for head-of-line blocking. It also just makes the `send_order` redundant, as a publisher could compute the send order using `group_id`, `object_id` and these rules. 

And what happens when there's a mismatch? Is a relay supposed to error if `send_order` does not follow these rules?

I mean this was the crux of the debate way back then and why there are two sections. I thought the publisher should bake the rules into the send_order, while Christian thought we should have explicit rules. This PR is effectively doing both by merging the sections together; allowing publishers to choose the send order per stream but dictating the rules?

kixelated:
Also +1 to renaming `send order` to `priority`.

suhasHere:
> Also +1 to renaming `send order` to `priority`.

+1

draft-ietf-moq-transport.md
suhasHere: Priority is saved as part of "Canonical Object" along with Object Forwarding preference . So that when they are served from  cache, regardless of stream mapping, right priority is applied and mapped to the right stream.

draft-ietf-moq-transport.md
```diff
@@ -522,91 +522,35 @@ This section is expected to cover details on:
...
-* Ensuring that objects are delivered in the order intended by the
-  emitter
-* Allowing nodes and relays to skip or delay some objects to deal with
-  congestion
-* Ensuring that emitters can accurately predict the behavior of relays
-* Ensuring that when relays have to skip and delay objects belonging to
-  different tracks that they do it in a predictable way if tracks are
-  explicitly coordinated and in a fair way if they are not.
+The publisher's Object Priority SHOULD be followed when possible. The
+contents of each object are still delivered in order.
```
afrind, line 536: s/still/always/ ?

draft-ietf-moq-transport.md
afrind: We've got text elsewhere that says that:

* when using stream-per-group, all objects have the same priority (sort of, see note)
* when using stream-per-group, the relay can't skip or reorder objects within the group

It might help clarify if this section referenced those bits, or explained the intersection of forwarding preferences and priorities.

Note: Our wire format allows the same group to be transmitted with different priorities if the group is carried on two or more streams (eg: disconnect in the middle of a group, objects 1-10 on stream 1, objects 11-20 on a different stream, or connection).

draft-ietf-moq-transport.md
afrind: I think this text should be scoped to tiebreakers:

"When objects from two or more groups have the same priority, send the objects from the more recent group first"

But for the "reliable live" case, you probably still want a subscriber side override to get groups in ascending order.

draft-ietf-moq-transport.md
afrind: Here priority == send order, so the qualifier (within the same group and priority) is that these objects have identical send order.

afrind (commented):
Individual Review:

Overall this is an improvement over the text that we have.  I don't think it's the end of the story but it's probably a good incremental step for now?

draft-ietf-moq-transport.md
ianswett, line 536:
```suggestion
contents of each object are delivered in order.
```

ianswett:
At the interim, we arrived at a much clearer direction on priorities, so closing this.

Issue #396, opened 2024-02-20
How does a relay/publisher know what to prioritize or drop?


ianswett:
It's unclear to me what objects a relay should prioritize or drop when congestion limited.  This is discussed in a few places, but I want to only focus on the question of what to send first and when to give up on sending an Object entirely.

Specific examples:

1. Is there a point at which Objects from a subscription should no longer be transmitted?  If so, what determines that time?
2. How does a relay decide whether to send the newest object or oldest?  Does it depend upon group boundaries?
3. Can a relay infer that some Objects are droppable if they have lower priority than other Objects in the same group?

My strawman proposal is that subscriptions should have numerical priorities, so a relay knows whether it should send Objects from one subscription or another given limited bandwidth.  Additionally, subscriptions should have a TTL so relays know at what point they should stop retransmitting Objects.

kixelated:
It depends on why we're dropping an OBJECT.

- Viewer has skipped past OR
- Relay is out storage.

In the first scenario, the receiver should be in charge since it's knows the jitter buffer size and whether it wants to block. Prioritization gets us most of the way here but maybe we add an explicit "I don't want GROUP X" message to save bandwidth.

In the second scenario, the sender must be in charge since. I would drop based on TTL with an explicit "I won't send GROUP X" message. In theory you could drop based on send order, but it's not time bounded so you end up catching way too much.

It does feel weird that they are two different units. Maybe the receiver could signal its TTL as well, but that feels wrong without advanced knowledge of the inbound content.

wilaw:
> Is there a point at which Objects from a subscription should no longer be transmitted? If so, what determines that time?

The publisher may add an explicit "expiration" absolute timestamp T to Objects. Any relay forwarding an object for which now() > T may drop that object or remove it from a pending send queue if there is congestion. 

> How does a relay decide whether to send the newest object or oldest? Does it depend upon group boundaries?

Objects can carry relative priorities, set by the publisher. These may correspond to newest or oldest, or they may not. The relay always sends highest priority items first when under congestion.

> Can a relay infer that some Objects are droppable if they have lower priority than other Objects in the same group?

No, a relay should NEVER arbitrarily drop objects, it should just delay their sending, until either they expire or the relay runs out of queue memory. In both these latter cases, to should report this condition to the subscriber.

ianswett:
[...]

Issue #419, opened 2024-03-18
Reliable Live use-case


kixelated:
I'm just going to coin a new term: **reliable live**

It's the use-case when a viewer wants to receive live content but doesn't want to skip anything, as quality is more important than latency. This is an extremely common scenario currently served by HLS/DASH and needs to be supported by MoQ too.

The player implements this by pausing playback (buffering) when the buffer is depleted until it reaches a minimum size. There's also a target latency, which is effectively the buffer size at steady state. The player will not skip over content unless there is an explicit gap. Let's assume fixed group sizes for simplicity. All of the below examples use:
```
group_size=2s
target_buffer=5s
min_buffer=1s
```

## Current Draft
In the current draft, a player start playback by issuing:

```
-> SUBSCRIBE 480p start=-2
<- SUBSCRIBE_OK 480p latest=6835
```

We added #343 so the player would know to start playback at group 6833, buffering any future groups. Unfortunately the player currently doesn't know about gaps, so it could block indefinitely waiting for 6834 (or any group). We need some sort of GROUP_DROPPED message to signal that the player should skip over, much like a DISCONTINUITY in HLS. This has been touted as a feature of FETCH, but it's something we also need for SUBSCRIBE.

The bigger issue is prioritization. If the send_order is decreasing, which is required to allow players to skip over old content, then the buffer will fill _backwards_.

```
-> SUBSCRIBE 480p start=-2
<- SUBSCRIBE_OK 480p latest=6835
<- OBJECT group=6835 send_order=0
<- OBJECT group=6834 send_order=1
<- OBJECT group=6833 send_order=2
```

Playback is blocked waiting for the first 1s of group 6833 but we receive newer groups first in their entirety, including possibly some of 6836 as it starts being generated. We will eventually be able to play this data, but this will dramatically increase startup and buffering time. This is also a problem for DVR, as the prioritization required for live does not make sense for VOD.

The broadcaster could remedy this by baking in head-of-line blocking into send_order but I don't want to do that. A participant on a conference stage may want interactive latency, while a viewer may want high quality (HoLB) instead. A viewer on a TV will want high quality, while a viewer using an interactive application will want low latency. It's not the end of the world but it's a missed opportunity.

The other option is to fetch each group one at a time like HLS/DASH:
```
-> SUBSCRIBE 480p start=-2 end=-1
<- SUBSCRIBE_OK 480p latest=6835
<- OBJECT group=6833
<- SUBSCRIBE_DONE 480p
[...]

Issue #411, opened 2024-03-06
Proposal: Subscriber Prioritization


kixelated:
I know we're not allowed to talk about priorities any longer but I wanted to float another option. 

The subscriber is in charge of prioritization:
- Add `order` to SUBSCRIBE: `ASC` or `DESC`
- Add `priority` to SUBSCRIBE: `VarInt`?
- Remove `send_order` from OBJECT.

The publisher prioritizes streams based on:
```rust
let order = subscribe.order == DESC ? 1 : -1;
subscribe.priority > object.group * order > -object.id
```

Depending on the QUIC API, publisher still needs to compute a send order for each stream. It can do this by inserting each stream into a priority queue based on the above comparison, using the resulting position in the queue as the send order. This was already a requirement for QUIC APIs that did not support a u64 send order.

Alternatively, a QUIC API could implement nested priority groups, perhaps something like [sendGroup](https://www.w3.org/TR/webtransport/#dom-webtransport-createsendgroup).

# Pros: Order
The benefits of `order=ASC` are similar to the proposed `FETCH` API in #368.

- Supports HLS/DASH-like experiences, where playback buffers instead of skips.
- Supports VOD playback.
- Supports advertisement playback.

Note that you could perform head-of-line blocking today* by issuing a SUBSCRIBE per group. However this is quite verbose, and it creates an RTT at the end of each group where the network is not being utilized.

# Pros: Priority
- Supports prioritizing unrelated tracks (from different encoders).
- Supports prioritizing based on focus state (sideline cam > main cam)
- Supports prioritizing high bitrate (main cam > sideline cam)
- The encoder is dumb, unaware of network priorities.

These could only be performed today by rewriting send order at a relay, which is impossible without catalog access.

# Cons
- SUBSCRIBE is even more complicated.
- No interleaved prioritizes between tracks/groups. (who cares?)
- Publisher has to compute/update send order on each stream/datagram created.
- Relay has to deduplicate conflicting SUBSCRIBEs
 
The main challenge is the last one. For example, two viewers could issue:
```
SUBSCRIBE track=720p order=ASC priority=3
SUBSCRIBE track=720p order=DESC priority=9
```

What does the relay send upstream? The idea behind publisher prioritization was to avoid this dilemma, but maybe there's another way we can solve it.

kixelated:
[...]

PR #421, opened 2024-03-19
Split SUBSCRIBE into SUBSCRIBE and FETCH

### LINKED ISSUE #269: Multiple subscriptions and updating a subscription

### LINKED ISSUE #296: Empty subscriptions
As mentioned in the sprawling #289, I think we should allow empty subscriptions, aka subscriptions that match no objects.

- It would be nice to have a way to check if a track exists without receiving any OBJECTs. The MoQ equivalent of a HTTP HEAD request.
- It could be useful to pre-warm the cache. An empty subscription could be the equivalent of "I'm interested, but don't have the bandwidth to receive objects yet". Again, similar to HTTP HEAD requests.

For example, a client could issue a `SUBSCRIBE /audio/japanese start=none` to accomplish two separate things:
1. The `/audio/japanese` track exists if it gets a SUBSCRIBE_OK.
2. The encoder/CDN starts producing the `/audio/japanese` track.
3. The client could resubscribe at any point to immediately start receiving the track.

### LINKED ISSUE #312: Clarify how to end a SUBSCRIBE at group N
If I wanted to receive all objects in group N, I would issue:

```
SUBSCRIBE
end_group=absolute/N
end_object=none
```

However the draft sneaks this restriction in, and it would actually be a protocol violation:

> EndObject's Mode MUST NOT be None if EndGroup's Mode is not None.

It wasn't clear to me until chatting with Alan how you're supposed to do it:

```
SUBSCRIBE
end_group=absolute/N+1
end_object=absolute/0
```

It's clarified in example 5, but I don't think this is obvious and it feels like an off-by-one. For example, if N was the last group in the broadcast, I can imagine some implementations incorrectly throwing an error or somehow mishandling it.

### LINKED ISSUE #326: Hints & sendOrder how can they work together

Finals questions:
- Do we need to add signals?

### LINKED ISSUE #350: When to use Track Alias or Subscribe ID?

### LINKED ISSUE #353: Problems with RelativeNext subscriptions
1. Consider the following sequence of groups and objects, some already produced , some that will be produced in the future. The largest group ID is currently 23 and the largest object ID is 3 (it also happens to be the last object in the group). 

Screenshot 2024-02-01 at 9 15 24 PM

Our definitions for RelativePrevious and RelativeNext are shown here. 

> RelativePrevious (0x2): Value is a delta from the largest sequence. 0 is the largest sequence, 1 is the largest sequence - 1, and so on.
> RelativeNext (0x3): Value is a delta from the largest sequence. 0 is the largest sequence + 1, 1 is the largest sequence + 2, and so on.

Example 6.4.3.1 for 'now' shows

```
Start Group: Mode=RelativePrevious, Value=0
Start Object: Mode=RelateiveNext, Value=0
End Group: Mode=None
End Object: Mode=None

StartGroup=Largest Group
StartObject=Largest Object + 1
```

In our example, largest group is 23, however Largest Object + 1 is 4, which does not exist. We need to adjust the definition of RelativeNext to allow for the fact that the next object may be in a different group.

2. There may be some players who simply want the relay to send the next available Object, in this example Object 0 of group 24. 

There is no current way to reliably ask for this, because StartGroup mode cannot be 'none', and we don't know which group that object will belong to. It may be in the currently largest group, or it may be in the next group. 

A solution here would be to relax the constraint that StartGroup mode cannot be 'none'. If we then request

```
Start Group: Mode=None
Start Object: Mode=RelativeNext, Value=0
```

it would instruct the relay "Send me your next object, no matter which group it belongs to".

5. In 6.4.3 we have examples titled 'Now',  'Current', "previous" and 'Next".  

'Now',  'Current'  are synonymous to most readers. We need to better titles to illustrate what is actually being requested here. Additionally, 
"Previous' doesn't indicate whether we want the previous group, or previous object. Similarly for next.

### LINKED ISSUE #377: What is the largest group/largest object if a relay has not received any objects?
This came up during interop - 

A relay received a subscribe with start={g=RelativeNext/0 o=Absolute/0}.  Since the relay had not received any objects, it defaulted largest to 0 and interpreted the subscribe as starting at {g=1, o=0}. When {g=0,o=0} arrived, it was not forwarded.  But there's a reasonable interpretation that it should have forwarded treated this as the beginning of the next subscription.

This is partially related to #337 and maybe #353

ianswett:
A PR corresponding to: https://datatracker.ietf.org/meeting/119/materials/slides-119-moq-subscribe-and-fetch

Fixes #269 
Fixes #296
Fixes #350 
Fixes #326
Fixes #377

Also relevant to #396, #326

Some open questions:
 - Do we want OBJECT_FETCH_DATAGRAM to be transmitted over datagrams only?

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1114,7 +1114,7 @@ ID` and the group indicated by `Group ID`.  All objects on the stream
 have the `Object Send Order` specified in the stream header.
 
 ~~~
-STREAM_HEADER_GROUP Message {
+STREAM_SUBSCRIBE_GROUP/STREAM_FETCH_GROUP Message {
   ID (i),
   Group ID (i)
   Object Send Order (i)
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -935,8 +935,6 @@ An OBJECT message contains a range of contiguous bytes from from the
 specified track, as well as associated metadata required to deliver,
 cache, and forward it.
 
-TODO: Determine whether FETCH can be delivered in all Object formats.
-
 ### Canonical Object Fields
 
 A canonical MoQ Object has the following information:
@@ -1178,17 +1176,16 @@ the most recently published Objects.
 
 A subscriber MUST NOT subscribe to the same track twice within a session.
 
-TODO: Decide if SUBSCRIBE should have an end.
-TODO: Decide if we want a new message for FETCH, or is it just a mode?
+There are 3 modes to indicate where to start the subscription:
 
-There are 3 modes:
+Last Group (0x0): Start from the the largest group for which the relay
+has received the last object.  If no complete group has been published,
+begin at the first group.
 
-Previous Group (0x0): Start the subscription from the Group prior to the most
-recent group, if any.
+Current Group (0x1): Start at the first Object from the largest received
+group.
 
-Current Group (0x1): Start the subscription from the currently open group.
-
-Next Group (0x2): Start the subscription from the next group.
+Next Group (0x2): Start at the first Object from from the next group.
 
 
 The format of SUBSCRIBE is as follows:
@@ -1200,6 +1197,7 @@ SUBSCRIBE Message {
   Track Namespace (b),
   Track Name (b),
   Mode (i),
+  Number of Groups (i),
   Number of Parameters (i),
   Track Request Parameters (..) ...
 }
@@ -1225,6 +1223,12 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
 * Mode: The subscription mode, as described above.
+
+* Number of Groups: The number of Groups a subscriber would like to receive.
+0 indicates there is no limit.
+TODO: Should 0 indicate the subscriber only wants to know what the most
+recent Group and Object are and doesn't want to to receive any data similar
+to a HTTP HEAD request?
 
 * Track Request Parameters: The parameters are defined in
 {{version-specific-params}}
@@ -1385,10 +1389,14 @@ message in the `Final Group` for this track.
 
 ## FETCH {#message-fetch}
 
-A receiver issues a FETCH to a publisher to request a range
-of Objects within a Track.
+A receiver issues a FETCH to a publisher to request a range of Objects
+within a Track.
 
-All avaialble Objects will be delivered in order.
+All available Objects will be delivered with the same delivery preference
+as Objects would be delivered with SUBSCRIBE.  When possible, earlier
+Objects SHOULD be delivered before later Objects, but it is possible to receive
+Objects out of order due to reordering, what is available in the cache, or
+other reasons.
 
 The format of FETCH is as follows:
 
@@ -1428,10 +1436,6 @@ EndObject's Mode MUST be None if EndGroup's Mode is None.
 
 * Track Request Parameters: The parameters are defined in
 {{version-specific-params}}
-
-On successful fetch, the publisher SHOULD start delivering
-objects from the StartGroup and StartObject.
-
 
 
 ## ANNOUNCE {#message-announce}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1071,26 +1071,26 @@ TODO: figure out how a relay closes these streams
 
 **Stream Header Track**
 
-When a stream begins with `STREAM_HEADER_TRACK`, all objects on the stream
-belong to the track requested in the Subscribe message identified by `Subscribe
-ID`.  All objects on the stream have the `Object Send Order` specified in the
-stream header.
+When a stream begins with `STREAM_SUBSCRIBE_TRACK` or 'STREAM_FETCH_TRACK', all
+objects on the stream belong to the track requested in the corresponding
+subscription or fetch as identified by the `Track Alias' or 'Fetch ID'.  All
+objects on the stream have the `Object Send Order` specified in the stream header.
 
 
 ~~~
-STREAM_HEADER_TRACK Message {
+STREAM_SUBSCRIBE_TRACK/STREAM_FETCH_TRACK Message {
   ID (i),
   Object Send Order (i),
 }
 ~~~
-{: #stream-header-track-format title="MOQT STREAM_HEADER_TRACK Message"}
+{: #stream-header-track-format title="MOQT STREAM_TRACK Message"}
 
-All Objects received on a stream opened with STREAM_HEADER_TRACK have an `Object
-Forwarding Preference` = `Track`.
+All Objects received on a stream opened with STREAM_SUBSCRIBE_TRACK or
+STREAM_FETCH_TRACK have an `Object Forwarding Preference` = `Track`.
 
 To send an Object with `Object Forwarding Preference` = `Track`, find the open
 stream that is associated with the subscription, or open a new one and send the
-`STREAM_HEADER_TRACK` if needed, then serialize the the following object fields.
+header, then serialize the the following object fields.
 
 ~~~
 {
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1007,7 +1007,7 @@ serialize object fields below, and terminate the stream.
 
 ~~~
 OBJECT_STREAM Message {
-  Track Alias (i),
+  ID (i),
   Group ID (i),
   Object ID (i),
   Object Send Order (i),
@@ -1016,10 +1016,7 @@ OBJECT_STREAM Message {
 ~~~
 {: #moq-transport-object-stream-format title="MOQT OBJECT_STREAM Message"}
 
-* Subscribe ID: Subscription Identifer as defined in {{message-subscribe-req}}.
-
-* Track Alias: Identifies the Track Namespace and Track Name as defined in
-{{message-subscribe-req}}.
+* ID: The Track Alias for subscriptions or the Fetch ID for Fetches.
 
 If the Track Namespace and Track Name identified by the Track Alias are
 different from those specified in the subscription identified by Subscribe ID,
@@ -1042,7 +1039,7 @@ will be dropped.
 
 ~~~
 OBJECT_DATAGRAM Message {
-  Track Alias (i),
+  ID (i),
   Group ID (i),
   Object ID (i),
   Object Send Order (i),
@@ -1074,7 +1071,7 @@ stream header.
 
 ~~~
 STREAM_HEADER_TRACK Message {
-  Track Alias (i),
+  ID (i),
   Object Send Order (i),
 }
 ~~~
@@ -1110,7 +1107,7 @@ have the `Object Send Order` specified in the stream header.
 
 ~~~
 STREAM_HEADER_GROUP Message {
-  Track Alias (i),
+  ID (i),
   Group ID (i)
   Object Send Order (i)
 }
@@ -1440,6 +1437,23 @@ EndObject's Mode MUST be None if EndGroup's Mode is None.
 
 * Track Request Parameters: The parameters are defined in
 {{version-specific-params}}
+
+## FETCH_CANCEL {#message-fetch-cancel}
+
+A subscriber issues a `FETCH_CANCEL` message to a publisher indicating it
+is no longer interested in receiving Objects for the specified track and
+Objects should stop being sent as soon as possible.
+
+The format of `FETCH_CANCEL` is as follows:
+
+~~~
+FETCH_CANCEL Message {
+  Fetch ID (i)
+}
+~~~
+{: #moq-transport-unsubscribe-format title="MOQT UNSUBSCRIBE Message"}
+
+* Fetch ID: Fetch Identifer as defined in {{message-fetch-req}}.
 
 
 ## ANNOUNCE {#message-announce}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -745,7 +745,7 @@ MOQT Message {
 |------:|:----------------------------------------------------|
 | 0x0   | OBJECT_STREAM ({{object-message-formats}})          |
 |-------|-----------------------------------------------------|
-| 0x1   | OBJECT_DATAGRAM ({{object-message-formats}}) |
+| 0x1   | OBJECT_DATAGRAM ({{object-message-formats}})        |
 |-------|-----------------------------------------------------|
 | 0x3   | SUBSCRIBE ({{message-subscribe-req}})               |
 |-------|-----------------------------------------------------|
@@ -768,6 +768,10 @@ MOQT Message {
 | 0xC   | SUBSCRIBE_RESET ({{message-subscribe-reset}})       |
 |-------|-----------------------------------------------------|
 | 0x10  | GOAWAY ({{message-goaway}})                         |
+|-------|-----------------------------------------------------|
+| 0x20  | FETCH ({{message-fetch}})                           |
+|-------|-----------------------------------------------------|
+| 0x21  | FETCH_CANCEL ({{message-fetch-cancel}})             |
 |-------|-----------------------------------------------------|
 | 0x40  | CLIENT_SETUP ({{message-setup}})                    |
 |-------|-----------------------------------------------------|
@@ -931,6 +935,8 @@ An OBJECT message contains a range of contiguous bytes from from the
 specified track, as well as associated metadata required to deliver,
 cache, and forward it.
 
+TODO: Determine whether FETCH can be delivered in all Object formats.
+
 ### Canonical Object Fields
 
 A canonical MoQ Object has the following information:
@@ -976,7 +982,6 @@ serialize object fields below, and terminate the stream.
 
 ~~~
 OBJECT_STREAM Message {
-  Subscribe ID (i),
   Track Alias (i),
   Group ID (i),
   Object ID (i),
@@ -1012,7 +1017,6 @@ will be dropped.
 
 ~~~
 OBJECT_DATAGRAM Message {
-  Subscribe ID (i),
   Track Alias (i),
   Group ID (i),
   Object ID (i),
@@ -1045,7 +1049,6 @@ stream header.
 
 ~~~
 STREAM_HEADER_TRACK Message {
-  Subscribe ID (i)
   Track Alias (i),
   Object Send Order (i),
 }
@@ -1082,7 +1085,6 @@ have the `Object Send Order` specified in the stream header.
 
 ~~~
 STREAM_HEADER_GROUP Message {
-  Subscribe ID (i),
   Track Alias (i),
   Group ID (i)
   Object Send Order (i)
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1171,46 +1173,23 @@ OBJECT_STREAM {
...
-
-~~~
-Location {
-  Mode (i),
-  [Value (i)],
-}
-~~~
-
-There are 4 modes:
+A subscriber MUST NOT subscribe to the same track twice within a session.
```
hardie, line 1204: I thought we had agreed that a client could want multiple copies for different uses (e.g. storage vs. render).  If it did that, using two different start locations, why would that be forbidden in a single session?

ianswett: The cases one would want multiple copies are covered by FETCH.  Having multiple subscribes can add complexity in some cases and I'm not aware of a clear usecase.

suhasHere: For use-case where i want to catch up last 5 mins, while i join the call live. I can see it implemented with subscribe , fetch combination ..

draft-ietf-moq-transport.md
```diff
@@ -1382,7 +1295,12 @@ end prior to the expiry time or last longer.
 If 0, then the Largest Group ID and Largest Object ID fields will not be
 present.
 
-* Largest Group ID: the largest Group ID available for this track. This field is only present if ContentExists has a value of 1.
+TODO: Should these be the first Group being delivered?  If the previous group
```
ianswett, line 1298:
```suggestion
TODO: Should these be the first Group being delivered?  If the last group
```

draft-ietf-moq-transport.md
```diff
@@ -1275,24 +1246,19 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
...
-Mode is None for an open-ended subscription.
-
-* EndObject: The last Object requested in the subscription, exclusive.
-EndObject's Mode MUST be None if EndGroup's Mode is None.  EndObject's Mode MUST
-NOT be None if EndGroup's Mode is not None.
+* Number of Groups: The number of Groups a subscriber would like to receive.
+0 indicates there is no limit.
+TODO: Should 0 indicate the subscriber only wants to know what the most
+recent Group and Object are and doesn't want to to receive any data similar
+to a HTTP HEAD request?
```
suhasHere, line 1255: wonder if we should make it invalid to set its value as 0 to avoid ambiguities or make it optional field and if it is omitted it means as long as track is alive?

draft-ietf-moq-transport.md
```diff
@@ -1469,6 +1387,75 @@ message in this track.
...
+  Track Name (b),
+  StartGroup (i),
+  StartObject (i),
+  EndGroup (i),
+  EndObject (i),
+  Number of Parameters (i),
+  Track Request Parameters (..) ...
+}
+~~~
+{: #moq-transport-subscribe-format title="MOQT FETCH Message"}
```
ianswett, line 1417:
```suggestion
{: #moq-transport-fetch-format title="MOQT FETCH Message"}
```

draft-ietf-moq-transport.md
```diff
@@ -1198,46 +1195,22 @@ OBJECT_STREAM {
...
-                                             ^
-                                      Largest Sequence
-RelativePrevious Value:  4    3    2    1    0
-RelativeNext Value:                               0    1  ...
-~~~
-{: title="Relative Indexing"}
+Current Group (0x1): Start at the first Object from the largest received
+group.
 
+Next Group (0x2): Start at the first Object from from the next group.
```
suhasHere, line 1212: s/from from/from

draft-ietf-moq-transport.md
ianswett: I think we need a way to say that there is no limit on the number of groups.  We can encode that in a number of ways, ie: 6 modes instead of 3, etc.

suhasHere (approved):
Thanks @ianswett  This seems to be heading in the right direction. I do have few nits and clarification questions (posted)

draft-ietf-moq-transport.md
```diff
@@ -1469,6 +1395,75 @@ message in this track.
...
 message in the `Final Group` for this track.
 
+
+## FETCH {#message-fetch}
+
+A receiver issues a FETCH to a publisher to request a range of Objects
+within a Track.
+
+All available Objects will be delivered with the same delivery preference
+as Objects would be delivered with SUBSCRIBE.  When possible, earlier
```
fluffy, line 1405: slight tweak, instead of  earlier, I think it would be better to use lower object id because order of things were delivered in time could be different than that

draft-ietf-moq-transport.md
```diff
@@ -1469,6 +1395,75 @@ message in this track.
...
+  Number of Parameters (i),
+  Track Request Parameters (..) ...
+}
+~~~
+{: #moq-transport-fetch-format title="MOQT FETCH Message"}
+
+* Fetch ID: Identifier of the Fetch that is unique within the session.
+`Fetch ID` is a monotonically increasing variable length integer which
+MUST not be reused within a session. `Fetch ID` is used by subscribers and
+the publishers to identify a given subscription.
```
fluffy, line 1430: perhaps change "subscription" to "fetch"

draft-ietf-moq-transport.md
```diff
@@ -1109,8 +1115,7 @@ have the `Object Send Order` specified in the stream header.
 
 ~~~
 STREAM_HEADER_GROUP Message {
```
suhasHere, line 1117: STREAM_SUBSCRIBE_GROUP/STREAM_FETCH_GROUP

draft-ietf-moq-transport.md
```diff
@@ -1382,7 +1303,12 @@ end prior to the expiry time or last longer.
 If 0, then the Largest Group ID and Largest Object ID fields will not be
 present.
 
-* Largest Group ID: the largest Group ID available for this track. This field is only present if ContentExists has a value of 1.
+TODO: Should these be the first Group being delivered?  If the last group
+isn't cached, but was requested, should that be indicated?
+ie: by making ContentExists an enum?
```
suhasHere, line 1308: If the last group was cached , but TTL expired on it and got removed. Subscribe should start from the data it has..

draft-ietf-moq-transport.md
```diff
@@ -1469,6 +1395,75 @@ message in this track.
...
+  Fetch ID (i),
+  Track Namespace (b),
+  Track Name (b),
+  StartGroup (i),
+  StartObject (i),
+  EndGroup (i),
+  EndObject (i),
+  Number of Parameters (i),
+  Track Request Parameters (..) ...
+}
```
fluffy, line 1423: This comment could be in a different PR, but I think it might be good to add an flag to the fetch that indicated the results should be delivered in reverse order.

fluffy (approved):

draft-ietf-moq-transport.md
ianswett, line 1117:
```suggestion
STREAM_SUBSCRIBE_GROUP/STREAM_FETCH_GROUP Message {
```

draft-ietf-moq-transport.md
```diff
@@ -1198,46 +1203,22 @@ OBJECT_STREAM {
...
-                                             ^
-                                      Largest Sequence
-RelativePrevious Value:  4    3    2    1    0
-RelativeNext Value:                               0    1  ...
-~~~
-{: title="Relative Indexing"}
+Current Group (0x1): Start at the first Object from the largest received
+group.
 
+Next Group (0x2): Start at the first Object from from the next group.
```
ianswett, line 1220:
```suggestion
Next Group (0x2): Start at the first Object from the next group.
```

draft-ietf-moq-transport.md
ianswett: Thanks, I would like to have that conversation along with the larger priorities conversation if possible, but it's a completely reasonable idea.

draft-ietf-moq-transport.md
ianswett: Agreed, my concern is that if last isn't available, how is that indicated?  If one finds out the largest group ID is 3, and it asked for last, it would expect to receive group 2, but if it's not available, it wouldn't.

draft-ietf-moq-transport.md
janaiyengar: @Ted: We had agreed on allowing this for FETCH, and explicitly disallowing dups for SUBSCRIBE, since FETCH is more like an HTTP GET.

janaiyengar (commented):
Broadly looks good, follows the arc of what we discussed.

draft-ietf-moq-transport.md
kixelated: What's the benefit to adding this restriction, effectively mandating deduplication? I don't see what's the harm in allowing dumb proxies that don't deduplicate subscriptions.

draft-ietf-moq-transport.md
kixelated: I'd like a flag too, see #411. We need it for SUBSCRIBE in the reliable live use-case.

draft-ietf-moq-transport.md
```diff
@@ -796,13 +796,25 @@ MOQT Message {
 |-------|-----------------------------------------------------|
 | 0x10  | GOAWAY ({{message-goaway}})                         |
 |-------|-----------------------------------------------------|
+| 0x20  | FETCH ({{message-fetch}})                           |
+|-------|-----------------------------------------------------|
+| 0x21  | FETCH_CANCEL ({{message-fetch-cancel}})             |
```
kixelated, line 801: There needs to be a `FETCH_OK`/`FETCH_ERROR` message too. I don't see how you can signal stuff like 404 or unauthorized otherwise.

draft-ietf-moq-transport.md
```diff
@@ -796,13 +796,25 @@ MOQT Message {
...
 | 0x40  | CLIENT_SETUP ({{message-setup}})                    |
 |-------|-----------------------------------------------------|
 | 0x41  | SERVER_SETUP ({{message-setup}})                    |
 |-------|-----------------------------------------------------|
-| 0x50  | STREAM_HEADER_TRACK ({{multi-object-streams}})      |
+| 0x48  | OBJECT_FETCH_STREAM ({{object-message-formats}})    |
+|-------|-----------------------------------------------------|
+| 0x49  | OBJECT_FETCH_DATAGRAM ({{object-message-formats}})  |
+|-------|-----------------------------------------------------|
+| 0x50  | STREAM_SUBSCRIBE_TRACK ({{multi-object-streams}})   |
```
kixelated, line 811: Can we shard the `id` (even/odd) based on if it's a FETCH/SUBSCRIBE?

Scratch that, why does it even matter? Why not have SUBSCRIBE and FETCH share the same ID space?

kixelated (changes requested):
I understand the desire to split the world into VOD and REALTIME via FETCH and SUBSCRIBE, but I think this proposal misses the nuances of live streams that startle between both worlds. I'd like some clarification in #419 as to how a player is supposed to use this API because it seems pretty terrible.

And I think the duplication between SUBSCRIBE/FETCH is just super messy. The justification is that now a relay doesn't need to implement relative ranges beyond +- 1 group. If that's a sticking point for folks, then I much rather add that restriction to the current SUBSCRIBE message along with `order=ASC|DESC` in #411.

draft-ietf-moq-transport.md
```diff
@@ -1198,46 +1203,22 @@ OBJECT_STREAM {
...
-~~~
-Location {
-  Mode (i),
-  [Value (i)],
-}
-~~~
+A subscriber MUST NOT subscribe to the same track twice within a session.
 
-There are 4 modes:
+There are 3 modes to indicate where to start the subscription:
```
kixelated, line 1211: Doesn't this break seamless ABR, since you can't subscribe using an absolute group in the future? The best you can do is subscribe to next group and pray for no race condition.

kixelated (changes requested):
And I think this breaks ABR if I understand correctly?

wilaw:
How would a live player that wants to start 3 groups back from the live head use this combination of SUBSCRIBE and FETCH? It could issue the SUBSCRIBE, which would deliver the latest group and all future groups as well as information about the latest (largest) group number. However, since group numbers increase monotonically by a variable increment, the player cannot use this information to construct the appropriate FETCH, because it cannot predict the two prior group numbers.

suhasHere:
> However, since group numbers increase monotonically by a variable increment, the player cannot use this information to construct the appropriate FETCH, because it cannot predict the two prior group numbers.

Player needs to know how the group numbers increment.  Its the application and if it has no clue , then something is already wrong.

suhasHere:
> If that's a sticking point for folks, then I much rather add that restriction to the current SUBSCRIBE message along with `order=ASC|DESC` as proposed in #411.

Can you please answer my question on the open issue on how can relay make such decision upstream when there are conflicting subscriptions with respect to the order.  May be last mile subscribe has a utility, but I am not sure how to solve the open issue on this.

draft-ietf-moq-transport.md
suhasHere: Deduping multiple conflicting subscriptions is not straighforward, also it leads to weird edge cases when there is partial data . I think these were reflected during Denver interim discussions at some point.

fluffy: Is there an use that drives need for duplicate subscriptions

kixelated: > Is there an use that drives need for duplicate subscriptions

As of the current draft, AUTH is in the SUBSCRIBE message. A relay that doesn't know the auth scheme would have to forward each SUBSCRIBE upstream.

I strongly think we should decouple auth anyway. I'd like to see an AUTH_NAMESPACE message that could be used for subsequent ANNOUNCE and SUBSCRIBE messages on the control stream.

> Deduping multiple conflicting subscriptions is not straighforward, also it leads to weird edge cases when there is partial data . I think these were reflected during Denver interim discussions at some point.

Yeah that's my point. This restriction forces a relay to deduplicate, for example if two viewers issue:

```
viewer1 -> SUBSCRIBE start=next
...
viewer2 -> SUBSCRIBE start=prev
```

A relay SHOULD deduplicate these for the upstream subscribe, but this text makes it a MUST. Doing this is tricky much like the existing SUBSCRIBE message.

kixelated: As a side note.

> A subscriber MUST NOT subscribe to the same track twice within a session.

I think the intent is to prevent duplicate concurrent subscriptions. However, a literal interpretation is that you're never allowed to subscribe again even after an unsubscribe, which of course breaks many things.

draft-ietf-moq-transport.md
```diff
@@ -1469,6 +1395,75 @@ message in this track.
...
+other reasons.
+
+The format of FETCH is as follows:
+
+~~~
+FETCH Message {
+  Fetch ID (i),
+  Track Namespace (b),
+  Track Name (b),
+  StartGroup (i),
```
kixelated, line 1417: What happens if the start group/object is in the future?

draft-ietf-moq-transport.md
kixelated: Speaking of 404, what happens if some of the groups/objects in the range requested don't exist or have expired? Without a way of signaling that, the subscriber would wait indefinitely for the FETCH to complete.

ex. a DVR client that issues a FETCH for 60s ago to replay a highlight. However the start of the highlight may have been evicted from the LRU by then, and is no longer available on origin either.

kixelated:
> > If that's a sticking point for folks, then I much rather add that restriction to the current SUBSCRIBE message along with `order=ASC|DESC` as proposed in #411.
> 
> Can you please answer my question on the open issue on how can relay make such decision upstream when there are conflicting subscriptions with respect to the order. May be last mile subscribe has a utility, but I am not sure how to solve the open issue on this.

What about `order=ASC | DESC | DEFAULT`? The relay could let the publisher/upstream choose like it does today.

Note that FETCH is not immune to this. There could be a continuous SUBSCRIBE from broadcaster -> origin (ascending) and a FETCH from origin -> relay (descending). Or a viewer could SUBSCRIBE and later another viewer could FETCH, leveraging some of the existing subscription.

viewer1:
```
-> SUBSCRIBE 
<- SUBSCRIBE_OK start=420
<- OBJECT 420
(upstream congestion)
<- OBJECT 422
<- OBJECT 423
<- OBJECT 421
```

viewer2:
```
-> FETCH start=419 end=421
<- OBJECT 420
<- OBJECT 422
(blocked on 423)
<- OBJECT 421
```

But I dunno, I kind of feel like this is a non-issue. Viewers will either request the same order, or upstream congestion is rare, or defaulting to ORDER=desc is required for live and good enough for VOD and reliable live. We're talking about edge cases of a prioritization scheme, which historically don't have much impact.

draft-ietf-moq-transport.md
suhasHere: +1 on needing a way for the fetch response to identify gaps.  We need to add FETCH_OK to include something on the lines of

Say if FETCH request was for the range 100-200 and and let's say for whatever reasons 110-120, 190-200 are missing, the response should have

```
{
    Gaps {
    type-group 
    [110-120,  190-200]
  }
}
```

Nice thing about fetch is it is able to find answers

suhasHere:
> What about `order=ASC | DESC | DEFAULT`? The relay could let the publisher/upstream choose like it does today.

The challenge here is any recommendation we give here will not be appropriate for Relays. I have not seen a case where given an application, say that publishes oldest to newest, have some percentage of subscribers wanting oldest to newest and rest asking newest to oldest. Subscribers (endpoints) knows what application they are participating it and they know the right expectation. 

OTOH I can see fetch being independent request, can influence some form of control in the receive order. But even there I am yet to be convinced if there is a real use-case/requirement that few endpoints fetch in one order vs few in a different order for a given application. 

My main concern is we are pushing recommendations/suggestions on relays for applications and Relays can't make a informed decision under ambiguities.

afrind:
Individual Comment:

> I am unaware of cases where given an application, say that publishes oldest to newest, have some percentage of subscribers wanting oldest to newest and rest asking newest to oldest

Here's a use case: A live stream with a recording option.  Some subscribers are watching live, so want newest first.  Recording subscriber doesn't care about latency and would prefer oldest first.  I think the SUBSCRIBE verb is still the right one for both cases, rather than forcing the recorder to use FETCH in a loop?

fluffy:
The requirements for recording are usually record all the objects that were sent. If a object was lost and never sent, the recording system would not want to wait for it and not record future packets.  I'm pretty skeptical of any use case where latency does not matter being a core case for MoQ. That said, a recoding system that did a subscribe to the whole namespace to get all the objects as they were produced sounds like the most likely thing for recording systems to build.

fluffy:
Please see my related email on the the list. Thanks

draft-ietf-moq-transport.md
```diff
@@ -1198,46 +1203,22 @@ OBJECT_STREAM {
...
-  [Value (i)],
-}
-~~~
+A subscriber MUST NOT subscribe to the same track twice within a session.
 
-There are 4 modes:
+There are 3 modes to indicate where to start the subscription:
 
-None (0x0): The Location is unspecified, Value is not present
+Last Group (0x0): Start from the the largest group for which the relay
```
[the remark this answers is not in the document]
afrind: Individual comment:

> the largest group for which the relay has received the last object

I find this strange and hard to reason about.  

First, we don't yet have a mechanism for the relay to know the last object in the group (though we all assume we will eventually have one).  But if we really want this semantic, maybe we want to delay adding this mode to the draft until we can actually implement it.

What's strange to me is that if the relay has received [group 7: all objects] and [group 8: all object *except the last*], and [group 9, object 0], a request for `Last Group` would serve 7 instead of 8.  Is there an application where that's the desired behavior?  

I find it hard to reason about because I could ask 3 different relays and get 3 different groups relative to the live head, depending on what happens to be in their cache. 

There are two other ways to handle last that seem more straightforward to me:

1. Last is the largest group smaller than current for which the relay has *any* objects
2. Last is `current group - 1`

Option 1 suffers from the same potential relay inconsistency as the proposed text, but it seems more likely to be useful.

I prefer option 2 because it is *predictable* - I know exactly what group I am going to get relative to the live head.  I get the vibe that some folks have a specific use case in mind for discontiguous group ids, which would mean option 2 wouldn't work for them.  If there is a use-case, can we write it down (maybe mail to the list) so we can evaluate pros and cons? I think Luke has given a use case for making group IDs sequential: they can be used for gap detection, and would allow relays to do simple math to compute relative groups.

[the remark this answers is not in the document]
kixelated: Oh yeah I didn't even read the full text. Any subscribe decision based on a relay's current cache is going to be problematic. 

`current-1` makes more sense. The wrinkle is that `group_ids` are apparently not sequential , so group-1 might not exist (hence why -N was more flexible).

I would remove this mode altogether.

draft-ietf-moq-transport.md
```diff
@@ -1469,6 +1395,75 @@ message in this track.
...
+
+~~~
+FETCH Message {
+  Fetch ID (i),
+  Track Namespace (b),
+  Track Name (b),
+  StartGroup (i),
+  StartObject (i),
+  EndGroup (i),
+  EndObject (i),
```
martinduke, line 1420: Is there some way to express "give me all the objects in the group" if we don't know the max object ID?

afrind: With SUBSCRIBE, the requested range was [Start, End), so you can request Start=X/0, End=X+1/0 to get all the objects in group X.

afrind: Individual Comment:

What is the normative behavior if `end` is larger than `largest` for a live track?

Here are some possibilities:

1) Fail the request
2) return everything up to min(end, live) at the time of the request, indicate end in a fetch OK
3) return everything up to min(end, live) at the time of the response, indicate end in a fetch trailer
4) When the fetch catches up to live, continue offering FETCH semantics for new objects (all objects delivered or gap notification), until end is reached

1 and 2 are fairly easy to implement.  3 is a little trickier but accounts for live head moving in the time it takes to serve the beginning of the request.  4 feels the hardest, but maybe handles the "reliable live" use case the best?

Are the other options?

wilaw: I'd vote for 4 - `When the fetch catches up to live, continue offering FETCH semantics for new objects (all objects delivered or gap notification), until end is reached .`. As you state, this satisfies a "reliable live" use case with a start behind the live head. 

I'd also suggest that EndGroup and EndObject can be omitted, in which it defaults to the behavior [4]. This prevents the ugliness of people asking for group 99999999999999 when all they really want is delivery until end-of -track.

ianswett:
The draft has evolved substantially, so if we want to revisit SUBSCRIBE vs FETCH, we will need a new PR.

PR #425, opened 2024-04-03
Add a way to get the current group ID

martinduke:
There is a bit of angst about using SUBSCRIBE to obtain the current Group ID. This PR is a straw man to provide an explicit means of obtaining the group ID.

I would be surprised if this ended up in the final RFC, but it might be useful to keep in the draft as a primitive while the other SUBSCRIBE/FETCH stuff moves around.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1615,8 +1615,7 @@ value is a malformed message.
 contain the highest Group and object ID known.
 
 0x04: The sender is a relay that cannot obtain the current track status from
-upstream. Subsequent fields contain the largest group and object ID known to
-the sender.
+upstream. Subsequent fields contain the largest group and object ID known.
 
 Any other value in the Status Code field is a malformed message.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -794,6 +794,10 @@ MOQT Message {
 |-------|-----------------------------------------------------|
 | 0xC   | ANNOUNCE_CANCEL ({{message-announce-cancel}})       |
 |-------|-----------------------------------------------------|
+| 0xD   | GET_CURRENT_GROUP ({{message-get-current-group}})   |
+|-------|-----------------------------------------------------|
+| 0xE   | CURRENT_GROUP ({{message-current-group}})           |
+|-------|-----------------------------------------------------|
 | 0x10  | GOAWAY ({{message-goaway}})                         |
 |-------|-----------------------------------------------------|
 | 0x40  | CLIENT_SETUP ({{message-setup}})                    |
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -794,6 +794,10 @@ MOQT Message {
 |-------|-----------------------------------------------------|
 | 0xC   | ANNOUNCE_CANCEL ({{message-announce-cancel}})       |
 |-------|-----------------------------------------------------|
+| 0xD   | GET_CURRENT_GROUP ({{message-get-current-group}})   |
```
kixelated, line 797: bikeshed: TRACK_INFO so it can contain more than the current group

fluffy (approved):
Looks like good start.

wilaw:
What about returning latest object in addition to latest group?

draft-ietf-moq-transport.md
```diff
@@ -1565,6 +1569,64 @@ ANNOUNCE_CANCEL Message {
...
+
+An endpoint sends a 'CURRENT_GROUP' message on the control stream in response
+to a GET_CURRENT_GROUP message.
+
+~~~
+CURRENT_GROUP Message {
+  Track Namespace (b),
+  Track Name (b),
+  Status Code (i),
+  Current Group ID (i),
```
suhasHere, line 1595: can we make it optional

martinduke:
OK, changes:
- I painted my bikeshed TRACK_STATUS, on the grounds that TRACK_INFO sounds like more permanent data that maybe out to be in the Catalog.
- I added last Object ID because Will asked for it and it seemed easy enough.
- I did not fill Suhas's suggestion to make it optional because the parameterized idea got a lot of pushback in the call, and this seemed adjacent.

Please have a look, and if Ian's happy with it he can merge.

draft-ietf-moq-transport.md
martinduke: As noted below, people didn't seem to like parameters, which is the most straightforward way to do this.

martinduke:
I'll also note that we could put a track alias hint in here, but I didn't care enough to add it without more positive feedback.

wilaw (approved):
This is a good start. We can massage the status codes later. They should correspond to some sort of state machine that we define for moqt in general.

suhasHere (approved):

draft-ietf-moq-transport.md
martinduke: Suhas approved, resolving issue

draft-ietf-moq-transport.md
```diff
@@ -1565,6 +1569,72 @@ ANNOUNCE_CANCEL Message {
...
+}
+~~~
+{: #moq-track-status-format title="MOQT TRACK_STATUS Message"}
+
+The 'Status Code' field provides additional information about the status of the
+track. It MUST hold one of the following values. Any other value is a malformed
+message.
+
+0x00: The track is in progress, and subsequent fields contain the highest group
+and object ID known to the sender for that track.
```
ianswett, line 1606:
```suggestion
and object ID for that track.
```

draft-ietf-moq-transport.md
```diff
@@ -1565,6 +1569,72 @@ ANNOUNCE_CANCEL Message {
...
+and object ID for that track.
+
+0x01: The track does not exist. Subsequent fields MUST be zero, and any other
+value is a malformed message.
+
+0x02: The track has not yet begun. Subsequent fields MUST be zero. Any other
+value is a malformed message.
+
+0x03: The track has finished, so there is no "live edge." Subsequent fields
+contain the highest Group and object ID known to the sender.
```
ianswett, line 1615:
```suggestion
contain the highest Group and object ID known.
```

draft-ietf-moq-transport.md
```diff
@@ -1565,6 +1569,72 @@ ANNOUNCE_CANCEL Message {
...
+
+0x02: The track has not yet begun. Subsequent fields MUST be zero. Any other
+value is a malformed message.
+
+0x03: The track has finished, so there is no "live edge." Subsequent fields
+contain the highest Group and object ID known.
+
+0x04: The sender is a relay that cannot obtain the current track status from
+upstream. Subsequent fields contain the largest group and object ID known to
+the sender.
```
ianswett, line 1619:
```suggestion
upstream. Subsequent fields contain the largest group and object ID known.
```

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1565,6 +1569,73 @@ ANNOUNCE_CANCEL Message {
...
+
+0x04: The sender is a relay that cannot obtain the current track status from
+upstream. Subsequent fields contain the largest group and object ID known.
+
+Any other value in the Status Code field is a malformed message.
+
+When a relay is subscribed to a track, it can simply return the highest group
+and object ID it has observed, whether or not that object was cached. If not
+subscribed, a relay SHOULD send a TRACK_STATUS_REQUEST upstream to obtain
+updated information.
```
ianswett, line 1627:
```suggestion
When a relay is subscribed to a track, it can simply return the highest group
and object ID it has observed, whether or not that object was cached or
completely delivered. If not subscribed, a relay SHOULD send a
TRACK_STATUS_REQUEST upstream to obtain updated information.
```

ianswett (approved):

Issue #423, opened 2024-03-28
Proposal: Gap Notifications


kixelated:
Feel free to `s/SUBSCRIBE/FETCH/g` in your head.

## Problem
The current draft is intentionally vague, but it allows objects to dropped in response to congestion. However, this ambiguity will cause issues for a few use-cases. 

Consider a subscriber that requests a range: 

```
-> SUBSCRIBE start=4 end=7
<- SUBSCRIBE_OK
<- OBJECT group=7
<- OBJECT group=6
<- OBJECT group=4
<- SUBSCRIBE_DONE
```

At this moment, group=5 has been promised as part of the SUBSCRIBE but has not arrived yet. This is perfectly normal as objects may be sent over multiple streams which arrive out of order. A reliable live (#419) and VOD player will block playback as they anticipate group=5 to arrive any moment now and they do not want to skip content.

However, if group=5 was dropped for any reason, potentially upstream, then playback will be stalled indefinitely. The only recourse is to set an arbitrary timeout and eventually assume loss, but this will cause additional buffering especially for reliable live.

Furthermore, a relay that issues this request upstream will mark group=5 in its cache as pending. If a FETCH comes in for start=2 end=5, then the relay could FETCH start=2 end=3 upstream, as it can serve group=4 immediately and group=5 once it arrives at any moment.

However, if group=5 was dropped due to a transient problem (ex. congestion), then the relay would be unaware that it could refetch group=5. Indicating that a dropped object can be retried would be very useful for patching up VODs, as the relay could refill these known holes after network recovery.

### 1. `SUBSCRIBE_OK` may contain a subset of the range requested in `SUBSCRIBE`.
The publisher commits to sending either an object or a drop notification within the specified range.

```
SUBSCRIBE_OK {
    Subscribe ID (i),
    Start Group (i)
    Start Object (i)
    End Group (i)
    End Object (i)
}
```
(some encoding finagling required)

Otherwise, a relay that receives a `SUBSCRIBE start=0 end=10000` would be expected to send an object or dropped message for every value within the requested range, even if they left cache long ago.  For example, a DVR player requesting content from 45s ago, but everything older than 30s was already dropped from the cache.

This modified `SUBSCRIBE_OK` message tightens the bounds to prevent this.

### 2. Add a `STREAM_DROPPED` that indicates drops within the `SUBSCRIBE_OK` range.
The publisher indicates that it cannot deliver the entire range as originally indicated, and one or more objects were dropped.

```
[...]

PR #426, opened 2024-04-03
END_OF_GROUP message

martinduke:
Here's a explicit signal for missing objects.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1579,7 +1579,7 @@ END_OF_GROUP Message {
   Subscribe ID (i),
   Track Alias (i),
   Group ID (i),
-  Next Object ID (i),
+  Object Count (i),
 }
 ~~~
 {: #moq-transport-end-of-group-format title="MOQT END_OF_GROUP Message"}
@@ -1589,8 +1589,8 @@ group, allowing receivers to easily determine if they are missing any objects
 from a group. END_OF_GROUP messages are not sent for groups whose subscription
 range does not include the last object of a group.
 
-The 'Next Object ID' field is one greater than the highest Object ID in the
-group. A publisher SHOULD send an END_OF_GROUP message with a Next Object ID of
+The 'Object Count' field is one greater than the highest Object ID in the
+group. A publisher MUST send an END_OF_GROUP message with a Next Object ID of
 zero when it skips a Group ID.
 
 When a track's forwading preference is 'Track', the sender interleaves
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -273,6 +273,11 @@ identical sequence of bytes regardless of how or where it is retrieved.
 An Object can become unavailable, but it's contents MUST NOT change over
 time.
 
+A publisher MAY skip Group IDs in a track. A publisher MUST NOT skip object IDs
+within a group and every group MUST start at Object ID zero. Therefore, any gap
+in object ID indicates that an object exists, or at one time existed, at the
+publisher.
+
 Objects are comprised of two parts: metadata and a payload.
 The metadata is never encrypted and is always
 visible to relays. The payload portion may be encrypted, in which case
@@ -803,6 +808,8 @@ MOQT Message {
 | 0x50  | STREAM_HEADER_TRACK ({{multi-object-streams}})      |
 |-------|-----------------------------------------------------|
 | 0x51  | STREAM_HEADER_GROUP ({{multi-object-streams}})      |
+|-------|-----------------------------------------------------|
+| 0x52  | END_OF_GROUP ({{message-end-of-group}})             |
 |-------|-----------------------------------------------------|
 
 ## Parameters {#params}
@@ -1565,7 +1572,50 @@ ANNOUNCE_CANCEL Message {
 * Track Namespace: Identifies a track's namespace as defined in
 ({{track-name}}).
 
+## END_OF_GROUP {#message-end-of-group}
+
+~~~
+END_OF_GROUP Message {
+  Subscribe ID (i),
+  Track Alias (i),
+  Group ID (i),
+  Next Object ID (i),
+}
+~~~
+{: #moq-transport-end-of-group-format title="MOQT END_OF_GROUP Message"}
+
+Senders use END_OF_GROUP messages to communicate the maximum Object ID in a
+group, allowing receivers to easily determine if they are missing any objects
+from a group. END_OF_GROUP messages are not sent for groups whose subscription
+range does not include the last object of a group.
+
+The 'Next Object ID' field is one greater than the highest Object ID in the
+group. A publisher SHOULD send an END_OF_GROUP message with a Next Object ID of
+zero when it skips a Group ID.
+
+When a track's forwading preference is 'Track', the sender interleaves
+END_OF_GROUP messages with OBJECT messages in the Track Stream. These SHOULD
+appear immediately after the last object from a group in the stream. In some use
+cases, they may not arrive in time to be relayed in that order.
+
+When a track's forwarding preference is 'Group', the END_OF_GROUP message MUST
+be the last message on the stream. If the publisher has skipped a group ID, or
+the sender is skipping delivery of a group for prioritization reasons, this
+involves opening a unidirectional stream solely to send the END_OF_GROUP
+message.
+
+When a track's forwarding preference is 'Object,' the sender MUST open a
+separate unidirectional stream per group to deliver the END_OF_GROUP message.
+
+When a track's forwarding preference is 'Datagram', END_OF_GROUP messages MUST
+NOT be sent.
+
+A sender MUST NOT send FIN on a unidirectional stream unless it has sent all
+END_OF_GROUP messages required for that stream as described above. Due to stream
+resets, a receiver may not observe an END_OF_GROUP message for all groups.
+
 ## GOAWAY {#message-goaway}
+
 The server sends a `GOAWAY` message to initiate session migration
 ({{session-migration}}) with an optional URI.
 


draft-ietf-moq-transport.md
```diff
@@ -1565,7 +1572,50 @@ ANNOUNCE_CANCEL Message {
...
+When a track's forwarding preference is 'Group', the END_OF_GROUP message MUST
+be the last message on the stream. If the publisher has skipped a group ID, or
+the sender is skipping delivery of a group for prioritization reasons, this
+involves opening a unidirectional stream solely to send the END_OF_GROUP
+message.
+
+When a track's forwarding preference is 'Object,' the sender MUST open a
+separate unidirectional stream per group to deliver the END_OF_GROUP message.
+
+When a track's forwarding preference is 'Datagram', END_OF_GROUP messages MUST
```
kixelated, line 1610: Why not?

draft-ietf-moq-transport.md
```diff
@@ -1565,7 +1572,50 @@ ANNOUNCE_CANCEL Message {
...
+NOT be sent.
+
+A sender MUST NOT send FIN on a unidirectional stream unless it has sent all
+END_OF_GROUP messages required for that stream as described above. Due to stream
+resets, a receiver may not observe an END_OF_GROUP message for all groups.
```
kixelated, line 1615: Can we eliminate this line and make this a reliable message? The point of this message is to remove the end group ambiguity due to unreliability, and we can't do that if this message itself is unreliable.

draft-ietf-moq-transport.md
```diff
@@ -1565,7 +1572,50 @@ ANNOUNCE_CANCEL Message {
...
+The 'Next Object ID' field is one greater than the highest Object ID in the
+group. A publisher SHOULD send an END_OF_GROUP message with a Next Object ID of
+zero when it skips a Group ID.
+
+When a track's forwading preference is 'Track', the sender interleaves
+END_OF_GROUP messages with OBJECT messages in the Track Stream. These SHOULD
+appear immediately after the last object from a group in the stream. In some use
+cases, they may not arrive in time to be relayed in that order.
+
+When a track's forwarding preference is 'Group', the END_OF_GROUP message MUST
```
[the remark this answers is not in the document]
kixelated: But I also like the existing behavior of using stream FIN to signal the end of a group. This END_OF_GROUP message feels like compliance, trying to address an issue that only impacts one delivery mode.

draft-ietf-moq-transport.md
```diff
@@ -1565,7 +1572,50 @@ ANNOUNCE_CANCEL Message {
...
+~~~
+{: #moq-transport-end-of-group-format title="MOQT END_OF_GROUP Message"}
+
+Senders use END_OF_GROUP messages to communicate the maximum Object ID in a
+group, allowing receivers to easily determine if they are missing any objects
+from a group. END_OF_GROUP messages are not sent for groups whose subscription
+range does not include the last object of a group.
+
+The 'Next Object ID' field is one greater than the highest Object ID in the
+group. A publisher SHOULD send an END_OF_GROUP message with a Next Object ID of
```
kixelated, line 1593: This `SHOULD` is a `MUST` imo, otherwise you'll break clients that actually use this message (ex. reliable live).

Also, this breaks being able to use the presentation timestamp as the group ID, as you would have to send a boatload of END_OF_GROUP messages. I'm all for that but I want to make sure we've discussed it.

kixelated (commented):
What's the use-case behind this message?

I think it's meant to support reliable live when using a stream per object. The player will buffer until it receives the entire group, but it can't do that if it doesn't know the full size of the group.

However, I don't think this message accomplishes that by itself since there can be gaps within the group. The player might know the intended size of the group, but any RESET_STREAM will cause it to block awaiting the full group to arrive.

I think we need the group size in conjunction with #423. Both messages are so similar that maybe there could be some overlap; I'm not sure.

kixelated:
Oh, and I'd like to only use this message when using a stream per object. It's not relevant for the other modes.

Alternatively, we could just make this a property of the mode itself. You don't know the final group size when using a stream per object, and it's probably just fine for the intended use-cases?

draft-ietf-moq-transport.md
```diff
@@ -1565,7 +1572,50 @@ ANNOUNCE_CANCEL Message {
...
+zero when it skips a Group ID.
+
+When a track's forwading preference is 'Track', the sender interleaves
+END_OF_GROUP messages with OBJECT messages in the Track Stream. These SHOULD
+appear immediately after the last object from a group in the stream. In some use
+cases, they may not arrive in time to be relayed in that order.
```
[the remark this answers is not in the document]
martinduke: Imagine a case where an MoQ publisher is receiving objects via a different transport. Could it not happen? If not, I'm happy to make it a MUST.

martinduke:
> What's the use-case behind this message?
> 
> I think it's meant to support reliable live when using a stream per object. The player will block until it receives the entire group, but it can't do that if it doesn't know the full size of the group.

I don't know use cases, but people keep saying we need a way to identify missing objects in a group.

> 
> However, I don't think this message accomplishes that by itself since there can be gaps within the group. The player might know the intended size of the group, but any RESET_STREAM will cause it to block awaiting the full group to arrive.

The first part of the PR says that publishers MUST NOT skip Object IDs. So if you're missing an Object ID, either the sender didn't get it, or it chose not to send it.

> 
> I think we need the group size in conjunction with #423. Both messages are so similar that maybe there could be some overlap; I'm not sure.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
martinduke: Q: If a sender decides not to send all of the objects in a group, can it close with a FIN, or does it have to use RESET_STREAM?

draft-ietf-moq-transport.md
martinduke: We could just have it open a stream, but I figure if you're in Datagram mode you are not concerned about reliability.

draft-ietf-moq-transport.md
martinduke: SGTM, changed. 

I don't think it "breaks" anything, but if you're concerned about message bandwidth we could allow this to apply to a range of groups or invent a similar message.

martinduke:
> Oh, and I'd like to only use this message when using a stream per object. It's not relevant for the other modes.
> 
> Alternatively, we could just make this a property of the mode itself. You don't know the final group size when using a stream per object, and it's probably just fine for the intended use-cases?

I don't think that's right. If it's a track stream, and the sender has to omit some objects, this is very much relevant.

You are presupposing some restrictions on how the various forwarding preferences can be used in conjunction with various reliability and latency requirements. If those restrictions exist, we need to write those down somewhere.

draft-ietf-moq-transport.md
martinduke: If a stream is reset, you a priori know you don't have everything. Don't reset a stream if getting all the objects is important!

martinduke:
> Oh, and I'd like to only use this message when using a stream per object. It's not relevant for the other modes.
> 
> Alternatively, we could just make this a property of the mode itself. You don't know the final group size when using a stream per object, and it's probably just fine for the intended use-cases?

But what if a track can be used in several use cases? It's stuck with one forwarding preference!

draft-ietf-moq-transport.md
```diff
@@ -1565,7 +1572,50 @@ ANNOUNCE_CANCEL Message {
...
+  Group ID (i),
+  Object Count (i),
+}
+~~~
+{: #moq-transport-end-of-group-format title="MOQT END_OF_GROUP Message"}
+
+Senders use END_OF_GROUP messages to communicate the maximum Object ID in a
+group, allowing receivers to easily determine if they are missing any objects
+from a group. END_OF_GROUP messages are not sent for groups whose subscription
+range does not include the last object of a group.
```
suhasHere, line 1590: END_OF_GROUP must be sent on the same unidirectional stream that sent the last object for the group.  This should be sent regardless of the request range/location.

draft-ietf-moq-transport.md
```diff
@@ -1565,7 +1572,50 @@ ANNOUNCE_CANCEL Message {
...
+
+The 'Object Count' field is one greater than the highest Object ID in the
+group. A publisher MUST send an END_OF_GROUP message with a Next Object ID of
+zero when it skips a Group ID.
```
suhasHere, line 1594: where is Next Object ID defined ?

draft-ietf-moq-transport.md
```diff
@@ -804,6 +809,8 @@ MOQT Message {
 |-------|-----------------------------------------------------|
 | 0x51  | STREAM_HEADER_GROUP ({{multi-object-streams}})      |
 |-------|-----------------------------------------------------|
+| 0x52  | END_OF_GROUP ({{message-end-of-group}})             |
+|-------|-----------------------------------------------------|
```
suhasHere, line 813: There was another proposal  here: https://github.com/moq-wg/moq-transport/pull/334/files

Not saying we do one vs the other. But I feel we need to agree upon requirements on
  - what such message indicate
  - when and where will it be sent,

draft-ietf-moq-transport.md
```diff
@@ -1565,7 +1572,50 @@ ANNOUNCE_CANCEL Message {
...
+cases, they may not arrive in time to be relayed in that order.
+
+When a track's forwarding preference is 'Group', the END_OF_GROUP message MUST
+be the last message on the stream. If the publisher has skipped a group ID, or
+the sender is skipping delivery of a group for prioritization reasons, this
+involves opening a unidirectional stream solely to send the END_OF_GROUP
+message.
+
+When a track's forwarding preference is 'Object,' the sender MUST open a
+separate unidirectional stream per group to deliver the END_OF_GROUP message.
```
suhasHere, line 1608: It must be sent on the same stream. not sure why it should be on a new stream

suhasHere (changes requested):
some comments.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
kixelated: Yeah, the purpose of a stream per track is to allow dumb decoders. All media is delivered in decode order so it's trivial to handle and there's no reassembly buffer required. That goes out the window if the encoder is allowed to emit out of order. It's meant to be a direct replacement for RTMP in that regard.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
kixelated: IMO `RESET_STREAM` combined with `STREAM_DROPPED` (#423). We don't want relays silently truncating streams and pretending like nothing happened.

draft-ietf-moq-transport.md
kixelated: I think we need to focus on the use-case here. A player needs know the final group size (and any gaps) for reliable playback (live and VOD) when using a stream per object. 

If this message is optional in any way then the use-case does not work. I think it needs to be mandatory to send an END_OF_GROUP message for each group (when using an object per stream).

draft-ietf-moq-transport.md
```diff
@@ -273,6 +273,11 @@ identical sequence of bytes regardless of how or where it is retrieved.
 An Object can become unavailable, but it's contents MUST NOT change over
 time.
 
+A publisher MAY skip Group IDs in a track. A publisher MUST NOT skip object IDs
+within a group and every group MUST start at Object ID zero. Therefore, any gap
+in object ID indicates that an object exists, or at one time existed, at the
+publisher.
```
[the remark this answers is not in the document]
kixelated: Yeah, I think this paragraph is awkward.

I think what you're trying to say is that there must be an `END_OF_GROUP` for each group within the subscription range, even if it's size 0.

kixelated (changes requested):
Let's identify the use-cases that needs to know the group size and make sure this meets the requirements. I don't think it does.

fluffy (changes requested):
I like the idea of a publisher being able to publish an object that indicates there will be no hither object in the group. Or not higher groups in the track. I think this should be allowed and explicit and on what I think of as the dataplane. I put in a similar proposal in #429. I like all these parts of this proposal. 

I do not like the relays needing to require that objects inside the group need to be produced in order or increment by one. I don't see that as needed for any of this to work int he relays work. Similarly I see no requirements for for oder that groups are produced or gaps between them. I can see how the catalog might indicate for a given track that the groups or object were sequential but does not seem to be a thing the relays should have to deal with. The reason I am concerned with this is moderne video codecs.

martinduke:
> I like the idea of a publisher being able to publish an object that indicates there will be no hither object in the group. Or not higher groups in the track. I think this should be allowed and explicit and on what I think of as the dataplane. I put in a similar proposal in #429. I like all these parts of this proposal.
> 
> I do not like the relays needing to require that objects inside the group need to be produced in order or increment by one. I don't see that as needed for any of this to work int he relays work. Similarly I see no requirements for for oder that groups are produced or gaps between them. I can see how the catalog might indicate for a given track that the groups or object were sequential but does not seem to be a thing the relays should have to deal with. The reason I am concerned with this is moderne video codecs.

The requirement for the publisher to not skip Object ID just makes the encoding more efficient. If you tell me there's a  use case for nonexistent object IDs, that's good enough for me, and we should probably do something closer to your PR.

PR #428, opened 2024-04-04
Delivery Preference for VOD and reliable live (and absolute start/end offsets)

kixelated:
The default behavior for live streams will be `prefer-skip` as encoded in the priority field. This doesn't work for VOD and reliable live, as they want to receive objects in creation order instead.

This is a counter-proposal to the FETCH keyword which attempts to accomplish something similar, but in a way that won't work for reliable live.

This PR also removes relative start/end offsets because SUBSCRIBE was getting too complex. An application that wants to start N groups back needs to use use #425 or a timeline to learn about the absolute group/object IDs.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1199,44 +1199,6 @@ OBJECT_STREAM {
 
 A receiver issues a SUBSCRIBE to a publisher to request a track.
 
-### Subscribe Locations {#subscribe-locations}
-
-The receiver specifies a start and optional end `Location` for the subscription.
-A location value may be an absolute group or object sequence, or it may be a
-delta relative to the largest group or the largest object in a group.
-
-~~~
-Location {
-  Mode (i),
-  [Value (i)],
-}
-~~~
-
-There are 4 modes:
-
-None (0x0): The Location is unspecified, Value is not present
-
-Absolute (0x1): Value is an absolute sequence
-
-RelativePrevious (0x2): Value is a delta from the largest sequence.  0 is the
-largest sequence, 1 is the largest sequence - 1, and so on.
-
-RelativeNext (0x3): Value is a delta from the largest sequence.  0 is the largest
-sequence + 1, 1 is the largest sequence + 2, and so on.
-
-The following table shows an example of how the RelativePrevious and RelativeNext
-values are used to determine the absolute sequence.
-
-~~~
-Sequence:                0    1    2    3    4   [5]  [6] ...
-                                             ^
-                                      Largest Sequence
-RelativePrevious Value:  4    3    2    1    0
-RelativeNext Value:                               0    1  ...
-~~~
-{: title="Relative Indexing"}
-
-
 ### SUBSCRIBE Format
 
 The format of SUBSCRIBE is as follows:
@@ -1247,10 +1209,11 @@ SUBSCRIBE Message {
   Track Alias (i),
   Track Namespace (b),
   Track Name (b),
-  StartGroup (Location),
-  StartObject (Location),
-  EndGroup (Location),
-  EndObject (Location),
+  Start Group (i),
+  [ Start Object (i), ]
+  End Group (i),
+  [ End Object (i), ]
+  Delivery Preference (i),
   Number of Parameters (i),
   Track Request Parameters (..) ...
 }
@@ -1275,24 +1238,24 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
-* StartGroup: The Location of the requested group.  StartGroup's Mode MUST NOT be
-None.
+* Start Group: The start Group ID, plus 1. A value of 0 indicates no start group and the latest group is used instead.
 
-* StartObject: The Location of the requested object.  StartObject's Mode MUST NOT
-be None.
+* Start Object: The start Object ID, plus 1. A value of 0 indicates no start object and the latest object is used instead. This field is not present when Start Group is 0.
 
-* EndGroup: The last Group requested in the subscription, inclusive.  EndGroup's
-Mode is None for an open-ended subscription.
+* End Group: The end Group ID, plus 1. A value of 0 indicates no end group and the subscription is open-ended.
 
-* EndObject: The last Object requested in the subscription, exclusive.
-EndObject's Mode MUST be None if EndGroup's Mode is None.  EndObject's Mode MUST
-NOT be None if EndGroup's Mode is not None.
+* End Object: The end Object ID, plus 1. A value of 0 indicates no end object and the entire group is requested. This field is not present when End Group is 0.
+
+* Delivery Preference: Indicates how objects should be transmitted and dropped.
+The following seciton ({{delivery-preference}}) enumerates the possible values.
 
 * Track Request Parameters: The parameters are defined in
 {{version-specific-params}}
 
 On successful subscription, the publisher SHOULD start delivering
-objects from the group ID and object ID described above.
+objects within the range described above.
+The publisher MUST reply with a SUBSCRIBE_OK to indicate the latest group/object,
+allowing the subscriber to determine the start group/object when not specified.
 
 If a publisher cannot satisfy the requested start or end for the subscription it
 MAY send a SUBSCRIBE_ERROR with code 'Invalid Range'. A publisher MUST NOT send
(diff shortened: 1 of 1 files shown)


kixelated:
This is half of #411. The other half would be to add a `priority` field to the SUBSCRIBE message itself, otherwise you can't prioritize between the tracks themselves. Note that `FETCH` would need something similar too.

We could also remove `priority` from the stream header and put `Default Delivery Preference` in `TRACK_INFO` instead. This would prevent a broadcaster from doing custom priorities, but that's probably fine?

draft-ietf-moq-transport.md
```diff
@@ -1301,58 +1264,31 @@ objects from outside the requested start and end.
...
-Start = Group 3, Object 0
-End = Group 3, Object <last>
-~~~
+### Delivery Preference
+The publisher will attempt to deliver every object within the requested range.
+However, in many situations this will be impossible.
+Congestion may cause the subscriber to fall behind until eventually the publisher hits resource limits and drops objects.
+
+The `Delivery Preference` is a hint from the subscriber indiciating which streams/datagrams should be transmitted first during congestion, and coorespondingly which streams/datagrams should be starved or dropped.
+A publisher OUGHT to respect the delivery prefernce however, especially when multiple subscribers share a cache, this may be on a best effort basis.
```
afrind, line 1273: https://datatracker.ietf.org/doc/html/rfc6919#section-4

I think you can just say "A publisher SHOULD respect the delivery preference"

draft-ietf-moq-transport.md
```diff
@@ -1301,58 +1264,31 @@ objects from outside the requested start and end.
...
+A subscriber MUST NOT expect objects to arrive in the indicated order.
+
+Note that objects are placed on QUIC streams according to the Object Forwarding Preference ({{object-message-formats})).
```
afrind, line 1276: streams or datagrams

draft-ietf-moq-transport.md
```diff
@@ -1301,58 +1264,31 @@ objects from outside the requested start and end.
...
+|---:|:--------------|
+| 0  | default       |
+|----|---------------|
+| 1  | prefer-stalls |
+|----|---------------|
+| 2  | prefer-skips  |
+|----|---------------|
+
+* `default`: The publisher transmits according to the `priority` field for each stream, as chosen by the broadcaster.
+* `prefer-stalls`: The publisher ignores the `priority` field, transmitting streams in ascending `group_id`, and then ascending `object_id` order.
```
afrind, line 1290: priority still matters within a group, when the preference is datagram or stream-per-object (eg: when group IDs are equal).

afrind (commented):
Thanks for writing the proposal.

Individual Comments

draft-ietf-moq-transport.md
kixelated: Should it? We could use the `priority` field instead of `object_id` as the `group_id` tiebreaker.

But suppose we have lower priority B-frames frames within a group. For VOD and reliable live, we want to receive and render all objects in order. These B-frames will actually block playback, so delivering them with lower priority will cause slightly more stalling.

Using `priority` instead of `object_id` makes sense for `prefer-skips` though.

draft-ietf-moq-transport.md
```diff
@@ -1247,10 +1209,11 @@ SUBSCRIBE Message {
   Track Alias (i),
   Track Namespace (b),
   Track Name (b),
-  StartGroup (Location),
-  StartObject (Location),
-  EndGroup (Location),
-  EndObject (Location),
+  Start Group (i),
+  [ Start Object (i), ]
+  End Group (i),
```
suhasHere, line 1214: I think even End Group can be optional too for open ended subscribes

draft-ietf-moq-transport.md
suhasHere, line 1273: This doesn't ensure consistent behavior when there is confliting preferences . Also as I mentioned earlier I am still unsure on how to match this upstream.

afrind: Individual Comment: 

I feel like the right answer here is to mandate relays subscribe with default (publisher preference), with an "application specific information" escape clause (eg: if you have a business arrangement with a CDN you can do whatever you want).  Presumably publishers know whether they want to optimize for realtime or reliable live viewers.

suhasHere: "mandate relays" is not in the scope of moqt. Also if we mandated something like that,  subscriber that is asking opposite behavior will experience the negative effects of such a choice.  



> (eg: if you have a business arrangement with a CDN you can do whatever you want

This is not exactly related to business agreement. 

On a diff note, I am not seeing why we have ASC/DESC like option. Groups are always published in the ascending order as @wilaw  mentioned.

kixelated: > "mandate relays" is not in the scope of moqt. Also if we mandated something like that, subscriber that is asking opposite behavior will experience the negative effects of such a choice.

This is a relay-specifc problem. Specifically, when one upstream subscribe is deduplicated into multiple downstream subscriptions with different preferences. 

I like the relay always asking for `default` as it punts the responsibility to the broadcaster/application. It's much like send order today: a broadcaster could push groups in descending order (via SUBSCRIBE) while the viewer pulls in ascending order (via FETCH). There's no getting around the fact that only one copy of the media will arrive and the order won't always be ideal.

suhasHere: I feel  we are trying to mix 2 different things here
 - Priority interaction
 - Client preference to wait or to continue without waiting
 
Groups in ascending vs descending order seems to be more of priority relayed.  Client preference can be dealt separately than the order discussion IIUC.

Also something like client preference if enabled has to be explicitly specified as last-mile only with an exception saying the client might not get what it is asking for probably ?

draft-ietf-moq-transport.md
```diff
@@ -1275,24 +1238,24 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
-* StartGroup: The Location of the requested group.  StartGroup's Mode MUST NOT be
-None.
+* Start Group: The start Group ID, plus 1. A value of 0 means the latest group is used instead.
```
wilaw, line 1241: Should clarify that its the latest group in an active subscription, and not the latest group that the relay happens to have in cache (which may be from 5 min ago).

draft-ietf-moq-transport.md
```diff
@@ -1275,24 +1238,24 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
...
+
+* Start Object: The start Object ID, plus 1. A value of 0 means the latest object is used instead. This field is not present when Start Group is 0.
```
wilaw, line 1243: The concern with coding start and end group/object using +1 is you refer to a non-existent entity. For example, if my content has groups 5, 10, 15, 20 and I want all of it, I need to subscribe to 6-21 to get all of it. Neither of those groups actually exist and semantically this feels wrong. What they really represent are non-inclusive threshold value, so that they relay can apply the test:

`forward if lower-threshold < groupID < upper-threshold. `

**Alternatively, can we shift to an inclusive scheme, where we actually specify the real group and object IDs that we want to begin and end with?** I that case, I would subscribe to 15 - 20. That is much cleaner. The challenge is in how do specify give me 'the latest group' and also an unspecified end group and/or end object? **A solution there is require Group and Object numbers to be > 0**. That frees up 0 to convey the following special cases:
```
StartGroup  = 0 give me the latest group
StartObject = 0, give me the latest object
EndGroup = 0, open-ended subscription, give me all future groups until EOT is signalled. 
EndObject = 0, open-ended subscription, give me all future objects in the EndGroup. 
```

draft-ietf-moq-transport.md
suhasHere: so just to be clear, none of this seems normative IIUC. A publisher implementation can always decide not to respect the hint, if needed ??

suhasHere: > This is a relay-specifc problem. Specifically, when one upstream subscribe is deduplicated into multiple downstream subscriptions with different preferences.

May be we need to specify that to some extent in the relay section too, probably saying a given relay implementation may decide not to aggregate subscriptions  altogether when there are conflicting delivery preferences OR relay may choose to pick the delivery preference implicitly whatever  it may choose when it prefers scalability OR relay may return error under conflicting subscriptions OR something else.

suhasHere:
How can a subscriber indicate upstream  (say a relay ) to fill in existing gaps before responding vs just feed me the data it has and as they come in

draft-ietf-moq-transport.md
```diff
@@ -1247,10 +1209,11 @@ SUBSCRIBE Message {
   Track Alias (i),
   Track Namespace (b),
   Track Name (b),
-  StartGroup (Location),
-  StartObject (Location),
-  EndGroup (Location),
-  EndObject (Location),
```
suhasHere, line 1253: +1 on removing the offsets . This makes reasoning very easy .

fluffy (changes requested):
I think this PR combines 3 differnt things that should all be seperated to their own PR. 

1) A way to specify the start and end of a subscribe that uses absolute group/object numbers or start are current group  /  open ended.   This seems like a reasonable proposal to me and I would support that.  It does seem like a way to request the next group is useful because the current group will not be decodable when a clients joins part way through the group. 

2) Stuff about what object a relay forwards next. This is the priority discussion. I feel very strongly this should wait till the interim meeting and have a bunch of discussion. It could easily be removed from the PR. I am strongly against merging this PR with this stuff in it - i'm not arguing against whatever is being proposed, just this is the wrong time to solve it. We need to get the other stuff done first. 

3) Blurs fetch and subscribe. I think we need the separate semantics of this based because what we expect the relay to do is different in the cases and we need some way of the relays knowing that. I don't care if that is a flag in the subscribe or two different messages or whatever but the relay needs to know which behavior.

draft-ietf-moq-transport.md
ianswett: It would be nice if 0 isn't special, but this works.

ianswett (commented):
This looks workable, though some clarifications:
 1) Subscriptions can overlap, and we'll just deliver the same data twice, correct?
 2) If 1, I think we need to remove track alias.

The 2nd and 3rd modes are fairly clear, but I don't know what default does in a number of cases.

kixelated:
Replaced partially by #432 

I'll make a separate PR for the prioritization stuff.

PR #429, opened 2024-04-10
Add object status used to indicate lost or nonexistent objects

### LINKED ISSUE #318: Impossible to know group FIN

### LINKED ISSUE #427: Group IDs and gaps
As far as I understand, we currently allow group IDs to have gaps in them, with only requirement for them is that they are monotonically increasing.  This presents a problem with caching.  Assume I am a relay, and I receive a range request for `[10, 20)`; if the origin has never produced group 15, then even if I have the entirety of that range in cache, I would not know that, since 15 would be just missing, and the only thing I can do with that is to request 15 from the origin every time.

Possible solutions to this problem:
1. Add some mechanism to indicate that 15 is not there (e.g. by putting the number of previous group in the beginning of every group).
2. Require group numbers to be incrementing by one.

I am fan of approach number 2 since it's conceptually simpler, but I think we can make some version of 1 work.

fluffy:
This PR adds a status field to the object that allows an relay to indicate an object or group was lost or dropped. 
 
 It also allows a producer to indicate a object or group will not be produced including the cases for end of group and end of track. 
 
 The PR probably needs a bit more text on what what producers and relays do but I think this is close enough to get some initial discussion on this solution. 

Fixes #318
Fixes #427
Fixes most of #423 

Closes #334
Closes #426

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1023,7 +1023,7 @@ session with a Protocol Violation ({{session-termination}}).
 
 There are times where some of the status information is redundant with
 information learned at the lower layer. For example, when using stream
-per object, the end of Group is redundant.
+per group, the end of Group is redundant.
 
 In most cases messages with a non zero status code are sent on the same
 stream that an object with that GroupID would have been sent on. The
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -985,7 +985,7 @@ be sent according to its `Object Forwarding Preference`, described below.
 * Object Payload: An opaque payload intended for the consumer and SHOULD
 NOT be processed by a relay.
 
-* Object Status: As enumeration used to indicated missing
+* Object Status: As enumeration used to indicate missing
   objects and as an marker of the end of group or track.
 
 The Object Status lets subscribers understand what objects will not be received
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1030,15 +1030,13 @@ status could no longer optional and only occur when the payload length
 was zero as it would be needed in non zero length packets. It would also
 have the issue that when a P frame is produced, the software getting
 data from the decoder often does not know if it is the last P frame
-until the next frame is encoded and the sotware gets an I frame. We do
-not want a situttion where we add a whole frame of latency on the
-encoder waiting for next frame so it can set the end of group field on
-the last last object in the group.
-
+until the next frame is encoded and the software gets an I frame. We do
+not want a solution where we add a whole frame of latency on the encoder
+waiting for next frame so it can set the end of group field on the last
+last object in the group.
 
 Any other value SHOULD be treated as a protocol error and terminate the
 session with a Protocol Violation ({{session-termination}}).
-
 
 ### Object Message Formats
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1020,6 +1020,7 @@ Every object has an associated
 
 Any other value SHOULD be treated as a protocol error and terminate the
 session with a Protocol Violation ({{session-termination}}).
+Any object with a status code other than zero will have an empty payload.
 
 There are times where some of the status information is redundant with
 information learned at the lower layer. For example, when using stream
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -982,19 +982,17 @@ group.
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
-* Object Status: As enumeration used to indicated missing
-  objects and as an marker of the end of group or track.
-
 * Object Payload: An opaque payload intended for the consumer and SHOULD
 NOT be processed by a relay.
 
-There is also a status field which helps subscribers understand what
-objects it will not receive due to then being dropped, lost, or never be
-produced including the case where they will not be produced because they
-are beyond the end of a group or track.
-The status code can also be used to indicated the end of a group or
-track.
-Every object has an associated
+* Object Status: As enumeration used to indicated missing
+  objects and as an marker of the end of group or track.
+
+The Object Status lets subscribers understand what objects will not be received
+because they were never produced, are no longer available, or because they
+are beyond the end of a group or track. The status code can also be used to
+indicate the end of a group or track.
+
 `Status` that can have following values:
 
 * 0x0 := Normal object. The payload is array of bytes and can by empty.
@@ -1026,7 +1024,7 @@ There are times where some of the status information is redundant with
 information learned at the lower layer. For example, when using stream
 per group, the end of Group is redundant.
 
-In most cases messages with a non zero status code are sent on the same
+In most cases, messages with a non zero status code are sent on the same
 stream that an object with that GroupID would have been sent on. The
 exception to this is when that stream has been reset; in that case they
 are sent on a new stream. This is to avoid the status message being lost
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1038,6 +1038,10 @@ last object in the group.
 Any other value SHOULD be treated as a protocol error and terminate the
 session with a Protocol Violation ({{session-termination}}).
 
+There are times where some of the status information is redundant with
+information learned at the lower layer. For example, when using stream
+per object, the end of Group is redundant.
+
 ### Object Message Formats
 
 Every Track has a single 'Object Forwarding Preference' and publishers
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1015,15 +1015,6 @@ are beyond the end of a group or track. Every object has an associated
          group. This is sent right after the last object in the
          track. This SHOULD be cached.
 
-* 0x5 := Indicates Object was dropped at a relay. Sent by a relay
-         indicating the object identified by the ObjectId was
-         dropped. This SHOULD NOT be cached.
-
-* 0x6 := Indicates Group was dropped at a relay. Sent by a relay
-         indicating some of the objects in the group identified by the
-         GroupId were dropped. The Object ID is one greater than last
-         object sent and MAY be zero. This SHOULD NOT be cached.
-
 Open Issue: We could make end of track and end of group just be a status
 set on the last object sent on the group or track. This would mean the
 status could no longer optional and only occur when the payload length
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -988,7 +988,7 @@ NOT be processed by a relay.
 * Object Status: As enumeration used to indicate missing
   objects or mark the end of a group or track.
 
-The Object Status lets subscribers understand what objects will not be received
+The Object Status informs subscribers what objects will not be received
 because they were never produced, are no longer available, or because they
 are beyond the end of a group or track.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -975,6 +975,8 @@ A canonical MoQ Object has the following information:
 IDs starts at 0, increasing sequentially for each object within the
 group.
 
+* Object Send Order: An integer indicating the object send order
+{{send-order}} or priority {{ordering-by-priorities}} value.
 
 * Object Forwarding Preference: An enumeration indicating how a sender sends an
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -994,7 +994,7 @@ are beyond the end of a group or track.
 
 `Status` that can have following values:
 
-* 0x0 := Normal object. The payload is array of bytes and can by empty.
+* 0x0 := Normal object. The payload is array of bytes and can be empty.
 
 * 0x1 := Indicates Object does not exist. Indicates that this object
          does not exist at any publisher and it will not be published in
@@ -1017,11 +1017,10 @@ are beyond the end of a group or track.
 
 Any other value SHOULD be treated as a protocol error and terminate the
 session with a Protocol Violation ({{session-termination}}).
-Any object with a status code other than zero will have an empty payload.
+Any object with a status code other than zero MUST have an empty payload.
 
-There are times where some of the status information is redundant with
-information learned at the lower layer. For example, when using stream
-per group, the end of Group is redundant.
+Though some status information could be inferred from QUIC stream state,
+that information is not reliable and cacheable.
 
 In most cases, messages with a non zero status code are sent on the same
 stream that an object with that GroupID would have been sent on. The
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1028,7 +1028,9 @@ are beyond the end of a group or track. Every object has an associated
          stream. If the group is dropped there will only one message
          with this status for the group.
 
-Any other value SHOULD be treated as a protcol error.
+Any other value SHOULD be treated as a protcol error and terminate the
+session with a Protocol Violation ({{session-termination}}).
+
 
 ### Object Message Formats
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -986,7 +986,7 @@ be sent according to its `Object Forwarding Preference`, described below.
 NOT be processed by a relay.
 
 * Object Status: As enumeration used to indicate missing
-  objects and as an marker of the end of group or track.
+  objects or mark the end of a group or track.
 
 The Object Status lets subscribers understand what objects will not be received
 because they were never produced, are no longer available, or because they
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1042,6 +1042,13 @@ There are times where some of the status information is redundant with
 information learned at the lower layer. For example, when using stream
 per object, the end of Group is redundant.
 
+In most cases messages with a non zero status code are sent on the same
+stream that an object with that GroupID would have been sent on. The
+exception to this is when that stream has been reset; in that case they
+are sent on a new stream. This is to avoid the status message being lost
+in cases such as a relay dropping a group and reseting the stream the
+group is being sent on.
+
 ### Object Message Formats
 
 Every Track has a single 'Object Forwarding Preference' and publishers
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -982,16 +982,19 @@ group.
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
-* Object Status: As enumeration used for missing or dropped objects that
-  indicates why data is missing.
+* Object Status: As enumeration used to indicated missing or dropped
+  objects and as an marker of the end of group or track.
 
 * Object Payload: An opaque payload intended for the consumer and SHOULD
 NOT be processed by a relay.
 
-There is also a status field which help subscribers understand what
+There is also a status field which helps subscribers understand what
 objects it will not receive due to then being dropped, lost, or never be
 produced including the case where they will not be produced because they
-are beyond the end of a group or track. Every object has an associated
+are beyond the end of a group or track.
+The status code can also be used to indicated the end of a group or
+track.
+Every object has an associated
 `Status` that can have following values:
 
 * 0x0 := Normal object. The payload is array of bytes and can by empty.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1028,6 +1028,7 @@ are beyond the end of a group or track. Every object has an associated
          stream. If the group is dropped there will only one message
          with this status for the group.
 
+Any other value SHOULD be treated as a protcol error.
 
 ### Object Message Formats
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -975,15 +975,57 @@ A canonical MoQ Object has the following information:
 IDs starts at 0, increasing sequentially for each object within the
 group.
 
-* Object Send Order: An integer indicating the object send order
-{{send-order}} or priority {{ordering-by-priorities}} value.
 
 * Object Forwarding Preference: An enumeration indicating how a sender sends an
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
+* Object Status: As enumeration used for missing or dropped ojects that
+  indicates why data is missing.
+
 * Object Payload: An opaque payload intended for the consumer and SHOULD
 NOT be processed by a relay.
+
+There is also a status field which help subscribers understand what
+objects it will not receive due to then being dropped, lost, or never be
+produced including the case where they will not be produced because they
+are beyond the end of a group or track. Every object has an associated
+`Status` that can have following values:
+
+* 0x0 := Object exists at the publisher and payload is non-null array of
+         bytes.
+
+* 0x1 := Object non existent at the publisher and has zero sized
+         payload. Sent by the media producer and sent to indicate that
+         the given object identified by the Object Id doesn't exist for
+         the reasons specific to the producer.
+
+* 0x2 := Group non existent. Sent by a publisher and indicates the group
+         identified by the groupId doesn't exist at the producer. This
+         is the only message sent on the unidirectional stream when
+         using Stream per group and Stream per object forwarding
+         preference.
+
+* 0x3 := Object dropped at the relay. Sent by a relay indicating the
+         object identified by the ObjectId was dropped. The payload will
+         be empty in this case.
+
+* 0x4 := Group dropped at the relay. Sent by a relay indicating the
+         group identified by the GroupId was dropped. The payload will
+         be empty in this case.
+
+* 0x5 := Indicates end of group. ObjectId identifies the largest object
+         produced in the group identified by the GroupID. This is sent
+         right after the last object in the group and on the same
+         stream. If the group is dropped there will only one message
+         with this status for the group.
+
+* 0x6 := Indicates end of track. GroupID and ObjectId identifies the
+         largest group and object produced in this track. This is sent
+         right after the last object in the track and on the same
+         stream. If the group is dropped there will only one message
+         with this status for the group.
+
 
 ### Object Message Formats
 
@@ -1012,6 +1054,7 @@ OBJECT_STREAM Message {
   Group ID (i),
   Object ID (i),
   Object Send Order (i),
+  Object Status (i),
   Object Payload (..),
 }
 ~~~
@@ -1048,6 +1091,7 @@ OBJECT_DATAGRAM Message {
   Group ID (i),
   Object ID (i),
   Object Send Order (i),
+  Object Status (i),
   Object Payload (..),
 }
 ~~~
@@ -1094,6 +1138,7 @@ stream that is associated with the subscription, or open a new one and send the
 {
   Group ID (i),
   Object ID (i),
+  Object Status (i),
   Object Payload Length (i),
   Object Payload (..),
 }
@@ -1132,6 +1177,7 @@ then serialize the following fields.
 ~~~
 {
   Object ID (i),
+  Object Status (i),
   Object Payload Length (i),
   Object Payload (..),
 }
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -982,11 +982,13 @@ group.
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
-* Object Payload: An opaque payload intended for the consumer and SHOULD
-NOT be processed by a relay.
-
 * Object Status: As enumeration used to indicate missing
-  objects or mark the end of a group or track.
+objects or mark the end of a group or track. See {{object-status}} below.
+
+* Object Payload: An opaque payload intended for the consumer and SHOULD
+NOT be processed by a relay. Only present when 'Object Status' is Normal (0x0).
+
+#### Object Status {#object-status}
 
 The Object Status informs subscribers what objects will not be received
 because they were never produced, are no longer available, or because they
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1136,13 +1136,14 @@ Forwarding Preference` = `Track`.
 To send an Object with `Object Forwarding Preference` = `Track`, find the open
 stream that is associated with the subscription, or open a new one and send the
 `STREAM_HEADER_TRACK` if needed, then serialize the following object fields.
+The Object Status field is only sent if the Object Payload Length is zero.
 
 ~~~
 {
   Group ID (i),
   Object ID (i),
-  Object Status (i),
   Object Payload Length (i),
+  [Object Status (i)],
   Object Payload (..),
 }
 ~~~
@@ -1176,12 +1177,13 @@ To send an Object with `Object Forwarding Preference` = `Group`, find the open
 stream that is associated with the subscription, `Group ID` and `Object
 Send Order`, or open a new one and send the `STREAM_HEADER_GROUP` if needed,
 then serialize the following fields.
+The Object Status field is only sent if the Object Payload Length is zero.
 
 ~~~
 {
   Object ID (i),
-  Object Status (i),
   Object Payload Length (i),
+  [Object Status (i)],
   Object Payload (..),
 }
 ~~~
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -994,41 +994,49 @@ produced including the case where they will not be produced because they
 are beyond the end of a group or track. Every object has an associated
 `Status` that can have following values:
 
-* 0x0 := Object exists at the publisher and payload is non-null array of
-         bytes.
+* 0x0 := Normal object. The payload is array of bytes and can by empty.
 
-* 0x1 := Object non existent at the publisher and has zero sized
-         payload. Sent by the media producer and sent to indicate that
-         the given object identified by the Object Id doesn't exist for
-         the reasons specific to the producer.
+* 0x1 := Indicates Object does not exist. Indicates that this object
+         does not exist at any publisher and it will not be published in
+         the future. This SHOULD be cached.
 
-* 0x2 := Group non existent. Sent by a publisher and indicates the group
-         identified by the groupId doesn't exist at the producer. This
-         is the only message sent on the unidirectional stream when
-         using Stream per group and Stream per object forwarding
-         preference.
+* 0x2 := Indicates Group does not exist. Indicates that objects with
+         this GroupID do not exist at any publisher and they will not be
+         published in the future. This SHOULD be cached.
 
-* 0x3 := Object dropped at the relay. Sent by a relay indicating the
-         object identified by the ObjectId was dropped. The payload will
-         be empty in this case.
+* 0x3 := Indicates end of Group. ObjectId is one greater that the
+         largest object produced in the group identified by the
+         GroupID. This is sent right after the last object in the
+         group. This SHOULD be cached.
 
-* 0x4 := Group dropped at the relay. Sent by a relay indicating the
-         group identified by the GroupId was dropped. The payload will
-         be empty in this case.
+* 0x4 := Indicates end of Track and Group. GroupID is one greater than
+         the largest group produced in this track and the ObjectId is
+         one greater than the largest object produced in that
+         group. This is sent right after the last object in the
+         track. This SHOULD be cached.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
+There is also a status field which help subscribers understand what
+objects it will not receive due to then being dropped, lost, or never be
+produced including the case where they will not be produced because they
+are beyond the end of a group or track. Every object has an associated
+`Status` that can have following values:
+
+* 0x0 := Object exists at the publisher and payload is non-null array of
+         bytes.
+
+* 0x1 := Object non existent at the publisher and has zero sized
```
ianswett, line 998: Should we reference caching here?  It seems like both 1 and 2 are cacheable, because they won't change, but 3 and 4 are not?

ianswett (approved):
This feels like it solves a number of problems in a good way.

One quirk is that the Object status can change, but I believe the other object properties, like forwarding preference, do not change.  I don't have a problem with that, but it might be worth calling out?

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
+         stream. If the group is dropped there will only one message
+         with this status for the group.
+
+* 0x6 := Indicates end of track. GroupID and ObjectId identifies the
+         largest group and object produced in this track. This is sent
+         right after the last object in the track and on the same
+         stream. If the group is dropped there will only one message
+         with this status for the group.
+
+
```
martinduke, line 1029: "Any other value for status is a malformed message"

martinduke (approved):
I think this fine. Like you, at this point I don't really care much how we spell it.

Compared with my proposal in #426, this is conceptually way simpler (no special rules for each forwarding preference) but is less byte-efficient. I'm happy to go with whatever the group decides.

ianswett:
> I think this fine. Like you, at this point I don't really care much how we spell it.
> 
> Compared with my proposal in #426, this is conceptually way simpler (no special rules for each forwarding preference) but is less byte-efficient. I'm happy to go with whatever the group decides.

I think I'm happy to use a byte for this.  We can always optimize it later in a number of ways if we want.

suhasHere (approved):
This looks fine to me. Like the simplicity of the proposal
   - Data plane handles objects as usual regardless of their status when caching
   - Fetch like api, can reach to to status of the object/group to decide if they need to go upstream 
   This will enable some of the requirements of reliable live use-cases where the player needs to know if the producer dropped a given group/object.

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
-* Object Send Order: An integer indicating the object send order
-{{send-order}} or priority {{ordering-by-priorities}} value.
 
 * Object Forwarding Preference: An enumeration indicating how a sender sends an
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
+* Object Status: As enumeration used for missing or dropped ojects that
+  indicates why data is missing.
+
```
suhasHere, line 985: I think this needs to added in the canonical object definition as well. As this will be along with the object in the cache.

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
 
-* Object Send Order: An integer indicating the object send order
-{{send-order}} or priority {{ordering-by-priorities}} value.
 
 * Object Forwarding Preference: An enumeration indicating how a sender sends an
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
+* Object Status: As enumeration used for missing or dropped ojects that
+  indicates why data is missing.
```
suhasHere, line 984: s/ojects/objects.

draft-ietf-moq-transport.md
suhasHere: I think 3/4 are also cacheable esp when serving chain of relays. A downstream relay, if it has a knowledge of object status (as expired), can satisfy fetch request by requesting upstream.

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
+
+* 0x2 := Group non existent. Sent by a publisher and indicates the group
+         identified by the groupId doesn't exist at the producer. This
+         is the only message sent on the unidirectional stream when
+         using Stream per group and Stream per object forwarding
+         preference.
+
+* 0x3 := Object dropped at the relay. Sent by a relay indicating the
+         object identified by the ObjectId was dropped. The payload will
+         be empty in this case.
```
wilaw, line 1011: We're signaling that we dropped an object by creating another object and sending that in its place. Is the belief that the removal of the payload will always ensure that this "message object" gets through, or need we deal with a condition in which this message object is delayed to the point that it is dropped too? At that point, do we send a placeholder object for the placeholder object for the real object?

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
-{{send-order}} or priority {{ordering-by-priorities}} value.
 
 * Object Forwarding Preference: An enumeration indicating how a sender sends an
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
+* Object Status: As enumeration used for missing or dropped ojects that
+  indicates why data is missing.
+
 * Object Payload: An opaque payload intended for the consumer and SHOULD
```
wilaw, line 986: Since most objects are not dropped or missing, the vast majority of the time we will be embedding a 0x0 byte for no reason. Since in all identified dropped or issuing situations the payload is 0, can we structure the header such that if a payload length of 0 is declared, then the next field carries the Object Status instead of the payload?

draft-ietf-moq-transport.md
vasilvv: I think we should either declare those non-cacheable, or have a flag in FETCH that says "refetch everything dropped at a relay", otherwise DVR/clipping use cases might not work.

ianswett:
> > I think this fine. Like you, at this point I don't really care much how we spell it.
> > Compared with my proposal in #426, this is conceptually way simpler (no special rules for each forwarding preference) but is less byte-efficient. I'm happy to go with whatever the group decides.
> 
> I think I'm happy to use a byte for this. We can always optimize it later in a number of ways if we want.

We can actually save a byte by omitting Object ID except for the first Object in the stream, because this allows the Objects to be sequential.  Similarly, we can omit GroupID if the delivery preference is track per stream.

draft-ietf-moq-transport.md
suhasHere: if this goes along with cached object and fetch happens for a object that has status other than 0x0, relay can issue fetch upstream.  Having it stored along with object be useful, isn't it ?

vasilvv (approved):
This is a dramatic improvement over the status quo, I think we should merge this without bikeshedding too much about the encoding for now.

draft-ietf-moq-transport.md
fluffy: I like that idea. For the control messages, I don't care about saving a byte but for the audio data, it makes a big difference.

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
+         preference.
+
+* 0x3 := Object dropped at the relay. Sent by a relay indicating the
+         object identified by the ObjectId was dropped. The payload will
+         be empty in this case.
+
+* 0x4 := Group dropped at the relay. Sent by a relay indicating the
+         group identified by the GroupId was dropped. The payload will
+         be empty in this case.
+
```
afrind, line 1016: The interaction with resetting the streams that carry these messages needs to be clarified.  Let's assume stream-per-group, and the sender has attempted to send several objects, but not all data is acked.  It then wants to drop the group, and cease retransmission of not-needed objects.  However, if it needs to write the drop marker into the same stream, it needs to continue retransmitting the objects in the group.

Also, we should specify what Object ID to use here in the case there is not a specific object ID when the relay wants to, say, tail drop the group.

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
+* 0x5 := Indicates end of group. ObjectId identifies the largest object
+         produced in the group identified by the GroupID. This is sent
+         right after the last object in the group and on the same
+         stream. If the group is dropped there will only one message
+         with this status for the group.
```
afrind, line 1021: It's slightly awkward to use this wire encoding, since I might know object 10 is the last object in the group, and I send:

```
OBJECT g=x, id=10, status=0, payload
OBJECT g=x, id=10, status=END_OF_GROUP
```

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
+
+* 0x6 := Indicates end of track. GroupID and ObjectId identifies the
+         largest group and object produced in this track. This is sent
```
afrind, line 1024: Does the end-of-track imply end-of-group?

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
+         right after the last object in the track and on the same
+         stream. If the group is dropped there will only one message
+         with this status for the group.
```
afrind, line 1027: I think this is copypasta from above?

draft-ietf-moq-transport.md
afrind: This is an important message and we need to clarify the stream reset semantics here as well.

afrind (commented):
Individual Review:

Most of the statuses are not mutually exclusive (eg: an object can be available at the publisher, the end of a group, the end of track and dropped by the relay), though some are (an object can't be the end of track *without* being the end of a group, I think?).  I wonder if there's a different spelling that captures this?

The dropped/end/never-existed signals seem more important than other objects and we may need to be more specific about reliability properties and/or stream placement.

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
 * Object Payload: An opaque payload intended for the consumer and SHOULD
 NOT be processed by a relay.
 
+There is also a status field which help subscribers understand what
+objects it will not receive due to then being dropped, lost, or never be
+produced including the case where they will not be produced because they
+are beyond the end of a group or track. Every object has an associated
+`Status` that can have following values:
+
+* 0x0 := Object exists at the publisher and payload is non-null array of
```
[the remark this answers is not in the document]
suhasHere: i wonder if it helps to not have status 0x0 and having payload length of non-zero for the object in the cache is a proof of existence ??

We definitely need a way to mark if a publisher drops an object or decides not to produce at all ... which i believe are covered by other status values

draft-ietf-moq-transport.md
ianswett: It would be nice to be able to send:

`OBJECT` g=x, id=10, status=END_OF_GROUP, payload

ianswett:
> Individual Review:
> 
> Most of the statuses are not mutually exclusive (eg: an object can be available at the publisher, the end of a group, the end of track and dropped by the relay), though some are (an object can't be the end of track _without_ being the end of a group, I think?). I wonder if there's a different spelling that captures this?
> 
> The dropped/end/never-existed signals seem more important than other objects and we may need to be more specific about reliability properties and/or stream placement.

This is a good point.  Conceptually we need two enums: One to indicate if it's a Group Fin or Track Fin and one to indicate why there's no payload.

There are a lot of ways to encode that, but it's a bit nice, because the first enum is an immutable property of the Object and the latter is sort of a payload alternative.

draft-ietf-moq-transport.md
afrind: It's also possible the publisher doesn't know that object 10 is the end of the group when it's published.

kixelated:
I've only had a chance to glance, but how does this interact with RESET_STREAM?

Putting drop notifications within the OBJECT itself means that an object can only be dropped prior to being written to a stream. This is okay when used in conjunction with stream priorities, but it's not great. Ideally we can renege after committing data to a stream by resetting it.

IMO there should be a message indicating which objects were dropped upon a RESET_STREAM.

draft-ietf-moq-transport.md
fluffy: oops

fluffy:
Notes to myself:

Need text on reset. 
- could send object group drop on new stream 
- open an issue if we need this it 

Way to deal with sending zero length object 
- add an enum for this object is empty 

Way to indicate end track and I dropped it.
- perhaps better way than sending same object twice

draft-ietf-moq-transport.md
suhasHere: My proposal is to remove status 0x0 .. since its redundant when payload length is > 0.
If we do that, then above will be simplified to following for marking end of the group
```
OBJECT g=x, id=10, status=END_OF_GROUP
```
where object id 10 is the largest group.

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
 IDs starts at 0, increasing sequentially for each object within the
 group.
 
-* Object Send Order: An integer indicating the object send order
-{{send-order}} or priority {{ordering-by-priorities}} value.
```
fluffy, line 979: Don't remove send order

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
 
 * Object Forwarding Preference: An enumeration indicating how a sender sends an
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
+* Object Status: As enumeration used for missing or dropped objects that
```
kixelated, line 983: IMO this status would only be sent in `OBJECT_DROPPED` on a separate stream,

draft-ietf-moq-transport.md
fluffy: @kixelated - I will fix this not to remove it.

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
+
+* 0x3 := Object dropped at the relay. Sent by a relay indicating the
+         object identified by the ObjectId was dropped. The payload will
+         be empty in this case.
+
+* 0x4 := Group dropped at the relay. Sent by a relay indicating the
+         group identified by the GroupId was dropped. The payload will
+         be empty in this case.
+
+* 0x5 := Indicates end of group. ObjectId identifies the largest object
```
kixelated, line 1017: IMO this should be a separate flag/message. `TRACK_FIN` and/or `GROUP_FIN` maybe.

draft-ietf-moq-transport.md
kixelated: IMO this should be a separate flag. `OBJECT_FIN` perhaps. 

Otherwise yeah it's confusing to sometimes receive a second OBJECT on a stream. Also it currently can't be encoded because OBJECT doesn't contain an explicit size (when per-stream).

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
+         the given object identified by the Object Id doesn't exist for
+         the reasons specific to the producer.
+
+* 0x2 := Group non existent. Sent by a publisher and indicates the group
+         identified by the groupId doesn't exist at the producer. This
+         is the only message sent on the unidirectional stream when
+         using Stream per group and Stream per object forwarding
+         preference.
+
+* 0x3 := Object dropped at the relay. Sent by a relay indicating the
```
kixelated, line 1009: Do we really care which hop dropped the object? A subscriber really only needs to know if it can retry.

There's currently no way of signaling that a publisher dropped an object due to flow control, but it's retryable.

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
+objects it will not receive due to then being dropped, lost, or never be
+produced including the case where they will not be produced because they
+are beyond the end of a group or track. Every object has an associated
+`Status` that can have following values:
+
+* 0x0 := Object exists at the publisher and payload is non-null array of
+         bytes.
+
+* 0x1 := Object non existent at the publisher and has zero sized
+         payload. Sent by the media producer and sent to indicate that
```
kixelated, line 999: What's the difference between status code = 0x1 and a payload of size 0? Do we need this status code at all?

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
kixelated: "object exists" is very strange.

Doesn't this status code just mean that the message contains a payload? Can we just state that?

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
+         the given object identified by the Object Id doesn't exist for
+         the reasons specific to the producer.
+
+* 0x2 := Group non existent. Sent by a publisher and indicates the group
```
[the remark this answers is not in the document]
kixelated: Why not send a group with a single (FIN) object of size 0?

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
+         identified by the groupId doesn't exist at the producer. This
+         is the only message sent on the unidirectional stream when
+         using Stream per group and Stream per object forwarding
+         preference.
```
kixelated, line 1007: (also the payload is empty in this case)

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
...
+are beyond the end of a group or track. Every object has an associated
+`Status` that can have following values:
+
+* 0x0 := Object exists at the publisher and payload is non-null array of
+         bytes.
+
+* 0x1 := Object non existent at the publisher and has zero sized
+         payload. Sent by the media producer and sent to indicate that
+         the given object identified by the Object Id doesn't exist for
+         the reasons specific to the producer.
```
kixelated, line 1001: (also the payload is empty in this case)

draft-ietf-moq-transport.md
suhasHere: end of track must imply end of the last group in the track

suhasHere:
> Notes to myself:
> 
> Need text on reset.
> 
> * could send object group drop on new stream
> * open an issue if we need this it
> 
> Way to deal with sending zero length object
> 
> * add an enum for this object is empty
> 
> Way to indicate end track and I dropped it.
> 
> * perhaps better way than sending same object twice

@fluffy  i did leave the following comment https://github.com/moq-wg/moq-transport/pull/429#discussion_r1569172957

draft-ietf-moq-transport.md
kixelated: But I do like using a second message, because it means you don't have to know if it's the last object prior to transmitting it. I would just put it on a separate (reliable) stream like drop messages.

For example, encoders spit out P-frames and we can immediately send it as an OBJECT. Only when an encoder crates an I-frame (not predictable) then we can end the current group. There's no latency added.

An alternative is to have the object contains a "last of group/track" flag, but 
that requires buffering 1 frame and adds additional latency.

draft-ietf-moq-transport.md
```diff
@@ -1094,6 +1138,7 @@ stream that is associated with the subscription, or open a new one and send the
 {
   Group ID (i),
   Object ID (i),
+  Object Status (i),
```
kixelated, line 1141: A lot of these end statuses become redundant when not using a stream per object. I think it's fine but needs to be called out.

draft-ietf-moq-transport.md
```diff
@@ -975,16 +975,58 @@ A canonical MoQ Object has the following information:
 IDs starts at 0, increasing sequentially for each object within the
 group.
 
-* Object Send Order: An integer indicating the object send order
```
kixelated, line 978: pls keep

kixelated (changes requested):

kixelated:
After reviewing fully, yeah I don't think we can decouple drops from resets.

When any publisher (including a relay) decides to drop an object, it will also reset the associated QUIC stream. Otherwise, the publisher either has to either delay writing to streams (TCP low watermark) or commit to delivering objects it wants to drop.

And the problem with TCP low watermark in MoQ is of course... _priorities_. Newer content is usually more important, so even if we already have N bytes queued in the QUIC layer, our new content needs to skip the queue. The only way to do that is to actually write to the QUIC stream marked at a higher priority... or reset streams currently queued.

draft-ietf-moq-transport.md
fluffy: made a protocol error in next version of PR

draft-ietf-moq-transport.md
fluffy: Added this in

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
fluffy: changed this to "normal"  - perhaps there is a better name. I think we will do need the 00 code because in same cases you don't have an explicit payload length to decide to send this or not. Check it out in next version of PR.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
fluffy: Changed to be more or less what Luke suggest.

draft-ietf-moq-transport.md
fluffy: Tried to clarify this. Part of the issue was can a normal object ( status code 0 ) have a zero length payload. I do not see any reason not to allow this. It makes it clear it is not "missing" but it just has no data. I think Alan had a use case for this. Check out next version of PR and see if it seems clearer on all this.

draft-ietf-moq-transport.md
fluffy: I tried to make the cache or not very explicit in new version. I think it resolves this.

kixelated:
Sorry I left a PR review a little too fast between meetings and taking care of the newborn. I had a chance to think things through a bit more.

> I think there's a better way to signal drops though than to retransmit the same OBJECT but with a different status/payload after a RESET_STREAM.

I'm warming up to the status in OBJECT, however it needs to be a bitmask. An object can be both dropped, the end of a group, and the end of a track. It's a little annoying to have to process duplicate OBJECTs with different statuses but it's fine I guess.

An alternative would be a reliable SUBSCRIBE control stream that contains reliable messages GROUP_END, TRACK_END, OBJECT_DROPPED messages. There would be no need to retransmit anything on a RESET_STREAM but the overhead would be marginally higher. 

ex. GROUP_END would consist of the message type, the group_id VarInt, and a object_id VarInt only. The track would be inferred from the subscribe control stream.

afrind:
> An object can be both dropped, the end of a group, and the end of a track

I think Cullen's update addressed this by making "end of group" and "end of track" have *different object IDs* than the last actual object.  Should we further state that only objects with status 0 can be "dropped by a relay"?  All the other statuses may be needed by the subscriber, and all should be very small.

Would you be ok with moving this PR forward if we split out 0x5 and 0x6 for now?

draft-ietf-moq-transport.md
```diff
@@ -1012,6 +1076,7 @@ OBJECT_STREAM Message {
   Group ID (i),
   Object ID (i),
   Object Send Order (i),
+  Object Status (i),
```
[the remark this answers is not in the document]
ianswett: I assumed we were doing that on a separate stream?

draft-ietf-moq-transport.md
```diff
@@ -982,9 +982,73 @@ group.
...
 * Object Payload: An opaque payload intended for the consumer and SHOULD
 NOT be processed by a relay.
 
+There is also a status field which help subscribers understand what
+objects it will not receive due to then being dropped, lost, or never be
+produced including the case where they will not be produced because they
+are beyond the end of a group or track. Every object has an associated
+`Status` that can have following values:
+
+* 0x0 := Normal object. The payload is array of bytes and can by empty.
```
ianswett, line 997: This is only a valid value for OBJECT_STREAM, correct?

draft-ietf-moq-transport.md
```diff
@@ -982,9 +982,73 @@ group.
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
+* Object Status: As enumeration used for missing or dropped objects that
```
ianswett, line 985: This it can also signal GROUP_END and TRACK_END.

draft-ietf-moq-transport.md
```diff
@@ -982,9 +982,73 @@ group.
...
+  indicates why data is missing.
+
 * Object Payload: An opaque payload intended for the consumer and SHOULD
 NOT be processed by a relay.
 
+There is also a status field which help subscribers understand what
```
ianswett, line 991: The status has a wider set of functionality than just drops and gaps.

Can you clarify that all status values except 0x0 do not have any payload?

ianswett (commented):
LG, but some minor comments to clarify that all statuses besides 0x0 don't have a payload.

And I would prefer to remove 0x5 and 0x6 for now.

draft-ietf-moq-transport.md
```diff
@@ -1128,11 +1196,13 @@ To send an Object with `Object Forwarding Preference` = `Group`, find the open
 stream that is associated with the subscription, `Group ID` and `Object
 Send Order`, or open a new one and send the `STREAM_HEADER_GROUP` if needed,
 then serialize the following fields.
+The Object Status field is only sent if the Object Payload Length is zero.
 
 ~~~
 {
   Object ID (i),
   Object Payload Length (i),
+  [Object Status (i)],
```
kixelated, line 1205: What happens if we get a stream FIN without status 0x3? Is that a fatal error or can we use it as an implicit signal?

draft-ietf-moq-transport.md
```diff
@@ -1089,12 +1155,14 @@ Forwarding Preference` = `Track`.
...
 stream that is associated with the subscription, or open a new one and send the
 `STREAM_HEADER_TRACK` if needed, then serialize the following object fields.
+The Object Status field is only sent if the Object Payload Length is zero.
 
 ~~~
 {
   Group ID (i),
   Object ID (i),
   Object Payload Length (i),
+  [Object Status (i)],
```
kixelated, line 1165: What happens if we get a stream FIN without status 0x4. Is that a fatal error or is it implicit?

kixelated (changes requested):

draft-ietf-moq-transport.md
kixelated: I'd like it to be implicit, or banned.

fluffy:
Note to cullen - remove the two codes. Clean up other comments. Ping Ian when done and ready to merge.

draft-ietf-moq-transport.md
```diff
@@ -982,9 +982,73 @@ group.
...
+         GroupID. This is sent right after the last object in the
+         group. This SHOULD be cached.
+
+* 0x4 := Indicates end of Track and Group. GroupID is one greater than
+         the largest group produced in this track and the ObjectId is
+         one greater than the largest object produced in that
+         group. This is sent right after the last object in the
+         track. This SHOULD be cached.
+
+* 0x5 := Indicates Object was dropped at a relay. Sent by a relay
```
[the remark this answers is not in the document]
fluffy: Removed in next version

draft-ietf-moq-transport.md
```diff
@@ -982,9 +982,73 @@ group.
...
+not want a solution where we add a whole frame of latency on the encoder
+waiting for next frame so it can set the end of group field on the last
+last object in the group.
+
+Any other value SHOULD be treated as a protocol error and terminate the
+session with a Protocol Violation ({{session-termination}}).
+
+There are times where some of the status information is redundant with
+information learned at the lower layer. For example, when using stream
+per object, the end of Group is redundant.
```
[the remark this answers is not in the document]
fluffy: oops - changed as you sugested

draft-ietf-moq-transport.md
kixelated: Actually can we remove this for now? Every status is redundant and confusing with this encoding.

0x1: Implies there can be gaps in group streams which I don't want to support. And if we did end up supporting it, intentional gaps can be detected via the sequence numbers.

0x2: Should be in the group stream header (or separate message). Using the first object to indicate the entire group is skipped is really confusing and error prone. Also I don't like intentional gaps in groups but that's a separate issue.

0x3: Conflicts with STREAM FIN. It can only create problems unless we allow multiple streams per group, but that's counterproductive.

0x4: Should be in the group header (or a separate message). Conflicts with SUBSCRIBE_END since it can indicate the end of the track too.

kixelated (commented):
I removed my -1 so if others think it's fine, merge it.

draft-ietf-moq-transport.md
fluffy: I'm struggling how to phrase this.  I was thinking  of the obj formats could have an empty payload.

draft-ietf-moq-transport.md
```diff
@@ -982,9 +982,73 @@ group.
...
+* 0x2 := Indicates Group does not exist. Indicates that objects with
+         this GroupID do not exist at any publisher and they will not be
+         published in the future. This SHOULD be cached.
+
+* 0x3 := Indicates end of Group. ObjectId is one greater that the
+         largest object produced in the group identified by the
+         GroupID. This is sent right after the last object in the
+         group. This SHOULD be cached.
+
+* 0x4 := Indicates end of Track and Group. GroupID is one greater than
```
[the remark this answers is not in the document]
fluffy: Lets say group had three object 0 to 2. And 2 is dropped. I was thinking what is sent is 0, 1 , then message with status code that indicated 2 was dropped, then message with status code that indicates obj 3 would be end of group marker.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
fluffy: I'm not following the problem here. There would be no change from what we have today in figuring out legth of message.

draft-ietf-moq-transport.md
fluffy: I don't think this changes what a FIN means in any way and I would not want to use these group and end markers in all cases. For applications where they make sense ( like streaming video ) they can can be used. For applications where they make no sense ( like audio only ) the publisher would not bother.

draft-ietf-moq-transport.md
fluffy: Sam thing, status 0x3 is optional, this PR does not change what a FIN does or does not mean.

suhasHere:
I did another read through and it looks good to me !! Thanks Cullen

afrind:
Individual Comment:

>  For applications where they make no sense ( like audio only ) the publisher would not bother.

This means a *relay* cannot assume the existence of these marker objects.  That's unfortunate, because I'd like the end-of-group marker to help my relay know when to close fanout streams.  I don't want to rely on QUIC FIN for this - a publisher may use a QUIC FIN to terminate a stream cleanly (delivering all written objects) as part of a graceful shutdown, without wanting to signal the end of group.  In this case, I explicitly *don't* want to close fanout streams, in case the publisher is reconnecting (make before break).

Would making end markers optional for DATAGRAM tracks sufficiently address the audio only use case?

fluffy:
> > For applications where they make no sense ( like audio only ) the publisher would not bother.
> 
> This means a _relay_ cannot assume the existence of these marker objects. That's unfortunate, because I'd like the end-of-group marker to help my relay know when to close fanout streams. I don't want to rely on QUIC FIN for this - a publisher may use a QUIC FIN to terminate a stream cleanly (delivering all written objects) as part of a graceful shutdown, without wanting to signal the end of group. In this case, I explicitly _don't_ want to close fanout streams, in case the publisher is reconnecting (make before break).
> 
> Would making end markers optional for DATAGRAM tracks sufficiently address the audio only use case?

I get your point and need to think about all of that a bit. Lets say the client did send them, would it be a problem if it just crashed without sending them. Would the relay also have a timer or some other way to clean up the state ? Or just wait for subscribe to end ? Just thinking through implications but you raise a good point.

draft-ietf-moq-transport.md
```diff
@@ -982,9 +982,57 @@ group.
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
+* Object Status: As enumeration used to indicated missing or dropped
```
ianswett, line 985:
```suggestion
* Object Status: As enumeration used to indicated missing
```

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -982,9 +982,73 @@ group.
...
+         does not exist at any publisher and it will not be published in
+         the future. This SHOULD be cached.
+
+* 0x2 := Indicates Group does not exist. Indicates that objects with
+         this GroupID do not exist at any publisher and they will not be
+         published in the future. This SHOULD be cached.
+
+* 0x3 := Indicates end of Group. ObjectId is one greater that the
+         largest object produced in the group identified by the
+         GroupID. This is sent right after the last object in the
```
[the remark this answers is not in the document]
ianswett: Yes, I added a MUST about this below.

draft-ietf-moq-transport.md
```diff
@@ -985,6 +985,50 @@ be sent according to its `Object Forwarding Preference`, described below.
 * Object Payload: An opaque payload intended for the consumer and SHOULD
 NOT be processed by a relay.
 
+* Object Status: As enumeration used to indicated missing
```
afrind, line 988: indicate

draft-ietf-moq-transport.md
```diff
@@ -985,6 +985,50 @@ be sent according to its `Object Forwarding Preference`, described below.
...
+  objects and as an marker of the end of group or track.
```
afrind, line 989: s/and as an marker/or mark/ ?

draft-ietf-moq-transport.md
```diff
@@ -985,6 +985,50 @@ be sent according to its `Object Forwarding Preference`, described below.
...
+
+The Object Status lets subscribers understand what objects will not be received
```
afrind, line 991: s/lets subscribers understand/informs subscribers/

draft-ietf-moq-transport.md
```diff
@@ -985,6 +985,50 @@ be sent according to its `Object Forwarding Preference`, described below.
...
+because they were never produced, are no longer available, or because they
+are beyond the end of a group or track.
+
+`Status` that can have following values:
```
afrind, line 995: remove 'that' ?

draft-ietf-moq-transport.md
```diff
@@ -985,6 +985,50 @@ be sent according to its `Object Forwarding Preference`, described below.
...
+
+* 0x0 := Normal object. The payload is array of bytes and can be empty.
```
afrind, line 997: Given this is in the canonical object part, and payload has already been defined above, maybe it's more clear to define Status first, then in Object Payload state that this property is only present when Status is Normal Object.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
afrind: > This is sent right after the last object

Can this be a MUST, at least for Forwarding Preference = Group?

afrind (commented):
Individual Review

draft-ietf-moq-transport.md
ianswett:
```suggestion
* Object Status: As enumeration used to indicate missing
```

draft-ietf-moq-transport.md
ianswett:
```suggestion
  objects or mark the end of a group or track.
```

draft-ietf-moq-transport.md
ianswett:
```suggestion
The Object Status informs subscribers what objects will not be received
```

draft-ietf-moq-transport.md
ianswett:
```suggestion
`Status` can have following values:
```

draft-ietf-moq-transport.md
ianswett: SGTM

ianswett: Ok, I created a separate section, I think this looks ok now.

PR #432, opened 2024-04-17
Make SUBSCRIBE start and end (typically) absolute

ianswett:
Based on #428

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1203,46 +1203,6 @@ OBJECT_STREAM {
 
 A receiver issues a SUBSCRIBE to a publisher to request a track.
 
-### Subscribe Locations {#subscribe-locations}
-
-The receiver specifies a start and optional end `Location` for the subscription.
-A location value may be an absolute group or object sequence, or it may be a
-delta relative to the largest group or the largest object in a group.
-
-~~~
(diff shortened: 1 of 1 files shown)


kixelated (approved):

draft-ietf-moq-transport.md
```diff
@@ -1279,24 +1239,23 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
-* StartGroup: The Location of the requested group.  StartGroup's Mode MUST NOT be
-None.
+* StartGroup: The start Group ID, plus 1. A value of 0 means the latest group.
```
martinduke, line 1242: This could use a tighter definition of "latest group". For example, if the track is complete, is it the last group in the track? Or is it an error?

afrind (commented):
Individual Review:

I like removing relative options from subscribe.  I don't love the group ID + 1 spelling, but I guess I can live with it.  An alternative is to add 4 code points for SUBSCRIBE to indicate the presence/absence of start/end (SUBSCRIBE, SUBSCRIBE_FROM, SUBSCRIBE_TO, SUBSCRIBE_RANGE).

fluffy (approved):

PR #434, opened 2024-04-17
Add OBJECTS_DROPPED on the control stream

### LINKED ISSUE #345: What if/Why will never deliver an Object/Track
If the relay/origin will never be able to fulfill an entire requested SUBSCRIBE, how can it indicate that?  For example, a sequence of relays and the origin may only keep video within 30 seconds of head and discard older content.

As a related question, does the relay need to indicate that there will be gaps in Object or Group IDs so the player doesn't wait for them?

High-Level Q: An object has not been delivered, why?

1. It doesn't exist anymore.
2. It never existed.
3. There wasn't enough bandwidth to send it based on... (ie: priority, the subscriber's latency requirements, etc)
4. It was sent in a stream/datagram by itself and reordered (I think we have this covered?)

ianswett:
This is intended to be complementary to #429 by adding a SUBSCRIBE specific mechanism to indicate hop-by-hop Objects that will not be delivered.  Because this is specific to a SUBSCRIBE, it's not intended to be cacheable.

Builds on ideas in #423 

Each SUBSCRIBE could be moved to their own bidi stream and this could be sent there, but that change can be made separately.

Fixes #345

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1551,9 +1551,9 @@ one or more Objects will not be delivered for this subscription due to
 congestion or other constraints.  This does not indicate the dropped
 Objects are permanantly unavailable.
 
-If Objects within the range have already been received or are subsequently
-received, they MAY be processed as normal, but there is no expectation
-they will be delivered.
+If Objects within the range have already been or are subsequently received,
+they MAY be processed as normal, but there is no expectation they will be
+delivered.
 
 ~~~
 OBJECTS_DROPPED Message {
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1551,9 +1551,9 @@ one or more Objects will not be delivered for this subscription due to
 congestion or other constraints.  This does not indicate the dropped
 Objects are permanantly unavailable.
 
-If Objects within the range have already been or are subsequently received,
-they MAY be processed as normal, but there is no expectation they will be
-delivered.
+If Objects between (inclusive) the first and last Object have already been
+or are subsequently received, they MAY be processed as normal, but there is no
+expectation they will be delivered.
 
 ~~~
 OBJECTS_DROPPED Message {
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1551,6 +1551,10 @@ one or more Objects will not be delivered for this subscription due to
 congestion or other constraints.  This does not indicate the dropped
 Objects are permanantly unavailable.
 
+If Objects within the range have already been received or are subsequently
+received, they MAY be processed as normal, but there is no expectation
+they will be delivered.
+
 ~~~
 OBJECTS_DROPPED Message {
   Subscribe ID (i),
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -798,6 +798,8 @@ MOQT Message {
 |-------|-----------------------------------------------------|
 | 0xE   | TRACK_STATUS ({{message-track-status}})             |
 |-------|-----------------------------------------------------|
+| 0xF   | OBJECTS_DROPPED ({{message-objects-dropped}})       |
+|-------|-----------------------------------------------------|
 | 0x10  | GOAWAY ({{message-goaway}})                         |
 |-------|-----------------------------------------------------|
 | 0x40  | CLIENT_SETUP ({{message-setup}})                    |
@@ -1541,6 +1543,44 @@ information with status code 0x04.
 The receiver of multiple TRACK_STATUS messages for a track uses the information
 from the latest arriving message, as they are delivered in order on a single
 stream.
+
+## OBJECTS_DROPPED {#message-objects-dropped}
+
+An endpoint sends a 'OBJECTS_DROPPED' message on the control stream when
+one or more Objects will not be delivered for this subscription due to
+congestion or other constraints.  This does not indicate the dropped
+Objects are permanantly unavailable.
+
+~~~
+OBJECTS_DROPPED Message {
+  Subscribe ID (i),
+  First Group (i),
+  First Object (i),
+  Last Group (i),
+  Last Object (i),
+  Fin (i),
+}
+~~~
+{: #moq-track-objects-dropped title="MOQT OBJECTS_DROPPED Message"}
+
+* Subscribe ID: Subscription Identifier as defined in {{message-subscribe-req}}.
+
+* First Group: The ID of the first group that contains an Object which won't be
+delivered.
+
+* First Object: The ID of the first Object within the group which won't be
+delivered.
+
+* Last Group: The ID of the last group that contains an Object which won't be
+delivered.
+
+* Last Object: The ID of the last Object within the group which won't be
+delivered.
+
+* Fin: When 1, the last Object is the final Object in the group, and when 2,
+the last Object is the final Object in the Track as well as the group.
+0 otherwise, and values greater than 2 are illegal.
+
 
 ## GOAWAY {#message-goaway}
 The server sends a `GOAWAY` message to initiate session migration


draft-ietf-moq-transport.md
```diff
@@ -1542,6 +1544,48 @@ The receiver of multiple TRACK_STATUS messages for a track uses the information
...
+* First Object: The ID of the first Object within the group which won't be
+delivered.
+
+* Last Group: The ID of the last group that contains an Object which won't be
+delivered.
+
+* Last Object: The ID of the last Object within the group which won't be
+delivered.
+
+* Fin: When 1, the last Object is the final Object in the group, and when 2,
```
kixelated, line 1584: I don't think this works when multiple groups are specified.

Could we specify a specific `group_id` instead of a range? It's fine sending multiple messages.

draft-ietf-moq-transport.md
```diff
@@ -1542,6 +1544,48 @@ The receiver of multiple TRACK_STATUS messages for a track uses the information
...
+* First Group: The ID of the first group that contains an Object which won't be
+delivered.
+
+* First Object: The ID of the first Object within the group which won't be
+delivered.
+
+* Last Group: The ID of the last group that contains an Object which won't be
+delivered.
+
+* Last Object: The ID of the last Object within the group which won't be
```
kixelated, line 1581: Do we want to allow gaps? I don't really want to support dropping only the middle of a group. 

I think we should only allow dropping the tail of a group. You can still prioritize objects separately, but eh it's just a pain to support dropping them out of order. Is there a use-case?

If so, then we can remove this `Last Object` and `Fin` since it's implied.

draft-ietf-moq-transport.md
```diff
@@ -1542,6 +1544,48 @@ The receiver of multiple TRACK_STATUS messages for a track uses the information
 from the latest arriving message, as they are delivered in order on a single
 stream.
 
+## OBJECTS_DROPPED {#message-objects-dropped}
+
+An endpoint sends a 'OBJECTS_DROPPED' message on the control stream when
+one or more Objects will not be delivered for this subscription due to
+congestion or other constraints.  This does not indicate the dropped
+Objects are permanantly unavailable.
```
kixelated, line 1552: A flag indicating as such would be good to prevent spurious retries. 404 versus 409.

kixelated (commented):
I like it more than #429, but think we can scope down what can be dropped.

draft-ietf-moq-transport.md
ianswett: If they're permanently unavailable, that should be cacheable and Cullen's PR can handle that.

draft-ietf-moq-transport.md
ianswett: If one prioritizes sending certain Objects over others, at some point, you can imagine dropping them, in which case we'd want a mechanism to indicate that.

draft-ietf-moq-transport.md
ianswett: I can go either way on this.

draft-ietf-moq-transport.md
kixelated: That's true. It's fine but I really don't like the temporal scalability use-case because it creates these jagged edges. At some point I think we should explicitly rule it out and require that SVC layers are separate tracks, not interleaved objects within a group.

draft-ietf-moq-transport.md
kixelated: I don't understand the relation to Cullen's PR then. I thought this PR effectively replaced status codes 0x0 - 0x3?

ianswett: No, the intent is only to replace the case when it's not deliverable due to congestion/etc, which I believe we cannot/should not cache.  If the Objects are immutable/cacheable, then they'd be delivered on unidirectional streams, even if they're missing/length 0.

kixelated: I think that's almost the case, but not quite.

A broadcaster will produce an OBJECT size=1234 and keep it in an internal cache for 5s. A subscriber might need it immediately so we sent it. But if a fetch comes in later, after the 5s expiration, then it sounds like you want the broadcaster to send OBJECT size=0 instead (or with an equivalent status code)? Otherwise there's no way of indicating that an object is permanently gone.

That kind of breaks the immutable property, and it gives a mixed signal because the object used to exist but no longer. IMO the broadcaster should send OBJECTS_DROPPED instead, just with a flag saying please don't try again; trust me.

ianswett: If an object goes from available to permanently gone, then that new status is cacheable.  The object didn't change, it's availability did.

From a functional perspective, we could definitely add a flag indicating the Objects were permanently dropped, but I think that mixes up the Track specific information with the Subscription specific information.

OBJECTS_DROPPED is intended to be specific to a subscription, and Objects sent on unidirectional streams are part of the track.

fluffy (commented):
I have one sort of meta issue about this and proposal to fix. 

Imagine a case that is audio only where each audio packet is on a new group and where the relay is dropping a large percentage of theses. The objects dropped message are pretty large and comparable in size to the actual audio packets that are getting dropped. 

I'm worried about the priorities - and yes, I know I am the person always saying lets leave the priorities to the next interim meeting.  I am assuming we will set the control channels at the highest priority because it is bad to loose the control messages that allow you shut off the flood of data from other things. ( I've seen phones where video is higher priority than control messages and you get on a congested link and you can not hang up - really funny ). My worry here is that in audio case above will cause the control link to become swamped with these messages. 

So few ideas on how to fix this... 

1. in this PR, make it so these happen on some to be determined in the future stream that is lower priority than control stream and we can sort out what stream that is and what priorities as part of the upcoming priority discussion 

2. Make it so "objects dropped" message could only be sent at a max rate over every 100 ms per stream and instead of trying to describe all the gaps, just indicated that on or more object were dropped between the start group/object and end group/object range. 

I'm trying to think about how this will be used and it seems like in most cases, it will be used to know to move on an where the good data that was not dropped is the key thing, not the fine grain details of what got dropped. 

All the details of the proposal other than this seem reasonable and a fine starting point.

kixelated:
Another 3rd option Cullen would be a control stream per SUBSCRIBE.

That way OBJECTS_DROPPED would only block SUBSCRIBE_END, which almost sounds like a feature?

fluffy:
> Another 3rd option Cullen would be a control stream per SUBSCRIBE.
> 
> That way OBJECTS_DROPPED would only block SUBSCRIBE_END, which almost sounds like a feature?

I can easily imagine something like that but I think we need to deal with the following requirements one way or another:

1. If we have three subscribes for three tracks A, B, C and A is always more important than B , which is more important than C.  You don't want the OBJ_DROP messages for B to have stop all the data on A track from flowing. (and I will hand wave the what is newer / older / priority etc but you can see how as we dig into that, we will still need to solve some version of this ) 

2. If a publisher is sending a ton of OBJ_DROP to a subscriber on a badly congested link,  it  must not result in a situation where the subscriber can not tell the publisher to end the subscription. Even when the bandwidth each direction is highly asymmetrical.

suhasHere (changes requested):
Not that I am against this PR. I am not seeing a need for this PR yet as most of these can be realized with Cullen's updated PR. I would wait until we decide on what cases Cullen's PR to better understand the requirements.

ianswett:
The audio stream use case is an interesting one.

I can imagine this instead being "Anything prior to this Object that you didn't receive, you probably won't"?  That means it doesn't scale with the number of Objects.

draft-ietf-moq-transport.md
suhasHere, line 1552: The PR #429  has another way to report similar information by sending a new stream with the information. Should these be same or are we expecting OBJECTS_DROPPED to be more informative as a reporting over ranges of things that can be dropped?

ianswett: The concept is that OBJECTS_DROPPED is specific to a subscription, so it's on the control stream, and that information isn't cacheable, because the Objects may exist and be retrievable.

afrind: > That kind of breaks the immutable property

It seems like TTL will mess with the immutable property regardless of how we communicate transient drops.  I suggest we discuss as part of adding TTLs.

afrind: I like this separation of track information and subscription specific information.

draft-ietf-moq-transport.md
afrind: I think for object-per-stream you might want the range feature to reduce the chattiness.  eg: the receiver got backed up and objects 5-10 never got sent, but 1-4 and 11-20 did.

I think we need to specify inclusivity/exclusivity of end.  Using the new Subscribe encoding may make sense, and that provides a way to say "to the end of the group" with End Object=0.

draft-ietf-moq-transport.md
afrind: Oh this is a very good point - I think we should have one group per drop message.

afrind (commented):
Individual Comments

draft-ietf-moq-transport.md
ianswett: I can make this really simple and just say have a group ID indicating that you're not going to get anything else from this group.

It's a bit chatty when you have one Object per group and drop a bunch of them, but if we put each subscription on its own stream, then this would be 1 byte for the message type and a varint for the group number, which seems reasonable.

ianswett: I sent out https://github.com/moq-wg/moq-transport/pull/443 as an alternative

fluffy:
I think I prefer #434 to this PR.

ianswett:
The current direction is to not include a specific signal for dropping, so closing this PR.

PR #435, opened 2024-04-18
SUBSCRIBE_UPDATE

### LINKED ISSUE #269: Multiple subscriptions and updating a subscription

ianswett:
Replace #385 based on the simplified SUBSCRIBE.

Fixes #269
Closes #385

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1354,7 +1354,7 @@ the new start Object.  The end Object MUST NOT increase and when it decreases,
 there is no guarantee that a publisher will not have already sent Objects after
 the new end Object.
 
-Unlike a new subscription, SUBSCRIBE_UPDATE does not cause an Object to be
+Unlike a new subscription, SUBSCRIBE_UPDATE can not cause an Object to be
 delivered multiple times.
 
 The format of SUBSCRIBE_UPDATE is as follows:
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1315,12 +1315,15 @@ SUBSCRIBE_UPDATE Message {
 {: #moq-transport-subscribe-update-format title="MOQT SUBSCRIBE_UPDATE Message"}
 
 * Subscribe ID: The subscription identifier that is unique within the session.
+This MUST match an existing Subscribe ID.
 
 * StartGroup: The start Group ID.
 
-* StartObject: The start Object ID.
+* StartObject: The start Object ID, plus 1. A value of 0 means the entire group
+is requested.
 
-* EndGroup: The end Group ID.
+* EndGroup: The end Group ID, plus 1.  A value of 0 means the subscription is
+open-ended.
 
 * EndObject: The end Object ID, plus 1. A value of 0 means the entire group is
 requested.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1261,9 +1261,43 @@ If a publisher cannot satisfy the requested start or end for the subscription it
 MAY send a SUBSCRIBE_ERROR with code 'Invalid Range'. A publisher MUST NOT send
 objects from outside the requested start and end.
 
-TODO: Define the flow where subscribe request matches an existing subscribe id
-(subscription updates.)
+## SUBSCRIBE_UPDATE {#message-subscribe-update-req}
 
+A receiver issues a SUBSCRIBE_UPDATE to a publisher to request a change to
+a prior subscription.  The start Object MUST NOT decrease and when it increases,
+there is no guarantee that a publisher will not send Objects after the new
+start Object.  The end Object MUST NOT increase and when it decreases,
+there is no guarantee that a publisher will not send Objects after the new
+end Object.
+
+The format of SUBSCRIBE_UPDATE is as follows:
+
+~~~
+SUBSCRIBE_UPDATE Message {
+  Subscribe ID (i),
+  StartGroup (i),
+  StartObject (i),
+  EndGroup (i),
+  EndObject (i),
+  Number of Parameters (i),
+  Track Request Parameters (..) ...
+}
+~~~
+{: #moq-transport-subscribe-update-format title="MOQT SUBSCRIBE_UPDATE Message"}
+
+* Subscribe ID: The subscription identifier that is unique within the session.
+
+* StartGroup: The start Group ID.
+
+* StartObject: The start Object ID.
+
+* EndGroup: The end Group ID.
+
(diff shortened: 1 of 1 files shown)


fluffy (approved):
What I like about this is we explicitly separate what can be setup in the initial subscription, and what can be changed about it later. I think that will make things easier to understand over the long run.

kixelated (approved):
Being able to increase the start group is a nice way for the subscriber to drop old content it doesn't care about. I like the simplicity.

draft-ietf-moq-transport.md
```diff
@@ -1261,9 +1261,43 @@ If a publisher cannot satisfy the requested start or end for the subscription it
...
 objects from outside the requested start and end.
 
-TODO: Define the flow where subscribe request matches an existing subscribe id
-(subscription updates.)
+## SUBSCRIBE_UPDATE {#message-subscribe-update-req}
 
+A receiver issues a SUBSCRIBE_UPDATE to a publisher to request a change to
+a prior subscription.  The start Object MUST NOT decrease and when it increases,
+there is no guarantee that a publisher will not send Objects after the new
+start Object.  The end Object MUST NOT increase and when it decreases,
```
afrind, line 1269: Do you mean "before the new start object" ?

draft-ietf-moq-transport.md
```diff
@@ -1261,9 +1261,43 @@ If a publisher cannot satisfy the requested start or end for the subscription it
...
+  StartObject (i),
+  EndGroup (i),
+  EndObject (i),
+  Number of Parameters (i),
+  Track Request Parameters (..) ...
+}
+~~~
+{: #moq-transport-subscribe-update-format title="MOQT SUBSCRIBE_UPDATE Message"}
+
+* Subscribe ID: The subscription identifier that is unique within the session.
```
suhasHere, line 1288: the subscribe Id must match an existing subscription or else its an error.

draft-ietf-moq-transport.md
```diff
@@ -1261,9 +1261,43 @@ If a publisher cannot satisfy the requested start or end for the subscription it
...
-(subscription updates.)
+## SUBSCRIBE_UPDATE {#message-subscribe-update-req}
 
+A receiver issues a SUBSCRIBE_UPDATE to a publisher to request a change to
+a prior subscription.  The start Object MUST NOT decrease and when it increases,
+there is no guarantee that a publisher will not send Objects after the new
+start Object.  The end Object MUST NOT increase and when it decreases,
+there is no guarantee that a publisher will not send Objects after the new
+end Object.
+
```
suhasHere, line 1272: I am bit confused on the requirements here. An update to an existing subscription updates the filter that the publisher looks at. A filter that is smaller than the original implies lesser objects than before.

suhasHere (changes requested):
+1 on needing subscribe update. But not sure of the proposal on object/group Ids and what gets delivered

draft-ietf-moq-transport.md
ianswett: I added text to clarify the intent, which is that one can only narrow a subscription, not widen it.

afrind:
Individual Comments:

Is there a response to a Subscribe Update?  For example, maybe the auth has changed and is no longer valid?

I think there is weirdness with respect to track and group forwarding preferences when you create a gap by moving start forward. I think we should require the publisher to close QUIC streams with objects that haven't been requested and open new ones.  eg:

-> Subscribe to a track with forward pref = track, start=0/0
<- Subscribe Ok
<- Open stream 1, sending objects starting from 0/0
-> Subscribe update, start=10/0
<- Close stream 1
<- Open stream 2, sending objects starting from 10/0

Alternately, should we only allow End to be modified by SUBSCRIBE_UPDATE?  This is should always succeed and doesn't have any hard edges.  To move start forward, or change parameters, end the current subscription and start a new one?

draft-ietf-moq-transport.md
```diff
@@ -1342,9 +1342,49 @@ If a publisher cannot satisfy the requested start or end for the subscription it
...
-TODO: Define the flow where subscribe request matches an existing subscribe id
-(subscription updates.)
+## SUBSCRIBE_UPDATE {#message-subscribe-update-req}
 
+A receiver issues a SUBSCRIBE_UPDATE to a publisher to request a change to
+a prior subscription.  Subscriptions can only become more narrower, not wider,
+because an attempt to widen a subscription could fail.  If Objects before the
+start or after the end of the current subscription are needed, a separate
+subscription can be made. The start Object MUST NOT decrease and when it increases,
+there is no guarantee that a publisher will not have already sent Objects after
```
ianswett, line 1352:
```suggestion
there is no guarantee that a publisher will not have already sent Objects before
```

suhasHere (approved):

ianswett:
I wasn't intending for there to be a response to SUBSCRIBE_UPDATE.  If auth fails, I think the right thing to do is terminate the subscription with an error.

Interesting point about forwarding preference.  Do we have to say anything about stream states/etc there?  This seems similar to the case when you have severe congestion and have to drop a bunch of Objects instead of delivering them.

I'd like to be able to modify start, because it is useful in live use cases, because it allows a client to tell the server 'stop sending content before this point'.

afrind:
Individual Comment:

> Interesting point about forwarding preference. Do we have to say anything about stream states/etc there? This seems similar to the case when you have severe congestion and have to drop a bunch of Objects instead of delivering them.

We're still spelling how this dropping works, but it seems like in those cases there will be OBJECT_DROPPED message or similar that explains the gap.

On the one hand, the publisher put a bunch of objects on a single stream for a reason,  so skipping over a bunch seems like it could break things.  On the other hand, it was the subscriber who asked for the gap, so it shouldn't be surprising.  I think we need some guidance about how to handle this, but I guess we can come back to it later as well.

draft-ietf-moq-transport.md
```diff
@@ -1342,9 +1342,52 @@ If a publisher cannot satisfy the requested start or end for the subscription it
...
+a prior subscription.  Subscriptions can only become more narrower, not wider,
+because an attempt to widen a subscription could fail.  If Objects before the
+start or after the end of the current subscription are needed, a separate
+subscription can be made. The start Object MUST NOT decrease and when it increases,
+there is no guarantee that a publisher will not have already sent Objects before
+the new start Object.  The end Object MUST NOT increase and when it decreases,
+there is no guarantee that a publisher will not have already sent Objects after
+the new end Object.
+
+Unlike a new subscription, SUBSCRIBE_UPDATE does not cause an Object to be
```
ianswett, line 1357:
```suggestion
Unlike a new subscription, SUBSCRIBE_UPDATE can not cause an Object to be
```

Issue #436, opened 2024-04-19
New SUBSCRIBE limitations


afrind:
As currently defined, you can SUBSCRIBE starting from the following points:

1. The latest group and latest object
2. An absolute group and latest object
3. An absolute group and absolute object

You cannot SUBSCRIBE to 

a. The latest group with an absolute object
b. The latest group from the beginning
c. The next group after latest, from the beginning

I'd like to remove 2) (absolute group, latest object).  If this group is in the past, then likely the group is complete and no new objects will come in that group, so it's equivalent to (absolute group + 1, object 0).  If this group is in the future, then it's equivalent to (absolute group, object 0).  If the group is current, it's equivalent to latest/latest.  Does anyone have a use case that requires this mode?

I'd also like to add b) (latest group, first object) and c) (next group, object 0).  This can be done now with an extra round trip via TRACK_STATUS_REQUEST, so perhaps this is an optimization we can tackle later.

kixelated:
> As currently defined, you can SUBSCRIBE starting from the following points:
> 
> 1. The latest group and latest object
> 2. An absolute group and latest object
> 3. An absolute group and absolute object
> 
> You cannot SUBSCRIBE to
> 
> a. The latest group with an absolute object
> b. The latest group from the beginning
> c. The next group after latest, from the beginning

It's currently ambiguous if `case 1` or `case b` happens. When `start_group=0`, then `start_object` is not present but doesn't have a specified default. If `default=0` then `case 1` occurs, and if `default=1` then `case b` occurs. The correct behavior is `case b` IMO since groups are sync points. 

+1 to removing `case 2`.

kixelated:
And I don't think we should support `case c` without a good use-case. It's too racey for ABR and too delayed for initial playback. 

It's also barely an optimization since the time spent making the TRACK_STATUS_REQUEST eats into time that would have been spent waiting for the next group to start anyway.

kixelated:
#437 means you can only start at:

b. The latest group from the beginning
3. An absolute group and absolute object

Everything else is accomplished with `TRACK_STATUS_REQUEST`, forcing the subscriber to converts relative values to absolute. I think we should eventually bring back Locations so the publisher can do this conversion too without an additionally RTT but the desire was to simplify start/end ranges.

ianswett:
FWIW, I thought/intended start_group=0 to correspond to b.

Also +1 to removing case 2.

fluffy:
Thanks for summary. 

[...]

PR #437, opened 2024-04-19
Exclusive SUBSCRIBE end range

kixelated:
This removes the `N/latest` functionality which is super confusing. It's also just cleaner to use exclusive end ranges instead of the previously implied inclusive ranges.

Note that to receive a single group, you need to use:
```
    start_group=N  start_object=0
    end_group=N+1  end_object=0
```

Previously that was accomplished with:
```
    start_group=N  start_object=1
    end_group=N    end_object=0
```

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1211,10 +1211,10 @@ SUBSCRIBE Message {
   Track Alias (i),
   Track Namespace (b),
   Track Name (b),
-  StartGroup (i),
-  [ StartObject (i), ]
-  EndGroup (i),
-  [ EndObject (i), ]
+  Start Group (i),
+  [ Start Object (i), ]
+  End Group (i),
+  [ End Object (i), ]
   Number of Parameters (i),
   Track Request Parameters (..) ...
 }
@@ -1239,16 +1239,14 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
-* StartGroup: The start Group ID, plus 1. A value of 0 means the latest group.
+* Start Group: The inclusive start Group ID, plus 1. A value of 0 means the latest Group ID.
 
-* StartObject: The start Object ID, plus 1. A value of 0 means the latest object.
-This field is not present when Start Group is 0.
+* Start Object: The inclusive start Object ID. When Start Group is 0, this field is not present and defaults to 0.
 
-* EndGroup: The end Group ID, plus 1. A value of 0 means the subscription is
-open-ended and continues to the end of the track.
+* End Group: The exclusive end Group ID, plus 1. A value of 0 means the subscription
+continues to the end of the track.
 
-* EndObject: The end Object ID, plus 1. A value of 0 means the entire group is
-requested. This field is not present when End Group is 0.
+* End Object: The exclusive end Object ID. When End Group is 0, this field is not present and the subscription continues to the end of the group.
 
 * Track Request Parameters: The parameters are defined in
 {{version-specific-params}}


afrind:
See related issue #436

draft-ietf-moq-transport.md
```diff
@@ -1239,16 +1239,14 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
-* StartGroup: The start Group ID, plus 1. A value of 0 means the latest group.
+* Start Group: The inclusive start Group ID, plus 1. A value of 0 means the latest Group ID.
 
-* StartObject: The start Object ID, plus 1. A value of 0 means the latest object.
-This field is not present when Start Group is 0.
+* Start Object: The inclusive start Object ID. When Start Group is 0, this field is not present and defaults to 0.
```
afrind, line 1244: Note: this changes the start group=0 from meaning latest/latest to meaning latest/0 and there is no way to express latest/latest after this change.

kixelated: Well the meaning was `latest/undefined`. My intent was `latest/0` but I can see how it could be interpreted as `latest/latest`.

draft-ietf-moq-transport.md
```diff
@@ -1239,16 +1239,14 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
...
 
-* EndGroup: The end Group ID, plus 1. A value of 0 means the subscription is
-open-ended and continues to the end of the track.
+* End Group: The exclusive end Group ID, plus 1. A value of 0 means the subscription
```
afrind, line 1246: The group is only exclusive when end object is 0.  I understand what you mean, but we should clarify the text. 

If I want to get up through g=7, o=5, I send end: g=7 (encoded 8) o=6.  If I want all of group 6 but none of group 7, I send start: g=6 (encoded 7), o=0, end: g=7 (encoded 8), o=0.

kixelated: Ah yeah, the tuple is exclusive.

draft-ietf-moq-transport.md
```diff
@@ -1211,10 +1211,10 @@ SUBSCRIBE Message {
   Track Alias (i),
   Track Namespace (b),
   Track Name (b),
-  StartGroup (i),
-  [ StartObject (i), ]
-  EndGroup (i),
-  [ EndObject (i), ]
+  Start Group (i),
+  [ Start Object (i), ]
```
afrind, line 1215: Suggestion to support both "latest/0" and "latest/latest": Start Object is always present, but when Start Group is 0, 0 means object 0, 1 means latest object and values >= 2 are illegal.

draft-ietf-moq-transport.md
fluffy, line 1244: I think "inclusive" and "exclusive" are not as good as just using  ">="  and  "< ".  I'd have a slight preference for that.  I can certainly live with  "inclusive" and "exclusive" and no big deal.

draft-ietf-moq-transport.md
```diff
@@ -1239,16 +1239,14 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
-* StartGroup: The start Group ID, plus 1. A value of 0 means the latest group.
+* Start Group: The inclusive start Group ID, plus 1. A value of 0 means the latest Group ID.
```
fluffy, line 1242: The more I read this +1, the less I like it. I can easily live with using 0 as the special value, but as an alternative, what would you think of:

We add a boolean flag field call "latest(f)," and make start group optional and only included if latest is false. Similarly we add a field called "forever(f)," and make end group optional only included if forever was false. Then we could get rid of +1. Thoughts ? and to be clear, I can easily live with it like it is.

draft-ietf-moq-transport.md
fluffy: The use cases for latest/latest are really weak - I think we could live without that for now.

draft-ietf-moq-transport.md
fluffy: We 100% need latest/0 so perhaps just some text to clear that up.

fluffy (changes requested):
This PR is making things better and clarifying stuff so I am in favor of merging, just think there are a few things that could get fixed up before merging.

suhasHere:
Submitted https://github.com/moq-wg/moq-transport/pull/439 as another way to clarify some ambiguities.

wilaw:
This PR assumes the end_group N+1 exists. What about the case of an asset which only has groups from 1..10. If I want to subscribe to all of these, I have request 

 start_group=1  start_object=0
 end_group=11  end_object=0

This is semantically awkward to me, since it references an entity (group 11) which does not exist. It would be so much cleaner and more intuitive if the API allowed us to ask for what we want, via inclusive threshold. I would rename them to "first" and "last" to make their syntax clearer. So to subscribe to groups 1..10 I could do

first_group=1  start_object=0
last_group=10  end_object=0

I then receive receive exactly what I ask for.

afrind:
This spiritually merged into #439

PR #439, opened 2024-04-22
Simplify/Clarify Subscribe by adding 4 modes

suhasHere:
I took a stab at another rewrite of Subscribe message which attempts to not use special group and object values. Also add support for multiple subscription flows that have been discussed.

As this is a control message, we should be Ok to be elaborate in describing the subscriber needs.

This leaves 0 as a special value for the StartObject and EndObject.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1276,7 +1276,8 @@ See ({{sub-filter}}).
 
 * EndGroup: The end Group ID. Only present for the "AbsoluteRange" filter type.
 
-* EndObject: The end Object ID. Only present for the "AbsoluteRange" filter type.
+* EndObject: The end Object ID, plus 1. A value of 0 means the entire group is
+requested. Only present for the "AbsoluteRange" filter type.
 
 * Track Request Parameters: The parameters are defined in
 {{version-specific-params}}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1203,21 +1203,24 @@ OBJECT_STREAM {
 
 ### Subscribe Location Filter {#sub-filter}
 
-The receiver specifies a filter value that allows publisher
-to identify the range of groups and objects that needs to be delivered, wherever applicable.
+The subscriber specifies a filter on the subscription to allow
+the publisher to identify which objects need to be delivered.
 
-There are 4 filter values:
+There are 4 types of filters:
 
-Current (0x1) : A filter value of "Current" specifies an open-ended subscription with objects from the beginning of the current group.
+Lastest Group (0x1) : Specifies an open-ended subscription with objects
+from the beginning of the current group.
 
-Latest (0x2): A filter value of "Latest" specifies an open-ended subscription beginning from the current object of the current group.
+Latest Object (0x2): Specifies an open-ended subscription beginning from
+the current object of the current group.
 
-AbsoluteStart (0x3):  A filter value of "AbsoluteStart" specifies an open-ended subscription with objects beginning from group/object identified in the StartGroup and StartObject fields.
+AbsoluteStart (0x3):  Specifies an open-ended subscription with objects
+beginning from the object identified in the StartGroup and StartObject fields.
 
-AbsoluteRange (0x4):  A filter value of "AbsoluteRange" specifies a closed subscription. The group information is obtained from the StartGroup and EndGroup fields, the object information is obtained from the StartObject and the EndObject fields.
+AbsoluteRange (0x4):  Specifies a closed subscription starting at StartObject
+in the StartGroup and ending at EndObject in EndGroup.
 
-
-A filter value other than the above MUST be treated as error.
+A filter type other than the above MUST be treated as error.
 
 
 ### SUBSCRIBE Format
@@ -1231,13 +1234,13 @@ SUBSCRIBE Message {
   Track Alias (i),
   Track Namespace (b),
   Track Name (b),
+  Filter Type (i),
+  [StartGroup (i),
+   StartObject (i)],
+  [EndGroup (i),
+   EndObject (i)],
   Number of Parameters (i),
   Track Request Parameters (..) ...,
-  Location Filter (i),
-  [StartGroup (i)],
-  [StartObject (i)],
-  [EndGroup (i)],
-  [EndObject (i)],
 }
 ~~~
 {: #moq-transport-subscribe-format title="MOQT SUBSCRIBE Message"}
@@ -1260,28 +1263,21 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
+* Location Filter: Identifies acceptable values for start and end group(s)/object(s)
+  to be delivered. See ({{sub-filter}}).
+
+* StartGroup: The start Group ID. Only present for "AbsoluteStart" and
+  "AbsoluteRange" filter types.
+
+* StartObject: The start Object ID within the StartGroup. Only present for
+  "AbsoluteStart" and "AbsoluteRange" filter types.
+
+* EndGroup: The end Group ID. Only present for the "AbsoluteRange" filter type.
+
+* EndObject: The end Object ID. Only present for the "AbsoluteRange" filter type.
+
 * Track Request Parameters: The parameters are defined in
 {{version-specific-params}}
-
-* Location Filter: Identifies acceptable values for start and end group(s)/object(s) to be delivered. See ({{sub-filter}}).
-
-For filter value of "AbsoluteStart", the following interpretation holds:
-
-* StartGroup: The start Group ID. This field MUST be present.
-
-* StartObject: The start Object ID within the group corresponding to the StartGroup. Omitting this field implies objects from the beginning of the group.
-
-The fields EndGroup and EndObject MUST NOT be present.
-
-For filter value of "AbsoluteRange", the following interpretation holds:
-
-* StartGroup: The start Group ID. This field MUST be present.
-
-* StartObject: The start Object ID within the group corresponding to the StartGroup. This field MUST be present.
-
-* EndGroup: The end Group ID. This field MUST be present.
-
-* EndObject: The end Object ID. Omitting this field implies until the end of group specified in EndGroup.
 
 
 On successful subscription, the publisher MUST reply with a SUBSCRIBE_OK,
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1201,6 +1201,25 @@ OBJECT_STREAM {
...
+The receiver specifies a filter value that allows publisher
+to identify the range of groups and objects that needs to be delivered, wherever applicable.
+
+There are 3 filter values:
+
+Current (0x1) : A filter value of "Current" implies objects from the beginning of the current group.
+
+Latest (0x2): A filter value of "Latest" implies beginning from the current object of the current group.
+
+Absolute (0x3):  For filter value of "Absolute", the group information is obtained from the StartGroup and EndGroup fields,
```
afrind, line 1215: Do we want this to apply only to start and not end, so we can support SUBSCRIBE(start=Latest, end=<x, y>)?  
Many different ways to spell that - eg: add another flag field for End Present.

suhasHere: the filter rules specify that when it is Absolute, only StartGroup is mandatory and rest are optional . The proposal here enables all these use-cases

Start at Current group and first object with open ended subscription_( filter : Current)_

Start at Current group and current object with open ended subscription _(filter: latest)_

Start at absolute group and absolute object and end  at absolute group and object _( filter: absolute, StartGroup and StartObject present, EndGroup and EndObject present)_

Start at absolute group, open ended subscription:  (filter: absolute, StartGroup and StartObject present, EndGroup/EndObject omitted)

Start at absolute group and absolute object and end after an absolute group   _( filter: absolute, StartGroup and StartObject present, EndGroup present but the value will be intended end-group +1  and EndObject is set to 0)_

@afrind  do you see other requirements that isn't met

afrind: @suhasHere : 

> Start at absolute group, open ended subscription

I don't see how this is encoded.  Do you need Filter=Absolute (Start only) and Filter=AbsoluteRange (Start and End)?  Or a Message Length to know if there are fields after Start Object?

> Do you see other requirements that isn't met

I don't know if (Start at current/latest, end at absolute) is an important use case, but I don't think it can be encoded.  Maybe the use case is to join a live stream with a bounded end, so you can switch resolutions at an explicit future group?  I'll defer to someone who knows the application space a bit better to argue for or against it.  This use case could be met by SUBSCRIBE_UPDATE maybe?

suhasHere: > I don't see how this is encoded. Do you need Filter=Absolute (Start only) and Filter=AbsoluteRange (Start and End)? Or a Message Length to know if there are fields after Start Object?

Yes Filter is Absolute and specify only StartGroup field

afrind: @suhasHere : you can't have:

Filter = Abs
Start Group=x
Start Obj=y

And

Filter = Abs
Start Group=x
Start Obj=y
End Group=z
End Obj=w

Your parser needs a way to know if End Group is there.

suhasHere: ah ! i get what you meant. yeah you are right .. you need to have something to indicate the same. What would be your preferred encoding for the same ?

afrind: @suhasHere : Here are two options, there are probably more

If you want to support an absolute end group with current or latest start:

```
~~~
SUBSCRIBE Message {
  Subscribe ID (i),
  Track Alias (i),
  Track Namespace (b),
  Track Name (b),
  Number of Parameters (i),
  Track Request Parameters (..) ...,
  Location Filter (i),
  [StartGroup (i)],
  [StartObject (i)],
  End Present(f),
  [EndGroup (i)],
  [ EndObject (i)],
}
```
Location Filter only controls Start.  End Present is a boolean flag indicating that End Group/Object are present, or the subcription is open ended.

If you don't want to support an absolute end group with current or latest start:

Add another filter mode:

AbsoluteRange 0x4

EndGroup and EndObject are only present when filter=AbsoluteRange.  Absolute (0x3) only specifies StartGroup/Object.

```
~~~
SUBSCRIBE Message {
  Subscribe ID (i),
  Track Alias (i),
  Track Namespace (b),
  Track Name (b),
  Number of Parameters (i),
  Track Request Parameters (..) ...,
  Location Filter (i),
  [StartGroup (i)],
  [StartObject (i)],
  [EndGroup (i)],
  [ EndObject (i)],
}
```

suhasHere: I prefer the second, but either works great.

suhasHere:
@afrind updated the PR .. please let me know .. thanks

fluffy:
I like that this would move us away from stealing special values from obejct id / group id space.

draft-ietf-moq-transport.md
```diff
@@ -1201,6 +1201,26 @@ OBJECT_STREAM {
...
+### Subscribe Location Filter {#sub-filter}
+
+The receiver specifies a filter value that allows publisher
+to identify the range of groups and objects that needs to be delivered, wherever applicable.
+
+There are 4 filter values:
+
+Current (0x1) : A filter value of "Current" specifies an open-ended subscription with objects from the beginning of the current group.
+
+Latest (0x2): A filter value of "Latest" specifies an open-ended subscription beginning from the current object of the current group.
```
ianswett, line 1213: I'd prefer Latest Group and Latest Object as names, because Current and Latest are quite similar.

ianswett (commented):
This looks like a good direction, but the group wanted some small changes, so I'll put those in.

draft-ietf-moq-transport.md
```diff
@@ -1211,12 +1235,13 @@ SUBSCRIBE Message {
...
-  EndGroup (i),
-  [ EndObject (i), ]
+  Filter Type (i),
+  [StartGroup (i),
+   StartObject (i)],
+  [EndGroup (i),
+   EndObject (i)],
   Number of Parameters (i),
-  Track Request Parameters (..) ...
+  Track Request Parameters (..) ...,
```
ianswett, line 1244:
```suggestion
  Track Request Parameters (..) ...
```

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1239,16 +1264,20 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
...
-* StartObject: The start Object ID, plus 1. A value of 0 means the latest object.
-This field is not present when Start Group is 0.
+* StartObject: The start Object ID within the StartGroup. Only present for
+"AbsoluteStart" and "AbsoluteRange" filter types.
 
-* EndGroup: The end Group ID, plus 1. A value of 0 means the subscription is
-open-ended and continues to the end of the track.
+* EndGroup: The end Group ID. Only present for the "AbsoluteRange" filter type.
 
 * EndObject: The end Object ID, plus 1. A value of 0 means the entire group is
```
afrind, line 1279: Heh, I just realized this encoding of inclusive is identical to exclusive on the wire (eg: if I want to end at 10, 11 goes on the wire in either case).  The only difference is the value of End Group when you want the whole group (before, ending at the end of group 7, you encoded 8/0, now it's 7/0).

draft-ietf-moq-transport.md
```diff
@@ -1201,6 +1201,30 @@ OBJECT_STREAM {
...
+There are 4 types of filters:
+
+Lastest Group (0x1) : Specifies an open-ended subscription with objects
+from the beginning of the current group.
+
+Latest Object (0x2): Specifies an open-ended subscription beginning from
+the current object of the current group.
+
+AbsoluteStart (0x3):  Specifies an open-ended subscription with objects
+beginning from the object identified in the StartGroup and StartObject fields.
```
suhasHere, line 1218: nit: may be we should converging on ones style for the naming here.

draft-ietf-moq-transport.md
```diff
@@ -1201,6 +1201,30 @@ OBJECT_STREAM {
...
 ## SUBSCRIBE {#message-subscribe-req}
 
+### Filter Types {#sub-filter}
+
+The subscriber specifies a filter on the subscription to allow
+the publisher to identify which objects need to be delivered.
+
+There are 4 types of filters:
+
+Lastest Group (0x1) : Specifies an open-ended subscription with objects
```
suhasHere, line 1211: s/Lastest/Latest

draft-ietf-moq-transport.md
ianswett: Good point.

draft-ietf-moq-transport.md
ianswett, line 1211:
```suggestion
Latest Group (0x1) : Specifies an open-ended subscription with objects
```

draft-ietf-moq-transport.md
```diff
@@ -1201,6 +1201,30 @@ OBJECT_STREAM {
...
+There are 4 types of filters:
+
+Latest Group (0x1) : Specifies an open-ended subscription with objects
+from the beginning of the current group.
+
+Latest Object (0x2): Specifies an open-ended subscription beginning from
+the current object of the current group.
+
+AbsoluteStart (0x3):  Specifies an open-ended subscription with objects
+beginning from the object identified in the StartGroup and StartObject fields.
```
ianswett, line 1218:
```suggestion
AbsoluteStart (0x3):  Specifies an open-ended subscription beginning
from the object identified in the StartGroup and StartObject fields.
```

draft-ietf-moq-transport.md
```diff
@@ -1201,6 +1201,30 @@ OBJECT_STREAM {
...
+Latest Group (0x1) : Specifies an open-ended subscription with objects
+from the beginning of the current group.
+
+Latest Object (0x2): Specifies an open-ended subscription beginning from
+the current object of the current group.
+
+AbsoluteStart (0x3):  Specifies an open-ended subscription beginning
+from the object identified in the StartGroup and StartObject fields.
+
+AbsoluteRange (0x4):  Specifies a closed subscription starting at StartObject
```
martinduke, line 1220: Would it be possible to make the end of the range exclusive?

That way, if I wanted groups 0 through N, I could request the end to be (N+1, 0) without knowing the size of N, and avoid getting an extra object.

martinduke (approved):

draft-ietf-moq-transport.md
afrind: See below, to request groups 0 - N, you say:

```
AbsoluteRange
StartGroup 0
StartObject 0
EndGroup N
EndObject 0
```
Note StartObject and EndObject = 0 mean different things

vasilvv (approved):

PR #443, opened 2024-04-29
Indicate a single group or object has been dropped on the control stream

### LINKED ISSUE #345: What if/Why will never deliver an Object/Track

ianswett:
An alternative to #434 that is extremely simple

Fixes #345 
Closes #434

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -798,6 +798,8 @@ MOQT Message {
 |-------|-----------------------------------------------------|
 | 0xE   | TRACK_STATUS ({{message-track-status}})             |
 |-------|-----------------------------------------------------|
+| 0xF   | OBJECTS_DROPPED ({{message-objects-dropped}})       |
+|-------|-----------------------------------------------------|
 | 0x10  | GOAWAY ({{message-goaway}})                         |
 |-------|-----------------------------------------------------|
 | 0x40  | CLIENT_SETUP ({{message-setup}})                    |
@@ -1622,6 +1624,25 @@ information with status code 0x04.
 The receiver of multiple TRACK_STATUS messages for a track uses the information
 from the latest arriving message, as they are delivered in order on a single
 stream.
+
+## OBJECTS_DROPPED {#message-objects-dropped}
+
+An endpoint sends a 'OBJECTS_DROPPED' message on the control stream when
+one or more Objects from a group will not be delivered for this subscription
+due to congestion or other constraints.  This does not indicate the dropped
+Objects are permanantly unavailable.
+
+~~~
+OBJECTS_DROPPED Message {
+  Subscribe ID (i),
+  Group ID (i),
+}
+~~~
+{: #moq-track-objects-dropped title="MOQT OBJECTS_DROPPED Message"}
+
+* Subscribe ID: Subscription Identifier as defined in {{message-subscribe-req}}.
+
+* Group ID: The ID of the group the subscriber does not expect any more Objects for.
 
 ## GOAWAY {#message-goaway}
 The server sends a `GOAWAY` message to initiate session migration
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1623,6 +1625,25 @@ The receiver of multiple TRACK_STATUS messages for a track uses the information
...
+OBJECTS_DROPPED Message {
+  Subscribe ID (i),
+  Group ID (i),
+}
+~~~
+{: #moq-track-objects-dropped title="MOQT OBJECTS_DROPPED Message"}
+
+* Subscribe ID: Subscription Identifier as defined in {{message-subscribe-req}}.
+
+* Group ID: The ID of the group the subscriber does not expect any more Objects for.
```
afrind, line 1645: The description makes it sound like the subscriber sends this message.  This is sent by the publisher, so maybe express it in terms of what won't be sent, and what the subscriber can expect.  

I'm also curious how it interacts with forwarding preference=track.

afrind (commented):
Individual Review:

We'd still have a message to indicate a single object drop, correct?

draft-ietf-moq-transport.md
ianswett: With forwarding preference of track, I don't think this would be needed, because if you skipped Objects, it would be very clear.

ianswett, line 1645:
```suggestion
* Group ID: The ID of the group the publisher will no longer send objects for.
```

ianswett:
> Individual Review:
> 
> We'd still have a message to indicate a single object drop, correct?

The intent here was to simplify the mechanism as much as possible, so I'd rather not unless there's a specific need?  It doesn't mean you can't drop individual objects, it just means the drop signal is on a group granularity, which I believe aligns with what we intend groups to be for.

draft-ietf-moq-transport.md
```diff
@@ -1666,6 +1668,25 @@ The receiver of multiple TRACK_STATUS messages for a track uses the information
 from the latest arriving message, as they are delivered in order on a single
 stream.
 
+## OBJECTS_DROPPED {#message-objects-dropped}
+
+An endpoint sends a 'OBJECTS_DROPPED' message on the control stream when
+one or more Objects from a group will not be delivered for this subscription
+due to congestion or other constraints.  This does not indicate the dropped
```
suhasHere, line 1675: trying to understand what could be examples of "other constraints".

suhasHere:
> Individual Review:
> 
> 
> 
> We'd still have a message to indicate a single object drop, correct?

I think this will be needed for TTL expiry cases too IIU

draft-ietf-moq-transport.md
ianswett: Flow control is one, I'm not sure what others would be.  I can imagine other resource limits coming into effect, like memory pressure, etc, but I'm not sure.

I can be more specific, but I'm not sure it's helpful?

draft-ietf-moq-transport.md
```diff
@@ -1666,6 +1670,49 @@ The receiver of multiple TRACK_STATUS messages for a track uses the information
...
+unavailable.
+
+~~~
+OBJECT_DROPPED Message {
+  Subscribe ID (i),
+  Group ID (i),
+  Object ID (i)
+}
+~~~
+{: #moq-track-objects-dropped title="MOQT OBJECT_DROPPED Message"}
```
ianswett, line 1687:
```suggestion
{: #moq-track-object-dropped title="MOQT OBJECT_DROPPED Message"}
```

ianswett:
I've added a message to indicate a single object has been dropped.

ianswett:
The current direction is to not include a specific signal for dropping, so closing this PR.

PR #444, opened 2024-04-30
The end MUST be after the start

### LINKED ISSUE #296: Empty subscriptions

ianswett:
Fixes #296

The other option is we can use a delta encoding to make it impossible to specify an earlier object, but I think that would make it more complex.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1271,8 +1271,8 @@ from the object identified in the StartGroup and StartObject fields.
 
 AbsoluteRange (0x4):  Specifies a closed subscription starting at StartObject
 in StartGroup and ending at EndObject in EndGroup.  The start and end of the
-range are inclusive.  EndGroup and EndObject MUST specify an Object after
-StartGroup and StartObject.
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

vasilvv (approved):
After or equals to?

(in case one requests a single object)

ianswett:
Good point, I changed it to the same or later object

fluffy (approved):
LGTM

PR #445, opened 2024-05-02
Add a priority field to subscriptions

### LINKED ISSUE #403: No way to prioritize one track above another
For low latency audio and video, it's helpful to be able to prioritize one track over another, whether it's audio vs video or one quality of video vs another quality of video.

This prioritization can be very helpful for ABR as well.  One might want to request the next quality up in hopes of upswitching, but continue playing the current quality until it's clear the next quality can be delivered, so the higher quality should be lower priority.

This is somewhat related to #396

ianswett:
Fixes #403 

It seems like we need this, and I'm open to suggestions, but this feels quite straightforward.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1046,9 +1046,9 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
-* Priority: Priority of the subscription relative to other active subscriptions.
-A lower number indicates Objects from the subscription SHOULD be sent before
-subscriptions of higher priorities.
+* Priority: Priority of the subscription relative to other subscriptions within
+the MoQ Session. A lower number indicates Objects from the subscription SHOULD
+be sent before subscriptions of higher priorities.
 
 * Filter Type: Identifies the type of filter, which also indicates whether
 the StartGroup/StartObject and EndGroup/EndObject fields will be present.
@@ -1111,9 +1111,9 @@ SUBSCRIBE_UPDATE Message {
 * Subscribe ID: The subscription identifier that is unique within the session.
 This MUST match an existing Subscribe ID.
 
-* Priority: Priority of the subscription relative to other active subscriptions.
-A lower number indicates Objects from the subscription SHOULD be sent before
-subscriptions of higher priorities.
+* Priority: Priority of the subscription relative to other subscriptions within
+the MoQ Session. A lower number indicates Objects from the subscription SHOULD
+be sent before subscriptions of higher priorities.
 
 * StartGroup: The start Group ID.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1288,6 +1288,7 @@ SUBSCRIBE Message {
   Track Alias (i),
   Track Namespace (b),
   Track Name (b),
+  Priority (i),
   Filter Type (i),
   [StartGroup (i),
    StartObject (i)],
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1288,6 +1288,7 @@ SUBSCRIBE Message {
   Track Alias (i),
   Track Namespace (b),
   Track Name (b),
+  Priority (i),
```
suhasHere, line 1291: Priority per subscritpion brings in few complications
- Subscriptions cannot be aggregated
- Upstream subscriptions between relays have to deal with ambiguities when conflicting priorities for the same track are requested.
- Original publisher's expectation cannot be mandated when it sets track/object priorities.

suhasHere (changes requested):

draft-ietf-moq-transport.md
ianswett: Why can't they be aggregated?  Clearly a relay will have to make a choice about what the upstream priority should be, but this at least gives it an ability to do that.

Priorities are within a track, why would one have priorities across tracks?  A subscriber may be subscribing to multiple different things at once, potentially from different namespaces, so having priorities apply between tracks is confusing.

suhasHere: in general , a way for a subscriber to express priority does makes sense. I do have few thoughts \

On aggregation: if 2 subscribers for the same track ask different priorities to a relay, the relay has to do some sort of logic to determine which one to pick. The easiest way to resolve this is "the publisher priority wins in such a case" . Perhaps we should document that in the relay section ?? Also I think the priority field needs to be optional.

On priority and track relation: Subscription is per track and if a subscriber asks for video track and audio track, it would ask that with 2 different priorities and we should allow that. isn't that the case or am I missing something ?

suhasHere: > I'm inclined not to say anything on aggregation.  CDNs and such deal with this today (HTTP priorities are similar to this) and I'm not sure being proscriptive helps.  This is very much intended as a property on a given subscription within a session, so I wouldn't expect it typically makes sense to go upstream, just like we don't act on HTTP priorities when coalescing upstream.
> 
> 
> 
> I'm fine with making it optional, but I have no idea what to make the default? This priority is purely relative, and there's no absolute meaning.

The default will be publisher set priority. 

+1 on not carrying over the priority upstream and thus applies to just the last mile. This allows for subscriptions aggregation. We should add a note to clarify this though.

LPardue (changes requested):
This is, IMO, too ambiguous about what happens when the values are equal. One option would be to bandwidth share, another would be to tie break based on some other parameter like the session ID (I.e. in HTTP we prefer lower stream IDs because that indicates an earlier request, not sure that translates to MoQ).

I'm not a huge fan of the name `Priority` because that is so overloaded. It seems like this single value is precedence.

Do we really need 2^62 levels? All the feedback I heard in the HTTP priorities discussion said that too many options is confusing. Do we have any understanding of how many concurrent sessions a MoQ use case has? Imy crystal ball suggests we can probably drop this field to 16-bits and be totally fine.

LPardue:
I might suggest that we move scheduling guidance out of field definitions and into its own section. That will make future edits simpler, for example if we take on additional refinements once there's some experience.

draft-ietf-moq-transport.md
```diff
@@ -1016,6 +1016,7 @@ SUBSCRIBE Message {
...
```
fluffy, line 1019: To me, calling this something other than priority would make more sense if the goal was to combine this with other priorities.  If this was called "preference" and relays sent things with highest preference first and then if the preferences of two thing were equal, then looked at priority to decide which next or something like this.

fluffy (changes requested):
I think we should talk about this at Seattle meeting. 

This needs to talk about how it interacts with other priorities. I need to talk about what happens upstream ( if anything ).  As it is written, I can't figure out if this works when the congested links is upstream of last relay.

ianswett:
Closing in favor of #470

PR #446, opened 2024-05-13
Add a Max Cache Duration (aka ttl) field to Objects

### LINKED ISSUE #415: Errors in second example in section 6.3.4 of draft-03
There are two issues with this example:

https://github.com/moq-wg/moq-transport/blob/80a1fc4eec92a05439e473edb60dcea399b3ca62/draft-ietf-moq-transport.md?plain=1#L1153-L1185

1. The Object id should be unique, not duplicated.
2. We shouldn't be mixing different forwarding preferences within the same track, per earlier text in this same version of the draft:

https://github.com/moq-wg/moq-transport/blob/80a1fc4eec92a05439e473edb60dcea399b3ca62/draft-ietf-moq-transport.md?plain=1#L975-L978

### LINKED ISSUE #440: Need expiry time / time to live

ianswett:
Based on discussion last week and on slack, I believe this is where the WG is heading.

It is possible this could be a per-track or per-subscription value to save bytes on the wire, particularly for the Object per Stream or Object per Datagram modes.

It's possible this value could be decreased by the cache dwell time at each hop, but I didn't hear clear consensus either way on that question.

Fixes #440
Fixes #415

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -982,6 +982,10 @@ group.
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
+* Cache Duration: The length of time in milliseconds a caching relay SHOULD
+cache an Object for.  The relay SHOULD NOT begin sending the Object if it
+has resided in the cache for longer than that time interval.
+
 * Object Status: As enumeration used to indicate missing
 objects or mark the end of a group or track. See {{object-status}} below.
 
@@ -1058,6 +1062,7 @@ OBJECT_STREAM Message {
   Group ID (i),
   Object ID (i),
   Object Send Order (i),
+  Cache Duration (i),
   Object Status (i),
   Object Payload (..),
 }
@@ -1095,6 +1100,7 @@ OBJECT_DATAGRAM Message {
   Group ID (i),
   Object ID (i),
   Object Send Order (i),
+  Cache Duration (i),
   Object Status (i),
   Object Payload (..),
 }
@@ -1127,6 +1133,7 @@ STREAM_HEADER_TRACK Message {
   Subscribe ID (i)
   Track Alias (i),
   Object Send Order (i),
+  Cache Duration (i),
 }
 ~~~
 {: #stream-header-track-format title="MOQT STREAM_HEADER_TRACK Message"}
@@ -1165,8 +1172,9 @@ have the `Object Send Order` specified in the stream header.
 STREAM_HEADER_GROUP Message {
   Subscribe ID (i),
   Track Alias (i),
-  Group ID (i)
-  Object Send Order (i)
+  Group ID (i),
+  Object Send Order (i),
+  Cache Duration (i),
 }
 ~~~
 {: #stream-header-group-format title="MOQT STREAM_HEADER_GROUP Message"}
@@ -1202,6 +1210,7 @@ STREAM_HEADER_TRACK {
   Subscribe ID = 1
   Track Alias = 1
   Object Send Order = 0
+  Cache Duration = 1000
 }
 {
   Group ID = 0
@@ -1228,6 +1237,7 @@ STREAM_HEADER_GROUP {
   Track Alias = 2
   Group ID = 0
   Object Send Order = 0
+  Cache Duration = 1000
 }
 {
   Object ID = 0
@@ -1247,6 +1257,8 @@ OBJECT_STREAM {
   Track Alias = 2
   Group ID = 0
   Object ID = 1
+  Object Send Order = 0
+  Cache Duration = 1000
   Payload = "moqrocks"
 }
 ~~~
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1253,8 +1253,8 @@ STREAM_HEADER_GROUP {
 Stream = 6
 
 OBJECT_STREAM {
-  Subscribe ID = 2
-  Track Alias = 2
+  Subscribe ID = 3
+  Track Alias = 3
   Group ID = 0
   Object ID = 1
   Object Send Order = 0
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -982,7 +982,7 @@ group.
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
-* Max Cache Duration: The maximum length of time in milliseconds a caching
+* Max Cache Duration: The maximum length of time in seconds a caching
 relay MAY cache an Object for.  A value of 0 indicates the relay should decide
 how long to cache the Object. The relay MUST NOT begin sending the Object if it
 has resided in the cache for longer than that time interval. When sending an
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1247,6 +1247,14 @@ Object, the amount of time the Object spent in cache is deducted from the
 Max Cache Duration value before the Object is sent. Exceeding the duration
 does not cause a change in the 'Object Status'.
 
+The Cache Duration is only expressed in the stream header, so the Object
+Forwarding Preference dictates the granularity of cache duration, because a
+single value applies to all Objects within the stream. For example, if a Group
+has two Objects, and the Object Forwarding Preference is 'Group', then
+only a single Max Cache Duration can be conveyed to a future subscriber,
+even if there was a substantial time gap between receiving the first and
+second Object.
+
 * Object Status: As enumeration used to indicate missing
 objects or mark the end of a group or track. See {{object-status}} below.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -983,7 +983,8 @@ object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
 * Cache Duration: The length of time in milliseconds a caching relay SHOULD
-cache an Object for.  The relay SHOULD NOT begin sending the Object if it
+cache an Object for.  A value of 0 indicates the relay should decide how
+long to cache the Object. The relay SHOULD NOT begin sending the Object if it
 has resided in the cache for longer than that time interval.
 
 * Object Status: As enumeration used to indicate missing
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -982,10 +982,12 @@ group.
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
-* Cache Duration: The length of time in milliseconds a caching relay SHOULD
-cache an Object for.  A value of 0 indicates the relay should decide how
-long to cache the Object. The relay SHOULD NOT begin sending the Object if it
-has resided in the cache for longer than that time interval.
+* Max Cache Duration: The maximum length of time in milliseconds a caching
+relay MAY cache an Object for.  A value of 0 indicates the relay should decide
+how long to cache the Object. The relay MUST NOT begin sending the Object if it
+has resided in the cache for longer than that time interval. When sending an
+Object, the amount of time the Object spent in cache is deducted from the
+Max Cache Duration value before the Object is sent.
 
 * Object Status: As enumeration used to indicate missing
 objects or mark the end of a group or track. See {{object-status}} below.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -982,6 +982,10 @@ group.
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
+* Cache Duration: The length of time in milliseconds a caching relay SHOULD
+cache an Object for.  The relay SHOULD NOT begin sending the Object if it
+has resided in the cache for longer than that time interval.
```
wilaw, line 987: Does a value of 0 mean that the object should never be cached?

If so, how does a publisher indicate that they have no suggested cache duration to convey?

ianswett: Good question.  Possibly 0 should indicate that there is no duration.  Which presumably means one can cache for however long it makes sense?

wilaw: > Which presumably means one can cache for however long it makes sense?

There are two possible (and quite different) meanings of 0:
1. Do not cache the object at all (ie a cache duration of 0 which is consistent with the interpretation of all other non-zero values
2.  The publisher has no opinion, so cache for however long you want.

ianswett: Thanks, and I agree.  In an ideal world, we'd make this optional and if it wasn't present, that would be case 2, but I don't see an easy way to allow for that.

afrind: Encode as duration + 1, and 0 means "no opinion"?  It's not pretty but it gets the job done.

Why SHOULD NOT send vs MUST NOT send?

afrind (commented):
Individual review

draft-ietf-moq-transport.md
alvestrand: https://www.rfc-editor.org/rfc/rfc9000#section-1.3 describes an optional field encoding. Why not use it?

draft-ietf-moq-transport.md
```diff
@@ -1165,8 +1172,9 @@ have the `Object Send Order` specified in the stream header.
 STREAM_HEADER_GROUP Message {
   Subscribe ID (i),
   Track Alias (i),
-  Group ID (i)
-  Object Send Order (i)
+  Group ID (i),
+  Object Send Order (i),
+  Cache Duration (i),
```
suhasHere, line 1177: This doesn't allow per object cache duration for stream per group /stream per track and excepts all the objects within a stream should have same expiry.

draft-ietf-moq-transport.md
```diff
@@ -1127,6 +1133,7 @@ STREAM_HEADER_TRACK Message {
   Subscribe ID (i)
   Track Alias (i),
   Object Send Order (i),
+  Cache Duration (i),
```
suhasHere, line 1136: This doesn't allow per object cache duration for stream per group /stream per track and excepts all the objects within a stream should have same expiry.

draft-ietf-moq-transport.md
afrind: The RFC 9000 notation just states that field *is* present sometimes and not others, but doesn't include a mechanism for a receiver to determine whether the field is present or not.

suhasHere (commented):
This seems to be a good start and we should consider taking this forward. 
However,  I feel we need to have mechanisms to specify object duration applied to objects for multi-object transport stream mappings. 

May be something like (for stream per track)
```
STREAM_HEADER_TRACK Message {
  Subscribe ID (i)
  Track Alias (i),
  Object Send Order (i),
 Cache Duration (i)
}
{
  Group ID (i),
  Object ID (i),
  Object Payload Length (i),
  [Object Status (i)],
  [Cache Duration(i)],
  Object Payload (..),
}

The cache duration is considered iff payload_length > 0 && not available in STREAM_HEADER_TRACK message
```
or some better way to do so.

draft-ietf-moq-transport.md
afrind: @suhasHere : That is also how I read this.  Does your application require unique cache duration per object within a group-per-stream, and if so, can you explain the use case the drives it?

VMatrix1900:
IIUC, this field is the TTL of per hop instead of end-to-end. The content origin can set the TTL to end-to-end [deadline](https://datatracker.ietf.org/doc/draft-ma-moq-relay-for-deadline/) and each hop decrease the value then pass it to the downstream relay. Is that correct?

vasilvv:
I'm not sure this PR solves either of the problems we're trying to solve with the cache TTL fields.

fluffy (approved):
I think this is good start and we should merge this. We need more things over time.  I am fine with 0 and 1 being special values for don't care , and don't cache.

suhasHere (dismissed):
LGTM

draft-ietf-moq-transport.md
suhasHere: I feel object properties and stream mapping are being overloaded. Object property like duration is application scoped property that indicates how long a object is valid. Stream mapping OTOH is transport mechanism.

draft-ietf-moq-transport.md
kixelated: That's absolutely the point. Introducing gaps in a QUIC stream is counterproductive.

draft-ietf-moq-transport.md
kixelated: > This doesn't allow per object cache duration for stream per group /stream per track and excepts all the objects within a stream should have same expiry.

That's absolutely the point. The ability to introduce gaps in a QUIC stream is counterproductive, and the application is explicitly trying to use a QUIC stream to avoid these gaps.

kixelated: > I feel object property and stream mapping are being mixed

I agree, the object model and stream mapping are in direct conflict and it needs to be fixed. The entire point of the stream mapping is to inherit the properties of QUIC streams, but that's pointless if the object model vetoes it.

ianswett: I think the goal of the stream mapping is to efficiently provide the transport features the application needs.  In some cases, HoL blocking is a feature, and sometimes it's a bug/negative for example.

suhasHere: +1 to what @ianswett said

draft-ietf-moq-transport.md
```diff
@@ -982,6 +982,14 @@ group.
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
+* Max Cache Duration: The maximum length of time in milliseconds a caching
+relay MAY cache an Object for.  A value of 0 indicates the relay should decide
+how long to cache the Object. The relay MUST NOT begin sending the Object if it
+has resided in the cache for longer than that time interval. When sending an
+Object, the amount of time the Object spent in cache is deducted from the
+Max Cache Duration value before the Object is sent. Exceeding the duration
+does not cause a change in the 'Object Status'.
```
suhasHere, line 991: I am fine with this, but don't see if we need to advice on deducting the time spent each hop.

draft-ietf-moq-transport.md
ianswett: Changed to MUST NOT BTW

draft-ietf-moq-transport.md
ianswett: Agreed.  I'd like to see if this is really necessary, since it adds more framing overhead and as I mentioned in the call, I think it might end up being better to use priorities to achieve what you're trying to achieve with TTL.

draft-ietf-moq-transport.md
ianswett: If we expect this to prevent an Object from being in cache for too long, I think we need to do the subtraction?

draft-ietf-moq-transport.md
```diff
@@ -1239,6 +1239,14 @@ group.
 object. The preferences are Track, Group, Object and Datagram.  An Object MUST
 be sent according to its `Object Forwarding Preference`, described below.
 
+* Max Cache Duration: The maximum length of time in seconds a caching
```
fluffy, line 1242: I would prefer milli seconds to seconds.

ianswett: I changed it because HTTP MaxAge is in seconds, and In the Object per Stream or Object per Datagram cases, using seconds saves at least a byte.

But I can change it back to ms if we're fine with the extra byte+ of overhead.

draft-ietf-moq-transport.md
suhasHere: agreed .. other than milliseconds vs seconds , this looks fine to me.

draft-ietf-moq-transport.md
```diff
@@ -1422,8 +1433,9 @@ have the `Object Send Order` specified in the stream header.
 STREAM_HEADER_GROUP Message {
   Subscribe ID (i),
   Track Alias (i),
-  Group ID (i)
-  Object Send Order (i)
+  Group ID (i),
+  Object Send Order (i),
+  Max Cache Duration (i),
```
fluffy, line 1438: I don't get how this will work with when combined with the removing the amount of time that has elapsed sitting in cache. Consider a stream per track mode where the timeout is set to 30 minutes from the use case Will had.  The first object I will call A is published at noon and a later object is polished at 15 minutes later at 12:15 and I will call this object B. 

Then a downstream relay subscribes 10 minutes after start to the track. It gets a stream header with a 20 minute time out since we are substracting the 10 min that track was sitting in ubstream realy. For object A, that is great, it will time out at 12:30. However, for object B, it seems it will time out 20 minutes after it was publshed, not 30 as request. So it will time out at 12:30 when it should be around will 12:40. 

Not sure I am explaining this well but hope you get idea.

fluffy: I am trying to say we need more clarity how the time is decrement for time in cache in the stream per track and stream per group modes. This problem only come up because the timeout is in the header not the object in those mode. Moving might be the easy fix.

draft-ietf-moq-transport.md
ianswett: I changed it back to ms.

draft-ietf-moq-transport.md
ianswett: I think moving them to be per-Object may create some other weird circumstances, such as the first few Objects of a Group/Track expiring, but the ones that arrived later haven't, but they aren't useful without the start of the Group.

ianswett: I updates this PR with another paragraph, PTAL

draft-ietf-moq-transport.md
```diff
@@ -1239,6 +1239,22 @@ group.
...
+Max Cache Duration value before the Object is sent. Exceeding the duration
+does not cause a change in the 'Object Status'.
+
+The Cache Duration is only expressed in the stream header, so the Object
+Forwarding Preference dictates the granularity of cache duration, because a
+single value applies to all Objects within the stream. For example, if a Group
+has two Objects, and the Object Forwarding Preference is 'Group', then
+only a single Max Cache Duration can be conveyed to a future subscriber,
+even if there was a substantial time gap between receiving the first and
+second Object.
```
ianswett, line 1256:
```suggestion
second Object in the Group.
```

draft-ietf-moq-transport.md
```diff
@@ -1239,6 +1239,22 @@ group.
...
+Max Cache Duration value before the Object is sent. Exceeding the duration
+does not cause a change in the 'Object Status'.
+
+The Cache Duration is only expressed in the stream header, so the Object
+Forwarding Preference dictates the granularity of cache duration, because a
+single value applies to all Objects within the stream. For example, if a Group
+has two Objects, and the Object Forwarding Preference is 'Group', then
+only a single Max Cache Duration can be conveyed to a future subscriber,
+even if there was a substantial time gap between receiving the first and
+second Object in the Group.
```
vasilvv, line 1256: Does this imply that you can have the beginning of the group expire before the end of the group?

vasilvv (commented):
It is still unclear to me which of many problems this specific field is trying to solve.

fluffy (changes requested):
still want to talk about this some more - there has been many changes

afrind:
Individual Review:

I think this PR needs additional, probably normative, guidance about when a caching relay starts the expiration timer.  For example, if using stream-per-object and the object is large, does the timer start when placing the first or last byte of the object in cache.  The same is true of other forwarding modes.

fluffy:
Just for notes ... Mo raised key point on call that if we tail drop things in some priority mechanism, then we may never get the end of group marker. We probably need to adjust text to be very careful about how we detect we are never going to get any more data on this group or track

fluffy:
Victor had point that perhaps we should use the start of next group to start timer instead end of current group

fluffy (approved):

ianswett:
Closing this in favor of a newer PR, #469

PR #447, opened 2024-05-13
Organize messages by Publisher and Subscriber

### LINKED ISSUE #309: Classify messages by ROLE
Here is the SUBSCRIBE family of messages:

```rust
SUBSCRIBE = 0x3,
SUBSCRIBE_OK = 0x4,
SUBSCRIBE_ERROR = 0x5,
UNSUBSCRIBE = 0xa,
SUBSCRIBE_FIN = 0xb,
SUBSCRIBE_RESET = 0xc,
```

I have two separate modules for publishing and subscribing as they are fully* independent in MoqTransport. Additionally, some applications will only publish or subscribe, so it's not necessary to implement/expose the full API surface.

However, at a glance, it's not clear which messages are sent by each role. I think it would help a lot if we classified which role sends which messages:

```rust
// Sent by a subscriber
SUBSCRIBE = 0x3,
UNSUBSCRIBE = 0xa,

// Sent by a publisher
SUBSCRIBE_OK = 0x4,
SUBSCRIBE_ERROR = 0x5,
SUBSCRIBE_FIN = 0xb,
SUBSCRIBE_RESET = 0xc,
```

For example, I think it makes it a lot clearer that a subscriber needs to accept 3(!) possible error messages.

### LINKED ISSUE #399: Organize messages by role
The discussion on PR #395 suggested a reorg of messages and how to discuss them in the Role section.

> _afrind_ 
> If we reorganized the document into "publisher messages" and "subscriber messages", then the text here (and below) could be improved. We could say:
> 
> The endpoint MUST NOT send any subscriber messages. If the endpoint receives any publisher messages, it MUST close the connection with "error code TBD".
> 
>  _kixelated_ 
> +1 I would love to see this.
> 
>  _kixelated_ 
> A simple way of doing it would be to add a "sent by (role)" column to the table of messages and IDs.

ianswett:
This only moves around the text, but I could create section titles for Publisher and Subscriber messages.

I can also add a column to the message type table if people would like.

Fixes #399 
Fixes #309

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -754,8 +754,8 @@ limits.  See section 2.2 in {{QUIC}}.
 
 # Messages {#message}
 
-Both unidirectional and bidirectional streams contain sequences of
-length-delimited messages.
+Unidirectional streams and datagrams contain Objects and all other
+messages are sent on a bidirectional stream.
 
 An endpoint that receives an unknown message type MUST close the session.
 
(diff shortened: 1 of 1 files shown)


fluffy (approved):
LGTM

ianswett:
I merged this to avoid conflicts, but I will do a follow-up or two.

PR #450, opened 2024-05-22
Max Cache Duration per Subscription

### LINKED ISSUE #440: Need expiry time / time to live

ianswett:
A variant of #446 that is on a Subscription granularity, instead of Object.

Fixes #440

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1400,6 +1400,7 @@ SUBSCRIBE_OK
 {
   Subscribe ID (i),
   Expires (i),
+  Max Cache Duration (i),
   ContentExists (f),
   [Largest Group ID (i)],
   [Largest Object ID (i)]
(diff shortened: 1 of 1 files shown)


suhasHere:
Per my read of the room, i would prefer #446  as it is gives better flexibility for the application than enforcing it per subscription ( as it applies the same value to all the objects) . Also not sure if there are cases where this can change between the subscriptions  either

VMatrix1900:
> Per my read of the room, i would prefer #446 as it is gives better flexibility for the application than enforcing it per subscription ( as it applies the same value to all the objects) . Also not sure if there are cases where this can change between the subscriptions either

+1, Cache Duration per object is more flexible.

fluffy (changes requested):
I don't understand how the relay would know the answer to this at the time the subscribe happened, and I don't understand what use case the applications would need to get this from the relay instead of some other way like the manifest.

draft-ietf-moq-transport.md
```diff
@@ -1519,6 +1519,7 @@ SUBSCRIBE_OK
 {
   Subscribe ID (i),
   Expires (i),
+  Max Cache Duration (i),
```
fluffy, line 1522: I think this might be slightly better as an optional parameter but clearly not a big deal either way

draft-ietf-moq-transport.md
```diff
@@ -1533,6 +1534,13 @@ longer valid. A value of 0 indicates that the subscription does not expire
 or expires at an unknown time.  Expires is advisory and a subscription can
 end prior to the expiry time or last longer.
 
+* Max Cache Duration: The maximum length of time in milliseconds a caching
+relay MAY cache the subscription for.  A value of 0 indicates there is no limit.
```
fluffy, line 1538: It is the objects that are cached not the subscription so I think this needs to say "MAY cache the object for"

draft-ietf-moq-transport.md
```diff
@@ -1533,6 +1534,13 @@ longer valid. A value of 0 indicates that the subscription does not expire
...
+The relay MUST NOT begin sending Objects for a new subscription from cache if
+it's older than Max Cache Duration. When a relay responds with SUBSCRIBE_OK,
```
fluffy, line 1540: Need to be clear the time starts at when the relay started to receive the object.

draft-ietf-moq-transport.md
```diff
@@ -1533,6 +1534,13 @@ longer valid. A value of 0 indicates that the subscription does not expire
...
+the amount of time since the corresponding upstream subscription was made is
+subtracted when sending the Max Cache Duration.
```
fluffy, line 1542: This sentence does not make sense if we are caching objects not subscriptions. Would just remove it.

fluffy (changes requested):
I think this was written before our current understanding of caching object not subscriptions so I have proposed a few updates but if it is meant to still be about refreshing subscriptions, then ignore all my comments.

ianswett:
Closing this in favor of #469

Issue #422, opened 2024-03-23
Redundant object_id with stream per group


kixelated:
When sending a stream per group, the object_id MUST increment by 1 based on my interpretation of the draft:

> Object ID: The order of the object within the group. The IDs starts at 0, increasing sequentially for each object within the group.

and

> A relay MUST not reorder or drop objects received on a multi-object stream when forwarding to subscribers, unless it has application specific information.

Technically, the broadcaster could produce a stream that doesn't increment by one, but that doesn't make any sense and feels like a loophole to close. I much prefer things to be sequential and I don't want the application stuffing context into these relay-specific headers anyway.

If this makes sense, then we can optimize the wire encoding. **Proposed**:
```
STREAM_HEADER_GROUP Message {
  Subscribe ID (i),
  Track Alias (i),
  Group ID (i)
  + First Object ID (i)
  Object Send Order (i)
}

{
  - Object ID (i),
  Object Payload Length (i),
  Object Payload (..),
}
```

We need to know the ID of the first object to handle subscriptions that start at a non-zero object. Otherwise, the receiver does +1 to compute the next object's ID. We'll save some bytes for larger groups but IMO more importantly, rule out some illegal encodings.

fluffy:
I think sending object ID out of order may end up being common for scalable video codecs. We should talk about what the case is her before deciding what to do with this.

kixelated:
> I think sending object ID out of order may end up being common for scalable video codecs. We should talk about what the case is her before deciding what to do with this.

That implies multiple streams per group, which I thought we ruled out. Otherwise you can't prioritize them separately.

I think SVC layers should be separate tracks. That way you can also subscribe to them independently.

afrind:
Individual Comment:

Is there a problem with this optimization if the publisher makes an object but doesn't send it for some reason?  I think I recall Jordi's implementation does that.

kixelated:
> Individual Comment:
> 
> Is there a problem with this optimization if the publisher makes an object but doesn't send it for some reason? I think I recall Jordi's implementation does that.

I think it comes down to the layer of  `object_id` and the intended usage.

In my opinion, `object_id` is for reassembly and gap detection when there's an object per stream/datagram. It's much like QUIC's `STREAM offset` in that regard, auto incrementing as each object is sent/received. It's also why my original warp draft didn't have a concept of object ID; everything was based on stream offsets.

[...]

PR #452, opened 2024-05-27
Delivery Order (aka Fetch lives)

### LINKED ISSUE #269: Multiple subscriptions and updating a subscription

### LINKED ISSUE #326: Hints & sendOrder how can they work together

### LINKED ISSUE #350: When to use Track Alias or Subscribe ID?

### LINKED ISSUE #396: How does a relay/publisher know what to prioritize or drop?

### LINKED ISSUE #419: Reliable Live use-case

### LINKED ISSUE #431: Is Forwarding Preference Independent of Use Case?
In his review of #426, @kixelated  indicated that some of the forwarding preferences seemed illogical for some use case. Indeed, the PR makes no judgments about what forwarding preferences make sense with different drop policies; it seeks a way to express missing objects and groups in any forwarding pref.

My reading of the current draft is that forwarding preference is an immutable track property. If you want to distribute a track for a use case that requires a different forwarding pref, you need a duplicate track. Maybe that's not explicit or what was intended; it could, in principle, be a per-subscription property based on the sender's understanding of the use case embodied by that subscription (or fetch, if we have those).

Or maybe we want to say that certain forwarding preferences imply certain behaviors. Maybe a track stream simply does not allow omitting objects, so no missing object marker is necessary for that format.

I don't have any preferred approach here, but we should agree about what is implied by a forwarding preference.

### LINKED ISSUE #442: Are end-of-group messages mandatory?
#429 introduced the Status field for Objects, which can have a value that signals the end of a group (0x3). 

The draft states:

> This is sent right after the last object in the group. 

I read this to mean that whenever a publisher publishes the last object in a group, it also publishes the end-of-group marker (eg: it is mandatory).   However, it seems like this might create high overhead - eg in an audio only track where every sample is its own group.

For a fanout relay handling group-per-stream tracks, the end of group marker is a signal that it can close the downstream streams.  Yes, the relay needs a timeout as well, to catch malicious or buggy producers, but I'd prefer that be the rare, and proactively clean up the state since it could be quite large.

The other modes seem less critical for a relay, I think? However, I worry about interoperability between producers and consumers if we don't make it required in other stream modes.  That could be part of a streaming format specification rather than the core transport.

Given that datagrams are inherently unreliable, I don't know that we need to mandate end-of-group markers there.

ianswett:
Takes an opinionated approach to adding 'Fetch' style functionality within the existing mechanisms.

Can be split up if people prefer.

Pulls in concepts from #411, #428 and #449 (Delivery Timeout)

Fixes #269
Fixes #326
Fixes #350
Fixes #396 
Fixes #419
Fixes #431 
Fixes #442

Relevant to #422, because we can omit the Object ID for 'In Order' delivery mode.

Closes #421
Closes #449

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1012,14 +1012,16 @@ completeness and Latest prioritizes receiving the most recent Objects.
 
 There are 2 delivery orders:
 
-In Order (0x1): Delivers Objects in ascending Group Id, and then ascending
-Object Id order, as fast as possible.  Objects are sent on a single stream at
+In Order (0x1): Delivers Objects in order of ascending Group Id and then
+ascending Object Id, as fast as possible.  Objects are sent on a single stream at
 once, starting with the 'Stream Header Track' header. When an Object or Group
 is no longer available, the status ({{object-status}}) indicates that.
 A SUBSCRIBE_UPDATE could narrow a subscription and it could make sense to skip
 over ranges of Objects. In this case, a publisher MAY reset the existing stream
 and start sending on a new one, with the same Subscribe ID.  Publishers MUST NOT
-skip Objects otherwise.
+skip Objects otherwise.  Because there are no unexplained Object gaps, it is
+not necessary to send Objects with the 'End of Group' status or 'End of Track'
+status.
 
 Latest (0x2): Objects from the most recent Group SHOULD be delivered first,
 in ascending Object ID order when possible, unless indicated otherwise by
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1592,7 +1592,7 @@ end prior to the expiry time or last longer.
 
 * Delivery Timeout: The number of milliseconds the publisher has agreed to attempt
 forwarding objects, or zero to indicate no limit.  MUST be 0 when the SUBSCRIBE
-specified a 'Delivery Order' of 'Strict In Order'. See ({{delivery-timeout}}) above.
+requested a 'Delivery Order' of 'Strict In Order'. See ({{delivery-timeout}}) above.
 
 * ContentExists: 1 if an object has been published on this track, 0 if not.
 If 0, then the Largest Group ID and Largest Object ID fields will not be
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1043,6 +1043,7 @@ would be sent as soon as the publisher or relay received them.
 
 ### Delivery Timeout {#delivery-timeout}
 
+The maximum number of milliseconds a publisher should attempt forwarding objects.
 The semantics of the timeout depend on the forwarding preference
 ({{object-message-formats}}) of the track:
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1004,6 +1004,30 @@ object than StartGroup and StartObject.
 
 A filter type other than the above MUST be treated as error.
 
+### Delivery Order {#delivery-order}
+
+A subscriber can prefer different Object delivery orders depending upon the
+use case.  Their responses to congestion are different: In Order prioritizes
+completeness and Latest prioritizes receiving the most recent Objects.
+
+There are 2 delivery orders:
+
+In Order (0x1): In Order subscriptions are delivered as fast as possible on
+a single stream at once, starting with the 'Stream Header Track' header.
+When an Object or Group is no longer available, the status ({{object-status}})
+should indicate that. A SUBSCRIBE_UPDATE could narrow a subscription and
+it could make sense to skip over ranges of Objects. In this case, a publisher
+MAY reset the existing stream and start sending on a new one, with the same
+Subscribe ID.  Publishers MUST NOT skip Objects otherwise.
+
+Latest (0x2): Objects from the most recent Group SHOULD be delivered first,
+in ascending Object ID order when possible, unless indicated otherwise by
+an Object's 'Object Send Order' or another mechanism.  Objects MAY be delivered
+via OBJECT_STREAM, OBJECT_DATAGRAM, or STREAM_HEADER_GROUP streams, but
+MUST NOT change the delivery preference during a subscription.
+A subscriber MUST NOT make two simultaneous 'Latest' subscriptions for the
+same Track.
+
 
 ### SUBSCRIBE Format
 A receiver issues a SUBSCRIBE to a publisher to request a track.
@@ -1021,6 +1045,7 @@ SUBSCRIBE Message {
    StartObject (i)],
   [EndGroup (i),
    EndObject (i)],
+  Delivery Order (i),
   Number of Parameters (i),
   Track Request Parameters (..) ...
 }
@@ -1059,6 +1084,9 @@ requested. Only present for "AbsoluteStart" and "AbsoluteRange" filter types.
 
 * EndObject: The end Object ID, plus 1. A value of 0 means the entire group is
 requested. Only present for the "AbsoluteRange" filter type.
+
+* Delivery Preference: Indicates how Objects should be transmitted and dropped.
+See ({{delivery-preference}}) above.
 
 * Track Request Parameters: The parameters are defined in
 {{version-specific-params}}
@@ -1310,7 +1338,6 @@ serialize object fields below, and terminate the stream.
 
 ~~~
 OBJECT_STREAM Message {
-  Subscribe ID (i),
   Track Alias (i),
   Group ID (i),
   Object ID (i),
@@ -1320,8 +1347,6 @@ OBJECT_STREAM Message {
 }
 ~~~
 {: #moq-transport-object-stream-format title="MOQT OBJECT_STREAM Message"}
-
-* Subscribe ID: Subscription Identifier as defined in {{message-subscribe-req}}.
 
 * Track Alias: Identifies the Track Namespace and Track Name as defined in
 {{message-subscribe-req}}.
@@ -1347,7 +1372,6 @@ will be dropped.
 
 ~~~
 OBJECT_DATAGRAM Message {
-  Subscribe ID (i),
   Track Alias (i),
   Group ID (i),
   Object ID (i),
@@ -1375,15 +1399,11 @@ TODO: figure out how a relay closes these streams
 
 When a stream begins with `STREAM_HEADER_TRACK`, all objects on the stream
 belong to the track requested in the Subscribe message identified by `Subscribe
-ID`.  All objects on the stream have the `Object Send Order` specified in the
-stream header.
-
+ID`.
 
 ~~~
 STREAM_HEADER_TRACK Message {
   Subscribe ID (i)
-  Track Alias (i),
-  Object Send Order (i),
 }
 ~~~
 {: #stream-header-track-format title="MOQT STREAM_HEADER_TRACK Message"}
@@ -1414,13 +1434,12 @@ previously sent Group ID on that stream, or if its Object ID is less than or
 equal to a previously sent Object ID within a given group on that stream.
 
 When a stream begins with `STREAM_HEADER_GROUP`, all objects on the stream
-belong to the track requested in the Subscribe message identified by `Subscribe
-ID` and the group indicated by `Group ID`.  All objects on the stream
+belong to the track requested in the Subscribe message identified by `Track
+Alias` and the group indicated by `Group ID`.  All objects on the stream
 have the `Object Send Order` specified in the stream header.
 
 ~~~
 STREAM_HEADER_GROUP Message {
-  Subscribe ID (i),
   Track Alias (i),
   Group ID (i)
   Object Send Order (i)
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1086,8 +1086,8 @@ requested. Only present for "AbsoluteStart" and "AbsoluteRange" filter types.
 * EndObject: The end Object ID, plus 1. A value of 0 means the entire group is
 requested. Only present for the "AbsoluteRange" filter type.
 
-* Delivery Preference: Indicates how Objects should be transmitted and dropped.
-See ({{delivery-preference}}) above.
+* Delivery Order: Indicates how Objects should be transmitted and dropped.
+See ({{delivery-order}}) above.
 
 * Track Request Parameters: The parameters are defined in
 {{version-specific-params}}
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1533,6 +1590,10 @@ longer valid. A value of 0 indicates that the subscription does not expire
 or expires at an unknown time.  Expires is advisory and a subscription can
 end prior to the expiry time or last longer.
 
+* Delivery Timeout: The number of milliseconds the publisher has agreed to attempt
+forwarding objects, or zero to indicate no limit.  MUST be 0 when the SUBSCRIBE
+specified a 'Delivery Order' of 'Strict In Order'. See ({{delivery-timeout}}) above.
```
ianswett, line 1595:
```suggestion
requested a 'Delivery Order' of 'Strict In Order'. See ({{delivery-timeout}}) above.
```

draft-ietf-moq-transport.md
```diff
@@ -1004,6 +1004,61 @@ object than StartGroup and StartObject.
...
+
+There are 3 delivery orders:
+
+Strict In Order (0x1): Delivers Objects in order of ascending Group Id and then
+ascending Object Id, as fast as possible.  Objects are sent on a single stream at
+once, starting with the 'Stream Header Track' header. When an Object or Group
+is no longer available, the status ({{object-status}}) indicates that.
+A SUBSCRIBE_UPDATE could narrow a subscription and it could make sense to skip
+over ranges of Objects. In this case, a publisher MAY reset the existing stream
+and start sending on a new one, with the same Subscribe ID.  Publishers MUST NOT
```
afrind, line 1021: The MAY + MUST NOT is somewhat confusing. I think you are trying to say the publisher MAY continue to send objects that are no longer in the subscription, or reset the stream and open a new one in order to skip them?

draft-ietf-moq-transport.md
```diff
@@ -1004,6 +1004,61 @@ object than StartGroup and StartObject.
...
+skip Objects otherwise.  Because there is no explicit Group ID, publishers MUST
+send Objects with the 'End of Group' status to indicate the end of the Group.
```
afrind, line 1023: Why do you say there is no explicit group id?  Don't the objects in Stream-per-track contain the group ID?

draft-ietf-moq-transport.md
```diff
@@ -1004,6 +1004,61 @@ object than StartGroup and StartObject.
...
+The 'Delivery Timeout' field is not present for this (0x1) mode.
+
+Latest In Order (0x2): Objects that have not exceeded the 'Delivery Timeout'
+({{delivery-timeout}}) SHOULD be delivered in order.  Objects can be delivered
```
afrind, line 1027: Can you clarify the expected behavior when a relay has a gap in their cache?  eg: The relay has object 1, 2 and 4.  

My read is that the relay MUST go upstream at least once to fill this gap, or get a tombstone for it.

draft-ietf-moq-transport.md
```diff
@@ -1004,6 +1004,61 @@ object than StartGroup and StartObject.
...
+in ascending Object ID order when possible, unless indicated otherwise by
+an Object's 'Object Send Order' or another mechanism.  Objects MAY be delivered
+via OBJECT_STREAM, OBJECT_DATAGRAM, or STREAM_HEADER_GROUP streams, but
+MUST NOT change the delivery preference during a subscription.
+
+A subscriber MUST NOT make two simultaneous subscriptions of type
+'Latest In Order' or 'Latest' for the same Track, because both could be
+delivered by the same Track Alias and they would compete for bandwidth
+and other resources.  When bandwidth is sufficient, a 'Latest In Order' and
+'Latest' subscription would be delivering the same Objects, because Objects
```
afrind, line 1041: What about subscribes behind the live head?  Wouldn't the two modes deliver the resources in opposite orders?

afrind (commented):
Individual Review

draft-ietf-moq-transport.md
suhasHere: Wonder why does this not apply for stream per group or stream per object cases.

draft-ietf-moq-transport.md
```diff
@@ -1005,6 +1005,61 @@ object than StartGroup and StartObject.
...
+once, starting with the 'Stream Header Track' header. When an Object or Group
+is no longer available, the status ({{object-status}}) indicates that.
+A SUBSCRIBE_UPDATE could narrow a subscription and it could make sense to skip
+over ranges of Objects. In this case, a publisher MAY reset the existing stream
+and start sending on a new one, with the same Subscribe ID.  Publishers MUST NOT
+skip Objects otherwise.  Because there is no explicit Group ID, publishers MUST
+send Objects with the 'End of Group' status to indicate the end of the Group.
+The 'Delivery Timeout' field is not present for this (0x1) mode.
+
+Latest In Order (0x2): Objects that have not exceeded the 'Delivery Timeout'
```
suhasHere, line 1027: by "latest" do we mean, new groups will be prioritized before prior groups?

draft-ietf-moq-transport.md
```diff
@@ -1005,6 +1005,61 @@ object than StartGroup and StartObject.
...
+({{delivery-timeout}}) SHOULD be delivered in order.  Objects can be delivered
+via OBJECT_STREAM, OBJECT_DATAGRAM, or STREAM_HEADER_GROUP streams, but
+MUST NOT change the delivery preference during a subscription.
+
+Latest (0x3): Objects from the most recent Group SHOULD be delivered first,
```
suhasHere, line 1032: Latest here is more clear compared to the previous option.

suhasHere (changes requested):
Thanks for the PR. I feel this PR is hitting some of the cases we need. 
I think this PR does need some more work to clarify few options - basically the interactions between priority across the delivery order options, what to do if there are gaps and cache duration. Relay upstreaming a subscription and what to pick or is it scoped to just last mile.

May be this is all some text clarification. But as-is I am not clear it addresses the issue completely.

draft-ietf-moq-transport.md
afrind: @suhasHere my read of this mode is that it overrides the publisher preference.  If subscriber asks for in-order, then the track is delivered in a single stream regardless.

suhasHere: ah! makes sense .. I missed that.

draft-ietf-moq-transport.md
suhasHere: +1 to @afrind  . we need some text to explain when to go upstream vs when to deliver what is there ? May be its already there and I am not seeing it ( so apologies if that's the case)

fluffy (changes requested):
I think we should talk about this at the Seattle interim starting with use cases we are trying to solve.

It is really hard to review a proposal like this without understanding if / how this propagates upstream.

draft-ietf-moq-transport.md
englishm: To me, this is the key question, not just in this context, but re: caching expectations as a whole. 

What behavior do we expect when a relay is serving active subscribers at the live edge, but a new subscribe comes in so far behind the live edge that all of the delivery timeouts have expired and none of the objects are in cache any longer? Then, what about the case where _some_ objects haven't yet expired, but others have? Should that behavior be any different?

draft-ietf-moq-transport.md
```diff
@@ -1005,6 +1005,61 @@ object than StartGroup and StartObject.
...
 
+### Delivery Order {#delivery-order}
+
+A subscriber can prefer different Object delivery orders depending upon the
+use case.  Their responses to congestion are different: In Order prioritizes
+completeness and Latest prioritizes receiving the most recent Objects.
+
+There are 3 delivery orders:
+
+Strict In Order (0x1): Delivers Objects in order of ascending Group Id and then
```
wilaw, line 1016: How can a relay implement this requirement that it deliver Groups in ascending order? Our group IDs are non-contiguous. As a consequence, if a relay receives group #10 and then group #12, should it wait for Group #11? It can never tell if 11 exists or not.

fluffy (changes requested):
I think we need to step back from this PR and talk about use cases and requirements with some face to face time. Right now I am strongly against merging this for a few reasons. 

1. This eliminates the mode mode that is actually used the most in existing conferencing systems and the mode I think we need here

2. All the data I have see indicates that latest is not needed and performs worse that other approaches 

3. Strict In Order seems like it would be extremely memory intensive to try and implement and I do not see the need for it. We have to buffer somewhere, clear the receiving clients, not the high scale relays, are the place to do that.  This seems like it would introduce all kinds of DOS potential.  System that want things to be in order inside a group will use stream per group and get this but the relays do not need to enforce it.

ianswett:
Most of these changes are in other PRs, so closing this.

PR #454, opened 2024-05-29
Remove special value for SUBSCRIBE StartObject

### LINKED ISSUE #453: Special Value for StartObject and EndObject for SUBSCRIBE
I'm not totally sure what to do with the special value of 0 for StartObject and EndObject.

If StartObject == 0, that's requesting the whole group. But if it's 1, that maps to a first object of zero, which is also consistent with  requesting the whole group.

For EndObject, the special value makes sense, because the subscriber might not know how many objects there are.

There's also the error case where StartObject == 0 and EndObject != 0, or the reverse.

IMO it would make more sense for EndObject to have the special value at zero, but not StartObject.

martinduke:
Fixes #453

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1107,11 +1107,9 @@ This MUST match an existing Subscribe ID.
 
 * StartGroup: The start Group ID.
 
-* StartObject: The start Object ID, plus 1. A value of 0 means the entire group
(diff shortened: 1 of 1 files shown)


ianswett (commented):
Can you fix SUBSCRIBE_UPDATE as well?

ianswett (approved):

Issue #438, opened 2024-04-21
Need better terminology for endpoints that are not relays 


fluffy:
Need a term for upstream, downstream and the original publisher that is not a relay and the final subscriber that is not a relay.

afrind:
We also have parked issue #157 referencing some terminology.  This comment https://github.com/moq-wg/moq-transport/issues/157#issuecomment-1553718108 seems to indicate that "Producer" and "Consumer" were meant to be the original producer/final subscriber, and perhaps even said that at one point, but the current doc doesn't reflect that.

fluffy:
Yah, my recollection was we failed to come to consensus on that one way or the other.  I certainly don't care what we call it but time and time again in describing this we need a few different words. 

Say  C is video display subscribed to a relay called B subscribed to a video camera called A. ( So  A sends video to B that sends it to C). We need words for:

1. The publisher like thing that is uniquely are original source like A. ( and origin server is not a good word for this) 

2. The final subscriber thing that is C in this case.

3. A thing that is received a subscribe but does not care if it came from a relay or not. So both A and B in this case gets subscribes and have a tone of behavior that is the same. 

4. a thing that is receiving object based on a subscribe and does not care if they are coming from a relay or not. So a word that covers both B and C in this case.

wilaw:
> The publisher like thing that is uniquely are original source like A. ( and origin server is not a good word for this)

I do actually like the term 'origin' for this :) Reasons:
1. It is a singleton - there is only one 'origin', while there can be many publishers, consumers, clients and servers. 
2. It defines its uniqueness - that of being the original source - while not overloading that with its behaviors, such as publish, or produce, which are shared by other nodes. 

> The final subscriber thing that is C in this case.

This is hard. I tend to use 'end client' or 'end consumer'. 

> Need a term for upstream, downstream

These seem clear enough - 'upstream' = towards the origin, 'downstream' = towards the end  client. Are you talking more about the node that is in that direction, or the direction itself?

TimEvens:
Any update on this? 

Section 1.2 in draft 04 still only defines `producer` and `consumer` yet the draft refers to them as publisher and subscriber. 

Interactions indicate that SUBSCRIBE is sent by the subscriber and SUBSCRIBE_OK is sent by the publisher, but the actual interaction is that the relay acts/proxies as a subscriber to publisher (client endpoint) and acts/proxies SUBSCRIBE_OK to subscriber.   

[...]

Issue #229, opened 2023-07-28
Define Upstream and Downstream 


fluffy:
Luke took some confused text I sent and made it really clear by using upstream and downstream. I think we should add those to are defined terminology.  They would help many places.

ianswett:
Is there a PR for Luke's changes?

kixelated:
There's no PR; it was just some terminology used in an issue. We used it within Twitch but I'm not sure how common it is across the industry.

Data is produced upstream and flows downstream, potentially branching out along the way. I think the branching is meant to symbolize a river but lul they work the other way around (multiple upstream tributaries join together).

The important this is that upstream and downstream changes based on your perspective. Upstream is in the direction of the broadcaster (ex. OBS) and downstream is in the direction of the viewer (ex. VLC). A relay may have multiple downstream subscriptions (fanout), but deduplicates them into one upstream subscription.

ianswett:
My experience with "upstream" and "downstream" is that it can be confusing, like "backend" and "frontend".  I tend to prefer "sender" and "receiver" or "producer" and "consumer" (or "subscriber"), but no term is perfect.

fluffy:
I think we wanted to use produce and consumer for the ends of the chain, but we need something for the next and prev in the chain. The sender and receiver is just too confusing because we send messages both direction even though media on flows one direction.  Not sure what names we need but I do think we should have a clear name for the prev and next hops for media as well as the first and last hop.

acbegen:
> My experience with "upstream" and "downstream" is that it can be confusing

Luke's definition is quite common. And I agree we need terms for next/previous, so I think we can use these.

suhasHere:
How about 
Producer and Consumer for the endpoints that represent media source and sink 

Publisher and subscriber for everything that participates in moqt transport pub/sub protocol

wilaw:
I am tempted to support 'Publisher' and 'Subscriber'  to define the hop-endpoints, however it is quite possible that both end-points publish and subscribe simultaneously, in which case these semantics become murky. For a robust hop end-point description, how about 'client' (the endpoint that initiates the connection) and  'server' (the endpoint that accepts the connection). This is the same definition as offered by RFC9000. 

Upstream and downstream define directions along a distribution system.  Upstream is towards the source. Downstream is away from the source. 

Since distribution is a chain of clients and servers, we still need clear terms to define the original server and the terminal client. Some candidates:

"original server"
- Source publisher 
- Origin publisher
- Origin
[...]

Issue #157, opened 2023-05-18
Proposal: Sync terminology between transport and arch draft


suhasHere:
PR #156  is updating the transport draft with few terminology , which is fine. However it would be nice to consider the roles/entities terminology defined in https://datatracker.ietf.org/doc/html/draft-nandakumar-moq-arch-00 

It would be good to drive these from security perspective and thus defining roles.

Proposal is to review the terminology in arch-00 spec and either bring in them once we are happy or refer to the arch spec as informative reference in the current spec.

afrind:
For reference, here's the delta between the draft (including 156) and the arch draft:

**Publisher**:
publishes tracks and transmits objects over the network.
vs
publish MoQ Objects belonging to a given track using the MoQ Transport Protocol.

**Subscriber:**
requests tracks and receives objects over the network.
vs
subscribe to receive MoQ Objects corresponding to a track using the MoQ Transport Protocol.

_These seem largely the same, just slightly different word choices._

**Producer** - The initial publisher that encodes tracks and objects.
vs
**Emitter** - Trusted with E2E encryption keys for the media and operate on one or more uncompressed and unencrypted media.  They compress and possibly encrypt and transmit over one or more Data Streams.

**Consumer** - The final subscriber that decodes tracks and objects.
vs
**Receiver** - Trusted with E2E keys and can transform the Encrypted Stream into Encoded Stream and decode the encoded stream to source stream for further consumptions

_For Producer/Emitter and Consumer/Receiver - I'm not highly opinionated and possible bikeshed?  We can vote (hum?!) in the next authors call._

_The main difference is the explicit mention of keys.  Do we need that in the transport draft, or is it sufficient perhaps to say these roles have access to the raw media (not opaque), considering some but not all tracks/objects will be e2e encrypted?_

**Relay**:
subscribes to endpoints and publishes to other endpoints. Relays typically perform deduplication and caching as part of a distributed system.
vs
Trusted with E2E encryption keys.
Entities performing Relay role implement store and forward behavior.  Such entities receives subscriptions and publishes data to the other endpoints that have subscribed to the named data.  They may cache the data as well for optimizing the delivery experience.  Since these entities are not trusted with the E2E keys, they cannot access unencrypted MoQ Object Payload, however they are allowed to access the MoQ Object Header.

_These definitions are largely the same, with some additional examples in the arch version, plus the mention of keys.  The same question about keys applies._

[...]

PR #463, opened 2024-06-12
Updated Terminology

afrind:
Replaces producer/consumer with publisher/subscriber, and introduces new terms for the Original Publisher, End Subscriber and relay.  Also replaced some instances of sender/receiver with publisher/subscriber, except instances that were referring to a particular message.

Added Upstream and Downstream to the definitions section.  Upstream has two uses so far and downstream has none.

Fixes: #157
Fixes: #229
Fixes: #438

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -171,13 +171,34 @@ Endpoint:
 
 : A Client or Server.
 
-Producer:
+Publisher:
 
-: An endpoint sending media over the network.
+: An endpoint that sends tracks.
 
-Consumer:
+Subscriber:
 
-: An endpoint receiving media over the network.
+: An endpoint that subscribes to and receives tracks.
+
+Original Publisher:
+
+: The first publisher of a given track.
+
+End Subscriber:
+
+: A subscriber that independently initiates a subscription.
+
+Relay:
+
+: An entitly that is both a Publisher and a Subscriber, but not the Original
+Publisher or End Subscriber.
+
+Upstream:
+
+: In the direction of the Original Publisher
+
+Downstream:
+
+: In the direction of the End Subscriber(s)
 
 Transport session:
 
@@ -274,14 +295,13 @@ identical sequence of bytes regardless of how or where it is retrieved.
 An Object can become unavailable, but it's contents MUST NOT change over
 time.
 
-Objects are comprised of two parts: metadata and a payload.
-The metadata is never encrypted and is always
-visible to relays. The payload portion may be encrypted, in which case
-it is only visible to the producer and consumer. The application is
-solely responsible for the content of the object payload. This includes
-the underlying encoding, compression, any end-to-end encryption, or
-authentication. A relay MUST NOT combine, split, or otherwise modify
-object payloads.
+Objects are comprised of two parts: metadata and a payload.  The metadata is
+never encrypted and is always visible to relays. The payload portion may be
+encrypted, in which case it is only visible to the Original Publisher and End
+Subscribers. The application is solely responsible for the content of the object
+payload. This includes the underlying encoding, compression, any end-to-end
+encryption, or authentication. A relay MUST NOT combine, split, or otherwise
+modify object payloads.
 
 ## Groups {#model-group}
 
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -171,13 +171,34 @@ Endpoint:
...
 
-Consumer:
+Subscriber:
 
-: An endpoint receiving media over the network.
+: An endpoint that subscribes to and receives tracks.
+
+Original Publisher:
+
+: The first publisher of a given track.
```
fluffy, line 184: I like this but I want to highlight a subtle part of this - if two clients are publishing different group to the same track. Only one of them is the Original publisher. I think this is fine.

fluffy (approved):
Look Good To Me.

draft-ietf-moq-transport.md
```diff
@@ -274,30 +295,29 @@ identical sequence of bytes regardless of how or where it is retrieved.
...
-The metadata is never encrypted and is always
-visible to relays. The payload portion may be encrypted, in which case
-it is only visible to the producer and consumer. The application is
-solely responsible for the content of the object payload. This includes
-the underlying encoding, compression, any end-to-end encryption, or
-authentication. A relay MUST NOT combine, split, or otherwise modify
-object payloads.
+Objects are comprised of two parts: metadata and a payload.  The metadata is
+never encrypted and is always visible to relays. The payload portion may be
+encrypted, in which case it is only visible to the Original Publisher and End
```
wilaw, line 300: In order to implement per-session ad insertion, we may want to give certain edge-servers the ability to read and modify catalogs in a edge-compute type of implementation. In this case it might better to say

> The payload portion may be encrypted, in which case it is only visible to the Original Publisher, End Subscribers and application-authorized relays.

draft-ietf-moq-transport.md
```diff
@@ -274,30 +295,29 @@ identical sequence of bytes regardless of how or where it is retrieved.
...
+Subscribers. The application is solely responsible for the content of the object
+payload. This includes the underlying encoding, compression, any end-to-end
+encryption, or authentication. A relay MUST NOT combine, split, or otherwise
+modify object payloads.
```
wilaw, line 304: Some edge relays may be authorized to modify content. Therefore we should relax this constraint to

> A relay MUST NOT combine, split, or otherwise modify object payloads unless authorized to by an application.

draft-ietf-moq-transport.md
englishm: While I agree that manipulating catalogs with edge compute is a valid use case, I don't think we should blur the lines of what a relay is here, but instead consider that node something different. One way of framing it would be to consider it both an end subscriber for the original catalog track and an original publisher for the modified catalog.

draft-ietf-moq-transport.md
englishm: As above, I'd rather we keep this strict and consider any modifying entity to be something other than a relay. We need a very minimal baseline for what's necessary to provide simple fan out capacity.

englishm: Follow up based on in-person discussion just now: my concerns here are primarily about whether other entities in a system can rely on immutability of objects within tracks of the same name regardless of where they received them from, or if that immutability guarantee is more limited. 

If we allow application-authorized modification to occur, I think it would be helpful to have a name for such authorized entities that is clearly distinct from simple fan out relays to make it more explicit when they are in play and when applications and systems designers may need to account for additional variability.

draft-ietf-moq-transport.md
ianswett, line 184:
```suggestion
: The originating publisher of a given track.
```

draft-ietf-moq-transport.md
```diff
@@ -171,13 +171,34 @@ Endpoint:
 
 : A Client or Server.
 
-Producer:
+Publisher:
 
-: An endpoint sending media over the network.
+: An endpoint that sends tracks.
```
ianswett, line 176:
```suggestion
: An endpoint that handles subscriptions by sending requested Objects from the requested track.
```

draft-ietf-moq-transport.md
```diff
@@ -171,13 +171,34 @@ Endpoint:
...
-: An endpoint receiving media over the network.
+: An endpoint that subscribes to and receives tracks.
+
+Original Publisher:
+
+: The first publisher of a given track.
+
+End Subscriber:
+
+: A subscriber that independently initiates a subscription.
```
ianswett, line 188:
```suggestion
: A subscriber that initiates a subscription and does not send the data on to other subscribers.
```

draft-ietf-moq-transport.md
afrind: circular definition?

draft-ietf-moq-transport.md
ianswett: The use case is very much valid, but this is intended as an editorial PR, so if you think that's a use case we need to address, please file an issue.

PR #469, opened 2024-06-18
Add MAX_CACHE_DURATION to SUBSCRIBE_OK Params

### LINKED ISSUE #249: Object/group TTL
I'm (properly) implementing caching in my relay and need to know how long to cache an object.

For example, a catalog object should be cached until the session terminates. Meanwhile a media segment should only be cached for X seconds. The relay could evict the cache earlier than that, and possibly refetch the object if it's still within the advertised expires window.

It doesn't seem valid to have objects with different cache expirations in a group. For example, there's never a reason why you would expire the I-frame, but not the P-frames that depend on it. Expiration should ideally be a group property but there's no message to express that currently.

### LINKED ISSUE #440: Need expiry time / time to live

ianswett:
A tweaked version of #448 

Fixes #440
Fixes #249

Closes #450 
Closes #448

I'll note that we might want to add params to TRACK_STATUS and allow this to be returned from that as well, which would allow a relay to keep it in its cache.

Attempts to write down where we ended up at the end of the June 12th interim.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1558,7 +1558,8 @@ The following property is defined in this document:
 MAX_CACHE_DURATION: An integer expressing a number of milliseconds. If
 present, the relay MUST NOT start forwarding any individual Object received
 through this subscription after the specified number of seconds has elapsed
-since the beginning of the Object was received.
+since the beginning of the Object was received.  This means Objects earlier
+in a stream will expire earlier than Objects later in the stream.
 
 ## SUBSCRIBE_ERROR {#message-subscribe-error}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1595,8 +1595,7 @@ is only present if ContentExists has a value of 1.
 * Largest Object ID: The largest Object ID available within the largest Group ID
 for this track. This field is only present if ContentExists has a value of 1.
 
-* Subscribe Parameters: Optional key-value pairs formatted as described in
-{{version-specific-params}}.
+* Subscribe Parameters: The parameters are defined in {{version-specific-params}}.
 
 
 ## SUBSCRIBE_ERROR {#message-subscribe-error}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1521,7 +1521,9 @@ SUBSCRIBE_OK
   Expires (i),
   ContentExists (f),
   [Largest Group ID (i)],
-  [Largest Object ID (i)]
+  [Largest Object ID (i)],
+  Number of Subscription Parameters (i),
+  Subscription Parameters (..) ...
 }
 ~~~
 {: #moq-transport-subscribe-ok format title="MOQT SUBSCRIBE_OK Message"}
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1537,12 +1539,27 @@ end prior to the expiry time or last longer.
...
+SUBSCRIBE_OK may contain zero or more parameters describing either the
+properties of the track, or the properties of an individual subscription.
+
+The following property is defined in this document:
+
+MAX_CACHE_DURATION: An integer expressing a number of milliseconds. If
+present, the relay MUST NOT start forwarding any individual Object received
+through this subscription after the specified number of seconds has elapsed
+since the beginning of the Object was received.  This means Objects earlier
+in a stream will expire earlier than Objects later in the stream.
```
ianswett, line 1562:
```suggestion
in a multi-object stream will expire earlier than Objects later in the stream.
```

draft-ietf-moq-transport.md
```diff
@@ -1557,12 +1559,27 @@ end prior to the expiry time or last longer.
...
+### Subscription Parameters  {#subscription-parameters}
+
+SUBSCRIBE_OK may contain zero or more parameters describing either the
+properties of the track, or the properties of an individual subscription.
+
+The following property is defined in this document:
+
+MAX_CACHE_DURATION: An integer expressing a number of milliseconds. If
+present, the relay MUST NOT start forwarding any individual Object received
+through this subscription after the specified number of seconds has elapsed
```
afrind, line 1580: seconds or milliseconds?

afrind (commented):
Individual Review:

Do we need some language about what to do when an object is not in the cache?

draft-ietf-moq-transport.md
ianswett, line 1580:
```suggestion
through this subscription after the specified number of milliseconds has elapsed
```

ianswett: Good catch, fixed to be millis.

ianswett:
> Individual Review:
> 
> Do we need some language about what to do when an object is not in the cache?

Thanks @afrind , I added a sentence to clarify that, let me know if you'd like more?

huitema (commented):
MAX_CACHE_DURATION is a bit of a foot gun, especially if we find long groups in which objects depend on all preceding objects. Would be nice to have some nuance...

fluffy (approved):
Looks good to me.

draft-ietf-moq-transport.md
```diff
@@ -1549,12 +1551,29 @@ Values of 0x0 and those larger than 0x2 are a protocol error.
...
+{{params}}.
+
+### Subscription Parameters  {#subscription-parameters}
+
+SUBSCRIBE_OK may contain zero or more parameters describing either the
+properties of the track, or the properties of an individual subscription.
+
+The following property is defined in this document:
+
+MAX_CACHE_DURATION (key 0x01): An integer expressing a number of milliseconds. If
```
mengelbart, line 1570: Should relays forward the parameter to downstream relays, and if so, what value should be used?

suhasHere (approved):
This looks good .

draft-ietf-moq-transport.md
```diff
@@ -1579,14 +1589,15 @@ Values of 0x0 and those larger than 0x2 are a protocol error.
...
+* Largest Group ID: The largest Group ID available for this track. This field
 is only present if ContentExists has a value of 1.
 
-* Largest Object ID: the largest Object ID available within the largest Group ID
+* Largest Object ID: The largest Object ID available within the largest Group ID
 for this track. This field is only present if ContentExists has a value of 1.
 
 * Subscribe Parameters: The parameters are defined in {{version-specific-params}}.
 
+
```
ianswett, line 1600:
```suggestion

```

PR #470, opened 2024-06-22
Add Priority and Group Send Order to SUBSCRIBE, clarify Publisher Priority

### LINKED ISSUE #326: Hints & sendOrder how can they work together

### LINKED ISSUE #396: How does a relay/publisher know what to prioritize or drop?

### LINKED ISSUE #403: No way to prioritize one track above another

### LINKED ISSUE #419: Reliable Live use-case

ianswett:
Adds a SUBSCRIBER_PRIORITY param that can be specified in SUBSCRIBE or SUBSCRIBE_UPDATE to indicate subscriber priority.  Adds Group Order to SUBSCRIBE to allow indicating Ascending or Descending.  Cleans up the existing priorities section and how Original Subscriber driven priorities are used.

Fixes #326 
Fixes #396 
Fixes #403
Fixes #419 

Closes #445

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -545,10 +545,10 @@ subscribed Objects.
 
 The subscriber indicates the priority of a subscription via the
 Subscriber Priority field and the original publisher indicates priority
-in every stream header.  As such, the subscriber's priority is a property
-of the subscription and the original publisher's priority is a property
-of the Track and the Objects it contains. In both cases, a lower value
-indicates a higher priority, with 0 being the highest priority.
+in every stream or datagram header.  As such, the subscriber's priority is a
+property of the subscription and the original publisher's priority is a
+property of the Track and the Objects it contains. In both cases, a lower
+value indicates a higher priority, with 0 being the highest priority.
 
 The Subscriber Priority is considered first when selecting
 a subscription to send data on within a given session. The original
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1094,7 +1094,6 @@ SUBSCRIBE_UPDATE Message {
   EndGroup (i),
   EndObject (i),
   Subscriber Priority (8),
-  Group Order (8),
   Number of Parameters (i),
   Subscribe Parameters (..) ...
 }
@@ -1117,11 +1116,6 @@ requested.
 * Subscriber Priority: Specifies the priority of a subscription relative to
 other subscriptions in the same session. Lower numbers get higher priority.
 See {{priorities}}.
-
-* Group Order: Allows the subscriber to request Objects be delivered in
-Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
-A value of 0x0 indicates the original publisher's Group Order SHOULD be
-used. Values larger than 0x2 are a protocol error.
 
 * Subscribe Parameters: The parameters are defined in {{version-specific-params}}.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -564,7 +564,7 @@ The subscriber's priority can be changed via a SUBSCRIBE_UPDATE message.
 
 In addition, SUBSCRIBE specifies a Delivery Order of either
 'Ascending' or 'Descending', which indicates whether the lowest or
-highest Group Id SHOULD be delivered first when multiple Groups are
+highest Group Id SHOULD be sent first when multiple Groups are
 available to send.
 
 Within the same group, and the same priority level,
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -550,15 +550,16 @@ property of the subscription and the original publisher's priority is a
 property of the Track and the Objects it contains. In both cases, a lower
 value indicates a higher priority, with 0 being the highest priority.
 
-The Subscriber Priority is considered first when selecting
-a subscription to send data on within a given session. When two or more tracks
+The Subscriber Priority is considered first when selecting a subscription
+to send data on within a given session. When two or more subscriptions
 have equal subscriber priority, the original publisher priority is considered
 next and can change within the track, so subscriptions are prioritized based
 on the highest priority data available to send. For example, if the subscription
 had data at priority 6 and priority 10 to send, the subscription priority would
 be 6. When both the subscriber and original publisher priorities for a
-subscription are equal, send order is implementation-dependent, but the
-expectation is that all subscriptions will be able to send some data.
+subscription are equal, how much data to send from each subscription is
+implementation-dependent, but the expectation is that all subscriptions will
+be able to send some data.
 
 The subscriber's priority can be changed via a SUBSCRIBE_UPDATE message.
 This updates the priority of all unsent data within the subscription,
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -85,7 +85,7 @@ discovery and subscription.
 
 * {{session}} covers aspects of setting up a MOQT session.
 
-* {{priority}} covers mechanisms for prioritizing subscriptions.
+* {{priorities}} covers mechanisms for prioritizing subscriptions.
 
 * {{relays-moq}} covers behavior at the relay entities.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -561,6 +561,8 @@ subscription are equal, send order is implementation-dependent, but the
 expectation is that all subscriptions will be able to send some data.
 
 The subscriber's priority can be changed via a SUBSCRIBE_UPDATE message.
+This updates the priority of all unsent data within the subscription,
+though the details of the reprioitization are implementation-specific.
 
 In addition, SUBSCRIBE or SUBSCRIBE_OK specifies a Group Order of
 either 'Ascending' or 'Descending', which indicates whether the lowest or
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -551,12 +551,12 @@ property of the Track and the Objects it contains. In both cases, a lower
 value indicates a higher priority, with 0 being the highest priority.
 
 The Subscriber Priority is considered first when selecting
-a subscription to send data on within a given session. The original
-publisher priority is considered next and can change within the track,
-so subscriptions are prioritized based on the highest priority data
-available to send. For example, if the subscription had data at
-priority 6 and priority 10 to send, the subscription priority would be 6.
-When both the subscriber and original publisher priorities for a
+a subscription to send data on within a given session. When two or more tracks
+have equal subscriber priority, the original publisher priority is considered
+next and can change within the track, so subscriptions are prioritized based
+on the highest priority data available to send. For example, if the subscription
+had data at priority 6 and priority 10 to send, the subscription priority would
+be 6. When both the subscriber and original publisher priorities for a
 subscription are equal, send order is implementation-dependent, but the
 expectation is that all subscriptions will be able to send some data.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1515,6 +1515,7 @@ SUBSCRIBE_OK
 {
   Subscribe ID (i),
   Expires (i),
+  Group Order (8),
   ContentExists (f),
   [Largest Group ID (i)],
   [Largest Object ID (i)]
@@ -1528,6 +1529,10 @@ SUBSCRIBE_OK
 longer valid. A value of 0 indicates that the subscription does not expire
 or expires at an unknown time.  Expires is advisory and a subscription can
 end prior to the expiry time or last longer.
+
+* Group Order: Indicates the subscrption will be delivered in
+Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
+Values of 0x0 and those larger than 0x2 are a protocol error.
 
 * ContentExists: 1 if an object has been published on this track, 0 if not.
 If 0, then the Largest Group ID and Largest Object ID fields will not be
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -539,6 +539,10 @@ MoQ priorities allow a subscriber and original publisher to influence
 the transmission order of Objects within a session in the presence of
 congestion.
 
+Given the critical nature of control messages and their relatively
+small size, the control stream SHOULD be prioritized higher than all
+subscribed Objects.
+
 The subscriber indicates the priority of a subscription via the
 Subscriber Priority field and the original publisher indicates priority
 in every stream header.  As such, the subscriber's priority is a property
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -539,14 +539,14 @@ MoQ priorities allow a subscriber and original publisher to influence
 the transmission order of Objects within a session in the presence of
 congestion.
 
-The subscriber can indicate the priority of a subscription via the
-SUBSCRIBER_PRIORITY param and the original publisher indicates priority
+The subscriber indicates the priority of a subscription via the
+Subscriber Priority field and the original publisher indicates priority
 in every stream header.  As such, the subscriber's priority is a property
 of the subscription and the original publisher's priority is a property
 of the Track and the Objects it contains. In both cases, a lower value
 indicates a higher priority, with 0 being the highest priority.
 
-When specified, the SUBSCRIBER_PRIORITY is considered first in selecting
+The Subscriber Priority is considered first when selecting
 a subscription to send data on within a given session.  The original
 subscriber priority can change within the track, so subscription is
 prioritized based on the highest priority data currently available to send.
@@ -570,9 +570,9 @@ Within the same group, and the same priority level,
 objects with a lower Object Id are always sent before objects with a
 higher Object Id, regardless of the specified Delivery Order.
 
-Relays SHOULD NOT directly use SUBSCRIBER_PRIORITY or Delivery Order
+Relays SHOULD NOT directly use Subscriber Priority or Delivery Order
 from incoming subscriptions for upstream subscriptions. Relays use of
-SUBSCRIBER_PRIORITY for upstream subscriptions is based on
+Subscriber Priority for upstream subscriptions is based on
 a number of factors specific to it, such as the populatity of the
 content or policy.
 
@@ -842,13 +842,6 @@ information in a SUBSCRIBE or ANNOUNCE message. This parameter is populated for
 cases where the authorization is required at the track level. The value is an
 ASCII string.
 
-#### SUBSCRIBER_PRIORITY Parameter
-
-The SUBSCRIBER_PRIORITY parameter (key 0x03) allows a subscriber to specify the
-priority of a subscription relative to other subscriptions in the same session.
-Lower numbers get higher priority and values range from 0 to 255 inclusive.
-More details on priorities are documented in {{priorities}}
-
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
 
 The `CLIENT_SETUP` and `SERVER_SETUP` messages are the first messages exchanged
@@ -1003,6 +996,7 @@ SUBSCRIBE Message {
   Track Alias (i),
   Track Namespace (b),
   Track Name (b),
+  Priority (8),
   Delivery Order (8),
   Filter Type (i),
   [StartGroup (i),
@@ -1032,6 +1026,10 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 ({{track-name}}).
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
+
+* Subscriber Priority: Specifies the priority of a subscription relative to
+other subscriptions in the same session. Lower numbers get higher priority.
+More details on priorities are documented in {{priorities}}.
 
 * Delivery Order: Requests Objects for the subscription be delievered in
 Ascending (0x0) or Descending (0x1) order. See {{priorities}}.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -588,7 +588,7 @@ factors specific to it, such as the popularity of the
 content or policy, or relays can specify the same value for all
 upstream subscriptions.
 
-MoQ Sessions can span multiple namespaces, and priorities may or may not
+MoQ Sessions can span multiple namespaces, and priorities might not
 be coordinated across namespaces.  The subscriber's priority is
 considered first, so there is a mechanism for a subscriber to fix
 incompatibilities between different namespaces prioritization schemes.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -573,7 +573,7 @@ higher Object Id, regardless of the specified Delivery Order.
 Relays SHOULD NOT directly use Subscriber Priority or Delivery Order
 from incoming subscriptions for upstream subscriptions. Relays use of
 Subscriber Priority for upstream subscriptions is based on
-a number of factors specific to it, such as the populatity of the
+a number of factors specific to it, such as the popularity of the
 content or policy.
 
 MoQ Sessions can span multiple namespaces, and priorities are treated
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -549,7 +549,7 @@ indicates a higher priority, with 0 being the highest priority.
 The publisher SHOULD respect the subscriber and original publisher's
 priorities.
 
-In addition, the SUBSCRIBE specifies a Delivery Order of either
+In addition, SUBSCRIBE specifies a Delivery Order of either
 'Ascending' or 'Descending', which indicates whether the lowest or
 highest Group Id SHOULD be delivered first when multiple Groups are
 available to send.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -562,8 +562,8 @@ expectation is that all subscriptions will be able to send some data.
 
 The subscriber's priority can be changed via a SUBSCRIBE_UPDATE message.
 
-In addition, SUBSCRIBE specifies a Group Order of either
-'Ascending' or 'Descending', which indicates whether the lowest or
+In addition, SUBSCRIBE or SUBSCRIBE_OK specifies a Group Order of
+either 'Ascending' or 'Descending', which indicates whether the lowest or
 highest Group Id SHOULD be sent first when multiple Groups are
 available to send.
 
@@ -1033,9 +1033,10 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 other subscriptions in the same session. Lower numbers get higher priority.
 See {{priorities}}.
 
-* Group Order: Requests Objects for the subscription be delievered in
-Ascending (0x0) or Descending (0x1) order by group. See {{priorities}}.
-Values larger than 0x1 are a protocol error.
+* Group Order: Allows the subscriber to requests Objects be delivered in
+Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
+A value of 0x0 indicates the original publisher's Group Order SHOULD be
+used. Values larger than 0x2 are a protocol error.
 
 * Filter Type: Identifies the type of filter, which also indicates whether
 the StartGroup/StartObject and EndGroup/EndObject fields will be present.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -588,9 +588,6 @@ possibly part of the same application.  In cases when pooling among
 namespaces is expected to cause issues, multiple MoQ sessions, either
 within a single connection or on multiple connections can be used.
 
-## Priority Algorithm
-
-
 
 # Relays {#relays-moq}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -573,6 +573,9 @@ Within the same group, and the same priority level,
 objects with a lower Object Id are always sent before objects with a
 higher Object Id, regardless of the specified Group Order.
 
+The Group Order cannot be changed via a SUBSCRIBE_UPDATE message, and
+instead an UNSUBSCRIBE and SUBSCRIBE can be used.
+
 Relays SHOULD respect the subscriber and original publisher's priorities.
 Relays SHOULD NOT directly use Subscriber Priority or Group Order
 from incoming subscriptions for upstream subscriptions. Relays use of
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -541,10 +541,13 @@ congestion.
 
 The subscriber can indicate the priority of a subscription via the
 SUBSCRIBER_PRIORITY param and the original publisher indicates priority
-in every stream header.  In both cases, a lower value indicates a
-higher priority, with 0 being the highest priority.
+in every stream header.  As such, the subscriber's priority is a property
+of the subscription and the original publisher's priority is a property
+of the Track and the Objects it contains. In both cases, a lower value
+indicates a higher priority, with 0 being the highest priority.
 
-The publisher SHOULD respect the subscriber and publisher's priorities.
+The publisher SHOULD respect the subscriber and original publisher's
+priorities.
 
 In addition, the SUBSCRIBE specifies a Delivery Order of either
 'Ascending' or 'Descending', which indicates whether the lowest or
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -547,11 +547,11 @@ of the Track and the Objects it contains. In both cases, a lower value
 indicates a higher priority, with 0 being the highest priority.
 
 When specified, the SUBSCRIBER_PRIORITY is considered first in selecting
-a subscription to send data on within a given session.  The effective
-original subscriber priority is that of the highest priority data
-currently available to send for that subscription.  For example, if
-the subscription had some data at priority 6 and other data at priority
-10, then the subscription priority would be 6.
+a subscription to send data on within a given session.  The original
+subscriber priority can change within the track, so subscription is
+prioritized based on the highest priority data currently available to send.
+For example, if the subscription had some data at priority 6 and other data
+at priority 10, then the subscription priority would be 6.
 
 The subscriber's priority can be changed via a SUBSCRIBE_UPDATE message.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1032,7 +1032,7 @@ other subscriptions in the same session. Lower numbers get higher priority.
 See {{priorities}}.
 
 * Delivery Order: Requests Objects for the subscription be delievered in
-Ascending (0x0) or Descending (0x1) order. See {{priorities}}.
+Ascending (0x0) or Descending (0x1) order by group. See {{priorities}}.
 Values larger than 0x1 are a protocol error.
 
 * Filter Type: Identifies the type of filter, which also indicates whether
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -597,9 +597,7 @@ similar in functionality to Content Delivery Networks
 (CDNs). Additionally, relays serve as policy enforcement points by
 validating subscribe and publish requests at the edge of a network.
 
-Relays are not required to cache Objects, but relays need to have
-the ability to distribute one upstream subscription to multiple
-downstream subscriptions.
+Relays can cache Objects, but are not required to.
 
 ## Subscriber Interactions
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1033,7 +1033,7 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 other subscriptions in the same session. Lower numbers get higher priority.
 See {{priorities}}.
 
-* Group Order: Allows the subscriber to requests Objects be delivered in
+* Group Order: Allows the subscriber to request Objects be delivered in
 Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
 A value of 0x0 indicates the original publisher's Group Order SHOULD be
 used. Values larger than 0x2 are a protocol error.
@@ -1089,6 +1089,7 @@ SUBSCRIBE_UPDATE Message {
   EndGroup (i),
   EndObject (i),
   Subscriber Priority (8),
+  Group Order (8),
   Number of Parameters (i),
   Subscribe Parameters (..) ...
 }
@@ -1111,6 +1112,11 @@ requested.
 * Subscriber Priority: Specifies the priority of a subscription relative to
 other subscriptions in the same session. Lower numbers get higher priority.
 See {{priorities}}.
+
+* Group Order: Allows the subscriber to request Objects be delivered in
+Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
+A value of 0x0 indicates the original publisher's Group Order SHOULD be
+used. Values larger than 0x2 are a protocol error.
 
 * Subscribe Parameters: The parameters are defined in {{version-specific-params}}.
 
@@ -1530,7 +1536,7 @@ longer valid. A value of 0 indicates that the subscription does not expire
 or expires at an unknown time.  Expires is advisory and a subscription can
 end prior to the expiry time or last longer.
 
-* Group Order: Indicates the subscrption will be delivered in
+* Group Order: Indicates the subscription will be delivered in
 Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
 Values of 0x0 and those larger than 0x2 are a protocol error.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -534,32 +534,19 @@ expects more OBJECTs to be delivered. The server closes the session with a
 'GOAWAY Timeout' if the client doesn't close the session quickly enough.
 
 
-# Prioritization and Congestion Response {#priority-congestion}
-
-TODO: This is a placeholder section to capture details on how the MOQT
-protocol deals with prioritization and congestion overall.
-
-This section is expected to cover details on:
-
-- Prioritization Schemes.
-- Congestion Algorithms and impacts.
-- Mapping considerations for one object per stream vs multiple objects
-  per stream.
-- Considerations for merging multiple streams across domains onto single
-  connection and interactions with specific prioritization schemes.
-
-## Priorities {#priorities}
+# Priorities {#priorities}
 
 MoQ priorities allow a subscriber and original publisher to influence
 the transmission order of Objects within a session in the presence of
 congestion.
 
-The subscriber can optionally indicate a priority in a SUBSCRIBE and
-the original publisher indicates a priority within every stream header.
+The subscriber can indicate the priority of a subscription via the
+SUBSCRIBER_PRIORITY param and the original publisher indicates priority
+in every stream header.
 
 The publisher SHOULD respect the subscriber and publisher's priorities.
 
-In addition, the SUBSCRIBE specifies a delivery_order of either
+In addition, the SUBSCRIBE specifies a Delivery Order of either
 'Ascending' or 'Descending', which indicates whether the lowest or
 highest Group Id SHOULD be delivered first when multiple Groups are
 available to send.
@@ -568,6 +555,11 @@ Within the same group, and the same priority level,
 objects with a lower Object Id are always sent before objects with a
 higher Object Id, regardless of the specified delivery_order.
 
+Relays SHOULD NOT directly use SUBSCRIBER_PRIORITY or Delivery Order 
+from incoming subscriptions for upstream subscriptions. Relays use of
+SUBSCRIBER_PRIORITY for upstream subscriptions is based on
+a number of factors specific to it, such as the populatity of the
+content or policy.
 
 # Relays {#relays-moq}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -885,6 +820,13 @@ information in a SUBSCRIBE or ANNOUNCE message. This parameter is populated for
 cases where the authorization is required at the track level. The value is an
 ASCII string.
 
+#### SUBSCRIBER_PRIORITY Parameter
+
+The SUBSCRIBER_PRIORITY parameter (key 0x03) allows a subscriber to specify the
+priority of a subscription relative to other subscriptions in the same session.
+Lower numbers get higher priority and values range from 0 to 255 inclusive.
+More details on priorities are documented in {{}}
```
ianswett, line 828:
```suggestion
More details on priorities are documented in {{priorities}}
```

draft-ietf-moq-transport.md
```diff
@@ -534,101 +533,32 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
-behavior.
-
-There have been proposals to allow emitters to coordinate the allocation
-of layer priorities across multiple coordinated tracks. At this point,
-these proposals have not reached consensus.
+Within the same group, and the same priority level,
+objects with a lower Object Id are always sent before objects with a
+higher Object Id, regardless of the specified delivery_order.
 
+Relays SHOULD NOT directly use SUBSCRIBER_PRIORITY or Delivery Order
```
ianswett, line 557:
```suggestion
Relays SHOULD NOT directly use SUBSCRIBER_PRIORITY or Delivery Order
```

draft-ietf-moq-transport.md
```diff
@@ -85,8 +85,7 @@ discovery and subscription.
 
 * {{session}} covers aspects of setting up a MOQT session.
 
-* {{priority-congestion}} covers protocol considerations on
-  prioritization schemes and congestion response overall.
+* {{priority}} covers mechanisms for prioritizing subscriptions.
```
ianswett, line 88:
```suggestion
* {{priorities}} covers mechanisms for prioritizing subscriptions.
```

draft-ietf-moq-transport.md
```diff
@@ -534,101 +533,36 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
-- Congestion Algorithms and impacts.
-- Mapping considerations for one object per stream vs multiple objects
-  per stream.
-- Considerations for merging multiple streams across domains onto single
-  connection and interactions with specific prioritization schemes.
+The publisher SHOULD respect the subscriber and original publisher's
+priorities.
 
-## Order Priorities and Options
+In addition, the SUBSCRIBE specifies a Delivery Order of either
```
ianswett, line 552:
```suggestion
In addition, SUBSCRIBE specifies a Delivery Order of either
```

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,61 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
+the transmission order of Objects within a session in the presence of
+congestion.
 
-This section is expected to cover details on:
+The subscriber indicates the priority of a subscription via the
+Subscriber Priority field and the original publisher indicates priority
+in every stream header.  As such, the subscriber's priority is a property
+of the subscription and the original publisher's priority is a property
+of the Track and the Objects it contains. In both cases, a lower value
+indicates a higher priority, with 0 being the highest priority.
```
vasilvv, line 547: Should this also specify lowest priority and default?

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,61 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
+the subscriber and original publisher priorities for a subscription are
+equal, send order is implmentation-depdendent, but the expectation
+is that all subscriptions will be able to send at least some data.
 
-## Order Priorities and Options
+The subscriber's priority can be changed via a SUBSCRIBE_UPDATE message.
 
-At the point of this writing, the working group has not reached
-consensus on several important goals, such as:
+The publisher SHOULD respect the subscriber and original publisher's
```
vasilvv, line 561: Maybe call out relays specifically? This does not make sense for original publishers.

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,61 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
+Within the same group, and the same priority level,
+objects with a lower Object Id are always sent before objects with a
+higher Object Id, regardless of the specified Delivery Order.
 
-### Proposal - Send Order {#send-order}
+Relays SHOULD NOT directly use Subscriber Priority or Delivery Order
+from incoming subscriptions for upstream subscriptions. Relays use of
+Subscriber Priority for upstream subscriptions is based on
+a number of factors specific to it, such as the populatity of the
+content or policy.
```
vasilvv, line 577: How about "SHOULD specify blank priority by default, MAY prioritize based on popularity/etc"?

draft-ietf-moq-transport.md
```diff
@@ -1039,6 +996,8 @@ SUBSCRIBE Message {
   Track Alias (i),
   Track Namespace (b),
   Track Name (b),
+  Priority (8),
+  Delivery Order (8),
```
vasilvv, line 1000: Would it make more sense to make those parameters? Since both of those have a reasonable default value.

draft-ietf-moq-transport.md
```diff
@@ -1252,8 +1219,8 @@ A canonical MoQ Object has the following information:
 IDs starts at 0, increasing sequentially for each object within the
 group.
 
-* Object Send Order: An integer indicating the object send order
-{{send-order}} or priority {{ordering-by-priorities}} value.
+* Priority: An 8 bit integer indicating the publisher's priority for the Object
```
vasilvv, line 1222: This field should probably be called "Publisher Priority".

draft-ietf-moq-transport.md
ianswett: They do, but making people explicitly specify them is also ok.

draft-ietf-moq-transport.md
ianswett: In the current PR, there is no default, but I can add the lowest priority.

draft-ietf-moq-transport.md
ianswett: I can change it, though if one was using MoQ in a P2P application, wouldn't the subscriber's priority matter and the publlsher would be the original publisher?

draft-ietf-moq-transport.md
ianswett: In this PR, there isn't any default, but I also don't think blank is inherently right.

ianswett (commented):
In this PR, there isn't currently a default.  I realize there could be, but unless there's different behavior for it, I'd prefer to make it required because the behavior is clear.

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,61 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
-is reached.
+Within the same group, and the same priority level,
+objects with a lower Object Id are always sent before objects with a
+higher Object Id, regardless of the specified Delivery Order.
 
-### Proposal - Send Order {#send-order}
+Relays SHOULD NOT directly use Subscriber Priority or Delivery Order
+from incoming subscriptions for upstream subscriptions. Relays use of
+Subscriber Priority for upstream subscriptions is based on
+a number of factors specific to it, such as the populatity of the
```
ianswett, line 576:
```suggestion
a number of factors specific to it, such as the popularity of the
```

suhasHere (commented):
This is coming along pretty well. thanks for the PR. I plan to give proper read next week, but overall shape of this is pretty good.

draft-ietf-moq-transport.md
```diff
@@ -1118,6 +1085,7 @@ SUBSCRIBE_UPDATE Message {
   StartObject (i),
   EndGroup (i),
   EndObject (i),
+  Subscriber Priority (8),
```
suhasHere, line 1088: I think we should make this optional or make a special value for "no preference, follow publisher prioirty"

draft-ietf-moq-transport.md
```diff
@@ -1039,6 +996,8 @@ SUBSCRIBE Message {
   Track Alias (i),
   Track Namespace (b),
   Track Name (b),
+  Subscriber Priority (8),
```
suhasHere, line 999: I think we need to make it optional or have a default of no-preference to follow original priority ?

draft-ietf-moq-transport.md
```diff
@@ -1068,6 +1027,14 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
+* Subscriber Priority: Specifies the priority of a subscription relative to
+other subscriptions in the same session. Lower numbers get higher priority.
+See {{priorities}}.
+
+* Delivery Order: Requests Objects for the subscription be delievered in
+Ascending (0x0) or Descending (0x1) order. See {{priorities}}.
+Values larger than 0x1 are a protocol error.
```
suhasHere, line 1036: Ascending or Descending order of the group.

draft-ietf-moq-transport.md
vasilvv: My understanding of the mechanism at which we arrived at the interim is that if the subscriber does not care, it can set all priorities to the same value (e.g. `0x80`), and in that case, the publisher priorities take action.  The advantage of this approach is that if, for some reason, the subscriber decides later that it wants one track higher, it can set it to something like `0x70`, without having to reason about what priorities publisher assigned to its tracks.  It's also really easy to implement (`Final Priority = Subscriber Priority * 256 + Publisher Priority`).

suhasHere: I am fine with it,  but I am not sure how to communicate not to set value of 0x1 implying the most important after the control stream for all the tracks, whereas the subscriber's intent was " i don't care, please follow publisher's priority" .
One way to make it is encode this field as optional and when missing follow publisher's priority for the track.

afrind: Individual Comment:

> most important after the control stream for all the tracks

Should we just add explicit text that control stream always has priority over non-control? 

In that case, any value indicates "don't care, default to publisher" so long as the client sends the *same* value for all subscribes.

draft-ietf-moq-transport.md
ianswett: Done.

draft-ietf-moq-transport.md
ianswett: Sure, I can add text about the control stream.

draft-ietf-moq-transport.md
```diff
@@ -1068,6 +1027,14 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
+* Subscriber Priority: Specifies the priority of a subscription relative to
+other subscriptions in the same session. Lower numbers get higher priority.
+See {{priorities}}.
+
+* Delivery Order: Requests Objects for the subscription be delievered in
+Ascending (0x0) or Descending (0x1) order. See {{priorities}}.
```
ianswett, line 1035:
```suggestion
Ascending (0x0) or Descending (0x1) order by group. See {{priorities}}.
```

draft-ietf-moq-transport.md
ianswett: Good point, done.

draft-ietf-moq-transport.md
suhasHere: yes .. control stream should go with the most important priority and clarifying that will be good.

ianswett: I added a normative statement, PTAL

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,65 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
 
-Media is produced with an intended order, both in terms of when media
-should be presented (PTS) and when media should be decoded (DTS).  As
-stated in the introduction, the network is unable to maintain this
-ordering during congestion without increasing latency.
+Relays SHOULD NOT directly use Subscriber Priority or Delivery Order
+from incoming subscriptions for upstream subscriptions. Relays use of
+Subscriber Priority for upstream subscriptions is based on
+a number of factors specific to it, such as the popularity of the
+content or policy.
```
suhasHere, line 581: I wonder if we should add the note  on publisher preference here too

draft-ietf-moq-transport.md
suhasHere: It would be good to say, if all the subscriber priorities are equal , publisher priorirty is followed for the objects correspnding to the subscribed tracks ?

suhasHere (changes requested):
Few nits and clarification. Overall looking good. Thanks for the PR

draft-ietf-moq-transport.md
afrind: @suhasHere : what's your suggestion?  Something like:

`In the absence of other specific reasons, relays SHOULD give all upstream subscriptions equal priority`

draft-ietf-moq-transport.md
afrind: Line 553-555 is *close* to what your are asking but could be clarified.

Current:

> The Subscriber Priority is considered first when selecting
a subscription to send data on within a given session. The original
publisher priority is considered next...

Maybe:

> The Subscriber Priority is considered first when selecting
a subscription to send data on within a given session. When two or more subscriptions have equal Subscriber Priority, the original
publisher priority is considered next...

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,65 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
+The subscriber indicates the priority of a subscription via the
+Subscriber Priority field and the original publisher indicates priority
+in every stream header.  As such, the subscriber's priority is a property
+of the subscription and the original publisher's priority is a property
+of the Track and the Objects it contains. In both cases, a lower value
+indicates a higher priority, with 0 being the highest priority.
 
-## Order Priorities and Options
+The Subscriber Priority is considered first when selecting
+a subscription to send data on within a given session. The original
```
kixelated, line 554:
```suggestion
a subscription to send data on within a given session. When two subscriptions have the same priority, the original
```

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,65 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
+publisher priority is considered next and can change within the track,
+so subscriptions are prioritized based on the highest priority data
```
kixelated, line 556: I think it's less confusing if you say publisher priority is per stream, instead of per track.

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,65 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
+available to send. For example, if the subscription had data at
+priority 6 and priority 10 to send, the subscription priority would be 6.
+When both the subscriber and original publisher priorities for a
+subscription are equal, send order is implementation-dependent, but the
+expectation is that all subscriptions will be able to send some data.
 
-At the point of this writing, the working group has not reached
-consensus on several important goals, such as:
+The subscriber's priority can be changed via a SUBSCRIBE_UPDATE message.
```
kixelated, line 563: Is this retroactive?

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,65 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
-is reached.
+In addition, SUBSCRIBE specifies a Delivery Order of either
+'Ascending' or 'Descending', which indicates whether the lowest or
+highest Group Id SHOULD be delivered first when multiple Groups are
+available to send.
 
-### Proposal - Send Order {#send-order}
+Within the same group, and the same priority level,
+objects with a lower Object Id are always sent before objects with a
+higher Object Id, regardless of the specified Delivery Order.
```
kixelated, line 575: A good reason to call it `Group Order` instead of `Delivery Order`; it doesn't apply to object ordering.

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,65 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
 
-The encoder determines how to behave during congestion by assigning each
-object a numeric send order.  The send order SHOULD be followed when
-possible, to ensure that the most important media is delivered when
-throughput is limited.  Note that the contents within each object are
-still delivered in order; this send order only applies to the ordering
-between objects.
+MoQ Sessions can span multiple namespaces, and priorities are treated
+equally, regardless of the namespace.  The subscriber's priority is
+considered first, so there is no incentive for the original publisher
```
kixelated, line 585: A subscriber will often want to use the default publisher priority, so this isn't always true.

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,65 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
+to attempt to prioritize all of its Tracks higher than another
+namespace.  Additionally, it is anticipated that when multiple namespaces
+are present within a session, the namespaces could be coordinating,
+possibly part of the same application.  In cases when pooling among
+namespaces is expected to cause issues, multiple MoQ sessions, either
+within a single connection or on multiple connections can be used.
```
kixelated, line 591: I don't think this will work in practice. There's too many use-cases where publishers are not coordinated because they're using different software or versions. 

ex. I want to show multiple Twitch streams on the same page. There's just no way that I can have OBS and ffmpeg agree to publish with the same priority scheme. It doesn't have to be malicious; one of the publishers is undoubtedly going to monopolize the connection.

Dialing a new connection is expensive and removes the ability to use subscriber priorities. I would totally agree with using multiple QUIC sessions if it had such a layer, but the alternative is WebTransport pooling...

And yes, you could use subscriber priorities only to avoid this issue, but that's a big limitation that could be easily solved by sandboxing namespaces. Relays can't use the subscriber's priority anyway and will have to sandbox namespaces if using MoQ over the backbone.

draft-ietf-moq-transport.md
```diff
@@ -638,6 +602,10 @@ similar in functionality to Content Delivery Networks
 (CDNs). Additionally, relays serve as policy enforcement points by
 validating subscribe and publish requests at the edge of a network.
 
+Relays are not required to cache Objects, but relays need to have
```
kixelated, line 605: A proxy is a relay too.

draft-ietf-moq-transport.md
```diff
@@ -1118,6 +1089,7 @@ SUBSCRIBE_UPDATE Message {
   StartObject (i),
   EndGroup (i),
   EndObject (i),
+  Subscriber Priority (8),
```
kixelated, line 1092: What about changing the group order too? It would be useful for toggling between reliable and unreliable playback.

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,65 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
 
-The working group has been considering two alternatives: marking objects
-belonging to a track with an explicit "send order"; and, defining
-algorithms combining tracks, priorities and object order within a
-group. The two proposals are listed in {{send-order}} and
-{{ordering-by-priorities}}.  We expect further work before a consensus
-is reached.
+In addition, SUBSCRIBE specifies a Delivery Order of either
+'Ascending' or 'Descending', which indicates whether the lowest or
+highest Group Id SHOULD be delivered first when multiple Groups are
```
kixelated, line 570: This is only true when groups have the same priority, since we're allowing the publisher priority to change per stream.

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,65 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
-### Proposal - Send Order {#send-order}
+Within the same group, and the same priority level,
+objects with a lower Object Id are always sent before objects with a
+higher Object Id, regardless of the specified Delivery Order.
 
-Media is produced with an intended order, both in terms of when media
-should be presented (PTS) and when media should be decoded (DTS).  As
-stated in the introduction, the network is unable to maintain this
-ordering during congestion without increasing latency.
+Relays SHOULD NOT directly use Subscriber Priority or Delivery Order
```
kixelated, line 577: As briefly discussed at the interim, we also need the `Publisher Group Order` in SUBSCRIBE_OK and TRACK_STATUS. Relays should use that instead of the `Subscriber's Group Order`.

draft-ietf-moq-transport.md
```diff
@@ -1068,6 +1031,14 @@ close the session with a Duplicate Track Alias error ({{session-termination}}).
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
+* Subscriber Priority: Specifies the priority of a subscription relative to
+other subscriptions in the same session. Lower numbers get higher priority.
+See {{priorities}}.
+
+* Delivery Order: Requests Objects for the subscription be delievered in
+Ascending (0x0) or Descending (0x1) order by group. See {{priorities}}.
+Values larger than 0x1 are a protocol error.
```
janaiyengar, line 1040:
```suggestion
* Group Send Order: Requests that objects be picked from Groups in the
subscription in Ascending (0x0) or Descending (0x1) order. See {{priorities}}.
Values larger than 0x1 are a protocol error.
```

janaiyengar (commented):
A few suggestions

ianswett (commented):
I made some updates, though I haven't clarified the text about priority changing between streams on a subscription.  PTAL

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,63 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
 
-- Prioritization Schemes.
-- Congestion Algorithms and impacts.
-- Mapping considerations for one object per stream vs multiple objects
-  per stream.
-- Considerations for merging multiple streams across domains onto single
-  connection and interactions with specific prioritization schemes.
+The subscriber indicates the priority of a subscription via the
+Subscriber Priority field and the original publisher indicates priority
+in every stream header.  As such, the subscriber's priority is a property
```
afrind, line 548: stream or datagram header

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,63 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
+of the subscription and the original publisher's priority is a property
+of the Track and the Objects it contains. In both cases, a lower value
+indicates a higher priority, with 0 being the highest priority.
 
-## Order Priorities and Options
+The Subscriber Priority is considered first when selecting
+a subscription to send data on within a given session. The original
```
afrind, line 554: > When two or more tracks have equal subscriber priority, the original publisher priority is considered next

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,63 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
+publisher priority is considered next and can change within the track,
+so subscriptions are prioritized based on the highest priority data
+available to send. For example, if the subscription had data at
+priority 6 and priority 10 to send, the subscription priority would be 6.
+When both the subscriber and original publisher priorities for a
+subscription are equal, send order is implementation-dependent, but the
```
afrind, line 560: It's not that `sub-pri == orig-pub-pri`, it's when `sub1.pri == sub2.pri` and `pub1.pri == pub2.pri`

> When the subscriber and original publisher priorities for two or more tracks are both equal

draft-ietf-moq-transport.md
afrind: You will need to update this bit, since we agreed it's latest, not highest.

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,63 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
 
-The working group has been considering two alternatives: marking objects
-belonging to a track with an explicit "send order"; and, defining
-algorithms combining tracks, priorities and object order within a
-group. The two proposals are listed in {{send-order}} and
-{{ordering-by-priorities}}.  We expect further work before a consensus
-is reached.
+Within the same group, and the same priority level,
+objects with a lower Object Id are always sent before objects with a
+higher Object Id, regardless of the specified Group Order.
```
afrind, line 572: I think we should revisit this for stream-per-object and datagram.  Forcing these to be in object ID order seems like a restriction of functionality that is not strictly required.

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,63 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
 
-### Proposal - Send Order {#send-order}
+Relays SHOULD respect the subscriber and original publisher's priorities.
+Relays SHOULD NOT directly use Subscriber Priority or Group Order
+from incoming subscriptions for upstream subscriptions. Relays use of
+Subscriber Priority for upstream subscriptions is based on
```
afrind, line 577: s/Subscriber Priority/priority fields/ ?

draft-ietf-moq-transport.md
afrind: I'd rather say that a relay shouldn't compare publisher priorities from uncoordinated publishers, but comparing them from coordinate publishers is fine.  Within a namespace, we can assert all tracks have coordinated priority, but between two namespaces, it would presently have to be implementation dependent.  That said, it shouldn't be hard to fix - break off part of namespace into another tuple field (application or something).

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,63 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
+namespaces is expected to cause issues, multiple MoQ sessions, either
+within a single connection or on multiple connections can be used.
 
-The encoder determines how to behave during congestion by assigning each
-object a numeric send order.  The send order SHOULD be followed when
-possible, to ensure that the most important media is delivered when
-throughput is limited.  Note that the contents within each object are
-still delivered in order; this send order only applies to the ordering
-between objects.
+## Priority Algorithm
```
afrind, line 591: This section left intentionally blank?

draft-ietf-moq-transport.md
afrind: Why can't a relay be a 1:1 relay?  I'd just chop the sentence at

> Relays are not required to cache Objects

The recently merged terminology says 

> Relay: An entitly that is both a Publisher and a Subscriber, but not the Original Publisher or End Subscriber.

draft-ietf-moq-transport.md
afrind: +1 add Group Order to Subscribe Update

afrind (commented):
Individual Review: 

You're still planning the publisher original group order bit in Subscribe OK too?

draft-ietf-moq-transport.md
ianswett: I worry that could create a number of edge cases.  A large benefit of SUBSCRIBE_UPDATE vs UNSUBSCRIBE/SUBSCRIBE is it avoids missing Objects or double-sending Objects.

But I'm not writing the code, so if you think it's useful and won't add unnecessary complexity, I'll add it.

draft-ietf-moq-transport.md
suhasHere: s/the same priority level/ the same object priority level

draft-ietf-moq-transport.md
afrind: Omitting this from subscribe update in favor of unsub/sub is ok, but you might want to add a note that it's intentional.

afrind (commented):
Individual Comment

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,63 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
 
-Media is produced with an intended order, both in terms of when media
-should be presented (PTS) and when media should be decoded (DTS).  As
-stated in the introduction, the network is unable to maintain this
-ordering during congestion without increasing latency.
+MoQ Sessions can span multiple namespaces, and priorities are treated
+equally, regardless of the namespace.  The subscriber's priority is
+considered first, so there is no incentive for the original publisher
+to attempt to prioritize all of its Tracks higher than another
+namespace.  Additionally, it is anticipated that when multiple namespaces
```
englishm, line 585: We may have a problem here. If we expect aggregating relays to not forward subscriber priority preferences, then, in fact, all middle hops are likely to fall to publisher priority and may incentivize priority inflation. On the other hand, relays are going to do what they're going to do, and, it's possible that publisher priorities may be ignored for middle hops in practice.

draft-ietf-moq-transport.md
englishm: This gets a bit off topic, but I think the more important property of a relay is that it republishes tracks from an upstream publisher without modification. I think there may be another entity we could define which could consume one or more tracks from upstream publishers and act as the original publisher for some new tracks (e.g. in the case of "manifest manipulation" use cases at the edge), but I think that's something we should discuss separately.

englishm: As for this bit of text, I'm fine with it either way. To me, it makes explicit that relays may optionally be relatively stateless fan out nodes without any significant local caches, so long as they still allow downstream subscribers to subscribe to tracks from an upstream publisher.

draft-ietf-moq-transport.md
suhasHere: We don't want to end up in the case where the susbcriber gets different experience on each new subscriber joining with different priority  than what has sent upstream. I feel keeping subscriber priorities not go beyond is a good design principle

draft-ietf-moq-transport.md
```diff
@@ -85,8 +85,7 @@ discovery and subscription.
 
 * {{session}} covers aspects of setting up a MOQT session.
 
-* {{priority-congestion}} covers protocol considerations on
-  prioritization schemes and congestion response overall.
+* {{priorities}} covers mechanisms for prioritizing subscriptions.
```
suhasHere, line 88: s/prioritizing subscriptions/prioritizing subscriptions and publications.

draft-ietf-moq-transport.md
fluffy: given we use "send order" to talk about Ascending / Descending, I got confused when I read this. I think it could be less confusing to replace "send order" with "track selected to send next object" or something like that.

draft-ietf-moq-transport.md
suhasHere: I feel this is right as it is worded ..

fluffy: We agreed on highest at this point, the latest is for other part. So I think this is correct as written here. Just to reason about this this, think about a track being send stream per object where publisher priorities changed. I don't think publisher priority is per stream - to start with it is for datagrams too.

draft-ietf-moq-transport.md
fluffy, line 572: This seems correct as written and covers the case where the publisher priority is the same but we need to also cover the case where where publisher priorities are not the same in which  case publisher priority determines which is sent first. This may just be implied here but better to say.

draft-ietf-moq-transport.md
fluffy: Agree with Alan

draft-ietf-moq-transport.md
fluffy: I imagine lots of implementations would  not apply the new priority to data that had already been send or queued to the outgoing QUIC stream and certainly not to UDP datagrams already sent to the OS kernel.  So I think this is a quality of implementation and implementation choice.

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,63 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
-* Allowing nodes and relays to skip or delay some objects to deal with
-  congestion
-* Ensuring that emitters can accurately predict the behavior of relays
-* Ensuring that when relays have to skip and delay objects belonging to
-  different tracks that they do it in a predictable way if tracks are
-  explicitly coordinated and in a fair way if they are not.
+In addition, SUBSCRIBE or SUBSCRIBE_OK specifies a Group Order of
+either 'Ascending' or 'Descending', which indicates whether the lowest or
+highest Group Id SHOULD be sent first when multiple Groups are
+available to send.
```
fluffy, line 568: This seems slightly under specified. If a last mile relay received two different values in both SUBSCRIBE and SUBSCRIBE_OK, we need to say which one takes priority.  We agreed on the end subscriber would take priority so this should have a bit more text to say value in SUBSCRIBE overrides in value in SUBSCRIBE_OK.

draft-ietf-moq-transport.md
fluffy: I like original text better because priority fields seems a bit undefined to me.

draft-ietf-moq-transport.md
fluffy: I get what you are saying here but relays have to deal with a reasonable solution to this just to even be able to run lots of different applications on the same relay network even when application never have more than one namespace. I think there are two keys parts of this 1) we are not trying to specify how a relay network moves data around inside the relays and 2) relays will do what relays do. 

So thought relays could have if they don't think about this problem, I don't think we can say much more that this.

draft-ietf-moq-transport.md
suhasHere: +1 on filling this with the psuedo-code ..

draft-ietf-moq-transport.md
fluffy: Alan, I don't think that would help, that just moves the same problem to the next layer and it would be turtles all the way down. The basic issue is when a content delivery network delivers content for lots of different applications/customers, it needs to decide how it will be "fair" between the application/customers and that is way more complicated that just use whatever the priority header asked for.

suhasHere: I agree with Ian. it is tough to be prescriptive and the proposed text pretty much is well defined

draft-ietf-moq-transport.md
suhasHere: +1 to @ianswett

draft-ietf-moq-transport.md
```diff
@@ -1118,6 +1088,8 @@ SUBSCRIBE_UPDATE Message {
   StartObject (i),
   EndGroup (i),
   EndObject (i),
+  Subscriber Priority (8),
+  Group Order (8),
```
suhasHere, line 1092: I don't think we need this in Subscribe_Update.

draft-ietf-moq-transport.md
fluffy: Luke, I thought you were complaining about too many options :-)  I don't see the use case with this but at first glance does not seem like it would complicated the implementation too much so I am fine with it getting added.

fluffy (approved):
Ian, I think this does an great job of covering what we agreed to the the interim. Thank you for the hard work on it. Few comments on things that I think could be a bit clearer and one small thing that seemed under specified.

afrind (commented):
Individual Comments: 

Buried in here is a proposal for a different algorithm to resolve the priority of a track when it has multiple different priorities in flight.

draft-ietf-moq-transport.md
ianswett: Good point, done

draft-ietf-moq-transport.md
ianswett: Removed for now

draft-ietf-moq-transport.md
ianswett: Based on the definition, I'll shorten this sentence and clarify some other text.

draft-ietf-moq-transport.md
ianswett: True, I added a bit more text to clarify what happens when priority varies within a group.

ianswett (commented):
I tried to address most comments, PTAL

draft-ietf-moq-transport.md
ianswett: Good point, I rewrote this paragraph to better match the current framing.

draft-ietf-moq-transport.md
ianswett: I expanded the sentence a bit more to note that a relay can use a single Subscriber Priority value for all upstream subscriptions.

draft-ietf-moq-transport.md
ianswett: Fair point, I changed this to text that points out the subscriber priority provides a mechanism to override.

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,70 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
+
+The Subscriber Priority is considered first when selecting
+a subscription to send data on within a given session. When two or more tracks
+have equal subscriber priority, the original publisher priority is considered
+next and can change within the track, so subscriptions are prioritized based
+on the highest priority data available to send. For example, if the subscription
+had data at priority 6 and priority 10 to send, the subscription priority would
+be 6. When both the subscriber and original publisher priorities for a
+subscription are equal, send order is implementation-dependent, but the
+expectation is that all subscriptions will be able to send some data.
```
suhasHere, line 561: s/  send order is / the track whose object is being selected is /

draft-ietf-moq-transport.md
```diff
@@ -1137,6 +1114,15 @@ open-ended.
...
 requested.
 
+* Subscriber Priority: Specifies the priority of a subscription relative to
+other subscriptions in the same session. Lower numbers get higher priority.
+See {{priorities}}.
+
+* Group Order: Allows the subscriber to request Objects be delivered in
+Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
+A value of 0x0 indicates the original publisher's Group Order SHOULD be
+used. Values larger than 0x2 are a protocol error.
```
suhasHere, line 1124: Line 580 indicates that group order cannot be changed via Subscribe update and I agree with that. We need to remove Group Order from the Update message though

suhasHere (approved):
@ianswett  the new changes look great. I have one change for removing the Group Order from the Subscribe Update message and define an error type when someone changes it in the update message.

draft-ietf-moq-transport.md
```diff
@@ -534,100 +533,70 @@ expects more OBJECTs to be delivered. The server closes the session with a
...
+though the details of the reprioitization are implementation-specific.
+
+Subscriptions have a Group Order of either 'Ascending' or 'Descending',
+which indicates whether the lowest or highest Group Id SHOULD be sent first
+when multiple Groups are available to send.  A subscriber can specify either
+'Ascending' or 'Descending' in the SUBSCRIBE message or they can specify they
+want to use the Original Publisher's Group Order, which is indicated in
+the corresponding SUBSCRIBE_OK.
+
+Within the same Group, and the same priority level,
```
vasilvv, line 574: I'm confused, does this imply that a single group can have multiple priorities within it?

vasilvv (approved):

draft-ietf-moq-transport.md
ianswett: Thanks for catching that, removed.

draft-ietf-moq-transport.md
ianswett: With Stream-per-Object, each Object can have a different priority in the current wire format.

draft-ietf-moq-transport.md
ianswett: Good point, this text was a bit unclear.  I rewrote it to hopefully be more clear.

draft-ietf-moq-transport.md
afrind: I am ok with merging the new language.  Long term, I don't plan to compare uncoordinated publisher priorities in my generic relay, and may make a proposal for how to avoid doing so.

draft-ietf-moq-transport.md
afrind: The new text looks good.

afrind (commented):
Individual Comments:

PR #479, opened 2024-07-04
Invalid SUBSCRIBE_UPDATE is a Protocol Violation

### LINKED ISSUE #455: SUBSCRIBE_UPDATE Message code and error cases
Table 4 is missing SUBSCRIBE_UPDATE, so there's no code for this message type.

Section 6.5 doesn't say what to do when the update is malformed, in some way. SUBSCRIBE_ERROR? 
 
Is a bogus SUBSCRIBE_UPDATE ignored, or does it nuke the original subscribe?

### LINKED ISSUE #467: SUBSCRIBE_UPDATE response/error handling
Draft 04 currently doesn't mention about sending any response when the SUBSCRIBE_UPDATE is successful or when it fails even though the draft hints on possible failures

>   If Objects before the start or after the end of the current
>   subscription are needed, a separate subscription can be made.  The
>   start Object MUST NOT decrease and when it increases, there is no
>   guarantee that a publisher will not have already sent Objects before
>   the new start Object.  The end Object MUST NOT increase and when it
>   decreases, there is no guarantee that a publisher will not have
>   already sent Objects after the new end Object.

The above suggests that an error should be sent to inform the client that it is making invalid subscription updates.

ianswett:
This case seems easily avoidable by a subscriber, so the simplest option is to close the session.

Fixes #455
Fixes #467

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1080,8 +1080,8 @@ there is no guarantee that a publisher will not have already sent Objects before
 the new start Object.  The end Object MUST NOT increase and when it decreases,
 there is no guarantee that a publisher will not have already sent Objects after
 the new end Object. A publisher SHOULD close the Session as a 'Protocol Violation'
-if the SUBSCRIBE_UPDATE violates either rule or if the subscriber specifies a Subscribe
-ID that has never existed within the Session.
+if the SUBSCRIBE_UPDATE violates either rule or if the subscriber specifies a
+Subscribe ID that does not exist within the Session.
 
 Unlike a new subscription, SUBSCRIBE_UPDATE can not cause an Object to be
 delivered multiple times.  Like SUBSCRIBE, EndGroup and EndObject MUST specify the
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

fluffy (approved):
LGTM

draft-ietf-moq-transport.md
```diff
@@ -1079,7 +1079,9 @@ subscription can be made. The start Object MUST NOT decrease and when it increas
 there is no guarantee that a publisher will not have already sent Objects before
 the new start Object.  The end Object MUST NOT increase and when it decreases,
 there is no guarantee that a publisher will not have already sent Objects after
-the new end Object.
+the new end Object. A publisher SHOULD close the Session as a 'Protocol Violation'
+if the SUBSCRIBE_UPDATE violates either rule or if it specifies a Subscribe ID
+that has never existed within the Session.
```
kixelated, line 1084: Since we have a single control stream, we can assume everything is ordered.

```suggestion
that does not exist within the Session.
```

kixelated (approved):

draft-ietf-moq-transport.md
janaiyengar, line 1084:
```suggestion
the new end Object. A publisher SHOULD close the Session as a 'Protocol Violation'
if the SUBSCRIBE_UPDATE violates either rule or if the subscriber specifies a Subscribe
ID that has never existed within the Session.
```

janaiyengar (approved):
One nit, but LGTM

draft-ietf-moq-transport.md
kixelated: also this way you don't have to remember every subscribe ID that has ever existed.

ianswett: Done

PR #483, opened 2024-07-11
STREAM_HEADER_TRACK and STREAM_HEADER_GROUP can in unidirectional stream

jiuhaizhang:
STREAM_HEADER_TRACK and STREAM_HEADER_GROUP can in unidirectional stream. 
But in https://datatracker.ietf.org/doc/html/draft-ietf-moq-transport-05#section-6.11.2-2 
'An OBJECT_STREAM message MUST be the first and only message on a unidirectional stream.'


ianswett (commented):
I think the existing text is what we intended to say, because it's describing the STREAM_HEADER_TRACK and STREAM_HEADER_GROUP as messages in themselves followed by Object Messages.

Issue #482, opened 2024-07-08
Challenges with being too vague on `subscribe_id` and `track_alias`


TimEvens:
As I read the draft, `SUBSCRIBE_ID` is the distinguisher of a given subscription. `TRACK_ALIAS`  is made unique for a subscription state by indicating *"If the Track Alias is already in use, the publisher **MUST** close the session with a Duplicate Track Alias error."* 

The `TRACK_ALIAS` is mentioned as *"A session specific identifier for the track."*. It is initially set by the subscriber, where the publisher **MAY** reject... likely not if it wants to avoid long subscribe times.   The draft does not indicate that the track alias is a hash of track `namespace + name` or not.  

If the client implementation uses a sequence ID for both subscribe ID and track alias, then it could result in duplicate subscribes that would result in the relay building state, propagating the subscribe toward the publisher to solicit a response of OK/ERROR. In some implementations that support headless publishers, the relay would send a subscribe OK immediately.

The edge relay that has the connection to the publisher performs state compression of subscribes, but that's only at the `track_alias` level per the draft.  In this case, the track alias is different, resulting in a new subscribe (duplicate when looking at the `namespace+name`) being created to the publisher. 

The publisher receives the subscribe with a new `SUBSCRIBE_ID` and `TRACK_ALIAS` and starts to publish data objects using the `SUBSCRIBE_ID` and `TRACK_ALIAS` that match internally the **track** `namespace + name`.  If one or more clients do this repeatedly, it would result in all relays and the actual publisher itself sending duplicate data that matches `namespace + name`.  

Below diagram illustrates the above.

```mermaid
sequenceDiagram
    actor P as Actual Publisher
    participant R1 as Relay 1
    participant R2 as Relay 2
    actor S as Lazy Subscriber

    S ->>R2: SUBSCRIBE (id=5, space=clock, name=second, alias=100)
    R2-->>S: SUBSCRIBE_OK (id=5, alias=100)
    R2 ->>R1:  SUBSCRIBE (id=99, space=clock, name=second, alias=9901)
    R1 -->>R2: SUBSCRIBE_OK (id=99, alias=9901)
    R1 ->>P:  SUBSCRIBE (id=14, space=clock, name=second, alias=31)
    P -->>R1: SUBSCRIBE_OK (id=14, alias=31)
    P -) R1: OBJECT (id=14, alias=31, g=1, id=1)
    R1 -)R2: OBJECT (id=99, alias=9901,  g=1, id=1)
[...]

PR #485, opened 2024-07-12
Clarify duplicate track alias in Subscribes

suhasHere:
Small editorial clarification on duplicate track alias for subscriptions that addresses aspects of #482


TimEvens (commented):
Does this mean we close the complete quic session? Can we just reject the subscribe?

ianswett (commented):
I feel like this is reasonable, but I'd like to sort out how we're using Track Alias and Subscribe ID anyway.  It would be ideal to do that next week, but if others would like this merged in the meantime, I have no objection.

fluffy (approved):

ianswett (approved):

Issue #355, opened 2024-02-05
Objects with unknown length and forwarding preferences


vasilvv:
In our current encoding, you can stream objects without knowing their length in advance if your forwarding preference is Object or Datagram, but not if it's Track or Stream (since those require explicit lengths in the front).  I feel like those two aspects should not be tied: we should either support or not support streaming for all preferences.

afrind:
Individual Comment:

In order to stream multiple objects with unknown lengths in a multi-object stream, you'd need additional framing.    

Alternately, we could create single-object variants of the TRACK and GROUP stream header.

kixelated:
I agree it's weird to have a mix of unframed and framed messages.

While I prefer unframed streams... I think we should state that MoQ objects are always framed, represent a single point in time, and have a known length. It's an optimization to withhold the length in the datagram and dedicated stream case but maybe we should include it.

fluffy:
In meeting we decided add a length to everything. We can add extension later to have no length if we need ( Thanks to Mike E for suggestion )

PR #486, opened 2024-07-12
Add a DELIVERY_TIMEOUT param to subscriptions

ianswett:
Pulls in some text from #449 and attempts to convey the conclusions of the Seattle interim.

Closes #449

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -850,6 +850,26 @@ information in a SUBSCRIBE or ANNOUNCE message. This parameter is populated for
 cases where the authorization is required at the track level. The value is an
 ASCII string.
 
+#### DELIVERY TIMEOUT Parameter {#delivery-timeout}
+
+DELIVERY TIMEOUT parameter (key 0x03) MAY appear in a SUBSCRIBE, SUBSCRIBE_OK,
+or a SUBSCRIBE_UDPATE message.  It is duration in milliseconds the relay
+SHOULD attempt forwarding objects.  If both the subscriber and publisher
+specify the field, the relay takes min of publisher and subscriber time for the
+subscription.  The publisher SHOULD always specify the value received from an
+upstream subscription when there is one, not the min. The start time for the
+timeout is based on when the beginning of the Object is received, and does not
+depend upon the forwarding preference. If an earlier Object arrives later than
+subsequent Objects, relays can consider the receipt time as that of the next
+later Object, with the assumption that the Object's data was reordered.
+
+If neither the subscriber or publisher specify DELIVERY TIMEOUT, Objects are
+delivered as indicated by their Group Order and Priority.
+
+When sent by a subscriber, this parameter is intended to be specific to a
+subscription, so it SHOULD NOT be forwarded upstream by a relay that intends
+to serve multiple subscriptions for the same track.
+
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
 
 The `CLIENT_SETUP` and `SERVER_SETUP` messages are the first messages exchanged


draft-ietf-moq-transport.md
```diff
@@ -850,6 +850,28 @@ information in a SUBSCRIBE or ANNOUNCE message. This parameter is populated for
...
 
+#### DELIVERY TIMEOUT Parameter {#delivery-timeout}
+
+The DELIVERY TIMEOUT parameter (key 0x03) MAY appear in a SUBSCRIBE,
+SUBSCRIBE_OK, or a SUBSCRIBE_UDPATE message.  It is the duration in milliseconds
+the relay SHOULD continue to attempt forwarding Objects after they have been
+received.  The start time for the timeout is based on when the beginning of the
+Object is received, and does not depend upon the forwarding preference. 
+
+If both the subscriber and publisher specify the field, the relay and subscriber
```
ianswett, line 861:
```suggestion
If both the subscriber and publisher specify the field, the publisher and subscriber
```

draft-ietf-moq-transport.md
```diff
@@ -850,6 +850,28 @@ information in a SUBSCRIBE or ANNOUNCE message. This parameter is populated for
...
 
+#### DELIVERY TIMEOUT Parameter {#delivery-timeout}
+
+The DELIVERY TIMEOUT parameter (key 0x03) MAY appear in a SUBSCRIBE,
+SUBSCRIBE_OK, or a SUBSCRIBE_UDPATE message.  It is the duration in milliseconds
+the relay SHOULD continue to attempt forwarding Objects after they have been
+received.  The start time for the timeout is based on when the beginning of the
+Object is received, and does not depend upon the forwarding preference. 
+
+If both the subscriber and publisher specify the field, they use the min of the
```
ianswett, line 861:
```suggestion
If both the subscriber and publisher specify the parameter, they use the min of the
```

draft-ietf-moq-transport.md
```diff
@@ -850,6 +850,32 @@ information in a SUBSCRIBE or ANNOUNCE message. This parameter is populated for
...
+Object's data was reordered.
+
+If neither the subscriber or publisher specify DELIVERY TIMEOUT, Objects are
+delivered as indicated by their Group Order and Priority.
+
+When sent by a subscriber, this parameter is intended to be specific to a
+subscription, so it SHOULD NOT be forwarded upstream by a relay that intends
+to serve multiple subscriptions for the same track.
+
+Publishers SHOULD consider whether the entire Object is likely to be delivered
```
wilaw, line 875: > Publishers SHOULD consider whether the entire Object is likely to be delivered

How can a relay comply with this requirement? In most cases it does not know the size of the object when it begins forwarding it.

ianswett: With the exception of Object per stream/datagram, the Object header always contains a length.  And at some point, we had consensus to add a length in the Object per stream/datagram case as well.  #355

fluffy (approved):
LGTM

draft-ietf-moq-transport.md
```diff
@@ -850,6 +850,32 @@ information in a SUBSCRIBE or ANNOUNCE message. This parameter is populated for
...
 
+#### DELIVERY TIMEOUT Parameter {#delivery-timeout}
+
+The DELIVERY TIMEOUT parameter (key 0x03) MAY appear in a SUBSCRIBE,
+SUBSCRIBE_OK, or a SUBSCRIBE_UDPATE message.  It is the duration in milliseconds
+the relay SHOULD continue to attempt forwarding Objects after they have been
+received.  The start time for the timeout is based on when the beginning of the
+Object is received, and does not depend upon the forwarding preference.
+
+If both the subscriber and publisher specify the parameter, they use the min of the
```
suhasHere, line 861: I think this is useful but IIRC i don't see a place where the publisher can indicate the delivery timeout.

suhasHere (changes requested):
This is useful addition. I have a minor clarification. Otherwise looks good.

draft-ietf-moq-transport.md
ianswett: By making it a param, the intent is that a subscriber can use it in SUBSCRIBE or publisher can use it in SUBSCRIBE_OK.

suhasHere: thanks @ianswett I totally missed it. Makes sense.

suhasHere (approved):

draft-ietf-moq-transport.md
ianswett: Good catch, for some reason I thought we'd already added parameters to SUBSCRIBE_OK, so I'll do that now.

suhasHere: Thanks Ian. Looks good

PR #487, opened 2024-07-15
Add an explicit length to OBJECT_STREAM and OBJECT_DATAGRAM

### LINKED ISSUE #355: Objects with unknown length and forwarding preferences

ianswett:
This had consensus in the past, but a PR wasn't written.

Particularly now that we have Object Status, which is optional for all other mappings, it makes sense to add the length and make things consistent.

Fixes #355


fluffy (approved):
LGTM

kixelated (approved):

PR #488, opened 2024-07-16
At least one SUBSCRIBE_OK to return SUBSCRIBE_OK

ianswett:
This is a bit different from HTTP, but having at least one SUBSCRIBE_OK succeed ensures the track exists in some sense, as well as allowing a relay to populate the Group Order field.

This is intended as a clarification, but it adds normative language.


fluffy (approved):
LGTM

PR #489, opened 2024-07-23
Clarify what Latest Group/Object send when there's no data

### LINKED ISSUE #377: What is the largest group/largest object if a relay has not received any objects?

ianswett:
Fixes #377


suhasHere (approved):
LGTM

PR #491, opened 2024-07-26
Add MAX_SUBSCRIBE_ID to limit number of subscriptions

### LINKED ISSUE #272: Multiple subscriptions could cause resource exhaustion
Currently, there's no way to limit the number of simultaneous subscriptions.  QUIC stream limits cannot effectively be used to limit the number of subscriptions, so I'd suggest MoQ have a limit.

The simplest idea would be to have a limit specified as a parameter in the server's SETUP message.

As a related question, even with limits, what happens if a client subscribes to a number of streams that it doesn't have the bandwidth to support?  ie: 10 4k video streams and they're on a slow link?  Possibly the subscriptions could fail with an error code indicating that the server didn't have enough bandwidth to request them and/or serve them to the client?

ianswett:
Fixes #272

Some diffs are due to me reflowing text to fit the line limits, so please ignore those.

Once we have this and a limit on announce IDs, it seems likely that we'll be able to get rid of 'Role', but for now it can stay.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -958,9 +958,8 @@ the `query` portion of the URI to the parameter.
 #### MAX_SUBSCRIBE_ID {#max-subscribe-id}
 
 The MAX_SUBSCRIBE_ID parameter (key 0x02) communicates an initial value for
-the Maximum Subscribe ID the peer can use when making subscriptions.
-The default value is 0, so if not specified, the peer MUST NOT create
-subscriptions.
+the Maximum Subscribe ID to the receiving subscriber. The default value is 0,
+so if not specified, the peer MUST NOT create subscriptions.
 
 
 ## GOAWAY {#message-goaway}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -504,7 +504,7 @@ code, as defined below:
   that was already in use.
 
 * Too Many Subscribes: The session was closed because the subscriber used
-  a Subscribe ID larger than the current Maximum Subscribe ID.
+  a Subscribe ID equal or larger than the current Maximum Subscribe ID.
 
 * GOAWAY Timeout: The session was closed because the client took too long to
   close the session in response to a GOAWAY ({{message-goaway}}) message.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1651,8 +1651,8 @@ MAX_SUBSCRIBE_ID
 ~~~
 {: #moq-transport-max-subscribe-id format title="MOQT MAX_SUBSCRIBE_ID Message"}
 
-* Subscribe ID: The new Maximums Subscribe ID for the session. If a Subscribe ID
-equal or larger than this is used in any message, including SUBSCRIBE,
+* Subscribe ID: The new Maximum Subscribe ID for the session. If a Subscribe ID
+equal or larger than this is received in any message, including SUBSCRIBE,
 the publisher MUST close the session with an error of 'Too Many Subscribes'.
 More on Subscribe ID in {{message-subscribe-req}}.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -864,9 +864,8 @@ ASCII string.
 
 The MAX_SUBSCRIBE_ID parameter (key 0x03) communicates an initial value for
 the Maximum Subscribe ID the peer can use when making subscriptions.
-If not specified, the peer MUST NOT create subscriptions. Subscribers (0x02)
-MUST NOT specify a MAX_SUBSCIBE_ID, because the role indicates it cannot
-handle subscriptions.
+The default value is 0, so if not specified, the peer MUST NOT create
+subscriptions.
 
 
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -404,7 +404,7 @@ The client can establish a connection to a MoQ server identified by a
 given URI by setting up a QUIC connection to the host and port
 identified by the `authority` section of the URI.  The `path-abempty`
 and `query` portions of the URI are communicated to the server using the
-PATH parameter ({{path}}) which is sent in the CLIENT_SETUP message at the
+PATH parameter ({{path}}) which is sent in the CLIENT_message at the
```
ianswett, line 407:
```suggestion
PATH parameter ({{path}}) which is sent in the CLIENT_SETUP message at the
```

draft-ietf-moq-transport.md
```diff
@@ -1620,6 +1634,29 @@ message in this track.
...
+
+~~~
+MAX_SUBSCRIBE_ID
+{
+  Subscribe ID (i),
+}
+~~~
+{: #moq-transport-max-subscribe-id format title="MOQT MAX_SUBSCRIBE_ID Message"}
+
+* Subscribe ID: The maximum valid Subscribe ID for the session. If a Subscribe ID
```
ianswett, line 1654:
```suggestion
* Subscribe ID: The new Maximums Subscribe ID for the session. If a Subscribe ID
```

draft-ietf-moq-transport.md
```diff
@@ -501,6 +503,9 @@ code, as defined below:
 * Duplicate Track Alias: The endpoint attempted to use a Track Alias
   that was already in use.
 
+* Too Many Subscribes: The session was closed because the subscriber used
+  a Subscribe ID larger than the current Maximum Subscribe ID.
```
ianswett, line 507:
```suggestion
  a Subscribe ID equal or larger than the current Maximum Subscribe ID.
```

draft-ietf-moq-transport.md
```diff
@@ -501,6 +503,9 @@ code, as defined below:
 * Duplicate Track Alias: The endpoint attempted to use a Track Alias
   that was already in use.
 
+* Too Many Subscribes: The session was closed because the subscriber used
+  a Subscribe ID equal or larger than the current Maximum Subscribe ID.
```
englishm, line 507: Is it expected that a publisher MUST close an entire session under these circumstances? Or could it just not fulfill the subscription request? I guess maybe we don't have another way to signal the latter.

englishm: My concern would be that the potential impact on aggregating subscribers (relays) could be substantial if they erroneously open too many subscriptions (or are somehow coerced into doing so).

englishm: That said, having a limit here is better than no limit, so I'd be fine revisiting this concern at a later date.

draft-ietf-moq-transport.md
```diff
@@ -993,6 +998,13 @@ client MUST set the PATH parameter to the `path-abempty` portion of the
 URI; if `query` is present, the client MUST concatenate `?`, followed by
 the `query` portion of the URI to the parameter.
 
+#### MAX_SUBSCRIBE_ID {#max-subscribe-id}
```
vasilvv, line 1001: This is unnecessary, since you can just send a MAX_SUBSCRIBE_ID immediately after the SETUP message.

draft-ietf-moq-transport.md
```diff
@@ -1668,6 +1680,29 @@ message in this track.
...
 message in the `Final Group` for this track.
 
+## MAX_SUBSCRIBE_ID {#message-max-subscribe-id}
+
+A publisher sends a MAX_SUBSCRIBE_ID message to increase the number of
+subscriptions a subscriber can create within a session.
+
+The Maximum Subscribe Id MUST only increase within a session, and
+receipt of a MAX_SUBSCRIBE_ID message with an equal or smaller Subscribe ID
+value is a 'Protocol Violation'.
```
suhasHere, line 1690: If a subscriber received  MAX_SUBSCRIBE_ID with value lesser than previous MAX_SUBSCRIBE_ID sent by the publisher, are we saying here that the subscriber needs to close the session ? I am not sure if that is feasible to do or can be enforced.

suhasHere:
Thanks @ianswett  . this looks like a good improvement. I have few clarification questions and bit of extra work for you to add some informative text, if my comment seems ok to you.

fluffy (commented):
This seems like a really complicated way to limit the subscribes. I understand how it mirror streams but that had pretty different constrains.  Why don't we just have an subscribe error of "too many subscriptions"?

draft-ietf-moq-transport.md
ianswett: Aggregating subscribers exist in HTTP proxies as well, and those deal with stream limits on both upstream and downstream fine, so I'd suggest we revisit if it turns out to be a problem people experience.

draft-ietf-moq-transport.md
ianswett: Yes.  I added it because we have the equivalent in QUIC.

ianswett:
> This seems like a really complicated way to limit the subscribes. I understand how it mirror streams but that had pretty different constrains. Why don't we just have an subscribe error of "too many subscriptions"?

Unfortunately, that makes it much more difficult for the server to tell abusive clients from valid ones, as we discovered with the recent HTTP/2 rapid reset attacks.  In the end we did develop heuristics, but they're imperfect and in the process we broke a bunch of applications in an effort to defend against DDoS attacks.  In contrast, QUIC had a system like this and experienced no issues and we didn't need to develop any new heuristics.

martinduke (approved):
Discussed open issues in the 7 August 2024 virtual, not recorded here. Remaining issue (just do a SUBSCRIBE_ERROR?) went to the list and chairs declared rough consensus, although at least 2 participants would prefer not to have this.

PR #494, opened 2024-07-26
Peeps First Attempt

martinduke:
I have foregone any premature optimization of the encodings here.

Asking @janaiyengar to take a first look.


vasilvv:
How does one close a peep? Currently, in our implementation, we rely in combination of delivery preference and object status for that.

draft-ietf-moq-transport.md
```diff
@@ -302,9 +302,58 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
+A peep is a collection of one or more objects and is a sub-unit of a group
+({{model-group}}). A peep consists of a set of objects within a group that
+have a dependency and priority relationship consistent with sharing a QUIC
+stream.
+
+When a Track's forwarding preference (see {{object-fields}}) is "Track" or
+"Datagram", Groups do not contain Peeps, no Peep IDs are assigned, and the
+description in the remainder of this section does not apply.
+
+QUIC streams offer in-order reliable delivery, differentiation of priority in
```
LPardue, line 316: QUIC streams as a transport feature don't really offer prioritization as described though

martinduke:
> How does one close a peep? Currently, in our implementation, we rely in combination of delivery preference and object status for that.

I think a Stream FIN would be sufficient.

draft-ietf-moq-transport.md
martinduke: I agree that it's not in 9000, but as you know it's a common feature of APIs, and streams are the lever by which QUIC can prioritize certain bits of data.

Is there a wording you would prefer?

draft-ietf-moq-transport.md
```diff
@@ -302,9 +302,58 @@ payload. This includes the underlying encoding, compression, any end-to-end
 encryption, or authentication. A relay MUST NOT combine, split, or otherwise
 modify object payloads.
 
+## Peeps {#model-peeps}
```
suhasHere, line 305: May be call it SubGroups :-) ?

draft-ietf-moq-transport.md
```diff
@@ -302,9 +302,58 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
+stream.
+
+When a Track's forwarding preference (see {{object-fields}}) is "Track" or
+"Datagram", Groups do not contain Peeps, no Peep IDs are assigned, and the
+description in the remainder of this section does not apply.
+
+QUIC streams offer in-order reliable delivery, differentiation of priority in
+delivery, and the ability to cancel sending and retransmission of data.
+
+Every object belongs to exactly one Peep.
```
suhasHere, line 319: s/Every object/Every object within a group, to be more specific that peepId are scoped to a given group

draft-ietf-moq-transport.md
```diff
@@ -302,9 +302,58 @@ payload. This includes the underlying encoding, compression, any end-to-end
 encryption, or authentication. A relay MUST NOT combine, split, or otherwise
 modify object payloads.
 
+## Peeps {#model-peeps}
+
+A peep is a collection of one or more objects and is a sub-unit of a group
+({{model-group}}). A peep consists of a set of objects within a group that
+have a dependency and priority relationship consistent with sharing a QUIC
+stream.
+
```
suhasHere, line 311: I think adding clarification that Peeps addresses the cases where a given MOQT group might need multiple streams  .. not sure of the exact wording though

draft-ietf-moq-transport.md
```diff
@@ -302,9 +302,58 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
+Alternatively, objects 1 and 2 could be in one peep and 3 in another. The
+optimal choice would depend on the dependencies of later objects.
+
+An original publisher might not follow these guidelines. Reasons would
+include implementation complexity, lack of foreknowledge of the dependencies
+of further objects, the knowledge that the transport is WebTransport over
+HTTP/2 (thus introducing head-of-line blocking across streams), or knowledge
+of low session stream limits that require aggregation of independent objects
+into the same stream.
+
```
suhasHere, line 350: I wonder if we need to provide these recommendations at all . May be if we leave it to say 
"Original publishers assigned objects within a group to different peeps driven by the prioritization and dependency requirements between the objects. This in certain cases allow mutiple QUIC streams to transport the objects with in a single group".
Then add few non-normative examples for the same
- Say example for temporal scalability
- Example where all the objects are of same priority and hence has one Peep.

draft-ietf-moq-transport.md
```diff
@@ -302,9 +302,58 @@ payload. This includes the underlying encoding, compression, any end-to-end
 encryption, or authentication. A relay MUST NOT combine, split, or otherwise
 modify object payloads.
 
+## Peeps {#model-peeps}
+
+A peep is a collection of one or more objects and is a sub-unit of a group
```
ianswett, line 307: I believe order matters, so maybe sequence instead of collection?

```suggestion
A peep is a sequence of one or more objects and is a sub-unit of a group
```

draft-ietf-moq-transport.md
```diff
@@ -302,9 +302,58 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
+({{model-group}}). A peep consists of a set of objects within a group that
+have a dependency and priority relationship consistent with sharing a QUIC
+stream.
```
ianswett, line 310:
```suggestion
have a dependency and priority relationship benefitting from in-order delivery,
and sharing a QUIC stream.
```

draft-ietf-moq-transport.md
```diff
@@ -302,9 +302,58 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
+
+When a Track's forwarding preference (see {{object-fields}}) is "Track" or
```
ianswett, line 312: Why not allow Peeps to be in "Datagram"?  If they fit, they fit?

Admittedly, the extra overhead of a Peep ID is unnecessary

draft-ietf-moq-transport.md
```diff
@@ -302,9 +302,58 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
+"Datagram", Groups do not contain Peeps, no Peep IDs are assigned, and the
+description in the remainder of this section does not apply.
+
+QUIC streams offer in-order reliable delivery, differentiation of priority in
+delivery, and the ability to cancel sending and retransmission of data.
+
+Every object belongs to exactly one Peep.
+
+Original publishers assign each Peep a Peep ID, and do so as they see fit.
```
ianswett, line 321: What's the scope of the Peep ID space, a Group?

draft-ietf-moq-transport.md
```diff
@@ -302,9 +302,58 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
+However, the following strategy is RECOMMENDED:
+
+- Higher object IDs are dependent on lower object IDs in the same peep.
+
+- If two objects are dependent on the same lower object ID, but are not
+dependent on each other, one ought to assigned the same peep as the lower
+ID, but the other ought to be in a different peep. The second peep would
+ideally have a lower priority than the one with two objects.
```
ianswett, line 329:
```suggestion
dependent on each other, the higher priority Object ought to be included in the same
peep as the lower object ID, and the other Object ought to be in a different peep.
```

ianswett (commented):
This looks good overall, some comments/questions

draft-ietf-moq-transport.md
LPardue: I'd extract the priority part and put it in another sentence afterwards such as

"Furthermore, many implementations offer the ability to control the relative priority of streams, which allows control over the scheduling of sending data on active streams"

Or something like that

draft-ietf-moq-transport.md
suhasHere: +1 on making it explicit that Peeps are scoped to a given Group

vasilvv:
> > How does one close a peep? Currently, in our implementation, we rely in combination of delivery preference and object status for that.
> 
> I think a Stream FIN would be sufficient.

How am I supposed to know when to send a FIN?

martinduke:
> > > How does one close a peep? Currently, in our implementation, we rely in combination of delivery preference and object status for that.
> > 
> > 
> > I think a Stream FIN would be sufficient.
> 
> How am I supposed to know when to send a FIN?

The publisher sends a FIN when the Peep is done. The relay knows a Peep is complete when it receives the FIN from upstream.

afrind:
> The publisher sends a FIN when the Peep is done. The relay knows a Peep is complete when it receives the FIN from upstream.

Is there ever a case where a publisher needs to close a stream and ensure all data on it is delivered, but does not know that the peep is done, or knows it is not done.  For example - in response to an unsubscribe or graceful closure.

vasilvv:
> The publisher sends a FIN when the Peep is done. The relay knows a Peep is complete when it receives the FIN from upstream.

How is this represented in the cache?  Are FINs guaranteed to be always at the same position?

One scenario where the sender might FIN a peep earlier is if the next object in the peep is past subscription end, though I'm not sure how we'd hit this in practice.

draft-ietf-moq-transport.md
martinduke: I invite you to start a discussion thread about naming; if there is consensus I am happy to adopt that name.

draft-ietf-moq-transport.md
martinduke: I don't think we want to make implementers figure out the implications of QUIC streams of their own. Is there harm in having a SHOULD here?

draft-ietf-moq-transport.md
martinduke: This would work, but IIUC we don't currently allow multiple objects in a Datagram, so this seems like a different discussion.

martinduke:
> > The publisher sends a FIN when the Peep is done. The relay knows a Peep is complete when it receives the FIN from upstream.
> 
> How is this represented in the cache? Are FINs guaranteed to be always at the same position?
> 
> One scenario where the sender might FIN a peep earlier is if the next object in the peep is past subscription end, though I'm not sure how we'd hit this in practice.

Good point, I've added an end of Peep object.

martinduke:
OK, I believe I've fully addressed all comments except for those I've left unresolved. @suhasHere @ianswett  If you are satisfied with my response, please resolve the comment.

draft-ietf-moq-transport.md
suhasHere: May be I wasn't clear in my above comment.  
Yes , each peep must maps to a QUIC Stream and making it explicit is good thing. 
All i was trying to say the conditions listed are examples and not conditions as such. If it was listed under non-normative example of few possible ways a group can have peeps in it , i think it would be more clearer.

draft-ietf-moq-transport.md
```diff
@@ -555,7 +610,7 @@ The subscriber indicates the priority of a subscription via the
 Subscriber Priority field and the original publisher indicates priority
 in every stream or datagram header.  As such, the subscriber's priority is a
 property of the subscription and the original publisher's priority is a
-property of the Track and the Objects it contains. In both cases, a lower
+property of the Track and the Peeps it contains. In both cases, a lower
```
suhasHere, line 613: We should s/Peeps/Objects. 

Objects have priorities and the objects in a group that has same priority can be assigned to a given peep (since peep == stream).

draft-ietf-moq-transport.md
```diff
@@ -582,9 +637,10 @@ the corresponding SUBSCRIBE_OK.
 
 Within the same Group, and the same priority level,
 Objects with a lower Object Id are always sent before objects with a
-higher Object Id, regardless of the specified Group Order. If the priority
-varies within a Group, higher priority Objects are sent before lower
-priority Objects.
+higher Object Id, regardless of the specified Group Order. If the group
+contains more than one Peep and the priority varies between these Peeps,
+higher priority Peeps are sent before lower priority Peeps, and Objects
+in peeps are sent in increasing Object Id order.
```
suhasHere, line 643: I don't think the new suggested changes are needed since the paragraph starts with "Within the same Group, and the same priority level," which implies if there were multiple peeps and differ in priorities, they are sent in the priority order.

draft-ietf-moq-transport.md
```diff
@@ -1637,7 +1658,7 @@ TODO: figure out how a relay closes these streams
 
 When a stream begins with `STREAM_HEADER_TRACK`, all objects on the stream
 belong to the track requested in the Subscribe message identified by `Subscribe
-ID`.  All objects on the stream have the `Publisher Priority` specified in the
+ID`.  All objects on the stream have the `Publisher Priority' specified in the
```
suhasHere, line 1661: nit s/'/`

draft-ietf-moq-transport.md
```diff
@@ -1670,32 +1691,32 @@ The Object Status field is only sent if the Object Payload Length is zero.
...
+`Object Forwarding Preference` = `Peep`.
 
-To send an Object with `Object Forwarding Preference` = `Group`, find the open
-stream that is associated with the subscription and `Group ID`, or open a new
-one and send the `STREAM_HEADER_GROUP` if needed, then serialize the following
-fields.
+To send an Object with `Object Forwarding Preference` = `Peep`, find the open
+stream that is associated with the subscription, `Group ID` and `Peep ID`,
+or open a new one and send the `STREAM_HEADER_PEEP`. Then serialize the
+ollowing fields.
```
suhasHere, line 1719: type in word following

suhasHere (commented):
thansk @martinduke for the new set of edits. I just had few nits/clarification points. PLease let me know your thoughts. 
I am liking the general direction.

draft-ietf-moq-transport.md
martinduke: No, I think I understand. You would prefer that there is no normative guidance, but that we provide an example of a separate mapping. I would prefer that we have a SHOULD with the recognition that there are reasons a publisher might choose a different mapping.

I'm not going to lie down in the road over this, but would like to hear the opinions of others. If most agree with you, I'm happy to make the change.

draft-ietf-moq-transport.md
martinduke: I guess I don't follow. What would it mean for two objects in a peep to have different priorities? Object priorities make sense in the current document, because each object can be its own stream. But in this PR, that's just a case where each Peep is only one object.

draft-ietf-moq-transport.md
martinduke: I agree that strictly speaking, the logic here is somewhat redundant. But I think the new text makes it clearer. Do you disagree?

draft-ietf-moq-transport.md
suhasHere: What i meant it  is , since peeps are priority grouping within a stream, all the objects with a same priority value belong to same peep. This enables delivering peep as a QUIC Stream.  Application sets priorities on objects and what is cached are objects.

martinduke: I understand what you mean now.

I had not envisioned that objects with the same priority must be in the same peep. For example, two objects could have the same priority but no dependency relationship. But if no one has a use case for that, I'm happy to make it impossible for two peeps to have the same priority: just make the Peep ID also be the priority (thus also shrinking the wire image).

martinduke:
No normative language, summarize what streams are.

draft-ietf-moq-transport.md
martinduke: Interim: Remove normative language, describe what streams do.

martinduke: Increasing object ID is a MUST

draft-ietf-moq-transport.md
martinduke: Interim: Ian is OK with this

draft-ietf-moq-transport.md
martinduke: Interim: people like all peeps having a different priority, but Alan may have a use case? We'll leave it as is for now.

martinduke:
@afrind wants to make a proposal, wait for that

wilaw:
For bikesheading on the name, I make these two proposals:
1. "Stream" - this correlates directly with the mapping to QUIC. Simple to comprehend and explain. 
2. "Block" - this denotes contiguity and avoids overloaded use of the term "stream".

draft-ietf-moq-transport.md
afrind: The language is confusing because Peeps don't have priority, Objects do.  All objects in the same peep have the same priority, so you can call that the priority of the Peep, but the wording isn't totally clear.

Mandating increasing object ID within a peep is a good thing, but using object ID as the tie-breaker between peeps with equivalent object priority is not good.

Consider this case, where all objects have priority 128:

Peep/Stream 1:

Object 1, Object 3

Peep/Stream 2:

Object 2, Object 4

Let' assume all data has been written to the QUIC stack.  At the point when all the bytes of Object 1 are sent, the priority of Peep/Stream 2 needs to change to be higher than Peep/Stream 1.  This is way more complex than most implementers what to deal with.

So we can:

1. Make Peep ID the tie-breaker - with the same priority, lower peep ID goes first
2. Collapse Peep ID and Priority:
  2a. Use 62 bits - priority gets larger in kind of an ugly way
  2b. Use 8 bits - restricts maximum peeps per group, which probably breaks some object-per-stream cases
3. Leave the language as is, which is super hard to implement
4. Leave the wire image the same, but add a SHOULD NOT set two peeps to the same priority (but realizing that they will be the same when numPeeps > 256).
5. Something else?

suhasHere: +1 on keeping priority as part of the Object and not a Peep's property. I think we need  to fix text to make sure that's reflected so. 
Also option 1 would be my choice

martinduke: @suhasHere, I have no idea how priority can meaningfully be an Object property? Priorities are implemented at the QUIC stream level, so how do two object in the same stream have different priority?

martinduke:
I have made all changes I know to do. I am awaiting @afrind making a concrete proposal on the question of whether or not two Peeps can have the same priority, and the wire image implications downstream from that.

draft-ietf-moq-transport.md
suhasHere: So a MOQ Application interfaces with moq object model and assigns priorities to the objects. The moq stack has a new object to send ( output of the priority decision algorithm), it looks at the object priority and does one of the following to match moq object priority to QUIC Stream.  
   - create a new stream , set the priority of the stream to match the priority of the object or
   - If there is already a stream created (as per first step) , pick that stream and send the object over that stream.
   
When relays receive the objects, the canonical object has priority saved as well. So relay for downstream can apply the similar logic.

draft-ietf-moq-transport.md
```diff
@@ -310,9 +310,56 @@ payload. This includes the underlying encoding, compression, any end-to-end
 encryption, or authentication. A relay MUST NOT combine, split, or otherwise
 modify object payloads.
 
+## Peeps {#model-peep}
+
+A peep is a collection of one or more objects and is a sub-unit of a group
+({{model-group}}). A peep consists of a set of objects within a Group that
```
ianswett, line 316:
```suggestion
A peep is a sequence of one or more objects from the same group
({{model-group}}) in ascending order by Object ID. Objects in a peep
```

draft-ietf-moq-transport.md
```diff
@@ -310,9 +310,56 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
+have a dependency and priority relationship consistent with sharing a QUIC
+stream. In some cases, a Group will be most effectively delivered using more
+than one stream.
```
ianswett, line 319:
```suggestion
than one QUIC stream.
```

draft-ietf-moq-transport.md
```diff
@@ -310,9 +310,56 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
+
+When a Track's forwarding preference (see {{object-fields}}) is "Track" or
+"Datagram", Groups do not contain Peeps, no Peep IDs are assigned, and the
+description in the remainder of this section does not apply.
```
ianswett, line 323:
```suggestion
"Datagram", Objects are not sent in Peeps, no Peep IDs are assigned, and the
description in the remainder of this section does not apply.
```

draft-ietf-moq-transport.md
```diff
@@ -310,9 +310,56 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
+
+QUIC streams offer in-order reliable delivery and the ability to cancel sending
+and retransmission of data. Furthermore, many implementations offer the ability
+to control the relative priority of streams, which allows control over the
+scheduling of sending data on active streams.
+
+Every object within a Group belongs to exactly one Peep.
```
ianswett, line 330: What if an Object appears in multiple Peeps?  Should the session be closed with a protocol error or something similar if it's detected?

draft-ietf-moq-transport.md
```diff
@@ -310,9 +310,56 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
+
+Original publishers assign each Peep a Peep ID, and do so as they see fit.  The
+scope of a Peep ID is a Group, so Peeps from different Groups MAY share a Peep
+ID without implying any relationship between them. In general, publishers assign
+objects to peeps in order to leverage the features of QUIC streams as described
+above.
+
+An example strategy for using QUIC stream properties follows. If object A is
+dependent on object B, then delivery of A can follow B, i.e. B and A can be
```
ianswett, line 339: Nit: usually I see examples B following A in examples?

draft-ietf-moq-transport.md
```diff
@@ -310,9 +310,56 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
+usefully delivered over a single QUIC stream. Furthermore, in this example:
+
+- If an object is dependent on all previous objects in a Peep, it is added to
+that Peep.
+
+- If an object is not dependent on some of the objects in a Peep, it goes in
+a different Peep.
+
+- There are often many ways to compose Peeps that meet these criteria. Where
+possible, choose the composition that results in the fewest Peeps in a group.
```
ianswett, line 349: This last point about minimizing the number of streams/peeps used could be important.  I wonder if that deserves a RECOMMENDED or SHOULD because it minimizes resource usage and the possibility of running out of available streams.

draft-ietf-moq-transport.md
```diff
@@ -310,9 +310,56 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
+
+For example, imagine that Object 3 and Object 2 are both dependent on Object
+1, but are have no depedency relationship with each other. It would
+therefore be appropriate for Objects 1 and 3 to be in one peep and Object 2
+in another, with lower priority. The separate stream prevents delivery of
+Object 3 from being blocked by the loss of Object 2.
+
+Alternatively, objects 1 and 2 could be in one peep and 3 in another. The
+optimal choice would depend on the dependencies of later objects.
```
ianswett, line 358:
```suggestion
```

I'm inclined to remove this example entirely.

draft-ietf-moq-transport.md
```diff
@@ -310,9 +310,56 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
+
 ## Groups {#model-group}
 
-A group is a collection of objects and is a sub-unit of a track
+A group is a collection of one or more peeps and is a sub-unit of a track
```
ianswett, line 362:
```suggestion
A group is a collection of objects and is a sub-unit of a track
```
I think it makes more sense to keep the existing definition here.

draft-ietf-moq-transport.md
ianswett: For now, I think making Peep ID the tiebreaker is probably best, but it does feel very duplicative to the priority then.

I also question whether a limit of 256 peeps in a Group is a problem?  I can't even think of a use case where one needs 8 peeps.

draft-ietf-moq-transport.md
```diff
@@ -1543,59 +1593,17 @@ are beyond the end of a group or track.
          group. This is sent right after the last object in the
          track. This SHOULD be cached.
 
+* 0x5 := Indicates end of Peep. Object ID is one greater than the largest
```
ianswett, line 1596: Do we need this?  I'd be inclined not to add it until we have a use case.

ianswett (commented):
Looks very good overall, some detailed suggestions.

draft-ietf-moq-transport.md
martinduke: I think this is the same as any other object that appears twice on the wire with different properties. If the sequences match but the payloads are different, what do we do?

draft-ietf-moq-transport.md
martinduke: There was a long discussion up-thread where Victor asked for this. As a relay, I don't know when to close the stream until I see this. The alternative is to leave all streams open until the Group is done, which I suppose would work?

draft-ietf-moq-transport.md
martinduke: I would be fine with that, but (1) I got pretty firm "no normative text" feedback, and more importantly (2) in the absence of any other normative guidance, I don't know how to write this. If there are no guidelines whatsoever for peep assignment except "minimize # of peeps", that collapses pretty quickly to stream-per-group.

martinduke:
Why have Peep ID be a tiebreaker if we also have priority? Just let the streams have the same priority!

draft-ietf-moq-transport.md
ianswett: Fair, i was more thinking about the exact same Object being included in two different peeps.

draft-ietf-moq-transport.md
ianswett: What you have now is fine as guidance, so given there was a lack of interest in normative text, let's keep it as is.

draft-ietf-moq-transport.md
ianswett: Interesting point.  I'm fine with adding it then.

draft-ietf-moq-transport.md
martinduke: Every object has one peep ID on the wire. So if the same object is delivered twice with different Peep IDs, that's like any other case where the same objects arrives with different metadata or content.

Maybe the interesting error case is a QUIC stream that carries multiple peeps.

martinduke: Correction: a QUIC stream can't carry multiple peeps because the Peep ID is only expressed once in the stream.

draft-ietf-moq-transport.md
```diff
@@ -582,9 +621,12 @@ the corresponding SUBSCRIBE_OK.
...
 Objects with a lower Object Id are always sent before objects with a
-higher Object Id, regardless of the specified Group Order. If the priority
-varies within a Group, higher priority Objects are sent before lower
-priority Objects.
+higher Object Id, regardless of the specified Group Order. If the group
+contains more than one Peep and the priority varies between these Peeps,
+higher priority Peeps are sent before lower priority Peeps. If the priority of
+two Peeps in a Group are equal, there is no normative requirement for which of
+those Peeps is sent next. Within a Peep, Objects MUST bs sent in increasing
+Object ID order.
```
afrind, line 629: Individual Comment: 

We had stream-per-object and object ID was the tie breaker.  To get the same functionality with stream-per-peep, you need peep ID as a tie breaker.  You might need > 256 peeps per group when you are using object-per-peep (eg: stream-per-object).

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -310,6 +310,45 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
+usefully delivered over a single QUIC stream. Furthermore, in this example:
+
+- If an object is dependent on all previous objects in a Peep, it is added to
+that Peep.
+
+- If an object is not dependent on all of the objects in a Peep, it goes in
+a different Peep.
+
+- There are often many ways to compose Peeps that meet these criteria. Where
+possible, choose the composition that results in the fewest Peeps in a group.
```
ianswett, line 349:
```suggestion
possible, choose the composition that results in the fewest Peeps in a group
to minimize the number of QUIC streams used.
```

draft-ietf-moq-transport.md
```diff
@@ -1543,59 +1587,17 @@ are beyond the end of a group or track.
          group. This is sent right after the last object in the
          track. This SHOULD be cached.
 
+* 0x5 := Indicates end of Peep. Object ID is one greater than the largest
```
fluffy, line 1590: I think how we indicate end of group /  end of peep deservers it own discussion and PR. I think I like this but see it as separable issue.

draft-ietf-moq-transport.md
```diff
@@ -1670,32 +1672,32 @@ The Object Status field is only sent if the Object Payload Length is zero.
 
 A publisher MUST NOT send an Object on a stream if its Group ID is less than a
 previously sent Group ID on that stream, or if its Object ID is less than or
-equal to a previously sent Object ID within a given group on that stream.
+equal to a previously sent Object ID with the same Group ID.
```
fluffy, line 1675: I think we should add a note here that there could be other streams with older objects. This is going to happen in reconnect corner cases.

fluffy (changes requested):
Conceptually agree with the big picture here but bunch of details I think need to be sorted out before we merge.

draft-ietf-moq-transport.md
martinduke: do you want me to revert this change?

draft-ietf-moq-transport.md
```diff
@@ -310,6 +310,46 @@ payload. This includes the underlying encoding, compression, any end-to-end
 encryption, or authentication. A relay MUST NOT combine, split, or otherwise
 modify object payloads.
 
+## Peeps {#model-peep}
```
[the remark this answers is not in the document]
afrind: > we are not at the point

I assume you meant to say "we *are* at the point"?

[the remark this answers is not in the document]
martinduke: I am happy to change it once someone identifies a consensus, but don't want to hold it up over a bikeshed

draft-ietf-moq-transport.md
```diff
@@ -555,7 +595,7 @@ The subscriber indicates the priority of a subscription via the
 Subscriber Priority field and the original publisher indicates priority
 in every stream or datagram header.  As such, the subscriber's priority is a
 property of the subscription and the original publisher's priority is a
-property of the Track and the Objects it contains. In both cases, a lower
+property of the Track and the Peeps it contains. In both cases, a lower
```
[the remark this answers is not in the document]
martinduke: this seems like semantics? the wire image only allows you to send one priority per peep, so whatever metaphysical priority the application assigns to objects has no discernible effect inside MoQT.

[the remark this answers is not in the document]
afrind: > We should jut say all the objects in the same group with the same peep-id MUST have the same publisher priority.

This is necessarily true from our wire encoding, which I think everyone is happy with, so we're only talking about how best to convey the concepts.

draft-ietf-moq-transport.md
martinduke: FYI there was a discussion on this point up thread.

draft-ietf-moq-transport.md
```diff
@@ -582,9 +622,11 @@ the corresponding SUBSCRIBE_OK.
 
 Within the same Group, and the same priority level,
 Objects with a lower Object Id are always sent before objects with a
-higher Object Id, regardless of the specified Group Order. If the priority
-varies within a Group, higher priority Objects are sent before lower
-priority Objects.
+higher Object Id, regardless of the specified Group Order. If the group
+contains more than one Peep and the priority varies between these Peeps,
+higher priority Peeps are sent before lower priority Peeps. If the specified
+priority of two Peeps in a Group are equal, the lower Peep ID has priority.
```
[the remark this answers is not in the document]
afrind: This is preserving what is in the current draft for stream-per-object - there was never a way to fair share objects in this mode, because Object ID was the tiebreaker.  

Adding Peeps opens the question of if we want to be able to fair share peeps.  In the interim I argued for a use case, but after consulting with the originators of said use case, I decided that a strict ordering is probably what they want.  I propose we advance this PR keeping as close to the existing behavior (Peep ID tiebreakers), and revisit fair-sharing them separately.

draft-ietf-moq-transport.md
afrind: Reverting this change will break my relay, so I don't want you to revert it.  However, I'm ok with switching it to "PeepContinues" and if a Peep ends without it, the stream FIN indicates end-of-peep.

draft-ietf-moq-transport.md
afrind: Let's add that in the batch of PRs that deal with reconnect, etc.

afrind (commented):
Individual Comments

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
fluffy: oops yes - sorry - i did mean we are at the point we should pick a good name.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
fluffy: Martin, I think you are missing the key point of what a moq priority is. Imagine QUIC had no priorities at all and at the QUIC layer things acted in the way QUIC would if the priorities were all set to the same thing.  In this case, MoQ priorities are still super useful and help the application decide which object gets sent next to the QUIC stack. And for implementations that don't send stuff to the QUIC stack until the congestion window opens up for the stream, this allows them moq level to get a wide range of priorities behavior to happen. The reason real time stuff is implemented this way is because once you put something on the QUIC stack, a huge limitation of QUIC APIs is there is no way to remove it, so if it expires before it is sent, you use up the bandwidth anyways. 

So what I am trying to say is priorities in MoQ are much more than just telling the quic stream what priority to use. They change the oder of what gets put into the quic stack.

draft-ietf-moq-transport.md
ianswett: From an Object model perspective, I think that's a good direction, @fluffy 

The wire format as proposed allows one to violate that Object model, which is where I was going with my normative suggestion.  It also allows Object ID to decrease, which we might want to fix in a follow-up PR.

draft-ietf-moq-transport.md
fluffy: Fair enough - that makes sense to deal with it then

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
fluffy: What would you think of instead of PeepID, calling it StreamAlias. Define it as things with different StreamAlias MUST NOT be on the same QUIC stream and things with the same StreamAlias inside the same group SHOULD be on the same QUIC stream. 

I'm not sure I love this name but I'm just trying to provide some sort of constructive suggestions.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
fluffy: this seems just less flexible than the alternative I can't see why we would do that

draft-ietf-moq-transport.md
fluffy: oh, it was marked resolved so I missed it. I will go comment on it

draft-ietf-moq-transport.md
fluffy: So this is basically the tombstone approach and I am a fan of that. However, I think we need to do this the same way end a group and end a track and this. Given we took the "last flag" approach to end of group and end track. I think we should stay consistent here and do this the same way. 

So my proposal is basically we put an "End of Peep" marker in the Object Status. I would restructure the Object Status to have bit 0 indicate End of Peep, bit 1 indicate End of Group, bit 2 indicate End of Track.  While we are at it, having a non optional Object Status in every Object would probably use less bandwidth than making it optional. 

Thoughts on this approach ?

ianswett: I'm ok with that, but I think I'd rather do it in a separate PR.

Also, I realize that I'm not sure what the difference is between "Group does not exist" and "End of Group" with am Object ID of 0?  A related qustion is what the Object ID should be for "Group does not exist"?  I'm starting to think "Group does not exist" and it's the only one where Object ID has a meaningless value.

ianswett: https://github.com/moq-wg/moq-transport/pull/509

ianswett:
> For bikesheading on the name, I make these two proposals:
> 
> 1. "Stream" - this correlates directly with the mapping to QUIC. Simple to comprehend and explain.
> 2. "Block" - this denotes contiguity and avoids overloaded use of the term "stream".

Block and Stream work for me.

I was going to say that I prefer "QUIC Stream", but given this might be over WebTransport, technically it'd be a WebTransport stream, so defining the term "Stream" at the top of the document should make that clear.

draft-ietf-moq-transport.md
ianswett: Reflecting on whether we need this, I think a stream FIN means the Peep is closed.  Otherwise someone could split a multi-object Peep into single Object streams, which is definitely not the intent and maybe could be used as an attack (thinking segment smack, etc)?

Session migration is important, but in that case I think streams should be reset or the session just suddenly goes away to indicate there is more data that couldn't be sent.  This is a reminder that writing up session migration could be really helpful for understanding how streams and MoQ interact.

But happy to bikeshed this issue afterwards.

draft-ietf-moq-transport.md
martinduke: Offline: Alan is fine with peep ID = priority, meaning that two Peeps can't have the same priority. Text is updated, closing this comment.

draft-ietf-moq-transport.md
martinduke: Text now has Peep ID = tiebreaker, @afrind is OK with it. Closing this comment, though Cullen has complaints downthread. Further discussion there.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
martinduke: This only works for the first hop for the publisher; relays will only be able to use the priority expressed on the wire, which is Peep-level.

That said, I'm happy to rewrite this to talk about what's on the wire and avoid any metaphysical discussion of object priority or telling MoQT what API to present here.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
martinduke: Cullen's use case: the relay might have received packets out of order, possibly because it's non-MoQT upstream. In this case, the only choices are drop it, send out of order in the stream, or open new stream. third choice is best.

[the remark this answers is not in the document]
fluffy: QuicStreamIndicator as name ?

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
martinduke: Cullen makes the point that the current algorithm works, if one views objects as the thing that holds priority. Within the group, the most important objects (peeps) will go first.

draft-ietf-moq-transport.md
fluffy: Martin convinced me this was fine and I was confused. And the more I thought about, we don't get many gains by closing the streams before the end of group so I would be fine with this as is and also fine with we just use end of group.

suhasHere:
One more name suggestion : SubGroup

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
martinduke: took the normative text (ish), closing although the naming bikeshed continues.

draft-ietf-moq-transport.md
martinduke: I don't understand how this allows Object ID to decrease, unless you mean that a higher object ID from one Peep could come before a lower one in a different peep. But that's a feature not a bug.

draft-ietf-moq-transport.md
martinduke: Talking offline to @fluffy, he is now fine with the end of Peep object, though we agreed that the publisher MAY use an end of Group object to implicitly close all open Peeps.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
martinduke: @afrind makes the point that it's hard to implement this because you'd have to change stream priorities mid-stream as the object IDs change, which QUIC APIs are generally bad it.

I'm not deeply concerned either way, as priorities are somewhat advisories. I'd just like these two to come to some agreement.

draft-ietf-moq-transport.md
ianswett: Sorry, I think my comments were confusing, please disregard them for now.

ianswett (approved):

ianswett:
Merging with the understanding that we'll continue to discuss how prioritization should work.

PR #498, opened 2024-08-05
SUBSCRIBE_NAMESPACE

### LINKED ISSUE #253: Wildcard SUBSCRIBE
SUBSCRIBE today requires a full track name, and specifies exactly one track.  Do we need a subscription mechanism that matches more than one track, including tracks that may not exist currently, but may come into existence in the future?

### LINKED ISSUE #484: Discovery of publishers via a relay instead of catalog
This is a redux of #228, now that I've heard from multiple other wg participants they are now interested in such a mechanism.

Consider using moq for a chat or other application where a relay has many publishers for the same application.  In order for subscribes to be routed to the correct publishing endpoints, each publisher/participant needs to announce a unique namespace (this is how moq-chat operates, see https://afrind.github.io/draft-frindell-moq-chat/draft-frindell-moq-chat.html#section-4.4).

This creates an issue: how do participants learn about the namespaces/tracks other clients have ANNOUNCED to the relay within the same chat room? I originally filed #228 thinking there should be a mechanism by which a client can declare their interest in new announcements and become a catalog-less means of track discovery. Based on wg feedback at the time, I updated moq-chat to use a different discovery mechanism involving a simple catalog.

Here's an alternate proposal:

```
SUBSCRIBE_NAMESPACE {
   Track Namespace Prefix (b),
   Auth Info (b)
 }
```

This would inform the relay to forward any ANNOUNCE with a prefix match, if authorized.

So in moq-chat, each new participant would do this:

```
ANNOUNCE(namespace="moq-chat/id/participant/<name>")
SUBSCRIBE_NAMESPACE(prefix="moq-chat/id/participant/")
```

And a relay would reply with:

```
ANNOUNCE_OK
SUBSCRIBE_NAMESPACE_OK
# existing announcements matching namespace prefix
ANNOUNCE(namespace="moq-chat/id/participant/<member1>") 
ANNOUNCE(namespace="moq-chat/id/participant/<member2>")
ANNOUNCE(namespace="moq-chat/id/participant/<member3>")
...
```

It would also forward any new ANNOUNCEs or ANNOUNCE_CANCELs matching the prefix to the participant.

The new participant then issues standard SUBSCRIBE to each existing chat member
```
SUBSCRIBE(namespace="moq-chat/id/participant/<member1>", name="chat") 
SUBSCRIBE(namespace="moq-chat/id/participant/<member2>", name="chat") 
...
```

This would remove the need for a catalog in some multi-publisher applications (eg: key distribution https://suhashere.github.io/moq-e2ee-mls/draft-jennings-moq-e2ee-mls.html), but also keep the scope of ANNOUNCE the same as it is today.

afrind:
This message suite allows subscribers to express interest in ANNOUNCE/UNANNOUNCE for set of namespaces that match a prefix.  To accomplish safe prefix matching, Track Namesapce is redefined from a single byte array to a N-Tuple of byte arrays.

Fixes #484
Closes #253

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -282,7 +282,7 @@ x (f):
 
 x (t):
 
-: Indicates that x consists of a variable length integer encoding as
+: Indicates that x is a tuple, consisting of a variable length integer encoded as
   described in ({{?RFC9000, Section 16}}), followed by that many variable length
   tuple fields, each of which are encoded as (b) above.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -280,11 +280,11 @@ x (f):
   protocol error and SHOULD terminate the session with a Protocol
   Violation ({{session-termination}}).
 
-x (t):
+x (tuple):
 
-: Indicates that x is a tuple, consisting of a variable length integer encoded as
-  described in ({{?RFC9000, Section 16}}), followed by that many variable length
-  tuple fields, each of which are encoded as (b) above.
+: Indicates that x is a tuple, consisting of a variable length integer encoded
+  as described in ({{?RFC9000, Section 16}}), followed by that many variable
+  length tuple fields, each of which are encoded as (b) above.
 
 
 To reduce unnecessary use of bandwidth, variable length integers SHOULD
@@ -384,8 +384,8 @@ In MOQT, every track has a track name and a track namespace associated
 with it.  A track name identifies an individual track within the
 namespace.
 
-Track namespace is an N-tuple of bytes where N can be between 1 and 32.  The
-structured nature of Track Namespace allows relays and applications to
+Track namespace is an ordered N-tuple of bytes where N can be between 1 and 32.
+The structured nature of Track Namespace allows relays and applications to
 manipulate prefixes of a namespace. Track name is a sequence of bytes.
 
 In this specification, both the Track Namespace tuple fields and the Track Name
@@ -1127,7 +1127,7 @@ The format of SUBSCRIBE is as follows:
 SUBSCRIBE Message {
   Subscribe ID (i),
   Track Alias (i),
-  Track Namespace (t),
+  Track Namespace (tuple),
   Track Name (b),
   Subscriber Priority (8),
   Group Order (8),
@@ -1273,7 +1273,7 @@ successful authorization and acceptance of an ANNOUNCE message.
 ~~~
 ANNOUNCE_OK
 {
-  Track Namespace (t),
+  Track Namespace (tuple),
 }
 ~~~
 {: #moq-transport-announce-ok format title="MOQT ANNOUNCE_OK Message"}
@@ -1289,7 +1289,7 @@ failed authorization.
 ~~~
 ANNOUNCE_ERROR
 {
-  Track Namespace (t),
+  Track Namespace (tuple),
   Error Code (i),
   Reason Phrase (b),
 }
@@ -1315,7 +1315,7 @@ receiving an ANNOUNCE_CANCEL, it SHOULD close the session as a
 
 ~~~
 ANNOUNCE_CANCEL Message {
-  Track Namespace (t),
+  Track Namespace (tuple),
 }
 ~~~
 {: #moq-transport-announce-cancel-format title="MOQT ANNOUNCE_CANCEL Message"}
@@ -1332,7 +1332,7 @@ A TRACK_STATUS message MUST be sent in response to each TRACK_STATUS_REQUEST.
 
 ~~~
 TRACK_STATUS_REQUEST Message {
-  Track Namespace (t),
+  Track Namespace (tuple),
   Track Name (b),
 }
 ~~~
@@ -1346,18 +1346,18 @@ the set.
 
 ~~~
 SUBSCRIBE_NAMESPACE Message {
-  Track Namespace Prefix (t),
+  Track Namespace Prefix (tuple),
   Number of Parameters (i),
   Parameters (..) ...,
 }
 ~~~
 {: #moq-transport-subscribe-ns-format title="MOQT SUBSCRIBE_NAMESPACE Message"}
 
-* Track Namespace Prefix: An N-Tuple of byte fields which are matched against
-track namespaces known to the publisher.  For example, if the publisher is a
-relay that has received ANNOUNCE messages for namespaces ("example.com",
-"meeting", "123", "participant=100") and ("example.com", "meeting", "123",
-"participant=200"), a SUBSCRIBE_NAMESPACE for ("example.com", "meeting", "123")
+* Track Namespace Prefix: An ordered N-Tuple of byte fields which are matched
+against track namespaces known to the publisher.  For example, if the publisher
+is a relay that has received ANNOUNCE messages for namespaces ("example.com",
+"meeting=123", "participant=100") and ("example.com", "meeting=123",
+"participant=200"), a SUBSCRIBE_NAMESPACE for ("example.com", "meeting=123")
 would match both.
 
 * Parameters: The parameters are defined in {{version-specific-params}}.
@@ -1367,6 +1367,12 @@ SUBSCRIBE_NAMESPACE_ERROR.  If the SUBSCRIBE_NAMESPACE is successful,
 the publisher will forward any matching ANNOUNCE messages to the subscriber
 that it has not yet sent.  If the set of matching ANNOUNCE messages changes, the
 publisher sends the corresponding ANNOUNCE or UNANNOUNCE message.
+
+A subscriber cannot have overlapping namespace subscriptions outstanding at the
+same time.  If a publisher receives a SUBSCRIBE_NAMESPACE with a Track Namespace
+Prefix that is a prefix of an earlier SUBSCRIBE_NAMESPACE request or vice versa,
+it MUST respond with SUBSCRIBE_ERROR, with error code
+SUBSCRIBE_NAMESPACE_OVERLAP.
 
 The publisher MUST ensure the subscriber is authorized to perform this
 namespace subscription.
@@ -1386,7 +1392,7 @@ The format of `UNSUBSCRIBE_NAMESPACE` is as follows:
 
 ~~~
 UNSUBSCRIBE_NAMESPACE Message {
-  Track Namespace Prefix (t)
+  Track Namespace Prefix (tuple)
 }
 ~~~
 {: #moq-transport-unsub-ns-format title="MOQT UNSUBSCRIBE Message"}
@@ -1531,7 +1537,7 @@ publish tracks under this namespace.
 
 ~~~
 ANNOUNCE Message {
-  Track Namespace (t),
+  Track Namespace (tuple),
   Number of Parameters (i),
   Parameters (..) ...,
 }
@@ -1552,7 +1558,7 @@ within the provided Track Namespace.
 
 ~~~
 UNANNOUNCE Message {
-  Track Namespace (t),
+  Track Namespace (tuple),
 }
 ~~~
 {: #moq-transport-unannounce-format title="MOQT UNANNOUNCE Message"}
@@ -1568,7 +1574,7 @@ to a TRACK_STATUS_REQUEST message.
 
 ~~~
 TRACK_STATUS Message {
-  Track Namespace (t),
+  Track Namespace (tuple),
   Track Name (b),
   Status Code (i),
   Last Group ID (i),
@@ -1621,7 +1627,7 @@ namespace subscriptions.
 ~~~
 SUBSCRIBE_NAMESPACE_OK
 {
-  Track Namespace Prefix (t),
+  Track Namespace Prefix (tuple),
 }
 ~~~
 {: #moq-transport-sub-ns-ok format title="MOQT SUBSCRIBE_NAMESPACE_OK Message"}
@@ -1636,7 +1642,7 @@ failed SUBSCRIBE_NAMESPACE.
 ~~~
 SUBSCRIBE_NAMESPACE_ERROR
 {
-  Track Namespace Prefix (t),
+  Track Namespace Prefix (tuple),
   Error Code (i),
   Reason Phrase (b),
 }
@@ -1946,6 +1952,7 @@ TODO: fill out currently missing registries:
 * Setup parameters
 * Subscribe parameters
 * Subscribe Error codes
+* Subscribe Namespace Error codes
 * Announce Error codes
 * Message types
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -272,6 +272,12 @@ x (f):
   protocol error and SHOULD terminate the session with a Protocol
   Violation ({{session-termination}}).
 
+x (t):
+
+: Indicates that x consists of a variable length integer encoding as
+  described in ({{?RFC9000, Section 16}}), followed by that many variable length
+  tuple fields, each of which are encoded as (b) above.
+
 
 To reduce unnecessary use of bandwidth, variable length integers SHOULD
 be encoded using the least number of bytes possible to represent the
@@ -325,6 +331,20 @@ In MOQT, every track has a track name and a track namespace associated
 with it.  A track name identifies an individual track within the
 namespace.
 
+Track namespace is an N-tuple of bytes where N can be between 1 and 32.  The
+structured nature of Track Namespace allows relays and applications to
+manipulate prefixes of a namespace. Track name is a sequence of bytes.
+
+In this specification, both the Track Namespace tuple fields and the Track Name
+are not constrained to a specific encoding. They carry a sequence of bytes and
+comparison between two Track Namespace tuple fields or Track Names is done by
+exact comparison of the bytes. Specifications that use MoQ Transport may
+constrain the information in these fields, for example by restricting them to
+UTF-8. Any specification that does needs to specify the canonicalization into
+the bytes in the Track Namespace or Track Name such that exact comparison works.
+
+### Scope {#track-scope}
+
 A MOQT scope is a set of servers (as identified by their connection
 URIs) for which the tuple of Track Name and Track Namespace are
 guaranteed to be unique and identify a specific track. It is up to
@@ -340,15 +360,6 @@ MOQT does not provide any in-band content negotiation methods similar to
 the ones defined by HTTP ({{?RFC9110, Section 10}}); if, at a given
 moment in time, two tracks within the same scope contain different data,
 they have to have different names and/or namespaces.
-
-In this specification, both the Track Namespace and the Track Name are
-not constrained to a specific encoding. They carry a sequence of
-bytes and comparison between two Track Namespaces or Track Names is
-done by exact comparison of the bytes. Specifications that use MoQ Transport
-may constrain the information in these fields, for example by restricting
-them to UTF-8. Any specification that does needs to specify the
-canonicalization into the bytes in the Track Namespace or Track Name
-such that exact comparison works.
 
 ### Connection URL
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -272,6 +272,12 @@ x (f):
   protocol error and SHOULD terminate the session with a Protocol
   Violation ({{session-termination}}).
 
+x (t):
+
+: Indicates that x consists of a variable length integer encoding as
+  described in ({{?RFC9000, Section 16}}), followed by that many variable length
+  tuple fields, each of which are encoded as (b) above.
+
```
suhasHere, line 280: thanks for this.

draft-ietf-moq-transport.md
```diff
@@ -325,6 +331,20 @@ In MOQT, every track has a track name and a track namespace associated
 with it.  A track name identifies an individual track within the
 namespace.
 
+Track namespace is an N-tuple of bytes where N can be between 1 and 32.  The
```
suhasHere, line 334: not sure if we need to limit the number of tuples. but we should instead have limit on the overall length.

suhasHere: I wonder if we need this to be "Ordered N-tuple" .

draft-ietf-moq-transport.md
```diff
@@ -851,9 +870,9 @@ these parameters to appear in Setup messages.
 #### AUTHORIZATION INFO Parameter {#authorization-info}
 
 AUTHORIZATION INFO parameter (key 0x02) identifies a track's authorization
-information in a SUBSCRIBE or ANNOUNCE message. This parameter is populated for
-cases where the authorization is required at the track level. The value is an
-ASCII string.
+information in a SUBSCRIBE, SUBSCRIBE_NAMESPACE or ANNOUNCE message. This
+parameter is populated for cases where the authorization is required at the
+track level. The value is an ASCII string.
```
suhasHere, line 875: just noted, the value of token needs to be opaque bytes not ASCII string :-D

draft-ietf-moq-transport.md
```diff
@@ -325,6 +331,20 @@ In MOQT, every track has a track name and a track namespace associated
...
+structured nature of Track Namespace allows relays and applications to
+manipulate prefixes of a namespace. Track name is a sequence of bytes.
+
+In this specification, both the Track Namespace tuple fields and the Track Name
+are not constrained to a specific encoding. They carry a sequence of bytes and
+comparison between two Track Namespace tuple fields or Track Names is done by
+exact comparison of the bytes. Specifications that use MoQ Transport may
+constrain the information in these fields, for example by restricting them to
+UTF-8. Any specification that does needs to specify the canonicalization into
+the bytes in the Track Namespace or Track Name such that exact comparison works.
```
suhasHere, line 344: I think need to add text on what happens when certain tuples are missing in subscribes and how to ensure the match is happening at the right tuple or should we constrain saying a if tuple is not specified, rest of the tuple elments MUST be ignored for comparision

draft-ietf-moq-transport.md
```diff
@@ -1250,13 +1269,66 @@ A TRACK_STATUS message MUST be sent in response to each TRACK_STATUS_REQUEST.
...
+  Number of Parameters (i),
+  Parameters (..) ...,
+}
+~~~
+{: #moq-transport-subscribe-ns-format title="MOQT SUBSCRIBE_NAMESPACE Message"}
+
+* Track Namespace Prefix: An N-Tuple of byte fields which are matched against
+track namespaces known to the publisher.  For example, if the publisher is a
+relay that has received ANNOUNCE messages for namespaces ("example.com",
+"meeting", "123", "participant=100") and ("example.com", "meeting", "123",
```
suhasHere, line 1296: this example is bit misleading but the intent is fine. May be better way to say is
("example.com", "meeting/123", "participant/100")

The mix of a=b, vs split info is confusing.

draft-ietf-moq-transport.md
```diff
@@ -1250,13 +1269,66 @@ A TRACK_STATUS message MUST be sent in response to each TRACK_STATUS_REQUEST.
...
+"participant=200"), a SUBSCRIBE_NAMESPACE for ("example.com", "meeting", "123")
+would match both.
```
suhasHere, line 1298: this is a good example. Same comment as missing tuple element applies here too.  Say if we  had following tuple

ANNOUNCE messages for namespaces ("example.com", "location/earth",
"meeting/123", "participant/100") and ("example.com", "meeting/123", "location/earth",
"participant/200"), a SUBSCRIBE_NAMESPACE for ("example.com",  "location/earth",) should match both the tuples as well. 

Another example 

ANNOUNCE messages for namespaces ("example.com", "location/earth",
"meeting/123", "participant/100") and ("example.com", "meeting/123", 
"participant/200"), a SUBSCRIBE_NAMESPACE for ("example.com",  "location/earth") should only the first one. 

@afrind what are your thoughts ?

draft-ietf-moq-transport.md
```diff
@@ -1250,13 +1269,66 @@ A TRACK_STATUS message MUST be sent in response to each TRACK_STATUS_REQUEST.
...
+
+* Parameters: The parameters are defined in {{version-specific-params}}.
+
+The publisher will respond with SUBSCRIBE_NAMESPACE_OK or
+SUBSCRIBE_NAMESPACE_ERROR.  If the SUBSCRIBE_NAMESPACE is successful,
+the publisher will forward any matching ANNOUNCE messages to the subscriber
+that it has not yet sent.  If the set of matching ANNOUNCE messages changes, the
```
suhasHere, line 1305: Today ANNOUNCE message doesn't get sent to the subscriber.  Do we intend for the subscriber to issue Subscribe once it receives announce ?

suhasHere (commented):
@afrind Huge thanks for this PR. This is coming along pretty well.
I do have few clarification questions and few details that, if added, will make it easy to reason about.

draft-ietf-moq-transport.md
afrind: The limit on the number of tuples allows the relay to bound the complexity of a tree-structure used to implement tuple prefix matching.

Ordered tuple seems more specific, I'll update it.

draft-ietf-moq-transport.md
afrind: As I have defined it, you cannot omit a tuple in SUBSCRIBE_NAMESPACE.  eg: if a namespace is (a, b, c) and your prefix is (a, c), c is compared against b, and does not match.

draft-ietf-moq-transport.md
afrind: Please open a new issue :P

draft-ietf-moq-transport.md
afrind: Good point.  I'll restructure the example to make it more clear.

draft-ietf-moq-transport.md
afrind: My first cut at this only performs prefix matching, rather than arbitrary non-prefix matches as in your example.  The goal was to limit the complexity required to implement SUBSCRIBE_NAMESPACE efficiently in relays.

Do you have use cases that cannot be represented as prefixes that would require arbitrary match?

draft-ietf-moq-transport.md
afrind: The ANNOUNCE message can be sent by any publisher to any subscriber.  Today, the draft does not require ANNOUNCE before subscribe and this PR does not change that.  I was looking for text that indicates the optionality of ANNOUNCE and we don't seem to have it, but we should.

As for how SUBSCRIBE_NAMESPACE works, the function is to make the subscriber aware of the announcements known to the publisher (which could be a relay, but doesn't have to be), hence this PR communicates those in the form of the ANNOUNCE / UNANNOUNCE messages themselves.  The general pattern I expect is that the subscriber would issue some subscribes in response to the ANNOUNCE's received in response to SUBSCRIBE_NAMESPACE.

afrind (commented):
Thanks for the review.

draft-ietf-moq-transport.md
suhasHere: ah .. makes sense .. thanks

draft-ietf-moq-transport.md
suhasHere: No prefix match works. May be i misread the details.

draft-ietf-moq-transport.md
suhasHere: That makes sense.  thanks

draft-ietf-moq-transport.md
wilaw, line 875: Do we require different AUTH tokens for different components of the namespace, or do we assume that the token structure is sophisticated enough to match the tuple boundaries? For example, do we envisage one token that allows you to subscribe to 

`example.com/conference/session12345/`

but at the same time blocks you from

`example.com/conference`

wilaw:
Philosophical question - we now have a N-32 tuple of namespace components and then a track name that hangs off the end. What's the utility in that? Could not the track name just be one of the namespace components? 

**Can we simplify MOQT by getting rid of the concept of namespace and instead having a N-32 tuple track name?**

draft-ietf-moq-transport.md
afrind: I think we need auth schemes that allow for that distinction.

draft-ietf-moq-transport.md
afrind: I misread your examples too but I think we're still on the same page.

If the namespaces are (a, b, c, x) and (a, b, c, y), SUBSCRIBE_NAMESPACE(a, b) matches both.

If the namespaces are (a, b, c, x) and (a, c, y), SUBSCRIBE_NAMEPACE(a, b) only matches the first one.

afrind (commented):
Regarding Will's question:

> Can we simplify MOQT by getting rid of the concept of namespace and instead having a N-32 tuple track name?

This also crossed my mind.  At present, ANNOUNCE only announces namespace and SUBSCRIBE still requires a FULL match on namespace to route through a relay, so the distinction between namespace and name is serving a purpose.  I guess we could say the last component is always the track name.  I was hesitant in any case to completely undo track naming in this PR.

suhasHere:
This is a good start. I am sure we will fix aspects of it as we explore futher, but overall LGTM

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -280,6 +280,12 @@ x (f):
   protocol error and SHOULD terminate the session with a Protocol
   Violation ({{session-termination}}).
 
+x (t):
+
+: Indicates that x consists of a variable length integer encoding as
```
ianswett, line 285:
```suggestion
: Indicates that x is a tuple, consisting of a variable length integer encoded as
```

draft-ietf-moq-transport.md
```diff
@@ -1249,13 +1268,66 @@ A TRACK_STATUS message MUST be sent in response to each TRACK_STATUS_REQUEST.
...
 {: #moq-track-status-request-format title="MOQT TRACK_STATUS_REQUEST Message"}
 
+## SUBSCRIBE_NAMESPACE {#message-subscribe-ns}
+
+The subscriber sends the SUBSCRIBE_NAMESPACE control message to a publisher to
+request the current set of matching announcements, as well as future updates to
+the set.
+
+~~~
+SUBSCRIBE_NAMESPACE Message {
```
ianswett, line 1284: If I do SUBSCRIBE_NAMESPACE to "foo" and "foo/bar", do I get announces for "foo/bar" once or twice?

ianswett (approved):
Some comments, but overall LGTM

draft-ietf-moq-transport.md
```diff
@@ -1249,13 +1268,66 @@ A TRACK_STATUS message MUST be sent in response to each TRACK_STATUS_REQUEST.
...
-  Track Namespace (b),
+  Track Namespace (t),
   Track Name (b),
 }
 ~~~
 {: #moq-track-status-request-format title="MOQT TRACK_STATUS_REQUEST Message"}
 
+## SUBSCRIBE_NAMESPACE {#message-subscribe-ns}
+
+The subscriber sends the SUBSCRIBE_NAMESPACE control message to a publisher to
```
ianswett, line 1279: Are there use cases where you'd only want new ANNOUNCEs and not old ones?  ie: Disconnect and reconnect?

We could make ANNOUNCE a special type of Object that relays understand, and then if you subscribe to a prefix of a track, you get a subscription to all ANNOUNCEs for that namespace.

Then the only real mechanism one has is subscriptions, which is sort of nice.  And you could ask for only new announces if you wanted, or to resume updates from some point?

I guess the negative here is you'd want a way to not retrieve any announces that were obsolete?

draft-ietf-moq-transport.md
ianswett: Also, if I could associate an ANNOUNCE with a SUBSCRIBE_NAMESPACE, I could omit the prefix portion, which might save quite a few bytes on the wire.

draft-ietf-moq-transport.md
afrind: CC: @kixelated who had a similar idea -- there's a special track that contains the ANNOUNCE messages as objects.

You could sort of mitigate the negative by marking cancelled announce objects as "Not Exists".  

You'd need to have a reserved name for such a track, which is kind of strange.  Should we pivot to that direction or folks like it as is?

draft-ietf-moq-transport.md
afrind: > If I do SUBSCRIBE_NAMESPACE to "foo" and "foo/bar", do I get announces for "foo/bar" once or twice?

I think twice?  Maybe we should disallow overlapping SUBSCRIBE_NAMESPACE for this reason?  We can say that if you receive a SUBSCRIBE_NAMESPACE for (foo) when you already had (foo, bar), you cancel the (foo, bar) one.

afrind: > Also, if I could associate an ANNOUNCE with a SUBSCRIBE_NAMESPACE, I could omit the prefix portion, which might save quite a few bytes on the wire.

That would have to be optional because the protocol allows for blind ANNOUNCE messages.  But I agree it's a nice optimization, and would allow for de-duplicating if we allowed overlapping SUBSCRIBE_NAMESPACE prefixes.

draft-ietf-moq-transport.md
```diff
@@ -280,6 +280,12 @@ x (f):
   protocol error and SHOULD terminate the session with a Protocol
   Violation ({{session-termination}}).
 
+x (t):
+
+: Indicates that x is a tuple, consisting of a variable length integer encoded as
+  described in ({{?RFC9000, Section 16}}), followed by that many variable length
+  tuple fields, each of which are encoded as (b) above.
```
fluffy, line 287: Tiny nit but I don't love this syntax for this. It would be hard for someone reading this out of the blue to figure this out. But agree with the encoding , just wish there was less confusing syntax to do this.

draft-ietf-moq-transport.md
fluffy: I think we should go with if you have token for a more general space, it covers all the more specific spaces. Anything else just get too slow to do all the ACL checks.  So if you have a token for a/b/* is covers a/b/c

fluffy (approved):
Like the direction of this and it makes a bunch of things much easier to do.

fluffy:
It would be good to also bring in the changes proposed in #508

draft-ietf-moq-transport.md
afrind: Are we ok for now but just including an AUTH blob in the request and punting the details?  Or do we need something specific here?

afrind:
> It would be good to also bring in the changes proposed in #508

I will try to take a stab at that but I'm worried the scope is going to blow up this PR, and we'd do better to handle that in a follow up.  I mentioned this briefly in the interim, but this PR adds significant functionality that is currently not possible so I am biased towards an iterative approach.

ianswett:
To be clear, I'm happy to land this and then evaluate other improvements like #508 subsequently.

draft-ietf-moq-transport.md
ianswett: This is honestly one of the things I never understood about ANNOUNCE.  When and how does one decide to send you a blind ANNOUNCE message?  If you want all the ANNOUNCE messages you could now use SUBSCRIBE_NAMESPACE with no prefix, correct?

suhasHere:
just 2 cents, #508  addresses important design aspects which i feel would be good to consider bringing in this PR as it applies not just to announces but subscribes in general. It would be good to evaulate this PR in the context of #508

draft-ietf-moq-transport.md
fluffy: I think we should just leave all auth out of this PR for now and deal with it when we fix auth across the whole protocol.

draft-ietf-moq-transport.md
ianswett: I agree that this is non-obvious.  Should we burn some characters and make it "x (tup)" or "x (tuple)"?

afrind: I'll try it with x (tuple) and see what folks think

draft-ietf-moq-transport.md
afrind: That sounds good.  I'm keeping the update here that indicates this track parameter can be used for SUBSCRIBE_NAMESPACE, which will be a good placeholder when we come to the auth conversation.

draft-ietf-moq-transport.md
afrind: > If you want all the ANNOUNCE messages you could now use SUBSCRIBE_NAMESPACE with no prefix, correct?

Correct.

> When and how does one decide to send you a blind ANNOUNCE message?

The only place where moqt requires ANNOUNCE is when the publisher is connected to a generic relay.  There are other uses but I think they will be described in higher level protocols?

draft-ietf-moq-transport.md
fluffy: yah, I like that x(tuple) proposal. ( and of course I don't care deeply if we keep it like it is, either way works for me )

fluffy:
I hope we can get agenda time to talk about this one at the next call. This add critical functionally to moq that I would really like to have included even if we still needed time and implementation experience to get all the details sorted.

draft-ietf-moq-transport.md
ianswett: Given this design, I think saying you can't have two overlapping SUBSCRIBE_NAMESPACEs works and I can't see any case it causes problems.

draft-ietf-moq-transport.md
ianswett: I like it, but I'd like to move forward with this first, because it unlocks some other follow-on changes.

draft-ietf-moq-transport.md
afrind: I added a section below stating overlap is an error

afrind (commented):
I believe all comments have been addressed, except "should we make this a special track instead".

draft-ietf-moq-transport.md
```diff
@@ -1368,6 +1368,12 @@ the publisher will forward any matching ANNOUNCE messages to the subscriber
 that it has not yet sent.  If the set of matching ANNOUNCE messages changes, the
 publisher sends the corresponding ANNOUNCE or UNANNOUNCE message.
 
+A subscriber cannot have overlapping namespace subscriptions outstanding at the
+same time.  If a publisher receives a SUBSCRIBE_NAMESPACE with a Track Namespace
+Prefix that is a prefix of an earlier SUBSCRIBE_NAMESPACE request or vice versa,
+it MUST respond with SUBSCRIBE_ERROR, with error code
+SUBSCRIBE_NAMESPACE_OVERLAP.
```
suhasHere, line 1375: I am not sure if this should be treated as an error. A relay can receive 2 different namespaces from 2 different subscribers that overlap.

afrind: I intended that it's an error for the same subscriber to make overlapping subscribes in the same session.  A relay should be fine to receive overlapping subscribes from different sessions.  This PR doesn't define any normative behavior for forwarding SUBSCRIBE_NAMESPACE, but this clause would enforce coalescing overlaps on a single session (via UNSUBSCRIBE_NAMESPACE).

draft-ietf-moq-transport.md
```diff
@@ -378,6 +384,20 @@ In MOQT, every track has a track name and a track namespace associated
 with it.  A track name identifies an individual track within the
 namespace.
 
+Track namespace is an ordered N-tuple of bytes where N can be between 1 and 32.
+The structured nature of Track Namespace allows relays and applications to
+manipulate prefixes of a namespace. Track name is a sequence of bytes.
```
ianswett, line 389:
```suggestion
match on prefixes of a namespace. Track name is a sequence of bytes.
```

Issue #83, opened 2023-01-31
Improve startup latency


kixelated:
Overview of the current draft.

Ingest
```
--> client CONNECT   url=auth
--> client SETUP     role=producer
<-- server CONNECT   status=200
<-- server SETUP     role=consumer
--> client CATALOG   broadcast=superbowl track=a track=b track=c
<-- server SUBSCRIBE broadcast=superbowl track=b
--> client OBJECT    broadcast=superbowl track=b
```

Distribution
```
--> client CONNECT   url=auth
--> client SETUP     role=consumer
<-- server CONNECT   status=200
<-- server SETUP     role=producer
<-- server CATALOG   broadcast=superbowl track=a track=b track=c
--> client SUBSCRIBE broadcast=superbowl track=b
<-- server OBJECT    broadcast=superbowl track=b
```

If we can avoid blocking on the SETUP, then that can reduce a round trip in the ingest case. If we can push OBJECTs before the SUBSCRIBE, especially if there's only one track, then that will reduce a round trip too.

fluffy:
The question to ask is what information can you move out of the CONNECT and SETUP so that the they can just be overlapped with the messages below. As far as I can tell, all of info 

Ingest 0 RTT
--> client PUBLISH url=auth role=producer broadcast=superbowl track=a track=b track=c
--> client OBJECT 

Distribution 2 RTT 
--> client SUBSCRIBE catalog role=consumer url=auth
<-- server CATALOG   broadcast=superbowl track=a track=b track=c
--> client SUBSCRIBE broadcast=superbowl track=b role=consumer url=auth
<-- server OBJECT

suhasHere:
This seems to be fine suggestion. There are some nuances, but we are at the point where we can start thinking about reducing the round trips

ianswett:
I don't see anything in the existing docs that prevent pipelining, assuming one knows what to SUBSCRIBE to prior to connecting.  Clearly adding a catalog request adds at least one round trip.

Should we add editorial text that makes it clear that when prior information is available, the goal is to enable 0-RTT Subscriptions?

kixelated:
We can pipeline now that there's a single control stream. CONNECT/SETUP/SUBSCRIBE cooould be sent in the first flight in theory. However, stuff like negotiating versions/extensions in SETUP will force an RTT unless SUBSCRIBE is the same across all offered versions/extensions.

[...]

PR #499, opened 2024-08-05
Always use ALPN for version negotiation

### LINKED ISSUE #242: Client and/or server initiated handshake
Splitting discussion in #212 into a separate issue.

The current draft states that the client initiates the session via a CLIENT_SETUP message and then server replies with a SERVER_SETUP. The client sends a list of supported versions and its role, while the server chooses a version and reaffirms the role.

However, depending on the underlying transport, the client won't always be the first endpoint allowed to send a flight, so this incurs a half-RTT of latency:

* WebTransport: Client is allowed to send first (before CONNECT response), but most implementations will let the server send first (after 2xx CONNECT response)
* QUIC: Client sends first with 0-RTT, while the server sends first with 1-RTT (most common).

### LINKED ISSUE #243: Separate handshakes, or unified handshake?
A few issues have cropped up because we support QUIC and WebTransport with a unified SETUP handshake.

- #232 was closed because there's a desire to use ALPN for native QUIC, however no such functionality exists in WebTransport.
- #135 added a path to the SETUP message, mirroring WebTransport because no such functionality exists in QUIC.
- #242 incurs an additional half RTT delay with native QUIC because the client sends the SETUP message first.

Do we separate the QUIC and WebTransport handshakes, or continue to combine them into the same mechanism?

ianswett:
Removes version negotiation from the MoQ SETUP messages.

Fixes #242 
Fixes #243
Related to #83

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -375,6 +375,14 @@ There is no definition of the protocol over other transports,
 such as TCP, and applications using MoQ might need to fallback to
 another protocol when QUIC or WebTransport aren't available.
 
+MoQ uses ALPN in QUIC and "WT-Available-Protocols" in WebTransport
+to negotiate the version of MoQ.
+
+The ALPN value {{!RFC7301}} for the final version of this specification
+is `moq`.  ALPNs used to identify IETF drafts are created by appending
+the draft number to "moq-". For example, draft-ietf-moq-transport-13
+would be identified as "moq-13".
+
 ### WebTransport
 
 A MOQT server that is accessible via WebTransport can be identified
@@ -405,25 +413,23 @@ given URI by setting up a QUIC connection to the host and port
 identified by the `authority` section of the URI.  The `path-abempty`
 and `query` portions of the URI are communicated to the server using the
 PATH parameter ({{path}}) which is sent in the CLIENT_SETUP message at the
-start of the session.  The ALPN value {{!RFC7301}} used by the protocol
-is `moq-00`.
+start of the session.
 
-## Version and Extension Negotiation {#version-negotiation}
+## Extension Negotiation {#version-negotiation}
 
-Endpoints use the exchange of Setup messages to negotiate the MOQT version and
-any extensions to use.
+Endpoints use the exchange of Setup messages to negotiate any MOQT extensions
+to use.
 
-The client indicates the MOQT versions it supports in the CLIENT_SETUP message
-(see {{message-setup}}). It also includes the union of all Setup Parameters
-{{setup-params}} required for a handshake by any of those versions.
+The client includes all Setup Parameters {{setup-params}} required for the
+negotiated MoQ version in CLIENT_SETUP.
 
 Within any MOQT version, clients request the use of extensions by adding Setup
 parameters corresponding to that extension. No extensions are defined in this
 document.
 
-The server replies with a SERVER_SETUP message that indicates the chosen
-version, includes all parameters required for a handshake in that version, and
-parameters for every extension requested by the client that it supports.
+The server replies with a SERVER_SETUP message that includes all parameters
+required for a handshake in that version, and parameters for every extension
+requested by the client that it supports.
 
 New versions of MOQT MUST specify which existing extensions can be used with
 that version. New extensions MUST specify the existing versions with which they
@@ -894,54 +900,30 @@ handles a subscription that includes those Objects re-requests them.
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
 
 The `CLIENT_SETUP` and `SERVER_SETUP` messages are the first messages exchanged
-by the client and the server; they allows the peers to establish the mutually
-supported version and agree on the initial configuration before any objects are
-exchanged. It is a sequence of key-value pairs called Setup parameters; the
-semantics and format of which can vary based on whether the client or server is
-sending.  To ensure future extensibility of MOQT, the peers MUST ignore unknown
-setup parameters. TODO: describe GREASE for those.
+by the client and the server; they allow the peers to agree on the initial
+configuration before any objects are exchanged. It is a sequence of key-value
+pairs called Setup parameters; the semantics and format of which can vary based
+on whether the client or server is sending.  To ensure future extensibility of
+MOQT, the peers MUST ignore unknown setup parameters.
+
+TODO: describe GREASE for those.
 
 The wire format of the Setup messages are as follows:
 
 ~~~
 CLIENT_SETUP Message Payload {
-  Number of Supported Versions (i),
-  Supported Version (i) ...,
   Number of Parameters (i) ...,
   Setup Parameters (..) ...,
 }
 
 SERVER_SETUP Message Payload {
-  Selected Version (i),
   Number of Parameters (i) ...,
   Setup Parameters (..) ...,
 }
 ~~~
 {: #moq-transport-setup-format title="MOQT Setup Messages"}
 
-The available versions and Setup parameters are detailed in the next sections.
-
-### Versions {#setup-versions}
-
-MoQ Transport versions are a 32-bit unsigned integer, encoded as a varint.
-This version of the specification is identified by the number 0x00000001.
-Versions with the most significant 16 bits of the version number cleared are
-reserved for use in future IETF consensus documents.
-
-The client offers the list of the protocol versions it supports; the
-server MUST reply with one of the versions offered by the client. If the
-server does not support any of the versions offered by the client, or
-the client receives a server version that it did not offer, the
-corresponding peer MUST close the session.
-
-\[\[RFC editor: please remove the remainder of this section before
-publication.]]
-
-The version number for the final version of this specification (0x00000001), is
-reserved for the version of the protocol that is published as an RFC.
-Version numbers used to identify IETF drafts are created by adding the draft
-number to 0xff000000. For example, draft-ietf-moq-transport-13 would be
-identified as 0xff00000D.
+The available Setup parameters are detailed in the next sections.
 
 ### Setup Parameters {#setup-params}
 
@@ -1780,7 +1762,7 @@ reaching a resource limit.
 
 TODO: fill out currently missing registries:
 
-* MOQT version numbers
+* MOQT ALPN values
 * Setup parameters
 * Subscribe parameters
 * Subscribe Error codes


draft-ietf-moq-transport.md
```diff
@@ -375,6 +375,14 @@ There is no definition of the protocol over other transports,
 such as TCP, and applications using MoQ might need to fallback to
 another protocol when QUIC or WebTransport aren't available.
 
+MoQ uses ALPN in QUIC and "WT-Available-Protocols" in WebTransport
```
suhasHere, line 378: MoQT

suhasHere (approved):

ianswett:
The group thinks this is a good direction, but we want to wait until Chrome supports this to merge this PR.

draft-ietf-moq-transport.md
```diff
@@ -894,54 +900,30 @@ handles a subscription that includes those Objects re-requests them.
...
-supported version and agree on the initial configuration before any objects are
-exchanged. It is a sequence of key-value pairs called Setup parameters; the
-semantics and format of which can vary based on whether the client or server is
-sending.  To ensure future extensibility of MOQT, the peers MUST ignore unknown
-setup parameters. TODO: describe GREASE for those.
+by the client and the server; they allow the peers to agree on the initial
+configuration before any objects are exchanged. It is a sequence of key-value
+pairs called Setup parameters; the semantics and format of which can vary based
+on whether the client or server is sending.  To ensure future extensibility of
+MOQT, the peers MUST ignore unknown setup parameters.
```
fluffy, line 907: This is one part I feel pretty strongly about. 

I think we need a way of naming keys or adding a flag to them that indicates that they are mandatory to understand and if the peer does not understand them it MUST end the connection with an error.  It does not cost of anything to makes it far easier to roll out new features across clients that don't all upgrade to the same implementation every 6 weeks.

fluffy (changes requested):
What is here looks good but I would like to see one small thing added.

draft-ietf-moq-transport.md
afrind: This text isn't new in this PR, the paragraph reflowed when `to establish the mutually
supported version and` was removed.  Can you open a separate issue or PR about MTI setup params?

fluffy (approved):

LPardue (changes requested):
The general direction of the PR is fine. However, the use of ALPN is imprecise and ideally should use more precise terminology. 

Appreciate the PR is very old, and maybe the statue of limitation had passed. Maybe the right thing to do is merge as is and take an action to followup my ideals. I trust editors to make the call here but wanted to flag it .

Sy0307:
+1 I hope this feature can be integrated into draft-11 as soon as possible, as it would be very beneficial for version negotiation.

Sy0307:
What's the status of this PR? Will it be included in MOQ draft-11?

TimEvens (commented):
Seems like we are addressing a version nightmare with draft changes when hopefully the RFC will stabilize and not have as much churn.  

The challenge I see with using QUIC (`ALPN`) or WT (`WT-Available-Protocols`) is that those are connection specific outside of the encoded iMoQT payloads.  If we were to capture and replay later a MoQT session, the version will be lost unless we also capture the out-of-band transport level version information of the protocol. Kind of a bit messy IMHO.

afrind:
Parked pending conflict resolution and update

ianswett (commented):
I updated this PR and resolved any conflicts.  PTAL

draft-ietf-moq-transport.md
```diff
@@ -618,6 +618,14 @@ There is no definition of the protocol over other transports,
 such as TCP, and applications using MoQ might need to fallback to
 another protocol when QUIC or WebTransport aren't available.
 
+MoQT uses ALPN in QUIC and "WT-Available-Protocols" in WebTransport
+({{WebTransport, Section 3.4}}) to negotiate the version of MoQ.
```
```suggestion
MOQT uses ALPN in QUIC and "WT-Available-Protocols" in WebTransport
({{WebTransport, Section 3.4}}) to perform version negotiation.
```

draft-ietf-moq-transport.md
```diff
@@ -618,6 +618,14 @@ There is no definition of the protocol over other transports,
...
+
+The ALPN value {{!RFC7301}} for the final version of this specification
+is `moq`.  ALPNs used to identify IETF drafts are created by appending
```
afrind: Should we use moqt?

draft-ietf-moq-transport.md
```diff
@@ -659,22 +667,21 @@ network hosts through which a track may be accessed. The syntax of the
...
-Endpoints use the exchange of Setup messages to negotiate the MOQT version and
-any extensions to use.
+Endpoints use the exchange of Setup messages to negotiate any MOQT extensions
+to use.
 
-The client indicates the MOQT versions it supports in the CLIENT_SETUP message
-(see {{message-setup}}). It also includes the union of all Setup Parameters
-(see {{setup-params}}) required for a handshake by any of those versions.
+The client includes all Setup Parameters {{setup-params}} required for the
+negotiated MoQ version in CLIENT_SETUP.
```
```suggestion
negotiated MOQT version in CLIENT_SETUP.
```

draft-ietf-moq-transport.md
```diff
@@ -659,22 +667,21 @@ network hosts through which a track may be accessed. The syntax of the
...
 
 Within any MOQT version, clients request the use of extensions by adding Setup
 parameters corresponding to that extension. No extensions are defined in this
 document.
 
-The server replies with a SERVER_SETUP message that indicates the chosen
-version, includes all parameters required for a handshake in that version, and
-parameters for every extension requested by the client that it supports.
+The server replies with a SERVER_SETUP message that includes all parameters
+required for a handshake in that version, and parameters for every extension
```
afrind, line 684: Is this bit about "handshake in that version" correct?

draft-ietf-moq-transport.md
```diff
@@ -1763,58 +1770,33 @@ does not expire or expires at an unknown time.
...
-supported version and agree on the initial configuration before any objects are
-exchanged. It is a sequence of key-value pairs called Setup parameters; the
-semantics and format of which can vary based on whether the client or server is
-sending.  To ensure future extensibility of MOQT, endpoints MUST ignore unknown
-setup parameters. TODO: describe GREASE for those.
+by the client and the server; they allow the peers to agree on the initial
+configuration before any objects are exchanged. It is a sequence of key-value
+pairs called Setup parameters; the semantics and format of which can vary based
+on whether the client or server is sending.  To ensure future extensibility of
+MOQT, the peers MUST ignore unknown setup parameters.
```
afrind: I liked endpoint a little better here?

draft-ietf-moq-transport.md
```diff
@@ -618,6 +618,14 @@ There is no definition of the protocol over other transports,
 such as TCP, and applications using MoQ might need to fallback to
 another protocol when QUIC or WebTransport aren't available.
 
+MoQT uses ALPN in QUIC and "WT-Available-Protocols" in WebTransport
+({{WebTransport, Section 3.4}}) to negotiate the version of MoQ.
+
```
afrind, line 624: Do you need something like this:

\[\[RFC editor: please remove the remainder of this section before publication.]]

ianswett:
```suggestion
[[RFC editor: please remove the remainder of this section before publication.]]

```

draft-ietf-moq-transport.md
ianswett: I have no strong opinion on that one.

draft-ietf-moq-transport.md
ianswett: I think so?

suhasHere (approved):

draft-ietf-moq-transport.md
ianswett: Good point, I redid this paragraph to be more closely based on the newer text.

afrind:
Should we mention that this only works for moq-15+?  For moqt versions 0 - 14 require VN in CLIENT/SERVER setup.

ianswett:
> Should we mention that this only works for moq-15+? For moqt versions 0 - 14 require VN in CLIENT/SERVER setup.

Sure, any suggested text?

lminiero:
> Should we mention that this only works for moq-15+? For moqt versions 0 - 14 require VN in CLIENT/SERVER setup.

Not sure if we should also mention how the ALPN or WebTransport protocol change as well, since in drafts <= 14 it was `moq-00`, and so that could cause interop issues initially (I'm modifying my own stack to support both flavours).

draft-ietf-moq-transport.md
afrind: Now that VN is alpn, is CLIENT/SERVER setup still a "handshake" or a capabilities / setting declaration, like HTTP?

draft-ietf-moq-transport.md
```diff
@@ -618,6 +618,15 @@ There is no definition of the protocol over other transports,
...
 
+MOQT uses ALPN in QUIC and "WT-Available-Protocols" in WebTransport
+({{WebTransport, Section 3.4}}) to perform version negotiation.
+[[RFC editor: please remove the remainder of this section before publication.]]
+
+The ALPN value {{!RFC7301}} for the final version of this specification
+is `moq`.  ALPNs used to identify IETF drafts are created by appending
+the draft number to "moq-". For example, draft-ietf-moq-transport-13
+would be identified as "moq-13".
+
```
afrind, line 629:
```suggestion

Note: Draft versions prior to -15 all used moq-00 ALPN, followed by version
negotiation in the CLIENT_SETUP and SERVER_SETUP messages.
```

Sy0307:
> Should we mention that this only works for moq-15+? For moqt versions 0 - 14 require VN in CLIENT/SERVER setup.

I suggest using moq-xx for older versions, like moq-11 for moq draft-11. Or is it possible to not make a hard guarantee here?

afrind:
> I suggest using moq-xx for older versions, like moq-11 for moq draft-11. Or is it possible to not make a hard guarantee here?

@Sy0307 I think you may be the only one that supported or expected that - I don't think it will break if we say that here.  What do you do with Version fields in CLIENT_SETUP/SERVER_SETUP when negotiating via ALPN?

lminiero:
> @Sy0307 I think you may be the only one that supported or expected that - I don't think it will break if we say that here. What do you do with Version fields in CLIENT_SETUP/SERVER_SETUP when negotiating via ALPN?

Actually I just modified my stack to do the same. What happens is that a negotiation on ALPN or WebTransport Protocol happens first: if that negotiates anything other than `moq-00`, then the stack constrains the original list to only the version that was negotiated there, which means `CLIENT_SETUP` or `SERVER_SETUP` only contains that version instead of the possibly many it would have negotiated before.

Sy0307:
> > I suggest using moq-xx for older versions, like moq-11 for moq draft-11. Or is it possible to not make a hard guarantee here?
> 
> @Sy0307 I think you may be the only one that supported or expected that - I don't think it will break if we say that here. What do you do with Version fields in CLIENT_SETUP/SERVER_SETUP when negotiating via ALPN?

My suggestion is based on the incompatibility between moq-draft-09 and subsequent versions, because here we have to maintain two different CLIENT_SETUP codepoints to ensure compatibility between the two. My current approach is to set different ALPN and corresponding application layer callbacks for the two versions. When the server detects that the client is currently on moq-10 or above, I will use the new setup codepoint and the new message type callback (note that the structure of some message types has changed to some extent)

Sy0307:
Or we can set it here, developers have the right to choose whether to set the previous version to moq-xx or moq-00, that is, we no longer force the past version to be moq-00 or moq-xx, is this acceptable?

draft-ietf-moq-transport.md
ianswett: Changed to moqt

ianswett:
> Or we can set it here, developers have the right to choose whether to set the previous version to moq-xx or moq-00, that is, we no longer force the past version to be moq-00 or moq-xx, is this acceptable?

I think that's acceptable.  This will get removed from the draft eventually.  If you'd like to write a PR to change this to moq-xx, feel free, but I'd like to get this PR in now, since it's been outstanding for so long.

Issue #433, opened 2024-04-17
Need MoQT Object Header Extensibility 


fluffy:
The whole draft needed a pass at some point for extensibility but a more immediate concern is to play with various things, we need a way of adding header to an object that the relay can read.  The LOC draft has a line about needing this, I think Will had some use cases that needed it, and it would help with experimenting and migrating some pre standard implementations to the current version of the draft. 

The requirement would be there was a way to define new headers in extensions, there was a way to indicate if the header needs to be understood by relays, reviewers, both, or neither.

kixelated:
IMO we should negotiate extensions in the SETUP message. I'm a little worried about adding arbitrary blobs to the OBJECT header that _might_ change the relay's behavior, although I understand that's how HTTP works today.

fluffy:
+1 on both parts of what you said. Let me give a more concrete example. 

Many systems end up traces defined roughly like  https://opentelemetry.io/docs/concepts/signals/traces/ - to debug things across clients, servers, and a distributed system.  To make this work for MoQT, we need to be able to add a trace id that is sent across the network. In many cases not every object would be traced as that would have a performance impact, but perhaps the first object in all the I frames might be traced from a client that was having issues.  

I was imaging something like a spec could define  an extension and get a IANA number for it ( small numbers for RFC, larger numbers for non IETF specs ). You would negotiate support for the extension in usual way in the SETUP.  Then in the  object there would be an array of extensions that could be zero size and each the data for each extension included the number that identified the extension. That way someone could right an extension for something like adding a trace_id to object and relays that supported that could log information about it that included the trace-id. 

We probably need a way for extensions to mandatory to understand or not by relays and if not mandatory to understand include a length or way to be able to skip over the data.

wilaw:
Thanks Cullen for raising this issue. 

>  I think Will had some use cases that needed it

Indeed I do. I was thinking about the usefulness of headers in HTTP and how much innovation, behavior and extensibility in workflows is enabled by their presence and relatively simple implementation. I'd like something similar for moqt. 

> there was a way to indicate if the header needs to be understood by relays, reviewers, both, or neither.

IMO, if a header needs to be understood by relays, then it should be part of the moqt specification and not part of some extensibility scheme. Therefore I'd say that headers are NEVER mandatory to be understood by relays. Headers, if present, are to be understood only by those nodes that know to look for them. All other nodes would ignore them. 

[...]

Issue #416, opened 2024-03-18
Extensibility and greasing


LPardue:
At IETF 119 we discussed very briefly extensibility of the protocol. It was suggested by Ian that having _some_ extension to test the extension mechanisms/points would be useful before publication of the draft. Greasing (whether as part of MoQT or a separate extension) might be a sufficient to exercise things.

fluffy:
An extension for a relay to repot metrics about a namespace to an authorized subscriber would be very nice to have.

ianswett:
Parking until we can discuss extensibility more, #433

wilaw:
In Extensibility PR #502 I added a simple greasing scheme for extensions.

> Header values: even types are followed by a single varint encoded value. Odd types are followed by a varint encoded length and then the header value. Every integer multiple of 17 or 18 is a reserved greasing value. These types MUST be treated as unsupported by a relay. Header types are registered in the IANA table 'MOQ Extension Headers'. See {{iana}}.

LPardue:
I'm just again reading that Control Messages are not flexibly extensibile https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#section-7

> An endpoint that receives an unknown message type MUST close the session. 

Is this a consious choice or an oversight? What model does MoQT want for allowing new control message types (new version, negotiation, etc.? 

The [QUIC transport frame](https://datatracker.ietf.org/doc/html/rfc9000#section-19.21) model is that all frames need to be known, because

> QUIC frames do not use a self-describing encoding. An endpoint therefore needs to understand the syntax of all frames before it can successfully process a packet. This allows for efficient encoding of frames, but it means that an endpoint cannot send a frame of a type that is unknown to its peer.

That doesn't seem to be the case for Control Messages, since they include a length field that allows for skipping over unknown fields to the next frame.

afrind:
@LPardue : the history is that originally the control messages *didn't* have lengths either, to be more QUIC-ish, but then we realized that QUIC had the advantage of always receiving messages in packet-length quanta.  Since moqt control messages come on a stream, the sender can force the receiver into trial parsing #474.  So we added the length.

So yes the extensibility model now is that you need to negotiate a new version to send a new control message and I think that was intentional.


As an individual I'm not sure, say the HTTP/2 unknown frame thing really works that well, but I'm open to persuasion.

afrind:
[...]

PR #502, opened 2024-08-06
Adding an Extensible Object Header

wilaw:
This PR adds an extensible header to Objects, as well as updating the SETUP parameters to allow negotiation of required extension header parameters. Extensible headers have types which are IANA registered. A new Extensible header is defined and reserved with type 0x0, allowing name/value data to be passed in a manner compatible with HTTP headers.  A new example is added showing serialization of an Object with an extensible header. 

Fixes: #433

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2337,7 +2337,8 @@ that information is not reliable and cacheable.
 
 #### Object Extension Header {#object-extensions}
 Object Extension Headers are visible to relays and allow the transmission of
-future metadata relevant to MOQT Object distribution. Any Object metadata never accessed by the transport or relays SHOULD be serialized as part of the Object
+future metadata relevant to MOQT Object distribution. Any Object metadata never
+accessed by the transport or relays SHOULD be serialized as part of the Object
 payload and not as an extension header.
 
 Extension Headers are defined in external specifications and registered in an
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -984,13 +984,20 @@ client MUST set the PATH parameter to the `path-abempty` portion of the
 URI; if `query` is present, the client MUST concatenate `?`, followed by
 the `query` portion of the URI to the parameter.
 
-#### REQUIRED-EXTENSION parameter {#required-extensions}
+#### REQUESTED-EXTENSION parameter {#requested-extensions}
 
-The REQUIRED-EXTENSION parameter (key 0x02) allows the client to specify
-multiple Extension Header types {{object-extensions}} which are required for
-operation. The value is a concatenation of varints, each describing a
-32-bit extension header type. This parameter is optional. If the server does
-not support a requested REQUIRED-EXTENSION, then it MUST close the connection.
+The REQUESTED-EXTENSION parameter (key 0x02) allows the client to request
+the server to acknowledge support for  multiple Extension Header types
+{{object-extensions}} which are required for operation. The value is a
+concatenation of varints, each describing a 32-bit extension header type.
+This parameter is optional. If this parameter is present in the
+CLIENT_SETUP message, then the server MUST respond with a
+REQUESTED-EXTENSION parameter in its SERVER_SETUP message. This parameter
+MUST list the subset of those extensions previously requested by the client
+which the server supports. If the server does not support any of the
+requested extensions, then it MUST respond with a parameter value length of 0.
+The client can then choose to continue or disconnect the session, at its
+discretion.
 
 ## GOAWAY {#message-goaway}
 The server sends a `GOAWAY` message to initiate session migration
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1578,18 +1578,13 @@ are sent on a new stream. This is to avoid the status message being lost
 in cases such as a relay dropping a group and reseting the stream the
 group is being sent on.
 
-#### Object Extension {#object-extensions}
-An Object Extension is a concatenation of optional Extension Headers. These
-headers are visible to relays. Extension headers MUST be forwarded and
-MUST NOT be modified by relays. The purpose of Extension Headers is to
-facilitate the future evolution of the transport protocol. Object
-Extensions are serialized as defined below:
+#### Object Extension Header {#object-extensions}
+Object Extension Headers are visible to relays. Extension Headers MUST be
+forwarded and MUST NOT be modified by relays. The purpose of Extension
+Headers is to facilitate the future evolution of the transport protocol.
+Object Extension Headers are serialized as defined below:
 
 ~~~
-Object Extension {
-  Extension Header (..) ...
-}
-
 Extension Header {
   Header Type (i),
   [Header Value (i)]
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -486,26 +486,16 @@ PATH parameter ({{path}}) which is sent in the CLIENT_SETUP message at the
 start of the session.  The ALPN value {{!RFC7301}} used by the protocol
 is `moq-00`.
 
-## Version and Extension Negotiation {#version-negotiation}
+## Version Negotiation {#version-negotiation}
 
-Endpoints use the exchange of Setup messages to negotiate the MOQT version and
-any extensions to use.
+Endpoints use the exchange of Setup messages to negotiate the MOQT version.
 
 The client indicates the MOQT versions it supports in the CLIENT_SETUP message
 (see {{message-setup}}). It also includes the union of all Setup Parameters
 {{setup-params}} required for a handshake by any of those versions.
 
-Within any MOQT version, clients request the use of extensions by adding Setup
-parameters corresponding to that extension. No extensions are defined in this
-document.
-
 The server replies with a SERVER_SETUP message that indicates the chosen
-version, includes all parameters required for a handshake in that version, and
-parameters for every extension requested by the client that it supports.
-
-New versions of MOQT MUST specify which existing extensions can be used with
-that version. New extensions MUST specify the existing versions with which they
-can be used.
+version, includes all parameters required for a handshake in that version.
 
 If a given parameter carries the same information in multiple versions,
 but might have different optimal values in those versions, there SHOULD be
@@ -2140,11 +2130,25 @@ Any object with a status code other than zero MUST have an empty payload.
 Though some status information could be inferred from QUIC stream state,
 that information is not reliable and cacheable.
 
-
 #### Object Extension Header {#object-extensions}
-Object Extension Headers are visible to relays. Extension Headers SHOULD be
-forwarded and SHOULD NOT be modified by relays. The purpose of Extension
-Headers is to facilitate the future evolution of the transport protocol.
+Object Extension Headers are visible to relays and allow the transmission of
+future metadata relevant to MOQT Object distribution. Extension Headers MUST NOT
+carry metadata bound to the Object payload. Such metadata should instead
+be serialized into the payload.
+
+Extension Headers are defined in external specifications and registered in an
+IANA table {{iana}}. These specifications define the type and value of the
+header, along with any rules concerning processing, modification, caching and
+forwarding. A relay which is coded to implement these rules is said to
+"support" the extension.
+
+If unsupported by the relay, Extension Headers MUST NOT be modified, MUST be
+cached as part of the Object and MUST be forwarded by relays
+
+If supported by the relay and subject to the processing rules specified in the
+definition of the extension, Extension Headers MAY be modified, MAY be cached
+and MAY be forwarded.
+
 Object Extension Headers are serialized as defined below:
 
 ~~~
@@ -2158,11 +2162,12 @@ Extension Header {
 {: #object-extension-format title="Object Extension Header Format"}
 
 * Header type: an unsigned integer, encoded as a varint, identifying the type
-  of extension and also the subsequent serialization. Header types in the range
-  0x00 - 0x20 inclusive are followed by a single varint encoded value. Header
-  types 0x21 and above are followed by a varint encoded length and then the
-  header value. Header types are registered in the IANA table 'MOQ Extension
-  Headers'. See {{iana}}.
+  of the extension and also the subsequent serialization.
+* Header values: even types are followed by a single varint encoded value. Odd
+  types are followed by a varint encoded length and then the header value.
+  Every integer multiple of 17 or 18 is a reserved greasing value. These types
+  MUST be treated as unsupported by a relay. Header types are registered in the
+  IANA table 'MOQ Extension Headers'. See {{iana}}.
 
 ## Object Datagram Message {#object-datagram}
 
@@ -2407,10 +2412,10 @@ STREAM_HEADER_GROUP {
 {
   Object ID = 0
   Extension Count  = 2
-    { Type = 5
+    { Type = 4
       Value = 2186796243
     },
-    { Type = 76
+    { Type = 77
       Length = 21
       Value = "traceID:123456"
     }
@@ -2464,7 +2469,9 @@ TODO: fill out currently missing registries:
 * MOQ Extension headers - we wish to reserve extension types 0-127 for
   standards utilization where space is a premium, 128 - 16383 for
   standards utilization where space is less of a concern, and 16384 and
-  above for first-come-first-served non-standardization usage.
+  above for first-come-first-served non-standardization usage. Additionally,
+  every integer multiple of 17 or 18 is a reserved greasing value and MUST
+  NOT be registered.
 
 TODO: register the URI scheme and the ALPN
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2327,9 +2327,9 @@ that information is not reliable and cacheable.
 
 #### Object Extension Header {#object-extensions}
 Object Extension Headers are visible to relays and allow the transmission of
-future metadata relevant to MOQT Object distribution. Extension Headers MUST NOT
-carry metadata bound to the Object payload. Such metadata should instead
-be serialized into the payload.
+future metadata relevant to MOQT Object distribution. Any Object metadata not
+used by the transport SHOULD be serialized as part of the Object and not as
+an extension header.
 
 Extension Headers are defined in external specifications and registered in an
 IANA table {{iana}}. These specifications define the type and value of the
@@ -2338,11 +2338,11 @@ forwarding. A relay which is coded to implement these rules is said to
 "support" the extension.
 
 If unsupported by the relay, Extension Headers MUST NOT be modified, MUST be
-cached as part of the Object and MUST be forwarded by relays
+cached as part of the Object and MUST be forwarded by relays.
 
 If supported by the relay and subject to the processing rules specified in the
-definition of the extension, Extension Headers MAY be modified, MAY be cached
-and MAY be forwarded.
+definition of the extension, Extension Headers MAY be modified, added, removed,
+and/or cached by relays.
 
 Object Extension Headers are serialized as defined below:
 
@@ -2360,9 +2360,8 @@ Extension Header {
   of the extension and also the subsequent serialization.
 * Header values: even types are followed by a single varint encoded value. Odd
   types are followed by a varint encoded length and then the header value.
-  Every integer multiple of 17 or 18 is a reserved greasing value. These types
-  MUST be treated as unsupported by a relay. Header types are registered in the
-  IANA table 'MOQ Extension Headers'. See {{iana}}.
+  Header types are registered in the IANA table 'MOQ Extension Headers'.
+  See {{iana}}.
 
 ## Object Datagram {#object-datagram}
 
@@ -2670,11 +2669,9 @@ TODO: fill out currently missing registries:
 * MOQ Extension headers - we wish to reserve extension types 0-127 for
   standards utilization where space is a premium, 128 - 16383 for
   standards utilization where space is less of a concern, and 16384 and
-  above for first-come-first-served non-standardization usage. Additionally,
-  every integer multiple of 17 or 18 is a reserved greasing value and MUST
-  NOT be registered.
+  above for first-come-first-served non-standardization usage.
 
-TODO: register the URI scheme and the ALPN
+TODO: register the URI scheme and the ALPN and grease the Extension types
 
 # Contributors
 {:numbered="false"}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1304,9 +1304,11 @@ MUST be sent according to its `Object Forwarding Preference`, described below.
 * Object Status: As enumeration used to indicate missing
 objects or mark the end of a group or track. See {{object-status}} below.
 
-* Object Extension count: the number of Object Extension Headers present. See
-  {{object-extensions}} below. A value of 0 indicates that no Object Extension
-  Headers are present.
+* Object Extension Count: The number of Object Extensions present. A value of 0
+  indicates that no Object Extension Headers are present.
+  
+* Object Extensions : an optional concatenation of Object Extension Headers. See
+  {{object-extensions}} below.
 
 * Object Payload: An opaque payload intended for an End Subscriber and SHOULD
 NOT be processed by a relay. Only present when 'Object Status' is Normal (0x0).
@@ -1356,13 +1358,11 @@ group is being sent on.
 
 #### Object Extension {#object-extensions}
 An Object Extension is a concatenation of optional Extension Headers. These
-headers are visible to relays. Extension headers specified via the
-REQUIRED-EXTENSIONS parameter {{required-extensions}} in the SETUP message
-MUST be parsed by relays. Extension headers not specified via the
-REQUIRED-EXTENSIONS parameter MAY be ignored by relays. The purpose of
-Extension Headers is to allow the transmission of application-specific
-data as well as future evolution of the transport protocol. Object
-Extensions are serialized as defined below:
+headers are visible to relays. Extension headers MUST be forwarded and
+MUST NOT be modified by relays. The purpose of Extension Headers is to
+allow the transmission of application-specific data as well as future
+evolution of the transport protocol. Object Extensions are serialized as
+defined below:
 
 ~~~
 Object Extension {
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2115,7 +2115,7 @@ objects or mark the end of a group or track. See {{object-status}} below.
 * Object Extension Count: The number of Object Extensions present. A value of 0
   indicates that no Object Extension Headers are present.
 
-* Object Extensions : an optional concatenation of Object Extension Headers. See
+* Object Extensions : A sequence of Object Extension Headers. See
   {{object-extensions}} below.
 
 * Object Payload: An opaque payload intended for an End Subscriber and SHOULD
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2327,9 +2327,8 @@ that information is not reliable and cacheable.
 
 #### Object Extension Header {#object-extensions}
 Object Extension Headers are visible to relays and allow the transmission of
-future metadata relevant to MOQT Object distribution. Any Object metadata not
-used by the transport SHOULD be serialized as part of the Object and not as
-an extension header.
+future metadata relevant to MOQT Object distribution. Any Object metadata never accessed by the transport or relays SHOULD be serialized as part of the Object
+payload and not as an extension header.
 
 Extension Headers are defined in external specifications and registered in an
 IANA table {{iana}}. These specifications define the type and value of the
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1592,14 +1592,19 @@ Object Extension {
 
 Extension Header {
   Header Type (i),
-  Header Length (i),
-  Header Value (..)
+  [Header Value (i)]
+  [Header Length (i),
+   Header Value (..)]
 }
 ~~~
 {: #object-extension-format title="Object Extension Header Format"}
 
-* Header type: an unsigned integer, encoded as a varint and registered in the
-  IANA table 'MOQ Extension Headers'. See {{iana}}.
+* Header type: an unsigned integer, encoded as a varint, identifying the type
+  of extension and also the subsequent serialization. Header types in the range
+  0x00 - 0x20 inclusive are followed by a single varint encoded value. Header
+  types 0x21 and above are followed by a varint encoded length and then the
+  header value. Header types are registered in the IANA table 'MOQ Extension
+  Headers'. See {{iana}}.
 
 ### Object Message Formats
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2675,8 +2675,8 @@ TODO: fill out currently missing registries:
 * Announce Error codes
 * Announce Cancel Reason codes
 * Message types
-* MOQ Extension headers - we wish to reserve extension types 0-127 for
-  standards utilization where space is a premium, 128 - 16383 for
+* MOQ Extension headers - we wish to reserve extension types 0-63 for
+  standards utilization where space is a premium, 64 - 16383 for
   standards utilization where space is less of a concern, and 16384 and
   above for first-come-first-served non-standardization usage.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -989,7 +989,7 @@ the `query` portion of the URI to the parameter.
 The REQUESTED-EXTENSION parameter (key 0x02) allows the client to request
 the server to acknowledge support for  multiple Extension Header types
 {{object-extensions}} which are required for operation. The value is a
-concatenation of varints, each describing a 32-bit extension header type.
+concatenation of varints, each describing an integer extension header type.
 This parameter is optional. If this parameter is present in the
 CLIENT_SETUP message, then the server MUST respond with a
 REQUESTED-EXTENSION parameter in its SERVER_SETUP message. This parameter
@@ -1599,8 +1599,8 @@ Extension Header {
 ~~~
 {: #object-extension-format title="Object Extension Header Format"}
 
-* Header type: an unsigned 32-bit integer, registered in the IANA table
-  'MOQ Extension Headers'. See {{iana}}.
+* Header type: an unsigned integer, encoded as a varint and registered in the
+  IANA table 'MOQ Extension Headers'. See {{iana}}.
 
 ##### Extension Header type 0 {#extension-header-zero}
 
@@ -1919,7 +1919,10 @@ TODO: fill out currently missing registries:
 * Subscribe Error codes
 * Announce Error codes
 * Message types
-* MOQ Extension headers
+* MOQ Extension headers - we wish to reserve extension types 0-127 for
+  standards utilization where space is a premium, 128 - 16383 for
+  standards utilization where space is less of a concern, and 16384 and
+  above for first-come-first-served non-standardization usage.
 
 TODO: register the URI scheme and the ALPN
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1370,7 +1370,7 @@ Extension Header {
 {: #object-extension-format title="Object Extension Header Format"}
 
 * Header type: an unsigned 32-bit integer, registered in the IANA table
-  'MOQ Extension Headers'. See {{#iana}}.
+  'MOQ Extension Headers'. See {{iana}}.
 
 ##### Extension Header type 0 {#extension-header-zero}
 
@@ -1397,7 +1397,6 @@ Header Value {
   and is not IANA registered.
 * Payload: the contents of the header. The combined size of the name and payload
   contents MUST NOT exceed 10240 bytes.
-  
 
 ### Object Message Formats
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1178,21 +1178,6 @@ The MAX_SUBSCRIBE_ID parameter (Parameter Type 0x02) communicates an initial
 value for the Maximum Subscribe ID to the receiving subscriber. The default
 value is 0, so if not specified, the peer MUST NOT create subscriptions.
 
-#### REQUESTED-EXTENSION parameter {#requested-extensions}
-
-The REQUESTED-EXTENSION parameter (key 0x03) allows the client to request
-the server to acknowledge support for  multiple Extension Header types
-{{object-extensions}} which are required for operation. The value is a
-sequence of varints, each describing an integer extension header type.
-This parameter is optional. If this parameter is present in the
-CLIENT_SETUP message, and if the then the server MUST respond with a
-REQUESTED-EXTENSION parameter in its SERVER_SETUP message. This parameter
-MUST list the subset of those extensions previously requested by the client
-which the server supports. If the server does not support any of the
-requested extensions, then it MUST respond with a parameter value length of 0.
-The client can then choose to continue or disconnect the session, at its
-discretion.
-
 ## GOAWAY {#message-goaway}
 
 An endpoint sends a `GOAWAY` message to inform the peer it intends to close
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -985,6 +985,15 @@ client MUST set the PATH parameter to the `path-abempty` portion of the
 URI; if `query` is present, the client MUST concatenate `?`, followed by
 the `query` portion of the URI to the parameter.
 
+#### REQUIRED-EXTENSION parameter {#required-extensions}
+
+The REQUIRED-EXTENSION parameter (key 0x02) allows the client to specify
+an Extension Header type {{object-extensions}} which is required for
+operation. The value is the 32-bit type expressed as a varint. This
+parameter is optional. Multiple of these parameters may be sent in a
+SETUP message. If the server does not support a requested
+REQUIRED-EXTENSION, then it MUST close the connection.
+
 ## GOAWAY {#message-goaway}
 The server sends a `GOAWAY` message to initiate session migration
 ({{session-migration}}) with an optional URI.
@@ -1287,6 +1296,10 @@ MUST be sent according to its `Object Forwarding Preference`, described below.
 * Object Status: As enumeration used to indicate missing
 objects or mark the end of a group or track. See {{object-status}} below.
 
+* Object Extension count: the number of Object Extension Headers present. See
+  {{object-extensions}} below. A value of 0 indicates that no Object Extension
+  Headers are present.
+
 * Object Payload: An opaque payload intended for an End Subscriber and SHOULD
 NOT be processed by a relay. Only present when 'Object Status' is Normal (0x0).
 
@@ -1333,6 +1346,59 @@ are sent on a new stream. This is to avoid the status message being lost
 in cases such as a relay dropping a group and reseting the stream the
 group is being sent on.
 
+#### Object Extension {#object-extensions}
+An Object Extension is a concatenation of optional Extension Headers. These
+headers are visible to relays. Extension headers specified via the
+REQUIRED-EXTENSIONS parameter {{required-extensions}} in the SETUP message
+MUST be parsed by relays. Extension headers not specified via the
+REQUIRED-EXTENSIONS parameter MAY be ignored by relays. The purpose of
+Extension Headers is to allow the transmission of application-specific
+data as well as future evolution of the transport protocol. Object
+Extensions are serialized as defined below:
+
+~~~
+Object Extension {
+  Extension Header (..) ...
+}
+
+Extension Header {
+  Header Type (i),
+  Header Length (i),
+  Header Value (..)
+}
+~~~
+{: #object-extension-format title="Object Extension Header Format"}
+
+* Header type: an unsigned 32-bit integer, registered in the IANA table
+  'MOQ Extension Headers'. See {{#iana}}.
+
+##### Extension Header type 0 {#extension-header-zero}
+
+This specification defines a utility extension header. The value of this header
+is itself a name-value pair.
+
+
+| Type |                         Value                        |
+| ---- | ---------------------------------------------------- |
+| 0x0  | Header Value  - see {{extension-header-zero-format}} |
+
+
+~~~
+Header Value {
+  Name Length (i),
+  Name Value (..)
+  Payload (..)
+}
+~~~
+{: #extension-header-zero-format title="Extension header 0 value format"}
+
+* Name Length: the size of the name value in bytes.
+* Name Value: a string encoded using ISO-8859-1. This name is application-defined
+  and is not IANA registered.
+* Payload: the contents of the header. The combined size of the name and payload
+  contents MUST NOT exceed 10240 bytes.
+  
+
 ### Object Message Formats
 
 Every Track has a single 'Object Forwarding Preference' and the Original
@@ -1359,6 +1425,8 @@ OBJECT_STREAM Message {
   Group ID (i),
   Object ID (i),
   Publisher Priority (8),
+  Extension Count (i),
+  [Extension headers (...)],
   Object Payload Length (i),
   [Object Status (i)],
   Object Payload (..),
@@ -1395,6 +1463,8 @@ OBJECT_DATAGRAM Message {
   Group ID (i),
   Object ID (i),
   Publisher Priority (8),
+  Extension Count (i),
+  [Extension headers (...)],
   Object Payload Length (i),
   [Object Status (i)],
   Object Payload (..),
@@ -1553,6 +1623,38 @@ OBJECT_STREAM {
 }
 ~~~
 
+Sending a group on one stream, with the first object containing an
+Extension Header.
+
+~~~
+Stream = 2
+
+STREAM_HEADER_GROUP {
+  Subscribe ID = 2
+  Track Alias = 2
+  Group ID = 0
+  Publisher Priority = 0
+}
+{
+  Object ID = 0
+  Extension Count  = 1
+    { Type = 0
+      Length = 21
+        { Name length = 15
+          Name value = "example-traceID"
+          Payload = "123456"
+        }
+    }
+  Object Payload Length = 4
+  Payload = "abcd"
+}
+{
+  Object ID = 1
+  Object Payload Length = 4
+  Payload = "efgh"
+}
+
+~~~
 
 ## SUBSCRIBE_OK {#message-subscribe-ok}
 
@@ -1786,6 +1888,7 @@ TODO: fill out currently missing registries:
 * Subscribe Error codes
 * Announce Error codes
 * Message types
+* MOQ Extension headers
 
 TODO: register the URI scheme and the ALPN
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2157,8 +2157,8 @@ that information is not reliable and cacheable.
 
 
 #### Object Extension Header {#object-extensions}
-Object Extension Headers are visible to relays. Extension Headers MUST be
-forwarded and MUST NOT be modified by relays. The purpose of Extension
+Object Extension Headers are visible to relays. Extension Headers SHOULD be
+forwarded and SHOULD NOT be modified by relays. The purpose of Extension
 Headers is to facilitate the future evolution of the transport protocol.
 Object Extension Headers are serialized as defined below:
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -485,16 +485,26 @@ PATH parameter ({{path}}) which is sent in the CLIENT_SETUP message at the
 start of the session.  The ALPN value {{!RFC7301}} used by the protocol
 is `moq-00`.
 
-## Version Negotiation {#version-negotiation}
+## Version Negotiation and Extension Negotiation {#version-negotiation}
 
-Endpoints use the exchange of Setup messages to negotiate the MOQT version.
+Endpoints use the exchange of Setup messages to negotiate the MOQT version and
+any extensions to use.
 
 The client indicates the MOQT versions it supports in the CLIENT_SETUP message
 (see {{message-setup}}). It also includes the union of all Setup Parameters
 {{setup-params}} required for a handshake by any of those versions.
 
+Within any MOQT version, clients request the use of extensions by adding Setup
+parameters corresponding to that extension. No extensions are defined in this
+document.
+
 The server replies with a SERVER_SETUP message that indicates the chosen
-version, includes all parameters required for a handshake in that version.
+version, includes all parameters required for a handshake in that version, and
+parameters for every extension requested by the client that it supports.
+
+New versions of MOQT MUST specify which existing extensions can be used with
+that version. New extensions MUST specify the existing versions with which they
+can be used.
 
 If a given parameter carries the same information in multiple versions,
 but might have different optimal values in those versions, there SHOULD be
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1521,7 +1521,7 @@ objects or mark the end of a group or track. See {{object-status}} below.
 
 * Object Extension Count: The number of Object Extensions present. A value of 0
   indicates that no Object Extension Headers are present.
-  
+
 * Object Extensions : an optional concatenation of Object Extension Headers. See
   {{object-extensions}} below.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1715,6 +1715,8 @@ The Object Status field is only sent if the Object Payload Length is zero.
 {
   Group ID (i),
   Object ID (i),
+  Extension Count (i),
+  [Extension headers (...)],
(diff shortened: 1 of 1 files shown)


afrind:
Individual Comment:

This seems generally ok.  Perhaps there's an optimization where Object encodings don't carry even the count field unless at least one extension is negotiated.

The wider problem I see with this design is that extension negotiation is hop-by-hop (HTTP has similar problems with SETTINGS).  What happens when, say, the publisher and relay negotiate use of an extension, but one of the subscribers to the relay does not?

draft-ietf-moq-transport.md
```diff
@@ -1295,6 +1304,10 @@ MUST be sent according to its `Object Forwarding Preference`, described below.
 * Object Status: As enumeration used to indicate missing
 objects or mark the end of a group or track. See {{object-status}} below.
 
+* Object Extension count: the number of Object Extension Headers present. See
+  {{object-extensions}} below. A value of 0 indicates that no Object Extension
+  Headers are present.
```
suhasHere, line 1309: I think this should allow actual extensions too not just the count. isn't it?  This will be needed especially when serving object from the cache, then those extensions needs to be sent along

draft-ietf-moq-transport.md
```diff
@@ -1341,6 +1354,58 @@ are sent on a new stream. This is to avoid the status message being lost
 in cases such as a relay dropping a group and reseting the stream the
 group is being sent on.
 
+#### Object Extension {#object-extensions}
+An Object Extension is a concatenation of optional Extension Headers. These
+headers are visible to relays. Extension headers specified via the
+REQUIRED-EXTENSIONS parameter {{required-extensions}} in the SETUP message
+MUST be parsed by relays. Extension headers not specified via the
```
suhasHere, line 1361: I understand the MUST. But I wonder if we should add a indicator in each extension to include if it needs to be forwarded as well.  These extension headers must not be modified by the relays either to be complete. Since some of the extensions are useful to downstream.

draft-ietf-moq-transport.md
```diff
@@ -1341,6 +1354,58 @@ are sent on a new stream. This is to avoid the status message being lost
...
+  Header Length (i),
+  Header Value (..)
+}
+~~~
+{: #object-extension-format title="Object Extension Header Format"}
+
+* Header type: an unsigned 32-bit integer, registered in the IANA table
+  'MOQ Extension Headers'. See {{iana}}.
+
+##### Extension Header type 0 {#extension-header-zero}
```
suhasHere, line 1383: I don't think we need this defined ? Can you please share your thoughts on why we need this ?

draft-ietf-moq-transport.md
```diff
@@ -1341,6 +1354,58 @@ are sent on a new stream. This is to avoid the status message being lost
...
+
+~~~
+Object Extension {
+  Extension Header (..) ...
+}
+
+Extension Header {
+  Header Type (i),
+  Header Length (i),
+  Header Value (..)
```
suhasHere, line 1375: Header Value should be opaque bytes.

draft-ietf-moq-transport.md
```diff
@@ -1341,6 +1354,58 @@ are sent on a new stream. This is to avoid the status message being lost
...
+}
+~~~
+{: #object-extension-format title="Object Extension Header Format"}
+
+* Header type: an unsigned 32-bit integer, registered in the IANA table
+  'MOQ Extension Headers'. See {{iana}}.
```
suhasHere, line 1381: I think we are missing iana section. 
At a high level, the iana registration must have
 - extension type, reference to rfc that defines the extenson
 -  this needs to be specification required

suhasHere (commented):
Thanks @wilaw  for this PR. I like the direction of this PR and have few clarification questions/nits.

draft-ietf-moq-transport.md
wilaw: The assumption is that these extensions headers must always be forwarded and must not be modified by any relay. I could add language stating that explicitly?

draft-ietf-moq-transport.md
wilaw: This default extension header provides a simple way to add ad-hoc name/value pairs to any object. This is similar to the ability of clients and servers to define custom request and response headers in HTTP-land. That functionality has proven extremely useful and efficient in enabling rapid development, testing and deployment of solutions without having to go to IANA registries or revise specifications. I'd like a similar capability within MOQT hence Extension Header Type Zero.

draft-ietf-moq-transport.md
wilaw: Yes, the [IANA section ](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#name-iana-considerations)of the whole MOQT document is a big todo. I added the requirement to define the MOQ extension Headers table to Section 8. I figure the editors plan to handle all the IANA stuff towards the end. It will be specifications required etc once it is defined.

draft-ietf-moq-transport.md
wilaw: Clarification - opaque to whom - the relays?  Object headers values are not opaque to relays and nor are Extension Header values. If an application wants to hide content from a relay then it needs to place it in the Object payload (which is opaque) and not the headers.

draft-ietf-moq-transport.md
wilaw: I was thinking that Extension headers are immutable. If the publisher adds them, then any relay forwarding that object must preserve them. The object would be cached along with any extension headers it was carrying.

draft-ietf-moq-transport.md
```diff
@@ -993,6 +993,15 @@ client MUST set the PATH parameter to the `path-abempty` portion of the
 URI; if `query` is present, the client MUST concatenate `?`, followed by
 the `query` portion of the URI to the parameter.
 
+#### REQUIRED-EXTENSION parameter {#required-extensions}
+
+The REQUIRED-EXTENSION parameter (key 0x02) allows the client to specify
+an Extension Header type {{object-extensions}} which is required for
+operation. The value is the 32-bit type expressed as a varint. This
+parameter is optional. Multiple of these parameters may be sent in a
```
mengelbart, line 1001: This contradicts [section 6.1](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#section-6.1), which states that there MUST NOT be duplicate parameters. I'm fine with lifting that restriction, though. If necessary, I think we could add this restriction to individual parameters that can't be duplicated.

```suggestion
parameter is optional. Multiple of these parameters MAY be sent in a
```

draft-ietf-moq-transport.md
mengelbart: I think @suhasHere meant that the actual object extensions should be included as a canonical field, not just the count.

draft-ietf-moq-transport.md
mengelbart: I think explicit language to state that extensions must be forwarded would be useful. Maybe it's enough to replace "MUST be parsed" with "MUST be forwarded"?

ianswett: I think we should say that headers MUST be forwarded and MUST NOT be modified if that's what we intend.

I'd like to have these authenticated when using MoQ per-object encryption, but that's a different draft.

suhasHere: agree with @ianswett  and @mengelbart  .. Making it explicit is good in these cases.

I see GroupId, ObjectId and Extension Header can fall in to this category. 

Secure objects uses ObjectId and GroupId to be authenticated and encrypted end to end, for example.

I am sure there will some use-cases that would have extensions whose scope is hop by hop too. I will be totally ok to have an scope field in the extension that identifies if it is between original-publisher and end-subscriber or not.

draft-ietf-moq-transport.md
suhasHere: Thanks for the context.  Still i don't see if this adds much value in end to end context.

draft-ietf-moq-transport.md
suhasHere: If a relay supports an extension, it can parse it.  We know that from setup message that if a relay supports an extension or not, isn't it.

draft-ietf-moq-transport.md
suhasHere: +1 .. makes sense. thanks

draft-ietf-moq-transport.md
wilaw: @mengelbart  - good point. We either 
1. redefine section 6.1 to indicate that no duplicate parameters are allowed except for Required-Extension, which may be duplicated. 
2. Or we change the serialization of the VALUE of the REQUIRED-EXTENSION parameter. Since each extension type is a varint, and the length of the parameter is specified, the parser can just keep reading varints until it reaches the length. In this way, multiple extension types can be communicated in a single REQUIRED-EXTENSION parameter and we can retain the existing requirement that parameters MUST NOT be duplicated. 

Preference?

draft-ietf-moq-transport.md
wilaw: I have updated the coanonical fields definitions to inlcude both fields, as shown below. 

> * Object Extension Count: The number of Object Extensions present. A value of 0
>   indicates that no Object Extension Headers are present.
>   
> * Object Extensions : an optional concatenation of Object Extension Headers. See
>   {{object-extensions}} below.

draft-ietf-moq-transport.md
wilaw: Given that we have no means, as Alan has pointed out, of propagating header requirements beyond the immediate hop, it means that if any relay wants to use the extension header then all relays must forward them and not modify them. I have therefore modified  the language to read that 

> Extension headers MUST be forwarded and MUST NOT be modified by relays.

draft-ietf-moq-transport.md
wilaw: I went with option 2. Now there is only a single Required Extension parameter key. It's value can hold a concatenation of multiple Extension Header numerical types

draft-ietf-moq-transport.md
```diff
@@ -984,6 +984,14 @@ client MUST set the PATH parameter to the `path-abempty` portion of the
 URI; if `query` is present, the client MUST concatenate `?`, followed by
 the `query` portion of the URI to the parameter.
 
+#### REQUIRED-EXTENSION parameter {#required-extensions}
+
+The REQUIRED-EXTENSION parameter (key 0x02) allows the client to specify
```
ianswett, line 989: This is a potentially painful way to negotiate extensions.  What about making it REQUESTED-EXTENSION and then the server responds with the ones it supports of the requested set?  At that point, if the client wants to close the connection because the server doesn't support a required extension, then it can.

ianswett (commented):
One comment about the negotiation approach, and I also am unsure of how to best use this, given it's negotiated hop-by-hop.

draft-ietf-moq-transport.md
```diff
@@ -984,6 +984,14 @@ client MUST set the PATH parameter to the `path-abempty` portion of the
 URI; if `query` is present, the client MUST concatenate `?`, followed by
 the `query` portion of the URI to the parameter.
 
+#### REQUIRED-EXTENSION parameter {#required-extensions}
```
fluffy, line 987: I don't think it makes sense to negotiate object extensions with the relay. This would be like negotiating HTTP header extensions with a socks proxy. Other types of extensions we need too negotiate with server but not these.   I think we can just remove this part of the PR.

draft-ietf-moq-transport.md
```diff
@@ -1511,6 +1519,12 @@ MUST be sent according to its `Object Forwarding Preference`, described below.
 * Object Status: As enumeration used to indicate missing
 objects or mark the end of a group or track. See {{object-status}} below.
 
+* Object Extension Count: The number of Object Extensions present. A value of 0
+  indicates that no Object Extension Headers are present.
+
+* Object Extensions : an optional concatenation of Object Extension Headers. See
+  {{object-extensions}} below.
```
fluffy, line 1526: I'd like to talk about ideas to bit pack the extension indictors to save space but conceptually good with this.

draft-ietf-moq-transport.md
```diff
@@ -1557,6 +1571,62 @@ are sent on a new stream. This is to avoid the status message being lost
...
+Extension Header {
+  Header Type (i),
+  Header Length (i),
+  Header Value (..)
+}
+~~~
+{: #object-extension-format title="Object Extension Header Format"}
+
+* Header type: an unsigned 32-bit integer, registered in the IANA table
+  'MOQ Extension Headers'. See {{iana}}.
```
fluffy, line 1596: I think we need to make sure that small numbers are reserved for thing where space is issue, medium size numbers are for things that are standards, and large numbers are totally first come first serve and super easy to get and use any way you want.

draft-ietf-moq-transport.md
```diff
@@ -1557,6 +1571,62 @@ are sent on a new stream. This is to avoid the status message being lost
...
+
+##### Extension Header type 0 {#extension-header-zero}
+
+This specification defines a utility extension header. The value of this header
```
fluffy, line 1600: I think this whole extension should be moved to a separate draft that helps illustrate how to define an extension and show that we got things defined in a way where a separate draft can define an extension. 

I'm not a huge fan of it, I would rather that it was easy to get a number code point for "fluffy-idea-of-perfectp-trace-id" rather than put strings in the protocol that would better in an IANA table but I agree with a low friction way of getting customer extensions.

fluffy (changes requested):
I basically like this but I think some parts that need some discution and some parts I think would be better as separate draft.

draft-ietf-moq-transport.md
```diff
@@ -984,6 +984,14 @@ client MUST set the PATH parameter to the `path-abempty` portion of the
 URI; if `query` is present, the client MUST concatenate `?`, followed by
 the `query` portion of the URI to the parameter.
 
+#### REQUIRED-EXTENSION parameter {#required-extensions}
+
+The REQUIRED-EXTENSION parameter (key 0x02) allows the client to specify
+multiple Extension Header types {{object-extensions}} which are required for
+operation. The value is a concatenation of varints, each describing a
+32-bit extension header type. This parameter is optional. If the server does
+not support a requested REQUIRED-EXTENSION, then it MUST close the connection.
```
suhasHere, line 993: I don't think closing the connection is right here. If a relay don't need to understand the extension, it just forwards it .

RichLogan (commented):
Should we also update the structures shown in the `Multi-Object Groups` section to show extension count + extensions in the track and group forwarding preference structures? 

**Track**:

https://github.com/moq-wg/moq-transport/blob/358c5d75d93b28c63a955fc2eda17bfdcb839e5e/draft-ietf-moq-transport.md?plain=1#L1735-L1741

**Group**:

https://github.com/moq-wg/moq-transport/blob/358c5d75d93b28c63a955fc2eda17bfdcb839e5e/draft-ietf-moq-transport.md?plain=1#L1777-L1782

draft-ietf-moq-transport.md
wilaw: @ianswett  - I have modified the PR with this [commit ](https://github.com/moq-wg/moq-transport/pull/502/commits/ab407f75b8954c445f6dd1efef919631f93b58d2) to modify the extension negotiation and to rename the parameter to REQUESTED_EXTENSION

> The REQUESTED-EXTENSION parameter (key 0x02) allows the client to request
> the server to acknowledge support for  multiple Extension Header types
> {{object-extensions}} which are required for operation. The value is a
> concatenation of varints, each describing a 32-bit extension header type.
> This parameter is optional. If this parameter is present in the
> CLIENT_SETUP message, then the server MUST respond with a
> REQUESTED-EXTENSION parameter in its SERVER_SETUP message. This parameter
> MUST list the subset of those extensions previously requested by the client
> which the server supports. If the server does not support any of the
> requested extensions, then it MUST respond with a parameter value length of 0.
> The client can then choose to continue or disconnect the session, at its
> discretion.

draft-ietf-moq-transport.md
wilaw: This has now changed. It is now a REQUESTED-EXTENSION parameter list from client to server with a response from the server. The client may then choose whether it wants to continue or disconnect.

wilaw: Additionally, all relays MUST forward and not modify all extensions.

draft-ietf-moq-transport.md
wilaw: This negotiation has now changed slightly. The clients uses the revised REQUESTED-EXTENSION parameter to indicate which extensions it would like and the server responds with the sub-set of those that it does support. The client may then chose to continue or disconnect. 

Since we have separate rules that relays MUST always forward all headers and MUST NOT modify them, any negotiation is practically only going to happen at the final hop. I agree that its utility is limited. I would like to remove it, but first I want agreement from those who wanted it in there in the first place.

draft-ietf-moq-transport.md
wilaw: With this [commit](https://github.com/moq-wg/moq-transport/pull/502/commits/3759564cffa9a066349de4d018aba601f9c117d3) I have added guidance language for the IANA registration of the extension header types. 

>  MOQ Extension headers - we wish to reserve extension types 0-127 for
>   standards utilization where space is a premium, 128 - 16383 for
>   standards utilization where space is less of a concern, and 16384 and
>   above for first-come-first-served non-standardization usage.

These ranges correspond to 1-bye, 2 bytes and 3 bytes and beyond with varint encoding.

draft-ietf-moq-transport.md
```diff
@@ -1557,6 +1578,62 @@ are sent on a new stream. This is to avoid the status message being lost
...
+allow the transmission of application-specific data as well as future
+evolution of the transport protocol. Object Extensions are serialized as
+defined below:
+
+~~~
+Object Extension {
+  Extension Header (..) ...
+}
+
+Extension Header {
```
fluffy, line 1594: A very common use case if an extension that caries a single integer - for example in voice applications the audio level specified in RFC6464. For some voice only applications this can be critical to keep the overall bandwidth very small. For these very special uses cases, I would like a shorter encoding so here is my proposed change

For type values above 20, it looks just like how you have it. But for type values less than that, it becomes just the type followed by  value which is a varint. Removing a byte out of every audio packet does make a real difference to some applications.

fluffy: Glad to send a PR to your PR if this sounds reasonable to you

tianji-sdo:
**A 5G use case to justify the support of the PR (Problem description & Requirements on MoQT per-object metadata header extension)**:
========
As we have presented in the IETF-121 MoQ session @ Dublin, the 3GPP 5G has standardized the solution of applying the MoQT to address the transmission of encrypted XRM media streams between MoQ server(s) and client(s) via MoQ relay. A MoQ relay is deployed in the 5G UPF. The corresponding 3GPP 5G specifications are TS 23.501 and TS 26.522. In the MoQT-based solution, there is per-object (or so-named per-PDU in 5G term) related metadata that is used to identify the PDU-set (i.e., a set of PDUs) related metadata information to 5G system. These PDU-set meta-info is identified and marked at a 5G UPF (i.e., where a MoQ replay runs) for RAN-based resource scheduling and optimization. Our IETF draft (https://datatracker.ietf.org/doc/draft-defoy-moq-relay-network-handling/)  has clearly explained the mechanism.  

Because an XRM media stream is encrypted (and so is the RTP header extension associated with an XRM PDU), the PDU-set metadata has to be mapped to the MoQT metadata field in an object header. The MoQ-based solution (as in 3GPP TS 23.501) is repurposing those object header fields, which leads to two problems: mapping accuracy and future extensibility.
 
Based on the discussions & rough consensus from the IETF-121 MoQ session, we think that the logics as specified in PR#502 , targeting at providing the OBJECT Extension Header, can address exactly the XRM ciphered stream challenges with a new revision of the MoQT draft. 

Those logics include REQUESTED-EXTENSION parameter, Object Extensions, as well as {{IANA}} part, and possibly more, in the current PR.

draft-ietf-moq-transport.md
```diff
@@ -1806,6 +1919,10 @@ TODO: fill out currently missing registries:
 * Subscribe Error codes
 * Announce Error codes
 * Message types
+* MOQ Extension headers - we wish to reserve extension types 0-127 for
+  standards utilization where space is a premium, 128 - 16383 for
+  standards utilization where space is less of a concern, and 16384 and
+  above for first-come-first-served non-standardization usage.
```
gloinul, line 1925: Should this space be greased or any particular picking algorithm?

draft-ietf-moq-transport.md
```diff
@@ -1557,6 +1578,62 @@ are sent on a new stream. This is to avoid the status message being lost
...
+}
+~~~
+{: #object-extension-format title="Object Extension Header Format"}
+
+* Header type: an unsigned integer, encoded as a varint and registered in the
+  IANA table 'MOQ Extension Headers'. See {{iana}}.
+
+##### Extension Header type 0 {#extension-header-zero}
+
+This specification defines a utility extension header. The value of this header
```
gloinul, line 1607: If this is kept in, it purpose needs to be clarified. But it feels like this is creating an additional extension space.

draft-ietf-moq-transport.md
```diff
@@ -1557,6 +1578,62 @@ are sent on a new stream. This is to avoid the status message being lost
 in cases such as a relay dropping a group and reseting the stream the
 group is being sent on.
 
+#### Object Extension {#object-extensions}
+An Object Extension is a concatenation of optional Extension Headers. These
```
gloinul, line 1582: I am bit worried by the usage of word Optional here? What exactly is optional? Looking below I get the impression that Extension header handling are depending on role of the node but in no case are they truly optional, other than it comes to the publisher including them. 

A Relay must forward and not modify
A client must be capable of receiving and ignoring them.

draft-ietf-moq-transport.md
```diff
@@ -984,6 +984,21 @@ client MUST set the PATH parameter to the `path-abempty` portion of the
 URI; if `query` is present, the client MUST concatenate `?`, followed by
 the `query` portion of the URI to the parameter.
 
+#### REQUESTED-EXTENSION parameter {#requested-extensions}
+
+The REQUESTED-EXTENSION parameter (key 0x02) allows the client to request
```
gloinul, line 989: I struggle to understand how this header is useful unless all relay's in the chain forward it upstream towards the publisher for those that are related to the media content. 

Are we envisioning both next hop extension as well as whole chain extensions?

draft-ietf-moq-transport.md
suhasHere: I am fine with this as a starting place. If we may decide to support a way to indicate HBH vs E2E extensions, then we can address how to best solve it. 

Starting with MoQ Relays forwarding extensions as-is seems fine to get some implementation experience. 

Also https://datatracker.ietf.org/doc/draft-mzanaty-moq-loc/ and https://datatracker.ietf.org/doc/html/draft-jennings-moq-secure-objects-01 defines extensions that might be more than hbh scoped for example.

draft-ietf-moq-transport.md
LPardue: +1 on greasing - let's get these reserved now and people adding them in their implementation from day one

afrind (commented):
Individual Review:

1. This needs a rebase because it doesn't include subgroups yet
2. I think we also might want a way to specify an extension per subgroup rather than per object (eg: an extension sent once at the beginning of the stream that applies to all objects that follow).

draft-ietf-moq-transport.md
wilaw: Based on your and others feedback, I have removed (via  commit https://github.com/moq-wg/moq-transport/pull/502/commits/e1764972a4dc1be7fac8e1777462c560df6d6385) the Extension header type 0 definition. It can be added back in the future if necessary.

draft-ietf-moq-transport.md
wilaw: Based on your and others feedback, I have removed (via commit https://github.com/moq-wg/moq-transport/commit/e1764972a4dc1be7fac8e1777462c560df6d6385) the Extension header type 0 definition. It can be added back in the future if necessary.

draft-ietf-moq-transport.md
wilaw: @fluffy  - this seems like a reasonable improvement.  I updated the PR with commit https://github.com/moq-wg/moq-transport/pull/502/commits/6f31642317f2fbddbc151db2f8994bf0fd10818b to reflect this improved compression scheme for extension types <=0x20

draft-ietf-moq-transport.md
wilaw: Based on your and others comments, I removed (via commit https://github.com/moq-wg/moq-transport/commit/e1764972a4dc1be7fac8e1777462c560df6d6385) the Extension header type 0 definition. It can be added back in the future if necessary.

wilaw:
@RichLogan 
> Should we also update the structures shown in the `Multi-Object Groups` section to show extension count + extensions in the track and group forwarding preference structures?

Yes. Commit https://github.com/moq-wg/moq-transport/pull/502/commits/ae714067b66e38ccc47d126ba9ea079e3976549c fixes this.

draft-ietf-moq-transport.md
```diff
@@ -984,6 +984,21 @@ client MUST set the PATH parameter to the `path-abempty` portion of the
...
+concatenation of varints, each describing an integer extension header type.
+This parameter is optional. If this parameter is present in the
+CLIENT_SETUP message, then the server MUST respond with a
+REQUESTED-EXTENSION parameter in its SERVER_SETUP message. This parameter
+MUST list the subset of those extensions previously requested by the client
+which the server supports. If the server does not support any of the
+requested extensions, then it MUST respond with a parameter value length of 0.
+The client can then choose to continue or disconnect the session, at its
+discretion.
+
```
suhasHere, line 1001: I am not too sure of the purpose of this parameter.
I am struggling understand what it means to say "server supports" .

suhasHere, line 1001: I am not too sure of the purpose of this parameter.
I am struggling understand what it means to say "server supports" .

draft-ietf-moq-transport.md
```diff
@@ -1511,6 +1526,12 @@ MUST be sent according to its `Object Forwarding Preference`, described below.
 * Object Status: As enumeration used to indicate missing
 objects or mark the end of a group or track. See {{object-status}} below.
 
+* Object Extension Count: The number of Object Extensions present. A value of 0
+  indicates that no Object Extension Headers are present.
+
+* Object Extensions : an optional concatenation of Object Extension Headers. See
+  {{object-extensions}} below.
```
suhasHere, line 1533: Like to understand what do we mean by concatenation here ?

draft-ietf-moq-transport.md
```diff
@@ -1557,6 +1578,35 @@ are sent on a new stream. This is to avoid the status message being lost
 in cases such as a relay dropping a group and reseting the stream the
 group is being sent on.
 
+#### Object Extension Header {#object-extensions}
+Object Extension Headers are visible to relays. Extension Headers MUST be
+forwarded and MUST NOT be modified by relays. The purpose of Extension
+Headers is to facilitate the future evolution of the transport protocol.
```
suhasHere, line 1584: I don't see the last sentence here is needed.

draft-ietf-moq-transport.md
```diff
@@ -1577,6 +1627,8 @@ OBJECT_STREAM Message {
   Group ID (i),
   Object ID (i),
   Publisher Priority (8),
+  Extension Count (i),
+  [Extension headers (...)],
```
suhasHere, line 1631: Updating to latest changes will get rid of most of the changes that follow

suhasHere (changes requested):
This is coming to be pretty close to finishing. I would like to see the PR based off the main branch to make sure of the encodings

draft-ietf-moq-transport.md
wilaw: Meaning a series of Object Extension headers, serialized in sequence without any delimiters between them.

draft-ietf-moq-transport.md
wilaw: Imagine in the future we introduce a new header which signals to a relay that it should transmit this object in its own QUIC stream (for whatever reason). By "supporting" this header, the publisher is signalling that it is aware of any relay processing rules which that header may invoke.

draft-ietf-moq-transport.md
wilaw: Rebased as of 20min ago.

wilaw:
Based on feedback in yesterday's interim meeting, I added Commit https://github.com/moq-wg/moq-transport/pull/502/commits/f2fba0c8ae0d31767f3dad058c3b949b5a3c080d which relaxes the constraints on relays having to always forward and not modify extension headers. New language reads as 

> Object Extension Headers are visible to relays. Extension Headers SHOULD be forwarded and SHOULD NOT be modified by relays.

This allows for the identified use cases in which a) an edge relay wishes to remove watermarking infomration before it reaches the final subscriber and b) relays may add data to tracing extensions as the Object moves through the distribution network.

wilaw:
A request was made by chairs to collate use-cases for extension headers. Documented use-cases are listed below. 

1. 5G XR - as per https://github.com/moq-wg/moq-transport/pull/502/#issuecomment-2471516973 and  https://github.com/moq-wg/moq-transport/pull/502/#issuecomment-2479800594 
2. Tracing - https://opentelemetry.io/docs/concepts/signals/traces/ 
3. LOC - https://datatracker.ietf.org/doc/draft-mzanaty-moq-loc/

If there are more use-cases, please add them.

draft-ietf-moq-transport.md
suhasHere: How about we have something along the lines in the core transport ( below formulation ,may be, exactly what you have, but i am writing it down in a single place to see if we all can agree)

### Relay Extension Processing Rules
1. If a relay understands an extension, it follows the rules as described by the specification that defines the extension, when processing such an extension. Such rules may be to  modify the extension ( by updating or by removing it), for example.
2. If the extension specification requires a given relay to understand an extension but relay doesn't support such an extension, it is treated as protocol violation.
3.  Else  the  MUST forwards the received extension as-is to downstream relays/subscribers.

Now on the point on  how to know if a relay understands an extension or not , there are couple of options
1. Reserve code points for such extensions OR
2. Use setup parameters to identify extensions

Extensions are per object property. I am not totally sure if we need stream level extensions at the moq layer though.

draft-ietf-moq-transport.md
```diff
@@ -1174,7 +1174,24 @@ the `query` portion of the URI to the parameter.
...
+The REQUESTED-EXTENSION parameter (key 0x03) allows the client to request
+the server to acknowledge support for  multiple Extension Header types
+{{object-extensions}} which are required for operation. The value is a
+concatenation of varints, each describing an integer extension header type.
+This parameter is optional. If this parameter is present in the
+CLIENT_SETUP message, then the server MUST respond with a
+REQUESTED-EXTENSION parameter in its SERVER_SETUP message. This parameter
+MUST list the subset of those extensions previously requested by the client
+which the server supports. If the server does not support any of the
+requested extensions, then it MUST respond with a parameter value length of 0.
```
ianswett, line 1192: Omitting the parameter also seems ok, but that's a detail?

draft-ietf-moq-transport.md
```diff
@@ -1174,7 +1174,24 @@ the `query` portion of the URI to the parameter.
...
 #### MAX_SUBSCRIBE_ID {#max-subscribe-id}
 
-The MAX_SUBSCRIBE_ID parameter (Parameter Type 0x02) communicates an initial value for the Maximum Subscribe ID to the receiving subscriber. The default value is 0, so if not specified, the peer MUST NOT create subscriptions.
+The MAX_SUBSCRIBE_ID parameter (Parameter Type 0x02) communicates an initial
+value for the Maximum Subscribe ID to the receiving subscriber. The default
+value is 0, so if not specified, the peer MUST NOT create subscriptions.
+
+#### REQUESTED-EXTENSION parameter {#requested-extensions}
+
+The REQUESTED-EXTENSION parameter (key 0x03) allows the client to request
```
ianswett, line 1183: This extension mechanism I believe means any extension is available to both client and server?  Alternatively, in QUIC extensions like Ack Frequency can be negotiated separately in each direction.

I can see the pros and cons of either, but if my understanding is correct, I think it's worth stating that explicitly.

draft-ietf-moq-transport.md
```diff
@@ -984,6 +984,21 @@ client MUST set the PATH parameter to the `path-abempty` portion of the
...
 
+#### REQUESTED-EXTENSION parameter {#requested-extensions}
+
+The REQUESTED-EXTENSION parameter (key 0x02) allows the client to request
+the server to acknowledge support for  multiple Extension Header types
+{{object-extensions}} which are required for operation. The value is a
+concatenation of varints, each describing an integer extension header type.
+This parameter is optional. If this parameter is present in the
+CLIENT_SETUP message, then the server MUST respond with a
+REQUESTED-EXTENSION parameter in its SERVER_SETUP message. This parameter
```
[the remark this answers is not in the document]
ianswett: See comment above, but I think it should be different if we want to allow for asymmetric negotiation?

draft-ietf-moq-transport.md
```diff
@@ -2095,6 +2112,12 @@ Track or Datagram.
 * Object Status: As enumeration used to indicate missing
 objects or mark the end of a group or track. See {{object-status}} below.
 
+* Object Extension Count: The number of Object Extensions present. A value of 0
+  indicates that no Object Extension Headers are present.
+
+* Object Extensions : an optional concatenation of Object Extension Headers. See
```
ianswett, line 2118: How about just "A sequence of ..."?

```suggestion
* Object Extensions : A sequence of Object Extension Headers. See
```

draft-ietf-moq-transport.md
```diff
@@ -2354,6 +2407,38 @@ STREAM_HEADER_SUBGROUP {
...
+STREAM_HEADER_GROUP {
+  Subscribe ID = 2
+  Track Alias = 2
+  Group ID = 0
+  Publisher Priority = 0
+}
+{
+  Object ID = 0
+  Extension Count  = 1
+    { Type = 0
```
ianswett, line 2425: I think this example needs to be updated?  I think types below a certain value are all varints?

draft-ietf-moq-transport.md
```diff
@@ -2132,6 +2155,30 @@ Any object with a status code other than zero MUST have an empty payload.
...
+  [Header Value (i)]
+  [Header Length (i),
+   Header Value (..)]
+}
+~~~
+{: #object-extension-format title="Object Extension Header Format"}
+
+* Header type: an unsigned integer, encoded as a varint, identifying the type
+  of extension and also the subsequent serialization. Header types in the range
+  0x00 - 0x20 inclusive are followed by a single varint encoded value. Header
```
ianswett, line 2177: This is very similar to 'Parameters'.  It might make sense to use the same encoding for both?

I understand the interest in optimizing for varints, but in both case we could have odd types always be a single varint and even be length+value if we wanted?

draft-ietf-moq-transport.md
ianswett: There's an issue filed for GREASING, FYI: https://github.com/moq-wg/moq-transport/issues/416

ianswett (commented):
Thanks for all the updates Will, this is looking good, but I had some questions and suggestions.

draft-ietf-moq-transport.md
wilaw: This is an Object header, and Objects flow from Publisher to Subscriber. The publisher would be writing the extension header and the subscriber would be reading and/or forwarding it. A "client' and "server" can fulfill either one, or both, of those roles. Not sure therefore that we need to state that the extension is available to be read by a subscriber, as that is a de facto property of all the Object headers.

draft-ietf-moq-transport.md
```diff
@@ -984,6 +984,21 @@ client MUST set the PATH parameter to the `path-abempty` portion of the
 URI; if `query` is present, the client MUST concatenate `?`, followed by
 the `query` portion of the URI to the parameter.
 
+#### REQUESTED-EXTENSION parameter {#requested-extensions}
+
+The REQUESTED-EXTENSION parameter (key 0x02) allows the client to request
+the server to acknowledge support for  multiple Extension Header types
+{{object-extensions}} which are required for operation. The value is a
+concatenation of varints, each describing an integer extension header type.
```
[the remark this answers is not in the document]
wilaw: Draft text has now changed to ".. a sequence of"

draft-ietf-moq-transport.md
ianswett: I understand the reasoning that if you understand the header, you understand the header.  But I'm unsure what the purpose of negotiation is in that context?  Is it to negotiate optimizations to headers or content, such as compression?

wilaw: Negotiation was introduced to support the enforcement of a a required per-hop object encoding between publisher and subscriber behavior. For example, imagine a hypothetical log tracing extension, carrying a log trace. The client may want to indicate that it does not want a connection to the server if the server doesn't support logging by way of the specified extension. 

However, after our first interim call on this subject, we agree to loosen the constraints for extensions to "Extension Headers SHOULD be forwarded and SHOULD NOT be modified by relays". This considerably weakens any attempt to enforce an end-to-end behavior across a delivery chain.

So one of the key questions to resolve this week is "Do we need to negotiate extensions at all?" We could leave them on a best-efforts basis for now and then get some implementation experience around whether or not we require negotiation and whether or not that negotiation should be bi-directional.

draft-ietf-moq-transport.md
wilaw: The same count+TLV encoding is already used for both Parameters and Extension Headers (modulo the varint optimization for extension types 0..20). Is your proposal that we don't call them Extension Headers and instead use Parameters? That would require migrating the varint version to Parameters which complicates things. Over the wire the bits are identical, so I'm inclined to leave the definition as it is.

ianswett: I think calling them Extension Headers is good, but I'd rather the wire format be described as being identical to Parameters.  It's close today, but as you point out, not quite identical.

In the past, someone, possibly @martinduke proposed making all even parameter values encoded as varints and odd encoded as variable length binary values.  I think that proposal may make sense for both extension headers and parameters.

draft-ietf-moq-transport.md
ianswett: I'd be happy to not negotiate them for now and reconsider that later if it turns out to be useful.

draft-ietf-moq-transport.md
```diff
@@ -2391,6 +2476,10 @@ TODO: fill out currently missing registries:
 * Announce Error codes
 * Announce Cancel Reason codes
 * Message types
+* MOQ Extension headers - we wish to reserve extension types 0-127 for
+  standards utilization where space is a premium, 128 - 16383 for
+  standards utilization where space is less of a concern, and 16384 and
+  above for first-come-first-served non-standardization usage.
```
fluffy, line 2482: This seems like about the right approach to IANA registrations

fluffy (changes requested):
I would like to change so that relays can change extentions

draft-ietf-moq-transport.md
wilaw: We can look at bit packing once we have some implementation experience. Resolving for this draft.

draft-ietf-moq-transport.md
wilaw: Following agreement at yesterday's interim meeting, I updated the language to be 

> Extension Headers are defined in external specifications and registered in an IANA table {{iana}}. These specifications define the type and value of the header, along with any rules concerning processing, modification, caching and forwarding. A relay which is coded to implement these rules is said to "support" the extension.
> 
> If unsupported by the relay, Extension Headers MUST NOT be modified, MUST be cached as part of the Object and MUST be forwarded by relays
> 
> If supported by the relay and subject to the processing rules specified in the definition of the extension, Extension Headers MAY be modified, MAY be cached and MAY be forwarded.

draft-ietf-moq-transport.md
wilaw: With yesterday's interim, resolved with agreed language

> Extension Headers are defined in external specifications and registered in an IANA table {{iana}}. These specifications define the type and value of the header, along with any rules concerning processing, modification, caching and forwarding. A relay which is coded to implement these rules is said to "support" the extension.
> 
> If unsupported by the relay, Extension Headers MUST NOT be modified, MUST be cached as part of the Object and MUST be forwarded by relays
> 
> If supported by the relay and subject to the processing rules specified in the definition of the extension, Extension Headers MAY be modified, MAY be cached and MAY be forwarded.

draft-ietf-moq-transport.md
wilaw: Negotiation was removed with the latest commit, following discussion at the interim.

draft-ietf-moq-transport.md
wilaw: I added a simple greasing scheme. 

> Header values: even types are followed by a single varint encoded value. Odd types are followed by a varint encoded length and then the header value. Every integer multiple of 17 or 18 is a reserved greasing value. These types MUST be treated as unsupported by a relay. Header types are registered in the IANA table 'MOQ Extension Headers'. See {{iana}}.

draft-ietf-moq-transport.md
wilaw: This text was replaced along with the processing rules. 

> Object Extensions : A sequence of Object Extension Headers. See {{object-extensions}} below.

> Object Extension Headers are visible to relays and allow the transmission of future metadata relevant to MOQT Object distribution. Extension Headers MUST NOT carry metadata bound to the Object payload. Such metadata should instead be serialized into the payload.

draft-ietf-moq-transport.md
```diff
@@ -1557,6 +1578,62 @@ are sent on a new stream. This is to avoid the status message being lost
 in cases such as a relay dropping a group and reseting the stream the
 group is being sent on.
 
+#### Object Extension {#object-extensions}
+An Object Extension is a concatenation of optional Extension Headers. These
+headers are visible to relays. Extension headers MUST be forwarded and
+MUST NOT be modified by relays. The purpose of Extension Headers is to
```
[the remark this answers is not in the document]
wilaw: Following the interim, we added new processing rules for relays and removed negotiation during setup.

draft-ietf-moq-transport.md
```diff
@@ -1557,6 +1578,62 @@ are sent on a new stream. This is to avoid the status message being lost
...
+allow the transmission of application-specific data as well as future
```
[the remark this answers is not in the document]
wilaw: Language changed to 

> Object Extension Headers are visible to relays and allow the transmission of future metadata relevant to MOQT Object distribution. Extension Headers MUST NOT carry metadata bound to the Object payload. Such metadata should instead be serialized into the payload.
> 
> Extension Headers are defined in external specifications and registered in an IANA table {{iana}}. These specifications define the type and value of the header, along with any rules concerning processing, modification, caching and forwarding. A relay which is coded to implement these rules is said to "support" the extension.
> 
> If unsupported by the relay, Extension Headers MUST NOT be modified, MUST be cached as part of the Object and MUST be forwarded by relays
> 
> If supported by the relay and subject to the processing rules specified in the definition of the extension, Extension Headers MAY be modified, MAY be cached and MAY be forwarded.

draft-ietf-moq-transport.md
```diff
@@ -1557,6 +1578,62 @@ are sent on a new stream. This is to avoid the status message being lost
...
+evolution of the transport protocol. Object Extensions are serialized as
```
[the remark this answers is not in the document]
wilaw: I went with 

> Object Extension Headers are visible to relays and allow the transmission of future metadata relevant to MOQT Object distribution.

draft-ietf-moq-transport.md
```diff
@@ -1577,6 +1654,8 @@ OBJECT_STREAM Message {
   Group ID (i),
   Object ID (i),
   Publisher Priority (8),
+  Extension Count (i),
```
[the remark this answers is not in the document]
wilaw: We can merge with the 1byte for now, but then I have opened a new issue #666  to capture this and other proposals for bit-packing the extension count.

draft-ietf-moq-transport.md
wilaw: I removed that sentence with latest commit.

draft-ietf-moq-transport.md
wilaw: We have now removed negotiation entirely.

draft-ietf-moq-transport.md
wilaw: Agreed. Negotiation has been removed with latest commit.

draft-ietf-moq-transport.md
wilaw: Based on yesterday's interim discussion, I modified the serialization to use the even|varint, odd|TLV scheme. 

> Header values: even types are followed by a single varint encoded value. Odd types are followed by a varint encoded length and then the header value. Every integer multiple of 17 or 18 is a reserved greasing value. These types MUST be treated as unsupported by a relay. Header types are registered in the IANA table 'MOQ Extension Headers'. See {{iana}}.

In a future edit, if parameters also adopt the even/odd scheme, then we can update Extensions to refer to the parameters serialization.

draft-ietf-moq-transport.md
```diff
@@ -2132,6 +2155,30 @@ Any object with a status code other than zero MUST have an empty payload.
 Though some status information could be inferred from QUIC stream state,
 that information is not reliable and cacheable.
 
+
+#### Object Extension Header {#object-extensions}
+Object Extension Headers are visible to relays. Extension Headers SHOULD be
+forwarded and SHOULD NOT be modified by relays. The purpose of Extension
```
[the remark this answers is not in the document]
wilaw: We now say

> Extension Headers are defined in external specifications and registered in an IANA table {{iana}}. These specifications define the type and value of the header, along with any rules concerning processing, modification, caching and forwarding. A relay which is coded to implement these rules is said to "support" the extension.
> 
> If unsupported by the relay, Extension Headers MUST NOT be modified, MUST be cached as part of the Object and MUST be forwarded by relays
> 
> If supported by the relay and subject to the processing rules specified in the definition of the extension, Extension Headers MAY be modified, MAY be cached and MAY be forwarded.

draft-ietf-moq-transport.md
```diff
@@ -2132,6 +2155,30 @@ Any object with a status code other than zero MUST have an empty payload.
 Though some status information could be inferred from QUIC stream state,
 that information is not reliable and cacheable.
 
+
+#### Object Extension Header {#object-extensions}
+Object Extension Headers are visible to relays. Extension Headers SHOULD be
```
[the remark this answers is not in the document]
wilaw: Relays can now modify extensions which they "support". 

> Extension Headers are defined in external specifications and registered in an IANA table {{iana}}. These specifications define the type and value of the header, along with any rules concerning processing, modification, caching and forwarding. A relay which is coded to implement these rules is said to "support" the extension.
> 
> If supported by the relay and subject to the processing rules specified in the definition of the extension, Extension Headers MAY be modified, MAY be cached and MAY be forwarded.

fluffy (changes requested):
Most my stuff is just comments that could be ignored but on the topic of can this be used for for meta data about stuff in payload, I feel very strongly we don't have consensus to merge this with MUST NOT. I thought the consensus from call was this was allowed. 

Other than that one point, looks great and thanks Will for fast turn around on thisl

draft-ietf-moq-transport.md
```diff
@@ -2592,6 +2667,12 @@ TODO: fill out currently missing registries:
 * Announce Error codes
 * Announce Cancel Reason codes
 * Message types
+* MOQ Extension headers - we wish to reserve extension types 0-127 for
+  standards utilization where space is a premium, 128 - 16383 for
+  standards utilization where space is less of a concern, and 16384 and
+  above for first-come-first-served non-standardization usage. Additionally,
+  every integer multiple of 17 or 18 is a reserved greasing value and MUST
```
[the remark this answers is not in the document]
ianswett: I agree that I'd prefer to do greasing of everything in a separate PR, and I don't think we have to use any of the 1 byte codepoints.  Those are the ones I'm pretty sure will work.

draft-ietf-moq-transport.md
```diff
@@ -2329,6 +2325,45 @@ Any object with a status code other than zero MUST have an empty payload.
...
+   Header Value (..)]
+}
+~~~
+{: #object-extension-format title="Object Extension Header Format"}
+
+* Header type: an unsigned integer, encoded as a varint, identifying the type
+  of the extension and also the subsequent serialization.
+* Header values: even types are followed by a single varint encoded value. Odd
+  types are followed by a varint encoded length and then the header value.
+  Every integer multiple of 17 or 18 is a reserved greasing value. These types
```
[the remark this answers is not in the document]
ianswett: I think in the Wednesday virtual interim we decided to do greasing as a separate PR?

ianswett (commented):
Thanks for the fast update.  I made a suggestion for text on extension headers and Object metadata and I agree with Cullen that I'd rather do greasing of everything in a separate PR.

draft-ietf-moq-transport.md
```diff
@@ -2329,6 +2325,45 @@ Any object with a status code other than zero MUST have an empty payload.
 Though some status information could be inferred from QUIC stream state,
 that information is not reliable and cacheable.
 
+#### Object Extension Header {#object-extensions}
+Object Extension Headers are visible to relays and allow the transmission of
+future metadata relevant to MOQT Object distribution. Extension Headers MUST NOT
+carry metadata bound to the Object payload. Such metadata should instead
```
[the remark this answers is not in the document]
wilaw: Modified text as suggested with commit https://github.com/moq-wg/moq-transport/pull/502/commits/10298c40d680d350282d4ad699f1b992034461f3

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
wilaw: Removed greasing with https://github.com/moq-wg/moq-transport/pull/502/commits/10298c40d680d350282d4ad699f1b992034461f3

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
wilaw: Removed greasing with https://github.com/moq-wg/moq-transport/pull/502/commits/10298c40d680d350282d4ad699f1b992034461f3

draft-ietf-moq-transport.md
```diff
@@ -2592,6 +2667,12 @@ TODO: fill out currently missing registries:
 * Announce Error codes
 * Announce Cancel Reason codes
 * Message types
+* MOQ Extension headers - we wish to reserve extension types 0-127 for
+  standards utilization where space is a premium, 128 - 16383 for
```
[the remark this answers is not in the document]
wilaw: Isn't 1 byte max 127 ?

> 1 Byte (7 bits for data)
> Largest decimal integer: 127
> 
> 2 Bytes (14 bits for data)
> Largest decimal integer: 16,383
> 
> 3 Bytes (21 bits for data)
> Largest decimal integer: 2,097,151

wilaw:
Commit https://github.com/moq-wg/moq-transport/pull/502/commits/10298c40d680d350282d4ad699f1b992034461f3 addressed the open rerquests of @fluffy and @ianswett .

draft-ietf-moq-transport.md
```diff
@@ -2329,6 +2325,44 @@ Any object with a status code other than zero MUST have an empty payload.
 Though some status information could be inferred from QUIC stream state,
 that information is not reliable and cacheable.
 
+#### Object Extension Header {#object-extensions}
+Object Extension Headers are visible to relays and allow the transmission of
+future metadata relevant to MOQT Object distribution. Any Object metadata not
+used by the transport SHOULD be serialized as part of the Object and not as
+an extension header.
```
afrind, line 2332: minor: s/part of the Object/part of the Object payload/

I can live with SHOULD here; perhaps RECOMMENDED is also ok?  My reasoning is that if the transport will never look at the extension, it shouldn't waste any time parsing out the fields.  However, as Suhas pointed out, it's unenforceable, and a relay cannot distinguish between an unsupported extension and payload.

afrind (commented):
Individual Review: lgtm.  I see we don't have any language about how relays treat receipt of the same object with different extensions.  I'm ok leaving that out here, and clarifying in a follow up too.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
fluffy: @wilaw - I think that MoQ should be changed to work the way you think it works (see https://github.com/moq-wg/moq-transport/issues/549)  but it does not

Right now the encoding is defined in section 16 of RFC 9000
https://www.rfc-editor.org/rfc/rfc9000#section-16

And for 1 byte, you only get 6 bits of data not 7.

draft-ietf-moq-transport.md
fluffy, line 2332: The problem with serializing in the object is that the object will be end to end encrypted in many cases and thus the rely can not read it. Take the example of capture timestamp for media. Relay can look at this timestamp to figure out how much lag. Then  metrics dashboard can take differences of lag to figure out the latency between two relays. This type of operational information is supper useful on running real systems. SRTP which we are trying to replace with MoQ is explicitly designed to keep that type of timing information in the readable hops along the media path. I feel very strongly we need the same sort of thing here. 

We could introduce two types of option , relay options, and client options so that the relay did not need to parse the client options. Or another design could be to introduce object headers that were not end to end encrypted.  But all of these options end up using another byte even if there are no headers and that is a big impact for audio only applications. 

I don't buy the parsing time problem. Anything reasonable relay is going to be able to parse to the level to check if it is interested in any of the extensions on single core at a rate that greatly exceeds whatever networking technology it uses and probably exceeds the bandwidth to non cached memory so you can probably parse this as fast as you can memcpy it. Either way, I just don't see any performance problem. 

I could live with this saying 

"Any metadata which will not be read or logged by any relay ever including use for operational or diagnostic reasons, 
 SHOULD NOT be sent in an extension header."

fluffy (changes requested):

draft-ietf-moq-transport.md
afrind: Individual Comment:

Let me start at the end, because I think we agree:

> "Any metadata which will not be read or logged by any relay ever including use for operational or diagnostic reasons, SHOULD NOT be sent in an extension header."

This is what I was trying to express.

The current text is:

> Any Object metadata not used by the transport SHOULD be serialized as part of the Object and not as an extension header.

Maybe just amending it to 

> Any Object metadata _never accessed_ by the transport _or relays_ SHOULD be serialized as part of the Object payload and not as an extension header.

?

draft-ietf-moq-transport.md
```diff
@@ -485,26 +485,16 @@ PATH parameter ({{path}}) which is sent in the CLIENT_SETUP message at the
 start of the session.  The ALPN value {{!RFC7301}} used by the protocol
 is `moq-00`.
 
-## Version and Extension Negotiation {#version-negotiation}
```
ianswett, line 488: I don't think we want to remove this text yet.  There could be extensions that aren't extension headers we may want to negotiate?

ianswett (approved):
This LG, but I don't think we should remove the "Extension Negotiation" text towards the beginning of the document, because that's different from extension headers.

draft-ietf-moq-transport.md
wilaw: I'm ok with below text in the interests of advancing the spec. perhaps @ianswett can edit when merging or do you want me to make another commit?

> Any Object metadata never accessed by the transport or relays SHOULD be serialized as part of the Object payload and not as an extension header.

At Denver interim, I'd like to debate the use case where we add timing data to extensions. While it appears useful, it binds the protocol to the payload, which I ardently believe we should strive to avoid. Media flows will want timestamps, gaming systems will want 3D coordinates, vehicle tracking systems will wants GPS coordinates etc. Do we add all of these use cases to the core transport, or do we try to architect a more generic "index" which can be exposed to relays and provide similar utility without the payload binding?

ianswett, line 2332:
```suggestion
future metadata relevant to MOQT Object distribution. Any Object metadata never accessed by the transport or relays SHOULD be serialized as part of the Object
payload and not as an extension header.
```

draft-ietf-moq-transport.md
ianswett: I agree with your intuition, but I'd like to minimize the size of this change, so I'm going to leave this text for now.

draft-ietf-moq-transport.md
```diff
@@ -485,7 +485,7 @@ PATH parameter ({{path}}) which is sent in the CLIENT_SETUP message at the
...
+## Version Negotiation and Extension Negotiation {#version-negotiation}
```
ianswett, line 488:
```suggestion
## Version and Extension Negotiation {#version-negotiation}
```

draft-ietf-moq-transport.md
```diff
@@ -2329,6 +2335,43 @@ Any object with a status code other than zero MUST have an empty payload.
 Though some status information could be inferred from QUIC stream state,
 that information is not reliable and cacheable.
 
+#### Object Extension Header {#object-extensions}
+Object Extension Headers are visible to relays and allow the transmission of
+future metadata relevant to MOQT Object distribution. Any Object metadata never accessed by the transport or relays SHOULD be serialized as part of the Object
```
ianswett, line 2340:
```suggestion
future metadata relevant to MOQT Object distribution. Any Object metadata never
accessed by the transport or relays SHOULD be serialized as part of the Object
```

afrind (approved):

Issue #410, opened 2024-03-06
What happens if we get OBJECT_DATAGRAM on a Stream?


martinduke:
Now that Datagram is no longer just a suggestion, should there be some sort of error if we get OBJECT_DATAGRAM on a Stream?

kixelated:
Absolutely, and it's something I'd like to address in #407

fluffy:
I view this as "protocol error" and any time we get that we should tear down the QUIC connection with appropriate error.

huitema:
There is a simple way to not have that error: just do not allocate a number for the type "OBJECT_DATAGRAM", and state that the content of a datagram is implicitly of state "OBJECT_DATAGRAM". Also saves one byte per datagram.

PR #503, opened 2024-08-07
Split definition of messages for unidirectional and bidirectional streams

vasilvv:
This is technically not purely editorial, but should be backwards-compatible in practice.

This PR does minimal editing to the existing text, and mostly moves things around.  The text in "Data Streams" section needs more cleanup (e.g. to stop referring to "OBJECT messages" that haven't been a thing for a while), but this is out of scope for this PR.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1458,6 +1458,8 @@ stream.
 
 # Data Streams {#data-streams}
 
+A publisher sends Objects matching a subscription on Data Streams. 
+
 All unidirectional MOQT streams, as well as all datagrams, start with a
 variable-length integer indicating the type of the stream in question.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -745,7 +745,7 @@ limits.  See section 2.2 in {{QUIC}}.
 
 # Control Messages {#message}
 
-MOQT uses a signle bidirectional stream to exchange control messages, as
+MOQT uses a single bidirectional stream to exchange control messages, as
 defined in {{session-init}}.  Every signle message on the control stream is
 formatted as follows:
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -743,15 +743,14 @@ increase latency over additional hops.  To mitigate this, a relay SHOULD
 read and write stream data out of order subject to flow control
 limits.  See section 2.2 in {{QUIC}}.
 
-# Messages {#message}
+# Control Messages {#message}
 
-Unidirectional streams and datagrams contain Objects and all other
-messages are sent on a bidirectional stream.
-
-An endpoint that receives an unknown message type MUST close the session.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -743,15 +743,14 @@ increase latency over additional hops.  To mitigate this, a relay SHOULD
...
 limits.  See section 2.2 in {{QUIC}}.
 
-# Messages {#message}
+# Control Messages {#message}
 
-Unidirectional streams and datagrams contain Objects and all other
-messages are sent on a bidirectional stream.
-
-An endpoint that receives an unknown message type MUST close the session.
+MOQT uses a signle bidirectional stream to exchange control messages, as
```
zafergurel, line 748: A small typo:

```suggestion
MOQT uses a single bidirectional stream to exchange control messages, as
```

draft-ietf-moq-transport.md
```diff
@@ -1258,7 +1251,230 @@ TRACK_STATUS_REQUEST Message {
...
+
+All unidirectional MOQT streams, as well as all datagrams, start with a
+variable-length integer indicating the type of the stream in question.
+
+|-------|-----------------------------------------------------|
+| ID    | Stream Type                                         |
+|------:|:----------------------------------------------------|
+| 0x0   | OBJECT_STREAM ({{object-stream}})                   |
+|-------|-----------------------------------------------------|
+| 0x1   | OBJECT_DATAGRAM ({{object-datagram}})               |
```
afrind, line 1463: Should we give OBJECT_DATAGRAM code point 0 also to avoid the problem in #410

vasilvv: Probably yes, though probably not in this PR.

mengelbart: Given that both OBJECT_STREAM and OBJECT_DATAGRAM have the same fields and objects with forwarding preference "Datagram" are either sent as datagrams or dropped, is there a reason to keep both messages instead of using the same type? That would avoid the problem of receiving datagrams on streams, too.

mengelbart: oh I see that was your suggestion. Please ignore :D

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1264,7 +1257,230 @@ TRACK_STATUS_REQUEST Message {
...
+information with status code 0x04.
+
+The receiver of multiple TRACK_STATUS messages for a track uses the information
+from the latest arriving message, as they are delivered in order on a single
+stream.
+
+
+
+# Data Streams {#data-streams}
+
```
suhasHere, line 1460: May be we need to add a note on what are Data Streams for ?

suhasHere (approved):
LGTM . Had a small nit .

draft-ietf-moq-transport.md
ianswett: Good point, I added a sentence.

PR #509, opened 2024-08-25
Remove unnecessary "Missing Group" status

ianswett:
As I observed in this comment: https://github.com/moq-wg/moq-transport/pull/494/files#r1730192507, the missing group status seems equivalent to "End of Group" with ObjectId of 0.  Additionally, there is no obvious value for ObjectId in Objects with a status of "Group Missing".


suhasHere (approved):

fluffy (approved):
makes sense to me

PR #510, opened 2024-08-25
Simplify SUBSCRIBE as a follow-up to FETCH

### LINKED ISSUE #350: When to use Track Alias or Subscribe ID?

### LINKED ISSUE #368: Subscribe vs Fetch and Object states
Posting this here, since I can't think of a better place, and we might want to make sure we have actual issues to track some of those.

_Note: this issue tries to focus on what is being requested, without focusing on how those objects should be delivered (in terms of reliability, etc)._

The proposal is to introduce two logical operations in the protocol:
- Subscribe.
  - Subscribe indicates that all of the future objects for the specified track should be forwarded to the subscriber.
  - The only argument is the starting point (do not forward objects that are before the specified point).
  - There may only be one subscribe per connection.
- Fetch.
  - Fetch indicates that the requester desires to receive all objects within some well-defined range.
  - The arguments are a starting point and an ending point.
  - Fetch is supposed to be “in the past”, but some objects that are technically in the past might arrive later due to reordering, so it’s unclear how useful declaring Fetch to be “in the past” actually is.

The main reason Fetch can have improved semantics over Subscribe is that it has a definite range of objects, and as a consequence:

- The relay can enumerate all of the objects in the range to determine whether they are in cache or not.  Because of that, we can place an expectation that the relay will request objects from the upstream if they are missing from the cache when performing Fetch.
- If the requested range is in the past, all objects can be resolved “synchronously” (i.e. there wouldn’t be objects that are “in the future”, so all objects are either available or permanently gone).
- While Subscribe is expected to just deliver objects as they arrive, Fetch allows for a possibility of objects being transmitted in various ways that would fit different use cases.

Let’s talk more about “resolving” objects.  A Fetch is supposed to resolve all objects in a range.  Potentially, every object can be in one of the following states:

1. Object is available from relay’s/publisher’s memory.
1. Object is partially available and is in the middle of being received.
1. Object is in the future and has not been seen yet.
1. Object is in the past, but the relay has not seen it yet (reordering).
1. Object has been previously received and cached, but then evicted from the cache.
1. Object has been explicitly described as permanently unavailable.
1. Object was supposed to be received, but was temporarily dropped due to network conditions.
1. (this list is meant to be exhaustive, but I am probably forgetting something)

We need to figure out how Fetch reacts to all of those.  We should also figure out how Subscribe deals with permanent and temporary drops, and whether it indicates those (or not).

My proposal was to add two messages that perform those operations:

- ATOMIC_SUBSCRIBE_AND_FETCH requires the recipient to determine the “head”, which is the largest group/object ID received.  A Fetch is started for the range before the head, and Subscribe is started for all of the objects after.
- FETCH is just a straightforward Fetch operation.

We need to analyze those with respect to various use cases, including:

- Subscribing to the live head with some immediately preceding objects (“catch-up objects”).  This is the most important use case, inasmuch as I believe other use cases listed below can be actually satisfied without MoQ.  Note that depending on the size of “catch-up objects”, the desired transmission behavior might be different (if the range is small, those should probably be sent “newest first”; if it’s large, it has to be “oldest first”, or the user would buffer).
- DVR.
- Clipping a fragment from the recent past as a stream highlight.
- (I believe some other use cases were mentioned before, but I probably forgot)

ianswett:
This is another effort at the SUBSCRIBE vs FETCH concept.

I've reduced the number of modes for SUBSCRIBE to 3, and only one of them subscribes to Objects in the future.

I removed Latest Group because the other two modes operate on the latest Object and I wasn't sure how practical it is to expect every relay to cache all of the latest group.  And if they don't, implementations get complex.

This is an opinionated approach, but it has quite a bit of flexibility now that the new priority design has landed.  In particular, it gives the subscriber lots of control over what order Objects on either side of the live head are delivered.

Fixes #368 
Could be extended to fix #350

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1019,7 +1019,9 @@ There are 3 types of filters:
 
 Latest Object (0x2): Specifies an open-ended subscription beginning from
 the current object of the current group.  If no content has been delivered yet,
-the subscription starts with the first published or received group.
+the subscription starts with the first published or received group. Subscribers
+MUST NOT issue more than one subscription for the Latest Object of a track
+within the same session.
 
 AbsoluteStart (0x3):  Specifies an open-ended subscription beginning
 from the Object identified in the StartGroup and StartObject fields and ending
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1014,6 +1014,12 @@ GOAWAY Message {
 The subscriber specifies a filter on the subscription to allow
 the publisher to identify which objects need to be delivered.
 
+If a subscriber wants to subscribe to Objects both before and after
+the Latest Object, it can send a SUBSCRIBE for the Latest Object
+followed by a SUBSCRIBE of type AbsoluteStart.  Depending upon the
+application, one might want to send both SUBSCRIBEs at the same time
+or wait for the first to return before sending the second.
+
 There are 3 types of filters:
 
 Latest Object (0x2): Specifies an open-ended subscription beginning from
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1214,7 +1214,6 @@ in the current group, via the 'Latest Group' filter.
 The subscriber specifies a filter on the subscription to allow
 the publisher to identify which objects need to be delivered.
 
-
 There are 4 types of filters:
 
 Latest Group (0x1) : Specifies an open-ended subscription with objects
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1215,7 +1215,7 @@ The subscriber specifies a filter on the subscription to allow
 the publisher to identify which objects need to be delivered.
 
 
-There are 3 types of filters:
+There are 4 types of filters:
 
 Latest Group (0x1) : Specifies an open-ended subscription with objects
 from the beginning of the current group.  If no content has been delivered yet,
@@ -1223,9 +1223,7 @@ the subscription starts with the first published or received group.
 
 Latest Object (0x2): Specifies an open-ended subscription beginning from
 the current object of the current group.  If no content has been delivered yet,
-the subscription starts with the first published or received group. Subscribers
-MUST NOT issue more than one subscription for the Latest Object of a track
-within the same session.
+the subscription starts with the first published or received group.
 
 AbsoluteStart (0x3):  Specifies an open-ended subscription beginning
 from the Object identified in the StartGroup and StartObject fields and ending
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1014,11 +1014,6 @@ GOAWAY Message {
 The subscriber specifies a filter on the subscription to allow
 the publisher to identify which objects need to be delivered.
 
-If a subscriber wants to subscribe to Objects both before and after
-the Latest Object, it can send a SUBSCRIBE for the Latest Object
-followed by a SUBSCRIBE of type AbsoluteStart.  Depending upon the
-application, one might want to send both SUBSCRIBEs at the same time
-or wait for the first to return before sending the second.
 
 There are 3 types of filters:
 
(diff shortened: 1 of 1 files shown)


wilaw:
This PR does not simplify subscription for non-real-time live stream clients (i.e live sports). These players need a variable buffer before start-up. To achieve this, they need to request several groups behind latest and then all future groups. The suggestion in the PR is for these clients to "it can send a SUBSCRIBE for the Latest Object followed by a SUBSCRIBE of type AbsoluteStart or AbsoluteRange.". The trouble with this is that it complicates the player's handling of the received data. It must first make a request for Latest Object. This will give it the latest and future objects and data will begin to flow immediately after it makes this request. However the player cannot place this data in a decode buffer. It must store it in a temporary buffer and hold it. Based the info it receives in the SUBSCRIBE_OK, it can then make a second request for the absolute range preceding this latest object. That data it can place it its decode buffer and begin decoding and it must then continuously copy data from the first receive buffer in to the decode buffer. This is unnecessarily messy, especially as the recommended workflow. 

A cleaner and simpler approach would be for the player to be able to make a single SUBSCRIBE request and be able to pipe that response directly in to its decode buffer. This can be achieve by either modifying the proposed AbsoluteStart filter so that it doesn't stop at the Latest Object, or adding a new filter type which is "AbsoluteStart and carry on" (bikeshead away). So the player wanting to start behind live would first make a TRACK_STATUS_REQUEST, which would give it the latest Group and Object. With this information it could calculate where it wanted to start behind live and then make a SUBSCRIBE request with the "AbsoluteStart and carry on" filter. It could pipe the response of this request directly in to its decode buffer and begin playback.

fluffy (changes requested):
So most of this looks good but I ams strongly against removing Latest Group. 

The alternative introduces too much unnecessary join time latency for the main case of people joining a conference call. Keep this as is does not complicate the protocol, We have spent extensive WG time getting to this consensus. I think we should direct are energy to more important things than redoing all the arguments about this.

kixelated (changes requested):
The LatestGroup mode is the reason why the concept of Groups even exists at all, and I don't see how the protocol can function without it. You could remove the other modes but absolutely not LatestGroup.

suhasHere:
My proposal would be:
 
1. Keep LatestGroup and LatestObject in subscribe with the restriction on only one of each can be issued. 
2. Remove AbsoluteStart. 
3. Define new Fetch API with AbsoluteRange. 

Applications using Fetch will use Track_status_Request to find the range of groups to request for and can optionally specify a control parameter for flow rate vs network rate. Fetch api also allow to fill in holes by requesting upstream and allows specifying group delivery order. 
This will allow players to use appropriate buffer for starting from history and at the right time issue Subscribe to continue from the live edge.

draft-ietf-moq-transport.md
```diff
@@ -1223,15 +1226,24 @@ the current object of the current group.  If no content has been delivered yet,
...
+from the object identified in the StartGroup and StartObject fields. If the
+StartGroup is prior to the current group, the publisher SHOULD reply with a
+SUBSCRIBE_ERROR with code 'Invalid Range'.
 
 AbsoluteRange (0x4):  Specifies a closed subscription starting at StartObject
 in StartGroup and ending at EndObject in EndGroup.  The start and end of the
 range are inclusive.  EndGroup and EndObject MUST specify the same or a later
-object than StartGroup and StartObject.
+object than StartGroup and StartObject. If the end has already been published,
+the publisher SHOULD reply with a SUBSCRIBE_ERROR with code 'Invalid Range'.
```
ianswett, line 1237:
```suggestion
object than StartGroup and StartObject. If the StartGroup is prior
to the current group, the publisher SHOULD reply with a
SUBSCRIBE_ERROR with code 'Invalid Range'.
```

draft-ietf-moq-transport.md
```diff
@@ -1206,12 +1206,15 @@ GOAWAY Message {
...
 single session and publishers SHOULD treat this as a protocol violation.
+The only objects prior to the current object that can be requested are those
+in the current group, via the 'Latest Group' filter.
 
 **Filter Types**
 
 The subscriber specifies a filter on the subscription to allow
 the publisher to identify which objects need to be delivered.
 
+
```
ianswett, line 1217:
```suggestion
```

ianswett:
I updated this PR, but it's changed enough I decided to create a new PR.

PR #516, opened 2024-08-29
Remove track per stream

### LINKED ISSUE #476: SUBSCRIBE with absolute ranges and flow control
This is another SUBSCRIBE-vs-FETCH issue, but with a specific implementation aspect of it.

Assume I have a group-per-stream track.  I can subscribe to groups 1000 to 2000; it is entirely plausible that the publisher has all of those in cache immediately available, so in a naive implementation it can just open 1000 streams and start sending things; it would run into a flow control, and starve everything else on the connection.

We could set some limit on how many streams an individual subscription has, and that would be one way of solving this: if you set deliver order to ASC, you would process old streams first, and once those are processed (something we don't actually have a signal either), you'd start sending the next group.  Is this what we actually want?

### LINKED ISSUE #480: Reconsider support for track-per-stream
When we originally added track-per-stream, I was skeptical of it being useful, as it's basically equivalent to sending media over TCP, which is a poor fit for latency-sensitive applications (and something that people already can do equally well using other widely-adopted protocols).

Now that we have some further experience with this, the drawbacks of supporting track-per-stream become more apparent:
1. There's a non-trivial amount of code required to support each individual stream mapping.
2. There's a non-trivial amount of design effort and specification text required to support each individual stream mapping.
3. Due to its nature, track-per-stream has its unique design challenges that other mappings don't have (if anything goes wrong, the entire subscription has to be thrown out, making things like timeouts hard to define).

In the narrow use cases where stream-per-track is actually useful, the application can use stream-per-group, and reassemble things itself.  Thus, I'm strongly in favor of removing it, at least as a publisher-selected stream mapping (we can discuss what FETCH does when we get back to that).

### LINKED ISSUE #490: QUIC flow control does not map cleanly to MoQ subscriptions
QUIC flow control doesn't work cleanly with MoQ, because a MoQ subscription can and is typically expected to span multiple streams, and can span multiple subscriptions per session. As a result, neither QUIC's session or stream flow control map cleanly to MoQ subscriptions.

As a contrast, HTTP is a single stream all the way through, and it's possible to exert backpressure, particularly within proxies.

For large requests to a proxy, this creates a challenge, because it can create a number of small range requests, but it doesn't know how large they should be.   The different stream mappings only make this more complex to reason about.  This is one case where stream per subscription/track is trivial.

One possible solution is to have each hop of MoQ decide how to use streams, and if there was a request for a large range that was in the past and subscription was in ascending group order, it could be delivered over a single QUIC stream, at least until it got close to the live edge.

ianswett:
As discussed at the Wednesday virtual interim.

Fixes #480
May fix #490
May fix #476

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1670,15 +1670,15 @@ message and is followed by one or more sets of serialized object fields.
 If a stream ends gracefully in the middle of a serialized Object, terminate the
 session with a Protocol Violation.
 
-Multiple Peeps MAY be sent on a single stream if a session does not have
-enough available streams or if the amount of data already available to send is
-large, and so the backpressure provided by a stream's flow control is valuable.
-This can happen when a subscriber requests a range of Objects that have already
-been published and the publisher can send Objects faster than the subscriber
-can receive them, due to bandwidth, flow control, or other resource limits.
-For a given Group ID, the Peep ID MUST increase with subsequent peeps sent on
-a stream.  The Group ID within a stream MUST be in the Group Order requested by
-the subscriber.
+Multiple Peeps from a single subscription MAY be sent on the same stream if a
+session does not have enough available streams or if the amount of data already
+available to send is large, and so the backpressure provided by a stream's flow
+control is valuable. This can happen when a subscriber requests a range of
+Objects that have already been published and the publisher can send Objects
+faster than the subscriber can receive them, due to bandwidth, flow control,
+or other resource limits. For a given Group ID, the Peep ID MUST increase with
+subsequent peeps sent on a stream.  The Group ID within a stream MUST be in the
+Group Order requested by the subscriber.
 
 A publisher SHOULD NOT open more than one stream at a time with the same stream
 header message type and fields.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1676,8 +1676,9 @@ large, and so the backpressure provided by a stream's flow control is valuable.
 This can happen when a subscriber requests a range of Objects that have already
 been published and the publisher can send Objects faster than the subscriber
 can receive them, due to bandwidth, flow control, or other resource limits.
-Either the Group ID or the Peep ID MUST increase with subsequent peeps sent on
-a stream.
+For a given Group ID, the Peep ID MUST increase with subsequent peeps sent on
+a stream.  The Group ID within a stream MUST be in the Group Order requested by
+the subscriber.
 
 A publisher SHOULD NOT open more than one stream at a time with the same stream
 header message type and fields.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1934,20 +1934,10 @@ OBJECT_DATAGRAM Message {
 
 ## Streams
 
-When objects are sent on streams, the stream begins with a stream header peep
+When objects are sent on streams, the stream begins with a stream header
 message and is followed by one or more sets of serialized object fields.
 If a stream ends gracefully in the middle of a serialized Object, terminate the
 session with a Protocol Violation.
-
-Multiple Peeps from a single subscription MAY be sent on the same stream if a
-session does not have enough available streams or if the amount of data already
-available to send is large, and so the backpressure provided by a stream's flow
-control is valuable. This can happen when a subscriber requests a range of
-Objects that have already been published and the publisher can send Objects
-faster than the subscriber can receive them, due to bandwidth, flow control,
-or other resource limits. For a given Group ID, the Peep ID MUST increase with
-subsequent peeps sent on a stream.  The Group ID within a stream MUST be in the
-Group Order requested by the subscriber.
 
 A publisher SHOULD NOT open more than one stream at a time with the same stream
 header message type and fields.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1555,8 +1555,6 @@ variable-length integer indicating the type of the stream in question.
 |------:|:----------------------------------------------------|
 | 0x1   | OBJECT_DATAGRAM ({{object-datagram}})               |
 |-------|-----------------------------------------------------|
-| 0x50  | STREAM_HEADER_TRACK ({{stream-header-track}})       |
-|-------|-----------------------------------------------------|
 | 0x52  | STREAM_HEADER_PEEP  ({{stream-header-peep}})        |
 |-------|-----------------------------------------------------|
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1667,10 +1667,19 @@ OBJECT_DATAGRAM Message {
 
 ## Streams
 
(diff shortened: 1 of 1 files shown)


kixelated:
IMO multiple peeps per stream is even worse than a stream per track. You inherit all of the problems (priorities, head-of-line blocking, inability to reset) but now it's implicit rather than explicit.

Even worse, the text doesn't require that peeps share the same track/subscription.

I don't think MAX_STREAMS is an issue worth degrading the protocol. Make implementations raise the limit rather than impose unnecessary HoLB.

So IMO just delete a track per stream and offer no replacement. Make a stream for each group, and if you're blocked by MAX_STREAMS before MAX_DATA then the limit needs to be raised.

suhasHere:
I think we just landed a PR saying a PEEP == Stream. I am bit confused on the wording of multiple peeps per stream. 
Also I am not too sure removing stream per track solves some of the issues raised and they do exist regardless. Some are fetch related and some as Luke pointed out are on the point.

I am not saying we definitely need Stream per Track or not, but I think it has its use. However I tend to disagree that it makes implementation complicated.

ianswett:
> IMO multiple peeps per stream is even worse than a stream per track. You inherit all of the problems (priorities, head-of-line blocking, inability to reset) but now it's implicit rather than explicit.

The intent is that you'd only use this when it's beneficial.

> 
> Even worse, the text doesn't require that peeps share the same track/subscription.

I clarified that, but the intent was always that they're Peeps from the same subscription.
> 
> I don't think MAX_STREAMS is an issue worth degrading the protocol. Make implementations raise the limit rather than impose unnecessary HoLB.

I'm a lot less concerned about MAX_STREAMS and more concerned about memory usage and buffering when a subscriber requests a large amount of Data/Objects from the past (aka Fetch) and there is a real need for backpressure/flow control of some sort.  Inventing a new mechanism in MoQ to do that is complex, but caching HTTP proxies deal with this today quite well for large requests. 

> 
> So IMO just delete a track per stream and offer no replacement. Make a stream for each group, and if you're blocked by MAX_STREAMS before MAX_DATA then the limit needs to be raised.

I'm fine with just deleting track per stream as well, and we can limit this PR to only that if people prefer.

vasilvv:
I'm not sure falling back from multi-stream to single-stream is a good idea: there is no good reason for the server to have large per-stream buffers but small total limit on streams, so I think the preferred approach here would be to buffer outgoing peeps until the credits appears.

Otherwise this looks good, though you might need to remove Track as a forwarding preference from other places in the spec.

draft-ietf-moq-transport.md
```diff
@@ -1667,55 +1665,27 @@ OBJECT_DATAGRAM Message {
...
 ## Streams
 
-When objects are sent on streams, the stream begins with a stream
-header message and is followed by one or more sets of serialized object fields.
+When objects are sent on streams, the stream begins with a stream header peep
+message and is followed by one or more sets of serialized object fields.
 If a stream ends gracefully in the middle of a serialized Object, terminate the
 session with a Protocol Violation.
 
+Multiple Peeps from a single subscription MAY be sent on the same stream if a
```
suhasHere, line 1673: This conflicts with the current definition of Peeps.

vasilvv:
Another problem is that if we ever assign semantics to closing a peep via a FIN, or require resetting a stream upon certain events (e.g. timeouts), "multiple peeps in a single stream" completely breaks that.

kixelated:
I definitely don't want a relay converting N streams into 1 stream by coalescing peeps, introducing HoLB in the process. The reverse is probably fine but it's still extremely situational; the broadcaster is trying to send more concurrent peeps than the QUIC connection allows?

But in general this just seems extremely fishy. We're worried about too many streams so we invent our own stream within a stream abstraction. Use fewer peeps or raise MAX_STREAMS.

suhasHere:
I am proposing to hold off merging this PR until fetch discussions are complete. Since the issues are related to flow control/range requests and  not specific to  forwarding mode preference. 
Having said that I am not in favor or against the stream-per-track mode, but hoping we come to a decision once we have spent some time figuring out range requests in general.

draft-ietf-moq-transport.md
```diff
@@ -1781,55 +1779,27 @@ OBJECT_DATAGRAM Message {
 
 ## Streams
 
-When objects are sent on streams, the stream begins with a stream
-header message and is followed by one or more sets of serialized object fields.
+When objects are sent on streams, the stream begins with a stream header peep
+message and is followed by one or more sets of serialized object fields.
 If a stream ends gracefully in the middle of a serialized Object, terminate the
 session with a Protocol Violation.
```
fluffy, line 1786: The following paragraph seem about peeps not removing stream per track so I think it should be in a seperate PR.

draft-ietf-moq-transport.md
fluffy: I do't think we should allow multiple peeps over the sam stream.  It is a fringe case that will not happen most the time and will not be tested. It will be hard to test and many receivers will fail when it happens. It will get very complicated to decide how to prioritize things. It will mean we need to communicate the information in each object instead of just the stream header.

If a publisher runs out of streams its needs to close the subscription with some sort of not enough streams error.

draft-ietf-moq-transport.md
```diff
@@ -1781,55 +1779,27 @@ OBJECT_DATAGRAM Message {
...
 
+Multiple Peeps from a single subscription MAY be sent on the same stream if a
+session does not have enough available streams or if the amount of data already
+available to send is large, and so the backpressure provided by a stream's flow
+control is valuable. This can happen when a subscriber requests a range of
```
fluffy, line 1790: I think this is just a good example to show why FETCH should go over a single stream even if it has many groups and peeps worth of data.

draft-ietf-moq-transport.md
```diff
@@ -1781,55 +1779,27 @@ OBJECT_DATAGRAM Message {
...
+Objects that have already been published and the publisher can send Objects
+faster than the subscriber can receive them, due to bandwidth, flow control,
+or other resource limits. For a given Group ID, the Peep ID MUST increase with
```
fluffy, line 1793: I don't think requiring this to increment will work. Even if the original publisher sent things in order of 1,2,3 an prioritization at an intermediate relay could result in things being sent in the order 1,3,2 so from the downstream point of view things are not in order.

draft-ietf-moq-transport.md
```diff
@@ -1781,55 +1779,27 @@ OBJECT_DATAGRAM Message {
...
+subsequent peeps sent on a stream.  The Group ID within a stream MUST be in the
+Group Order requested by the subscriber.
```
fluffy, line 1795: with remove of stream per track I don't think you can have more than one group in a stream so I am not following this. If this is for multiple groups in a single stream then does the the relay have to buffer all future groups waiting for the last object in the first group.

fluffy (changes requested):
There is a bunch of stuff in here about peeps that I don't think works.

ianswett:
I think we'll end up removing track per stream, but let's do so after we resolve the conversation about Fetch.

PR #518, opened 2024-09-04
Rewrite the Priority section to improve clarity

### LINKED ISSUE #517: Change in publisher priority is underdefined
The current text says "When two or more subscriptions have equal subscriber priority, the original publisher priority is considered next and can change within the track, so subscriptions are prioritized based on the highest priority data available to send."

There are two issues here.  The first is that "data available to send" is not well-defined for a protocol running over QUIC, in the sense that it's unclear as to when the data should be deemed "unavailable" (does it become unavailable after we put it into the QUIC buffer? after we send it? after it gets acked?).  The second is that doing this requires extra bookkeeping to keep track of the current publisher priority per subscription (@martinduke implemented this in our MoQT prototype, and it was definitely quite a bit of complexity and overheard that seems unnecessary, to me at least).

(As an aside, I am not quite sure as to where the "publisher priority of a track is the max priority for all objects in the track" came from; I remember myself arguing for "publisher priority of a _stream_ is the max priority for all objects in the track" at the interim, but that point was moot since we don't actually allow specifying different priorities on a single stream in the current wire format anyways.)

I can see at least two ways we can simplify this:
1. Make the publisher priority of a track to be equal to the priority of the latest (=highest sequence number) object in the track.
2. Make it so that the publisher priority of a peep/datagram is just what is in its header, and does not retroactively change priorities of other peeps and datagrams on that track.  This effectively makes publisher priority have the same expressive power as peep priority, so we may want to merge those if we decide to go this way.

### LINKED ISSUE #585: Alternate Priority Clarification Proposal
This is an individual technical proposal for modifying the priority algorithm added in draft-05 (herein called "Seattle").

Draft-05 contains the following text:

> When two or more subscriptions have equal subscriber priority, the original publisher priority is considered next and can change within the track, so subscriptions are prioritized based on the highest priority data available to send. For example, if the subscription had data at priority 6 and priority 10 to send, the subscription priority would be 6. 

My interpretation of this text is that when a relay receives a new object with higher priority than any queued objects on that subscription, it elevates the publisher subscription priority to this new higher priority.  When it receives objects with lower priority than any queued objects, there is no impact to the publisher subscription priority.  When the relay dequeues an object of higher priority than remaining queued objects, it lowers the priority of the subscription.

Maintaining the subscription priority in this way is complex and potentially computationally expensive. The following proposal is intended to address that while otherwise maintaining the properties of Seattle algorithm, specifically selecting the next group within a prioritized subscription before selecting a subgroup from that group.

**MoQ Priority Proposal “Boston”**

The proposal supports the following priority signals:

1. **Subscriber Subscription Priority**: Subscriber specified priority of a subscription relative to other subscriptions
2. **Publisher Subscription Priority**: Publisher specified priority of a subscription relative to other subscriptions
3. **Subscriber Group Order**: Subscriber specified group order preference within a subscription, Ascending, Descending or Default to Publisher
4. **Publisher Group Order**: Publisher specified group order preference within a subscription; only considered when Subscriber Group Order is Default to Publisher.
5. **Object Priority**: Publisher specified priority of subgroup and datagram relative to other subgroups/datagrams in the same group

When “highest” and “lowest” are used to compare priorities, “highest” means the highest priority, which has the lowest numerical value, and vice versa.

**The priority algorithm is:**

* A relay selects the subscription with the highest **Subscriber Subscription Priority**;
	* If more than one subscription has this priority; select the subscription with the highest **Publisher Subscription Priority**;
		* If more than one subscription has this priority; the implementation chooses, with the assumption that every subscription makes some progress

* From the selected subscription, use the **Group Order** to choose the group to send from (either highest or lowest available).  

* If using stream-per-track, send the object with the lowest Object ID.

* If using stream-per-subgroup, select the subgroup with the highest **Object Priority**
	* if more than one subgroup has the same priority; choose the one with the lower subgroup ID.

* If using datagrams, select the Object with the highest **Object Priority**
	* if more than one datagram has the same priority; choose the one with the lower Object ID

**Analysis:**

The proposal in #518 uses the Object Priority as both the Publisher Subscription Priority and Object Priority.  In order to express increased importance of a particular subscription or track, it must also elevate the priority of the subgroup or datagrams relative to the subgroups and datagrams in the same group and other groups.  

Both Seattle and 518 can only change subscription priority by sending new subgroups or datagrams, although the wire format proposed below could also address this.

The 518 proposal also relegates Group Order to a less significant place in the priority hierarchy.  Consider a track with a single subgroup per group.  The subscriber may have asked for Group Order Ascending, but the publisher sets Object Priority as follows:

```
Group 0, Subgroup 0, Priority 1
Group 1, Subgroup 0, Priority 3
Group 2, Subgroup 0, Priority 2
```

Assuming queuing happens at the relay, it will send the groups 0, 2, 1 which is neither ascending nor descending, and overrides the subscriber preference.

In the new proposal, when the relay receives a new publisher priority signal, all unsent data in the subscription updates to that priority, the send order of groups and objects within that subscription remains unchanged.  Subscription Priority can be changed at any time and does not require producing an object or beginning a new subgroup.

“Boston” Wire Format changes:

Identical to draft-06 with the following additions:

1. - Subscribe OK contains a mandatory field for **Publisher Subscription Priority**
2. - One or more new messages to change the **Publisher Subscription Priority**:

* A new Object Status “Family”, eg: 

`UPDATE_PUBLISHER_SUBSCRIPTION_PRIORITY 0x100 - 0x1FF` where the last 8 bits are the new priority

* A new unidirectional stream header

```
UPDATE_PUBLISHER_SUBSCRIPTION_PRIORITY {
	Subscribe ID (i)
	Priority (8)
}
```

* A control message 

```
UPDATE_PUBLISHER_SUBSCRIPTION_PRIORITY {
	0xTBD
	Length (i)
	Subscribe Id (i)
	Priority (8)
}
```

Any of these messages could work but they have tradeoffs that can be discussed.

vasilvv:
This change provides a step-by-step algorithm for how to pick the next object to send, and which priorities take precedence over another, all in one place.

This PR may or may not change the normative behavior of priorities depending on one's interpretation of the current text (which, as we learned in today's call, appears to vary between people).

Fixes #517

Closes #585

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -630,7 +630,7 @@ Since a single peep or datagram has a single publisher priority, it can be
 useful to conceptualize this process as scheduling peeps or datagrams
 instead of individual objects on them.
 
-A _priority_ is an unsigned integer with a value between 0 and 255.
+A _priority number_ is an unsigned integer with a value between 0 and 255.
 A lower priority number indicates higher priority; the highest priority is 0.
 
 _Subscriber Priority_ is a priority number associated with an individual
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -630,7 +630,7 @@ Since a single peep or datagram has a single publisher priority, it can be
 useful to conceptualize this process as scheduling peeps or datagrams
 instead of individual objects on them.
 
-A _priority number_ is an unsigned integer with a value between 0 and 255.
+A _priority_ is an unsigned integer with a value between 0 and 255.
 A lower priority number indicates higher priority; the highest priority is 0.
 
 _Subscriber Priority_ is a priority number associated with an individual
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -637,7 +637,10 @@ _Subscriber Priority_ is a priority number associated with an individual
 subscription.  It is specified in the SUBSCRIBE message, and can be updated via
 SUBSCRIBE_UPDATE message.  The subscriber priority of an individual schedulable
 object is the subscriber priority of the subscription that caused that object
-to be sent.
+to be sent. When subscriber priority is changed, a best effort SHOULD be made
+to change the apply that to all objects that have not been sent, but it is
+implementation dependent what happens to objects that have already been
+received and possibly scheduled.
 
 _Publisher Priority_ is a priority number associated with an indiviaul
 schedulable object.  It is specified in the header of the respective peep or
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -617,9 +617,9 @@ A schedulable object in MoQT is either:
    to be sent.
 1. An object that belongs to a track with delivery preference Datagram.
 
-Since a single peep can contain at most one schedulable object, it can be
-useful to conceptualize this process as scheduling peeps themselves instead
-of individaul objects on them.
+Since a single peep or datagram has a single publisher priority, it can be
+useful to conceptualize this process as scheduling peeps or datagrams
+instead of individual objects on them.
 
 A _priority number_ is an unsigned integer with a value between 0 and 255.
 A lower priority number indicates higher priority; the highest priority is 0.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -608,52 +608,67 @@ MoQ priorities allow a subscriber and original publisher to influence
 the transmission order of Objects within a session in the presence of
 congestion.
 
+## Definitions
+
+MoQT maintains priorities between different _schedulable objects_.
+A schedulable object in MoQT is either:
+
+1. An object that belongs to a peep where that object would be the next object
+   to be sent.
+1. An object that belongs to a track with delivery preference Datagram.
+
+Since a single peep can contain at most one schedulable object, it can be
+useful to conceptualize this process as scheduling peeps themselves instead
+of individaul objects on them.
+
+A _priority number_ is an unsigned integer with a value between 0 and 255.
+A lower priority number indicates higher priority; the highest priority is 0.
+
+_Subscriber Priority_ is a priority number associated with an individual
+subscription.  It is specified in the SUBSCRIBE message, and can be updated via
+SUBSCRIBE_UPDATE message.  The subscriber priority of an individual schedulable
+object is the subscriber priority of the subscription that caused that object
+to be sent.
+
+_Publisher Priority_ is a priority number associated with an indiviaul
+schedulable object.  It is specified in the header of the respective peep or
+datagram, and is the same for all objects in a single peep.
+
+_Group Order_ is a property of an invidual subscription.  It can be either
+'Ascending' (groups with lower group ID are sent first), or 'Descending'
+(groups with higher group ID are sent first).  The publisher communicates its
+group order in the SUBSCRIBE_OK message; the subscriber can override it in its
+SUBSCRIBE message.  The group order of an existing subscription cannot be
+changed.
+
+## Scheduling Algorithm
+
+When an MoQT publisher has multiple schedulable objects it can choose between,
+the objects SHOULD be selected as follows:
+
+1. If two objects have a different subscriber priority associated with them,
+   the one with **the highest subscriber priority** is sent first.
+1. If two objects have the same subscriber priority, but a different publisher
+   priority, the one with **the highest publisher priority** is sent first.
+1. If two objects have the same subscriber and publisher priority, but belong
+   to two different groups of the same track received through the same
+   subscription, **the group order** of the associated subscription is used to
+   decide the one that is sent first.
+1. If two objects belong to the same group of the same track received through
+   the same subscription, the one with **the lowest Peep ID** (for tracks with
+   delivery preference Peep), or **the lowest Object ID** (for tracks with
+   delivery preference Datagram) is sent first.
+
+This algorithm does not provide a well-defined ordering for objects that belong
+to different subscriptions, but have the same subscriber and publisher
+priority.  The ordering in those cases is implementation-defined, though the
+expectation is that all subscriptions will be able to send some data.
+
 Given the critical nature of control messages and their relatively
 small size, the control stream SHOULD be prioritized higher than all
 subscribed Objects.
 
-The subscriber indicates the priority of a subscription via the
-Subscriber Priority field and the original publisher indicates priority
-in every stream or datagram header.  As such, the subscriber's priority is a
-property of the subscription and the original publisher's priority is a
-property of the Track and the Objects it contains. In both cases, a lower
-value indicates a higher priority, with 0 being the highest priority.
-
-When Objects are contained in Peeps, all Objects in the Peep have the same
-priority.
-
-The Subscriber Priority is considered first when selecting a subscription
-to send data on within a given session. When two or more subscriptions
-have equal subscriber priority, the original publisher priority is considered
-next and can change within the track, so subscriptions are prioritized based
-on the highest priority data available to send. For example, if the subscription
-had data at priority 6 and priority 10 to send, the subscription priority would
-be 6. When both the subscriber and original publisher priorities for a
-subscription are equal, how much data to send from each subscription is
-implementation-dependent, but the expectation is that all subscriptions will
-be able to send some data.
-
-The subscriber's priority can be changed via a SUBSCRIBE_UPDATE message.
-This updates the priority of all unsent data within the subscription,
-though the details of the reprioitization are implementation-specific.
-
-Subscriptions have a Group Order of either 'Ascending' or 'Descending',
-which indicates whether the lowest or highest Group Id SHOULD be sent first
-when multiple Groups are available to send.  A subscriber can specify either
-'Ascending' or 'Descending' in the SUBSCRIBE message or they can specify they
-want to use the Original Publisher's Group Order, which is indicated in
-the corresponding SUBSCRIBE_OK.
-
-Within the same Group, and the same priority level,
-Objects with a lower Object Id are always sent before objects with a
-higher Object Id, regardless of the specified Group Order. If the group
-contains more than one Peep and the priority varies between these Peeps,
-higher priority Peeps are sent before lower priority Peeps. If the specified
-priority of two Peeps in a Group are equal, the lower Peep ID has priority.
-Within a Peep, Objects MUST bs sent in increasing Object ID order.
-
-The Group Order cannot be changed via a SUBSCRIBE_UPDATE message, and
-instead an UNSUBSCRIBE and SUBSCRIBE can be used.
+## Considerations for Setting Priorities
 
 Relays SHOULD respect the subscriber and original publisher's priorities.
 Relays SHOULD NOT directly use Subscriber Priority or Group Order
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -623,7 +623,7 @@ MoQT maintains priorities between different _schedulable objects_.
 A schedulable object in MoQT is either:
 
 1. An object that belongs to a peep where that object would be the next object
-   to be sent.
+   to be sent in that peep.
 1. An object that belongs to a track with delivery preference Datagram.
 
 Since a single peep or datagram has a single publisher priority, it can be


vasilvv:
@fluffy You've expressed interest in seeing a step-by-step algorithm for sending things to be explicitly spelled out before.  Does the text in this PR match your expectations of how this should look like?

draft-ietf-moq-transport.md
```diff
@@ -608,52 +608,67 @@ MoQ priorities allow a subscriber and original publisher to influence
...
+
+## Scheduling Algorithm
+
+When an MoQT publisher has multiple schedulable objects it can choose between,
+the objects SHOULD be selected as follows:
+
+1. If two objects have a different subscriber priority associated with them,
+   the one with **the highest subscriber priority** is sent first.
+1. If two objects have the same subscriber priority, but a different publisher
+   priority, the one with **the highest publisher priority** is sent first.
```
afrind, line 652: This implies that publisher priority trumps subscriber group order.  That was not how I read the previous algorithm.  Higher publisher priority would *select a subscription* to send from, then you go through to group order.  This is how we got min.

I like the property that says group order is ascending or descending *only*, and priority doesn't influence group order.  Trying to use the object priority as both intra and inter track is weird.

draft-ietf-moq-transport.md
```diff
@@ -608,52 +608,67 @@ MoQ priorities allow a subscriber and original publisher to influence
...
 congestion.
 
+## Definitions
+
+MoQT maintains priorities between different _schedulable objects_.
+A schedulable object in MoQT is either:
+
+1. An object that belongs to a peep where that object would be the next object
+   to be sent.
+1. An object that belongs to a track with delivery preference Datagram.
```
martinduke, line 618: Nit: fix numbering here

martinduke (approved):
LGTM

afrind (commented):
Individual Comment

draft-ietf-moq-transport.md
```diff
@@ -608,52 +608,67 @@ MoQ priorities allow a subscriber and original publisher to influence
...
+
+Since a single peep can contain at most one schedulable object, it can be
+useful to conceptualize this process as scheduling peeps themselves instead
+of individaul objects on them.
+
+A _priority number_ is an unsigned integer with a value between 0 and 255.
```
ianswett, line 624:
```suggestion
A _priority_ is an unsigned integer with a value between 0 and 255.
```

nit: Can we just call it priority?

draft-ietf-moq-transport.md
```diff
@@ -608,52 +608,67 @@ MoQ priorities allow a subscriber and original publisher to influence
...
+MoQT maintains priorities between different _schedulable objects_.
+A schedulable object in MoQT is either:
+
+1. An object that belongs to a peep where that object would be the next object
+   to be sent.
+1. An object that belongs to a track with delivery preference Datagram.
+
+Since a single peep can contain at most one schedulable object, it can be
+useful to conceptualize this process as scheduling peeps themselves instead
+of individaul objects on them.
```
ianswett, line 622:
```suggestion
Since a single peep or datagram has a single publisher priority, it can be
useful to conceptualize this process as scheduling peeps or datagrams
instead of individual objects on them.
```

ianswett (approved):
Some nits, but I think this is very close.

draft-ietf-moq-transport.md
vasilvv: > Trying to use the object priority as both intra and inter track is weird.

I agree that it might feel a little bit weird, but if priority is applied before group order, this makes sense, since this should be more expressive than just having two of those.

As to publisher priority vs group order, I feel like for the use cases that I can think of, the group order should always apply _after_ the publisher priority.  E.g. for the active vs non-active speaker case, you'd always want to prioritize the more-useful audio over less-useful regardless of stream position; or for SVC, you'd probably rather have all of the base layer frames than have the newest group in the best quality and a missing group right before it.

(Of course, even if that is not true, you can choose the opposite behavior by setting publisher priority as the same for all track, and then using peep IDs to prioritize those within the group.)

afrind: The problem that I see if that the publisher gets the final say on group order here -- if the subscriber asks for one track with pri=0 and order=asc, but the publisher sets the peep priorities for each group in descending order, the data will be delivered dsc.  I'd find it more consistent and easier to think about if we separate publisher track priority from peep/object priority (Martin's Option 3).  Then it's `sub pri -> pub pri -> group order -> peep/obj pri -> obj ID`.

In any case, this PR doesn't feel like a clarification so much as a change to the algorithm.

ianswett: Adding an explicit track priority is completely reasonable, though it'd be good to understand what cases it was necessary?  For SVC, I'm quite sure you'd want the peep priority to be more important than group order.

We only have 256 priority values, so your example won't last that long for most use cases.  Also, your example is somewhat adversarial, so I'm not sure how realistic it is?

ianswett: I created https://github.com/moq-wg/moq-transport/issues/519 to discuss the question of whether publishers should specify a priority in addition to the subscriber priority.

draft-ietf-moq-transport.md
```diff
@@ -608,52 +608,67 @@ MoQ priorities allow a subscriber and original publisher to influence
...
+Since a single peep or datagram has a single publisher priority, it can be
+useful to conceptualize this process as scheduling peeps or datagrams
+instead of individual objects on them.
+
+A _priority number_ is an unsigned integer with a value between 0 and 255.
+A lower priority number indicates higher priority; the highest priority is 0.
+
+_Subscriber Priority_ is a priority number associated with an individual
+subscription.  It is specified in the SUBSCRIBE message, and can be updated via
+SUBSCRIBE_UPDATE message.  The subscriber priority of an individual schedulable
```
fluffy, line 629: I would add here something along lines of "when it is changed, in a relay it is implementation dependent on how it will apply to object that the relay has already received".

Clearly if a something had been pushed into the QUIC stack but not yet pushed to the wire, it would be pretty hard to change. I think it is fine to leave this very implementation dependent because this is something that is happening for a short time on the edge of the change and weather a given object gets the old or new priority during the switch over does not matter much to apps.

draft-ietf-moq-transport.md
```diff
@@ -608,52 +608,67 @@ MoQ priorities allow a subscriber and original publisher to influence
 the transmission order of Objects within a session in the presence of
 congestion.
 
+## Definitions
+
+MoQT maintains priorities between different _schedulable objects_.
+A schedulable object in MoQT is either:
```
fluffy, line 614: Let me ask a questions first.  If every object was a schedulable object, does it change the algorithm in the rest of this in any way ? When I read the rest of this I think the answer is No but I feel like others might feel the answer is yes and that I am missing something that is a key issue for other people. 

The way I would implement this is put all the object on a priority queue where the priority on the q was an integer basically formed by concatenating subscriber priority, publisher priority, and either the group id or reverse group id based on direction, then peep id, then object id.  Then when there is space available send an object, you grab the next object out of queue and and send it.  I want to understand if this is wrong. If it is wrong, I am probably fine with what people want but I think it will help me get to understanding why people keep saying this is complicated. If this is right, then I don't think we need the concept of a schedulable object.

draft-ietf-moq-transport.md
```diff
@@ -608,52 +608,67 @@ MoQ priorities allow a subscriber and original publisher to influence
...
+1. If two objects have a different subscriber priority associated with them,
+   the one with **the highest subscriber priority** is sent first.
+1. If two objects have the same subscriber priority, but a different publisher
+   priority, the one with **the highest publisher priority** is sent first.
+1. If two objects have the same subscriber and publisher priority, but belong
+   to two different groups of the same track received through the same
+   subscription, **the group order** of the associated subscription is used to
+   decide the one that is sent first.
+1. If two objects belong to the same group of the same track received through
+   the same subscription, the one with **the lowest Peep ID** (for tracks with
```
fluffy, line 658: Not a big deal but I would have a preference of limiting peep ID to 8 bits if they are going to be used as one of the priority fields.

fluffy: I do not see any reason to add peep into this. If the publisher wanted the object in one peep to be delivered ahead of another peep, it should have set the publisher priority. I can probably live with this but I'd like to understand why we need it.

fluffy (changes requested):
This is a huge improvement and I like it. 

I do have one questions about if we need the concept of scheduleable that I would like to get to resolution on.

draft-ietf-moq-transport.md
fluffy: Looking at wills slides from seattle interim it is clear that publisher priority override group id other wise nearly all the uses cases that use publisher priority don't work. I think if we just step back and look at uses cases, we can reason our way though the conversation we have had several times and end up back at the space spot of what Victor has in the PR now 

There are not enough priorities codes for a publisher to even do the example Alan raised at top of this thread. I view that sort of use cases as just a broken application if the publisher and subscriber are trying to do totally conflicting things and does not bother me much that weird things happen in that case.

afrind: Individual Comment:

I agree the Seattle proposal was: pick a track by sub priority, then min(pub-priority-of-pending-objects), select a group by group order, order by object ID.

There's a complaint that min priority is too hard to implement.  Sounds like the new proposal is:

Pick a track by sub priority, then pub priority, then group by group order, order by peep/object ID.

Here's what I think is a weird case:

The subscriber asked for ascending groups.
In group 1, all objects have the same pub pri = 100.
In group 2, some objects have pub pri 100 but some are more important and get pub pri 90
  -> The publisher wants to express relative priority of the objects *within the group*

The new algorithm will deliver group 2 pri=90 objects first and that seems wrong.  The Seattle algorithm would send group 1 object in obj ID order, then group 2 pri=90 objects, then group 2 pri=100 objects.  This seemed more correct, if harder to implement.  You can sort of fix this by incrementing the base priority of objects in every group, but you will run out of priority levels after 256 groups.

There seemed to be general consensus that groups would only be delivered ascending or descending and I think we should try to preserve that property -- do people care less about this now?  I'm inclined to think we need to set aside some time to discuss at the hybrid interim next month.

fluffy: One observation about ascending / descending order is that it seems the use case for it are all times where it is a fetch not a subscribe. It also seems that for flow control reason fetch will likely flow over one stream. I wonder if we could just remove descending order for subscribe.  I've pointed out in the past it is totally useless and in most cases where the things are not congested and the subscribe is on the live edge, the subscribe will get things in the order the publisher published them regardless of how any of the priorities and orders are set. This will mean a subscriber asks for things in reverse order and will then get most things in forward order with perhaps a few object in reverse order then there was a bit of congestion. I don't object to it being here if others want it because it is all trivial to implement but I remain of the opinion order is unless for subscribe and critical for fetch.

draft-ietf-moq-transport.md
fluffy: @vasilvv - This question here is the only question I have of significance on this PR. Love the PR - thanks.

fluffy:
At interim meeting, decided to give Will 2 weeks to write an alternative PR to this one and can't get something better, merge this.

wilaw:
> decided to give Will 2 weeks to write an alternative PR

Are you sure that was me? I don't have an alternate PR in mind.

martinduke:
> > decided to give Will 2 weeks to write an alternative PR
> 
> Are you sure that was me? I don't have an alternate PR in mind.

@afrind has the token to produce an alternative

afrind:
Alternate proposal is here: https://github.com/moq-wg/moq-transport/issues/585

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -608,52 +608,67 @@ MoQ priorities allow a subscriber and original publisher to influence
 the transmission order of Objects within a session in the presence of
 congestion.
 
+## Definitions
+
+MoQT maintains priorities between different _schedulable objects_.
+A schedulable object in MoQT is either:
+
+1. An object that belongs to a peep where that object would be the next object
+   to be sent.
```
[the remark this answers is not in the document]
ianswett:
```suggestion
   to be sent in that peep.
```

draft-ietf-moq-transport.md
ianswett: I tried to add a clarifying sentence, PTAL

fluffy (approved):
Just updating my review after all the conversation in the WG. We might need to tweek a bit of this later but I am in favour of merging this more or less as is.

PR #522, opened 2024-09-09
Error code for announce cancel

### LINKED ISSUE #412: ANNOUNCE_CANCEL is missing a code/reason
It's meant to be sent by the subscriber, indicating that the announce is no longer accepted. The fields should mirror ANNOUNCE_ERROR so the subscriber can indicate stuff like authentication expired, another announce was received, etc.

suhasHere:
fixes #412

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1316,12 +1316,19 @@ receiving an ANNOUNCE_CANCEL, it SHOULD close the session as a
 ~~~
 ANNOUNCE_CANCEL Message {
   Track Namespace (tuple),
+  Error Code (i),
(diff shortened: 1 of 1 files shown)


afrind (approved):
lgtm

ianswett (approved):
LG, but I noticed:
"If a publisher receives new subscriptions for that namespace after
receiving an ANNOUNCE_CANCEL, it SHOULD close the session as a
'Protocol Violation'."

I believe that's racy, because the SUBSCRIBE and ANNOUNCE_CANCEL can be inflight at the same time?
Also, shouldn't it be "after sending an ANNOUNCE_CANCEL", or am I misunderstanding the text?

Issue #267, opened 2023-10-08
Graceful session migration


kixelated:
I initially added a recommended session migration algorithm in #264. However it's too prescriptive and non-normative and I removed it. However I do think we should talk about what it takes to gracefully migrate a MoqTransport session, if it's non-obvious and should be in the draft (maybe an appendix), and if we need more mechanisms in the protocol.

This recommended algorithm requires servers to not error on a duplicate ANNOUNCE. It also results in an RTT where duplicate OBJECTs will arrive on both sessions, which is not necessarily ideal. Here's the original text before I removed it:

> ### Graceful {#session-migration-graceful}
> A graceful session migration involves migrating all subscriptions between
> sessions without impeding data flow. The application should be unaware that
> a session migration even occured.
> 
> The client SHOULD establish the new MoqTransport session in the background
> since it may take a few round trips to resolve the address and perform any
> handshakes. Once the the new session is established, any active subscriptions
> and announcements are removed from the old session and issued on the new
> session.
> 
> An endpoint MAY choose to abruptly transition between sessions. However it is
> RECOMMENDED that both sessions are simultaneously active for a brief window
> while subscriptions are individually migrated. The client and server migrate any
> active announcements and subscriptions depending on their role(s) as covered
> below.
> 
> The client terminates the old session once there are no more active
> subscriptions or announcements. The client MAY choose to delay if OBJECTs are
> still being sent/received as queues are drained, but should be prepared to
> receive a GOAWAY Failure from the server if it waits too long.
> 
> #### Publisher
> A publisher needs to migrate all active announcements to the new session. For
> each active announcement:
> 
> 1. The publisher sends an ANNOUNCE on the new session.
> 2. The publisher receives an ANNOUNCE\_OK on the new session.
> 3. The publisher sends an UNANNOUNCE on the old session.
> 4. The publisher receives an ANNOUNCE\_ERROR on the old session, acknowledging
>    the UNANNOUNCE.
> 
> Once these steps are complete, new subscriptions will arrive on the new session
> instead of the old session. The publisher may receive new subscriptions in the
> meantime which should be served normally.
> 
> #### Subscriber
> A subscriber needs to move all active subscriptions to the new session. For each
> active subscription:
> 
> 1. The client sends a SUBSCRIBE message on the new session, starting one past
>    the maximum received object sequence.
> 2. The client sends an UNSUBSCRIBE message on the old session.
> 3. The client receives a SUBSCRIBE\_OK on the new session.
[...]

Issue #225, opened 2023-07-24
Clarify conflicting announces across Producers


suhasHere:
While implementing MOQT relay,  I stumbled upon #204. Thinking through authorization implementation a bit more , I think the current spec needs to clarify this behavior.

Proposal is to say, when a conflicting announce happens,  the Relay MUST reject with an appropriate error type in the ANNOUNCE_ERROR message when the authorization disallows

```
ERROR_CONFLICT_TRACK
```

However, if the authorization policy allows such an allocation of namespaces, then clarify Relay interaction to indicate the most recent ANNOUNCE wins.

afrind:
I'm unclear how this issue is different from #227 - can we merge them or clarify how they are different?

suhasHere:
It can be treated same, if the solution applies to 2 different producers and same producers announcing more than once has the same behavior rules on the relay. 

@afrind do you think we would be fine with that ?

afrind:
Ah, I see.

So this issue is specific to two different producers announce the same namespace (eg: different auth fields in ANNOUNCE), and 227 is the generic case (two ANNOUNCE messages with the same namespace, regardless of auth)?

I'll defer to you and @ianswett if it makes sense to keep separate or merge the discussion.

suhasHere:
After considering few more use-cases, for cases where we need to support make before break, it would be good to have multiple announces be allowed. 

So here is the updated proposal

Proposal is to say, when a conflicting announce happens, the Relay MUST reject with an appropriate error type in the ANNOUNCE_ERROR message when the authorization disallows

``` ERROR_CONFLICT_TRACK ```

However, if the authorization policy allows such an allocation of namespaces, then clarify Relay interaction to indicate that multiple announces for the same namespace is allowed.

afrind:
Individual Comment:

Bikeshed: ERROR_CONFLICT_TRACK -> ERROR_CONFLICTING_TRACK_NAMESPACE

Maybe simplify it to say, 

If a server receives multiple Announce's for the same namespace, it MAY send an Announce Error = ERROR_CONFLICTING_TRACK_NAMESPACE depending on authorization policy.

To clarify: if a server sends this error it doesn't support make-before-break?

kixelated:
I'm borrowing HTTP status codes in my implementation, and yeah we need an equivalent of 409.

ianswett:
Conclusion of the discsusion was to allow identical ANNOUNCEs and provide a varint or token to allow deciding which to prefer.

[...]

Issue #79, opened 2023-01-31
Broadcast publisher reconnect support


kpugin:
In [Model](https://github.com/kixelated/warp-draft/blame/main/draft-lcurley-warp.md#L186) section it says that each Media object is uniquely identified within broadcast, track, etc. I am wondering how reconnect would work - wouldn't it require publisher to maintain state? What happens if one encoder box fails and broadcast needs to be resumed from another box?

fluffy:
What happens in a HTTP CDN today if a cache asks for an file from origin, origin says it is cachable for a long time, then the origin servers crashes when it comes up generates a different file, with the same name. What will happen is pretty broken on different clients get different version of the file.

We have the same issue here, if two clients publish an object with different data but same full track name, group ID, and object ID, then it is undefined which version will get served to any given subscription. This is going to be the nature of any design that provides both aggregation and caching and has unique names.

kixelated:
> I am wondering how reconnect would work - wouldn't it require publisher to maintain state? What happens if one encoder box fails and broadcast needs to be resumed from another box?

The publisher MUST retain state for any reconnect to work. Even something as trivial as group sequence numbers needs to be maintained, otherwise you run into split brain and potential cache poisoning. Consider a player that receives group 69 and then suddenly group 0 (because of a stateless reconnect): it will just throw it out. And of course tracks need to be consistent; the new encoder can't just use different encodings or identification on a whim.

We should absolutely support publishers that can seamlessly continue a broadcast if they can persist the required state. Like @afrind mentioned, I think this would look more like an ETag as a way to specify that the new ANNOUNCE is a continuation of the original resource.

suhasHere:
IOT media publishers will for sure not maintain any state on restarts.   I think a system where we are using static names and maintain caches based on that and expect it to work over restarts is a brittle system and doesn't scale really well where the system is dynamic and networks/connections come and go.

ianswett:
Possibly this can be fixed as part of @fluffy PRs for make before break?

ianswett:
Fixed by #525

PR #525, opened 2024-09-13
Clarify subscribe / annouce interaction

### LINKED ISSUE #362: Add text to clarify relay behavior when there are multiple announces
At interim, to support make before break use-cases, we need to clarify relay behavior when a new announce arrives for existing published.

fluffy:
Fixes #362 #267 #227 #225 #79


draft-ietf-moq-transport.md
```diff
@@ -759,24 +768,47 @@ SUBSCRIBE_DONE, as defined below:
...
   associated with the set of tracks whose Track Namespace matches the
-  announced namespace. Specifics of where the authorization happens,
-  either at the relays or forwarded for further processing, depends on
-  the way the relay is managed and is application specific (typically
-  based on prior business agreement).
+  announced namespace. Specifics of where the authorization and
+  identification of the publisher  happens either at the relays or
+  forwarded for further processing, depends on the way the relay is
+  managed and is application specific (typically based on prior business
+  agreement).
```
afrind, line 799: We should offer a default behavior, like 

> relays MAY have pre-negotiated mechanisms of handling authorization on behalf of publishers.  In the absence of such mechanisms, relays MUST forward SUBSCRIBE requests containing non-empty Auth Info parameters to the original publisher.

draft-ietf-moq-transport.md
```diff
@@ -759,24 +768,47 @@ SUBSCRIBE_DONE, as defined below:
...
 Relays respond with an ANNOUNCE_OK or ANNOUNCE_ERROR control message
 providing the result of announcement. The entity receiving the
 ANNOUNCE MUST send only a single response to a given ANNOUNCE of
-either ANNOUNCE_OK or ANNOUNCE_ERROR.  When a publisher wants to stop
+either ANNOUNCE_OK or ANNOUNCE_ERROR.
+
+A Relay can receive announcements from multiple publishers for the same
+tracknamespace and it MUST respond with appropriate response to each of the
+publishers, in the same way as it would responds when processing ANNOUNCE
+from a single publisher for a given tracknamespace.
```
afrind, line 809: Can relays limit the number of publishers for a given namespace?  It seems like for DoS reasons they can.  Maybe we should specify a minimum: "Relays MUST support at least X publishers announcing the same namespace"

draft-ietf-moq-transport.md
```diff
@@ -788,13 +820,52 @@ match on track namespace unless otherwise negotiated by the application.
 For example, a SUBSCRIBE namespace=foobar message will be forwarded to
 the session that sent ANNOUNCE namespace=foobar.
 
+When a relay receives an incoming SUBSCRIBE for a given namespace, for
+each publisher that has announced that namespace, the relay MUST send a
+SUBSCRIBE to that publisher unless it already has an active subscription
+to that publisher for the full track name in the incoming SUBSCRIBE.
```
afrind, line 826: What about cases where the relay can fill from cache?

draft-ietf-moq-transport.md
```diff
@@ -710,7 +710,16 @@ subscribers for each track. Each new OBJECT belonging to the
...
+track with a SUBSCRIBE_DONE (see {{message-subscribe-done}}). A
+caching relay saves objects to its cache identified by the object's
+fulltrackname, group ID and object ID. Relays MUST be
+ready to process objects for the same fulltrackname from
+multiple publishers and the objects received are forwarded
+to active matching subscribers. If multiple objects are received
+with the same fulltrackname, group ID and object ID - a caching
+Relays MUST drop the received objects as duplicate if there exists
+already an object with the matching fulltrackame, group ID and
+object ID, in its cache.
```
afrind, line 722: Caching and non-caching relays also need to drop (rather than forward) duplicate objects, right?

```
Pub A ---+
         +---> Relay ---> Sub
Pub B ---+
```

I think relays can figure out the best algorithms to do this, we just need to say that the subscriber only receives one copy.

Might want to add an example of why this is useful (graceful failover, redundant publishers, etc)

afrind (commented):
Individual Review

This is really helpful clarification as I didn't necessarily think it had to work this way.

You have a number of random typos etc but just giving high-level feedback.

draft-ietf-moq-transport.md
fluffy: I don't think CDNs are going forward any traffic unless it is both clear who is paying for it and who is responsible for it from a DDoS point of view. We definitely need to get a better handle on that as part of the overall all issue. Glad to remove this sentence if that makes it better but just to talk about it a bit. It seems that if the CDN can look at the track name and say, yep, this is one that I bill for either by policy of by an authenticated annouce for that trackname, then it will likely forward, but if just gets a request for some random track name from some random users, seems very unlikely that it would forward that. 

At the other end of the spectrum from CDNs, a relay running in my home AP gets a a request from the local WAN side it make make complete sense for it to forward that.

draft-ietf-moq-transport.md
fluffy: well ... I suspect this is going to fall in policy for relays region. Lets say you have 10 million viewers of a Taylor Swift livestream that can all publish text comments to various chat tracks. You OK with the number of max publishers on the same namespace being in the millions ?  It seems that different relays will want very different limits here.

draft-ietf-moq-transport.md
fluffy: You mean a fetch of things that are all in the cache ? In this case definitely no subscription needed at all but I assumed that would be handled by the whatever PR eventually deals with fetch.

draft-ietf-moq-transport.md
fluffy: Lets say we had just one publisher, one high available relay, and one subscriber. I don't know how to make an HA Relay that will guarantee to never deliver the same object exactly once even during failures with any reasonable QUIC stack I have today. I think subscribers are going to have to deal with rare events. I don't think we can ask non caching relays to keep track of ever object they have forwarded to stop duplicates. 

What we certainly can do is make sure that most duplicates could be suppress particularly if the duplicated happened within the MAX_CACHE_DURATION for the object. 

I'm pretty on the fence on what to say about non caching relay. The problem really comes down to how much storage space they will use for this ( for both caching and non caching ). Clearly at some point they will free up storage and an that point any MUST is probably going to fail in some cases.   If we proposed the relay MUST cache this for at least MAX_CACHE_DURATION, I am pretty sure CDN implementers will tell us they are not going to do that 100% of the time and it should be a SHOULD. 

This is a long rambling way to get to, yah, you are right. I should have something more here. But I am thinking it will be a SHOULD level recommendation. That work for you ?

fluffy:
> Individual Review
> 
> This is really helpful clarification as I didn't necessarily think it had to work this way.
> 
> You have a number of random typos etc but just giving high-level feedback.

Many thanks on review and any help of fixing all the small stuff (point me too it, or PR on the PR, anything) much appreciated.

draft-ietf-moq-transport.md
afrind: I think a SHOULD will probably work?  Keeping state of everything sent to every client is not scalable, but in some common cases it should be pretty easy (eg within a peep, if you see the same object ID or it goes backwards from the most recent, just drop it).  Implementation experience will help us see what's possible.

I'm thinking of a make-befrore-break case, during the point when the publisher has 2 connections to the relay, it would be nice if the relay is deduplicating as much as it can as it forwards.

draft-ietf-moq-transport.md
afrind: My concern is that without a normative default for what to do, there's very little interoperability.  The only other normative behavior I can think of is "MUST reject any subscribe with authorization info", but that seems less functional.

I 100% agree that CDNs are going to have policies and such that govern this and we should acknowledge that, but it should also work "out of the box".

We will likely revisit some of this when we tackle auth in more detail too.

draft-ietf-moq-transport.md
afrind: I was hoping SUBSCRIBE_NAMESPACE would solve some of that problem -- all 10M viewers have their own namespace `tswift/chat/unique_id`.  We may need to indicate here what limits are likely to exist, because it influences how higher level protocols will use moqt in terms of namespace and track naming.

Agreed the max will be different, but is there a min that we need to specify to get basic interop, eg: for make-before-break, min=2.

draft-ietf-moq-transport.md
afrind: I mean, 2 publishers have announced track T, and published some objects into the cache.  A subscriber issue a subscribe for track T but whatever is specified as the start and/or end is already in cache - there's no need to forward subscribe.  Maybe you can reword this to say - if you would send a subscribe to pub 1, you MUST send it to all N.

afrind (commented):
Individual Comments

draft-ietf-moq-transport.md
wilaw: Rather than drop the later arriving object, I would rather that it overwrite the item existing in cache (i.e latest wins). This allows updating caches and purging corrupt content by overwriting. If we go with the proposal that the original write always wins, then we have no method for publishers to flush content from upstream caches other than waiting for its cache duration to expire.

draft-ietf-moq-transport.md
wilaw: In the absence of such mechanisms, relays MUST forward SUBSCRIBE requests containing non-empty Auth Info parameters to the original publisher.

How is that expected to work if a thousand clients make the same subscription with different auth info to an edge? Normally we would want the edge to coalesce these requests and make a single upstream subscription. But it would now have to make 1000 upstream subscriptions? I don't think that is practical. I think Cullen's proposed text is fine.

afrind: > How is that expected to work if a thousand clients make the same subscription with different auth info to an edge?
 
 I see three possibilities:
 
 1.  pre-negotiated scheme for relay to perform auth
 2. forward the auth credentials to OP for validation 
 3. subscribe error (no way to auth)
 
If we dig into 2, right now, only SUBSCRIBE carries the auth field, so you'd end up blowing up your coalescing, unless you UNSUBSCRIBE duplicates immediately I guess.  An alternate would be to add a new message like AUTH which would perform the auth at the OP but not initiate the actual subscription.   This retains the coalescing while also providing a way for generic relays to do auth.  Maybe we can defer this to a later PR.

ianswett (commented):
This looks great, and clarifies a lot about how one might use ANNOUNCE.  I have one normative comment, which is that I think some of the MUSTs should be a SHOULD.

suhasHere (approved):
This clarifies lot of ambiguities on implementing make before break use-cases.
LGTM

draft-ietf-moq-transport.md
```diff
@@ -710,7 +710,16 @@ subscribers for each track. Each new OBJECT belonging to the
 track within the subscription range is forwarded to each active
 subscriber, dependent on the congestion response. A subscription
 remains active until the publisher of the track terminates the
-track with a SUBSCRIBE_DONE (see {{message-subscribe-done}}).
+subscription with a SUBSCRIBE_DONE (see {{message-subscribe-done}}). A
```
martinduke, line 713: or UNSUBSCRIBE? Or is there a SUBSCRIBE_DONE in reply?

draft-ietf-moq-transport.md
```diff
@@ -759,24 +768,46 @@ SUBSCRIBE_DONE, as defined below:
...
+all active subscriptions to the new relay. The new relay will send a
+response to the subscribes and once this has happened, the subscription
+to the old relay can be stopped with an UNSUBSCRIBE.
+
+
 ## Publisher Interactions
 
 Publishing through the relay starts with publisher sending ANNOUNCE
 control message with a `Track Namespace` ({{model-track}}).
+The annouce allows the relays to know which publisher to forward a
```
martinduke, line 785:
```suggestion
The announce allows the relays to know which publisher to forward a
```

draft-ietf-moq-transport.md
```diff
@@ -788,13 +819,51 @@ match on track namespace unless otherwise negotiated by the application.
 For example, a SUBSCRIBE namespace=foobar message will be forwarded to
 the session that sent ANNOUNCE namespace=foobar.
 
+When a relay receives an incoming SUBSCRIBE for a given namespace, for
+each publisher that has announced that namespace, the relay MUST send a
+SUBSCRIBE to that publisher unless it already has an active subscription
```
martinduke, line 822: Unless the entire range is already cached?

draft-ietf-moq-transport.md
```diff
@@ -788,13 +819,51 @@ match on track namespace unless otherwise negotiated by the application.
...
+to that publisher for the full track name in the incoming SUBSCRIBE.
+
+When a relay receives an incoming ANNOUCE for a given namespace, for
+each active subscription that matches that namespace, it MUST send a
+SUBSCRIBE to that publisher that send the ANNOUCE.
```
martinduke, line 827:
```suggestion
SUBSCRIBE to that publisher that send the ANNOUNCE.
```

draft-ietf-moq-transport.md
```diff
@@ -788,13 +820,52 @@ match on track namespace unless otherwise negotiated by the application.
 For example, a SUBSCRIBE namespace=foobar message will be forwarded to
 the session that sent ANNOUNCE namespace=foobar.
 
+When a relay receives an incoming SUBSCRIBE for a given namespace, for
+each publisher that has announced that namespace, the relay MUST send a
```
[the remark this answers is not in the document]
martinduke: +1

martinduke (changes requested):

draft-ietf-moq-transport.md
fluffy: Going to leave the auth stuff for a later work on auth for the whole draft

draft-ietf-moq-transport.md
fluffy: yah .... let me see what to put here. A bunch of this changes depending on how subscribe is also fetch or if they are separate.

fluffy: I think the text now deals with that

draft-ietf-moq-transport.md
fluffy: I'm adding "or all the downstream subscriptions are ended". I started with just SUBSCRIBE_DONE  but there are also all the timeout and connection close things that can end a subscription. The draft overall is a bit light on all the clean up and timeouts of things and I can't fix that in this PR.

draft-ietf-moq-transport.md
```diff
@@ -759,24 +768,47 @@ SUBSCRIBE_DONE, as defined below:
...
+
+A Relay can receive announcements from multiple publishers for the same
+tracknamespace and it MUST respond with appropriate response to each of the
+publishers, in the same way as it would responds when processing ANNOUNCE
+from a single publisher for a given tracknamespace.
+
+When a publisher wants to stop
 new subscriptions for an announced namespace it sends an UNANNOUNCE.
 A subscriber indicates it will no longer route subscriptions for a
 namespace it previously responded ANNOUNCE_OK to by sending an
```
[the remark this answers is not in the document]
fluffy: I've got no idea :-) We need an error to ANNOUNCE for auth errors, so it makes sense to have an OK just to know there was no error. But I have no idea why we need CANCEL

draft-ietf-moq-transport.md
fluffy: I like that - added in as you sugesteed.

fluffy:
@martinduke - thanks for review. I think I fixed all the stuff you raised. Can you take another look 

@ianswett - handing the baton to you on this one. Thanks you

draft-ietf-moq-transport.md
```diff
@@ -710,7 +710,18 @@ subscribers for each track. Each new OBJECT belonging to the
 track within the subscription range is forwarded to each active
 subscriber, dependent on the congestion response. A subscription
 remains active until the publisher of the track terminates the
-track with a SUBSCRIBE_DONE (see {{message-subscribe-done}}).
+subscription with a SUBSCRIBE_DONE (see {{message-subscribe-done}})
+or all the downstream subscriptions are ended. A
```
martinduke, line 714: The "publisher" is a term local to the session, unless you mean "original publisher." So a subscription ends when the publisher sends SUBSCRIBE_DONE or when the subscriber sends UNSUBSCRIBE. Downstream subscriptions don't enter into it.

draft-ietf-moq-transport.md
```diff
@@ -759,24 +770,46 @@ SUBSCRIBE_DONE, as defined below:
 | 0x6  | Expired                   |
 |------|---------------------------|
 
+### Graceful Publisher Relay Switchover
+
+This section describes behavior a subscriber MAY implement
+to allow for a better user experience when a relay sends a GOAWAY.
+
+When a subscriber receives the GOAWAY message, it starts of the process
```
martinduke, line 778:
```suggestion
When a subscriber receives the GOAWAY message, it starts the process
```

draft-ietf-moq-transport.md
```diff
@@ -759,24 +770,46 @@ SUBSCRIBE_DONE, as defined below:
...
 
 Relays respond with an ANNOUNCE_OK or ANNOUNCE_ERROR control message
 providing the result of announcement. The entity receiving the
 ANNOUNCE MUST send only a single response to a given ANNOUNCE of
-either ANNOUNCE_OK or ANNOUNCE_ERROR.  When a publisher wants to stop
+either ANNOUNCE_OK or ANNOUNCE_ERROR.
+
+A Relay can receive announcements from multiple publishers for the same
+tracknamespace and it MUST respond with appropriate response to each of the
+publishers, in the same way as it would responds when processing ANNOUNCE
```
martinduke, line 809:
```suggestion
publishers, in the same way as it would respond when processing ANNOUNCE
```

martinduke: alternately, "it responds"

draft-ietf-moq-transport.md
```diff
@@ -788,13 +821,52 @@ match on track namespace unless otherwise negotiated by the application.
 For example, a SUBSCRIBE namespace=foobar message will be forwarded to
 the session that sent ANNOUNCE namespace=foobar.
 
+When a relay receives an incoming SUBSCRIBE for a given namespace
+that would cause it to send an upstream subscription, for
+each publisher that has announced that namespace, the relay SHOULD send a
+SUBSCRIBE to that publisher unless it already has an active subscription
+to that publisher for the Full Track Name in the incoming SUBSCRIBE.
```
martinduke, line 828: s/for the Full Track Name/for the objects in the specified range

draft-ietf-moq-transport.md
```diff
@@ -788,13 +821,52 @@ match on track namespace unless otherwise negotiated by the application.
...
+
+When a relay receives an incoming ANNOUCE for a given namespace, for
+each active subscription that matches that namespace, it SHOULD send a
+SUBSCRIBE to that publisher that send the ANNOUNCE.
```
martinduke, line 832: I think this means that if there is an ANNOUNCE for a track where a relay has N incoming subscriptions, it should send N SUBSCRIBEs. That seems to negate the whole purpose of relays?

I hope you actually mean the relay should send *one* SUBSCRIBE assuming the ranges are the same. If not, we are not on the same page at all.

draft-ietf-moq-transport.md
```diff
@@ -788,13 +821,52 @@ match on track namespace unless otherwise negotiated by the application.
 For example, a SUBSCRIBE namespace=foobar message will be forwarded to
 the session that sent ANNOUNCE namespace=foobar.
 
+When a relay receives an incoming SUBSCRIBE for a given namespace
+that would cause it to send an upstream subscription, for
+each publisher that has announced that namespace, the relay SHOULD send a
+SUBSCRIBE to that publisher unless it already has an active subscription
```
afrind, line 827: meta.ai suggests:

When a relay receives an incoming SUBSCRIBE request for a given namespace that would trigger an upstream subscription, it SHOULD send a SUBSCRIBE request to each publisher that has announced that namespace, unless it already has an active subscription to that publisher for the Full Track Name specified in the incoming SUBSCRIBE request.

draft-ietf-moq-transport.md
afrind: Maybe "each active *upstream* subscription" would help clarify?  

I think some of this language will need a second pass once we sort fetch and ranges.

draft-ietf-moq-transport.md
```diff
@@ -788,13 +821,52 @@ match on track namespace unless otherwise negotiated by the application.
...
+to that publisher for the Full Track Name in the incoming SUBSCRIBE.
+
+When a relay receives an incoming ANNOUCE for a given namespace, for
+each active subscription that matches that namespace, it SHOULD send a
```
ianswett, line 831:
```suggestion
each active upstream subscription that matches that namespace, it SHOULD send a
```

draft-ietf-moq-transport.md
ianswett: I added upstream.

ianswett (approved):

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
suhasHere: ANNOUNCE_CANCEL  is used in the below flow IIYC

```mermaid
sequenceDiagram
    Publisher->>Relay(Subscriber): Announce
    Relay(Subscriber)->>Publisher: Announce Ok
    Relay(Subscriber)->>Publisher: Subscribe
    Publisher->>Relay(Subscriber): Subscribe OK
    Publisher->Relay(Subscriber): Objects Flow
    Note right of Relay(Subscriber):  Relays is unable to support new subscriptions 
    Relay(Subscriber)->>Publisher: Announce Cancel 
```

Existing subscriptions will still get served, but new subscriptions will not be sent to the publisher anymore.

draft-ietf-moq-transport.md
fluffy: Thanks - yes, upstream

PR #530, opened 2024-09-18
Rename Peep to Subgroup

martinduke:


ianswett (approved):

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -316,48 +316,48 @@ payload. This includes the underlying encoding, compression, any end-to-end
...
 scheduling of sending data on active streams.
 
-Every object within a Group belongs to exactly one Peep.
+Every object within a Group belongs to exactly one Subgroup.
 
-Objects from two peeps cannot be sent on the same QUIC stream. Objects from the
-same Peep MUST NOT be sent on different QUIC streams, unless one of the streams
-was reset prematurely, or upstream conditions have forced objects from a Peep
+Objects from two subgroups cannot be sent on the same QUIC stream. Objects from the
+same Subgroup MUST NOT be sent on different QUIC streams, unless one of the streams
```
wilaw, line 339: > Objects from two subgroups cannot be sent on the same QUIC stream. Objects from the
> same Subgroup MUST NOT be sent on different QUIC streams
> 

Subgroups in sentence are referred to as both 'Subgroup' and 'subgroup' throughout the PR. The capitalization should be consistent.

ianswett: Agreed, though we're quite inconsistent today on Object, Group, etc.  I think all should be capitalized, but I'm happy to fix capitalization later in a single PR if we agree that they should all be capitalized.

PR #533, opened 2024-09-20
Don't close the control stream

### LINKED ISSUE #532: Clarify "abruptly closed" on control stream
Section 3.3 says:

"The control stream MUST NOT be abruptly closed at the underlying transport layer. Doing so results in the session being closed as a 'Protocol Violation'."

I know we have multiple transports, but it would be good to clarify what an abrupt close is. Presumably any sort of RESET_STREAM, but what happens if there's a FIN? Is this a "clean close" of the session? Does it imply or force a reset of data streams?

ianswett:
Fixes #532

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -513,8 +513,9 @@ sent on unidirectional streams.  Because there are no other uses of
 bidirectional streams, a peer MAY currently close the session as a
 'Protocol Violation' if it receives a second bidirectional stream.
 
-The control stream MUST NOT be abruptly closed at the underlying transport
-layer.  Doing so results in the session being closed as a 'Protocol Violation'.
+The control stream MUST NOT be closed at the underlying transport layer while the
+session is active.  Doing so results in the session being closed as a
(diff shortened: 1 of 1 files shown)


martinduke (approved):
LGTM

LPardue:
Sorry for the late response but I dont think this text is quite correct. If you're running MoQ over WebTransport, you're likely to be following the instructions in https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-10#section-5, which states

> An endpoint that sends a CLOSE_WEBTRANSPORT_SESSION capsule MUST immediately send a FIN. The endpoint MAY send a STOP_SENDING to indicate it is no longer reading from the CONNECT stream. The recipient MUST either close or reset the stream in response. If any additional stream data is received on the CONNECT stream after receiving a CLOSE_WEBTRANSPORT_SESSION capsule, the stream MUST be reset with code H3_MESSAGE_ERROR.

My reading of the current text means that endpoints are likely to blend up treating things as a violation when there is no violation.

Issue #336, opened 2023-11-10
Can something be considered a MoQT relay if it does not terminate QUIC?


afrind:
My understanding is that MoQ is always hop-by-hop encrypted by virtue of working over QUIC or WebTransport, and therefore a "relay" as described in the draft can only inspect Track/Group/Object metadata if it terminates QUIC.  The document should clarify the requirements.

tmengKOP:
I agree with Alan. As a QUIC enthusiast, I guess the current draft implicitly suggests hop-by-hop QUIC termination at relay, e.g., which is necessary to enable cache at relay while having MOQT Message Type and OBJECT metadata as QUIC payload. Yet it is definitely more clarified for readers to explicitly see that information in an early section (e.g., sec 1.1.4).

ianswett:
Given MoQ is over either QUIC or WebTransport, I believe it's impractical to restrict it to only QUIC based transports.

afrind:
> Given MoQ is over either QUIC or WebTransport, I believe it's impractical to restrict it to only QUIC based transports.

The context of the question is whether one can build an MoQ relay that can view object properties without terminating the TLS session underpinning MoQ or WebTransport (eg an untrusted middlebox).

PR #538, opened 2024-09-23
Relays terminate QUIC connections

afrind:
Fixes: #336

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -309,12 +309,12 @@ An Object can become unavailable, but its contents MUST NOT change over
 time.
 
 Objects are comprised of two parts: metadata and a payload.  The metadata is
-never encrypted and is always visible to relays. The payload portion may be
-encrypted, in which case it is only visible to the Original Publisher and End
-Subscribers. The application is solely responsible for the content of the object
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -683,6 +683,9 @@ similar in functionality to Content Delivery Networks
 (CDNs). Additionally, relays serve as policy enforcement points by
 validating subscribe and publish requests at the edge of a network.
 
+Relays are MoQ endpoints, which means they terminate the underlying QUIC
+connections in order to have visibility of MoQ Object metadata.
```
ianswett, line 687:
```suggestion
Relays are MoQ endpoints, which means they terminate the MoQ session
in order to have visibility of MoQ Object metadata.
```

ianswett (approved):
LG, but given Webtransport is also a substrate MoQ runs over, I think we want to avoid using QUIC connection?

suhasHere (approved):

fluffy (approved):
LGTM

ianswett (approved):
LG, but given Webtransport is also a substrate MoQ runs over, I think we want to avoid using QUIC connection?

Issue #542, opened 2024-09-23
Why define a boolean type but be lenient with its usage?


LPardue:
https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#section-1.3-5.4.1 states 

> Indicates that x is a flag and is encoded as a single byte with the value 0 or 1. A value of 0 indicates the flag is false or off, while a value of 1 indicates the flag is true or on. Any other value is a protocol error and SHOULD terminate the session with a Protocol Violation ([Section 3.5](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#session-termination)).

This seems at odds with the QUIC design ethos of being very strict about receiving things that are obviously broken. This matter is wrapped up in https://github.com/moq-wg/moq-transport/issues/536 and the proposed change (https://github.com/moq-wg/moq-transport/pull/543), so I wanted to track the question separately.

I suspect we might want a bigger pass over the recomendations about when to detect a protocol error and how to handle it but I don't want to miss this specific issue in case it was done for a good reason.

LPardue:
I don't think this is editorial, since it affects the expectations on what a receiver is expected to do

ianswett:
To clarify, the proposal is to change the SHOULD terminate to a MUST terminate?

I expect that's what everyone will do anyway, so this SG

LPardue:
Yes that would be my proposal here.

And in addition, a separate review of other fields for consistency.

LPardue:
Discussed in today's interim. MUST rolled into https://github.com/moq-wg/moq-transport/pull/543 and merged. So closing issue as done.

Issue #536, opened 2024-09-20
Extending QUIC notation 


LPardue:
In https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#section-1.3 three new notations are presented thst extend the usual QUIC notations defined in RFC 9000.

Personally, with no hats, I don't see much value in adding all the notation. It introduces overlap and divergence from how we already use the existing notation to convey similar types e.g. the `x (b)` form

> SUBSCRIBE_NAMESPACE_ERROR
{
  Track Namespace Prefix (tuple),
  Error Code (i),
  Reason Phrase (b),
}

vs two separate fields

> CONNECTION_CLOSE Frame {
  Type (i) = 0x1c..0x1d,
  Error Code (i),
  [Frame Type (i)],
  Reason Phrase Length (i),
  Reason Phrase (..),
}

The `x (tuple)`  format seems potentially useful but grates me a bit given its just a group of things, where the group size is restricted to 1..32. It could be represented for example as a type

``` 
Track Namespace Parameters {
  Parameter Count (8) = 1..32,
  Track Namespace Parameter (..) ...,
}

Track Namespace Parameter {
  Track Namespace Paramater Length (i),
  Track Namespace Parameter Value (..),
}
```

Personally, I don't quite think the value of the new things overcomes the additional cognitive load that readers need to carry. There's many other folks that already use or consume RFC 9000 notation. This issue is sort of already present in MoQT, where the `Parameter` type doesn't use the `x (b)` form even though it could to cover the length and value parts.

```
Parameter {
  Parameter Type (i),
  Parameter Length (i),
  Parameter Value (..),
}
```

So one suggestion is to remove the special notations and go back to RFC 9000 style. This shouldn't change the wire encoding at all.

A different suggeation, if I'm in the rough and the MoQ WG has consensus that there is value in extending the notation and dealing with the overlap and cognitive load, it's possible other WGs might too. Having to import MoQT for a notational convention is messy. So it might make sense to spin them out into a separate I-D that could update RFC 9000, and send that over to the QUIC WG to manage.

fluffy:
I don't really care what we do here either but I would like to point at what I suspect is one of the root causes of all the pain around this is given 

```
Track {
   Len (i), 
   Val (..),
  } 
  ''' 
  
It is not possible to write a encoder/parser for that because the length of val may or may not be encoded in Len. You have to have extra information to know how to write an parser for that structure. That seems like a pretty big lacking in any formal syntax.

[...]

PR #543, opened 2024-09-23
Replace x(f) notation with 8-bit field

LPardue:
This is a proposal to illustrate the changes needed to address one aspect of https://github.com/moq-wg/moq-transport/issues/536 .

In creating this, I explicitly didn't state copy the normative action on dealing with a 0 or 1 value.  See https://github.com/moq-wg/moq-transport/issues/542 for that.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1511,7 +1511,8 @@ Values of 0x0 and those larger than 0x2 are a protocol error.
 
 * ContentExists: 1 if an object has been published on this track, 0 if not.
 If 0, then the Largest Group ID and Largest Object ID fields will not be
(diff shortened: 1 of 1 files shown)


ianswett (approved):
This change seems small and sensible, especially since every time I see (f) I think it's a float, not a flag.

suhasHere (approved):

martinduke:
9/25 interim: No objections to this direction, but make it MUST close if not 0 or 1. @ianswett can merge when he's satisfied the discussion has completed.

fluffy (approved):
Don't feel strongly one way or the other but this looks good to me.

ianswett:
I updated this to use MUST.

PR #541, opened 2024-09-23
Expand Type field in control messages

LPardue:
Fixes https://github.com/moq-wg/moq-transport/issues/537

PR #544, opened 2024-09-23
Reduce use of x(b) notation

LPardue:
This is a proposal to illustrate the changes needed to address one aspect of https://github.com/moq-wg/moq-transport/issues/536 .

In creating this, I realised that the [recent change](https://github.com/moq-wg/moq-transport/commit/b5ad15c1f4320271cd515b20e55275efe573256a) to add a length to control messages used the `x(b)` notation. This makes it awkward to then fully expand the full message structure as is done in RFC 9000 and RFC 9114 (and proposed for the `Type` field here on https://github.com/moq-wg/moq-transport/pull/541). Its also arguable if the "followed by that many bytes  of binary data" description for the `x (b)` notation is really supposed to apply to fields that are also formally defined but I think that debate is a distraction.

This PR reduces usage of `x (b)` by adding an explicit length field before the respective value. This reveals that some control message types have a redundant length field, which is rather pointless in some cases. For example, with this change we have

```
GOAWAY Message {
  Length (i),
  New Session URI Length (i),
  New Session URI (..),
}
```

this could be simplified to just 

```
GOAWAY Message {
  Length (i),
  New Session URI (..),
}
```

However, I avoided doing that in this PR because it is a breaking wire format change. 

I'd suggest we do the editorial changes in this PR first, and then survey the control types for opportunities to remove redundant lengths. We could even go so far as rearranging the order of fields so that the variable length ones go at the end, for example HTTP/3 does this with [HEADERS](https://datatracker.ietf.org/doc/html/rfc9114#name-headers) and [PUSH_PROMISE](https://datatracker.ietf.org/doc/html/rfc9114#section-7.2.5)

The last remaining usage of `x (b)` after this change relates to `x (tuple)`. I left that one since I didn't want to conflate things.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -899,7 +899,8 @@ formatted as follows:
 ~~~
 MOQT Control Message {
   Message Type (i),
-  Message Payload (b),
+  Message Length (i),
+  Message Payload (..),
 }
 ~~~
 {: #moq-transport-message-format title="MOQT Message"}
@@ -1060,6 +1061,7 @@ The wire format of the Setup messages are as follows:
 
 ~~~
 CLIENT_SETUP Message Payload {
+  Length (i),
   Number of Supported Versions (i),
   Supported Version (i) ...,
   Number of Parameters (i) ...,
@@ -1067,6 +1069,7 @@ CLIENT_SETUP Message Payload {
 }
 
 SERVER_SETUP Message Payload {
+  Length (i),
   Selected Version (i),
   Number of Parameters (i) ...,
   Setup Parameters (..) ...,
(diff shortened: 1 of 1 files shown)


suhasHere:
I am fine with moving away from (b) as that was the original structure of the messages. 
I would see if we can combine this change with Type change as well to handle all the control message changes in one place.

ianswett (approved):
I'm happy with this direction as long as the WG agrees.

fluffy (approved):
I don't feel strongly one way or the other but all looks good to me.

martinduke:
25 Sep 2024 interim: No objections to this direction. @ianswett Merge when github discussion has petered out.

LPardue:
closing in favor of the mega PR that people asked for: https://github.com/moq-wg/moq-transport/pull/550

Issue #545, opened 2024-09-24
Error code definitions use proper English


LPardue:
There are several error cocdes defined in MoQT which state a code and a reason, that is proper English. This contrasts with the QUIC and HTTP/3 specs, which define error codes using Upper Case format. E.g. "Foo Bar" vs. "FOO_BAR". 

This is purely an editorial thing but given that people will probably be cross referencing all these specs, the deviation in MoQT seems unnecessary.

fluffy:
+1 on using the names that don't have spaces and easily translate into ENUM values in code.

PR #546, opened 2024-09-24
Give codepoints UPPER_CASE names

LPardue:
Not 100% sure if we want to go the whole hog, but this addresses the issue I raise on #545.

Importantly, there's a few things that have naming clashes, which I solved by adding a long prefix. Something to bash on.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -504,18 +504,18 @@ the peers exchange Setup messages ({{message-setup}}).  All messages defined in
 this draft except OBJECT and OBJECT_WITH_LENGTH are sent on the control stream
 after the Setup message. Control messages MUST NOT be sent on any other stream,
 and a peer receiving a control message on a different stream closes the session
-as a 'Protocol Violation'. Objects MUST NOT be sent on the control stream, and a
-peer receiving an Object on the control stream closes the session as a 'Protocol
-Violation'.
+as a SESSION_PROTOCOL_VIOLATION. Objects MUST NOT be sent on the control stream,
+and a peer receiving an Object on the control stream closes the session as a
+SESSION_PROTOCOL_VIOLATION.
 
 This draft only specifies a single use of bidirectional streams. Objects are
 sent on unidirectional streams.  Because there are no other uses of
 bidirectional streams, a peer MAY currently close the session as a
-'Protocol Violation' if it receives a second bidirectional stream.
+SESSION_PROTOCOL_VIOLATION if it receives a second bidirectional stream.
 
 The control stream MUST NOT be closed at the underlying transport layer while the
 session is active.  Doing so results in the session being closed as a
-'Protocol Violation'.
+SESSION_PROTOCOL_VIOLATION.
 
 ## Stream Cancellation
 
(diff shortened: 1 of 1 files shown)


suhasHere:
I like this change and upper case naming convention.

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -504,18 +504,18 @@ the peers exchange Setup messages ({{message-setup}}).  All messages defined in
 this draft except OBJECT and OBJECT_WITH_LENGTH are sent on the control stream
 after the Setup message. Control messages MUST NOT be sent on any other stream,
 and a peer receiving a control message on a different stream closes the session
-as a 'Protocol Violation'. Objects MUST NOT be sent on the control stream, and a
-peer receiving an Object on the control stream closes the session as a 'Protocol
-Violation'.
+as a SESSION_PROTOCOL_VIOLATION. Objects MUST NOT be sent on the control stream,
```
ianswett, line 507: nit: Isn't it a bit redundant to close the session with a SESSION_PROTOCOL_VIOLATION?  I think PROTOCOL_VIOLATION might be sufficient?

Similar comment for the other error codes, where I don't think SESSION_ is necessary

draft-ietf-moq-transport.md
```diff
@@ -535,45 +535,42 @@ Section 5}}).
...
-* Duplicate Track Alias: The endpoint attempted to use a Track Alias
-  that was already in use.
-
-* Too Many Subscribes: The session was closed because the subscriber used
-  a Subscribe ID equal or larger than the current Maximum Subscribe ID.
-
-* GOAWAY Timeout: The session was closed because the client took too long to
-  close the session in response to a GOAWAY ({{message-goaway}}) message.
-  See session migration ({{session-migration}}).
+SESSION_NO_ERROR (0x0):
```
ianswett, line 538: When using native QUIC, are we currently conflicting with QUIC error codes?

draft-ietf-moq-transport.md
```diff
@@ -738,42 +735,62 @@ request is cached and shared among the pending subscribers.
...
-|------|---------------------------|
-| 0x2  | Retry Track Alias         |
-|------|---------------------------|
-| 0x3  | Track Does Not Exist      |
-|------|---------------------------|
-| 0x4  | Unauthorized              |
-|------|---------------------------|
-| 0x5  | Timeout                   |
-|------|---------------------------|
+SUBSCRIBE_ERROR_INTERNAL_ERROR (0x0):
```
ianswett, line 738: It'd be nice to not have the 'ERROR_'  in the names, but I see you're trying to distinguish them from 'STATUS_'.  The fact we have two sets of error codes may be an error, which this PR is pointing out?

vasilvv (approved):

ianswett (approved):
Some nits, but this looks like the right direction.

draft-ietf-moq-transport.md
suhasHere: agree  with @ianswett  . Looks like these needs to be collapsed

draft-ietf-moq-transport.md
LPardue: the text above says

> When native QUIC is used, the session is closed using the CONNECTION\_CLOSE frame ({{QUIC, Section 19.19}}).

It should probably be clarified that this is an application-layer close, and that MoQT over native QUIC is an application mapping that defines its own error code registry (like HTTP/3 does in https://datatracker.ietf.org/doc/html/rfc9114#name-http-3-error-codes)

In other words, the QUIC errors are in the "transport namespace" and so there is no conflict.

draft-ietf-moq-transport.md
LPardue: Totally agree. I was very confused by these two different similar but different error code sets. On https://github.com/moq-wg/moq-transport/pull/546/files#r1773547847 I also highlight that differentiating SESSION and SUBSCRIBE is annoying. Maybe we just want a unified space?

afrind: I don't have a strong opinion, though a downside of a unified space is that you can receive an error code that has no meaning in a particular context.

LPardue: Yes that's true. But the mantra in QuIc and H3 has been that you can't always expect the peer to be able to (or want to) use the error code that the RFC states.

I'm not familiar enough with MoQ to know what it's mantra wants to be

fluffy (approved):
I have wanted this for a long time. +1

martinduke:
25 Sep 2024 Interim: do another pass consolidating into a single error space, then @ianswett merge after discussion wanes.

draft-ietf-moq-transport.md
TimEvens: Isn't it that there is only one/single bidirectional stream used solely for control messages and unidirectional streams only for data objects? Why would the implementation even process a control message over a unidirectional stream? IMO, this should be reworded a bit.... 

It's a `DATA_STREAM_VIOLATION` if any other message type is received (regardless of control or other stream header/data object header) over a unidirectional stream that isn't expected based on the stream mode in progress.  

If a new bidir stream is created or if data objects/stream headers are received over the control bidir stream, it would make sense to be a  protocol violation that would result in session/connection close.  

My two-cents are that if a data (aka unidirectional) stream violates something, such as a bad/malformed message header (e.g., believed to be control message type erroneously), it shouldn't result in a connection/session close. It should only result in the unidirectional stream being closed.  A use-case for this is a forced reset of a stream to transition to a new stream.

draft-ietf-moq-transport.md
TimEvens: +1. We currently map these values as the QUIC application error code.

draft-ietf-moq-transport.md
LPardue: I'd rather track adjusting the "when" of errors in a separate issue. This PR is already going to be hard to review, even when it focuses on the "what" of the error codes names/values.

afrind:
This PR is useful but overtaken by events.  Closing for now, but keeping the issue open to make another attempt after the document re-org.

PR #562, opened 2024-09-30
remove example that is incoherent

fluffy:


ianswett (approved):
I tend to agree, but in this case, I'm leaning towards doing this as part of whatever larger change we make.

But happy to merge this now while we wait, depending on what others think?

fluffy:
After the discussion at the interim, I think this is not even the tip of the iceberg on confusion on priority. This could be a snowflake sitting on top of the iceberg. I'm going to just close this and it can get fixed in the PR that clarifies priority.

PR #565, opened 2024-09-30
rephrase object ID ordering to match what a relay can implement

fluffy:


suhasHere (approved):

ianswett (commented):
I think we really want to be sending Objects in increasing Object ID order within a stream, so I'm unsure why remove the normative MUST here?

PR #567, opened 2024-09-30
strawman for when to close streams

fluffy:


draft-ietf-moq-transport.md
```diff
@@ -1926,8 +1926,20 @@ session with a Protocol Violation.
 A publisher SHOULD NOT open more than one stream at a time with the same stream
 header message type and fields.
 
+Stream SHOULD be closed shortly after it is likely no more data will be
```
TimEvens, line 1929: I believe you are suggesting FIN close of stream. What happens with RST? Can we use that instead of FIN?

draft-ietf-moq-transport.md
```diff
@@ -1926,8 +1926,20 @@ session with a Protocol Violation.
...
+send on them. Indications of this include:
+
+* Receiving an object status of "end of subgroup" indicates streams for
+  that subgroup could be closed.
+
+* Receiving an object status with "end of group" indicates all streams
+  for that group and its subgroup can be closed.
```
afrind, line 1936: With subgroups and reordering this seems too aggressive.

afrind (commented):
Individual Comment: 

I also made a different proposal here: https://github.com/moq-wg/moq-transport/issues/442#issuecomment-2380048334

draft-ietf-moq-transport.md
martinduke: I think it's fine to do this if all objects with an ID lower than end-of-group have been received.

draft-ietf-moq-transport.md
```diff
@@ -1926,8 +1926,20 @@ session with a Protocol Violation.
...
+
+* Receiving an object status with "end of track" indicates all streams
+  for this track, its groups, and sub groups, can be closed.
```
martinduke, line 1939: likewise, assuming that all previous sequence numbers have arrived.

martinduke (commented):
As an individual:

I would like to wordsmith something involving stream FIN, but it doesn't have to be in this PR.

draft-ietf-moq-transport.md
fluffy: Good point - the more I thought about it, end of group is really so the subscriber can know it has rendered the last frame in group X before starting rendering frames for group X+1. It should not be used for this. The stream is  for a sub group so end of sub group is what should be the indicator to close the stream. I will update.

draft-ietf-moq-transport.md
fluffy: fixed in updated PR

draft-ietf-moq-transport.md
fluffy: fixed in updated PR

draft-ietf-moq-transport.md
```diff
@@ -1926,8 +1934,15 @@ session with a Protocol Violation.
 A publisher SHOULD NOT open more than one stream at a time with the same stream
 header message type and fields.
 
+Stream SHOULD be closed (using a QUIC CONNECTION_CLOSE frame) shortly
```
martinduke, line 1937: CONNECTION_CLOSE is used for connections, not streams.

The ways to close a stream are FIN and RESET_STREAM/RESET_STREAM_AT.

I would argue the sender should use FIN if the subgroup is complete, RESET_STREAM_AT otherwise.

ianswett: Yes, streams should typically be closed with a fin, and if not, a reset.

ianswett:
Discussion at the virtual interim on Oct 16th indicated interest in clarification about when to use Fin, Reset, etc, with interest in @martinduke writing a PR.

fluffy:
I'll leave this open for a bit but I expect we will use #589 and just ignore this.

ianswett:
Closing this now that #589 has been merged.

PR #581, opened 2024-10-07
Fetch Objects if you can

### LINKED ISSUE #350: When to use Track Alias or Subscribe ID?

### LINKED ISSUE #368: Subscribe vs Fetch and Object states

### LINKED ISSUE #476: SUBSCRIBE with absolute ranges and flow control

### LINKED ISSUE #482: Challenges with being too vague on `subscribe_id` and `track_alias`

### LINKED ISSUE #490: QUIC flow control does not map cleanly to MoQ subscriptions

### LINKED ISSUE #578: Need to add SubGroupID to the data model and make it cacheable
Make the SubGroupID something that is stored in cache 

Make sure it is transported by FETCH data stream

ianswett:
Attempt to convert the consensus from the recent hybrid interim on October 2nd/3rd into a PR with the help of @suhasHere 

There's a lot more editorial cleanup to do, such as the priorities section, because many things now need to say "subscriptions or fetches", etc.

It does not do some things we may want to do later:
 * add support for multiple ranges
 * add a FETCH_UPDATE
 * add a MAX_FETCH_ID or equivalent
 * Put fetches on their own **bidirectional** stream

Fixes #350 
Fixes #368
Fixes #476
Fixes #482
Fixes #490 
Fixes #578

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2183,12 +2183,7 @@ FETCH_HEADER Message {
 {: #fetch-header-format title="MOQT FETCH_HEADER Message"}
 
 
-When a publisher sends objects in response to a FETCH request, it finds the
-stream that is associated with the fetch request, or opens a new one and sends
-the `FETCH_HEADER`, and then serializes the following object fields. Each object
-sent on a fetch stream after the header has the following format:
-
-The Object Status field is only sent if the Object Payload Length is zero.
+Each object sent on a fetch stream after the FETCH_HEADER has the following format:
 
 ~~~
 {
@@ -2202,6 +2197,8 @@ The Object Status field is only sent if the Object Payload Length is zero.
 }
 ~~~
 {: #object-fetch-format title="MOQT Fetch Object Fields"}
+
+The Object Status field is only sent if the Object Payload Length is zero.
 
 
 ## Examples
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1248,7 +1248,7 @@ SUBSCRIBE Message {
   Filter Type (i),
   [StartGroup (i),
    StartObject (i)],
-  Group (i),
+  [EndGroup (i),
    EndObject (i)],
   Number of Parameters (i),
   Subscribe Parameters (..) ...
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1201,12 +1201,16 @@ GOAWAY Message {
 
 ## SUBSCRIBE {#message-subscribe-req}
 
-A subscription causes the publisher to send newly published objects for a track
-and can start at either the latest Group or latest Object. A subscriber MUST NOT
-make multiple active subscriptions for a track within a single session and
-publishers SHOULD treat this as a protocol violation.
+A subscription causes the publisher to send newly published objects for a track.
+A subscriber MUST NOT make multiple active subscriptions for a track within a
+single session and publishers SHOULD treat this as a protocol violation.
 
-There are 2 types of subscriptions:
+**Filter Types**
+
+The subscriber specifies a filter on the subscription to allow
+the publisher to identify which objects need to be delivered.
+
+There are 4 types of filters:
 
 Latest Group (0x1) : Specifies an open-ended subscription with objects
 from the beginning of the current group.  If no content has been delivered yet,
@@ -1216,9 +1220,15 @@ Latest Object (0x2): Specifies an open-ended subscription beginning from
 the current object of the current group.  If no content has been delivered yet,
 the subscription starts with the first published or received group.
 
-A subscribe type other than the above MUST be treated as error.
+AbsoluteStart (0x3):  Specifies an open-ended subscription beginning
+from the object identified in the StartGroup and StartObject fields.
 
-A subscriber issues a SUBSCRIBE to a publisher to request a track.
+AbsoluteRange (0x4):  Specifies a closed subscription starting at StartObject
+in StartGroup and ending at EndObject in EndGroup.  The start and end of the
+range are inclusive.  EndGroup and EndObject MUST specify the same or a later
+object than StartGroup and StartObject.
+
+A filter type other than the above MUST be treated as error.
 
 The format of SUBSCRIBE is as follows:
 
@@ -1286,9 +1296,9 @@ On successful subscription, the publisher MUST reply with a SUBSCRIBE_OK,
 allowing the subscriber to determine the start group/object when not explicitly
 specified and the publisher SHOULD start delivering objects.
 
-If a publisher cannot satisfy the requested start or end for the subscription it
-MAY send a SUBSCRIBE_ERROR with code 'Invalid Range'. A publisher MUST NOT send
-objects from outside the requested start and end.
+If a publisher cannot satisfy the requested start or end or if the end has
+already been published it SHOULD send a SUBSCRIBE_ERROR with code 'Invalid Range'.
+A publisher MUST NOT send objects from outside the requested start and end.
 
 ## SUBSCRIBE_UPDATE {#message-subscribe-update-req}
 
@@ -1381,17 +1391,9 @@ stream that is used to send the Objects.
 
 The Object Forwarding Preference does not apply to fetches.
 
-There are 2 types of fetches:
-
-AbsoluteStart (0x3):  Specifies an open-ended fetch beginning
-from the object identified in the StartGroup and StartObject fields.
-
-AbsoluteRange (0x4):  Specifies a closed fetch starting at StartObject
-in StartGroup and ending at EndObject in EndGroup.  The start and end of the
-range are inclusive.  EndGroup and EndObject MUST specify the same or a later
-object than StartGroup and StartObject.
-
-A 'Fetch Type' other than the above MUST be treated as error.
+Fetch specifies an inclusive range of Objects starting at StartObject
+in StartGroup and ending at EndObject in EndGroup. EndGroup and EndObject MUST
+specify the same or a later object than StartGroup and StartObject.
 
 The format of FETCH is as follows:
 
@@ -1405,7 +1407,6 @@ FETCH Message {
   Track Name (..),
   Subscriber Priority (8),
   Group Order (8),
-  Fetch Type (i),
   StartGroup (i),
   StartObject (i),
   [EndGroup (i),
@@ -1416,8 +1417,8 @@ FETCH Message {
 ~~~
 {: #moq-transport-fetch-format title="MOQT FETCH Message"}
 
-* Fetch ID: The Fetch ID identifies a given fetch request. Fetch ID is a
-variable length integer that MUST be unique and monotonically increasing
+* Subscribe ID: The Subscribe ID identifies a given fetch request. Subscribe ID
+is a variable length integer that MUST be unique and monotonically increasing
 within  a session.
 
 * Track Namespace: Identifies the namespace of the track as defined in
@@ -1434,17 +1435,14 @@ Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
 A value of 0x0 indicates the original publisher's Group Order SHOULD be
 used. Values larger than 0x2 are a protocol error.
 
-* Filter Type: Identifies the type of filter, which also indicates whether
-the StartGroup/StartObject fields will be present.
-
 * StartGroup: The start Group ID.
 
 * StartObject: The start Object ID.
 
-* EndGroup: The end Group ID. Only present for the "AbsoluteRange" filter type.
+* EndGroup: The end Group ID.
 
 * EndObject: The end Object ID, plus 1. A value of 0 means the entire group is
-requested. Only present for the "AbsoluteRange" filter type.
+requested.
 
 * Parameters: The parameters are defined in {{version-specific-params}}.
 
@@ -1465,7 +1463,7 @@ subgroup ID is not used for ordering.
 ## FETCH_CANCEL {#message-fetch-cancel}
 
 A subscriber issues a `FETCH_CANCEL` message to a publisher indicating it is no
-longer interested in receiving Objects for the fetch specified by 'Fetch ID'.
+longer interested in receiving Objects for the fetch specified by 'Subscribe ID'.
 The publisher SHOULD close the unidirectional stream as soon as possible.
 
 The format of `FETCH_CANCEL` is as follows:
@@ -1474,12 +1472,12 @@ The format of `FETCH_CANCEL` is as follows:
 FETCH_CANCEL Message {
   Type (i) = 0x17,
   Length (i),
-  Fetch ID (i)
+  Subscribe ID (i)
 }
 ~~~
 {: #moq-transport-fetch-cancel title="MOQT FETCH_CANCEL Message"}
 
-* Fetch ID: Subscription Identifier as defined in {{message-fetch}}.
+* Subscribe ID: Subscription Identifier as defined in {{message-fetch}}.
 
 
 ## ANNOUNCE_OK {#message-announce-ok}
@@ -1724,7 +1722,7 @@ FETCH_OK
 {
   Type (i) = 0x18,
   Length (i),
-  Fetch ID (i),
+  Subscribe ID (i),
   Group Order (8),
   End Of Track (8),
   Largest Group ID (i),
@@ -1735,7 +1733,7 @@ FETCH_OK
 ~~~
 {: #moq-transport-fetch-ok format title="MOQT FETCH_OK Message"}
 
-* Fetch ID: Fetch Identifier as defined in {{message-fetch}}.
+* Subscribe ID: Fetch Identifier as defined in {{message-fetch}}.
 
 * Group Order: Indicates the fetch will be delivered in
 Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
@@ -1763,7 +1761,7 @@ FETCH_ERROR
 {
   Type (i) = 0x19,
   Length (i),
-  Fetch ID (i),
+  Subscribe ID (i),
   Error Code (i),
   Reason Phrase Length (i),
   Reason Phrase (..),
@@ -1771,7 +1769,7 @@ FETCH_ERROR
 ~~~
 {: #moq-transport-fetch-error format title="MOQT FETCH_ERROR Message"}
 
-* Fetch ID: Subscription Identifier as defined in {{message-subscribe-req}}.
+* Subscribe ID: Subscription Identifier as defined in {{message-subscribe-req}}.
 
 * Error Code: Identifies an integer error code for fetch failure.
 
@@ -2203,11 +2201,11 @@ previously sent Object ID within a given group in that stream.
 ### Fetch Header {#fetch-header}
 
 When a stream begins with `FETCH_HEADER`, all objects on the stream belong to the
-track requested in the Fetch message identified by `Fetch ID`.
+track requested in the Fetch message identified by `Subscribe ID`.
 
 ~~~
 FETCH_HEADER Message {
-  Fetch ID (i),
+  Subscribe ID (i),
   Publisher Priority (8),
 }
 ~~~
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1201,12 +1201,12 @@ GOAWAY Message {
 
 ## SUBSCRIBE {#message-subscribe-req}
 
-### Filter Types {#sub-filter}
+A subscription causes the publisher to send newly published objects for a track
+and can start at either the latest Group or latest Object. A subscriber MUST NOT
+make multiple active subscriptions for a track within a single session and
+publishers SHOULD treat this as a protocol violation.
 
-The subscriber specifies a filter on the subscription to allow
-the publisher to identify which objects need to be delivered.
-
-There are 2 types of filters:
+There are 2 types of subscriptions:
 
 Latest Group (0x1) : Specifies an open-ended subscription with objects
 from the beginning of the current group.  If no content has been delivered yet,
@@ -1216,10 +1216,8 @@ Latest Object (0x2): Specifies an open-ended subscription beginning from
 the current object of the current group.  If no content has been delivered yet,
 the subscription starts with the first published or received group.
 
-A filter type other than the above MUST be treated as error.
+A subscribe type other than the above MUST be treated as error.
 
-
-### SUBSCRIBE Format
 A subscriber issues a SUBSCRIBE to a publisher to request a track.
 
 The format of SUBSCRIBE is as follows:
@@ -1419,7 +1417,7 @@ FETCH Message {
 ~~~
 {: #moq-transport-fetch-format title="MOQT FETCH Message"}
 
-* Fetch ID: The Fetch ID identifies a given fetch request. Fetch ID is a 
+* Fetch ID: The Fetch ID identifies a given fetch request. Fetch ID is a
 variable length integer that MUST be unique and monotonically increasing
 within  a session.
 
@@ -1429,7 +1427,7 @@ within  a session.
 * Track Name: Identifies the track name as defined in ({{track-name}}).
 
 * Priority: Specifies the priority of a fetch request relative to
-other subscriptions or fetches in the same session. Lower numbers get higher 
+other subscriptions or fetches in the same session. Lower numbers get higher
 priority. See {{priorities}}.
 
 * Group Order: Allows the subscriber to request Objects be delivered in
@@ -1481,7 +1479,7 @@ FETCH_CANCEL Message {
 ~~~
 {: #moq-transport-unsubscribe-format title="MOQT FETCH_CANCEL Message"}
 
-* Fetch ID: Subscription Identifier as defined in {{message-fetch-req}}.
+* Fetch ID: Subscription Identifier as defined in {{message-fetch}}.
 
 
 ## ANNOUNCE_OK {#message-announce-ok}
@@ -2095,7 +2093,6 @@ will be dropped.
 
 ~~~
 OBJECT_DATAGRAM Message {
-  Subscribe ID (i),
   Track Alias (i),
   Group ID (i),
   Object ID (i),
@@ -2129,7 +2126,6 @@ stream header.
 
 ~~~
 STREAM_HEADER_TRACK Message {
-  Subscribe ID (i)
   Track Alias (i),
   Publisher Priority (8),
 }
@@ -2167,7 +2163,6 @@ ID` and the subgroup indicated by 'Group ID' and `Subgroup ID`.
 
 ~~~
 STREAM_HEADER_SUBGROUP Message {
-  Subscribe ID (i),
   Track Alias (i),
   Group ID (i),
   Subgroup ID (i),
@@ -2238,7 +2233,6 @@ Sending a track on one stream:
 
 ~~~
 STREAM_HEADER_TRACK {
-  Subscribe ID = 1
   Track Alias = 1
   Publisher Priority = 0
 }
@@ -2262,7 +2256,6 @@ Sending a subgroup on one stream:
 Stream = 2
 
 STREAM_HEADER_SUBGROUP {
-  Subscribe ID = 2
   Track Alias = 2
   Group ID = 0
   Subgroup ID = 0
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1268,7 +1268,6 @@ used. Values larger than 0x2 are a protocol error.
 
 * Filter Type: Identifies the type of filter, which also indicates whether
 the StartGroup/StartObject and EndGroup/EndObject fields will be present.
-See ({{sub-filter}}).
 
 * StartGroup: The start Group ID. Only present for "AbsoluteStart" and
 "AbsoluteRange" filter types.
@@ -1477,7 +1476,7 @@ FETCH_CANCEL Message {
   Fetch ID (i)
 }
 ~~~
-{: #moq-transport-unsubscribe-format title="MOQT FETCH_CANCEL Message"}
+{: #moq-transport-fetch-cancel title="MOQT FETCH_CANCEL Message"}
 
 * Fetch ID: Subscription Identifier as defined in {{message-fetch}}.
 
@@ -1732,7 +1731,7 @@ FETCH_OK
   Subscribe Parameters (..) ...
 }
 ~~~
-{: #moq-transport-subscribe-ok format title="MOQT FETCH_OK Message"}
+{: #moq-transport-fetch-ok format title="MOQT FETCH_OK Message"}
 
 * Fetch ID: Fetch Identifier as defined in {{message-fetch}}.
 
@@ -1764,7 +1763,7 @@ FETCH_ERROR
   Reason Phrase (..),
 }
 ~~~
-{: #moq-transport-subscribe-error format title="MOQT FETCH_ERROR Message"}
+{: #moq-transport-fetch-error format title="MOQT FETCH_ERROR Message"}
 
 * Fetch ID: Subscription Identifier as defined in {{message-subscribe-req}}.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1701,6 +1701,64 @@ SUBSCRIBE_ERROR
   ({{session-termination}}).
 
 
+## FETCH_OK {#message-fetch-ok}
+
+A publisher sends a FETCH control message for successful fetches.
+
+~~~
+FETCH_OK
+{
+  Type (i) = 0x18,
+  Length (i),
+  Fetch ID (i),
+  Group Order (8),
+  Largest Group ID (i),
+  Largest Object ID (i),
+  Number of Parameters (i),
+  Subscribe Parameters (..) ...
+}
+~~~
+{: #moq-transport-subscribe-ok format title="MOQT FETCH_OK Message"}
+
+* Fetch ID: Fetch Identifier as defined in {{message-fetch}}.
+
+* Group Order: Indicates the fetch will be delivered in
+Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
+Values of 0x0 and those larger than 0x2 are a protocol error.
+
+* Largest Group ID: The largest Group ID available for this track. This field
+is only present if ContentExists has a value of 1.
+
+* Largest Object ID: The largest Object ID available within the largest Group ID
+for this track. This field is only present if ContentExists has a value of 1.
+
+* Subscribe Parameters: The parameters are defined in {{version-specific-params}}.
+
+## FETCH_ERROR {#message-fetch-error}
+
+A publisher sends a FETCH_ERROR control message in response to a
+failed FETCH.
+
+~~~
+FETCH_ERROR
+{
+  Type (i) = 0x19,
+  Length (i),
+  Fetch ID (i),
+  Error Code (i),
+  Reason Phrase Length (i),
+  Reason Phrase (..),
+}
+~~~
+{: #moq-transport-subscribe-error format title="MOQT FETCH_ERROR Message"}
+
+* Fetch ID: Subscription Identifier as defined in {{message-subscribe-req}}.
+
+* Error Code: Identifies an integer error code for fetch failure.
+
+* Reason Phrase: Provides the reason for fetch error.
+
+
 ## SUBSCRIBE_DONE {#message-subscribe-done}
 
 A publisher sends a `SUBSCRIBE_DONE` message to indicate it is done publishing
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1206,18 +1216,8 @@ Latest Object (0x2): Specifies an open-ended subscription beginning from
 the current object of the current group.  If no content has been delivered yet,
 the subscription starts with the first published or received group.
 
-AbsoluteStart (0x3):  Specifies an open-ended subscription beginning
-from the object identified in the StartGroup and StartObject fields.
+A subscribe type other than the above MUST be treated as error.
```
afrind, line 1219: Did we want to handle subscribe changes in a separate PR?

draft-ietf-moq-transport.md
```diff
@@ -1371,6 +1366,121 @@ UNSUBSCRIBE Message {
...
 * Subscribe ID: Subscription Identifier as defined in {{message-subscribe-req}}.
 
+
+## FETCH {#message-fetch}
+
+A subscriber issues a FETCH to a publisher to request a range of already published
+objects within a track. The publisher responding to a FETCH is responsible for
+retrieving all available Objects, and if there are gaps between Objects, it
+indicates these Objects are not avaiable. A fetch MUST NOT omit sending objects
+due to congestion, like a subscription can.
```
afrind, line 1376: Is the last sentence is needed? The previous sentence says you fetch them all, and if fetch omits an object it will be considered gone by the receiver.

draft-ietf-moq-transport.md
```diff
@@ -1371,6 +1366,121 @@ UNSUBSCRIBE Message {
...
+message.  If it responds with FETCH_OK, the publisher creates a new unidirectional
+stream that is used to send the Objects.
+
+The Object Forwarding Preference does not apply to fetches.
+
+There are 2 types of fetches:
+
+AbsoluteStart (0x3):  Specifies an open-ended fetch beginning
+from the object identified in the StartGroup and StartObject fields.
+
```
afrind, line 1388: Wasn't the consensus that Fetch requires an endpoint in the request?  You can do HTTP GET Range style and ask for 1-99999999.

draft-ietf-moq-transport.md
```diff
@@ -1371,6 +1366,121 @@ UNSUBSCRIBE Message {
...
+* StartGroup: The start Group ID.
+
+* StartObject: The start Object ID.
+
+* EndGroup: The end Group ID. Only present for the "AbsoluteRange" filter type.
+
+* EndObject: The end Object ID, plus 1. A value of 0 means the entire group is
+requested. Only present for the "AbsoluteRange" filter type.
+
+* Parameters: The parameters are defined in {{version-specific-params}}.
```
afrind, line 1449: Which params are valid here?  Just AUTH?

draft-ietf-moq-transport.md
```diff
@@ -1371,6 +1366,121 @@ UNSUBSCRIBE Message {
...
+
+Any Objects that are not yet published will not be retrieved by a FETCH.
+The latest available Object is indicated in the FETCH_OK, and is the last
+Object a fetch will return if the 'Fetch Type is AbsoluteStart or if the
+type is AbsoluteRange and the EndGroup and EndObject have not yet been
+published.
+
+If StartGroup/StartObject is greater than the latest published Object group,
+the publisher MUST return FETCH_ERROR with error code 'No Objects'. A publisher
+MUST send fetched objects in ascending order and MUST NOT send objects from
```
afrind, line 1460: Should we describe object ordering in more detail (eg how to use group ID, subgroup ID and object ID)?

Also, what about group order (ascending / descending)?

draft-ietf-moq-transport.md
```diff
@@ -1371,6 +1366,121 @@ UNSUBSCRIBE Message {
...
+requested. Only present for the "AbsoluteRange" filter type.
+
+* Parameters: The parameters are defined in {{version-specific-params}}.
+
+
+Any Objects that are not yet published will not be retrieved by a FETCH.
+The latest available Object is indicated in the FETCH_OK, and is the last
+Object a fetch will return if the 'Fetch Type is AbsoluteStart or if the
+type is AbsoluteRange and the EndGroup and EndObject have not yet been
+published.
```
afrind, line 1456: This reads somewhat awkwardly.  Isn't the object specified in fetch OK is the last one you will get?

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1714,64 @@ SUBSCRIBE_ERROR
...
+{
+  Type (i) = 0x18,
+  Length (i),
+  Fetch ID (i),
+  Group Order (8),
+  Largest Group ID (i),
+  Largest Object ID (i),
+  Number of Parameters (i),
+  Subscribe Parameters (..) ...
+}
```
afrind, line 1732: Didn't the consensus include an indicator for "This range extends to the end of the track"?

draft-ietf-moq-transport.md
```diff
@@ -1371,6 +1366,121 @@ UNSUBSCRIBE Message {
...
+
+A 'Fetch Type' other than the above MUST be treated as error.
+
+The format of FETCH is as follows:
+
+~~~
+FETCH Message {
+  Type (i) = 0x16,
+  Length (i),
+  Subscribe ID (i),
```
afrind, line 1402: Did you mean to say Fetch ID here?  Re-using the subscribe ID space is interesting though.

draft-ietf-moq-transport.md
```diff
@@ -1371,6 +1366,121 @@ UNSUBSCRIBE Message {
...
+   EndObject (i)],
+  Number of Parameters (i),
+  Parameters (..) ...
+}
+~~~
+{: #moq-transport-fetch-format title="MOQT FETCH Message"}
+
+* Fetch ID: The Fetch ID identifies a given fetch request. Fetch ID is a
+variable length integer that MUST be unique and monotonically increasing
+within  a session.
```
afrind, line 1421: nit: double space between 'within' and 'a'

draft-ietf-moq-transport.md
```diff
@@ -2024,13 +2193,45 @@ The Object Status field is only sent if the Object Payload Length is zero.
...
+
+### Fetch Header {#fetch-header}
+
+When a stream begins with `FETCH_HEADER`, all objects on the stream belong to the
+track requested in the Fetch message identified by `Fetch ID`.
+
+~~~
+FETCH_HEADER Message {
+  Fetch ID (i),
+  Publisher Priority (8),
```
afrind, line 2205: How does this get set by a relay with then entire fetch range already cached?

draft-ietf-moq-transport.md
```diff
@@ -2024,13 +2193,45 @@ The Object Status field is only sent if the Object Payload Length is zero.
...
+}
+~~~
+{: #fetch-header-format title="MOQT FETCH_HEADER Message"}
+
+
+When a publisher sends objects in response to a FETCH request, it finds the
+stream that is associated with the fetch request, or opens a new one and sends
+the `FETCH_HEADER`, and then serializes the following object fields.
```
afrind, line 2213: Do we want different language about how to send FETCH objects -- eg: do we need the bit about when to create the stream already?

draft-ietf-moq-transport.md
```diff
@@ -2024,13 +2193,45 @@ The Object Status field is only sent if the Object Payload Length is zero.
...
+  Group ID (i),
+  Object ID (i),
+  Object Payload Length (i),
+  [Object Status (i)],
+  Object Payload (..),
+}
+~~~
+{: #object-fetch-format title="MOQT Fetch Object Fields"}
+
+
```
afrind, line 2228: Are we missing the bit that says if you receive `G=1,O=0; G=1,O=2` then `G=1,O=1` is implied `Object Status=Object does not exist`

afrind (commented):
Individual Review

draft-ietf-moq-transport.md
```diff
@@ -1371,6 +1366,121 @@ UNSUBSCRIBE Message {
 
 * Subscribe ID: Subscription Identifier as defined in {{message-subscribe-req}}.
 
+
+## FETCH {#message-fetch}
+
+A subscriber issues a FETCH to a publisher to request a range of already published
+objects within a track. The publisher responding to a FETCH is responsible for
+retrieving all available Objects, and if there are gaps between Objects, it
```
suhasHere, line 1374: I don't see a reason to specify gaps explicitly.  How about we say "It is possible from a fetch response to infer if there exists any gaps "

draft-ietf-moq-transport.md
suhasHere: good point , may be we should separate it out.

draft-ietf-moq-transport.md
suhasHere: I am not sure we need 2 types of fetches. We agreed that if a fetch's end goes past a live edge, then the response include object only up to the live edge.

draft-ietf-moq-transport.md
suhasHere: I think that is implied. A note wouldn't hurt.

draft-ietf-moq-transport.md
ianswett: But then we risk leaving the draft in the state it's in, which is awkward at best.

I think I want everyone to be ok with the final functional outcome, and so we need to change Subscribe now.

Sorry to be difficult.

draft-ietf-moq-transport.md
ianswett: Good point, I made it non-normative.

draft-ietf-moq-transport.md
ianswett: I don't know, but SUBSCRIBE has params, so I'm assuming we'll need them for FETCH?

draft-ietf-moq-transport.md
ianswett: My understanding is that the latest Object in FETCH_OK is intended to be accurate, even if the end of the fetch is prior to that object?

draft-ietf-moq-transport.md
ianswett: We had consensus that if you did that, and if it returned at least one Object, you would get FETCH_OK.

But given we already had the concept of AbsoluteStart, it seemed in keeping with the draft to maintain that type of behavior, as Christian suggested.

draft-ietf-moq-transport.md
ianswett: Yes I did mean to change it to Fetch ID, so it's an error, but I think writing this as two modes of Subscribe would be much easier to read.

draft-ietf-moq-transport.md
ianswett: Yes, added a bool.

draft-ietf-moq-transport.md
ianswett: Are you suggesting the text is awkward, which it is?  Or asking for something else?

draft-ietf-moq-transport.md
ianswett: TBH, I have no idea how this gets set, but it was in @suhasHere slides so maybe he has ideas?

Alternatively, we could just serialize Peeps on a stream and then the only special thing would be that it's on a single stream.

draft-ietf-moq-transport.md
ianswett: It's said above in the FETCH message description, but we can repeat it here or move it.

ianswett (commented):
Thanks for the quick reviews!

draft-ietf-moq-transport.md
afrind: The language made sense when it was introduced for Forwarding Preferences.  An object could show up at a relay and the relay needed instructions for how to send it.  But fetches are a little simpler, you can open the stream as soon as you send the FETCH OK, or have the first object to send, and there's nothing else to do on subsequent objects.

Maybe something like: 

> Each object sent on a fetch stream after the header has the following format

draft-ietf-moq-transport.md
```diff
@@ -2024,13 +2198,45 @@ The Object Status field is only sent if the Object Payload Length is zero.
...
+The Object Status field is only sent if the Object Payload Length is zero.
+
+~~~
+{
+  Group ID (i),
+  Object ID (i),
+  Object Payload Length (i),
+  [Object Status (i)],
+  Object Payload (..),
+}
```
afrind, line 2229: Do we need other canonical MoQ fields like Subgroup ID and Forwarding Preference?

afrind (commented):
Individual Review

draft-ietf-moq-transport.md
suhasHere: I feel the Publisher Priority should not be in the fetch header. IIRC the discussions, i thought the idea was objects carry priority ( as carried over from the canonical object)  so that the consumer fetching knows the relative priorities between the objects.  This means that we need to add priority filed to the object not the stream header ?

Thoughts ?

draft-ietf-moq-transport.md
```diff
@@ -957,6 +959,14 @@ MOQT Control Message {
 |-------|-----------------------------------------------------|
 | 0x15  | MAX_SUBSCRIBE_ID ({{message-max-subscribe-id}})     |
 |-------|-----------------------------------------------------|
```
suhasHere, line 961: General comment. May be we should add a non-normative example text on how transition to live can be envisioned with combination of fetch and subscribe?

draft-ietf-moq-transport.md
suhasHere: AbsoluteStart in this cases seems like subset of AbsoluteRange. One can still set a real large end group/object. Regardless, the property (or the principle) that the fetch ends at the live edge at the time of request needs to be specified (and i think it is already).  Also fetch's start is whatever the relay choose it's start is ( going back to Christian's example).
 

Having said that , AbsoluteStart is redundant ?

draft-ietf-moq-transport.md
```diff
@@ -1191,12 +1201,12 @@ GOAWAY Message {
 
 ## SUBSCRIBE {#message-subscribe-req}
 
-### Filter Types {#sub-filter}
+A subscription causes the publisher to send newly published objects for a track
+and can start at either the latest Group or latest Object. A subscriber MUST NOT
```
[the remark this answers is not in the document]
afrind: So to clarify, you want to leave AbsStart and AbsRange as subscribe filters, but the numbers are only valid if they are in the future?  I still think Latest Group would be equivalent to Latest Object in that case, right?

draft-ietf-moq-transport.md
afrind: I think removing this priority and adding the other canonical object fields to the object serialization below sounds right.

draft-ietf-moq-transport.md
afrind: >  Also fetch's start is whatever the relay choose it's start is

Can you clarify?  My recollection in all the examples is the fetch start has to be the first object in the requested range that has a status other than "Object Not Available".

afrind (commented):
Individual Comments

draft-ietf-moq-transport.md
suhasHere: Agree. Let's start with that.  so the new object looks like the below for fetch ?
{
  Group ID = 0
  Subgroup ID = 0
  Publisher Priority = 0
  Object Payload Length
  [Object Status (i)],
  Payload
}

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
suhasHere: yes .. LatestGroup will point to latest group's latest object if the subscription is into the latest group ..

draft-ietf-moq-transport.md
ianswett: AbsoluteStart is redundant, so I'm happy to get rid of it.  Especially given we don't have restrictions on numbering of groups or objects today, I think the only way to guarantee you get all Objects that have already been published is to put 2^62-1 on the wire, which is awkward, but completely functional.

suhasHere: +1 on removing AbsoluteStart

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
martinduke: 10/9 Interim result: I believe we settled leaving SUBSCRIBE as-ism for now, but ranges fully in the past or a second subscribe for the same track = error.

Reasons: There is much disagreement on the details, and implementers would not prefer implementing a half-backed new subscribe for Dublin and then have to churn it when we settle on a design.

draft-ietf-moq-transport.md
martinduke: 10/9 interim: It was contentious, but I believe the result was to have a single fetch/subscribe ID space for now.

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1715,69 @@ SUBSCRIBE_ERROR
   ({{session-termination}}).
 
 
+## FETCH_OK {#message-fetch-ok}
+
+A publisher sends a FETCH control message for successful fetches.
```
martinduke, line 1720: An important outcome in Boston was that the publisher cannot send FETCH_OK (nit: s/FETCH/FETCH_OK) until it has determined the largest Group ID/Object ID, which may require the result of an upstream FETCH.

So a common pattern may be that the stream opens and data arrives well before FETCH_OK does (which is already a case the receiver has to handle because they're on different streams).

This should be explicitly stated.

martinduke, line 1720:
```suggestion
A publisher sends a FETCH_OK control message for successful fetches.
```

draft-ietf-moq-transport.md
martinduke: It depends on whether one can FETCH data that eventually is served via SUBSCRIBE. If not, it's useless. If so, yes, we need both.

But forwarding preference can go in the FETCH header, while (see my comment above) pub priority should go down here.

draft-ietf-moq-transport.md
```diff
@@ -1371,6 +1366,122 @@ UNSUBSCRIBE Message {
...
 * Subscribe ID: Subscription Identifier as defined in {{message-subscribe-req}}.
 
+
+## FETCH {#message-fetch}
+
+A subscriber issues a FETCH to a publisher to request a range of already published
+objects within a track. The publisher responding to a FETCH is responsible for
+retrieving all available Objects, and if there are gaps between Objects, it
+indicates these Objects are not available. As a result a fetch does not omit objects
+due to congestion.
```
vasilvv, line 1376: I feel like this does not spell out the most important property of FETCH explicitly, which is, FETCH `a...b` will return *all objects that exist* between `a` and `b`.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
suhasHere: After thinking through this a bit, fetch until live seems like a good addition regardless.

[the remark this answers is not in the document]
wilaw: > fetch until live seems like a good addition regardless.

The problem with 'FETCH until live' is that the live point is constantly moving (especially with 60fps stream-per-frame), therefore there is a possibility of a race condition in which by the time the relay has retrieved content it thought was the live edge, the live edge has already moved forward and it gets in a loop of constantly being behind.

draft-ietf-moq-transport.md
```diff
@@ -1371,6 +1376,110 @@ UNSUBSCRIBE Message {
...
```
ianswett, line 1386:
```suggestion
objects within a track. The publisher responding to a FETCH is responsible for retrieving
all available Objects. If there are gaps between Objects, the publisher omits them from the
fetch response. All omitted objects have status Object Not Available.
```

draft-ietf-moq-transport.md
```diff
@@ -1371,6 +1366,121 @@ UNSUBSCRIBE Message {
...
+requested. Only present for the "AbsoluteRange" filter type.
+
+* Parameters: The parameters are defined in {{version-specific-params}}.
+
+
+Any Objects that are not yet published will not be retrieved by a FETCH.
+The latest available Object is indicated in the FETCH_OK, and is the last
+Object a fetch will return if the 'Fetch Type is AbsoluteStart or if the
+type is AbsoluteRange and the EndGroup and EndObject have not yet been
+published.
```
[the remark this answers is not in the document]
ianswett: Agreed.  Last week's consensus was SUBSCRIBEs purely in the past were errors, but what they return still seems TBD.

draft-ietf-moq-transport.md
ianswett: Yes, but in a follow-up PR.

draft-ietf-moq-transport.md
ianswett: I assume we're killing forwarding preference in the very near future?

draft-ietf-moq-transport.md
ianswett: Alan, I think the text you suggested makes that clear now?

draft-ietf-moq-transport.md
TimEvens: It's going to be changed to `fetch_id` right? It would be nice that `fetch_id` has a different space than `subscribe_id`

draft-ietf-moq-transport.md
```diff
@@ -1371,6 +1380,109 @@ UNSUBSCRIBE Message {
...
+
+~~~
+FETCH Message {
+  Type (i) = 0x16,
+  Length (i),
+  Subscribe ID (i),
+  Track Namespace (tuple),
+  Track Name Length (i),
+  Track Name (..),
+  Subscriber Priority (8),
```
suhasHere, line 1411: should we call this as Fetch Priority,

draft-ietf-moq-transport.md
```diff
@@ -1371,6 +1380,109 @@ UNSUBSCRIBE Message {
...
+  Group Order (8),
+  StartGroup (i),
+  StartObject (i),
+  [EndGroup (i),
+   EndObject (i)],
```
suhasHere, line 1416: given that we are not having multiple fetch types at this point, should we make these non-optional?

draft-ietf-moq-transport.md
```diff
@@ -1371,6 +1380,109 @@ UNSUBSCRIBE Message {
...
+* Subscribe ID: The Subscribe ID identifies a given fetch request. Subscribe ID
+is a variable length integer that MUST be unique and monotonically increasing
+within  a session.
+
+* Track Namespace: Identifies the namespace of the track as defined in
+({{track-name}}).
+
+* Track Name: Identifies the track name as defined in ({{track-name}}).
+
+* Priority: Specifies the priority of a fetch request relative to
```
suhasHere, line 1432: s/Priority/Fetch Priority?

draft-ietf-moq-transport.md
```diff
@@ -1371,6 +1380,109 @@ UNSUBSCRIBE Message {
...
+
+* EndObject: The end Object ID, plus 1. A value of 0 means the entire group is
+requested.
+
+* Parameters: The parameters are defined in {{version-specific-params}}.
+
+
+Any Objects that are not yet published will not be retrieved by a FETCH.
+The latest available Object is indicated in the FETCH_OK, and is the last
+Object a fetch will return if the 'Fetch Type is AbsoluteStart or if the
```
suhasHere, line 1455: I don't see Fetch Type defined. May be this needs to reflect the same ?

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1716,69 @@ SUBSCRIBE_ERROR
...
+  Type (i) = 0x18,
+  Length (i),
+  Subscribe ID (i),
+  Group Order (8),
+  End Of Track (8),
+  Largest Group ID (i),
+  Largest Object ID (i),
+  Number of Parameters (i),
+  Subscribe Parameters (..) ...
+}
```
suhasHere, line 1735: wondering if Fetch_OK needs any parameters at all ?

draft-ietf-moq-transport.md
```diff
@@ -2024,13 +2199,45 @@ The Object Status field is only sent if the Object Payload Length is zero.
...
+
+### Fetch Header {#fetch-header}
+
+When a stream begins with `FETCH_HEADER`, all objects on the stream belong to the
+track requested in the Fetch message identified by `Fetch ID`.
+
+~~~
+FETCH_HEADER Message {
+  Fetch ID (i),
+  Publisher Priority (8),
```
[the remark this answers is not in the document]
suhasHere: For now, may be we should remove the Publisher Priority and revisit this later, if needed ?

draft-ietf-moq-transport.md
```diff
@@ -2024,13 +2200,45 @@ The Object Status field is only sent if the Object Payload Length is zero.
...
+The Object Status field is only sent if the Object Payload Length is zero.
+
+~~~
+{
+  Group ID (i),
+  Object ID (i),
+  Object Payload Length (i),
+  [Object Status (i)],
+  Object Payload (..),
+}
```
suhasHere, line 2231: we need to add SubGroupId to the object as well.

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1716,69 @@ SUBSCRIBE_ERROR
...
+  End Of Track (8),
+  Largest Group ID (i),
+  Largest Object ID (i),
+  Number of Parameters (i),
+  Subscribe Parameters (..) ...
+}
+~~~
+{: #moq-transport-fetch-ok format title="MOQT FETCH_OK Message"}
+
+* Subscribe ID: Fetch Identifier as defined in {{message-fetch}}.
```
fluffy, line 1739: Should this be "Fetch ID" ?

fluffy (approved):
I think this is close enough we should merge and then tweek.

draft-ietf-moq-transport.md
ianswett: For now, I'm going to keep it as is, because I don't want to rewrite the prioritization section.

draft-ietf-moq-transport.md
ianswett: Good point, fixed

draft-ietf-moq-transport.md
ianswett: As discussed, for now we're going to stick with one ID space.

draft-ietf-moq-transport.md
ianswett: Good point, done.

draft-ietf-moq-transport.md
```diff
@@ -1236,7 +1248,7 @@ SUBSCRIBE Message {
   Filter Type (i),
   [StartGroup (i),
    StartObject (i)],
-  [EndGroup (i),
+  Group (i),
```
```suggestion
  EndGroup (i),
```

ianswett (commented):
Thanks for the comments, I think this is ready to land.

draft-ietf-moq-transport.md
```diff
@@ -1986,13 +2165,47 @@ The Object Status field is only sent if the Object Payload Length is zero.
...
+  Subscribe ID (i),
+  Publisher Priority (8),
+}
+~~~
+{: #fetch-header-format title="MOQT FETCH_HEADER Message"}
+
+
+When a publisher sends objects in response to a FETCH request, it finds the
+stream that is associated with the fetch request, or opens a new one and sends
+the `FETCH_HEADER`, and then serializes the following object fields.
```
ianswett, line 2185:
```suggestion
the `FETCH_HEADER`, and then serializes the following object fields. Each object
sent on a fetch stream after the header has the following format:
```

draft-ietf-moq-transport.md
```diff
@@ -1986,13 +2165,48 @@ The Object Status field is only sent if the Object Payload Length is zero.
...
+sent on a fetch stream after the header has the following format:
+
+The Object Status field is only sent if the Object Payload Length is zero.
+
+~~~
+{
+  Group ID (i),
+  Subgroup ID (i),
+  Object ID (i),
+  Publisher Priority (8),
```
suhasHere, line 2195: I think Publisher Priority is repeated in 2 places (header and object message)

suhasHere (approved):
One minor nit on the PublisherPriority being duplicated in feath-header and object data. 
Otherwise, looks good to me.

Thanks @ianswett

draft-ietf-moq-transport.md
```diff
@@ -1954,7 +2138,6 @@ ID` and the subgroup indicated by 'Group ID' and `Subgroup ID`.
 
 ~~~
 STREAM_HEADER_SUBGROUP Message {
-  Subscribe ID (i),
```
vasilvv, line 1957: I don't remember when we agreed on this, I don't understand why this is in this PR, and I also don't like it because this will break our implementation badly.

ianswett: I can leave this here for now, but then this doesn't close the various issues around track alias vs Subscribe ID.

draft-ietf-moq-transport.md
ianswett: Good point, I removed it from FETCH_HEADER.

draft-ietf-moq-transport.md
ianswett: I'm not sure, but SUBSCRIBE_OK had them, so I put them in FETCH_OK.

draft-ietf-moq-transport.md
ianswett: I think it should be clear now, but I think we should clarify when you don't get everything for SUBSCRIBE?

draft-ietf-moq-transport.md
ianswett: SG, I took your simpler suggestion.

draft-ietf-moq-transport.md
ianswett: I did remove Subscribe ID from these, but we can revisit that if needed.

NEUDitao: Seems the description of the Stream Header Group missed the removal of Subscrib eID

Issue #582, opened 2024-10-07
Clarify the fidelity of SUBSCRIBE


afrind:
During the interim last week, there seemed to be consensus that SUBSCRIBE's default behavior is a high-fidelity mode (eg all new objects will get to all subscribers, with the exception of DATAGRAM tracks).  Based on the FETCH discussion - the status of a cache on any relay in the chain has no impact on new object fidelity.  The only parameter that can affect this is the subscriber's `DELIVERY_TIMEOUT`, which opt-in.

The only other issue to address is #462 - a subscriber falls too far behind live head to sustain the subscription based on sender resource constraints.  I recommend we add an error code to `SUBSCRIBE_DONE` such as `TOO_FAR_BEHIND`, include the final queued object ID, and remove the subscriber from the forwarding table so no new data is queued for them.  When those objects are delivered, the subscriber can FETCH later objects and/or resubscribe as needed.

afrind:
Individual Comment:

> I do not think this was the consensus

Here's a clip from the minutes:
```
Conclusion on allowing fetches to future objects:
Responses before live edge have fetch semantics, and after live edge
have subscribe semantics, provided:

Transition from past to live needs to be handled cleanly
Subscribe must be high fidelity
```

> It has been very clear that one of the main use case for sub groups was temporal scaled video where there was enough bandwidth for one sub group but not enough bandwidth for all the subgroups.

My understanding of that scenario is that prioritization can deprioritize some subgroups but they are queued indefinitely unless the objects within pass the DELIVERY_TIMEOUT.  If the delivery timeout is infinity, how does a relay know which subgroups can be skipped over (or when) and which are critical?  Do you have a different mechanism in mind?

afrind:
Individual Comment:

If a publisher doesn't have enough MAX_STREAM_DATA, MAX_DATA or MAX_STREAM_ID to send objects, it seems to me that it needs to queue until the peer grants these credits, up to the delivery timeout, right?

PR #587, opened 2024-10-11
clarify gaps in object/group IDs at relay 

fluffy:


draft-ietf-moq-transport.md
```diff
@@ -852,6 +852,32 @@ the active subscribers for that track. Relays MUST forward OBJECT messages to
...
+seeing the some of the objects in a group. It is possible for a client
+doing scalable video to publish the base layer over cellular, and the
+enhancement layers over WiFi. This could result in some relays getting
+the objects for both layers but other relays might only see one of the
+layer. Object being delivered over unreliable datagrams can also create
+gaps and out of order reception. What can be assumed is that the objects
+received on a single QUIC stream are in order but may have gaps.  These
+reasons can also impact whole groups and the relay cannot assume that it
+will receive all groups or that it will see all the earlier groups in
+the Track.
```
suhasHere, line 878: @fluffy I took editorial liberty to fix few nits. Hope that's ok.

I like the clarification this brings. Thanks for writing this up.

suhasHere (approved):

draft-ietf-moq-transport.md
fluffy: thank you

draft-ietf-moq-transport.md
```diff
@@ -852,6 +852,32 @@ the active subscribers for that track. Relays MUST forward OBJECT messages to
 matching subscribers in accordance to each subscription's priority, group order,
 and delivery timeout.
 
+Limited bandwidth upstream of the relay may result in gaps in the
+sequences of ObjectID values that the relay receives. For example if
```
afrind, line 856: If I understood the comments made in Boston, the *only* way such a gap can happen right now for non-datagrams is due to DELIVERY_TIMEOUT expiration and limited bandwidth by itself is not a cause of gaps.  Is that correct?

See also #582

draft-ietf-moq-transport.md
```diff
@@ -852,6 +852,32 @@ the active subscribers for that track. Relays MUST forward OBJECT messages to
...
+objects can be delivered out of order to the downstream relay. A relay
+cannot assume that it will see the first object in a group. Due to
+order of clients joining as well as timeouts, a relay may end up only
+seeing the some of the objects in a group. It is possible for a client
+doing scalable video to publish the base layer over cellular, and the
+enhancement layers over WiFi. This could result in some relays getting
+the objects for both layers but other relays might only see one of the
+layer. Object being delivered over unreliable datagrams can also create
+gaps and out of order reception. What can be assumed is that the objects
+received on a single QUIC stream are in order but may have gaps.  These
```
afrind, line 875: This PR is adding non-normative text which is in conflict normative text here:

https://github.com/moq-wg/moq-transport/blob/main/draft-ietf-moq-transport.md?plain=1#L740-L741

What happens to a subgroup with an object in the middle that misses the delivery timeout?  This PR says: omit the object and keep going.  Previous text says the relay cannot omit the object.  I'm not sure how to implement what is suggested in this PR for subgroups:

When I receive the beginning of an object from upstream, it's not past the delivery timeout -- that starts when the object is received -- so I will start forwarding it (eg: write it to a downstream QUIC stream).  If the end of the object does not arrive by the delivery timeout, I cannot retract the bytes already written.  I think the only recourse is to reset the entire stream (eg: drop the tail of the subgroup).  Is there another implementation you have in mind?

afrind (commented):
Individual Comments:

Can you link the issue this PR is addressing (or create one)?

draft-ietf-moq-transport.md
suhasHere: IIUC delivery timeout is considered even before inserting object in to the QUIC stream.  If the object stays within QUIC stream for ever,  it is still considered moving out of moq layer into transport layer.

afrind: Let me restate that to see if I got it:

You envision a system where the incoming object is queued outside of QUIC, and if the delivery timeout fires before it gets written to QUIC, you can skip over it.

Is that correct?

I imagine some relays might operate that way, and others might write incoming objects directly to the downstream stream (necessitating stream reset on timeout).  Is there an interoperability problem if clients assume one behavior or the other?

A related question is, when does the relay stop the delivery timeout timer - when the last byte of the object is written to the stream, when it's first transmitted, when it's ACK'd, or something else?  Maybe this one needs a separate issue.

suhasHere: > I imagine some relays might operate that way, and others might write incoming objects directly to the downstream stream (necessitating stream reset on timeout). Is there an interoperability problem if clients assume one behavior or the other?

This should work. Logic is same in both the case. 
Before inserting the object into the downstream QUIC stream, check for the delivery timeout and insert it.  Once inserted into QUIC stream, it needs to be treated as handed over from the moq  layer to the transport.

draft-ietf-moq-transport.md
englishm: This is my understanding, as well.

ianswett: Agreed, also the subsequent paragraph is a bit daunting

fluffy:
Few things to think about in updating from the call. 

1. when talking about gaps, should be clear of timescale and if means not arriving

2. bandwidth causes delay, cause timout, cuases drop

fluffy:
@afrind is going to help me on clarifying this.

martinduke:
Cullen,

I think one of the problems is that you keep using the words "out of order" which does not fully capture what I think we're agreeing on.

Say a subgroup is object IDs 2, 4, 6, 8:

sending 2, 4, 6, 8 = legal
sending 2, 4, 6 = legal
sending 2, 4, 8, 6 = illegal (out of order)
sending 2, 4, 8 = illegal (in order, but with an omitted object in the middle.)

fluffy:
Thanks Martin - i will try and rephrase. I also think that Alan and I were using gap differently so will get more specific with that too.

fluffy:
@afrind - Can you read the latest version of this and see if it matches your thinking. I think it should.

afrind:
Individual Comment:

Thanks for rewriting.  I would prefer to see this expressed normatively and relatively short in terms of what a relay can and cannot write to a stream, and then we can add an examples or informative motivation.

I was trying to write something that fit the bill, but stumbled on this case.

> sending 2, 4, 8 = illegal (in order, but with an omitted object in the middle.)

If a relay gets a subgroup with objects `2, 4, 6, 8`, I expect it can send one of the following

```
Nothing
2
2, 4
2, 4, 6
2, 4, 6, 8
```

However, what is the relay gets 

```
<GOAWAY>
2, 4 <RELIABLE_RESET>
```

It starts making a new subscription to the GOAWAY location, but doesn't close anything downstream.  By the time it resubscribes it gets another stream for this subgroup:

```
8
```

Would the relay be allowed to send `2, 4, 8` on one stream downstream?  If not, then what is it supposed to do?

* FETCH to fill in the gap?  
* If it gets non-sequential object (even 6) it has no idea if it missed and should just reset all the downstream subgroup streams
  * If so, can it open a new subgroup stream containing just `8` or stop trying to forward this subgroup?
* Something else?

martinduke:
IMO if there is a gap between subscriptions it has to FETCH the stuff
between the highest old and first new object ID. (Or subscribe to it, I
guess)

fluffy:
So in the case where the relay gets for group=77 subgroup=11
```
<GOAWAY>
2, 4 <RELIABLE_RESET>
```

after that it would not allowed for the publisher to resent that subsgroup with just 
```
8
```

Instead it would need to send all the groups in the substream so it could be in order so it would send 
```
2,4,6,8
```

fluffy:
What I want to get into the PR is something that helps people understand how this works and a bit on how the cases come up. Once we get there, we might need to tweek a few places in the draft with actual normative language.

afrind:
> after that it would not allowed for the publisher to resent that subsgroup with just
> 
> ```
> 8
> ```

I agree with that bit 100%.  The example I gave was a new subscription that started at latest object.  eg, maybe like this:

```
          /--- relay 1 ---\
Publisher                  relay3 ---> Subscriber 
          \--- relay 2 ---/
```

relay3 gets GOAWAY/2, 4<RELIABLE_RESET> from relay1 and a new ANNOUNCE from relay2.  relay3 issues SUBSCRIBE(latest object) to relay2, and gets a stream with just `8` for group=77 subgroup=11.

Also, we should probably move this discussion to a PR or the mailing list rather than a PR where it's less discoverable.

> What I want to get into the PR is something that helps people understand how this works and a bit on how the cases come up. Once we get there, we might need to tweek a few places in the draft with actual normative language.

I'll defer to the editor on the approach.

fluffy:
I don't know enough about reliable reset. Will web transport support that ?

afrind:
> Will web transport support that

I think it has too, but we should double check.  @vasilvv do you know?

draft-ietf-moq-transport.md
```diff
@@ -870,6 +870,38 @@ the active subscribers for that track. Relays MUST forward OBJECT messages to
...
+single QUIC stream are in the same order the original publisher intended
+in that sub group. For example, if the original publisher put objects
+with object ID 1,3,5,7 in the same subgroup, any downstream receiver
+will get the object in the same order in the stream, if they are
+received at all. So they might only get 1,3 and then have the stream
+close but they will never get 1,5 then 3 on a stream.
+
+There are also cases where a publisher lost its connection to an
+upstream relay and then reconnects, in which case objects can be
+delivered on different streams to the downstream relay.  It is possible
```
TimEvens, line 896:
```suggestion
delivered on different streams to the other relay(s).  It is possible
```

TimEvens (commented):
Curious if fetch is supposed to fill in the gaps here. I hope not as we would be reinventing a reliable \[overlay\] transport. I believe the expected behavior is to expect gaps and be okay with it. No fetch is required in this case to fill the gaps. This results in the relay having a cache with partial data, which is expected... but with fetch it gets interesting in that the fetch cannot be processed if there is any missing data... Might be worth clarifying that a bit more as it pertains to fetch request with gap data in cache on the relays.

ianswett:
This is an editorial PR, and I'm not sure what issue we're solving with the text, so I'm going to close for now and we can re-open if needed.

draft-ietf-moq-transport.md
```diff
@@ -870,6 +870,38 @@ the active subscribers for that track. Relays MUST forward OBJECT messages to
...
 
+Any real time system, which by definition has constraints on how late
+the data can be delivered, when running on the limited bandwidth
+internet, is not going to be able to guarantee delivery of everything
+but what is guaranteed is the order or object inside a stream if they
+are delivered at all. In the case where the original publisher put the
+even and odd frames on separate sub groups to do temporal scalability,
+the stream carrying one of those sub groups will see the objectID
+incrementing by more than one from one object in the stream to the next.
+Limited bandwidth upstream of the relay, combined with object delivery
```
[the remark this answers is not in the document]
fluffy: So I want to keep this only about relay to clients not relay to relay as that is a bit out of scope for MoQ

> Suggest removing terms `upstream` as that suggests a topology that is not defined in this spec.

draft-ietf-moq-transport.md
```diff
@@ -870,6 +870,56 @@ the active subscribers for that track. Relays MUST forward OBJECT messages to
 matching subscribers in accordance to each subscription's priority, group order,
 and delivery timeout.
 
+Any real time system, which by definition has constraints on how late
+the data can be delivered, when running on the limited bandwidth
+internet, is not going to be able to guarantee delivery of everything
+but what is guaranteed is the order or object inside a stream if they
```
afrind, line 876:
```suggestion
but what is guaranteed is the order of objects inside a stream if they
```

draft-ietf-moq-transport.md
```diff
@@ -870,6 +870,56 @@ the active subscribers for that track. Relays MUST forward OBJECT messages to
...
+are delivered at all. In the case where the original publisher put the
+even and odd frames on separate sub groups to do temporal scalability,
+the stream carrying one of those sub groups will see the objectID
+incrementing by more than one from one object in the stream to the next.
+Limited bandwidth upstream of the relay, combined with object delivery
+timeouts, may result in some of the objects never being delivered.
+Object being delivered over unreliable datagrams can loose objects and
```
afrind, line 883:
```suggestion
Tracks being delivered over unreliable datagrams can lose objects and
```

draft-ietf-moq-transport.md
```diff
@@ -870,6 +870,56 @@ the active subscribers for that track. Relays MUST forward OBJECT messages to
...
+have out of order reception.
+
+Subscribers (and relays) can assume that the objects received on a
+single QUIC stream are in the same order the original publisher intended
+in that sub group. For example, if the original publisher put objects
+with object ID 1,3,5,7 in the same subgroup, any receiver
+will get the object in the same order in the stream, if they are
```
afrind, line 890:
```suggestion
will get the objects in the same order in the stream, if they are
```

draft-ietf-moq-transport.md
```diff
@@ -870,6 +870,56 @@ the active subscribers for that track. Relays MUST forward OBJECT messages to
...
+upstream relay and then reconnects, in which case objects can be
+delivered on different streams to the other relay(s).  It is possible
+for a client doing scalable video to publish the base layer over
+cellular, and the enhancement layers over WiFi.  This could result in
+some relays getting the objects for both layers but other relays might
+only see one of the layer.  These reasons can also impact whole groups
+and the relay cannot assume that it will receive all groups or that it
+will see all the earlier groups in the Track.
+
+If there is no downstram congestion, group are forwarded by a
```
afrind, line 904:
```suggestion
If there is no downstream congestion, groups are forwarded by a 
```

draft-ietf-moq-transport.md
```diff
@@ -870,6 +870,56 @@ the active subscribers for that track. Relays MUST forward OBJECT messages to
...
+some relays getting the objects for both layers but other relays might
+only see one of the layer.  These reasons can also impact whole groups
+and the relay cannot assume that it will receive all groups or that it
+will see all the earlier groups in the Track.
+
+If there is no downstram congestion, group are forwarded by a 
+relay as soon as they are received. But as describe above, an 
+upstream failure might have group 3 arrive before group 2. The 
+rules in {{priorities}} are used to determine the order. If a 
+publisher chooses to send groups 30 then group 3, thoese rules
```
afrind, line 908:
```suggestion
publisher chooses to send group 30 then group 3, those rules 
```

afrind (commented):
Suggestions for typos

zafergurel:
To detect a gap, the previous group id can be added to the group header. So,  incrementing group ids or incrementing by one is not necessary.
For instance we have group ids: 1, 3, 5. Using the previous group ids, the relay can understand that there is not gap when it receives  the groups in this order 5, 3, 1. It can order them using the previous group ids as follows:

`1 <- 3 <- 5`

Even if you don't have 2 and 4 in the cache, the relay can understand there is no gap in this sequence by looking at every previous group id.

PR #589, opened 2024-10-18
FIN, RESET_STREAM, RESET_STREAM_AT, STOP_SENDING

martinduke:
This is explicitly not dealing with FETCH at all, but it won't be hard for me to write something similar for the FETCH stream once the rubble settles on that proposal.

PR #593, opened 2024-10-22
SUBSCRIBEs end at a group boundary

ianswett:
This is a dependency of Martin's Stream FIN/RESET PR, #589


martinduke (approved):
Besides the fact that this simplifies the FIN logic, I really like that this frees the subscriber from either finding out the last object or sending object ID 2^62-1.

PR #596, opened 2024-10-23
Remove End Of Subgroup status

ianswett:
We don't discuss how to use it anyway.

This was suggested in #589


martinduke (approved):
There are references to EndOfSubgroup in the FIN PR, so one of the two PRs is going to have to delete them.

ianswett:
> There are references to EndOfSubgroup in the FIN PR, so one of the two PRs is going to have to delete them.

Done

PR #601, opened 2024-10-27
Fix a typo in a section ref in the Intro

levaitamas:
There is an extra `}` that leads to rendering artifact:
```
*  Section 7} covers how data messages are encoded on the wire.
```


ianswett (approved):

ianswett:
Thanks!

Issue #572, opened 2024-10-01
When should MAX_SUBSCRIBE be sent ?


fluffy:
Do we need text that talks about when to send MAX_SUBSCRIBE. It defaults to zero and there does not seems to be normative text about using it.

afrind:
Individual Comment:

It's also a setup param, so each side can declare it's max before any subscriptions are sent.  That said, some text about when to send this could be helpful.  

Aside: Do we need a BLOCKED frame to inform the peer when you want to send a SUBSCRIBE but you are out of IDs?  QUIC has these.

ianswett:
We can base some text on https://datatracker.ietf.org/doc/html/rfc9000#flow-control

PR #602, opened 2024-10-28
Add SUBSCRIBES_BLOCKED

ianswett:
We said we might do this after adding MAX_SUBSCRIBE_ID, so here it is.

Motivated by a comment on #572


afrind:
Individual Review:

lgtm

fluffy (approved):

PR #603, opened 2024-10-28
Remove 'Role'

### LINKED ISSUE #574: Do we need the Role parameter
Not clear what the use case is. Seems like a foot gun in that it can be set wrong for how the connection is used.

### LINKED ISSUE #586: Behavior when unexpected Role Parameter arrives
I understand that a ClientSetup is sent that includes the role of `Publisher` if the client only wants to send objects, the role of `Subscriber` if it only wants to receive objects, and the role of `PubSub` if it wants to both send and receive objects.
And I believe that basically only the following relationships can be established:
- Client: `Publisher` vs Server: `Subscriber`
- Client: `Subscriber` vs Server: `Publisher`
- Client: `PubSub` vs Server: `PubSub`

My questions are as follows:
1. If a ClientSetup with `Subscriber` or `PubSub` role parameters were to arrive at an ingestion-only server, how would the server handle it? Is it required that the server returns `Subscriber` that the client does not expect and the client understands the server's intention?

2. If a client that has already sent `Publisher` or `Subscriber` receives a ServerSetup containing `PubSub`, is it acceptable to use that connection as is? There may be some security risks in this case.

Thank you!

ianswett:
The role somewhat limited what actions were valid, but it's not clear that was important.  Also, some things can be limited by MAX_SUBSCRIBE_ID now.

Fixes #574 
Fixes #586

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1128,33 +1128,6 @@ identified as 0xff00000D.
 
 ### Setup Parameters {#setup-params}
 
-#### ROLE {#role}
-
-The ROLE parameter (key 0x00) allows each endpoint to independently specify what
-functionality they support for the session. It has three possible values,
-which are of type varint:
-
-0x01: Publisher
-
-: The endpoint can process subscriptions and send objects, but not subscribe.
-  The endpoint MUST NOT send a SUBSCRIBE message and an ANNOUNCE MUST NOT be
-  sent to it.
-
-0x02: Subscriber
-
-: The endpoint can send subscriptions and receive objects, but not publish.
-  The endpoint MUST NOT send an ANNOUNCE message and a SUBSCRIBE MUST NOT be
-  sent to it.
-
-0x03: PubSub
-
-: The endpoint can act as a publisher or subscriber, and can send or process
-  any message type.
-
-Both endpoints MUST send a ROLE parameter with one of the three values
-specified above. Both endpoints MUST close the session if the ROLE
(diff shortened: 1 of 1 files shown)


wilaw:
Since you are removing parameter key 0x00, you should shift the other setup parameters down by one. So path currently 0x01 should become 0x00 and max-subscribe_ID currently 0x02  should become 0x01.

draft-ietf-moq-transport.md
```diff
@@ -1128,33 +1128,6 @@ identified as 0xff00000D.
...
-: The endpoint can act as a publisher or subscriber, and can send or process
-  any message type.
-
-Both endpoints MUST send a ROLE parameter with one of the three values
-specified above. Both endpoints MUST close the session if the ROLE
-parameter is missing or is not one of the three above-specified values.
-
 #### PATH {#path}
 
 The PATH parameter (key 0x01) allows the client to specify the path of
```
wilaw, line 1133: Should this now shift to key 0x01?

wilaw (commented):
Only suggestion to adjust other setup parameter keys now that you have removed 0x00.

draft-ietf-moq-transport.md
ianswett: In the past, we've commonly left codepoints where they are to avoid unnecessary code churn in implementations.  I don't have a strong opinion, so I'll let implementers decide what they prefer.

fluffy (approved):
Given we don't have a strong need for this, glad to see it go for now and we can add back later if we find a strong need for it.

PR #609, opened 2024-11-07
Do something useful with SUBSCRIBE_DONE

### LINKED ISSUE #424: SUBSCRIBE_DONE and terminal state
**tl;dr** when does a MoqTransport subscriber signal SUBSCRIBE_DONE to the application or clean up subscription state?

I ran into a bug when refactoring my relay to better support error conditions.

For subscriptions to the catalog or init track, my publisher would send a single OBJECT followed by an immediate SUBSCRIBE_DONE, since updates are (currently) not supported. It then removes the ID from the lookup table since this state is terminal.

Upon receiving the SUBSCRIBE_DONE, my subscriber also removes the ID from the lookup table and returns the error to the application. However, this causes a race, as the OBJECTs in flight (on another stream) will be dropped since they reference a newly non-existent subscription. This occurs the vast majority of the time especially because the control stream is higher priority.

Since SUBSCRIBE_DONE contains a final group/object, the intended behavior seems to be for the subscriber to delay until the final object is received.

But what if that object is dropped? What about the rest of the group? What happens to any other objects that might be queued (lower priority) or awaiting a retransmission?

Could the subscriber wait until all objects within subscribe range are received instead? This seems useful especially for something like FETCH, however it's currently impossible for a few reasons:

 - group_ids are not sequential and there can be intentional gaps
 - objects can be dropped, causing gaps with a group 
 - the latest object ID per group is unknown when using a stream per group.

Reliable live actually has the same exact issues. To generalize, its currently not possible to wait for all objects to arrive over the network.

### LINKED ISSUE #456: SUBSCRIBE_DONE final object
Section 6.14 says:

"Final Group: The largest Group ID sent by the publisher in an OBJECT message in this track.

Final Object: The largest Object ID sent by the publisher in an OBJECT message in the Final Group for this track."

Is it for the track, or for the subscribe? 

I support it being for the track, as that requires less publisher bookkeeping. But if it's that, we should probably make it clear that we really mean "for the track", unrelated to whatever was delivered from that subscribe_id.

### LINKED ISSUE #465: SUBSCRIBE_DONE logic is gruesome
How does a relay know when to send SUBSCRIBE_DONE?

The purpose of relay is to aggregate subscribes, so if overlapping subscribes come in with different endpoints I can't rely on a SUBSCRIBE_DONE from upstream.

If a SUBSCRIBE is entirely for future objects, the easiest thing to do is to send it when I've sent the last object ID, or a higher ID is published. One can add a timer per subscribe to allow for reordering, or just accept that reordered objects arriving after the last won't be delivered. I can't simply keep state for each object in the subscribe because I have no idea where there are object or group gaps. (maybe this is the use case for explicitly communicating these gaps after all?)

If a SUBSCRIBE is entirely for past objects, either I am serving them from cache (in which case I can provide an API to the cache to call when it is fully served), or I have to send a SUBSCRIBE upstream. If the SUBSCRIBE is unique, then I'll get a SUBSCRIBE_DONE from upstream. If it's aggregated somehow, once again I have to either watch for the last sequence number and run a timer, or keep score of all the objects I expect to see. IIUC, as it stands there is no guarantee that a subscribe for past objects will give me the complete set of objects.

If a SUBSCRIBE is for both past and future objects, then the union of both algorithms must both return true in order for the relay to send SUBSCRIBE_DONE. So the sender has to keep track of two separate object ranges independently. Because the SUBSCRIBE may arrive just as the publisher is incrementing the group ID, I need to track the max object ID in each group so I can tell when the past portion of the subscription is done.

At the origin, this is somewhat simpler in that published objects can be assumed to arrive in order and without gaps, and past objects can arrive in their own order and without gaps.

martinduke:
Fixes #424
Fixes #456
Fixes #465

Comes up with some rules on when to send SUBSCRIBE_DONE, and what to do with it.

Deletes the ContentExists and Final Group/Object fields, because there is nothing the receiver does with it.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1797,8 +1797,8 @@ stream state may persist until delivery completes.
 A subscriber that receives SUBSCRIBE_DONE SHOULD set a timer of at least 2 seconds
 in case some datagrams or unopened streams are still inbound due to
 prioritization or packet loss. Once the timer has expired, the receiver destroys
-subscription state once all open streams for the subscription have closed. Ai
-receiver MAY destroy subscription state earlier, at the cost of potentially not
+subscription state once all open streams for the subscription have closed. A
+subscriber MAY discard subscription state earlier, at the cost of potentially not
 delivering some late objects to the application.
 
 The format of `SUBSCRIBE_DONE` is as follows:
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1780,12 +1780,15 @@ FETCH_ERROR
 
 ## SUBSCRIBE_DONE {#message-subscribe-done}
 
-A publisher sends a `SUBSCRIBE_DONE` message to indicate it will not open any
-additional streams or send any more datagrams for a subscription. When all
-streams for the subscription are fully closed, each endpoint can destroy its
-subscription state.
+A publisher sends a `SUBSCRIBE_DONE` message when it is not going to send
+additional objects for a subscription. Because SUBSCRIBE_DONE is sent on the
+control stream, it is likely to arrive at the receiver before late-arriving
+objects, and often even late-opening streams. However, the receiver uses it
+as an indication that it should receive any late-opening streams in a relatively
+short time. When all streams for the subscription are fully closed, each
+endpoint can destroy its subscription state.
 
-Note that some objects in the subscribed groups might never be delivered,
+Note that some objects in the subscribed track might never be delivered,
 because a stream was reset, or never opened in the first place, due to the
 delivery timeout.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -725,7 +725,7 @@ For successful subscriptions, the publisher maintains a list of
 subscribers for each track. Each new OBJECT belonging to the
 track within the subscription range is forwarded to each active
 subscriber, dependent on the congestion response. A subscription
-remains active until the publisher of the track terminates the
+remains active until soon after the publisher of the track terminates the
 subscription with a SUBSCRIBE_DONE (see {{message-subscribe-done}}).
 
 A caching relay saves Objects to its cache identified by the Object's
@@ -1343,8 +1343,7 @@ See {{priorities}}.
 A subscriber issues a `UNSUBSCRIBE` message to a publisher indicating it is no
 longer interested in receiving media for the specified track and Objects
 should stop being sent as soon as possible.  The publisher sends a
-SUBSCRIBE_DONE to acknowledge the unsubscribe was successful and indicate
-the final Object.
+SUBSCRIBE_DONE to acknowledge the unsubscribe was successful.
 
 The format of `UNSUBSCRIBE` is as follows:
 
@@ -1767,9 +1766,25 @@ FETCH_ERROR
 
 ## SUBSCRIBE_DONE {#message-subscribe-done}
 
-A publisher sends a `SUBSCRIBE_DONE` message to indicate it is done publishing
-Objects for that subscription.  The Status Code indicates why the subscription ended,
-and whether it was an error.
+A publisher sends a `SUBSCRIBE_DONE` message to indicate it will not open any
+additional streams for a subscription. When all streams for the subscription are
+fully closed, each endpoint can destroy its subscription state.
+
+Note that some objects in the subscribed groups might not have been delivered,
+because a stream was reset, or never opened in the first place, due to the
+delivery timeout.
+
+A sender MUST NOT send SUBSCRIBE_DONE until it has closed all streams it will
+ever open for a subscription. After sending SUBSCRIBE_DONE, MoQT can immediately
+destroy subscription state. The QUIC implementation might still be resolving the
+closing of the stream on the wire.
+
+A receiver that receives SUBSCRIBE_DONE SHOULD set a timer of at least 2 seconds
+in case some unopened streams are still inbound due to prioritization or packet
+loss. Once the timer has expired, the receiver destroys subscription state once
+all open streams for the subscription have closed. A receiver MAY destroy
+subscription state earlier, at the cost of potentially not delivering some late
+objects to the application.
 
 The format of `SUBSCRIBE_DONE` is as follows:
 
@@ -1781,9 +1796,6 @@ SUBSCRIBE_DONE Message {
   Status Code (i),
   Reason Phrase Length (i),
   Reason Phrase (..),
-  ContentExists (8),
-  [Final Group (i)],
-  [Final Object (i)],
 }
 ~~~
 {: #moq-transport-subscribe-fin-format title="MOQT SUBSCRIBE_DONE Message"}
@@ -1793,17 +1805,6 @@ SUBSCRIBE_DONE Message {
 * Status Code: An integer status code indicating why the subscription ended.
 
 * Reason Phrase: Provides the reason for subscription error.
-
-* ContentExists: 1 if an object has been published for this subscription, 0 if
-not. If 0, then the Final Group and Final Object fields will not be present.
-Any other value is a protocol error and MUST terminate the session with a
-Protocol Violation ({{session-termination}}).
-
-* Final Group: The largest Group ID sent by the publisher in an OBJECT
-message in this track.
-
-* Final Object: The largest Object ID sent by the publisher in an OBJECT
-message in the `Final Group` for this track.
 
 ## MAX_SUBSCRIBE_ID {#message-max-subscribe-id}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1770,13 +1770,13 @@ A publisher sends a `SUBSCRIBE_DONE` message to indicate it will not open any
 additional streams for a subscription. When all streams for the subscription are
 fully closed, each endpoint can destroy its subscription state.
 
-Note that some objects in the subscribed groups might not have been delivered,
+Note that some objects in the subscribed groups might never be delivered,
 because a stream was reset, or never opened in the first place, due to the
 delivery timeout.
 
 A sender MUST NOT send SUBSCRIBE_DONE until it has closed all streams it will
 ever open for a subscription. After sending SUBSCRIBE_DONE, MoQT can immediately
-destroy subscription state. The QUIC implementation might still be resolving the
+destroy subscription state. The QUIC layer might still be resolving the
 closing of the stream on the wire.
 
 A receiver that receives SUBSCRIBE_DONE SHOULD set a timer of at least 2 seconds
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1767,24 +1767,25 @@ FETCH_ERROR
 ## SUBSCRIBE_DONE {#message-subscribe-done}
 
 A publisher sends a `SUBSCRIBE_DONE` message to indicate it will not open any
-additional streams for a subscription. When all streams for the subscription are
-fully closed, each endpoint can destroy its subscription state.
+additional streams or send any more datagrams for a subscription. When all
+streams for the subscription are fully closed, each endpoint can destroy its
+subscription state.
 
 Note that some objects in the subscribed groups might never be delivered,
 because a stream was reset, or never opened in the first place, due to the
 delivery timeout.
 
 A sender MUST NOT send SUBSCRIBE_DONE until it has closed all streams it will
-ever open for a subscription. After sending SUBSCRIBE_DONE, MoQT can immediately
-destroy subscription state. The QUIC layer might still be resolving the
-closing of the stream on the wire.
+ever open, and has no pending datagrams, for a subscription. After sending
+SUBSCRIBE_DONE, the sender  can immediately destroy subscription state, although
+stream state may persist until delivery completes.
 
 A receiver that receives SUBSCRIBE_DONE SHOULD set a timer of at least 2 seconds
-in case some unopened streams are still inbound due to prioritization or packet
-loss. Once the timer has expired, the receiver destroys subscription state once
-all open streams for the subscription have closed. A receiver MAY destroy
-subscription state earlier, at the cost of potentially not delivering some late
-objects to the application.
+in case some datagrams or unopened streams are still inbound due to
+prioritization or packet loss. Once the timer has expired, the receiver destroys
+subscription state once all open streams for the subscription have closed. Ai
+receiver MAY destroy subscription state earlier, at the cost of potentially not
+delivering some late objects to the application.
 
 The format of `SUBSCRIBE_DONE` is as follows:
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1785,24 +1785,33 @@ additional objects for a subscription. Because SUBSCRIBE_DONE is sent on the
 control stream, it is likely to arrive at the receiver before late-arriving
 objects, and often even late-opening streams. However, the receiver uses it
 as an indication that it should receive any late-opening streams in a relatively
-short time. When all streams for the subscription are fully closed, each
-endpoint can destroy its subscription state.
+short time.
 
 Note that some objects in the subscribed track might never be delivered,
 because a stream was reset, or never opened in the first place, due to the
 delivery timeout.
 
 A sender MUST NOT send SUBSCRIBE_DONE until it has closed all streams it will
-ever open, and has no pending datagrams, for a subscription. After sending
-SUBSCRIBE_DONE, the sender can immediately destroy subscription state, although
-stream state may persist until delivery completes.
+ever open, and has no further datagrams to send, for a subscription. After
+sending SUBSCRIBE_DONE, the sender can immediately destroy subscription state,
+although stream state can persist until delivery completes. The sender might
+persist subscription state to enforce the delivery timeout by resetting streams
+on which it has already sent FIN, only deleting it when all such streams have
+received ACK of the FIN.
 
-A subscriber that receives SUBSCRIBE_DONE SHOULD set a timer of at least 2 seconds
-in case some datagrams or unopened streams are still inbound due to
-prioritization or packet loss. Once the timer has expired, the receiver destroys
-subscription state once all open streams for the subscription have closed. A
-subscriber MAY discard subscription state earlier, at the cost of potentially not
-delivering some late objects to the application.
+A sender MUST NOT destroy subscription state until it sends SUBSCRIBE_DONE,
+though it can choose to stop sending objects (and thus send SUBSCRIBE_DONE) for
+any reason.
+
+A subscriber that receives SUBSCRIBE_DONE SHOULD set a timer of at least its
+delivery timeout in case some objects are still inbound due to prioritization
+or packet loss. The subscriber MAY dispense with a timer if it sent UNSUBSCRIBE
+or is otherwise no longer interested in objects from the track. Once the timer
+has expired, the receiver destroys subscription state once all open streams for
+the subscription have closed. A subscriber MAY discard subscription state
+earlier, at the cost of potentially not delivering some late objects to the
+application. The subscriber SHOULD send STOP_SENDING on all streams related to
+the subscription when it deletes subscription state.
 
 The format of `SUBSCRIBE_DONE` is as follows:
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1791,7 +1791,7 @@ delivery timeout.
 
 A sender MUST NOT send SUBSCRIBE_DONE until it has closed all streams it will
 ever open, and has no pending datagrams, for a subscription. After sending
-SUBSCRIBE_DONE, the sender  can immediately destroy subscription state, although
+SUBSCRIBE_DONE, the sender can immediately destroy subscription state, although
 stream state may persist until delivery completes.
 
 A receiver that receives SUBSCRIBE_DONE SHOULD set a timer of at least 2 seconds
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1794,7 +1794,7 @@ ever open, and has no pending datagrams, for a subscription. After sending
 SUBSCRIBE_DONE, the sender can immediately destroy subscription state, although
 stream state may persist until delivery completes.
 
-A receiver that receives SUBSCRIBE_DONE SHOULD set a timer of at least 2 seconds
+A subscriber that receives SUBSCRIBE_DONE SHOULD set a timer of at least 2 seconds
 in case some datagrams or unopened streams are still inbound due to
 prioritization or packet loss. Once the timer has expired, the receiver destroys
 subscription state once all open streams for the subscription have closed. Ai


martinduke:
Note: Another design would not use a timer at the receiver. Instead, it would require that all reset streams use RESET_STREAM_AT with at least the track alias delivered reliably, and the SUBSCRIBE_DONE would indicate how many streams were opened over the life of the subscription.

draft-ietf-moq-transport.md
```diff
@@ -1767,9 +1766,25 @@ FETCH_ERROR
 
 ## SUBSCRIBE_DONE {#message-subscribe-done}
 
-A publisher sends a `SUBSCRIBE_DONE` message to indicate it is done publishing
-Objects for that subscription.  The Status Code indicates why the subscription ended,
-and whether it was an error.
+A publisher sends a `SUBSCRIBE_DONE` message to indicate it will not open any
+additional streams for a subscription. When all streams for the subscription are
```
afrind, line 1770: Or send additional datagrams

draft-ietf-moq-transport.md
```diff
@@ -1767,9 +1766,25 @@ FETCH_ERROR
...
+fully closed, each endpoint can destroy its subscription state.
+
+Note that some objects in the subscribed groups might never be delivered,
+because a stream was reset, or never opened in the first place, due to the
+delivery timeout.
+
+A sender MUST NOT send SUBSCRIBE_DONE until it has closed all streams it will
+ever open for a subscription. After sending SUBSCRIBE_DONE, MoQT can immediately
```
afrind, line 1778: s/MoQT/a sender (or publisher?)/

draft-ietf-moq-transport.md
```diff
@@ -1767,9 +1766,25 @@ FETCH_ERROR
...
+destroy subscription state. The QUIC layer might still be resolving the
+closing of the stream on the wire.
```
afrind, line 1780:
```suggestion
destroy subscription state, even though the underlying transport might not have 
finished closing the associated streams.
```

afrind: Also do we need to say something about datagrams?

draft-ietf-moq-transport.md
```diff
@@ -1767,9 +1766,25 @@ FETCH_ERROR
...
+
+A receiver that receives SUBSCRIBE_DONE SHOULD set a timer of at least 2 seconds
+in case some unopened streams are still inbound due to prioritization or packet
```
afrind, line 1783: streams or datagrams

afrind (commented):
Individual Review:

I can live with using a timer, but I prefer putting an opened stream count in SUBSCRIBE_DONE and requiring any reset streams to use RELIABLE_RESET_STREAM with an offset >= the stream header.

draft-ietf-moq-transport.md
```diff
@@ -1767,9 +1766,25 @@ FETCH_ERROR
...
+Note that some objects in the subscribed groups might never be delivered,
+because a stream was reset, or never opened in the first place, due to the
+delivery timeout.
+
+A sender MUST NOT send SUBSCRIBE_DONE until it has closed all streams it will
+ever open for a subscription. After sending SUBSCRIBE_DONE, MoQT can immediately
+destroy subscription state. The QUIC layer might still be resolving the
+closing of the stream on the wire.
+
+A receiver that receives SUBSCRIBE_DONE SHOULD set a timer of at least 2 seconds
```
afrind, line 1782: Should there be an exception if the SUBSCRIBE_DONE has code UNSUBSCRIBED (eg: the subscriber already expressed disinterest)?

draft-ietf-moq-transport.md
martinduke: Done. #downwithdatagrams

draft-ietf-moq-transport.md
martinduke: As the text says, the receiver MAY delete early if the application has lost interest in the data.

I don't think we need special language for that.

draft-ietf-moq-transport.md
```diff
@@ -1781,9 +1780,26 @@ FETCH_ERROR
...
+additional streams or send any more datagrams for a subscription. When all
+streams for the subscription are fully closed, each endpoint can destroy its
+subscription state.
+
+Note that some objects in the subscribed groups might never be delivered,
+because a stream was reset, or never opened in the first place, due to the
+delivery timeout.
+
+A sender MUST NOT send SUBSCRIBE_DONE until it has closed all streams it will
+ever open, and has no pending datagrams, for a subscription. After sending
```
ianswett, line 1793:
```suggestion
ever open, and has no unsent datagrams, for a subscription. After sending
```

draft-ietf-moq-transport.md
```diff
@@ -1781,9 +1780,26 @@ FETCH_ERROR
...
+SUBSCRIBE_DONE, the sender  can immediately destroy subscription state, although
```
ianswett, line 1794:
```suggestion
SUBSCRIBE_DONE, the sender can immediately destroy subscription state, although
```

draft-ietf-moq-transport.md
```diff
@@ -1781,9 +1780,26 @@ FETCH_ERROR
...
+stream state may persist until delivery completes.
+
+A receiver that receives SUBSCRIBE_DONE SHOULD set a timer of at least 2 seconds
```
ianswett, line 1797:
```suggestion
A subscriber that receives SUBSCRIBE_DONE SHOULD set a timer of at least 2 seconds
```

draft-ietf-moq-transport.md
ianswett: 2 seconds seems arbitrary.  Can we use 3 * PTO or leave it open to implementations?

draft-ietf-moq-transport.md
```diff
@@ -1781,9 +1780,26 @@ FETCH_ERROR
...
+in case some datagrams or unopened streams are still inbound due to
+prioritization or packet loss. Once the timer has expired, the receiver destroys
+subscription state once all open streams for the subscription have closed. Ai
+receiver MAY destroy subscription state earlier, at the cost of potentially not
+delivering some late objects to the application.
```
ianswett, line 1802:
```suggestion
subscription state once all open streams for the subscription have closed. A
subscriber MAY discard subscription state earlier, at the cost of potentially not
delivering some late objects to the application.
```

draft-ietf-moq-transport.md
```diff
@@ -1795,9 +1811,6 @@ SUBSCRIBE_DONE Message {
   Status Code (i),
   Reason Phrase Length (i),
   Reason Phrase (..),
-  ContentExists (8),
-  [Final Group (i)],
-  [Final Object (i)],
```
ianswett, line 1800: Was this removed because it was difficult, or because you think it's useless or ?

ianswett (commented):
Overall looks good, but I'm curious why you removed the Final Group and Final Object?

ianswett (approved):

ianswett (changes requested):
Oops, I forgot I would like to change the 2 second constant before merging this.

draft-ietf-moq-transport.md
martinduke: It is totally arbitrary. The MoQ application is unlikely to have an API for the PTO, but there's no number I'd be unhappy with.

draft-ietf-moq-transport.md
martinduke: IIUC Datagrams can't be pending at the application level. If QUIC can't take them, they just drop on the floor?

draft-ietf-moq-transport.md
martinduke: As the description above says, I couldn't think of anything the receiver could do with it. This also clears the issue about if it's the final object of the track or the subscription.

martinduke:
I thought of a third way this could work:

If we got WebTransport APIs to expose the QUIC Stream ID of a stream, we could include the largest stream ID opened on behalf of a subscription. If all unidirectional sender-opened streams up to that ID have been observed, it's safe to destroy the state.

draft-ietf-moq-transport.md
```diff
@@ -1781,9 +1780,26 @@ FETCH_ERROR
 
 ## SUBSCRIBE_DONE {#message-subscribe-done}
 
-A publisher sends a `SUBSCRIBE_DONE` message to indicate it is done publishing
-Objects for that subscription.  The Status Code indicates why the subscription ended,
-and whether it was an error.
+A publisher sends a `SUBSCRIBE_DONE` message to indicate it will not open any
+additional streams or send any more datagrams for a subscription. When all
+streams for the subscription are fully closed, each endpoint can destroy its
```
suhasHere, line 1785: by the phrase "fully closed" , do you mean stream fin or does it cover stream reset cases too ?

draft-ietf-moq-transport.md
```diff
@@ -1781,9 +1780,26 @@ FETCH_ERROR
...
+subscription state.
+
+Note that some objects in the subscribed groups might never be delivered,
```
suhasHere, line 1788: s/subscribed groups/subscribed tracks

draft-ietf-moq-transport.md
```diff
@@ -1781,9 +1780,26 @@ FETCH_ERROR
...
+
+A sender MUST NOT send SUBSCRIBE_DONE until it has closed all streams it will
+ever open, and has no pending datagrams, for a subscription. After sending
+SUBSCRIBE_DONE, the sender can immediately destroy subscription state, although
+stream state may persist until delivery completes.
+
+A subscriber that receives SUBSCRIBE_DONE SHOULD set a timer of at least 2 seconds
+in case some datagrams or unopened streams are still inbound due to
+prioritization or packet loss. Once the timer has expired, the receiver destroys
+subscription state once all open streams for the subscription have closed. A
```
suhasHere, line 1800: "Once the timer has expired, the receiver can cleanup its subscription state" 

I don't think we need to have "once al open streams for the subscriptions have closed".

draft-ietf-moq-transport.md
suhasHere: i tend to agree with @martinduke  .. a receiver knows the last object it received before it cleans up its subscription state and adding more info in subscribe_done is not that helpful.

draft-ietf-moq-transport.md
martinduke: It covers Reset -- I'm not sure how it could work just on FIN

draft-ietf-moq-transport.md
fluffy: The quicr relay queues datagram packets at the applications layer and only passing them to QUIC layer when there is space in the CWIN for them. This allows the relay to control the times outs and priorities for object sent with datagram. I think that will deliver a much better user experience than something that just drops them on the floor (thought both implementations are valid spec compliant AFAIC )

draft-ietf-moq-transport.md
```diff
@@ -1781,9 +1780,26 @@ FETCH_ERROR
 
 ## SUBSCRIBE_DONE {#message-subscribe-done}
 
-A publisher sends a `SUBSCRIBE_DONE` message to indicate it is done publishing
-Objects for that subscription.  The Status Code indicates why the subscription ended,
-and whether it was an error.
+A publisher sends a `SUBSCRIBE_DONE` message to indicate it will not open any
+additional streams or send any more datagrams for a subscription. When all
```
fluffy, line 1784: So lets say I have a track with a single group and single sub group and a million objects in the group. (this might be a common patter for deliver software updates or AI models updates. This new text say the DONE would be sent as soon as the subscribe was received. This seems very wrong and will cause weird race conditions when the DONE was sent after the publisher opened the last stream, but the subscriber got the DONE before they new stream had been opened.

draft-ietf-moq-transport.md
```diff
@@ -1781,9 +1780,26 @@ FETCH_ERROR
...
+Note that some objects in the subscribed groups might never be delivered,
+because a stream was reset, or never opened in the first place, due to the
+delivery timeout.
+
+A sender MUST NOT send SUBSCRIBE_DONE until it has closed all streams it will
+ever open, and has no pending datagrams, for a subscription. After sending
+SUBSCRIBE_DONE, the sender can immediately destroy subscription state, although
+stream state may persist until delivery completes.
+
+A subscriber that receives SUBSCRIBE_DONE SHOULD set a timer of at least 2 seconds
```
fluffy, line 1797: 2 seconds sounds insanely fragile.

fluffy (changes requested):
I don't think this would work.

draft-ietf-moq-transport.md
martinduke: @fluffy is it your contention that the Final Object field refers to the final object in the track, or the subscription? If the subscription, is it the final one that was delivered, or the final one that was in range (because those are not the same thing).

> I don't see how this works to remove this. If client A does a subscribe of group 10 to 20, will the relay cache that 20 is the end of track ? I just don't see much connection between end of subscription and end of track.

Why would a relay cache an end of track at all in that case? It is perfectly normal for a cache to not know the end of a track.

> I'm even more baffled how, for a group that is not the last group in the track, this allows the client to figure out when it has got all the objects in the group.

Final Object doesn't ensure that. If a stream is never opened by the sender due to delivery timeout, or the track uses DATAGRAM, the receiver has no idea if those Object IDs were dropped or never minted in the first place.

draft-ietf-moq-transport.md
martinduke: Would you like to suggest a different timeout, or one of the other proposals discussed in this thread? I'm open to any of them but was reluctant to impose more requirements on the QUIC/WebTransport stack.

draft-ietf-moq-transport.md
martinduke: That is not what it says at all. A few paragraphs down, it says:

"A sender MUST NOT send SUBSCRIBE_DONE until it has closed all streams it will
ever open, and has no pending datagrams, for a subscription."

Race conditions are inevitable, even with that rule, because we can't guarantee the order in which QUIC will deliver things. That's why there has to be either a timer or a careful accounting of what streams have been opened.

draft-ietf-moq-transport.md
fluffy: Sorry - in the diff I missed you were removing from the subscribe done only not the dataplane. I don't really case what is there

draft-ietf-moq-transport.md
fluffy: So 

> "A sender MUST NOT send SUBSCRIBE_DONE until it has closed all streams it will
> ever open, and has no pending datagrams, for a subscription."

and 

> "A publisher sends a `SUBSCRIBE_DONE` message to indicate it will not open any
> additional streams or send any more datagrams for a subscription."

Seem to contradict each other about when DONE is sent.

martinduke: I agree that this is confusing and I'll fix it.

What I'm getting at is this: we don't need a FETCH_DONE because it's all on one stream, so the FIN/RESET is adequate to communicate the end of the FETCH. If the subscriber somehow knew for sure how many streams would open for a SUBSCRIBE, then stream FIN/RESETs would be adequate to know when to destroy subscription state.

SUBSCRIBE_DONE is only needed because it doesn't know for sure; the sender has to tell it that it's done opening streams, so once the current crop of streams is closed it's safe to tear down.

Unfortunately, it is quite likely that SUBSCRIBE_DONE outraces the last stream openings. So either there has to be a timer, or there has to be some indication of how many streams there were (which would levy new requirements on QUIC/WebTransport)

draft-ietf-moq-transport.md
martinduke: The effect of the timer expiring early is that some objects may be dropped on the floor instead of delivered to the  application.

If the receiver is interested in objects beyond that, it should set a longer timer.

martinduke:
I went with the timer design because I thought it would be easy to write. But it turns out it's confusing to people, so I'm starting to think that we either require RESET_STREAM_AT or expose QUIC stream IDs to the application, and get rid of the timer.

draft-ietf-moq-transport.md
ianswett: I'm fine with removing this.

ianswett:
I'm fine with this design, minus explicitly specifying a timeout.

mengelbart:
I think RESET_STREAM_AT might work for QUIC, but if I understand WebTransport correctly, it doesn't expose it to the application and only uses it for the WebTransport session IDs. Referencing the highest stream ID sounds cleaner to me than setting an arbitrary timeout, but if that is hard for some QUIC APIs, I'm fine with a timeout.

Another option I could think of is to send SUBSCRIBE_DONE on a separate last stream which must not be reset. But that would mean we no longer send all control messages on the control stream...

mengelbart:
oh and the separate stream also requires exposing the stream ID...

martinduke:
> Another option I could think of is to send SUBSCRIBE_DONE on a separate last stream which must not be reset. But that would mean we no longer send all control messages on the control stream...

Even if that stream has a very low priority, I don't think we can guarantee that it arrives after any object streams that are sent around the same time.

martinduke:
> I think RESET_STREAM_AT might work for QUIC, but if I understand WebTransport correctly, it doesn't expose it to the application and only uses it for the WebTransport session IDs.

It's not clear to me if WebTransport is specifying the APIs, but to me this is less of an IETF question and more a "what are WebTransport implementers willing to do" question.

martinduke:
@vasilvv points out that the Webtrans API has a signal when the stream FIN has been ACKed. This makes everything way easier.

afrind:
Individual Comment:

I don't like using FIN-ACK:

1. It moves the waiting period from the subscriber to the publisher.  The high-scale/long wait cases are likely when relays publish fan-out to client subscribers, and the relay is the one with reason to free up resources faster.  The publisher otherwise has no reason to keep this state around but the subscriber might.
2. Receiving a FIN-ACK does not mean the MoQ layer has received the stream, only the QUIC stack, which could be in the peer's kernel (or someday, NIC).  The SUBSCRIBE_DONE could still be racing delivery of the QUIC data and result in missed objects.  HTTP/3 explicitly chose *not* to use QUIC signals in QPACK for these reasons.
3. The subscriber doesn't have to wait around after SUBSCRIBE_DONE if it is uninterested in remaining data, but by withholding the signal, it doesn't get a choice.

I still prefer using RESET_STREAM_AT -- the downside from my perspective is pushing WebTransport to include this surface, which I think they should anyway, and perhaps that raw QUIC implementations MUST negotiate this extension.  Timers are inelegant and a waste when everything was already received.

martinduke:
> LGTM. I assume this works with existing WebTransport APIs?

It's in both the spec and our implementation.

draft-ietf-moq-transport.md
```diff
@@ -1358,8 +1358,7 @@ See {{priorities}}.
 A subscriber issues a `UNSUBSCRIBE` message to a publisher indicating it is no
 longer interested in receiving media for the specified track and Objects
 should stop being sent as soon as possible.  The publisher sends a
-SUBSCRIBE_DONE to acknowledge the unsubscribe was successful and indicate
-the final Object.
+SUBSCRIBE_DONE to acknowledge the unsubscribe was successful.
```
TimEvens, line 1361: Seems weird to me that it's named `SUBSCRIBE_DONE` for what is defined as `UNSUBSCRIBE_ACK`

draft-ietf-moq-transport.md
```diff
@@ -1782,9 +1781,24 @@ FETCH_ERROR
...
-and whether it was an error.
+A publisher sends a `SUBSCRIBE_DONE` message to indicate it has finished sending
+objects for a subscription, is discarding its subscription state, and the
+subscriber can discard its state as well.
+
+A sender MUST NOT send SUBSCRIBE_DONE until all the following conditions are met:
+
+1. All DATAGRAMs sent for the subscription have reached their expiration time.
+
+2. All data streams have either been reset, or the stream API has provided a
```
TimEvens, line 1792: Is stream API clarified? Suggest rewording as that makes an assumption of an API.

draft-ietf-moq-transport.md
```diff
@@ -1782,9 +1781,24 @@ FETCH_ERROR
...
+notification that the FIN has been ACKed.
+
+3. Any data streams closed with RESET_STREAM_AT have received notification from
+the stream API that the reliable size has been acknowledged.
```
TimEvens, line 1796: Same regarding using "stream API". It suggests an implementation that this protocol doesn't appear to define.

TimEvens (commented):
Minor comments... but I'm unclear on how `SUBSCRIBE_DONE` is to be used with `Object Status` type **0x4** which indicates end of track.  If the publisher sends end of track object status, isn't that the same as `SUBSCRIBE_DONE`?

martinduke:
> Minor comments... but I'm unclear on how `SUBSCRIBE_DONE` is to be used with `Object Status` type **0x4** which indicates end of track. If the publisher sends end of track object status, isn't that the same as `SUBSCRIBE_DONE`?

No. If the EndOfTrack object is delivered, there might still be other streams out there that are open and delivering earlier objects, or even streams that haven't been opened yet.

draft-ietf-moq-transport.md
martinduke: It's not just an UNSUBSCRIBE_ACK. It's also used when the SUBSCRIBE reaches its end.

draft-ietf-moq-transport.md
martinduke: How would you reword it?

draft-ietf-moq-transport.md
TimEvens: isn't that redundant with object status 0x4/end of track?

TimEvens: TBH, I don't believe it adds much value to ack unsubscribe.  It should be assumed that it was successful as it was transferred over a reliable control stream, which had to be processed by the remote otherwise the control stream would be invalidated.  If the remote doesn't do anything with the unsubscribe (e.g., noop the message) and continues to publish data, the relay should then respond with error.  I don't recommend using protocol violation in this case as that would close a connection causing a potential large impact that spans more than one track. This is a track error and should, IMHO, only affect that track.  With QUIC,  either side can reset a stream regardless of being bidir or unidir, right?  I would suggest that we need a track level error message that is followed by a RESET of the incoming stream. RESET of stream could indicate error code as well.

Out of scope for this PR but keep in mind as it relates, what happens when an app sends end of track or end of group object status and continues to send data that would violate that status???

TimEvens:
> > Minor comments... but I'm unclear on how `SUBSCRIBE_DONE` is to be used with `Object Status` type **0x4** which indicates end of track. If the publisher sends end of track object status, isn't that the same as `SUBSCRIBE_DONE`?
> 
> No. If the EndOfTrack object is delivered, there might still be other streams out there that are open and delivering earlier objects, or even streams that haven't been opened yet.

@martinduke , that needs to be clarified as that's unclear as described. If I understand your statement above, would it be correct to rephrase it as:

> "End of track" (EoT) indicates that this group, subgroup, and object id indicates the last in the track. The track alias **cannot be reused** after this status. Any group, subgroup, object id that is greater than what was indicated at the time of EoT are invalid and would result in a track error. Relays that receive EoT should mark the track as done, but linger it for some period of time to allow for missing/gap data that is inflight and as mentioned in #587?  If by chance the relay or receiving client is tracking gaps and detect no gaps upon receiving EoT, they can immediately close the track without having to wait a linger period for inflight groups/subgroups pending.

draft-ietf-moq-transport.md
TimEvens: I believe the context here is the QUIC transport stack/API, correct?  If so, I would recommend "QUIC transport" instead of "stream API"

draft-ietf-moq-transport.md
martinduke: It is not redundant to end of track, as other objects could be delivered later and out of order, and sometimes the subscribe ends before the track does

martinduke:
After more thought, I really think #628 (Alan's stream-counting PR) is a superior approach, as it gives the option for the client to short-cut its timer. Some were concerned about messing up counting; worst case is it falls back to using the timer, which is this proposal.

Issue #458, opened 2024-06-06
Session Initialization Clarification with datagrams


TimEvens:
Section 3.3 states *"The first stream opened is a client-initiated bidirectional control stream where the peers exchange Setup messages ([Section 6.2](https://datatracker.ietf.org/doc/html/draft-ietf-moq-transport-04#message-setup))"* but does not state what we should do about datagrams being sent before the control stream as well as before **CLIENT_SETUP**/**SERVER_SETUP**.  

Should the connection be closed by protocol violation if DATAGRAM is received matching **any** of the following conditions:

* Control bi-dir stream not yet created; session hasn't been established
* `CLIENT_SETUP` and `SERVER_SETUP` not compete; session hasn't been established
* `CLIENT_SETUP` role is `0x02` **SUBSCRIBER**; publishing is not allowed

ianswett:
If you get a Datagram or Unidirectional stream with Objects from a subscription or Fetch you've never made, then it's a protocol violation.

@TimEvens can you write a PR to this effect?

PR #616, opened 2024-11-15
Invalid Track Alias/Subscribe ID are a protocol violation

afrind:
Fixes: #458

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2194,6 +2194,12 @@ the datagram.
 An endpoint that receives an unknown stream or datagram type MUST close the
 session.
 
+The publisher only sends Objects after receiving a SUBSCRIBE or FETCH.  The
+publisher MUST NOT send unsolicited objects.  If an endpoint receives an
+unsolicited, it SHOULD terminate the session with a protocol violation.  Objects
+that arrive after a subscription or fetch has been cancelled are not considered
+unsolicited.
+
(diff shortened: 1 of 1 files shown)


martinduke:
The lack of normative language is telling here. Killing a session because a subscriber has set its timer too short ("recently closed") is overkill and can lead to failures.

If the intent is to throw an error when there has not been a SUBSCRIBE or FETCH at all, then write that. If it's to throw an error when a SUBSCRIBE or FETCH has never been opened, then say the Sender MUST NOT do that, and the receiver SHOULD do a protocol error if it has the ability to detect that.

If it's to strictly ban the sending of data for any SUBSCRIBE/FETCH that has closed, it can be a MUST  NOT for the sender, but I don't see how you can enforce that at the receiver except to reset the stream, because data can arrive arbitrarily late.

suhasHere (approved):

fluffy (approved):
Looks close enough to me and we can add more later if needed.

draft-ietf-moq-transport.md
```diff
@@ -2194,6 +2194,12 @@ the datagram.
 An endpoint that receives an unknown stream or datagram type MUST close the
 session.
 
+The publisher only sends Objects after receiving a SUBSCRIBE or FETCH.  The
+publisher MUST NOT send unsolicited objects.  If an endpoint receives an
```
ianswett, line 2198: I think I prefer "unrequested objects" here and below, but no strong preference.

ianswett (approved):

Issue #462, opened 2024-06-11
Limits on the amount of buffered data


vasilvv:
If the relay is sending data to a subscriber, and the link between the subscriber and the relay cannot sustain the bitrate of the track, eventually the relay will have to buffer more and more data.  At some point, the relay would have to drop the data; we should specify how that happens (lowest priority stream gets dropped first?).

afrind:
Individual Comment:

+1  We need to specify what happens in this case.  I think there's two parts: 

1) selecting what gets dropped (this is a priority question)
2) the mechanism for dropping (stream resets, etc).

englishm:
+1

Expanding Alan's list, I think we should determine:

- (1a?) valid boundaries for dropping data (group? object? byte? bit?)
- (2a?) under what circumstances the subscriber needs to be informed of the dropped data (do we need different verbs with different semantics in this case, e.g. fetch/subscribe? is dropping data always permitted and something all subscribers must be able to tolerate?)
- (2b?) the means of informing subscribers of dropped data (implied or explicit)

PR #620, opened 2024-11-15
Add an error code for Too Far Behind

afrind:
This error code can be sent by a publisher to a subscriber that is not consuming data fast enough, causing a queue to form at the publisher that exceeds the publisher's limits.

Fixes: #462
Fixes: #582

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -803,6 +803,8 @@ SUBSCRIBE_DONE, as defined below:
 |------|---------------------------|
 | 0x6  | Expired                   |
 |------|---------------------------|
+| 0x7  | Too Far Behind            |
+|------|---------------------------|
 
 ### Graceful Publisher Relay Switchover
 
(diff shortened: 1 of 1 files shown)


afrind:
I added text here on when to send Too Far Behind that attempts to clarify the fidelity of subscribe.  It ended up in the delivery timeout section, which may not be where it belongs, but we don't have a generic section that explains how subscriptions work, it's spread amongst the relay section, SUBSCRIBE message and Delivery Timeout.

mengelbart (approved):

draft-ietf-moq-transport.md
```diff
@@ -1065,8 +1067,14 @@ If an earlier Object arrives later than subsequent Objects, relays can consider
 the receipt time as that of the next later Object, with the assumption that the
 Object's data was reordered.
 
-If neither the subscriber or publisher specify DELIVERY TIMEOUT, Objects are
-delivered as indicated by their Group Order and Priority.
+If neither the subscriber or publisher specify DELIVERY TIMEOUT, all Objects
+in the track matching the subscription filter are delivered as indicated by
+their Group Order and Priority.  If a subscriber exceeds the publisher's
+resource limits by failing to consume objects at a sufficient rate, the
+publisher MAY terminate the subscription with Too Far Behind.
```
ianswett, line 1074: I think we've been putting single quotes around errors?
```suggestion
publisher MAY terminate the subscription with error 'Too Far Behind'.
```

ianswett (approved):

PR #623, opened 2024-11-15
Datagram FETCH subgroup id=object ID

### LINKED ISSUE #600: What is the Subgroup ID of a Datagram FP object?
if a Datagram object is delivered via FETCH (which doesn't really make sense, see #599 and #588), what Subgroup ID do I put in the FETCH object header?

The only answers that make sense are zero, or the object ID.

afrind:
Fixes #600

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2280,8 +2280,8 @@ Each object sent on a fetch stream after the FETCH_HEADER has the following form
 
 The Object Status field is only sent if the Object Payload Length is zero.
 
-The Subgroup ID of an object with a Forwarding Preference of "Datagram"
-(see {{object-fields}}) is zero.
+The Subgroup ID field of an object with a Forwarding Preference of "Datagram"
+(see {{object-fields}}) is set to the Object ID.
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

afrind:
> I'm fine with this, since it resolves the issue, but I would think having all Datagrams in a Group have the same subgroup ID would indicate I should be sending them over a single stream? Making the subgroup ID the same as the Object ID is more in keeping with the SubGroup == Stream transmission model.

Oh that makes more sense.  Let's add this to the issue and see if others agree there.

martinduke (approved):

afrind:
@suhasHere @martinduke you both approved when this PR set the value to 0 but I changed it.  Please take another look.

mengelbart (approved):

ianswett (approved):
I'm fine with this, since it resolves the issue, but I would think having all Datagrams in a Group have the same subgroup ID would indicate I should be sending them over a single stream?  Making the subgroup ID the same as the Object ID is more in keeping with the SubGroup == Stream transmission model.

Issue #551, opened 2024-09-26
Is a namespace tuple with >32 elements a protocol violation?


DanielFay22:
Current definition specifies namespace is an N-tuple, with 1<=N<=32. If the server/relay receives a namespace with N>32 is it expected to return a protocol violation, or do we leave that undefined?

LPardue:
+1 I think we should address this as part of refactoring the tuple type

afrind:
I put up a PR addressing the issue but didn't do a larger refactor of the tuple type, which seems editorial.

PR #626, opened 2024-11-16
Protocol Violation for namespace tuples outside 1..32

afrind:
Fixes: #551

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1597,8 +1597,9 @@ against track namespaces known to the publisher.  For example, if the publisher
 is a relay that has received ANNOUNCE messages for namespaces ("example.com",
 "meeting=123", "participant=100") and ("example.com", "meeting=123",
 "participant=200"), a SUBSCRIBE_ANNOUNCES for ("example.com", "meeting=123")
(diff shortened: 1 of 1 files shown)


DanielFay22 (approved):

draft-ietf-moq-transport.md
```diff
@@ -395,6 +395,8 @@ namespace.
 Track namespace is an ordered N-tuple of bytes where N can be between 1 and 32.
 The structured nature of Track Namespace allows relays and applications to
 manipulate prefixes of a namespace. Track name is a sequence of bytes.
+If an endpoint receives a Track Namespace tuple with N less than 1 or more
```
LPardue, line 398: Would it better to put this receiver requirement next to the bounds definition further down? Right now it's expressed slightly differently and about 1000 lines away

afrind: The reason it's in two places is that the latter one is in the definition of SUBSCRIBE_ANNOUNCES, which takes a Namespace Prefix, while the definition here is for Track Namespace.  I can align the language though.

ianswett (approved):
Editorial suggestion, but LG

LPardue (approved):

suhasHere (approved):

PR #628, opened 2024-11-19
SUBSCRIBE_DONE with Stream Count

afrind:
This is an alternative to #609 that requires a dependency on RESET_STREAM_AT.

Some lines reflowed so the delta looks larger than it is.


martinduke:
So how does this work if Webtrans is over HTTP/2, where there is no RESET_STREAM_AT?

afrind:
> So how does this work if Webtrans is over HTTP/2, where there is no RESET_STREAM_AT?

H2 WT RESET_STREAM has reliable size.

suhasHere (approved):

ianswett (approved):

PR #635, opened 2024-12-03
Clarify End of track

### LINKED ISSUE #505: ID of EndOfTrackObjects
7.1.1.1 currently says:

> 0x4 := Indicates end of Track and Group. GroupID is one greater than the largest group produced in this track and the ObjectId is one greater than the largest object produced in that group. This is sent right after the last object in the track. This SHOULD be cached.

This is confusing. I *think* this means that if the last real object in a track is (10,3), then we'll send an end of Track and Group object with sequence (11, 4)?

I can imagine use cases where the publisher ends the group, and then finds out the track is ending. So would it send EndOfGroup(10,4) and EndOfTrackAndGroup(11, 5)?

Maybe we could instead separate EndOfGroup and EndOfTrack and make EOT's sequence {last_group + 1, 0}?

martinduke:
Fixes #505.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2123,7 +2123,7 @@ are beyond the end of a group or track.
          the largest group produced in this track and the ObjectId is
          zero. An object with this status that has a Group ID less than or
          equal to any other Group ID, or an Object ID other than zero, is
-         a protocol error, and the receiver MUST terminate the session. 
+         a protocol error, and the receiver MUST terminate the session.
 
 Any other value SHOULD be treated as a protocol error and terminate the
 session with a Protocol Violation ({{session-termination}}).
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2119,7 +2119,14 @@ are beyond the end of a group or track.
          in this Group. This SHOULD be cached. A publisher MAY use an end of
          Group object to signal the end of all open Subgroups in a Group.
 
-* 0x4 := Indicates end of Track. GroupID is one greater than the largest group
+* 0x4 := Indicates end of Track and Group. GroupID is the largest group produced
+         in this track and the ObjectId is one greater than the largest object
+         produced in this track. An object with this status that has a Group ID
+         less than any other Group ID, or an Object ID less than or equal to the
+         largest in the group, is a protocol error, and the receiver MUST
+         terminate the session. This SHOULD be cached.
+
+* 0x5 := Indicates end of Track. GroupID is one greater than the largest group
          produced in this track and the ObjectId is zero. An object with this
          status that has a Group ID less than or equal to any other Group ID, or
          an Object ID other than zero, is a protocol error, and the receiver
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2121,9 +2121,9 @@ are beyond the end of a group or track.
 
 * 0x4 := Indicates end of Track and Group. GroupID is one greater than
          the largest group produced in this track and the ObjectId is
-         one greater than the largest object produced in that
-         group. This is sent right after the last object in the
-         track. This SHOULD be cached.
+         zero. An object with this status that has a Group ID less than or
+         equal to any other Group ID, or an Object ID other than zero, is
+         a protocol error, and the receiver MUST terminate the session.
```
afrind, line 2126: Individual Comment: 

It's weird that this ends "track and group" but has a new group ID -- which group did it end?

I suggest we rename it to just "end of track" if we're going to spell it like this.  If we want a single status to end both group and track, it needs to have the group ID of the last group and object ID > previous object in this group.

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -2119,11 +2119,11 @@ are beyond the end of a group or track.
...
-* 0x4 := Indicates end of Track and Group. GroupID is one greater than
-         the largest group produced in this track and the ObjectId is
-         one greater than the largest object produced in that
-         group. This is sent right after the last object in the
-         track. This SHOULD be cached.
+* 0x4 := Indicates end of Track. GroupID is one greater than the largest group
+         produced in this track and the ObjectId is zero. An object with this
+         status that has a Group ID less than or equal to any other Group ID, or
+         an Object ID other than zero, is a protocol error, and the receiver
+         MUST terminate the session.
```
ianswett, line 2126:
```suggestion
         MUST terminate the session. This SHOULD be cached.
```

draft-ietf-moq-transport.md
```diff
@@ -2119,11 +2119,11 @@ are beyond the end of a group or track.
...
-* 0x4 := Indicates end of Track and Group. GroupID is one greater than
-         the largest group produced in this track and the ObjectId is
-         one greater than the largest object produced in that
-         group. This is sent right after the last object in the
-         track. This SHOULD be cached.
+* 0x4 := Indicates end of Track. GroupID is one greater than the largest group
+         produced in this track and the ObjectId is zero. An object with this
+         status that has a Group ID less than or equal to any other Group ID, or
+         an Object ID other than zero, is a protocol error, and the receiver
+         MUST terminate the session. This SHOULD be cached.
```
afrind, line 2126: Individual Comment:

The new formulation of END_OF_TRACK always has to be sent in it's own subgroup/stream, because it's in a new group.  I think the intention behind END_OF_TRACK_AND_GROUP was to prevent this.

1.  Should we specify subgroup 0 for END_OF_TRACK or just let it be whatever?
2. Can we add back END_OF_TRACK_AND_GROUP (0x4) with group ID=largest Group and objectID = largest Object +1, and move END_OF_TRACK to (0x5)

Issue #598, opened 2024-10-25
SUBSCRIBE(current group) splits past and future


afrind:
One of the principles the group seemed to agree on is that FETCH is for the past and SUBSCRIBE is for the future.  Draft-07 however left in SUBSCRIBE(current group) which spans both and seems to violate the principle.  I understand that subscribing starting from the current group is likely to be a very common use case, as groups are MOQT join points.

My concern is that all the reasons we decided FETCH semantics were good for past data are being discarded here.  If the current group consists of an large number of subgroups or datagrams, the sender can overwhelm the receiver.  If the relay has a cache with gaps, we have not explained the interaction.

My strawman is to respond to SUBSCRIBE(current group) with both a FETCH OK and SUBSCRIBE OK, and objects from (current group, 0) to (current group, live edge) are returned in a FETCH stream using those semantics.

englishm:
My thought was to resolve this with a new message, JOIN, which is decomposed into a FETCH+SUBSCRIBE at a publisher with sufficient local knowledge to do so.

englishm:
I'll make slides for Dublin

ianswett:
I'm not arguing for a specific approach, but people did not like removing current group in the past, ie: https://github.com/moq-wg/moq-transport/pull/510#pullrequestreview-2260597641

A large number of datagrams is a congestion issue, but not a flow control one.  For congestion, I'd expect the relay could lower the subscriber priority once it had too much data for a subscription?

If there's only one peep in a group, then starting at the current group doesn't create any problems.  If the peeps were all going to be open at once anyway (ie: 3 or 4 enhancement layers), then the stream usage is no different than starting at the current object.  As the number of peeps increases, it does start being problematic.  If you're doing one Object per peep, for example.

I'd rather not add a 3rd way to retrieve Objects, but I can see an argument for merging the existing functionality of FETCH and SUBSCRIBE and saying:
- You're getting things before the latest Object on a single stream, and ones after it in peeps or datagrams.

If we want to use track alias, I think we still need to prohibit having two "JOINs" that end beyond the current object?  Alternatively, we could just say that you'll only receive one copy of things after the current object, even if there are multiple matching JOINs/SUBSCRIBEs/etc?

vasilvv:
The problem here is that since the groups are arbitrarily large, asking to subscribe starting with object 0 of the current group is not that different from starting at arbitrary group in the past.

wilaw:
[...]

Issue #358, opened 2024-02-07
Tighten definition of ObjectID and GroupID 


wilaw:
Current prose should be updated to indicate that 
1. ObjectID and GroupID are integers. The Object header implicitly defines this, but it should be stated explicitly.
2. ObjectID MUST only increment by 1
3. The GroupID MUST always increase. 
4. The GroupID may start at any value >= 0
5. The GroupID can increase by any value >= 1.

suhasHere:
on 2 , i don't see a need for us to enforce monotonic unitary sequence for objectIds either. If the application needs it for certain purposes, it will generate the objectIds to match that pattern.

I would just say GroupId and ObjectId needs to be increasing and SHOULD NOT be duplicated. (fine to not even have the latter part)

vitaly-castLabs:
Wouldn't it be beneficial to have ObjectIDs incrementing strictly by 1 in order to detect dropped objects? It's an extremely common scenario in video decoding: run into a gap in media and stop decoding until a new key frame arrives.
Are there any real-life scenarios when arbitrary increments might be necessary?

suhasHere:
Its application defined.  Even with today's video applications, one can have b-frames sent over different track/different stream encoding and thus causing gaps within a group. Only application is aware of that end to end and MoQT shouldn't enforce application rules.

martinduke:
Would ObjectID have to start at zero?

martinduke:
Will discuss this in Boston.

kixelated:
Now that there's an object status code of 0x1 to signal gaps, object ID MUST start at 0 and increment by 1.

fluffy:
The more I think about, I'm not seeing why the moq layer needs any restrictions on any of these. Yes, an applications like warp on top might define how they are used but behavior of things at the moq level does not seem to require constraints on any of this. I guess I am trying to understand what problem we are trying to solve ?

martinduke:
It makes it quite a bit easier to implement. If I'm doing stream-per-group, I know that object ID 0 starts a group and a stream.

afrind:
Individual Comment:

These bullets seem like editorial clarifications that are implicit:

> ObjectID and GroupID are integers. The Object header implicitly defines this, but it should be stated explicitly.

[...]

PR #638, opened 2024-12-04
Joining Fetch

englishm:
This PR adds the desired "Join" functionality as [discussed](https://meetecho-player.ietf.org/playout/?session=IETF121-MOQ-20241104-1730) [at IETF 121 in Dublin](https://datatracker.ietf.org/meeting/121/session/moq)[^1]

Adding this functionality to Fetch unblocks further simplifications of Subscribe, allowing us to more clearly delineate between "past" (Fetch) and "future" (Subscribe). (see https://github.com/moq-wg/moq-transport/issues/598, https://github.com/moq-wg/moq-transport/pull/510#pullrequestreview-2260597641, etc.)

There are several ways we could allow a publisher to "atomically" join a Fetch and a Subscribe.
At IETF 121, @wilaw presented [slides](https://datatracker.ietf.org/meeting/121/materials/slides-121-moq-join-api-proposal-00) showing API options:
1. a new macro-like singular JOIN message that is decomposed at the publisher into both a SUBSCRIBE and a FETCH
2. a modified form of FETCH that can be joined together with an existing Subscribe 

The consensus of the discussion seemed to heavily favor the latter design, so, as was requested, this PR is written as a modified form of Fetch: **Joining Fetch**.

This PR:
- Adds "**Fetch Type**" field to Fetch messages
- Redefines pre-existing behavior as a Fetch Type: "**Standalone Fetch**"
- Defines new Fetch Type: "**Joining Fetch**"
- Defines new error code: "**Invalid Subscribe ID**"

[^1]: [21:18 ](https://youtu.be/zs5Y424tASE?feature=shared&t=1278) in the recording. Note: the [Meetecho page](https://meetecho-player.ietf.org/playout/?session=IETF121-MOQ-20241104-1730) includes the chat, synced to the video.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1522,7 +1522,7 @@ Standalone Fetch (0x1) : A Fetch of Objects performed independently of any Subsc
 
 Joining Fetch (0x2) : A Fetch joined together with a Subscribe. A Joining Fetch
 shares the same Subscribe ID as an already-sent Subscribe. A publisher receiving a
-Joining Fetch should use properties of the associated Subscribe to determine the
+Joining Fetch uses properties of the associated Subscribe to determine the
 Track Namespace, Track, StartGroup, StartObject, EndGroup, and EndObject for the
 Joining Fetch such that it is contiguous with the associated Subscribe and begins
 Preceding Group Offset prior.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1607,7 +1607,7 @@ Fields present only for Standalone Fetch (0x1):
 * EndObject: The end Object ID, plus 1. A value of 0 means the entire group is
 requested.
 
-Field present only for Joining Fetch (0x2):
+Fields present only for Joining Fetch (0x2):
 
 * Joining Subscribe ID: The Subscribe ID of the existing subscription to be
 joined. If a publisher receives a Joining Fetch with a Subscribe ID that does
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1536,7 +1536,8 @@ The following values are used:
   * For Subscribes with Filter Type LatestObject or LatestGroup, this is equal to Largest Group ID.
   * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange, this is equal to the StartGroup field of the Subscribe message
 * Resolved Subscribe Start Object:
-  * For Subscribes with Filter Type LatestObject or LatestObject, this is equal to Largest Object ID.
+  * For Subscribes with Filter Type LatestObject, this is equal to Largest Object ID.
+  * For Subscribes with Filter Type LatestGroup, this is 0
   * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange, this is equal to the StartObject field of the Subscribe message
 * Preceding Group Offset: A field in the Joining Fetch message indicating the relative offset from the start of the Subscribe
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1540,7 +1540,7 @@ The following values are used:
   * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange, this is equal to the StartObject field of the Subscribe message
 * Preceding Group Offset: A field in the Joining Fetch message indicating the relative offset from the start of the Subscribe
 
-Note: If a relay does not yet have LatestGroup and LatestObject for a given track, it may choose to either forward both the Subscribe and
+Note: If a relay does not yet have LatestGroup and LatestObject for a given track, it can choose to either forward both the Subscribe and
 the Joining Fetch upstream or to wait until the Joining Fetch can be resolved locally. In either case, the Resolved Subscribe Start values
 for a Joining Fetch MUST correspond to the Subscribe within the same session so the Fetch and Subscribe can be contiguous and non-overlapping.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1546,7 +1546,7 @@ The following values are used:
 The Resolved Subscribe Start values for a Joining Fetch MUST correspond to the referenced Subscribe within the same session so that the Objects delivered across the Fetch and Subscribe are contiguous and non-overlapping.
 If a relay answers the referenced Subscribe with a `SUBSCRIBE_OK` that has ContentExists set to 0, it MUST respond to the Joining Fetch with a `FETCH_ERROR`.
 
-Using that information and the following algorithm, these values are computed:
+Using the above information and the following algorithm, these values are computed:
 
 * Fetch Start Group: The StartGroup for the Fetch
 * Fetch Start Object: The StartObject for the Fetch (always 0)
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1542,7 +1542,7 @@ with the following fields dynamically determined from the corresponding Subscrib
 
 * EndGroup: StartGroup of the corresponding Subscribe minus 1 (double check minus 1)
 
-* EndObject: StartObject of the corresponding SUBSCRIBE minus 1 (double check minus 1)
+* EndObject: StartObject of the corresponding Subscribe minus 1 (double check minus 1)
 
 A Joining Fetch MUST be sent in ascending group order.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1474,7 +1474,7 @@ Fields common to all Fetch Types:
 
 * Subscribe ID: The Subscribe ID identifies a given fetch request. Subscribe ID
 is a variable length integer that MUST be unique and monotonically increasing
-within  a session. For a Standalone Fetch a new Subscribe ID MUST be used. For
+within a session. For a Standalone Fetch a new Subscribe ID MUST be used. For
 a Joining Fetch, the Subscribe ID MUST correspond to a Subscribe which has already
 been sent. If a publisher receives a Joining Fetch with a Subscribe ID that does
 not correspond to an existing Subscribe, it MUST close the session with an
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1536,7 +1536,7 @@ with the following fields dynamically determined from the corresponding Subscrib
 
 * Track Name: Same as in the corresponding Subscribe
 
-* StartGroup: LatestGroup as determined for the SUBSCRIBE minus PreviousGroups from the JOIN
+* StartGroup: LatestGroup as determined for the Subscribe minus Previous Groupe Count from the Joining Fetch
 
 * StartObject: Always 0
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1542,28 +1542,20 @@ The following values are used:
 The Resolved Subscribe Start values for a Joining Fetch MUST correspond to the referenced Subscribe within the same session so that the ranges of Objects covered by the Fetch and Subscribe are contiguous and non-overlapping.
 If a relay answers the referenced Subscribe with a `SUBSCRIBE_OK` that has ContentExists set to 0, it MUST respond to the Joining Fetch with a `FETCH_ERROR`.
 
-Using the above information and the following algorithm, these values are computed:
-
-* Fetch Start Group: The StartGroup for the Fetch
-* Fetch Start Object: The StartObject for the Fetch (always 0)
-* Fetch End Group: The EndGroup for the Fetch
-* Fetch End Object: The EndObject for the Fetch (represented as 1 more than the end Object ID
-  or a value of 0 indicating that the entire group is requested)
-
 The publisher receiving a Joining Fetch computes the fetch range as follows:
 
-* Fetch Start Group: Resolved Subscribe Start Group - Preceding Group Offset
-* Fetch Start Object: 0
+* Fetch StartGroup: Resolved Subscribe Start Group - Preceding Group Offset
+* Fetch StartObject: 0
 
 If Resolved Subscribe Start Object is 0:
-* Fetch End Group: Resolved Subscribe Start Group - 1
-* Fetch End Object: 0 (all objects in the group)
+* Fetch EndGroup: Resolved Subscribe Start Group - 1
+* Fetch EndObject: 0 (all objects in the group)
 
 Else, if Resolved Subscribe Start Object is 1 or more:
-* Fetch End Group: Resolved Subscribe Start Group
-* Fetch End Object: Resolved Subscribe Start Object
+* Fetch EndGroup: Resolved Subscribe Start Group
+* Fetch EndObject: Resolved Subscribe Start Object
 
-If Fetch End Group < Fetch Start Group respond with a Fetch Error.
+If Fetch EndGroup < Fetch StartGroup respond with a `FETCH_ERROR`.
 
 ## FETCH_CANCEL {#message-fetch-cancel}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1430,7 +1430,7 @@ Joining Fetch (0x2) : A Fetch joined together with a Subscribe. A Joining Fetch
 shares the same Subscribe ID as an already-sent Subscribe. A publisher receiving a Joining
 Fetch should use properties of the associated Subscribe to determine the Track Namespace,
 Track, StartGroup, StartObject, EndGroup, and EndObject for the Joining Fetch such that it is
-contiguous with the associated Subscribe and begins Previous Group Count prior.
+contiguous with the associated Subscribe and begins Preceding Group Offset prior.
 
 A Fetch Type other than the above MUST be treated as an error.
 
@@ -1463,7 +1463,7 @@ FETCH Message {
    StartObject (i),
    EndGroup (i),
    EndObject (i),]
-  [Previous Group Count (i),]
+  [Preceding Group Offset (i),]
   Number of Parameters (i),
   Parameters (..) ...
 }
@@ -1511,7 +1511,7 @@ requested.
 
 Field present only for Joining Fetch (0x2):
 
-* Previous Group Count: The number of groups to Fetch prior to the StartGroup of the corresponding Subscribe
+* Preceding Group Offset: The group offset for the Fetch prior and relative to the StartGroup of the corresponding Subscribe
 
 Objects that are not yet published will not be retrieved by a FETCH.
 The latest available Object is indicated in the FETCH_OK, and is the last
@@ -1532,7 +1532,7 @@ subgroup ID is not used for ordering.
 ### Calculating the Range of a Joining Fetch
 
 A publisher which receives a Fetch message with a Fetch Type of 0x2 must treat it as a Fetch
-with a range dynamically determined by the Previous Group Count (PGC) field and values derived
+with a range dynamically determined by the Preceding Group Offset (PGO) field and values derived
 from to the corresponding Subscribe message (hereafter "the Subscribe").
 
 The following values are used:
@@ -1544,7 +1544,7 @@ The following values are used:
 * Resolved Subscribe Start Object:
   * For Subscribes with Filter Type LatestObject or LatestObject, this is equal to Largest Object ID.
   * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange, this is equal to the StartObject field of the Subscribe message
-* Previous Group Count: A field in the Joining Fetch message indicating how many prior groups to include
+* Preceding Group Offset: A field in the Joining Fetch message indicating the relative offset from the start of the Subscribe
 
 Using that information and the following algorithm, these values are computed:
 
@@ -1559,7 +1559,7 @@ The publisher receiving a Joining Fetch computes the fetch range as follows:
 If ContentExists is not 1 and Largest Group ID and Largest Object ID are not available, a relay SHOULD
 forward the Fetch with the 0x2 Fetch Type upstream.
 
-* Fetch Start Group: Resolved Subscribe Start Group - Previous Group Count
+* Fetch Start Group: Resolved Subscribe Start Group - Preceding Group Offset
 * Fetch Start Object: 0
 
 If Resolved Subscribe Start Object is 0:
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1597,7 +1597,8 @@ The Largest Group ID and Largest Object ID values from the corresponding
 subscription are used to calculate the end of a Joining Fetch so the Objects
 retrieved by the FETCH and SUBSCRIBE are contiguous and non-overlapping.
 If no Objects have been published for the track, and the SUBSCRIBE_OK has a
-ContentExists value of 0, the publisher responds with a FETCH_ERROR.
+ContentExists value of 0, the publisher responds with a FETCH_ERROR with
+error code 'No Objects'.
 
 The publisher receiving a Joining Fetch computes the range as follows:
 
@@ -1605,6 +1606,10 @@ The publisher receiving a Joining Fetch computes the range as follows:
 * Fetch StartObject: 0
 * Fetch EndGroup: Subscribe Largest Group
 * Fetch EndObject: Subscribe Largest Object
+
+A Fetch EndObject of 0 requests the entire group, but Fetch will not
+retrieve Objects that have not yet been published, so 1 is subtracted from
+the Fetch EndGroup if Fetch EndObject is 0.
 
 ## FETCH_CANCEL {#message-fetch-cancel}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1661,9 +1661,6 @@ Else, if Resolved Subscribe Start Object is 1 or more:
 * Fetch EndGroup: Resolved Subscribe Start Group
 * Fetch EndObject: Resolved Subscribe Start Object
 
-If Fetch EndGroup < Fetch StartGroup, there are no Objects to retrieve,
-so respond with a `FETCH_ERROR`.
-
 ## FETCH_CANCEL {#message-fetch-cancel}
 
 A subscriber issues a `FETCH_CANCEL` message to a publisher indicating it is no
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1520,12 +1520,12 @@ There are two types of Fetch messages:
 
 Standalone Fetch (0x1) : A Fetch of Objects performed independently of any Subscribe.
 
-Joining Fetch (0x2) : A Fetch joined together with a Subscribe. A Joining Fetch
-shares the same Subscribe ID as an already-sent Subscribe. A publisher receiving a
+Joining Fetch (0x2) : A Fetch joined together with a Subscribe by specifying
+the Subscribe ID of an active subscription. A publisher receiving a
 Joining Fetch uses properties of the associated Subscribe to determine the
-Track Namespace, Track, StartGroup, StartObject, EndGroup, and EndObject for the
-Joining Fetch such that it is contiguous with the associated Subscribe and begins
-Preceding Group Offset prior.
+Track Namespace, Track, StartGroup, StartObject, EndGroup, and EndObject such that
+it is contiguous with the associated Subscribe. The Joining Fetch begins the
+Preceding Group Offset prior to the associated subscription.
 
 A Subscriber can use a Joining Fetch to, for example, fill a playback buffer with a
 certain number of groups prior to the live edge of a track.
@@ -1533,7 +1533,7 @@ certain number of groups prior to the live edge of a track.
 A Joining Fetch is only permitted when the associated Subscribe has the Filter
 Type Latest Object.
 
-A Fetch Type other than the above MUST be treated as an error.
+A Fetch Type other than 0x1 or 0x2 MUST be treated as an error.
 
 A publisher responds to a FETCH request with either a FETCH_OK or a FETCH_ERROR
 message.  If it responds with FETCH_OK, the publisher creates a new unidirectional
@@ -1564,7 +1564,8 @@ FETCH Message {
    StartObject (i),
    EndGroup (i),
    EndObject (i),]
-  [Preceding Group Offset (i),]
+  [Joining Subscribe ID (i),
+   Preceding Group Offset (i),]
   Number of Parameters (i),
   Parameters (..) ...
 }
@@ -1575,12 +1576,7 @@ Fields common to all Fetch Types:
 
 * Subscribe ID: The Subscribe ID identifies a given fetch request. Subscribe ID
 is a variable length integer that MUST be unique and monotonically increasing
-within a session. For a Standalone Fetch a new Subscribe ID MUST be used. For
-a Joining Fetch, the Subscribe ID MUST correspond to a Subscribe which has already
-been sent. If a publisher receives a Joining Fetch with a Subscribe ID that does
-not correspond to an existing Subscribe, it MUST respond with a Fetch Error.
-Though they share an ID, cancelling or updating the subscription has no impact
-on the Fetch and vice versa.
+within a session.
 
 * Subscriber Priority: Specifies the priority of a fetch request relative to
 other subscriptions or fetches in the same session. Lower numbers get higher
@@ -1612,6 +1608,10 @@ Fields present only for Standalone Fetch (0x1):
 requested.
 
 Field present only for Joining Fetch (0x2):
+
+* Joining Subscribe ID: The Subscribe ID of the existing subscription to be
+joined. If a publisher receives a Joining Fetch with a Subscribe ID that does
+not correspond to an existing Subscribe, it MUST respond with a Fetch Error.
 
 * Preceding Group Offset: The group offset for the Fetch prior and relative
 to the StartGroup of the corresponding Subscribe. A value of 0 indicates
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1636,30 +1636,18 @@ as a Fetch with a range dynamically determined by the Preceding Group Offset
 field and field values derived from the corresponding SUBSCRIBE message
 (hereafter "the Subscribe").
 
-The following values are used:
-
-* Resolved Subscribe Start Group: the Largest Group ID of the associated Subscribe.
-* Resolved Subscribe Start Object: the Largest Object ID of the associated Subscribe.
-
-The Resolved Subscribe Start values for a Joining Fetch MUST correspond to the
-referenced Subscribe within the same session so that the ranges of Objects covered
-by the Fetch and Subscribe are contiguous and non-overlapping.
+The Largest Group ID and Largest Object ID values from the corresponding
+subscription are used to calculate the end of a Joining Fetch so the Objects
+retrieved by the Fetch and Subscribe are contiguous and non-overlapping.
 If no Objects have been published for the track, so the SUBSCRIBE_OK has a
-ContentExists value of 0, the publisher MUST respond to the Joining Fetch with a
-FETCH_ERROR.
+ContentExists value of 0, the publisher responds with a FETCH_ERROR.
 
 The publisher receiving a Joining Fetch computes the fetch range as follows:
 
-* Fetch StartGroup: Resolved Subscribe Start Group - Preceding Group Offset
+* Fetch StartGroup: Subscribe Largest Group - Preceding Group Offset
 * Fetch StartObject: 0
-
-If Resolved Subscribe Start Object is 0:
-* Fetch EndGroup: Resolved Subscribe Start Group - 1
-* Fetch EndObject: 0 (all objects in the group)
-
-Else, if Resolved Subscribe Start Object is 1 or more:
-* Fetch EndGroup: Resolved Subscribe Start Group
-* Fetch EndObject: Resolved Subscribe Start Object
+* Fetch EndGroup: Subscribe Largest Group
+* Fetch EndObject: Subscribe Largest Object
 
 ## FETCH_CANCEL {#message-fetch-cancel}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1626,15 +1626,14 @@ field and field values derived from the corresponding SUBSCRIBE message
 The following values are used:
 
 * Resolved Subscribe Start Group:
-  * For Subscribes with Filter Type Latest Object or Latest Group,
-    this is equal to Largest Group ID.
-  * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange,
-    this is equal to the StartGroup field of the Subscribe message
+  * For Latest Object or Latest Group filter types, this is Largest Group ID.
+  * For AbsoluteStart or AbsoluteRange filter types, this is the StartGroup field
+    of the SUBSCRIBE message
 * Resolved Subscribe Start Object:
-  * For Subscribes with Filter Type Latest Object, this is equal to Largest Object ID.
-  * For Subscribes with Filter Type Latest Group, this is 0
-  * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange,
-    this is equal to the StartObject field of the Subscribe message
+  * For the Latest Object filter type, this is the Largest Object ID.
+  * For the Latest Group filter type, this is 0
+  * For AbsoluteStart or AbsoluteRange filter types, this is the StartObject field
+    of the SUBSCRIBE message
 * Preceding Group Offset: A field in the Joining Fetch message indicating the
   relative offset from the start of the Subscribe
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1540,7 +1540,7 @@ with the following fields dynamically determined from the corresponding Subscrib
 
 * StartObject: Always 0
 
-* EndGroup: StartGroup of the corresponding SUBSCRIBE minus 1 (double check minus 1)
+* EndGroup: StartGroup of the corresponding Subscribe minus 1 (double check minus 1)
 
 * EndObject: StartObject of the corresponding SUBSCRIBE minus 1 (double check minus 1)
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1529,20 +1529,47 @@ A publisher MUST send fetched groups in group order, either ascending or
 descending. Within each group, objects are sent in Object ID order;
 subgroup ID is not used for ordering.
 
-A publisher which receives a Fetch message with a Fetch Type of 0x2 should treat it as a Fetch
-with the following fields dynamically determined from the corresponding Subscribe:
+### Calculating the Range of a Joining Fetch
 
-* Track Namespace: Same as in the corresponding Subscribe
+A publisher which receives a Fetch message with a Fetch Type of 0x2 must treat it as a Fetch
+with a range dynamically determined by the Previous Group Count (PGC) field and values derived
+from to the corresponding Subscribe message (hereafter "the Subscribe").
 
-* Track Name: Same as in the corresponding Subscribe
+The following values are used:
 
-* StartGroup: LatestGroup as determined for the Subscribe minus Previous Group Count from the Joining Fetch
+* Filter Type: The type of filter specified in the Subscribe
+* Subscribe Start Group: The Start Group specified in the Subscribe (if AbsoluteStart or AbsoluteRange)
+* Subscribe Start Object: The Start Object specified in the Subscribe (if AbsoluteStart or AbsoluteRange)
+* Largest Group ID: The largest Group ID available for the track (if ContentExists is 1)
+* Largest Object ID: The largest Object ID within the largest Group ID (if ContentExists is 1)
+* Previous Group Count: A field in the Joining Fetch message indicating how many prior groups to include
 
-* StartObject: Always 0
+Using that information and the following algorithm, these values are computed:
 
-* EndGroup: StartGroup of the corresponding Subscribe minus 1 (double check minus 1)
+* Fetch Start Group: The StartGroup for the Fetch
+* Fetch Start Object: The StartObject for the Fetch (always 0)
+* Fetch End Group: The EndGroup for the Fetch
+* Fetch End Object: The EndObject for the Fetch (represented as 1 more than the end Object ID
+  or a value of 0 indicating that the entire group is requested)
 
-* EndObject: StartObject of the corresponding Subscribe minus 1 (double check minus 1)
+The publisher receiving a Joining Fetch computes the fetch range as follows:
+
+If ContentExists is not 1 and Largest Group ID and Largest Object ID are not available, a relay SHOULD
+forward the Fetch with the 0x2 Fetch Type upstream.
+
+1. For Subscribes with a Filter Type of AbsoluteStart or AbsoluteRange:
+* Fetch Start Group: max(Largest Group ID, Subscribe Start Group) - Previous Group Count
+* Fetch Start Object: 0
+* Fetch End Group: the maximum of Largest Group ID and Subscribe Start Group
+* Fetch End Object:
+  * Largest Object ID if Largest Group > Subscribe Start Group
+  * Subscribe Start Object otherwise
+
+2. For Subscribes with a Filter Type of LatestObject or LatestGroup:
+* Fetch Start Group: Largest Group ID - Previous Group Count
+* Fetch Start Object: 0
+* Fetch End Group: Largest Group ID
+* Fetch End Object: Largest Object ID
 
 A Joining Fetch MUST be sent in ascending group order.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1452,11 +1452,11 @@ FETCH Message {
   Length (i),
   Subscribe ID (i),
   Subscriber Priority (8),
+  Group Order (8),
   Fetch Type (i),
   [Track Namespace (tuple),
    Track Name Length (i),
    Track Name (..),
-   Group Order (8),
    StartGroup (i),
    StartObject (i),
    EndGroup (i),
@@ -1481,6 +1481,11 @@ not correspond to an existing Subscribe, it MUST respond with a Fetch Error.
 other subscriptions or fetches in the same session. Lower numbers get higher
 priority. See {{priorities}}.
 
+* Group Order: Allows the subscriber to request Objects be delivered in
+Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
+A value of 0x0 indicates the original publisher's Group Order SHOULD be
+used. Values larger than 0x2 are a protocol error.
+
 * Fetch Type: Identifies the type of Fetch, whether joining or standalone.
 
 * Parameters: The parameters are defined in {{version-specific-params}}.
@@ -1491,11 +1496,6 @@ Fields present only for Standalone Fetch (0x1):
 ({{track-name}}).
 
 * Track Name: Identifies the track name as defined in ({{track-name}}).
-
-* Group Order: Allows the subscriber to request Objects be delivered in
-Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
-A value of 0x0 indicates the original publisher's Group Order SHOULD be
-used. Values larger than 0x2 are a protocol error.
 
 * StartGroup: The start Group ID.
 
@@ -1564,8 +1564,6 @@ Else, if Resolved Subscribe Start Object is 1 or more:
 * Fetch End Object: Resolved Subscribe Start Object
 
 If Fetch End Group < Fetch Start Group respond with a Fetch Error.
-
-A Joining Fetch MUST be sent in ascending group order.
 
 ## FETCH_CANCEL {#message-fetch-cancel}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1531,7 +1531,6 @@ from to the corresponding Subscribe message (hereafter "the Subscribe").
 
 The following values are used:
 
-* ContentExists: whether we have any objects and therefore a Largest Group ID and Largest Object ID for the track
 * Resolved Subscribe Start Group:
   * For Subscribes with Filter Type LatestObject or LatestGroup, this is equal to Largest Group ID.
   * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange, this is equal to the StartGroup field of the Subscribe message
@@ -1540,6 +1539,10 @@ The following values are used:
   * For Subscribes with Filter Type LatestGroup, this is 0
   * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange, this is equal to the StartObject field of the Subscribe message
 * Preceding Group Offset: A field in the Joining Fetch message indicating the relative offset from the start of the Subscribe
+
+Note: If a relay does not yet have LatestGroup and LatestObject for a given track, it may choose to either forward both the Subscribe and
+the Joining Fetch upstream or to watch until the Joining Fetch can be resolved locally. However it is handled, the Resolved Subscribe Start values
+for a Joining Fetch MUST correspond to the Subscribe within the same session so the Fetch and Subscribe can be contiguous and non-overlapping.
 
 Using that information and the following algorithm, these values are computed:
 
@@ -1550,9 +1553,6 @@ Using that information and the following algorithm, these values are computed:
   or a value of 0 indicating that the entire group is requested)
 
 The publisher receiving a Joining Fetch computes the fetch range as follows:
-
-If ContentExists is not 1 and Largest Group ID and Largest Object ID are not available, a relay SHOULD
-forward the Fetch with the 0x2 Fetch Type upstream.
 
 * Fetch Start Group: Resolved Subscribe Start Group - Preceding Group Offset
 * Fetch Start Object: 0
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1540,9 +1540,8 @@ The following values are used:
   * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange, this is equal to the StartObject field of the Subscribe message
 * Preceding Group Offset: A field in the Joining Fetch message indicating the relative offset from the start of the Subscribe
 
-Note: If a relay does not yet have LatestGroup and LatestObject for a given track, it can choose to either forward both the Subscribe and
-the Joining Fetch upstream or to wait until the Joining Fetch can be resolved locally. In either case, the Resolved Subscribe Start values
-for a Joining Fetch MUST correspond to the Subscribe within the same session so the Fetch and Subscribe can be contiguous and non-overlapping.
+The Resolved Subscribe Start values for a Joining Fetch MUST correspond to the referenced Subscribe within the same session so that the Objects delivered across the Fetch and Subscribe are contiguous and non-overlapping.
+If a relay answers the referenced Subscribe with a `SUBSCRIBE_OK` that has ContentExists set to 0, it MUST respond to the Joining Fetch with a `FETCH_ERROR`.
 
 Using that information and the following algorithm, these values are computed:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -585,7 +587,10 @@ code, as defined below:
 * Too Many Subscribes: The session was closed because the subscriber used
   a Subscribe ID equal or larger than the current Maximum Subscribe ID.
 
-* GOAWAY Timeout: The session was closed because the peer took too long to
+* Invalid Subscribe ID: The session was closed because the subscriber sent
+  a Joining Fetch with a Subscribe ID that does not exist.
+
+* GOAWAY Timeout: The session was closed because the client took too long to
```
englishm, line 593: Looks like a one word change (from f89745d) was lost in a rebase here maybe?

```suggestion
* GOAWAY Timeout: The session was closed because the peer took too long to
```

draft-ietf-moq-transport.md
```diff
@@ -1485,6 +1519,23 @@ A publisher MUST send fetched groups in group order, either ascending or
 descending. Within each group, objects are sent in Object ID order;
 subgroup ID is not used for ordering.
 
+A publisher which receives a Fetch message with a Fetch Type of 0x2 should treat it as a Fetch
+with the following fields dynamically determined from the corresponding Subscribe:
+
+* Track Namespace: Same as in the corresponding Subscribe
+
+* Track Name: Same as in the corresponding SUBSCRIBE
```
englishm, line 1527:
```suggestion
* Track Name: Same as in the corresponding Subscribe
```

draft-ietf-moq-transport.md
```diff
@@ -1485,6 +1519,23 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+
+* StartGroup: LatestGroup as determined for the SUBSCRIBE minus PreviousGroups from the JOIN
```
englishm, line 1529:
```suggestion
* StartGroup: LatestGroup as determined for the Subscribe minus Previous Groupe Count from the Joining Fetch
```

draft-ietf-moq-transport.md
```diff
@@ -1485,6 +1519,23 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+
+* StartObject: Always 0
+
+* EndGroup: StartGroup of the corresponding SUBSCRIBE minus 1 (double check minus 1)
```
englishm, line 1533:
```suggestion
* EndGroup: StartGroup of the corresponding Subscribe minus 1 (double check minus 1)
```

draft-ietf-moq-transport.md
```diff
@@ -1485,6 +1519,23 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+
+* EndObject: StartObject of the corresponding SUBSCRIBE minus 1 (double check minus 1)
```
englishm, line 1535:
```suggestion
* EndObject: StartObject of the corresponding Subscribe minus 1 (double check minus 1)
```

englishm (commented):
Need to fix some capitalization inconsistencies

draft-ietf-moq-transport.md
```diff
@@ -1424,34 +1443,48 @@ FETCH Message {
...
 }
 ~~~
 {: #moq-transport-fetch-format title="MOQT FETCH Message"}
 
+Fields common to all Fetch Types:
+
 * Subscribe ID: The Subscribe ID identifies a given fetch request. Subscribe ID
 is a variable length integer that MUST be unique and monotonically increasing
-within  a session.
+within  a session. For a Standalone Fetch a new Subscribe ID MUST be used. For
```
afrind, line 1467: nit: extra space 'within  a'

draft-ietf-moq-transport.md
```diff
@@ -1405,6 +1410,20 @@ objects within a track. The publisher responding to a FETCH is responsible for r
...
+**Fetch Types**
+
+There are two types of Fetch messages:
+
+Standalone Fetch (0x1) : A Fetch of Objects performed independently of any Subscribe.
+
+Joining Fetch (0x2) : A Fetch joined together with a Subscribe. A Joining Fetch
+shares the same Subscribe ID as an already-sent Subscribe. A publisher receiving a Joining
+Fetch should use properties of the associated Subscribe to determine the Track Namespace,
+Track, StartGroup, StartObject, EndGroup, and EndObject for the Joining Fetch such that it is
```
afrind, line 1422: I don't think EndGroup and EndObject are used by the joining fetch.

draft-ietf-moq-transport.md
```diff
@@ -1485,6 +1519,23 @@ A publisher MUST send fetched groups in group order, either ascending or
...
 subgroup ID is not used for ordering.
 
+A publisher which receives a Fetch message with a Fetch Type of 0x2 should treat it as a Fetch
+with the following fields dynamically determined from the corresponding Subscribe:
+
+* Track Namespace: Same as in the corresponding Subscribe
+
+* Track Name: Same as in the corresponding Subscribe
+
+* StartGroup: LatestGroup as determined for the Subscribe minus Previous Groupe Count from the Joining Fetch
```
afrind, line 1529: spelling: Groupe

afrind, line 1529: How does it work if the relay has no content for the subscription yet?  Does it wait for the first published object to resolve the joining fetch?

draft-ietf-moq-transport.md
```diff
@@ -1485,6 +1519,23 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+
+* StartObject: Always 0
+
+* EndGroup: StartGroup of the corresponding Subscribe minus 1 (double check minus 1)
```
afrind, line 1533: With this definition you can have EndGroup < StartGroup --

SUBSCRIBE(10, 0)
JOINING_FETCH(1)
SUBSCRIBE_OK(20, 0)

StartGroup = 19
EndGroup = 9

At least you can if StartGroup in the past is not an error, which I don't think it should be based on how hard it is to guess where Latest is.

draft-ietf-moq-transport.md
```diff
@@ -1485,6 +1519,23 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+
+* EndObject: StartObject of the corresponding Subscribe minus 1 (double check minus 1)
+
+A Joining Fetch MUST be sent in ascending group order.
```
afrind, line 1537: Seems like we should allow an order param here, right?

draft-ietf-moq-transport.md
afrind, line 1529: I'll note that "Previous Group Count" is doing subtraction on group IDs and applications need to be generally aware what will happen based on the publisher's group ID scheme.  Eg: if the publisher is using group IDs (1000, 2000, 3000), PGC=1 has no effect.

afrind (commented):
Individual Review:

Thanks for this.  I think it's an improvement over "Current Group" in subscribe.

draft-ietf-moq-transport.md
englishm, line 1467:
```suggestion
within a session. For a Standalone Fetch a new Subscribe ID MUST be used. For
```

draft-ietf-moq-transport.md
englishm: If the relay does not have an active upstream subscription for the track already then I think it should probably forward both the Subscribe and the Joining Fetch upstream.

If it hasn't seen the Subscribe ID yet within this downstream session, then that's an error (Invalid Subscribe ID).

englishm: This is a good point. If group IDs are not contiguous (currently allowed) then "count" in this name cannot be an accurate name. Maybe we need a different name here to avoid confusion.

draft-ietf-moq-transport.md
englishm: Say more? The Fetch has to end somewhere.

draft-ietf-moq-transport.md
englishm, line 1529:
```suggestion
* StartGroup: LatestGroup as determined for the Subscribe minus Previous Group Count from the Joining Fetch
```

afrind: I'm asking about the case that the upstream subscribe OK has arrived but no content has been produced yet.

draft-ietf-moq-transport.md
afrind: nvm I misread the text.  I thought you were using the End* from the Subscribe for something.

draft-ietf-moq-transport.md
englishm: The omission was intentional. I think the use cases we're serving here will all want "ascending" order. It seemed better to have a smaller and simpler message.

I'm not wholly opposed to re-adding that field to Joining Fetch if there are use cases for it, but another option could be to add another Fetch Type down the road if we find use cases that need different behavior here.

draft-ietf-moq-transport.md
englishm: Whoops! Good point! I see now that I forgot to resolve my `(double check minus 1)` TODO before submitting this PR and also left some confusing mixed use of LatestGroup and StartGroup here. The intent is to define the fields of the Fetch range to be prior to and contiguous, but non-overlapping with the _resolved_ starting point of the corresponding Subscribe.

Here's the relevant text from the editor's draft's section on Fetch:

> Fetch specifies an inclusive range of Objects starting at StartObject
> in StartGroup and ending at EndObject in EndGroup. EndGroup and EndObject MUST
> specify the same or a later object than StartGroup and StartObject.

> If StartGroup/StartObject is greater than the latest published Object group, the publisher MUST return FETCH_ERROR with error code 'No Objects'.

and Subscribe (as it currently stands):

> Latest Group (0x1) : Specifies an open-ended subscription with objects from the beginning of the current group. If no content has been delivered yet, the subscription starts with the first published or received group.

> Latest Object (0x2): Specifies an open-ended subscription beginning from the current object of the current group. If no content has been delivered yet, the subscription starts with the first published or received group.

> AbsoluteStart (0x3): Specifies an open-ended subscription beginning from the object identified in the StartGroup and StartObject fields. If the StartGroup is prior to the current group, the publisher MUST reply with a SUBSCRIBE_ERROR with code 'Invalid Range'.

> AbsoluteRange (0x4): Specifies a closed subscription starting at StartObject in StartGroup and ending at EndObject in EndGroup. The start and end of the range are inclusive. EndGroup MUST specify the same or a later group than StartGroup. If the StartGroup is prior to the current group, the publisher MUST reply with a SUBSCRIBE_ERROR with code 'Invalid Range'.

and Subscribe OK:

> Largest Group ID: The largest Group ID available for this track. This field is only present if ContentExists has a value of 1.

> Largest Object ID: The largest Object ID available within the largest Group ID for this track. This field is only present if ContentExists has a value of 1.

So, at present it's not permitted to send a Subscribe (AbsoluteStart or AbsoluteRange) with a Group ID < Largest Group ID, even though you can't definitively know what that is ahead of time.

englishm: That said, I think we do still need to clarify this a bit further...

If we consider the case of Latest Group, then our "resolved StartGroup" for the Subscribe will == Largest Group ID and our "resolved StartObject" will == 0.

If we consider the case of Latest Object, then our "resolved StartGroup" will == Largest Group ID and our "resolved StartObject" will == Largest Object ID? Maybe? Or will the first object a subscriber receives from the Subscribe be Largest Object ID plus one? Let's assume for now the latter is true because we want to move in the direction of treating Subscribe more like an entry in a forwarding table and handle already "known" / "cached" objects with Fetch.

Examples:

Groups of 30 objects (0 through 29)
Contiguous, monotonically incrementing Group IDs
Largest Group ID = 20
Largest Object ID = 5

Subscribe(LatestObject) gets 20-6, 20-7, 20-8, ...
Fetch(JoiningFetch,PGC=1) gets 19-0, 19-1, 19-2, ..., 19-29, 20-0, 20-1, ... 20-5
Fetch(JoiningFetch,PGC=0) gets 20-0, 20-1, 20-2, ..., 20-5

And since we still have LatestGroup, let's look at that, too:
Subscribe(LatestGroup) gets: 20-0, 20-1, 20-2, ...
Fetch(JoiningFetch,PGC=1) gets: 19-0, 19-1, 19-2, ... 19-29
Fetch(JoiningFetch,PGC=0) gets: nothing, but not an error?

You wouldn't want to send PGC=0 with a LatestGroup Subscribe, but we'd have a similar case whenever Largest Object ID happens to fall on a group boundary so we should handle it gracefully.

I'll make a pass a rewriting this to match these examples and use "Largest Group ID" / "Largest Object ID" as our  points of reference since we have those defined for Subscribe OK already.

englishm:
> Individual Review:
> 
> Thanks for this. I think it's an improvement over "Current Group" in subscribe.

Thanks for the quick feedback @afrind! 

I'll fix up the definitions for how to resolve the range relative to the Subscribe so the terminology matches the current draft text and the details are more clear.

draft-ietf-moq-transport.md
englishm: As I'm revising these definitions, I've realized that there's an annoying boundary case we need to handle. 

Thinking aloud about how best to handle it...

Example:

Groups of 30 objects (0 through 29)
Contiguous, monotonically incrementing Group IDs
Largest Group ID = 20
Largest Object ID = 29

Assuming a Subscribe with Filter Type LatestObject gets the next object _after_ (Largest Group ID, Largest Object ID) then...

Subscribe(LatestObject) gets 21-0, 21-1, 21-2, ...
Fetch(JoiningFetch,PGC=1) "should" get 20-0, 20-1, 20-2, ..., 20-29
Fetch(JoiningFetch,PGC=0) "should" get nothing (because the subscribe already starts at an object 0)

But we can't know ahead of time that 29 is the last object in the group.

If we define our Joining Fetch range purely in terms relative to (Largest Group ID, Largest Object ID) consistently with the other examples above we'd end up fetching an entire extra Group.

Subscribe(LatestObject) gets 21-0, 21-1, 21-2, ...
Fetch(JoiningFetch,PGC=1) gets 19-0, 19-1, 19-2, ..., 20-29
Fetch(JoiningFetch,PGC=0) gets 20-0, ..., 20-29

I think the simplest option would be to just accept that and say that the subscriber can always cancel the Fetch portion if we get "lucky" and find the Subscribe has given us what we need (an Object ID 0 and "sufficient" buffer depth for our purposes - which might be zero if we _really_ want to live on the edge).

In the future maybe we could consider a further optimization with a different Fetch Type that allows the _publisher_ to cancel the Fetch as soon as any Object ID 0 is sent, but there are only some circumstances in which that would be useful, and I don't think it's likely the common case for most live streaming applications.

draft-ietf-moq-transport.md
```diff
@@ -1424,34 +1443,48 @@ FETCH Message {
...
+Fields common to all Fetch Types:
+
 * Subscribe ID: The Subscribe ID identifies a given fetch request. Subscribe ID
 is a variable length integer that MUST be unique and monotonically increasing
-within  a session.
+within a session. For a Standalone Fetch a new Subscribe ID MUST be used. For
+a Joining Fetch, the Subscribe ID MUST correspond to a Subscribe which has already
+been sent. If a publisher receives a Joining Fetch with a Subscribe ID that does
+not correspond to an existing Subscribe, it MUST close the session with an
+Invalid Subscribe ID error.
```
wilaw, line 1471: Closing the session seems a bit drastic. Wouldn't you just respond with a FETCH_ERROR?

draft-ietf-moq-transport.md
```diff
@@ -1485,6 +1519,23 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+
+* Track Namespace: Same as in the corresponding Subscribe
+
+* Track Name: Same as in the corresponding Subscribe
+
+* StartGroup: LatestGroup as determined for the Subscribe minus Previous Group Count from the Joining Fetch
+
+* StartObject: Always 0
+
+* EndGroup: StartGroup of the corresponding Subscribe minus 1 (double check minus 1)
```
wilaw, line 1533: Editorial - remove your note to "double check minus 1"

draft-ietf-moq-transport.md
```diff
@@ -1466,8 +1499,9 @@ used. Values larger than 0x2 are a protocol error.
 * EndObject: The end Object ID, plus 1. A value of 0 means the entire group is
 requested.
 
-* Parameters: The parameters are defined in {{version-specific-params}}.
+Field present only for Joining Fetch (0x2):
 
+* Previous Group Count: The number of groups to Fetch prior to the StartGroup of the corresponding Subscribe
```
wilaw, line 1504: Are there limits on this value? Is zero allowed? Is 1000 allowed? Is -1 allowed?

draft-ietf-moq-transport.md
wilaw: For simplicity, we should disallow PGC = 0. It essentially overlaps the FETCH with the SUBSCRIBE. Forcing PGC >= 1 seems cleaner.

wilaw (commented):
Looking good Mike. I added a few points for clarification.

draft-ietf-moq-transport.md
afrind: The Largest group/object are *optional* in Subscribe OK:

> ContentExists: 1 if an object has been published on this track, 0 if not. If 0, then the Largest Group ID and Largest Object ID fields will not be present. Any other value is a protocol error and MUST terminate the session with a Protocol Violation ([Section 3.5](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#session-termination)).

draft-ietf-moq-transport.md
afrind: Individual Comment: 

I opened #639 to relax the start constraints in Subscribe.

I think with Joining Fetch, the Current Group option needs to be removed from Subscribe.  The only options should be Current Object, or a beginning or range in the future.  So I wouldn't worry to much about how Joining Fetch interacts with it.

> ... I think the simplest option would be to just accept that

I think so too.  In some ways this is the risk posed to a subscriber using Joining Fetch.  Depending on the group length/buffer size you may want to wait for Subscribe OK and issue the Fetch yourself.  As you say, the subscriber can cancel the Fetch too, so let's keep this PR as simple as possible.

draft-ietf-moq-transport.md
wilaw: > I thought I heard a use-case like filling backwards and cancelling as soon as the sum of the buffered data == playback buffer size.

Players would know the starting buffer they want before they JOIN FETCH, so the notion of asking for more than that seems like a non-use case and doesn't lead to any faster startup.

draft-ietf-moq-transport.md
suhasHere, line 1504: Since joining fetch is using an existing subscription, we should just specify the group_id to start from . My proposal is to use absolute group Id rather than counts.

draft-ietf-moq-transport.md
```diff
@@ -1485,6 +1519,23 @@ A publisher MUST send fetched groups in group order, either ascending or
...
 
+A publisher which receives a Fetch message with a Fetch Type of 0x2 should treat it as a Fetch
+with the following fields dynamically determined from the corresponding Subscribe:
+
+* Track Namespace: Same as in the corresponding Subscribe
+
+* Track Name: Same as in the corresponding Subscribe
+
+* StartGroup: LatestGroup as determined for the Subscribe minus Previous Group Count from the Joining Fetch
+
```
suhasHere, line 1530: same comment on using absolute group Id to fetch from than using count, as it assumes properties on the group distribution.

draft-ietf-moq-transport.md
suhasHere, line 1533: The end group must just be the latest group at the publisher when retrieving the joining fetch.

draft-ietf-moq-transport.md
wilaw: Start Group **IS** already an absolute group ID. As the text indicates, it is the Latest Group ID from the subscribe (which is absolute) minus the relative offset provided by the Joining Fetch.

draft-ietf-moq-transport.md
wilaw: The trouble with that proposal to use use absolute group Id rather than an offset, is that the information for the current live edge, obtained from SUBSCRIBE_OK, has to propagate back to the client before it can calculate an absolute join point for the fetch. This incurs an extra RTT in join time, which is what we are trying to avoid with this FETCH improvement. 

By using a relative offset, the client can issue the SUBSCRIBE and FETCH together, avoiding the 1xRTT delay.

draft-ietf-moq-transport.md
wilaw: Maybe call it "Group ID offset"? You can specify an offset of 2000, so it the subscribe returns a latest group of 16000 then you would start the fetch at 16000 - 2000 = 14000

draft-ietf-moq-transport.md
wilaw: The current definition of End Group for FETCH is inclusive, as is the definition of Largest Group ID for SUBSCRIBE. So they cannot be the same number or else that group will sent twice - once as part of the subscribe and again for the fetch. 

The problem with minus 1 is that Group IDs may not be contiguous. IF group numbers are every 1000 and you FETCH 5000 - 15999, the relay will need to go back to original publisher to determine if an object exists between 15001 and 15999.

draft-ietf-moq-transport.md
```diff
@@ -1485,6 +1519,52 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+If Resolved Subscribe Start Object is 0:
+* Fetch End Group: Resolved Subscribe Start Group - 1
+* Fetch End Object: 0 (all objects in the group)
+
+Else, if Resolved Subscribe Start Object is 1 or more:
+* Fetch End Group: Resolved Subscribe Start Group
+* Fetch End Object: Resolved Subscribe Start Object
+
+If Fetch End Group < Fetch Start Group no Fetch will be generated.
+(TODO: How can we signal this?)
```
afrind, line 1564: If you make a FETCH with a range that results in no objects I think you get a FETCHE_ERROR?

afrind (commented):
Individual Review

fluffy (approved):
I think we need to fix up the "what happens" when no content case but I like the overall PR here and seems to meet the use cases.

draft-ietf-moq-transport.md
englishm: As discussed on the call today, non-contiguous group IDs do still present some issues like this, but the same issues also exist for standalone fetch. Since a joining fetch is defined as resolving to the equivalent to a standalone fetch, I suggest that we address concerns about non-contiguous group IDs separately from this PR. 

It looks like we already have a couple open issues about how Group IDs are defined so I don't think we need another issue to track this case specifically. (See https://github.com/moq-wg/moq-transport/issues/423, https://github.com/moq-wg/moq-transport/issues/358, for example)

draft-ietf-moq-transport.md
englishm: Consolidating [discussion](https://github.com/moq-wg/moq-transport/pull/638#discussion_r1870922477) of this point.
> For simplicity, we should disallow PGC = 0. It essentially overlaps the FETCH with the SUBSCRIBE. Forcing PGC >= 1 seems cleaner.

I'm in favor of allowing "Preceding Group Offset" = 0 so that we can have the option of using Fetch to retrieve the first portion of a group when Subscribe(LatestObject) is used to retrieve the rest. The updated text hopefully makes it clear that the Fetch and Subscribe do not overlap.

>  Is 1000 allowed?

Yes. Depending on how group ID are set for a given track this may make sense. 

That said, it is not necessarily advisable to use a Joining Fetch to Fetch a large number of groups. In that case it may be better to wait for Subscribe OK and then make a series of smaller Fetch requests to fill the desired range.

I'm not opposed to putting some kind of limit on this value, but it's hard to say what it ought to be for all circumstances. I think it may be better to just allow publishers to reject joining fetch requests with PGOs larger than they wish to fulfill by responding with a Fetch Error. (I just created a separate issue about defining more error codes: https://github.com/moq-wg/moq-transport/issues/644)

> Is -1 allowed?

No. This field is defined as `[Preceding Group Offset (i),]` and this [Variable-Length Integer Encoding](https://www.rfc-editor.org/rfc/rfc9000#name-variable-length-integer-enc) does not permit negative values.

englishm:
@ianswett let me know if there are any other editorial changes you'd like to see here.

draft-ietf-moq-transport.md
```diff
@@ -1482,6 +1510,52 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+  * For Subscribes with Filter Type LatestObject or LatestGroup, this is equal to Largest Group ID.
+  * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange, this is equal to the StartGroup field of the Subscribe message
+* Resolved Subscribe Start Object:
+  * For Subscribes with Filter Type LatestObject, this is equal to Largest Object ID.
+  * For Subscribes with Filter Type LatestGroup, this is 0
+  * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange, this is equal to the StartObject field of the Subscribe message
+* Preceding Group Offset: A field in the Joining Fetch message indicating the relative offset from the start of the Subscribe
+
+Note: If a relay does not yet have LatestGroup and LatestObject for a given track, it may choose to either forward both the Subscribe and
+the Joining Fetch upstream or to watch until the Joining Fetch can be resolved locally. However it is handled, the Resolved Subscribe Start values
```
englishm, line 1531:
```suggestion
the Joining Fetch upstream or to wait until the Joining Fetch can be resolved locally. In either case, the Resolved Subscribe Start values
```

draft-ietf-moq-transport.md
```diff
@@ -1482,6 +1510,52 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+* Resolved Subscribe Start Group:
+  * For Subscribes with Filter Type LatestObject or LatestGroup, this is equal to Largest Group ID.
+  * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange, this is equal to the StartGroup field of the Subscribe message
+* Resolved Subscribe Start Object:
+  * For Subscribes with Filter Type LatestObject, this is equal to Largest Object ID.
+  * For Subscribes with Filter Type LatestGroup, this is 0
+  * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange, this is equal to the StartObject field of the Subscribe message
+* Preceding Group Offset: A field in the Joining Fetch message indicating the relative offset from the start of the Subscribe
+
+Note: If a relay does not yet have LatestGroup and LatestObject for a given track, it may choose to either forward both the Subscribe and
```
afrind, line 1530: There's a non-normative 'may' - suggest 'can'.

I like leaving some flexibility here. 

The case I brought up in the interim (SUBSCRIBE_OK seen but with contentExists=false) implicitly goes to "wait until you can resolve fetch locally" because you cannot send another SUBSCRIBE upstream.  Maybe we could say a bit more here.

draft-ietf-moq-transport.md
```diff
@@ -1482,6 +1510,52 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+### Calculating the Range of a Joining Fetch
+
+A publisher which receives a Fetch message with a Fetch Type of 0x2 must treat it as a Fetch
+with a range dynamically determined by the Preceding Group Offset (PGO) field and values derived
+from to the corresponding Subscribe message (hereafter "the Subscribe").
+
+The following values are used:
+
+* Resolved Subscribe Start Group:
+  * For Subscribes with Filter Type LatestObject or LatestGroup, this is equal to Largest Group ID.
```
afrind, line 1522: I think we need to resolve what's going on with Subscribe(LatestGroup).  Options:

1) Remove it (it splits past and future)
2) Subscribe(LatestGroup) is identical to Subscribe(LatestObject) + Joining Fetch(PGO=0) -- it sends a Subscribe Ok, a FetchOK, a Fetch stream and any future subgroups.
3) Special rules for publishers/relays that definitively have the entire current group cached

I guess it can be in another PR.

afrind (commented):
Individual Review

draft-ietf-moq-transport.md
englishm: I agree. I've tried to write this PR in a way where 1 is a viable option, but I think that actually modifying Subscribe should be done in a separate PR.

draft-ietf-moq-transport.md
englishm, line 1530:
```suggestion
Note: If a relay does not yet have LatestGroup and LatestObject for a given track, it can choose to either forward both the Subscribe and
```

englishm: Good catch, switched that to 'can'. 

Do we want to suggest behavior for particular cases like the one you describe?
The other one we discussed as seeming fairly straightforward is the case of receiving a Joining Fetch that joins to the first Subscribe for a track. Since the relay would send a new Subscribe upstream in that case, it could also forward the Joining Fetch so that the join point is resolved at the upstream publisher.

That said, the exact details of how relays choose to handle aggregation may vary from implementation to implementation and I agree with Will's sentiments that it would not be good to be too prescriptive here. If we spell out specific scenarios where a relay _can_ choose to take a particular course of action, does that make this note sound more normative than it should be? Or is just avoiding normative language (i.e. sticking to 'can' instead of MAY/MUST/SHOULD) sufficient?

I'll take a stab at spelling out these two cases non-normatively if folks think that makes sense.

wilaw: I'd replace 

> Note: If a relay does not yet have LatestGroup and LatestObject for a given track, it can choose to either forward both the Subscribe and the Joining Fetch upstream or to wait until the Joining Fetch can be resolved locally. In either case, the Resolved Subscribe Start values for a Joining Fetch MUST correspond to the Subscribe within the same session so the Fetch and Subscribe can be contiguous and non-overlapping.

with

> The Resolved Subscribe Start values for a Joining Fetch MUST correspond to the referenced Subscribe so that the Objects delivered across the Fetch and Subscribe are contiguous and non-overlapping.

englishm: I like that @wilaw - simple and to the point, thanks!

@afrind @fluffy do either of you feel it's necessary to spell out the cases outlined above or could we just go with something like Will's suggestion, focusing on only the normative requirement and omitting reference to what will likely be implementation-specific behavior to meet that requirement?

draft-ietf-moq-transport.md
afrind: I agree.  I'll wait for this to land since I don't think it will impact the decision and then open a separate issue.

draft-ietf-moq-transport.md
afrind: I liked your language "within the same session" because it made clear that the resolution had to be on the same endpoint.

I think the "already subscribed upstream, content-exists=false' case warrants a sentence explaining what to do.  Eg if the relay sends SUBSCRIBE_OK with content-exists=false, it MUST send a FETCH_ERROR for the joining fetch.

afrind (commented):
Individual Comments

draft-ietf-moq-transport.md
englishm: Incorporated these suggestions, thanks!

draft-ietf-moq-transport.md
```diff
@@ -1495,6 +1526,51 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+* Fetch End Group: Resolved Subscribe Start Group - 1
+* Fetch End Object: 0 (all objects in the group)
+
+Else, if Resolved Subscribe Start Object is 1 or more:
+* Fetch End Group: Resolved Subscribe Start Group
+* Fetch End Object: Resolved Subscribe Start Object
+
+If Fetch End Group < Fetch Start Group respond with a Fetch Error.
+
+A Joining Fetch MUST be sent in ascending group order.
```
wilaw, line 1572: > A Joining Fetch MUST be sent in ascending group order.

Why? See Alan's comment [here](https://github.com/moq-wg/moq-transport/issues/598#issuecomment-2605532478). The suggests that the client execute the FETCH in descending order and waits until it has sufficient groups to start. Those groups are the sum of those delivered by the FETCH and those delivered by the SUBSCRIBE. If you go in ascending order, you are always forced to wait for the complete range to be sent before you can stitch them together. But by going in reverse, you assemble the fastest path to your starting buffer. This should be an option for the player.

draft-ietf-moq-transport.md
```diff
@@ -1495,6 +1526,51 @@ A publisher MUST send fetched groups in group order, either ascending or
 descending. Within each group, objects are sent in Object ID order;
 subgroup ID is not used for ordering.
```
suhasHere, line 1527: Looks like this paragraph is duplicated

draft-ietf-moq-transport.md
```diff
@@ -1495,6 +1526,51 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+
+The following values are used:
+
+* Resolved Subscribe Start Group:
+  * For Subscribes with Filter Type Latest Object or Latest Group, this is equal to Largest Group ID.
+  * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange, this is equal to the StartGroup field of the Subscribe message
+* Resolved Subscribe Start Object:
+  * For Subscribes with Filter Type Latest Object, this is equal to Largest Object ID.
+  * For Subscribes with Filter Type Latest Group, this is 0
+  * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange, this is equal to the StartObject field of the Subscribe message
```
suhasHere, line 1543: This section seems to repeat the text to some extent in subscribe section. When PR https://github.com/moq-wg/moq-transport/pull/652 lands, some of these values may not match exactly especially when subscription filters are in the past to the live edge.

My proposal would be to refer to Subscribe section for these filter types and not repeat the text.

draft-ietf-moq-transport.md
```diff
@@ -1495,6 +1526,51 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+* Preceding Group Offset: A field in the Joining Fetch message indicating the relative offset from the start of the Subscribe
+
+The Resolved Subscribe Start values for a Joining Fetch MUST correspond to the referenced Subscribe within the same session so that the Objects delivered across the Fetch and Subscribe are contiguous and non-overlapping.
```
suhasHere, line 1546: not sure that we can guarantee "contigous" property but we can say it will be non-overlapping though

draft-ietf-moq-transport.md
```diff
@@ -1495,6 +1526,51 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+If a relay answers the referenced Subscribe with a `SUBSCRIBE_OK` that has ContentExists set to 0, it MUST respond to the Joining Fetch with a `FETCH_ERROR`.
+
+Using that information and the following algorithm, these values are computed:
```
suhasHere, line 1549: s/that/above information/

draft-ietf-moq-transport.md
```diff
@@ -1495,6 +1526,51 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+
+* Fetch Start Group: The StartGroup for the Fetch
+* Fetch Start Object: The StartObject for the Fetch (always 0)
+* Fetch End Group: The EndGroup for the Fetch
+* Fetch End Object: The EndObject for the Fetch (represented as 1 more than the end Object ID
+  or a value of 0 indicating that the entire group is requested)
```
suhasHere, line 1555: not sure if we need these definitions here ?

draft-ietf-moq-transport.md
```diff
@@ -1495,6 +1526,51 @@ A publisher MUST send fetched groups in group order, either ascending or
...
+
+The publisher receiving a Joining Fetch computes the fetch range as follows:
+
+* Fetch Start Group: Resolved Subscribe Start Group - Preceding Group Offset
+* Fetch Start Object: 0
+
+If Resolved Subscribe Start Object is 0:
```
suhasHere, line 1562: s/Resolved Subscribe Start Object is 0/ Subscribe FilterType is LatestGroup

draft-ietf-moq-transport.md
suhasHere: The whole point of joining fetch is solving is to have just enough buffer to have a smooth playout and not to overlap with standalone fetch use-cases. Given that, you want the preceding group offset ( of one or 2 probably) to be available fully before you start the playout and allowing just ascending order makes more sense for this usecase.

suhasHere (commented):
@englishm  thanks for the PR. This is looking pretty good. I have put in few thoughts on some observations made. Please let me know your thoughts.

draft-ietf-moq-transport.md
gwendalsimon: >  you want the preceding group offset ( of one or 2 probably) to be available fully before you start the playout
+
> Those groups are the sum of those delivered by the FETCH and those delivered by the SUBSCRIBE. If you go in ascending order, you are always forced to wait for the complete range to be sent before you can stitch them together. 

You can decide to start playing even if "the complete range" has not arrived yet (not "available fully" yet). You may still have a gap between the latest FETCH object you received and the earliest SUBSCRIBE object but you are confident that this gap will be filled when the playhead will reach that point. It is a Relay decision, and it can be more or less aggressive wrt to the offset (which can be 5-ish in a typical live streaming implementation).

Anyway, I don't see a strong motivation to force Ascending Order (the MUST in the sentence).

draft-ietf-moq-transport.md
englishm, line 1549:
```suggestion
Using the above information and the following algorithm, these values are computed:
```

draft-ietf-moq-transport.md
englishm: Looks like that was present before this PR, but I can fix it while I'm working in this section.

draft-ietf-moq-transport.md
englishm: For Joining Fetch it's important that the _ranges_ of the Fetch and Subscribe portions at least be both contiguous and non-overlapping. Actual delivery of objects will of course be subject to whatever Fetch and Subscribe each provide. Perhaps I could reword slightly to keep this scoped to the properties we care about for Joining Fetch and not overstate anything with regard to delivery?

draft-ietf-moq-transport.md
englishm: This is defining how these computed values are used to fill out the fields of what could otherwise be a standalone Fetch.

draft-ietf-moq-transport.md
englishm: That's not the only case here. This can also be true if the StartObject field of a Subscribe with Filter Type of AbsoluteStart or AbsoluteRange is 0. This is an example of why I'd prefer to retain that above section defining "Resolved Subscribe..." values rather than just referring directly to the Subscribe section's definitions.

draft-ietf-moq-transport.md
englishm: I first wrote this with the relevant details collapsed into the below section and it was very verbose and hard to follow. I split it out by defining these intermediate "Resolved" values for clarity. I see your point about duplication, and you're right that #652 may also need to touch this, but I still prefer to state these things explicitly here to avoid confusion about how Joining Fetch values should be determined. Maybe @ianswett can weigh in as both editor and author of #652.

draft-ietf-moq-transport.md
englishm: Addressed in [7d4a99d](https://github.com/moq-wg/moq-transport/pull/638/commits/7d4a99dbddc0c513bcbfb4355a474fd354adb9e6)

draft-ietf-moq-transport.md
suhasHere: My only concern is #652 will invalidate some of the resolved values as it is defined in this PR.  Given we had general agreement that Susbcribe needs updates if we add Joining Fetch, it would be nice to land these two PRs around the same time to keep things unambiguous.

draft-ietf-moq-transport.md
suhasHere: By contiguous, if we mean groups will be in sequence with no gaps, we cannot guarantee that for variety of reasons.  Am I misunderstanding it ?

draft-ietf-moq-transport.md
suhasHere: True .. I was wondering can't this be simplified by combining it with the next paragraph that follows

draft-ietf-moq-transport.md
suhasHere: I am fine with having the section on Resolved subscribes , but seems bit incorrect to compute those absolute values given those are filters and may not map to the exact values that we mention here

draft-ietf-moq-transport.md
englishm: > I don't see a strong motivation to force Ascending Order (the MUST in the sentence)

The main reason is really just that we wanted to omit the Group Order field in the message format for a Fetch with a Fetch Type of Joining Fetch.

I can either add Group Order back in (shifting it up into the shared fields just after `Subscriber Priority`), or we could make some other simplifying association here like saying that "the Group Order of the referenced Subscribe will be used." I don't think the latter would always be what we want though, so I'm inclined to just add the Group Order field back in here. I'll go ahead and do that shortly unless anyone strongly objects to reinstating that flexibility.

draft-ietf-moq-transport.md
englishm: Ah, I think maybe we're focusing on different things here and I think it may relate to some of your other comments, too. I'm thinking of this primarily in terms of what is being _requested_ - we can make our requests contiguous and non-overlapping, even if the way the objects are published and delivered means there could be gaps. What actually gets delivered will be subject to whatever limitations any other Subscribe and Fetch are subject to. I updated this line of the text to clarify that, too.

draft-ietf-moq-transport.md
englishm: As above, I'm thinking of this in terms of computing an absolute value for the _requested_ range so we can delineate which portion would be provided by a Fetch (subject to whatever limitations a standalone Fetch is subject to) and which portion would be provided by a Subscribe (subject to whatever limitations a Subscribe is subject to).

draft-ietf-moq-transport.md
englishm: On a re-read, I think you may be right. I think we can eliminate these definitions.

draft-ietf-moq-transport.md
englishm: OK, I added the Group Order field back in with [7db23b5](https://github.com/moq-wg/moq-transport/pull/638/commits/7db23b587cb0665ab4a60ccbe519238f11061b82) - it is no longer fixed to Ascending for Joining Fetch.

draft-ietf-moq-transport.md
englishm: Done in [86ec7ff](https://github.com/moq-wg/moq-transport/pull/638/commits/86ec7ffe924e64f75f2659862f75ae910ce9bf57).

ianswett (commented):
I find this approach a bit awkward, so I wrote #657, but if we got rid of "Recent Group" mode in SUBSCRIBE, I think it'd make more sense.

draft-ietf-moq-transport.md
englishm: @ianswett if we pursue 1 here, does Joining Fetch seem less awkward?

draft-ietf-moq-transport.md
```diff
@@ -1566,8 +1598,11 @@ used. Values larger than 0x2 are a protocol error.
 * EndObject: The end Object ID, plus 1. A value of 0 means the entire group is
 requested.
 
-* Parameters: The parameters are defined in {{version-specific-params}}.
+Field present only for Joining Fetch (0x2):
 
+* Preceding Group Offset: The group offset for the Fetch prior and relative
+to the StartGroup of the corresponding Subscribe. A value of 0 indicates
```
ianswett, line 1604: What if the StartGroup of the Subscribe is in the future?  The current text and algorithm below seem to indicate I could issue a joining fetch for Objects that haven't been published?

draft-ietf-moq-transport.md
```diff
@@ -1524,29 +1543,31 @@ FETCH Message {
...
 * Subscribe ID: The Subscribe ID identifies a given fetch request. Subscribe ID
 is a variable length integer that MUST be unique and monotonically increasing
-within  a session.
-
-* Track Namespace: Identifies the namespace of the track as defined in
-({{track-name}}).
-
-* Track Name: Identifies the track name as defined in ({{track-name}}).
+within a session. For a Standalone Fetch a new Subscribe ID MUST be used. For
+a Joining Fetch, the Subscribe ID MUST correspond to a Subscribe which has already
```
ianswett, line 1568: If I cancel the Subscribe, does it cancel the Fetch?  I assume SUBSCRIBE_UPDATE doesn't have any impact on the Fetch?

draft-ietf-moq-transport.md
```diff
@@ -1581,9 +1616,50 @@ subgroup ID is not used for ordering.
...
+  * For Subscribes with Filter Type Latest Object or Latest Group,
+    this is equal to Largest Group ID.
+  * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange,
+    this is equal to the StartGroup field of the Subscribe message
+* Resolved Subscribe Start Object:
+  * For Subscribes with Filter Type Latest Object, this is equal to Largest Object ID.
+  * For Subscribes with Filter Type Latest Group, this is 0
+  * For Subscribes with Filter Type AbsoluteStart or AbsoluteRange,
+    this is equal to the StartObject field of the Subscribe message
+* Preceding Group Offset: A field in the Joining Fetch message indicating the
```
ianswett, line 1637: This is defined above, so I think we can omit this definition here.

draft-ietf-moq-transport.md
```diff
@@ -1581,9 +1616,49 @@ subgroup ID is not used for ordering.
 If StartGroup/StartObject is greater than the latest published Object group,
 the publisher MUST return FETCH_ERROR with error code 'No Objects'.
 
-A publisher MUST send fetched groups in group order, either ascending or
-descending. Within each group, objects are sent in Object ID order;
-subgroup ID is not used for ordering.
+### Calculating the Range of a Joining Fetch
+
+A publisher which receives a Fetch message with a Fetch Type of 0x2 treats it
+as a Fetch with a range dynamically determined by the Preceding Group Offset (PGO)
```
ianswett, line 1622:
```suggestion
as a Fetch with a range dynamically determined by the Preceding Group Offset
```
You don't use PGO elsewhere in the PR

draft-ietf-moq-transport.md
```diff
@@ -1581,9 +1616,49 @@ subgroup ID is not used for ordering.
...
+
+If Resolved Subscribe Start Object is 0:
+* Fetch EndGroup: Resolved Subscribe Start Group - 1
+* Fetch EndObject: 0 (all objects in the group)
+
+Else, if Resolved Subscribe Start Object is 1 or more:
+* Fetch EndGroup: Resolved Subscribe Start Group
+* Fetch EndObject: Resolved Subscribe Start Object
+
+If Fetch EndGroup < Fetch StartGroup, there are no Objects to retrieve,
```
ianswett, line 1660: It's always true that if there are no Objects in the Fetch range, it results in a FETCH_ERROR, so I'm not sure we need to say it again here.  If you are going to restate it, I think Fetch EndGroup == Fetch StartGroup && Fetch EndObject < Fetch StartObject also results in a FETCH_ERROR?

ianswett (commented):
I think this would read better without the Latest Group mode of SUBSCRIBE.  We can get rid of it later, but it'd simplify this text a bit.

I'm curious about the idea of FETCH in the future and in general any joining fetches that aren't joining SUBSCRIBEs to the current Object.  The joining fetch is very much a latency optimization and/or convenience, so I'm not sure there's any reason to allow a joining fetch to join anything besides a Current Object SUBSCRIBE?

draft-ietf-moq-transport.md
englishm: Hah, good point! Thanks!

draft-ietf-moq-transport.md
```diff
@@ -1524,29 +1541,31 @@ FETCH Message {
   Type (i) = 0x16,
   Length (i),
   Subscribe ID (i),
-  Track Namespace (tuple),
-  Track Name Length (i),
-  Track Name (..),
   Subscriber Priority (8),
   Group Order (8),
```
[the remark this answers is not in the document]
englishm: Yes, but it was requested that we continue to allow Group Order instead. I don't have strong feelings either way, but see the discussion here:

https://github.com/moq-wg/moq-transport/pull/638#discussion_r1927117659

draft-ietf-moq-transport.md
englishm: That is a good point I had not previously considered. We probably should preclude joining a Fetch to a Subscribe in the future.

draft-ietf-moq-transport.md
englishm: Canceling the Subscribe would not cancel the Fetch, but if you cancel the Subscribe before sending the Joining Fetch the Joining Fetch would fail with a FETCH_ERROR. SUBSCRIBE_UPDATE would only update the Subscribe with the given id, not the Fetch.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
ianswett: SG, let's keep it as is for now.

draft-ietf-moq-transport.md
englishm: Made a pass at limiting Joining Fetch to only Subscribes with the Latest Object Filter Type. It definitely reduces the complexity here quite a bit. See a70455e473f83a2d6e9381bd855ce06dbff5c1e4.

draft-ietf-moq-transport.md
```diff
@@ -1590,9 +1627,40 @@ subgroup ID is not used for ordering.
...
+as a Fetch with a range dynamically determined by the Preceding Group Offset
+field and field values derived from the corresponding SUBSCRIBE message
+(hereafter "the Subscribe").
+
+The following values are used:
+
+* Resolved Subscribe Start Group: the Largest Group ID of the associated Subscribe.
+* Resolved Subscribe Start Object: the Largest Object ID of the associated Subscribe.
+
+The Resolved Subscribe Start values for a Joining Fetch MUST correspond to the
```
ianswett, line 1642: Nit: I don't think you need a MUST here.  This is just explaining how the mechanism works and why.

```suggestion
The Resolved Subscribe Start values for a Joining Fetch correspond to the
```

ianswett (approved):
Thanks for the update to make it only apply to Latest Object subscriptions, this is MUCH simpler IMO.

draft-ietf-moq-transport.md
englishm: This is actually a requirement, but I see how it's easy to read it as just descriptive. The important part is the "within the same session" which precludes responding to Subscribe at relay, but forwarding the Joining Fetch and having it resolved against a different Subscribe at an upstream relay. We don't necessarily want to entirely preclude forwarding upstream, but it's important that the Joining Fetch be resolved **within the same session** as the associated Subscribe so that this join point is consistent.

This text originated after the [2024-12-18 weekly interim](https://datatracker.ietf.org/doc/minutes-interim-2024-moq-30-202412181700/) where the main issue raised (mainly by Cullen and Alan, if I recall correctly) was a desire to clarify requirements relating to forwarding behavior. My initial stab at text for that was in this commit: https://github.com/moq-wg/moq-transport/commit/9dfceaf9c6e3bc71761a1f61ecfed2fed3357c7f followed by some discussion tightening it up here: https://github.com/moq-wg/moq-transport/pull/638#discussion_r1893250280

If this is too subtle, we could expand it a bit again or rephrase it to make the requirement more clear, but the MUST here is intentional.

draft-ietf-moq-transport.md
englishm: Makes sense to me - added that clarifying language in [411d42e](https://github.com/moq-wg/moq-transport/pull/638/commits/411d42e864776007ce9d91ef867072bda705947b).

draft-ietf-moq-transport.md
englishm: Removed in 
[c3442de](https://github.com/moq-wg/moq-transport/pull/638/commits/c3442de318bef65db36f27e6b70d4820d28b20ce). 

Would it be worth explicitly stating (again) that once these values are resolved, the Fetch is handled as an ordinary Standalone Fetch? I'm leaning towards 'no' since we already say something to that effect [above](https://github.com/moq-wg/moq-transport/pull/638/files#diff-7bc267642f4c6d5c039e5cd0ab26784907b240eb24ca1e9fedb4d5feccca49ffR1634-R1637).

> A publisher which receives a Fetch message with a Fetch Type of 0x2 treats it
> as a Fetch with a range dynamically determined by the Preceding Group Offset
> field and field values derived from the corresponding SUBSCRIBE message
> (hereafter "the Subscribe").

draft-ietf-moq-transport.md
ianswett: I think everything in this document is in the context of a single session, except for some advice on relays, so I thought that aspect was a bit out of place.  For example, SUBSCRIBE_UPDATE and SUBSCRIBE_CANCEL don't clarify that you're altering a subscription within the session because it's implied.

draft-ietf-moq-transport.md
```diff
@@ -1575,8 +1611,11 @@ used. Values larger than 0x2 are a protocol error.
 * EndObject: The end Object ID, plus 1. A value of 0 means the entire group is
 requested.
 
-* Parameters: The parameters are defined in {{version-specific-params}}.
+Field present only for Joining Fetch (0x2):
 
+* Preceding Group Offset: The group offset for the Fetch prior and relative
+to the StartGroup of the corresponding Subscribe. A value of 0 indicates
```
suhasHere, line 1617: s/StartGroup/Current Group , since joining fetch is always relative to the current group.

draft-ietf-moq-transport.md
```diff
@@ -1575,8 +1611,11 @@ used. Values larger than 0x2 are a protocol error.
...
+the current Group.
```
suhasHere, line 1618: s/the current Group / beginning of the current Group

draft-ietf-moq-transport.md
```diff
@@ -1590,9 +1629,37 @@ subgroup ID is not used for ordering.
...
+The Resolved Subscribe Start values for a Joining Fetch MUST correspond to the
+referenced Subscribe within the same session so that the ranges of Objects covered
+by the Fetch and Subscribe are contiguous and non-overlapping.
+If no Objects have been published for the track, so the SUBSCRIBE_OK has a
+ContentExists value of 0, the publisher MUST respond to the Joining Fetch with a
+FETCH_ERROR.
+
+The publisher receiving a Joining Fetch computes the fetch range as follows:
+
+* Fetch StartGroup: Resolved Subscribe Start Group - Preceding Group Offset
```
suhasHere, line 1653: Since ResolveSubscribe Start Group is Largest Group ID of the subscription, it may be clearer if we don't introduce new terminology and just define as follows

Fetch StartGroup: Largest Subscribe  Group ID - Preceding Group Offset

suhasHere: same comment applies to the Resolved Subscribe Object too ?

fluffy:
The more I think about the joining case where the Fetch and associated Subscribe share the same subscribe ID seems like it might cause problems for us later. Say we added an extension that did something that worked on a Fetch or Subscribe, there would be know way to know what it referred to. 

I think it would be better to have joining fetch have a new ID and in the request parameters cary the ID of the associated subscribe.

draft-ietf-moq-transport.md
```diff
@@ -1533,29 +1554,33 @@ FETCH Message {
...
-  EndObject (i),
+  Fetch Type (i),
+  [Track Namespace (tuple),
+   Track Name Length (i),
+   Track Name (..),
+   StartGroup (i),
+   StartObject (i),
+   EndGroup (i),
+   EndObject (i),]
+  [Preceding Group Offset (i),]
```
fluffy, line 1567: in the joining case, I think we should add the subscribe ID of the associated subscribe

draft-ietf-moq-transport.md
```diff
@@ -1514,6 +1514,27 @@ objects within a track. The publisher responding to a FETCH is responsible for r
...
 fetch response. All omitted objects have status Object Does Not Exist.
 
+**Fetch Types**
+
+There are two types of Fetch messages:
+
+Standalone Fetch (0x1) : A Fetch of Objects performed independently of any Subscribe.
+
+Joining Fetch (0x2) : A Fetch joined together with a Subscribe. A Joining Fetch
+shares the same Subscribe ID as an already-sent Subscribe. A publisher receiving a
```
fluffy, line 1524: I think we should use a new Subscribe ID for this and put the id of the already-sent Subscribe as one of the paramaters

ianswett: That does seem reasonable, though admittedly I don't have a strong opinion.

afrind: Individual Comment:

Having separate subscribe IDs for the join and fetch might reduce the chance of getting them mixed up at some point.  If it's not required that they're the same, maybe this is a good approach.

englishm: I don't feel especially strongly one way or the other here either. I think the reason we originally shared the ID here was mostly about opportunistically saving a field. There may be some other benefits, too, but I'm happy to separate them if that's the way most folks are leaning.

ianswett: I don't think saving a varint is a good reason to have them be the same, given how infrequently this will be sent.

draft-ietf-moq-transport.md
```diff
@@ -1514,6 +1514,27 @@ objects within a track. The publisher responding to a FETCH is responsible for r
...
+Joining Fetch should use properties of the associated Subscribe to determine the
```
ianswett, line 1525:
```suggestion
Joining Fetch uses properties of the associated Subscribe to determine the
```

ianswett (commented):
I do think it's less confusing to have a separate ID for the Fetch, since it makes some of the questions I previously had (ie: If I do SUBSCRIBE_CANCEL, does it also cancel the Fetch with the same ID?) obvious.

draft-ietf-moq-transport.md
```diff
@@ -1575,8 +1607,15 @@ used. Values larger than 0x2 are a protocol error.
...
 
-* Parameters: The parameters are defined in {{version-specific-params}}.
+Field present only for Joining Fetch (0x2):
+
+* Joining Subscribe ID: The Subscribe ID of the existing subscription to be
+joined. If a publisher receives a Joining Fetch with a Subscribe ID that does
+not correspond to an existing Subscribe, it MUST respond with a Fetch Error.
 
+* Preceding Group Offset: The group offset for the Fetch prior and relative
+to the StartGroup of the corresponding Subscribe. A value of 0 indicates
```
ianswett, line 1617:
```suggestion
to the Current Group of the corresponding Subscribe. A value of 0 indicates
```

draft-ietf-moq-transport.md
```diff
@@ -1575,8 +1607,15 @@ used. Values larger than 0x2 are a protocol error.
...
+the current Group.
```
ianswett, line 1618:
```suggestion
the Fetch starts at the beginning of the Current Group.
```

ianswett (approved):

afrind (approved):

draft-ietf-moq-transport.md
```diff
@@ -1575,8 +1607,15 @@ used. Values larger than 0x2 are a protocol error.
 * EndObject: The end Object ID, plus 1. A value of 0 means the entire group is
 requested.
 
-* Parameters: The parameters are defined in {{version-specific-params}}.
+Field present only for Joining Fetch (0x2):
```
ianswett, line 1610:
```suggestion
Fields present only for Joining Fetch (0x2):
```

draft-ietf-moq-transport.md
```diff
@@ -1636,30 +1636,18 @@ as a Fetch with a range dynamically determined by the Preceding Group Offset
...
+ContentExists value of 0, the publisher responds with a FETCH_ERROR.
 
 The publisher receiving a Joining Fetch computes the fetch range as follows:
 
-* Fetch StartGroup: Resolved Subscribe Start Group - Preceding Group Offset
+* Fetch StartGroup: Subscribe Largest Group - Preceding Group Offset
 * Fetch StartObject: 0
-
-If Resolved Subscribe Start Object is 0:
-* Fetch EndGroup: Resolved Subscribe Start Group - 1
```
englishm: I think we do still need to handle this zero case separately.

e.g. Subscribe Largest Group = 3, Subscribe Largest Object = 0, Preceding Group Offset = 2

The new text here would have us create a Fetch with:
- StartGroup = 1
- StartObject = 0
- EndGroup = 3
- EndObject = 0

This would overlap an entire group since an EndObject of 0 in a Fetch means "all objects in the group"

I think the rest of the text consolidation here is fine, but we do need to keep this `- 1` for the Fetch EndGroup when Subscribe Largest Object is 0.

PR #643, opened 2024-12-13
Message flows and endpoint state management (Was: ANNOUNCE)

### LINKED ISSUE #556: ANNOUNCE (and maybe SUBSCRIBE?) infinite loops
I believe that if we're not careful, we can end up with ANNOUNCE infinite loops today.

If there are three relays connected: A -> B -> C -> A, then we need to ask relays to see if they've already seen an ANNOUNCE before forwarding it along to applicable SUBSCRIBE_NAMESPACEs.  Similarly, I can imagine an ANNOUNCE/UNANNOUNCE loop, which aren't as trivially prevented by ensuring you don't forward the same ANNOUNCE twice.

SUBSCRIBE has a temporal ordering, so this feels more tractable, but given we allow arbitrary gaps in Objects and Groups, tracking everything you've already forwarded for a subscription is potentially very costly/complex.  If a relay is subscribing to the same track twice because two different relays announced it, I think it needs to keep state to ensure it doesn't forward the same Object multiple times?

BGP has loop prevention mechanisms and so do QUIC stateless resets.

Possibly deployments can ensure data flows in only one direction, but given this is pub/sub, not request/response, it's a lot easier to create loops.

### LINKED ISSUE #557: How to use ANNOUNCE_OK, ANNOUNCE_ERROR, and ANNOUNCE_CANCEL?
QUIC provides reliability, so I know the peer received the ANNOUNCE (or UNANNOUNCE).

We have almost no text on what might constitute an ANNOUNCE_ERROR and the only text on ANNOUNCE_CANCEL is you're supposed to send it if you're not going to send subscriptions to the publisher for something you sent an ANNOUNCE_OK for previously.

As it's written today, I don't see why an implementation would send or take any action based on any of these messages?

I believe there is some functionality we'd like in MoQT, which is we don't want anyone to be able to announce a track without appropriate auth, but I don't think this trio of messages gives us that, certainly not in the current draft.  I'd hope pretending to be authoritative for an entity you're not would be rare, and closing the session or another more abrupt approach would be fine.

### LINKED ISSUE #577: Errors from Announce
Need better explanation about what error codes ara generated from Annouce

### LINKED ISSUE #583: Does SUBSCRIBE_NAMESPACE return your own ANNOUNCE's
In moq-chat, a subscriber does an ANNOUNCE for their namespace and a SUBSCRIBE_NS for a level higher prefix.  Should the relay filter these announces, or forward them back to the originator?

### LINKED ISSUE #584: Is it necessary for the Relay Server to send the ANNOUNCE to the End Subscriber without waiting for a SUBSCRIBE_NAMESPACE?
SUBSCRIBE_NAMESPACE allows ANNOUNCE filtered by TrackNamespace prefix to be sent from the relay server to the End Subscriber. This is to prevent all ANNOUNCEs from being sent to the End Subscriber in case of a large number of TrackNamespaces.

However, the current specification appears to allow ANNOUNCEs to be sent from the Relay Server to the End Subscriber without SUBSCRIBE_NAMESPACE.

I think it would be better to allow the relay server to send an ANNOUNCE to the End Subscriber only if the End Subscriber sends a SUBSCRIBE_NAMESPACE. 
This method makes the sequence very clear.

martinduke:
Fixes #556
Fixes #557 
Fixes #583 
Fixes #584 

Also eliminates text from the relay section that is redundant, because it applies to endpoints whether or not they are relays.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -895,8 +895,8 @@ control message with a `Track Namespace` ({{model-track}}).
 The ANNOUNCE enables the relay to know which publisher to forward a
 SUBSCRIBE to.
 
-Relays MUST ensure that publishers are authorized by verifying that the
-publisher is authorized to publish the content associated with the set of
+Relays MUST verify that publishers are authorized to publish
+the content associated with the set of
 tracks whose Track Namespace matches the announced namespace. Where the
 authorization and identification of the publisher occurs depends on the way the
 relay is managed and is application specific.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -600,7 +600,11 @@ code, as defined below:
   stream. If an endpoint times out waiting for a new object header on an
   open subgroup stream, it MAY send a STOP_SENDING on that stream, terminate
   the subscription, or close the session with an error.
+<<<<<<< HEAD
 
+=======
+  
+>>>>>>> 297094f (Add session errors for control and data timeout)
 ## Migration {#session-migration}
 
 MOQT requires a long-lived and stateful session. However, a service
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -874,6 +874,48 @@ allows relays to make only a single upstream subscription for the
 track. The published content received from the upstream subscription
 request is cached and shared among the pending subscribers.
 
+The application SHOULD use a relevant error code in SUBSCRIBE_ERROR,
+as defined below:
+
+|------|---------------------------|
+| Code | Reason                    |
+|-----:|:--------------------------|
+| 0x0  | Internal Error            |
+|------|---------------------------|
+| 0x1  | Invalid Range             |
+|------|---------------------------|
+| 0x2  | Retry Track Alias         |
+|------|---------------------------|
+| 0x3  | Track Does Not Exist      |
+|------|---------------------------|
+| 0x4  | Unauthorized              |
+|------|---------------------------|
+| 0x5  | Timeout                   |
+|------|---------------------------|
+
+The application SHOULD use a relevant status code in
+SUBSCRIBE_DONE, as defined below:
+
+|------|---------------------------|
+| Code | Reason                    |
+|-----:|:--------------------------|
+| 0x0  | Unsubscribed              |
+|------|---------------------------|
+| 0x1  | Internal Error            |
+|------|---------------------------|
+| 0x2  | Unauthorized              |
+|------|---------------------------|
+| 0x3  | Track Ended               |
+|------|---------------------------|
+| 0x4  | Subscription Ended        |
+|------|---------------------------|
+| 0x5  | Going Away                |
+|------|---------------------------|
+| 0x6  | Expired                   |
+|------|---------------------------|
+| 0x7  | Too Far Behind            |
+|------|---------------------------|
+
 ### Graceful Publisher Relay Switchover
 
 This section describes behavior a subscriber MAY implement
@@ -1600,22 +1642,6 @@ message for which this response is provided.
 
 * Reason Phrase: Provides the reason for announcement error.
 
-The following error codes are defined:
-
-|------|---------------------------|
-| Code | Reason                    |
-|-----:|:--------------------------|
-| 0x0  | Internal Error            |
-|------|---------------------------|
-| 0x1  | Uninterested              |
-|------|---------------------------|
-| 0x2  | Unauthorized              |
-|------|---------------------------|
-| 0x3  | Timeout                   |
-|------|---------------------------|
-| 0x4  | Announce Not Supported    |
-|------|---------------------------|
-
 ## ANNOUNCE_CANCEL {#message-announce-cancel}
 
 The subscriber sends an `ANNOUNCE_CANCEL` control message to
@@ -1803,24 +1829,6 @@ SUBSCRIBE_ERROR
   the subscriber MUST close the connection with a Duplicate Track Alias error
   ({{session-termination}}).
 
-The following error codes are defined:
-
-|------|---------------------------|
-| Code | Reason                    |
-|-----:|:--------------------------|
-| 0x0  | Internal Error            |
-|------|---------------------------|
-| 0x1  | Invalid Range             |
-|------|---------------------------|
-| 0x2  | Retry Track Alias         |
-|------|---------------------------|
-| 0x3  | Track Does Not Exist      |
-|------|---------------------------|
-| 0x4  | Unauthorized              |
-|------|---------------------------|
-| 0x5  | Timeout                   |
-|------|---------------------------|
-
 ## FETCH_OK {#message-fetch-ok}
 
 A publisher sends a FETCH_OK control message in response to successful fetches.
@@ -1885,23 +1893,6 @@ FETCH_ERROR
 
 * Reason Phrase: Provides the reason for fetch error.
 
-The following error codes are defined:
-
-|------|---------------------------|
-| Code | Reason                    |
-|-----:|:--------------------------|
-| 0x0  | Internal Error            |
-|------|---------------------------|
-| 0x1  | Invalid Range             |
-|------|---------------------------|
-| 0x3  | Track Does Not Exist      |
-|------|---------------------------|
-| 0x4  | Unauthorized              |
-|------|---------------------------|
-| 0x5  | Timeout                   |
-|------|---------------------------|
-
-
 ## SUBSCRIBE_DONE {#message-subscribe-done}
 
 A publisher sends a `SUBSCRIBE_DONE` message to indicate it is done publishing
@@ -1962,29 +1953,6 @@ opened for this subscription.  The subscriber can remove all subscription state
 once the same number of streams have been processed.
 
 * Reason Phrase: Provides the reason for subscription error.
-
-The application SHOULD use a relevant status code in SUBSCRIBE_DONE, as defined
- below:
-
-|------|---------------------------|
-| Code | Reason                    |
-|-----:|:--------------------------|
-| 0x0  | Unsubscribed              |
-|------|---------------------------|
-| 0x1  | Internal Error            |
-|------|---------------------------|
-| 0x2  | Unauthorized              |
-|------|---------------------------|
-| 0x3  | Track Ended               |
-|------|---------------------------|
-| 0x4  | Subscription Ended        |
-|------|---------------------------|
-| 0x5  | Going Away                |
-|------|---------------------------|
-| 0x6  | Expired                   |
-|------|---------------------------|
-| 0x7  | Too Far Behind            |
-|------|---------------------------|
 
 ## MAX_SUBSCRIBE_ID {#message-max-subscribe-id}
 
@@ -2176,25 +2144,6 @@ title="MOQT SUBSCRIBE_ANNOUNCES_ERROR Message"}
 failure.
 
 * Reason Phrase: Provides the reason for the namespace subscription error.
-
-The application SHOULD use a relevant error code in a
-SUBSCRIBE_NAMESPACES_ERROR, as defined below.
-
-|------|--------------------------------|
-| Code | Reason                         |
-|-----:|:-------------------------------|
-| 0x0  | Internal Error                 |
-|------|--------------------------------|
-| 0x1  | Message Not Supported          |
-|------|--------------------------------|
-| 0x2  | Will Never Announce Namespace  |
-|------|--------------------------------|
-| 0x3  | Unauthorized                   |
-|------|------------------------------- |
-| 0x4  | Timeout                        |
-|------|--------------------------------|
-
-
 
 # Data Streams {#data-streams}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -631,14 +631,6 @@ expects more OBJECTs to be delivered. The server closes the session with a
 Discovery of MoQT servers is always done out-of-band. Namespace discovery can be
 done in the context of an established MoQT session.
 
-Throughout this section, "publishers" refer to endpoints that can deliver
-objects in response to SUBSCRIBE or FETCH messages for a particular track, and
-"subscribers" refer to endpoints that might send SUBSCRIBE or FETCH messages for
-a track. An endpoint can be both a publisher and subscriber in a session, for
-different tracks. If a relay, it can be a publisher for a track in one session
-and a subscriber for it in another session. But an endpoint cannot be publisher
-and subscriber for the same track in the same session.
-
 Given sufficient out of band information, it is valid for a subscriber
 to send a SUBSCRIBE or FETCH message to a publisher (including a relay) without
 any previous MoQT messages besides SETUP. However, SUBSCRIBE_ANNOUNCES and
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -837,7 +837,7 @@ Relays MAY cache Objects, but are not required to.
 
 ## Subscriber Interactions
 
-Subscribers subscribe to tracks by sending  a SUBSCRIBE
+Subscribers subscribe to tracks by sending a SUBSCRIBE
 ({{message-subscribe-req}}) control message for each track of
 interest. Relays MUST ensure subscribers are authorized to access the
 content associated with the track. The authorization
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -631,11 +631,19 @@ expects more OBJECTs to be delivered. The server closes the session with a
 Discovery of MoQT servers is always done out-of-band. Namespace discovery can be
 done in the context of an established MoQT session.
 
+Throughout this section, "publishers" refer to endpoints that can deliver
+objects in response to SUBSCRIBE or FETCH messages for a particular track, and
+"subscribers" refer to endpoints that might send SUBSCRIBE or FETCH messages for
+a track. An endpoint can be both a publisher and subscriber in a session, for
+different tracks. If a relay, it can be a publisher for a track in one session
+and a subscriber for it in another session. But an endpoint cannot be publisher
+and subscriber for the same track in the same session.
+
 Given sufficient out of band information, it is valid for a subscriber
 to send a SUBSCRIBE or FETCH message to a publisher (including a relay) without
 any previous MoQT messages besides SETUP. However, SUBSCRIBE_ANNOUNCES and
 ANNOUNCE messages provide an in-band means of discovery of subscribers and
-publishers, respectively.
+publishers for a namespace.
 
 The syntax of these messages is described in {{message}}.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -600,11 +600,7 @@ code, as defined below:
   stream. If an endpoint times out waiting for a new object header on an
   open subgroup stream, it MAY send a STOP_SENDING on that stream, terminate
   the subscription, or close the session with an error.
-<<<<<<< HEAD
 
-=======
-  
->>>>>>> 297094f (Add session errors for control and data timeout)
 ## Migration {#session-migration}
 
 MOQT requires a long-lived and stateful session. However, a service
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -679,8 +679,7 @@ application to inform the search for additional subscribers for a namespace,
 or abandoning the attempt to publish under this namespace. This might be
 especially useful in upload or chat applications. A subscriber MUST send exactly
 one ANNOUNCE_OK or ANNOUNCE_ERROR in response to an ANNOUNCE. The publisher
-SHOULD close the session with a protocol error if it detects receiving more than
-one.
+SHOULD close the session with a protocol error if it receives more than one.
 
 An UNANNOUNCE message withdraws a previous ANNOUNCE, although it is not a
 protocol error for the subscriber to send a SUBSCRIBE or FETCH message after
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -838,8 +838,8 @@ Relays MAY cache Objects, but are not required to.
 
 ## Subscriber Interactions
 
-Subscribers interact with the Relays by sending a SUBSCRIBE
-({{message-subscribe-req}}) control message for the tracks of
+Subscribers subscribe to tracks by sending  a SUBSCRIBE
+({{message-subscribe-req}}) control message for each track of
 interest. Relays MUST ensure subscribers are authorized to access the
 content associated with the track. The authorization
 information can be part of subscription request itself or part of the
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -888,20 +888,21 @@ to the old relay can be stopped with an UNSUBSCRIBE.
 
 Publishing through the relay starts with publisher sending ANNOUNCE
 control message with a `Track Namespace` ({{model-track}}).
-The ANNOUNCE enables the relay to know which publisher to forward a
-SUBSCRIBE to, if the track name belongs to the namespace indicated in
-the ANNOUNCE.
+The announce enables the relay to know which publisher to forward a
+SUBSCRIBE to.
 
-Relays MUST ensure that publishers are authorized by verifying that the
-publisher is authorized to publish the content associated with the set of
-tracks whose Track Namespace matches the announced namespace. Where the
-authorization and identification of the publisher occurs depends on the way the
-relay is managed and is application specific.
+Relays MUST ensure that publishers are authorized by:
+
+- Verifying that the publisher is authorized to publish the content
+  associated with the set of tracks whose Track Namespace matches the
+  announced namespace. Where the authorization and identification of
+  the publisher occurs depends on the way the relay is managed and
+  is application specific.
 
 A Relay can receive announcements from multiple publishers for the same
 Track Namespace and it SHOULD respond with the same response to each of the
 publishers, as though it was responding to an ANNOUNCE
-from a single publisher for a given track namespace.
+from a single publisher for a given tracknamespace.
 
 When a publisher wants to stop
 new subscriptions for an announced namespace it sends an UNANNOUNCE.
@@ -918,7 +919,7 @@ publisher that has announced the subscription's namespace, unless it
 already has an active subscription for the Objects requested by the
 incoming SUBSCRIBE request from all available publishers.
 
-When a relay receives an incoming ANNOUNCE for a given namespace, for
+When a relay receives an incoming ANNOUCE for a given namespace, for
 each active upstream subscription that matches that namespace, it SHOULD send a
 SUBSCRIBE to the publisher that sent the ANNOUNCE.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -720,7 +720,7 @@ with FETCH_CANCEL, but MUST send FETCH_CANCEL.
 The Publisher can destroy subscription or fetch state as soon as it has received
 UNSUBSCRIBE or FETCH_CANCEL, respectively. It MUST reset any open streams
 associated with the SUBSCRIBE or FETCH. It can also destroy state after closing
-on the FETCH data stream.
+the FETCH data stream.
 
 The publisher can immediately delete SUBSCRIBE state after sending
 SUBSCRIBE_DONE, but MUST NOT send it until it has closed all related streams. It
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -646,9 +646,10 @@ SUBSCRIBE_ANNOUNCES to publishers/relays it has discovered. The recipient of
 this message will send any relevant ANNOUNCE messages for that namespace, or
 subset of that namespace.
 
-A publisher MUST send exactly one SUBSCRIBE_NAMESPACES_OK or
-SUBSCRIBE_NAMESPACES_ERROR. The subscriber SHOULD close the session with a
-protocol error if it detects receiving more than one.
+A publisher MUST send exactly one SUBSCRIBE_ANNOUNCES_OK or
+SUBSCRIBE_ANNOUNCES_ERROR  in response to a SUBSCRIBE_ANNOUNCES.
+The subscriber SHOULD close the session with a protocol error if it detects receiving
+more than one.
 
 An UNSUBSCRIBE_NAMESPACES withdraws a previous SUBSCRIBE_NAMESPACES. It does
 not prohibit the receiver (publisher) from sending further ANNOUNCE messages.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -712,8 +712,8 @@ of a SUBSCRIBE_DONE or SUBSCRIBE_ERROR. Note that SUBSCRIBE_DONE does not
 usually indicate that state can immediately be destroyed, see
 {{message-subscribe-done}}.
 
-A subscriber keeps FETCH state until it sends FETCH_CANCEL; receives
-FETCH_ERROR; or receives a FIN or RESET_STREAM for the FETCH data stream. If the
+A subscriber keeps FETCH state until it sends FETCH_CANCEL, receives
+FETCH_ERROR, or receives a FIN or RESET_STREAM for the FETCH data stream. If the
 data stream is already open, it MAY send STOP_SENDING for the data stream along
 with FETCH_CANCEL, but MUST send FETCH_CANCEL.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -911,7 +911,10 @@ namespace it previously responded ANNOUNCE_OK to by sending an
 ANNOUNCE_CANCEL.
 
 A relay manages sessions from multiple publishers and subscribers,
-connecting them based on the track namespace.
+connecting them based on the track namespace. This MUST use an exact
+match on track namespace unless otherwise negotiated by the application.
+For example, a SUBSCRIBE namespace=foobar message will be forwarded to
+the session that sent ANNOUNCE namespace=foobar.
 
 When a relay receives an incoming SUBSCRIBE request that triggers an
 upstream subscription, it SHOULD send a SUBSCRIBE request to each
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -653,7 +653,7 @@ more than one.
 
 An UNSUBSCRIBE_NAMESPACES withdraws a previous SUBSCRIBE_NAMESPACES. It does
 not prohibit the receiver (publisher) from sending further ANNOUNCE messages.
-The receiver of a SUBSCRIBE_NAMESPACES_OK or SUBSCRIBE_NAMESPACES_ERROR should
+The receiver of a SUBSCRIBE_NAMESPACES_OK or SUBSCRIBE_NAMESPACES_ERROR ought to
 forward the result to the application, so that it can make decisions about
 further publishers to contact.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -657,10 +657,13 @@ protocol error if it detects receiving more than one.
 
 ## ANNOUNCE
 
-A publisher MAY send ANNOUNCE messages to any subscriber. It SHOULD send them
-if it has received a SUBSCRIBE_ANNOUNCES for that namespace, or a superset of
-that namespace. An ANNOUNCE increases the likelihood of a subsequent successful
-SUBSCRIBE or FETCH.
+A publisher MAY send ANNOUNCE messages to any subscriber. An ANNOUNCE increases
+the likelihood of a subsequent successful SUBSCRIBE or FETCH. It SHOULD send
+an ANNOUNCE if it has received a SUBSCRIBE_ANNOUNCES for that namespace, or a
+superset of that namespace. However, a publisher SHOULD NOT send an ANNOUNCE
+advertising a namespace equal to, or a subset of, a namespace for which the
+receiver sent an earlier ANNOUNCE (i.e. an ANNOUNCE ought not to be echoed back
+to its sender).
 
 An UNANNOUNCE message negates the meaning of an ANNOUNCE, although it is not a
 protocol error for the subscriber to send a SUBSCRIBE or FETCH message anyway.
@@ -710,48 +713,6 @@ A publisher MUST send exactly one SUBSCRIBE_OK or SUBSCRIBE_ERROR in response to
 a SUBSCRIBE. It MUST send exactly one FETCH_OK or FETCH_ERROR in response to a
 FETCH. The subscriber SHOULD close the session with a protocol error if it
 detects receiving more than one.
-
-The application SHOULD use a relevant error code in SUBSCRIBE_ERROR or
-FETCH_ERROR, as defined below:
-
-|------|---------------------------|
-| Code | Reason                    |
-|-----:|:--------------------------|
-| 0x0  | Internal Error            |
-|------|---------------------------|
-| 0x1  | Invalid Range             |
-|------|---------------------------|
-| 0x2  | Retry Track Alias         |
-|------|---------------------------|
-| 0x3  | Track Does Not Exist      |
-|------|---------------------------|
-| 0x4  | Unauthorized              |
-|------|---------------------------|
-| 0x5  | Timeout                   |
-|------|---------------------------|
-
-The application SHOULD use a relevant status code in
-SUBSCRIBE_DONE or a FETCH RESET_STREAM, as defined below:
-
-|------|---------------------------|
-| Code | Reason                    |
-|-----:|:--------------------------|
-| 0x0  | Unsubscribed              |
-|------|---------------------------|
-| 0x1  | Internal Error            |
-|------|---------------------------|
-| 0x2  | Unauthorized              |
-|------|---------------------------|
-| 0x3  | Track Ended               |
-|------|---------------------------|
-| 0x4  | Subscription Ended        |
-|------|---------------------------|
-| 0x5  | Going Away                |
-|------|---------------------------|
-| 0x6  | Expired                   |
-|------|---------------------------|
-| 0x7  | Too Far Behind            |
-|------|---------------------------|
 
 # Priorities {#priorities}
 
@@ -1623,6 +1584,23 @@ message for which this response is provided.
 
 * Reason Phrase: Provides the reason for announcement error.
 
+The application SHOULD use a relevant error code in an ANNOUNCE_ERROR, as
+defined below.
+
+|------|---------------------------|
+| Code | Reason                    |
+|-----:|:--------------------------|
+| 0x0  | Internal Error            |
+|------|---------------------------|
+| 0x1  | Uninterested              |
+|------|---------------------------|
+| 0x2  | Unauthorized              |
+|------|---------------------------|
+| 0x3  | Timeout                   |
+|------|---------------------------|
+| 0x4  | Announce Not Supported    |
+|------|---------------------------|
+
 ## ANNOUNCE_CANCEL {#message-announce-cancel}
 
 The subscriber sends an `ANNOUNCE_CANCEL` control message to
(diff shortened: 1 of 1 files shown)


ianswett (commented):
I still find some of the ANNOUNCE flow overly complex, but this does a good job of describing how you might use the defined messages.

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,93 @@ and announcements. The client can choose to delay closing the session if it
...
+
+## SUBSCRIBE_ANNOUNCES
+
+If the subscriber is aware of a namespace of interest, it can send
+SUBSCRIBE_ANNOUNCES to publishers/relays it has discovered. This message
+increases the likelihood that publishers will send relevant ANNOUNCE messages
+for that namespace.
+
+An UNSUBSCRIBE_NAMESPACES negates the effect of a SUBSCRIBE_NAMESPACES. It does
+not prohibit the receiver from sending further ANNOUNCE messages. The receiver
```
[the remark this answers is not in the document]
martinduke: yes.

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,93 @@ and announcements. The client can choose to delay closing the session if it
...
+ANNOUNCE messages provide an in-band means of discovery of subscribers and
+publishers.
+
+The syntax of these messages is described in {{message}}.
+
+## SUBSCRIBE_ANNOUNCES
+
+If the subscriber is aware of a namespace of interest, it can send
+SUBSCRIBE_ANNOUNCES to publishers/relays it has discovered. This message
+increases the likelihood that publishers will send relevant ANNOUNCE messages
```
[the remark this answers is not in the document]
martinduke: The normative words are in the ANNOUNCE section. I had SHOULD under the general theory of "server's gonna do..." but I'll change it to MUST because I don't think it matters that much.

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,98 @@ and announcements. The client can choose to delay closing the session if it
...
+for that namespace.
+
+An UNSUBSCRIBE_NAMESPACES negates the effect of a SUBSCRIBE_NAMESPACES. It does
```
afrind, line 636: "negates the effect" has me reaching for a D20

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,98 @@ and announcements. The client can choose to delay closing the session if it
...
+The receiver of a SUBSCRIBE_NAMESPACES_OK or SUBSCRIBE_NAMESPACES_ERROR should
+forward the result to the application, so that it can make decisions about
+further publishers to contact. A publisher MUST send exactly one
+SUBSCRIBE_NAMESPACES_OK or SUBSCRIBE_NAMESPACES_ERROR. The subscriber SHOULD
+close the session with a protocol error if it detects receiving more than one.
+
+## ANNOUNCE
+
+A publisher MAY send ANNOUNCE messages to any subscriber. An ANNOUNCE increases
+the likelihood of a subsequent successful SUBSCRIBE or FETCH.
```
afrind, line 647: I don't think "increases the likelihood" is the right framing, it's not a +1 modifier to your successful subscribe roll.  ANNOUNCE does what ANNOUNCE does -- it indicates where to route SUBSCRIBE/FETCH that match a namespace.

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,98 @@ and announcements. The client can choose to delay closing the session if it
...
+
+If a publisher is authoritative for a given namespace, or has received an
+ANNOUNCE for that namespace from an upstream publisher, it MUST send an ANNOUNCE
```
afrind, line 650: `or *is a relay that* as received`, since generic publishers don't have the concept of upstream publishers.

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,98 @@ and announcements. The client can choose to delay closing the session if it
...
+to any subscriber that has subscribed to ANNOUNCE for that namespace, a
+superset of that namespace, or a subset of that namespace. A publisher MAY send
```
afrind, line 652: We should definitely discuss the superset case.

I forget, does a relay route SUBSCRIBE based on exact namespace match pr prefix match?  If it requires an exact match, then what use is knowing who is authoritative for "Foo" if you need to subscribe to ("Foo", "Bar")?

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,98 @@ and announcements. The client can choose to delay closing the session if it
...
+the ANNOUNCE to any other subscriber.
+
+However, a publisher SHOULD NOT send an ANNOUNCE advertising a namespace equal
+to, or a subset of, a namespace for which the subscriber sent an earlier ANNOUNCE
```
afrind, line 656: I think you mean superset?  But again it depends on exact vs prefix matching.

C->R ANNOUNCE Foo/Bar (I am authoritative for Foo/Bar)
R->C ANNOUNCE Foo (Someone else connected to the relay is authoritative for all of Foo)

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,98 @@ and announcements. The client can choose to delay closing the session if it
...
+(i.e. an ANNOUNCE ought not to be echoed back to its sender).
+
+An UNANNOUNCE message negates the meaning of an ANNOUNCE, although it is not a
+protocol error for the subscriber to send a SUBSCRIBE or FETCH message anyway.
```
afrind, line 660: after receiving an UNANNOUNCE

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,98 @@ and announcements. The client can choose to delay closing the session if it
...
+A subscriber can send ANNOUNCE_CANCEL, meaning it is no longer interested in a
```
afrind, line 661: ANNOUNCE_CANCEL is a weird one if it's only function is to prevent UNANNOUNCE?  But if we want to remove it let's not do it in this PR.

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,98 @@ and announcements. The client can choose to delay closing the session if it
...
+namespace, which also negates the ANNOUNCE: the publisher need not send
+UNANNOUNCE. A publisher and subscriber might seek alternate subscribers and
+publishers, respectively, for potential SUBSCRIBE and FETCH interactions.
+
+The receiver of an ANNOUNCE_OK or ANNOUNCE_ERROR SHOULD report this to the
+application to inform the search for additional subscribers for a namespace,
+or abandoning the attempt to publish a track. This might be especially useful
+in upload or chat applications. A subscriber MUST send exactly one ANNOUNCE_OK
+or ANNOUNCE_ERROR in response to an ANNOUNCE. The publisher SHOULD close the
+session with a protocol error if it detects receiving more than one.
```
afrind, line 671: Why not MUST?

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,98 @@ and announcements. The client can choose to delay closing the session if it
...
+
+A subscriber MAY send a SUBSCRIBE or FETCH for a track to any publisher. If it
+has received an ANNOUNCE with a namespace that includes that track, it SHOULD
+only request it from the senders of those ANNOUNCE messages.
+
+The central interaction with a publisher is to send SUBSCRIBE and/or FETCH for
+a particular track in a namespace. The subscriber expects to receive a
+SUBSCRIBE_OK/FETCH_OK and objects from the track.
+
+A subscriber keeps SUBSCRIBE state until it sends UNSUBSCRIBE, or after receipt
```
afrind, line 683: Do we need to mention SUBSCRIBE_ERROR?

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,98 @@ and announcements. The client can choose to delay closing the session if it
...
+of a SUBSCRIBE_DONE.  Note that SUBSCRIBE_DONE does not usually indicate that
+state can immediately be destroyed, see {{message-subscribe-done}}.
+
+A subscriber keeps FETCH state until it sends FETCH_CANCEL, or a FIN or
```
afrind, line 687: Do we need to mention FETCH_ERROR?

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,98 @@ and announcements. The client can choose to delay closing the session if it
...
+RESET_STREAM for the FETCH data stream. If the data stream is already open, it
+MAY send STOP_SENDING for the data stream, but MUST send FETCH_CANCEL.
```
afrind, line 689: Why does it have to send FETCH_CANCEL?

SHOULD send STOP_SENDING?

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,98 @@ and announcements. The client can choose to delay closing the session if it
...
+
+The Publisher can destroy subscription or fetch state as soon as it has received
+UNSUBSCRIBE or FETCH_CANCEL, respectively. It MUST reset any open streams
```
afrind, line 692: Why not STOP_SENDING on the fetch stream?

afrind: Do we talk about STOP_SENDING on subscribe streams anywhere?   Seems like those should elicit resets but shouldn't affect subscribe state.

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,98 @@ and announcements. The client can choose to delay closing the session if it
...
+SUBSCRIBE_DONE, but MUST NOT send it until it has closed all related streams. It
+can destroy all FETCH state after closing the data stream.
+
+A SUBSCRIBE_ERROR or FETCH_ERROR indicates no objects will be delivered, and
+both endpoints can immediately destroy relevant state. Objects MUST NOT be sent
+for requests that end with an error.
+
+A publisher MUST send exactly one SUBSCRIBE_OK or SUBSCRIBE_ERROR in response to
+a SUBSCRIBE. It MUST send exactly one FETCH_OK or FETCH_ERROR in response to a
+FETCH. The subscriber SHOULD close the session with a protocol error if it
```
afrind, line 705: MUST

draft-ietf-moq-transport.md
```diff
@@ -721,31 +813,22 @@ Relays MAY cache Objects, but are not required to.
...
-will need to make an upstream subscription.  The relay SHOULD NOT
-return a SUBCRIBE_OK until at least one SUBSCRIBE_OK has been
-received for the track, to ensure the Group Order is correct.
+encompassing session. The specifics of how a relay authorizes a user are outside
+the scope of this specification.
+
+The relay will have to send an upstream SUBSCRIBE and/or FETCH if it does not
+have all the objects in the FETCH, or is not currently subscribed to the full
+requested range. In this case, it MUST withhold sending its own SUBSCRIBE_OK or
+FETCH_OK until receiving one from upstream.
```
afrind, line 826: Let's make this change in a separate PR, since it's not ANNOUNCE related.

draft-ietf-moq-transport.md
```diff
@@ -754,63 +837,20 @@ publishers and forward objects to active matching subscriptions.
 If multiple objects are received with the same Full Track Name,
 Group ID and Object ID, Relays MAY ignore subsequently received Objects
 or MAY use them to update the cache. Implementations that update the
-cache need to be protect against cache poisoning.
+cache need to protect against cache poisoning.
 
-Objects MUST NOT be sent for unsuccessful subscriptions, and if a subscriber
-receives a SUBSCRIBE_ERROR after receiving objects, it MUST close the session
-with a 'Protocol Violation'.
+Caching can reduce the number of upstream FETCH requests.
```
afrind, line 842: Where did this come from -- also not ANNOUNCE

draft-ietf-moq-transport.md
```diff
@@ -754,63 +837,20 @@ publishers and forward objects to active matching subscriptions.
...
 
 A relay MUST NOT reorder or drop objects received on a multi-object stream when
 forwarding to subscribers, unless it has application specific information.
 
-Relays MAY aggregate authorized subscriptions for a given track when
-multiple subscribers request the same track. Subscription aggregation
+Relays SHOULD aggregate authorized subscriptions for a given track when
+multiple subscribers request the same track, using SUBSCRIBE_UPDATE when new
+subscriptions change the aggregate properties. Subscription aggregation
```
afrind, line 849: Not ANNOUNCE

draft-ietf-moq-transport.md
```diff
@@ -855,18 +888,15 @@ namespace it previously responded ANNOUNCE_OK to by sending an
 ANNOUNCE_CANCEL.
 
 A relay manages sessions from multiple publishers and subscribers,
-connecting them based on the track namespace. This MUST use an exact
```
afrind, line 858: You removed the normative exact match

draft-ietf-moq-transport.md
```diff
@@ -1378,8 +1408,7 @@ See {{priorities}}.
 
 A subscriber issues a `UNSUBSCRIBE` message to a publisher indicating it is no
 longer interested in receiving media for the specified track and Objects
-should stop being sent as soon as possible.  The publisher sends a
-SUBSCRIBE_DONE to acknowledge the unsubscribe was successful.
+should stop being sent as soon as possible.
```
afrind, line 1411: non-normative should

draft-ietf-moq-transport.md
```diff
@@ -1545,16 +1574,29 @@ message for which this response is provided.
 
 * Reason Phrase: Provides the reason for announcement error.
 
+The application SHOULD use a relevant error code in an ANNOUNCE_ERROR, as
+defined below.
```
afrind, line 1578: Maybe just "the following error codes are defined"?

draft-ietf-moq-transport.md
```diff
@@ -1736,6 +1778,23 @@ SUBSCRIBE_ERROR
   the subscriber MUST close the connection with a Duplicate Track Alias error
   ({{session-termination}}).
 
+The application SHOULD use a relevant error code in SUBSCRIBE_ERROR, as defined below:
```
afrind, line 1781: SHOULD seems silly here

draft-ietf-moq-transport.md
```diff
@@ -1801,6 +1860,22 @@ FETCH_ERROR
 
 * Reason Phrase: Provides the reason for fetch error.
 
+The application SHOULD use a relevant error code in FETCH_ERROR, as defined below:
```
afrind, line 1863: And here

afrind (commented):
Individual Review:

Thanks for tackling this.  There's a lot going on here.  

Moving the error codes is helpful but seems like a great candidate for a standalone editorial PR.

There are some normative changes to relay handling not related to ANNOUNCE which should also be in a separate PR.

There's a subtle deletion of the "exact" namespace matching behavior.  This is less problematic in the face of structured namespaces, but I forget if that was something we discussed and had consensus on?

draft-ietf-moq-transport.md
suhasHere: Agree with Alan. Announce acts as anchor point to route subscriptions to as it identifies where the original publisher can be found.

draft-ietf-moq-transport.md
martinduke: It allows the publisher to delete ANNOUNCE state.

draft-ietf-moq-transport.md
martinduke: Clearly if I want to know where to get Foo/Bar, the provider of Foo is sufficient?

The text is unclear; I'll add something. Yes, it must be a prefix match to make any sense. all track names are subsets of a namespace; there is no exact match.

draft-ietf-moq-transport.md
martinduke: It might not detect it. When I get ANNOUNCE_ERROR the state is gone. If another ANNOUNCE_ERROR or _OK arrives, I don't know if it's a protocol error or a response to an ANNOUNCE where I already sent UNANNOUNCE and nuked the state.

I don't believe in forcing endpoints to keep a bunch of state just to enforce protocol errors, unless they are security risks.

draft-ietf-moq-transport.md
martinduke: As you mentioned in a different issue, in case the data stream isn't open yet. The state of the stream may not be the same at both endpoints, so this makes it easier.

FETCH_CANCEL results in RESET_STREAM, just the same as STOP_SENDING. so the advantage is... what? that the QUIC control message is slightly faster than an MoQT control message?

draft-ietf-moq-transport.md
martinduke: I added that it can delete state after closing the data stream, which is a more general case than getting STOP_SENDING.

STOP_SENDING is irrelevant to SUBSCRIBE state. If it's not discussed, file a new issue.

draft-ietf-moq-transport.md
martinduke: Again, my implementation will not detect a second message after _ERROR.

draft-ietf-moq-transport.md
martinduke: This is eliminating duplicate text, with some cleanup at the same time.

draft-ietf-moq-transport.md
martinduke: Multi-tiered namespaces don't work if we don't update this.

draft-ietf-moq-transport.md
martinduke: It seems silly to have a separate PR for this editorial line.

draft-ietf-moq-transport.md
martinduke: this is just housekeeping. There is already a ton of editing in this section.

martinduke:
> Individual Review:
> 
> Thanks for tackling this. There's a lot going on here.
> 
> Moving the error codes is helpful but seems like a great candidate for a standalone editorial PR.

This PR is already heavily editing the relay section for redundant and poorly located text. This falls into that theme; it is not specific to relays.

> 
> There are some normative changes to relay handling not related to ANNOUNCE which should also be in a separate PR.

Again, a huge rewrite of the Relay section is in scope for this.

> 
> There's a subtle deletion of the "exact" namespace matching behavior. This is less problematic in the face of structured namespaces, but I forget if that was something we discussed and had consensus on?

I don't see how structured namespaces can work with exact matching.

draft-ietf-moq-transport.md
martinduke: Moved to #645

draft-ietf-moq-transport.md
martinduke: moved to #645

draft-ietf-moq-transport.md
martinduke: I thought this was uncontroversial, but I guess not. I'll just revert it, because I don't really care

draft-ietf-moq-transport.md
martinduke: reverted, because I don't think SHOULD/MAY matters.

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
 expects more OBJECTs to be delivered. The server closes the session with a
 'GOAWAY Timeout' if the client doesn't close the session quickly enough.
 
+# Track Discovery and Retrieval (#track-discovery}
+
+Discovery of MoQT servers is always done out-of-band. Track discovery is done in
```
afrind, line 617: s/is done/can be done/ ?

There are cases where track discovery is also out of band (eg catalog, hard coded in specifications), right?

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+the context of an established MoQT session. The session client might be a
+subscriber, publisher, or both.
```
afrind, line 619: > The session client might be a subscriber, publisher, or both.

This is true, but is it related to track discovery?

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+
+Given sufficient out of band information, it is valid for a subscriber
+to send a SUBSCRIBE or FETCH message to a publisher (including a relay) without
+any previous MoQT messages besides SETUP. However, SUBSCRIBE_NAMESPACES and
```
afrind, line 623: SUBSCRIBE_ANNOUNCES

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+ANNOUNCE messages provide an in-band means of discovery of subscribers and
```
afrind, line 624: I don't think you can discover subscribers via these messages.  

...discovery of publishers and track namespaces?

Aside: I've been thinking there might be value in an optional `track name` (or names) in ANNOUNCE, to assist with catalog-less discovery.

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+publishers.
+
+The syntax of these messages is described in {{message}}.
+
+## SUBSCRIBE_ANNOUNCES
+
+If the subscriber is aware of a namespace of interest, it can send
+SUBSCRIBE_ANNOUNCES to publishers/relays it has discovered. This message
+increases the likelihood that publishers will send relevant ANNOUNCE messages
```
afrind, line 633: I don't like 'increases the likelihood'.  Maybe:

The recipient of the message will send any relevant ANNOUNCE messages for that namespace.

draft-ietf-moq-transport.md
afrind: 'withdraws a previous SUBSCRIBE_NAMESPACES' ?

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+for that namespace.
+
+An UNSUBSCRIBE_NAMESPACES negates the effect of a SUBSCRIBE_NAMESPACES. It does
+not prohibit the receiver (publisher) from sending further ANNOUNCE messages.
+The receiver of a SUBSCRIBE_NAMESPACES_OK or SUBSCRIBE_NAMESPACES_ERROR should
+forward the result to the application, so that it can make decisions about
+further publishers to contact. A publisher MUST send exactly one
```
afrind, line 640: 'futher publishers or relays' ?

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+SUBSCRIBE_NAMESPACES_OK or SUBSCRIBE_NAMESPACES_ERROR. The subscriber SHOULD
+close the session with a protocol error if it detects receiving more than one.
```
afrind, line 642: This bit should be moved ahead of the UNSUBSCRIBE_ANNOUNCES description

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+
+## ANNOUNCE
+
+A publisher MAY send ANNOUNCE messages to any subscriber. An ANNOUNCE indicates
+to the subscriber where to route a SUBSCRIBE or FETCH for that namespace. A
+subscriber MAY send SUBSCRIBE or FETCH for a namespace without having received
+an ANNOUNCE for it.
+
+If a publisher is authoritative for a given namespace, or is a relay that has
+received an ANNOUNCE for that namespace from an upstream publisher, it MUST send
```
afrind, line 652: 'an authorized ANNOUNCE' maybe?

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+an ANNOUNCE to any subscriber that has subscribed to ANNOUNCE for that
+namespace, a superset of that namespace, or a subset of that namespace. A
```
afrind, line 654: Walking through an example:

A: SUBSCRIBE_ANNOUCNES /meeting123
B: SUBSCRIBE_ANNOUNCES /meeting123/video
C: SUBSCRIBE_ANNOUNCES /meeting123/video/afrind

D: ANNOUNCE /meeting123/video

Why does C get ANNOUNCE /meeting123/video?  Given that SUBSCRIBE only matches ANNOUNCE when the namespaces are an *exact* match (not a prefix match), this ANNOUNCE is not relevant to C.

I'm inclined to remove "superset" above.  If the wg changes SUBSCRIBE/ANNOUNCE matching, we can make the matching change here.

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+publisher MAY send the ANNOUNCE to any other subscriber.
```
afrind, line 655: This MAY opens up the door for "full-mesh" relays that just forward all ANNOUNCEs everywhere.  Maybe we need a little note that indiscriminate broadcasts of ANNOUNCE messages is discouraged?  Or warns of DoS/perf problems?

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+
+However, a publisher SHOULD NOT send an ANNOUNCE advertising a namespace that
```
afrind, line 657: nit: Leading with 'However, ' feels jarring, maybe

A publisher SHOULD NOT, however, send ...

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+exactly matches a namespace for which the subscriber sent an earlier ANNOUNCE
+(i.e. an ANNOUNCE ought not to be echoed back to its sender).
+
+An UNANNOUNCE message negates the meaning of an ANNOUNCE, although it is not a
```
afrind, line 661: s/negates the meaning of an/withdraws a previous/

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+protocol error for the subscriber to send a SUBSCRIBE or FETCH message after
+receiving an UNANNOUNCE.
+
+A subscriber can send ANNOUNCE_CANCEL, meaning it is no longer interested in a
+namespace, which also negates the ANNOUNCE: the publisher need not send
```
afrind, line 666: I'd say something like:

A subscriber can send ANNOUNCE_CANCEL to indicate cancellation of a previously accepted ANNOUNCE, for example due to expiration of authorization credentials.  The message enables the publisher to ANNOUNCE again with refreshed authorization, or discard associated state.  After receiving an ANNOUNCE_CANCEL, the publisher does not send UNANNOUNCE.

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+UNANNOUNCE. A publisher and subscriber might seek alternate subscribers and
+publishers, respectively, for potential SUBSCRIBE and FETCH interactions.
+
+The receiver of an ANNOUNCE_OK or ANNOUNCE_ERROR SHOULD report this to the
+application to inform the search for additional subscribers for a namespace,
+or abandoning the attempt to publish a track. This might be especially useful
```
afrind, line 672: > attempt to publish a track

attempt to publish under this namespace

(could be multiple tracks)

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+An UNANNOUNCE message negates the meaning of an ANNOUNCE, although it is not a
+protocol error for the subscriber to send a SUBSCRIBE or FETCH message after
+receiving an UNANNOUNCE.
+
+A subscriber can send ANNOUNCE_CANCEL, meaning it is no longer interested in a
+namespace, which also negates the ANNOUNCE: the publisher need not send
+UNANNOUNCE. A publisher and subscriber might seek alternate subscribers and
+publishers, respectively, for potential SUBSCRIBE and FETCH interactions.
+
+The receiver of an ANNOUNCE_OK or ANNOUNCE_ERROR SHOULD report this to the
```
afrind, line 670: Should this section move closer to ANNOUNCE (above UNANNOUNCE/ANNOUNCE_CANCEL)

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+of a SUBSCRIBE_DONE or SUBSCRIBE_ERROR. Note that SUBSCRIBE_DONE does not
+usually indicate that state can immediately be destroyed, see
+{{message-subscribe-done}}.
+
+A subscriber keeps FETCH state until it sends FETCH_CANCEL; receives
+FETCH_ERROR; or receives a FIN or RESET_STREAM for the FETCH data stream. If the
+data stream is already open, it MAY send STOP_SENDING for the data stream along
+with FETCH_CANCEL, but MUST send FETCH_CANCEL.
+
+The Publisher can destroy subscription or fetch state as soon as it has received
```
afrind, line 697: The order you have is:

subscriber SUBSCRIBE state
subscriber FETCH state
publisher FETCH state
publisher SUBSCRIBE state

which I find dissatisfying. Can you reorder (either `sub sub/sub fetch/pub sub/pub fetch`, or `sub sub/pub sub/sub fetch/pub fetch` is fine)

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+or ANNOUNCE_ERROR in response to an ANNOUNCE. The publisher SHOULD close the
+session with a protocol error if it detects receiving more than one.
+
+## SUBSCRIBE/FETCH
+
+A subscriber MAY send a SUBSCRIBE or FETCH for a track to any publisher. If it
+has received an ANNOUNCE with a namespace that includes that track, it SHOULD
+only request it from the senders of those ANNOUNCE messages.
+
+The central interaction with a publisher is to send SUBSCRIBE and/or FETCH for
```
afrind, line 683: Maybe reorder these to paragraphs?

draft-ietf-moq-transport.md
```diff
@@ -1545,16 +1585,28 @@ message for which this response is provided.
 
 * Reason Phrase: Provides the reason for announcement error.
 
+The following error codes are defined:
```
afrind, line 1588: I'm not going to die on this hill, but moving the error code tables could easily go in another PR.

draft-ietf-moq-transport.md
```diff
@@ -612,6 +612,105 @@ and announcements. The client can choose to delay closing the session if it
...
+
+The publisher can immediately delete SUBSCRIBE state after sending
+SUBSCRIBE_DONE, but MUST NOT send it until it has closed all related streams. It
+can destroy all FETCH state after closing the data stream.
+
+A SUBSCRIBE_ERROR or FETCH_ERROR indicates no objects will be delivered, and
+both endpoints can immediately destroy relevant state. Objects MUST NOT be sent
+for requests that end with an error.
+
+A publisher MUST send exactly one SUBSCRIBE_OK or SUBSCRIBE_ERROR in response to
```
afrind, line 710: Move closer to the top of the section, before the state bits.

afrind (commented):
This looks better.  A bunch of editorial comments/suggestions.  The substantive comment is about superset.

draft-ietf-moq-transport.md
martinduke: SUBSCRIBE_ANNOUNCES discovers subscribers.

draft-ietf-moq-transport.md
martinduke: A relay is a publisher.

draft-ietf-moq-transport.md
martinduke: That does not accurately describe the publisher paragraphs.

One is via UNSUBSCRIBE/FETCH_CANCEL.

One is SUBSCRIBE_DONE/close the fetch stream.

The first is publisher destroying state to a subscriber message, the second is a local decision.

draft-ietf-moq-transport.md
martinduke: Added "respectively" to make it clearer.

draft-ietf-moq-transport.md
martinduke: I think it totally depends on the use case. I personally don't know enough to make a recommendation here.

martinduke:
The flexibility built into this architecture can lead to routing loops.

Take relays A, B, and C, which are connected to each other.

C sends SUBSCRIBE_ANNOUNCES for namespace Foo to B, but not A. B sends ("forwards") SUBSCRIBE_ANNOUNCES to A for Foo.

Then A gets an ANNOUNCE for Foo from Upstream. it sends it to B, which passes it on to C. C forwards that ANNOUNCE unsolicited to A. Now A has a backup path to Foo, so even if its upstream session dies, it will not send UNANNOUNCE.

Even without SUBSCRIBE_ANNOUNCE, there could be this chain of events.

One way out of this would be to require SUBSCRIBE_ANNOUNCES to send an ANNOUNCE.

draft-ietf-moq-transport.md
```diff
@@ -626,6 +626,108 @@ and announcements. The client can choose to delay closing the session if it
...
+usually indicate that state can immediately be destroyed, see
+{{message-subscribe-done}}.
+
+A subscriber keeps FETCH state until it sends FETCH_CANCEL; receives
+FETCH_ERROR; or receives a FIN or RESET_STREAM for the FETCH data stream. If the
+data stream is already open, it MAY send STOP_SENDING for the data stream along
+with FETCH_CANCEL, but MUST send FETCH_CANCEL.
+
+The Publisher can destroy subscription or fetch state as soon as it has received
+UNSUBSCRIBE or FETCH_CANCEL, respectively. It MUST reset any open streams
```
afrind, line 720: > It MUST reset

Is this new normative language and if so it doesn't seem very ANNOUNCE related.  That said it seems like the right behavior -- if you want less abrupt termination, use SUBSCRIBE_UPDATE?

draft-ietf-moq-transport.md
```diff
@@ -868,10 +912,7 @@ namespace it previously responded ANNOUNCE_OK to by sending an
 ANNOUNCE_CANCEL.
 
 A relay manages sessions from multiple publishers and subscribers,
-connecting them based on the track namespace. This MUST use an exact
-match on track namespace unless otherwise negotiated by the application.
-For example, a SUBSCRIBE namespace=foobar message will be forwarded to
-the session that sent ANNOUNCE namespace=foobar.
```
afrind, line 874: Did we lose the example?

afrind (commented):
Individual Review: 

lgtm with a couple minor comments

draft-ietf-moq-transport.md
```diff
@@ -626,6 +626,108 @@ and announcements. The client can choose to delay closing the session if it
...
+
+A subscriber keeps FETCH state until it sends FETCH_CANCEL; receives
+FETCH_ERROR; or receives a FIN or RESET_STREAM for the FETCH data stream. If the
+data stream is already open, it MAY send STOP_SENDING for the data stream along
+with FETCH_CANCEL, but MUST send FETCH_CANCEL.
+
+The Publisher can destroy subscription or fetch state as soon as it has received
+UNSUBSCRIBE or FETCH_CANCEL, respectively. It MUST reset any open streams
+associated with the SUBSCRIBE or FETCH. It can also destroy state after closing
+on the FETCH data stream.
```
ianswett, line 722:
```suggestion
the FETCH data stream.
```

draft-ietf-moq-transport.md
```diff
@@ -735,17 +837,13 @@ Relays MAY cache Objects, but are not required to.
 
 ## Subscriber Interactions
 
-Subscribers interact with the Relays by sending a SUBSCRIBE or FETCH
-control message for the tracks of interest. Relays MUST ensure subscribers are
-authorized to access the content associated with the track. The authorization
+Subscribers interact with the Relays by sending a SUBSCRIBE
+({{message-subscribe-req}}) control message for the tracks of
+interest. Relays MUST ensure subscribers are authorized to access the
```
ianswett, line 842:
```suggestion
Subscribers subscribe to tracks by sending  a SUBSCRIBE
({{message-subscribe-req}}) control message for each track of
interest. Relays MUST ensure subscribers are authorized to access the
```

draft-ietf-moq-transport.md
```diff
@@ -769,63 +865,16 @@ Group ID and Object ID, Relays MAY ignore subsequently received Objects
...
-receives a SUBSCRIBE_ERROR after receiving objects, it MUST close the session
-with a 'Protocol Violation'.
-
 A relay MUST NOT reorder or drop objects received on a multi-object stream when
 forwarding to subscribers, unless it has application specific information.
 
-Relays MAY aggregate authorized subscriptions for a given track when
-multiple subscribers request the same track. Subscription aggregation
+Relays SHOULD aggregate authorized subscriptions for a given track when
+multiple subscribers request the same track, using SUBSCRIBE_UPDATE when new
```
ianswett, line 872: Is the "using SUBSCRIBE_UPDATE..." clause useful without explanation about how/when to use it?

draft-ietf-moq-transport.md
```diff
@@ -843,19 +892,14 @@ to the old relay can be stopped with an UNSUBSCRIBE.
 Publishing through the relay starts with publisher sending ANNOUNCE
 control message with a `Track Namespace` ({{model-track}}).
 The ANNOUNCE enables the relay to know which publisher to forward a
-SUBSCRIBE to if the track belongs to the namespace in the ANNOUNCE.
+SUBSCRIBE to.
 
-Relays MUST verify that publishers are authorized to publish
-the content associated with the set of
+Relays MUST ensure that publishers are authorized by verifying that the
+publisher is authorized to publish the content associated with the set of
```
ianswett, line 898:
```suggestion
Relays MUST verify that publishers are authorized to publish
the content associated with the set of
```

ianswett (commented):
Some more comments and suggestions

draft-ietf-moq-transport.md
martinduke: This whole section is about managing state related to SUBSCRIBE_ANNOUNCES, ANNOUNCE, SUBSCRIBE and FETCH. I don't see how this bit can be sensibly separated.

afrind: Maybe retitle PR to "Message flows and endpoint state management"?

draft-ietf-moq-transport.md
```diff
@@ -626,6 +626,108 @@ and announcements. The client can choose to delay closing the session if it
...
+this message will send any relevant ANNOUNCE messages for that namespace, or
+subset of that namespace.
+
+A publisher MUST send exactly one SUBSCRIBE_NAMESPACES_OK or
+SUBSCRIBE_NAMESPACES_ERROR. The subscriber SHOULD close the session with a
+protocol error if it detects receiving more than one.
+
+An UNSUBSCRIBE_NAMESPACES withdraws a previous SUBSCRIBE_NAMESPACES. It does
+not prohibit the receiver (publisher) from sending further ANNOUNCE messages.
+The receiver of a SUBSCRIBE_NAMESPACES_OK or SUBSCRIBE_NAMESPACES_ERROR should
```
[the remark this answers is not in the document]
martinduke: I changed it to "ought to".

I'm surprised to hear you think it would be handled internally. Within a session, the decision to subscribe to a namespace has to come from the application, no? And if rejected, the application is going to have to initiate an MoQT session with someone else (or fix their authentication)?

draft-ietf-moq-transport.md
martinduke: Well, it says "when new subscriptions change the aggregate properties"

I was going to use the example of a new subscription having a  later end group, and therefore sending SUBSCRIBE_UPDATE. But of course that's illegal! So I just eliminated the clause.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
ianswett: I guess, but you can also SUBSCRIBE without ever receiving an ANNOUNCE, so I'm having a difficult time reasoning about how ANNOUNCE will be used.  But I'm fine with keeping it as is for now.

draft-ietf-moq-transport.md
```diff
@@ -626,6 +626,109 @@ and announcements. The client can choose to delay closing the session if it
 expects more OBJECTs to be delivered. The server closes the session with a
 'GOAWAY Timeout' if the client doesn't close the session quickly enough.
 
+# Track Discovery and Retrieval (#track-discovery}
+
+Discovery of MoQT servers is always done out-of-band. Track discovery can be
+done in the context of an established MoQT session.
```
ianswett, line 632: Maybe I'm confused, but I think you can discover announced namespaces, but not tracks, correct?

draft-ietf-moq-transport.md
```diff
@@ -626,6 +626,109 @@ and announcements. The client can choose to delay closing the session if it
...
+
+If the subscriber is aware of a namespace of interest, it can send
+SUBSCRIBE_ANNOUNCES to publishers/relays it has discovered. The recipient of
+this message will send any relevant ANNOUNCE messages for that namespace, or
+subset of that namespace.
+
+A publisher MUST send exactly one SUBSCRIBE_ANNOUNCES_OK or
+SUBSCRIBE_ANNOUNCES_ERROR  in response to a SUBSCRIBE_ANNOUNCES.
+The subscriber SHOULD close the session with a protocol error if it detects receiving
+more than one.
```
ianswett, line 652:
```suggestion
The subscriber SHOULD close the session with a protocol error if it detects
receiving more than one.
```

draft-ietf-moq-transport.md
```diff
@@ -626,6 +626,109 @@ and announcements. The client can choose to delay closing the session if it
...
+exactly matches a namespace for which the subscriber sent an earlier ANNOUNCE
+(i.e. an ANNOUNCE ought not to be echoed back to its sender).
+
+The receiver of an ANNOUNCE_OK or ANNOUNCE_ERROR SHOULD report this to the
+application to inform the search for additional subscribers for a namespace,
+or abandoning the attempt to publish under this namespace. This might be
+especially useful in upload or chat applications. A subscriber MUST send exactly
+one ANNOUNCE_OK or ANNOUNCE_ERROR in response to an ANNOUNCE. The publisher
+SHOULD close the session with a protocol error if it detects receiving more than
+one.
```
ianswett, line 683:
```suggestion
SHOULD close the session with a protocol error if it receives more than one.
```

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
martinduke: Sure, but if the app sends a SUBSCRIBE_ANNOUNCE surely it needs one, and needs to know if the one it asked for failed?

ianswett (approved):

draft-ietf-moq-transport.md
afrind: Still there

martinduke:
> Minor nits, lgtm.

The SUBSCRIBE_DONE thing is intended! It's part of the logic this protocol describes.

afrind:
> The SUBSCRIBE_DONE thing is intended! It's part of the logic this protocol describes.

"Still there" was supposed to refer to a non-normative _should_ on line 1411.  Removing the SUBSCRIBE_DONE ack to UNSUBSCRIBE seems fine to me.

draft-ietf-moq-transport.md
```diff
@@ -626,6 +626,108 @@ and announcements. The client can choose to delay closing the session if it
...
 
+# Track Discovery and Retrieval (#track-discovery}
+
+Discovery of MoQT servers is always done out-of-band. Namespace discovery can be
+done in the context of an established MoQT session.
+
+Given sufficient out of band information, it is valid for a subscriber
+to send a SUBSCRIBE or FETCH message to a publisher (including a relay) without
+any previous MoQT messages besides SETUP. However, SUBSCRIBE_ANNOUNCES and
+ANNOUNCE messages provide an in-band means of discovery of subscribers and
```
suhasHere, line 637: I still not sure if we can describe subscriber though ;)

martinduke:
Deleted #577 from the list of issues, since that is now solved by #656

draft-ietf-moq-transport.md
martinduke: I explicity defined the term.

martinduke:
> > The SUBSCRIBE_DONE thing is intended! It's part of the logic this protocol describes.
> 
> "Still there" was supposed to refer to a non-normative _should_ on line 1411. Removing the SUBSCRIBE_DONE ack to UNSUBSCRIBE seems fine to me.

Ah, got it, fixed.

martinduke:
I removed the "subscriber" and "publisher" definitions I just added, because these terms are already well defined in Section 1.2

draft-ietf-moq-transport.md
suhasHere: not sure how can discovery subscribers using announces/subscribe_announces.  These messages allow a way to discover a publisher(s) for a given namespace/namespace-prefix and they aid in routing a subscription to the publishers.

martinduke: An endpoint that receives a SUBSCRIBE_ANNOUNCES can now identify the sender as a subscriber.

ianswett:
I and the chairs acknowledge that this might not be perfect, but it clarifies the behavior in a meaningful way.

PR #645, opened 2025-01-08
Add FETCH to Relay section

martinduke:
Taking this out of PR #643 as a separate change.

This just updates the text to reflect that FETCH exists too.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -735,8 +735,9 @@ either SUBSCRIBE_OK or SUBSCRIBE_ERROR.
 
 The relay will have to send an upstream SUBSCRIBE and/or FETCH if it does not
 have all the objects in the FETCH, or is not currently subscribed to the full
-requested range. In this case, it MUST withhold sending its own SUBSCRIBE_OK or
-FETCH_OK until receiving one from upstream.
+requested range. In this case, it SHOULD withhold sending its own SUBSCRIBE_OK 
+until receiving one from upstream. It MUST withhold FETCH_OK until receiving
+one from upstream.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -734,11 +733,10 @@ SUBSCRIBE_OK ({{message-subscribe-ok}}) or SUBSCRIBE_ERROR
...
 
-If a relay does not already have a subscription for the track,
-or if the subscription does not cover all the requested Objects, it
-will need to make an upstream subscription.  The relay SHOULD NOT
-return a SUBCRIBE_OK until at least one SUBSCRIBE_OK has been
-received for the track, to ensure the Group Order is correct.
+The relay will have to send an upstream SUBSCRIBE and/or FETCH if it does not
+have all the objects in the FETCH, or is not currently subscribed to the full
+requested range. In this case, it MUST withhold sending its own SUBSCRIBE_OK or
+FETCH_OK until receiving one from upstream.
```
afrind, line 739: I believe some folks have designs where they want to send SUBSCRIBE_OK before hearing anything from upstream.  Given the presence of latest object in SUBSCRIBE_OK, that may be challenging, but perhaps a SHOULD is better than a MUST?   Maybe those relays will just set content-exists=false?

CC: @suhasHere

afrind (commented):
Individual Comment: 

This lgtm, modulo the MUST/SHOULD on subscribe OK

draft-ietf-moq-transport.md
martinduke: I disagree but don't care enough to fight about it. Please approve so we can ship this.

PR #651, opened 2025-01-20
Object Does Not Exist not Object Not Available

### LINKED ISSUE #629: No such status as "object not available"
In 6.7 FETCH it says

"All omitted objects have status Object Not Available."

This status doesn't exist. The closest status is "Object Does Not Exist"

I was surprised that we are sending EndOfGroup status, etc, but I can definitely live with that.

ianswett:
Fixes #629


martinduke (approved):

afrind:
lgtm

PR #652, opened 2025-01-20
Do not fail SUBSCRIBEs that start in the past

### LINKED ISSUE #595: Is a SUBSCRIBE_UPDATE with Start Group in the past an error?
It is for SUBSCRIBE. Maybe SUBSCRIBE_UPDATE shouldn't have a  start group at all?

ianswett:
Fixes #595 

It's difficult or possibly even impossible to prevent it.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1343,13 +1343,13 @@ the subscription starts with the first published or received group.
 AbsoluteStart (0x3):  Specifies an open-ended subscription beginning
 from the object identified in the StartGroup and StartObject fields. If the
 StartGroup is prior to the current group, the subscription starts at the
-beginning of the current object.
+beginning of the current object like the 'Latest Object' filter.
 
 AbsoluteRange (0x4):  Specifies a closed subscription starting at StartObject
 in StartGroup and ending at EndObject in EndGroup.  The start and end of the
 range are inclusive.  EndGroup MUST specify the same or a later group than
 StartGroup. If the StartGroup is prior to the current group, the subscription
-starts at the beginning of the current object.
+starts at the beginning of the current object like the 'Latest Object' filter.
 
 A filter type other than the above MUST be treated as error.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1343,13 +1343,13 @@ the subscription starts with the first published or received group.
 AbsoluteStart (0x3):  Specifies an open-ended subscription beginning
 from the object identified in the StartGroup and StartObject fields. If the
 StartGroup is prior to the current group, the subscription starts at the
-beginning of the current group.
+beginning of the current object.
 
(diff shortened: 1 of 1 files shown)


wilaw:
Contrary to the title, this PR is not allowing a SUBSCRIBE to start in the past. It is taking a SUBSCRIBE that wants to start in the past, and coercing it to instead start at the beginning of the current group.

ianswett:
> Contrary to the title, this PR is not allowing a SUBSCRIBE to start in the past. It is taking a SUBSCRIBE that wants to start in the past, and coercing it to instead start at the beginning of the current group?

Good point will, I retitled this PR.

draft-ietf-moq-transport.md
```diff
@@ -1248,14 +1248,14 @@ the subscription starts with the first published or received group.
 
 AbsoluteStart (0x3):  Specifies an open-ended subscription beginning
 from the object identified in the StartGroup and StartObject fields. If the
-StartGroup is prior to the current group, the publisher MUST reply with a
-SUBSCRIBE_ERROR with code 'Invalid Range'.
+StartGroup is prior to the current group, the subscription starts at the
+beginning of the current group.
```
afrind, line 1252: Individual Comment: 

I'd say if the specified start point is prior to the latest seen object, the subscription starts at the next object .

suhasHere: +1 with @afrind with minor tweaks .. "If the specified StartGroup is prior to the latest seen object, then the subscriptions starts at the latest group and object seen at the publisher"

draft-ietf-moq-transport.md
```diff
@@ -1248,14 +1248,14 @@ the subscription starts with the first published or received group.
...
 
 AbsoluteRange (0x4):  Specifies a closed subscription starting at StartObject
 in StartGroup and ending at EndObject in EndGroup.  The start and end of the
 range are inclusive.  EndGroup MUST specify the same or a later group than
-StartGroup. If the StartGroup is prior to the current group, the publisher MUST
-reply with a SUBSCRIBE_ERROR with code 'Invalid Range'.
+StartGroup. If the StartGroup is prior to the current group, the subscription
+starts at the beginning of the current group.
```
suhasHere, line 1258: same comment as above that we need to start at the current group/object at the publisher ( in a way it becomes the latest object filter)

suhasHere (commented):
few minor suggestions to clarify.

ianswett:
I made some minor updates, but I think a larger change is likely required to clarify what Objects one expects to receive.

draft-ietf-moq-transport.md
ianswett: I mention the Latest Object filter now.  I think it might be worth filing an issue for exactly what Object is delivered first for the Latest Object filter.

englishm (approved):
This seems like a useful incremental change to get to where we want to with clarifying Subscribe and Fetch. LGTM

PR #654, opened 2025-01-20
When to send MAX_SUBSCRIBE_ID

### LINKED ISSUE #572: When should MAX_SUBSCRIBE be sent ?

ianswett:
Adds some text and references the similar mechanism of MAX_STREAMS in RFC9000

Fixes #572

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1185,7 +1185,9 @@ the `query` portion of the URI to the parameter.
 
 #### MAX_SUBSCRIBE_ID {#max-subscribe-id}
 
-The MAX_SUBSCRIBE_ID parameter (Parameter Type 0x02) communicates an initial value for the Maximum Subscribe ID to the receiving subscriber. The default value is 0, so if not specified, the peer MUST NOT create subscriptions.
+The MAX_SUBSCRIBE_ID parameter (Parameter Type 0x02) communicates an initial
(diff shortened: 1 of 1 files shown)


afrind (approved):
Individual Comment: 

lgtm

draft-ietf-moq-transport.md
```diff
@@ -1185,7 +1185,9 @@ the `query` portion of the URI to the parameter.
 
 #### MAX_SUBSCRIBE_ID {#max-subscribe-id}
 
-The MAX_SUBSCRIBE_ID parameter (Parameter Type 0x02) communicates an initial value for the Maximum Subscribe ID to the receiving subscriber. The default value is 0, so if not specified, the peer MUST NOT create subscriptions.
+The MAX_SUBSCRIBE_ID parameter (Parameter Type 0x02) communicates an initial
+value for the Maximum Subscribe ID to the receiving subscriber. The default value
+is 0, so if not specified, the peer MUST NOT create subscriptions.
```
suhasHere, line 1190: i wonder if we should explicitly say this is sent in Server Setup message and is useful only of the Role is either subscriber or pub/sub

ianswett: I only reflowed this because it was a long line.

PR #656, opened 2025-01-21
Move error codes to message descriptions, include newer error messages

### LINKED ISSUE #577: Errors from Announce

martinduke:
The error codes are currently in the "Relays" section, which doesn't make any sense.

It's also missing codes for SUBSCRIBE_ANNOUNCES_ERROR, ANNOUNCE_ERROR, and FETCH_ERROR.

Also fixes #577.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -786,7 +786,8 @@ to the old relay can be stopped with an UNSUBSCRIBE.
 Publishing through the relay starts with publisher sending ANNOUNCE
 control message with a `Track Namespace` ({{model-track}}).
 The ANNOUNCE enables the relay to know which publisher to forward a
-SUBSCRIBE to if the track belongs to the namespace in the ANNOUNCE.
+SUBSCRIBE to, if the track name belongs to the namespace indicated in
+the ANNOUNCE.
 
 Relays MUST verify that publishers are authorized to publish
 the content associated with the set of
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -631,19 +631,11 @@ expects more OBJECTs to be delivered. The server closes the session with a
 Discovery of MoQT servers is always done out-of-band. Namespace discovery can be
 done in the context of an established MoQT session.
 
-Throughout this section, "publishers" refer to endpoints that can deliver
-objects in response to SUBSCRIBE or FETCH messages for a particular track, and
-"subscribers" refer to endpoints that might send SUBSCRIBE or FETCH messages for
-a track. An endpoint can be both a publisher and subscriber in a session, for
-different tracks. If a relay, it can be a publisher for a track in one session
-and a subscriber for it in another session. But an endpoint cannot be publisher
-and subscriber for the same track in the same session.
-
 Given sufficient out of band information, it is valid for a subscriber
 to send a SUBSCRIBE or FETCH message to a publisher (including a relay) without
 any previous MoQT messages besides SETUP. However, SUBSCRIBE_ANNOUNCES and
 ANNOUNCE messages provide an in-band means of discovery of subscribers and
-publishers for a namespace.
+publishers, respectively.
 
 The syntax of these messages is described in {{message}}.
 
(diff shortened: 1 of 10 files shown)


afrind (commented):
Thank you for doing this.  Do we need to look at Track Status responses too?

martinduke:
> Thank you for doing this. Do we need to look at Track Status responses too?

I don't want to take that on in this PR. These are mostly obvious tweaks from SUBSCRIBE_ERROR to *_ERROR.

draft-ietf-moq-transport.md
```diff
@@ -1545,6 +1503,22 @@ message for which this response is provided.
...
+|-----:|:--------------------------|
+| 0x0  | Internal Error            |
+|------|---------------------------|
+| 0x1  | Uninterested              |
+|------|---------------------------|
+| 0x2  | Unauthorized              |
+|------|---------------------------|
+| 0x3  | Timeout                   |
+|------|---------------------------|
+| 0x4  | Announce Not Supported    |
```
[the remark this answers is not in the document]
martinduke: OK, I made the mapping as uniform as possible.

It does presume that #643 goes through, so SUBSCRIBE_DONE no longer follows UNSUBSCRIBE

suhasHere (commented):
this seems ok , but rather duplicated ..

draft-ietf-moq-transport.md
```diff
@@ -2054,7 +2088,24 @@ failure.
...
+|------|---------------------------|
+| 0x1  | Unauthorized              |
+|------|---------------------------|
+| 0x2  | Timeout                   |
+|------|---------------------------|
+| 0x3  | Not Supported             |
+|------|---------------------------|
+| 0x4  | Namespace Does Not Exist  |
+|------|---------------------------|
+| 0x5  | Invalid Range             |
```
afrind, line 2107: How can you get Invalid Range in SUBSCRIBE_ANNOUNCES?

draft-ietf-moq-transport.md
```diff
@@ -2054,7 +2088,24 @@ failure.
...
+|-----:|:--------------------------|
+| 0x0  | Internal Error            |
+|------|---------------------------|
+| 0x1  | Unauthorized              |
+|------|---------------------------|
+| 0x2  | Timeout                   |
+|------|---------------------------|
+| 0x3  | Not Supported             |
+|------|---------------------------|
+| 0x4  | Namespace Does Not Exist  |
```
afrind, line 2105: There's not a namespace in this message, but a namespace prefix, and they don't have to "exist" per se.  Not sure exactly what to call this.

draft-ietf-moq-transport.md
```diff
@@ -1545,6 +1503,23 @@ message for which this response is provided.
...
+below:
+
+|------|---------------------------|
+| Code | Reason                    |
+|-----:|:--------------------------|
+| 0x0  | Internal Error            |
+|------|---------------------------|
+| 0x1  | Unauthorized              |
+|------|---------------------------|
+| 0x2  | Timeout                   |
```
afrind, line 1516: Are these timeout errors meant to be returned in the case where the receiver of the message times out waiting for eg: the upstream?  I mean, it never hurts to have a timeout error code.

afrind (commented):
Individual Comments

martinduke:
> this seems ok , but rather duplicated ..

Agreed, but each of the messages has a couple of slightly different error conditions, unfortunately.

draft-ietf-moq-transport.md
martinduke: I think it means "I will never ANNOUNCE within this prefix".  I'm open to other names.

draft-ietf-moq-transport.md
martinduke: I have no particular vision for these, but I didn't see a reason to take it out.

draft-ietf-moq-transport.md
martinduke: Namespace Prefix Unknown

draft-ietf-moq-transport.md
martinduke: offline: Alan relented

ianswett (approved):
LG once you update the name and fix the merge conflicts.

martinduke:
I guess this is done, although git, as usual, is being stupid. I don't think any errors crept in, but I'm not sure.

ianswett (approved):

fluffy (approved):
I assume this is 100% editorial but LGTM.

PR #657, opened 2025-01-30
Allow SUBSCRIBE to start at prior groups

ianswett:
This is an alternative to #638, Joining Fetch.

This also allows for subscribed Objects to be delivered over a single stream when behind the 'live edge'.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1322,8 +1322,13 @@ GOAWAY Message {
 A subscription causes the publisher to send newly published objects for a track.
 A subscriber MUST NOT make multiple active subscriptions for a track within a
 single session and publishers SHOULD treat this as a protocol violation.
-The only objects prior to the current object that can be requested are those
-in the current group.
+
+If a subscription has multiple Objects to deliver, either because the subscription
+begins prior to the current Object or because delivery of the subscription has
+been delayed because another subscription is higher priority, the publisher
+SHOULD send queued Objects on a single stream to allow the subscriber to
+use flow control to control the delivery rate.  This is particularly important
+at relays where the subscriber may not be the end subscriber.
 
 **Filter Types**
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1322,8 +1322,13 @@ GOAWAY Message {
 A subscription causes the publisher to send newly published objects for a track.
 A subscriber MUST NOT make multiple active subscriptions for a track within a
 single session and publishers SHOULD treat this as a protocol violation.
-The only objects prior to the current object that can be requested are those
-in the current group.
+
+If a subscription has multiple Objects to deliver, either because the subscription
```
martinduke, line 1326: Individual comment:

For objects in the past, this is just another way of saying "subscribe + fetch" and it's fine.  For queued objects that are post-subscribe, I don't like the uncertainty of how many streams things are going to arrive on.

Furthermore, FETCH and SUBSCRIBE have different rules for omitting objects, and what to do with non-Normal objects. This introduces a lot of ambiguity that requires careful thought.

ianswett: FETCH and SUBSCRIBE actually have quite similar rules for omitting objects.  The main difference is that SUBSCRIBE has more flexibility in the transmission order, since it can use more than one stream.

englishm: I agree with @martinduke here. I appreciate the sentiment that we could conceivably collapse subscribe and fetch into a singular message (again), but I think it would require quite a bit more articulation of the behavior than is currently written. I also agree that the difference in requirements for omitting objects is an important nuance we should not overlook because my current understanding is that Fetch is more likely to need to go upstream (potentially even back to the original publisher) to get a definitive answer about object non-existence. If we don't provide the same guarantee here, then we still need Fetch for that. If we do, we risk unduly increasing upstream requests for clients who may not need that same level of fidelity for the entire range of objects requested by a Subscribe.

ianswett: Fair points.  The Joining FETCH PR is quite far along, and as long as we restrict it a bit more (ie: You can't join a SUBSCRIBE that starts 100 groups in the future), I think it will accomplish what we want without too much complexity.

PR #665, opened 2025-02-04
Remove ContentExists references from FETCH_OK

### LINKED ISSUE #655: FETCH_OK doesn't contain a ContentExists field
Hi all! I noticed a small inconsistency while implementing FETCH message encoding/decoding. 

The description of FETCH_OK's "Largest Group ID" and "Largest Object ID" fields state "This field is only present if ContentExists has a value of 1.", but there is no ContentExists field in FETCH_OK, like there is in SUBSCRIBE_OK.

I think this just editorial and a straightforward fix so I've [prepped a diff](https://github.com/moq-wg/moq-transport/compare/main...mcintyrehh:moq-transport:fetch-ok-cleanup) and I'd be happy to submit a PR, but let me know if you think this needs discussion first. I might be missing some context.

mcintyrehh:
Resolves #655

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1782,11 +1782,10 @@ Values of 0x0 and those larger than 0x2 are a protocol error.
 the Largest Group ID and Object Id indicate the last Object in the track,
 0 if not.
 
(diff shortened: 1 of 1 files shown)


ianswett (approved):
Thanks!

PR #687, opened 2025-02-24
Clean up confusing all caps text that seems false 

fluffy:
add bit in terms defined after avoid confusion of if the "all capitals, as shown here" applies to the following text


suhasHere (approved):

ianswett (commented):
I think this goes back to the question about whether to consistently capitalize Object, Group, etc?

PR #707, opened 2025-02-24
Remove unused x(b) notation

LPardue:
This is leftover and unused after #550


LPardue:
Actually I lied, its used by `tuple`. But my recollection from previous discussion was that we wanted to also change the generic `tuple` into something object specific.

LPardue:
See https://github.com/moq-wg/moq-transport/pull/708 instead

PR #710, opened 2025-02-25
Replace use of tuple with concrete type

LPardue:
Strawman proposal to follow up on what was discussed in #536 

In the process, I spotted other uses of "tuple" that seemed like they could be substituted for other words (I dunno, you tell me). However, with the removal of the N-tuple thing, switching back to the old verbiage might be fine (you tell me).

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -263,21 +263,6 @@ x (L) ...:
 : Indicates that x is repeated zero or more times and that each instance
   has a length of L
 
-This document extends the RFC9000 syntax and with the additional field types:
-
-x (b):
-
-: Indicates that x consists of a variable length integer encoding as
-  described in ({{?RFC9000, Section 16}}), followed by that many bytes
-  of binary data
-
-x (tuple):
-
-: Indicates that x is a tuple, consisting of a variable length integer encoded
-  as described in ({{?RFC9000, Section 16}}), followed by that many variable
-  length tuple fields, each of which are encoded as (b) above.
-
-
 To reduce unnecessary use of bandwidth, variable length integers SHOULD
 be encoded using the least number of bytes possible to represent the
 required value.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -390,36 +375,79 @@ active.
...
-Track namespace is an ordered N-tuple of bytes where N can be between 1 and 32.
-The structured nature of Track Namespace allows relays and applications to
-manipulate prefixes of a namespace. Track name is a sequence of bytes.
-If an endpoint receives a Track Namespace tuple with an N of 0 or more
-than 32, it MUST close the session with a Protocol Violation.
+A Track Namespace Prefix is a list of byte sequence entries. It is used for
+matching different Track Namespaces based on a prefix. For example, if two are
+two Track Namespaces of ("example.com", "meeting=123", "participant=100") and
+("example.com", "meeting=123", "participant=200") the Track Namespace Prefix
+("example.com", "meeting=123")would match both.
```
afrind, line 413:
```suggestion
("example.com", "meeting=123") would match both.
```

draft-ietf-moq-transport.md
```diff
@@ -1659,16 +1687,16 @@ ANNOUNCE_ERROR Message
...
   Reason Phrase Length (i),
   Reason Phrase (..),
 }
 ~~~
 {: #moq-transport-announce-error format title="MOQT ANNOUNCE_ERROR Message"}
 
-* Track Namespace: Identifies the track namespace in the ANNOUNCE
-message for which this response is provided.
+* Track Namespace: Identifies the track namespace (as defined in
+{{track-name}})in the ANNOUNCE message for which this response is provided.
```
afrind, line 1699:
```suggestion
{{track-name}}) in the ANNOUNCE message for which this response is provided.
```

draft-ietf-moq-transport.md
```diff
@@ -2180,6 +2204,9 @@ TRACK_STATUS Message {
 ~~~
 {: #moq-track-status-format title="MOQT TRACK_STATUS Message"}
 
+* Track Namespace: Identifies a track's namespace as defined in
+({{track-name}})
+
```
afrind, line 2209: Add an entry for Track Name while you are here?

afrind (commented):
Individual review: 

This looks fine.  It's a little strange that the definition of namespace and namespace prefix is identical.  I wonder if we need both or if SUBSCRIBE_ANNOUNCES should just take a Track Namespace?

afrind:
Parked pending document reorg

afrind:
This is a good idea but has been made stale.  Let's close this and try again.

Issue #679, opened 2025-02-20
Subscribe Filters need revision (SUBSCRIBE_UPDATE fields as well)


martinduke:
As an Individual:

Now that we have Joining Fetch, some subscribe filters are redundant.

LatestGroup is equivalent to LatestObject + JoiningFetch with 0 previous groups.

Are there important use cases for AbsoluteStart, AbsoluteEnd, or AbsoluteRange, which would all need to be entirely in the future?

For transitioning from one ABR track to another, I can see a use for AbsoluteEnd in SUBSCRIBE_UPDATE and AbsoluteStart in SUBSCRIBE. If people agree that it's important, and there are not other use cases, I would propose the following:

SUBSCRIBE is LatestObject by default with an optional StartGroup which MUST be beyond the live edge.

SUBSCRIBE_UPDATE just has an EndGroup field.

martinduke:
I can also see a case for NextGroup, but have no strong opinion.

afrind:
Individual Comment:

> LatestGroup is equivalent to LatestObject + JoiningFetch with 0 previous groups.

+1 remove LatestGroup

> Are there important use cases for AbsoluteStart, AbsoluteEnd, or AbsoluteRange, which would all need to be entirely in the future?

Starting in draft-08, AbsoluteStart and AbsoluteRange can be in the past, and SUBSCRIBE_OK tells you where the response actually begins, or you can get INVALID_RANGE for AbsoluteRange *entirely* in the past.

afrind:
Individual Comment:

> what is the use case for it?

I think it's something like:

AbsoluteStart = "Start playing from Time=X"

AbsoluteRange = "Get a clip from Time=X to Time=Y"

Since the client may not know where X and Y are relative to the live head, I think it's easiest just to SUBSCRIBE with X and Y (translated to group numbers), and FETCH the remainder.  You cannot JOINING_FETCH pgo=N if you don't know the live head. 
 You can implement both use cases with TRACK_STATUS_REQUEST or starting with FETCH, but starting with SUBSCRIBE also ensures if you happen to cross live you won't miss something and have to re-FETCH.

martinduke:
Thanks for clarifying. I still have trouble imagining a client getting a fixed clip from beyond the live edge. How would an application even determine the correct time indices (to say nothing of group ID)?

[...]

Issue #468, opened 2024-06-17
enum types have inconsistent numbering


afrind:
Role, Subscribe Filter = start at 0x1
Setup Params, Object Status, Track Status codes, Subscribe Done, other error codes = start at 0x0

kixelated:
`ROLE` starts at 1 too

fluffy:
In general, I think it is really nice to have a value defined as Invalid and that will never be used by the enum or extension. That value is often zero. Agree we should be consistent and don't really care much one way or another. (and if you receive an enum with this value, it is a protcol error ).

afrind:
Individual Comment:

I'll note that Role has a reason to start at 1 -- Pub (0x01) | Sub (0x02) = PubSub (0x03).

afrind:
Stream Types are currently 

DATAGRAM=1, TRACK=2, SUBGROUP=4

ianswett:
The plan is to park this for now, and then renumber everything towards the end.

afrind:
Tracking in #881

PR #713, opened 2025-02-25
Remove Latest Group from SUBSCRIBE

### LINKED ISSUE #598: SUBSCRIBE(current group) splits past and future

ianswett:
Fixes #598 

May address #679 

Now that Joining Fetch is in the draft, we don't need this.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1288,11 +1288,7 @@ in the current group.
 The subscriber specifies a filter on the subscription to allow
 the publisher to identify which objects need to be delivered.
 
-There are 4 types of filters:
-
-Latest Group (0x1) : Specifies an open-ended subscription with objects
-from the beginning of the current group.  If no content has been delivered yet,
-the subscription starts with the first published or received group.
+There are 3 types of filters:
 
 Latest Object (0x2): Specifies an open-ended subscription beginning from
 the current object of the current group.  If no content has been delivered yet,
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1288,11 +1288,7 @@ in the current group.
 The subscriber specifies a filter on the subscription to allow
 the publisher to identify which objects need to be delivered.
 
-There are 4 types of filters:
-
-Latest Group (0x1) : Specifies an open-ended subscription with objects
-from the beginning of the current group.  If no content has been delivered yet,
-the subscription starts with the first published or received group.
+There are 3 types of filters:
```
suhasHere, line 1291: should we change the numbering to start from 0x1 ?

ianswett: I can go either way on this.  Also, we could renumber to start at 0x0?

afrind: See also #468.  I don't mind either way now, since we'll make a clean-up pass at some point.

afrind (commented):
Individual Review:

Lgtm

fluffy (approved):
LGTM

PR #723, opened 2025-02-25
Count object extension headers in bytes, not fields

### LINKED ISSUE #718: Change from extensions count number to length of all extensions
We agreed interim to Change from extensions count number to length of all extensions 

Victor doing PR

vasilvv:
This makes it possible to skip those without parsing, and easier to parse without trial decoding.

Fixes #718

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2350,8 +2350,8 @@ Track or Datagram.
 * Object Status: As enumeration used to indicate missing
 objects or mark the end of a group or track. See {{object-status}} below.
 
-* Object Extension Count: The number of Object Extensions present. A value of 0
-  indicates that no Object Extension Headers are present.
+* Object Extension Size: The total size of the Object Extension Headers block,
+  in bytes.
 
 * Object Extensions : A sequence of Object Extension Headers. See
   {{object-extensions}} below.
@@ -2455,7 +2455,7 @@ OBJECT_DATAGRAM {
   Group ID (i),
   Object ID (i),
   Publisher Priority (8),
-  Extension Count (i),
+  Extension Headers Size (i),
   [Extension headers (...)],
   Object Payload Length (i),
   [Object Status (i)],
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2350,8 +2350,8 @@ Track or Datagram.
 * Object Status: As enumeration used to indicate missing
 objects or mark the end of a group or track. See {{object-status}} below.
 
-* Object Extension Count: The number of Object Extensions present. A value of 0
-  indicates that no Object Extension Headers are present.
+* Object Extension Size: The total size of the Object Extension Headers block,
+  in bytes.
```
suhasHere, line 2354: we need to update Canonical Field section as well.

draft-ietf-moq-transport.md
```diff
@@ -2680,7 +2680,7 @@ STREAM_HEADER_GROUP {
 }
 {
   Object ID = 0
-  Extension Count  = 2
+  Extension Headers Size = 33
```
suhasHere, line 2683: :) did you count the bytes to be 33 :-)

draft-ietf-moq-transport.md
vasilvv: This is the Canonical Object Fields section, though?

draft-ietf-moq-transport.md
vasilvv: Yes, but I did it in a calculator, so it may be not 33 (and of course, varint serialization is ambiguous).

draft-ietf-moq-transport.md
```diff
@@ -2455,7 +2455,7 @@ OBJECT_DATAGRAM {
   Group ID (i),
   Object ID (i),
   Publisher Priority (8),
-  Extension Count (i),
+  Extension Headers Size (i),
```
ianswett, line 2458: I'd prefer length over size, since that's what we use elsewhere in the doc.

ianswett (approved):

afrind (commented):
Individual Review:

lgtm.

Do we need to specify an error when the length of the extensions block exceeds the Extensions Length?

suhasHere (approved):

fluffy (approved):
LGTM

Issue #592, opened 2024-10-22
Clarification on what current object means in SUBSCRIBE, SUBSCRIBE_OK and FETCH


wilaw:
Given the sequence below, of Groups each containing 4 objects, with the request  time of subscription shown midway through receipt of object 1 of group 2. The relay has received the first bytes of Object 1-2, but not the complete object. 

![Screenshot 2024-10-22 at 10 19 27](https://github.com/user-attachments/assets/3e363843-e4df-4884-9e3c-9aa455ca0750)

1. Current SUSBCRIBE text indicates

> Latest Group (0x1) : Specifies an open-ended subscription with objects from the beginning of the current group. If no content has been delivered yet, the subscription starts with the first published or received group.
> 
> Latest Object (0x2): Specifies an open-ended subscription beginning from the current object of the current group. If no content has been delivered yet, the subscription starts with the first published or received group.

Does a SUBSCRIBE of Latest Object (0x2) begin with 
a) 2-0 because it the latest complete object at the relay
b) 2-1 because some of the bytes have been received
c) 2-2 because we don't want to start a subscription on a partially received object

2. FETCH text indicates

> EndObject: The end Object ID, plus 1. A value of 0 means the entire group is requested.

What is the largest EndObject object a player can issue a FETCH for?  Is it
a) Object 2-0 because that is the last fully received object?
b) Object 2-1 because some bytes have been received?

3. SUBSCRIBE_OK text indicates

> Largest Object ID: The largest Object ID available within the largest Group ID for this track. This field is only present if ContentExists has a value of 1.

What largest Object ID would be returned? Is it 
a) 2-0 because it the latest complete object at the relay
b) 2-1 because some of the bytes have been received

afrind:
Speaking as someone with a non-caching relay implementation:

SUBSCRIBE_OK will return 2-1
SUBSCRIBE (Latest obj) begins with 2-2

I haven't implemented FETCH yet, but I think FETCH can set the end object to 2^62-1.  The question is, what comes in the FETCH OK?  Since I don't have a cache, I can just forward upstream (side question: should the relay rewrite the upstream FETCH end object to 2-1?).  If I had a cache, I'm not sure.  I want to set the end object in FETCH OK to 2-1, but because it would be incomplete in cache, it's probably easier to return 2-0, but that might create a gap that needs refetching.

ianswett:
[...]

PR #726, opened 2025-02-26
Replace latest Group/Object with StartGroup and StartObject

ianswett:
Also clarify what Object is expected to be delivered first, fixing #592

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1296,7 +1296,7 @@ the subscription starts with the first published or received group.
 
 Latest Object (0x2): Specifies an open-ended subscription, beginning
 with the next published Object.  If no content has been delivered
-yet, the subscription starts with the first published or received group.
+yet, the subscription starts with the first published Object.
 
 AbsoluteStart (0x3):  Specifies an open-ended subscription beginning
 from the object identified in the StartGroup and StartObject fields. If the
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1835,7 +1835,7 @@ Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
 Values of 0x0 and those larger than 0x2 are a protocol error.
 
 * StartGroup: The resolved start Group of the subscription filter.
-This SHOULD be the same as the largest Group ID.
+This SHOULD be the same as the largest observed Group ID.
 
 * StartObject: The resolved start Object of the subscription filter.
 This SHOULD be one more than the the largest Object ID available within the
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1294,9 +1294,9 @@ Latest Group (0x1) : Specifies an open-ended subscription with objects
 from the beginning of the current group.  If no content has been delivered yet,
 the subscription starts with the first published or received group.
 
-Latest Object (0x2): Specifies an open-ended subscription beginning from
-the current object of the current group.  If no content has been delivered yet,
-the subscription starts with the first published or received group.
+Latest Object (0x2): Specifies an open-ended subscription, beginning
+with the next published Object.  If no content has been delivered
+yet, the subscription starts with the first published or received group.
```
ianswett, line 1299:
```suggestion
yet, the subscription starts with the first published Object.
```

draft-ietf-moq-transport.md
```diff
@@ -1294,9 +1294,9 @@ Latest Group (0x1) : Specifies an open-ended subscription with objects
 from the beginning of the current group.  If no content has been delivered yet,
 the subscription starts with the first published or received group.
 
-Latest Object (0x2): Specifies an open-ended subscription beginning from
-the current object of the current group.  If no content has been delivered yet,
-the subscription starts with the first published or received group.
+Latest Object (0x2): Specifies an open-ended subscription, beginning
+with the next published Object.  If no content has been delivered
```
afrind, line 1298: The next published object may not be larger than Start Group/Start Object.  eg: with reordering the relay has seen 2-2 when my subscribe arrives, it will say '2-3', but the next object the relay sees may be 2-1.

My current relay implementation will not deliver 2-1, so I think I agree with what's below, but maybe just this wording needs something else.

Also, given the name in the response "Start Group/Object", we might want to clearly explain there is no guarantee that such object will ever be delivered or even exists.  It's just the equivalent to the field with the same name in AbsoluteStart/Range.

draft-ietf-moq-transport.md
```diff
@@ -1815,9 +1815,8 @@ SUBSCRIBE_OK
   Subscribe ID (i),
   Expires (i),
   Group Order (8),
-  ContentExists (8),
-  [Largest Group ID (i)],
-  [Largest Object ID (i)],
+  [StartGroup (i),
+   StartObject (i)],
```
afrind, line 1819: These are in brackets - when does the receiver know they are there?  Or should we always include them?  For AbsoluteStart/AbsoluteRange, we just echo back what was in the request?

afrind (commented):
Individual Review

draft-ietf-moq-transport.md
suhasHere: I wonder if we should instead call it as "LastSeenGroup" and "LastSeenObject"  .. this is the statement of truth when the subscription happened. End Subscribers now know the last known point and adjust accordingly.

ianswett:
I believe @martinduke was going to take a crack at this issue as well.  I'm not sure I'm completely happy with this text, so I'm going to mark it parked for now.

PR #727, opened 2025-02-26
Add language to define how Group IDs increment

wilaw:


draft-ietf-moq-transport.md
```diff
@@ -379,6 +379,11 @@ subscriber that does not want to receive the entire track can opt to receive onl
 the latest group(s).  The publisher then selectively transmits objects based on
 their group membership.  Groups can contain any number of objects.
 
+Groups are assigned integer Group IDs. Within a track, the original publisher
+MUST produce Group IDs which increase with time. Note that Groups delivered over
+different Subgroups or Datagrams might be received by the subscriber out of numeric
+Group ID order.
```
suhasHere, line 385: I wonder how do we support the demo use-case on synchronized video fit with the above model.

draft-ietf-moq-transport.md
```diff
@@ -379,6 +379,11 @@ subscriber that does not want to receive the entire track can opt to receive onl
...
+
```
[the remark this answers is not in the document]
wilaw: That can be accomplished by the synchronized clients all subscribing to a control track produced by the leader. That control track instructs them to play a VOD asset (play, pause, seek) in sync with the leader. The VOD asset group IDs run from {0,1,2,…,N}.

That demo could also show a live stream control, with (play,plause) controls if the relay has no caching, or (play,pause,seek) if the relay has caching. In this case the live streams Group IDs also run {0,1,2,…,N}.

draft-ietf-moq-transport.md
```diff
@@ -379,6 +379,11 @@ subscriber that does not want to receive the entire track can opt to receive onl
 the latest group(s).  The publisher then selectively transmits objects based on
 their group membership.  Groups can contain any number of objects.
 
+Groups are assigned integer Group IDs. Within a track, the original publisher
+MUST produce Group IDs which increase with time. Note that Groups delivered over
```
gwendalsimon, line 383: Thanks for the re-wording.
Though, we are back to the point that it says nothing about the +1 issue (incremental increase of group_ID). 

Problem: if a Publisher publishes a track where group_ID contains only even incremental numbers. When a Relay caches the objects, it cannot know whether the missing odd group_ID are gaps or not. When receiving a FETCH(0,1000) request, it can serve the 500 objects from its cache but it generates 500 individual FETCH requests to the Publisher to check the 500 gaps. It may be exploited for DDoS attacks.

To reach a potential consensus, it was mentioned during Denver's interim meeting that we could add a track property which states whether the original publisher produces Group ID with gaps or not. I would be in favor of having such a track property.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
suhasHere: yes .. we can do it in different ways .. but what we saw yesterday was things work pretty well the way the protocol is defined today .. It is surprising to instruct app developers to change their implementation than innovate !!!

draft-ietf-moq-transport.md
suhasHere: @gwendalsimon  for my understanding, how will knowing that a group can have gaps help in relay implementation ?

zafergurel: To detect a gap, previous group id can be added to the group header. So, you won't need incrementing group ids or it's not necessary to increment by one.
1 <- 3 <- 5 is a perfectly acceptable scheme. Even if you don't have 2 and 4 in the cache, the relay can understand there is no gap in this sequence by looking at every previous group id.

Issue #688, opened 2025-02-24
Track Defn seems wrong.


fluffy:
In section 1.2, the definition for Track should remove the part that says "An encoded bitstream" - this does not make sense in this context and could mean anything. The resit of the description is fine for the definition without this.

PR #736, opened 2025-02-27
Streamline Track definition in Terminology section

afrind:
Fixes: #688

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -229,9 +229,7 @@ Object:
 
 Track:
 
-: An encoded bitstream. Tracks contain a sequential series of one or
-  more groups and are the subscribable entity with MOQT.
-  See ({{model-track}}).
+: A track is a collection of one or more groups. See ({{model-track}}).
 
 
 ## Notational Conventions
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -229,7 +229,7 @@ Object:
 
 Track:
 
-: A track is a collection of one or more groups. See ({{model-track}}).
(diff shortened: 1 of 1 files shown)


ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -229,9 +229,7 @@ Object:
 
 Track:
 
-: An encoded bitstream. Tracks contain a sequential series of one or
-  more groups and are the subscribable entity with MOQT.
-  See ({{model-track}}).
+: A track is a collection of one or more groups. See ({{model-track}}).
```
ianswett, line 232: We can keep this as is for now, but I imagine there are cases a track will have 0 groups?

If so, can we just say:
```suggestion
: A track is a collection of groups. See ({{model-track}}).
```

ianswett (approved):

draft-ietf-moq-transport.md
ianswett: Also, would it be better to use 'sequence' rather than collection?

suhasHere: Since we recently decided to with non sequential groups ..  I wonder if we use the word sequence , it may add the confusion ???

suhasHere (commented):
One meta comment:  I think for each of Track, Group and Object, we should just point to the model section and not define anything in the terminology section.

draft-ietf-moq-transport.md
ianswett: SG, let's keep this wording for now, but I think 0 has to be allowable, so I'd prefer less text.

I do think it's nice to have a definition, just a very short one.

PR #749, opened 2025-03-01
Group ID SHOULD increase with time

ianswett:
The text that came out of the Denver interim.

Closes #727
Closes #587

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -379,6 +379,12 @@ subscriber that does not want to receive the entire track can opt to receive onl
 the latest group(s).  The publisher then selectively transmits objects based on
 their group membership.  Groups can contain any number of objects.
 
+Within a track, the original publisher SHOULD produce Group IDs which increase
+with time. Subscribers to tracks which do not follow this requirement SHOULD NOT
+use range filters which span multiple groups in FETCH or SUBSCRIBE. SUBSCRIBE and
+FETCH delivery use Group Order, so a FETCH cannot deliver Groups out of order
+and subscription could have unexpected delivery order if Group IDs do not increase
+with time.
 
 ## Track {#model-track}
 


draft-ietf-moq-transport.md
```diff
@@ -379,6 +379,12 @@ subscriber that does not want to receive the entire track can opt to receive onl
 the latest group(s).  The publisher then selectively transmits objects based on
 their group membership.  Groups can contain any number of objects.
 
+Within a track, the original publisher SHOULD produce Group IDs which increase
+with time. Subscribers to tracks which do not follow this requirement SHOULD NOT
+use range filters which span multiple groups in FETCH or SUBSCRIBE. SUBSCRIBE and
+FETCH delivery use Group Order, so a FETCH cannot deliver Groups out of order
+and subscription could have unexpected delivery order if Group IDs do not increase
```
ianswett, line 386:
```suggestion
and a subscription could have unexpected delivery order if Group IDs do not increase
```

draft-ietf-moq-transport.md
```diff
@@ -379,6 +379,12 @@ subscriber that does not want to receive the entire track can opt to receive onl
 the latest group(s).  The publisher then selectively transmits objects based on
 their group membership.  Groups can contain any number of objects.
 
+Within a track, the original publisher SHOULD produce Group IDs which increase
+with time. Subscribers to tracks which do not follow this requirement SHOULD NOT
+use range filters which span multiple groups in FETCH or SUBSCRIBE. SUBSCRIBE and
+FETCH delivery use Group Order, so a FETCH cannot deliver Groups out of order
+and a subscription could have unexpected delivery order if Group IDs do not increase
+with time.
```
suhasHere, line 387: this looks good ..  One observation: At a time a fetch is answered from cache, the group order is applied to the data in the cache. Wouldn't fetch always get the things in the order for a given fetch request since it is serving data from cache ?

ianswett: Very true, but if you tried to Fetch everything, and Groups are published out of order, a Group with a smaller ID might be published after you're past that Group ID.

Happy to rephrase if you have a different suggestion?

suhasHere: good point. Thinking about it again,  I feel what you have makes sense and we should keep it.

suhasHere (approved):

draft-ietf-moq-transport.md
ianswett: Thanks Suhas.  I think more text might be appropriate, but for today, I'll focus on shipping a draft.

Issue #680, opened 2025-02-20
Collapse END_OF_TRACK and END_OF_TRACK_AND_GROUP


afrind:
I know I was the one who suggested this but now that I've implemented I think the two are equivalent.

If you send END_OF_TRACK with a non-zero object id, it means "END_OF_TRACK_AND_GROUP" for the group it is in.  If you send it with object ID 0, it means the track ended.

ianswett:
SGTM, can you write a PR?

PR #755, opened 2025-03-03
Merge END_OF_TRACK with END_OF_TRACK_AND_GROUP

afrind:
Fixes: #680

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2388,7 +2388,7 @@ are beyond the end of a group or track.
          in this track and the ObjectID is one greater than the largest object
          produced in that group, or GroupID is one greater than the largest
          group produced in this track and the ObjectID is zero. This status
-         implies the end of the specified group. An object with this status
+         also indicates the last group has ended. An object with this status
          that has a Group ID less than any other GroupID, or an ObjectID less
          than or equal to the largest in the specified group, is a protocol
          error, and the receiver MUST terminate the session. This SHOULD be
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2396,18 +2396,15 @@ are beyond the end of a group or track.
          in this Group. This SHOULD be cached. A publisher MAY use an end of
          Group object to signal the end of all open Subgroups in a Group.
 
-* 0x4 := Indicates end of Track and Group. GroupID is the largest group produced
-         in this track and the ObjectId is one greater than the largest object
-         produced in that group. An object with this status that has a Group ID
-         less than any other Group ID, or an Object ID less than or equal to the
-         largest in the group, is a protocol error, and the receiver MUST
-         terminate the session. This SHOULD be cached.
-
-* 0x5 := Indicates end of Track. GroupID is one greater than the largest group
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2396,18 +2396,15 @@ are beyond the end of a group or track.
...
-* 0x5 := Indicates end of Track. GroupID is one greater than the largest group
-         produced in this track and the ObjectId is zero. An object with this
-         status that has a Group ID less than or equal to any other Group ID, or
-         an Object ID other than zero, is a protocol error, and the receiver
-         MUST terminate the session. This SHOULD be cached.
+* 0x4 := Indicates end of Track. GroupID is either the largest group produced
+         in this track and the ObjectID is one greater than the largest object
+         produced in that group, or GroupID is one greater than the largest
+         group produced in this track and the ObjectID is zero. This status
+         implies the end of the specified group. An object with this status
```
suhasHere, line 2403: s/This status/ The latter status/

suhasHere (commented):
Looks good . one minor comment

draft-ietf-moq-transport.md
```diff
@@ -2396,18 +2396,15 @@ are beyond the end of a group or track.
...
+         that has a Group ID less than any other GroupID, or an ObjectID less
+         than or equal to the largest in the specified group, is a protocol
+         error, and the receiver MUST terminate the session. This SHOULD be
```
ianswett, line 2406: Should we say something about how this shouldn't be propagated either?  Otherwise a publisher might be able to close every downstream subscriber's session.

ianswett (approved):
One potential 'attack' concern, but LG

draft-ietf-moq-transport.md
afrind: I guess a note to that effect might be helpful.

ianswett: I filed #775 as a general issue.  It's possible we should say that once in the doc and then we don't need to call it out for each individual message?

draft-ietf-moq-transport.md
```diff
@@ -2386,18 +2386,15 @@ are beyond the end of a group or track.
...
-* 0x5 := Indicates end of Track. GroupID is one greater than the largest group
-         produced in this track and the ObjectId is zero. An object with this
-         status that has a Group ID less than or equal to any other Group ID, or
-         an Object ID other than zero, is a protocol error, and the receiver
-         MUST terminate the session. This SHOULD be cached.
+* 0x4 := Indicates end of Track. GroupID is either the largest group produced
+         in this track and the ObjectID is one greater than the largest object
+         produced in that group, or GroupID is one greater than the largest
+         group produced in this track and the ObjectID is zero. This status
+         implies the end of the specified group. An object with this status
```
ianswett, line 2393:
```suggestion
         also indicates the last group has ended. An object with this status
```

draft-ietf-moq-transport.md
afrind: #806 addresses this

draft-ietf-moq-transport.md
afrind: There's only one status now, so I think "This status" is correct.

Issue #722, opened 2025-02-25
Need to be able to provide multiple auth tokens in one request


fluffy:

wilaw:
Since auth tokens are provided by the 0x02 parameter, we could easily provide multiple tokens if we simply attached multiple instances of that 0x02 parameter. 

However the text currently states that  `Senders MUST NOT repeat the same parameter type in a message.`.

We should adjust that restriction to not apply to AUTHORIZATION INFO or we can remove that blanket restriction completely and then allow the definition of each parameter to specify whether duplicate parameter entries are allowed in a particular message.

PR #756, opened 2025-03-03
Adds AUTHORIZATION ALIAS parameter

wilaw:
The AUTHORIZATION ALIAS parameter allows a small alias to be associated with a larger AUTHORIZATION INFO payload, such as a token. This reduces message wireline size and token processing time (decryption etc) when future operations (such as rapid subscriptions and fetches) wish to reapply the same token. 

Also removes the restriction that AUTHORIZATION INFO is an ASCII string.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1175,7 +1175,7 @@ a relay that handles a subscription that includes those Objects re-requests them
 AUTHORIZATION ALIAS parameter (Parameter Type 0x05) identifies an alias by which
 AUTHORIZATION INFO (Parameter Type 0x02) can be referenced within the current
 session. Transmission of both AUTHORIZATION ALIAS and AUTHORIZATION INFO parameters
-in the same parameter set is an instruction to the receiver to associate the 
+in the same parameter set is an instruction to the receiver to associate the
 AUTHORIZATION ALIAS with the AUTHORIZATION INFO. Receivers processing future
 parameters sets in which only the AUTHORIZATION ALIAS is sent MUST process the
 parameter set as if the AUTHORIZATION INFO had been retransmitted.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1170,6 +1170,25 @@ means Objects earlier in a multi-object stream will expire earlier than Objects
...
+AUTHORIZATION INFO (Parameter Type 0x02) can be referenced within the current
+session. Transmission of both AUTHORIZATION ALIAS and AUTHORIZATION INFO parameters
+in the same parameter set is an instruction to the receiver to associate the
+AUTHORIZATION ALIAS with the AUTHORIZATION INFO. Receivers processing future
+parameters sets in which only the AUTHORIZATION ALIAS is sent MUST process the
+parameter set as if the AUTHORIZATION INFO had been retransmitted.
+
+If an AUTHORIZATION ALIAS is received without first having been associated with an
+AUTHORIZATION INFO in that session, then the receiver SHALL process the parameter
+set as if no AUTHORIZATION INFO had been included.
```
suhasHere, line 1185: I think we Should go stronger her and say terminate the connection.

afrind:
Individual Comment: 

Would it be better to define a structure for Auth Info that has an optional alias?  If we allow sending > 1 AUTH (#722), then there's no safe way to correlate them.

wilaw:
> Would it be better to define a structure for Auth Info that has an optional alias? If we allow sending > 1 AUTH (https://github.com/moq-wg/moq-transport/issues/722), then there's no safe way to correlate them.

Since parameters have a serialization order, one alternative option is to specify that an AUTHORIZATION ALIAS refers to the AUTHORIZATION INFO which it follows. This would allows multiple auth blobs and aliases to be communicated in one message.

draft-ietf-moq-transport.md
suhasHere: Typically In security , it is recommended to be clear with the intent.  
No auth parameters implies , there is nothing to authroize against. 
Having authz parameters implies there is something we need to check before allowing the action. 
But this one is in-between and we should disallow it.

suhasHere:
+1 @afrind  Having a structure seems like a nice idea

wilaw:
Closing and replacing with #760 due to rebasing issues.

Issue #724, opened 2025-02-25
Do we need a way knowing what type of token each token is


fluffy:

ianswett:
To clarify, is this related to auth?

wilaw:
I think for now (v1 of MOQT) we should not create a registry and identity scheme for token types. The reasons are:
1. There is a high cardinality of token schemes that can be used now and going forward with moqt
2. These schemes mutate and evolve with their own versioning
3. Tracking and managing all these permutations is unnecessary overhead for MOQT.
4. In the HTTP world, there is equal diversity in possible tokens and there is no centralized registry. 

In HTTP tokens work at scale through business relationships between token issuers and CDNs. I think in MOQT, as long as we do not mandate any type of particular type of token format, that we can rely upon similar out-of-band business rules to ensure that the token processor understands the token type. If we find its a mess, then we can introduce a registry in v2.

afrind:
Individual Comment:

As someone with a simple relay and simple publishers, I'm not sure what I should do.  Let's say I build support for MoQ/CAT tokens - do I just decode any AUTH INFO as such, and tell all my clients they better use a token of that type?

Now if I add support for some other scheme, privacypass or basic auth, and I want to allow either, how do I know what the peer is sending me in the blob?  Or I can support at most one auth scheme per track?

afrind:
#760 now includes a field to indicate the token type, with 0 meaning "unspecified".  This works for me.

Issue #513, opened 2024-08-29
Auth Cookies Not Specified


fluffy:
This is a placeholder that we need to sort out how Auth and auth tokens work in MoQ. 

Requirements:

1. Can control what users can connect to a relay. 
2. Can control if a user can publish to a given namespace.
3. Can control if a user can subscribe to a given namespace. 
4. Can update auth for long lived track that is longer than token lifetime
5. Design that allows for relay to validate tokens locally 
6. As much as possible, reuse approaches used for existing systems 

What am I missing on requirements

wilaw:
For point [6], we might consider support for CTA 5007 - Common Access Token (CAT), a standard (soon) published by the CTA. This is a CBOR token for compactness,  with many claim fields as well a boolean operations for composite claims. It is likely to become a popular standard for OTT content, so harmonizing MoQT with that side of the industry brings some efficiency gains. 


For additional requirements, support both symmetric and asymmetric ciphers.

gwendalsimon:
> 1. Can control what users can connect to a relay.
> 2. Can control if a user can publish to a given namespace.
> 3. Can control if a user can subscribe to a given namespace.
It can be made even more specific to track name level. For example no access to UHD video track for some users. Or watch the video but no access to some associated chatrooms or premium metadata tracks.

> 4. Can update auth for long lived track that is longer than token lifetime
> 5. Design that allows for relay to validate tokens locally
Relay may also generate tokens locally. It is especially useful for token renewal (or update).

> 6. As much as possible, reuse approaches used for existing systems
+1 on Common Access Token (CAT).

> What am I missing on requirements
Note sure if it would require a separate Issue, but the question of watermark can also be raised. Today's OTT watermarking solutions are based on some WM tokens, which are some forms of auth tokens.

martinduke:
Parked for next hybrid Interim.

PR #760, opened 2025-03-04
Adding AUTHORIZATION TOKEN parameter

wilaw:
The AUTHORIZATION ALIAS parameter allows a small alias to be associated with a larger AUTHORIZATION INFO payload, such as a token. This reduces message wireline size and token processing time (decryption etc) when future operations (such as rapid subscriptions and fetches) wish to reapply the same token.

Also removes the restriction that AUTHORIZATION INFO is an ASCII string.

This PR replaces PR #756 (same changes, cleaner base)

Fixes https://github.com/moq-wg/moq-transport/issues/906
Fixes: #513 
Fixes: #722
Fixes: #724

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1120,8 +1120,8 @@ Protocol Violation.
 Some messages include a Parameters field that encode optional message
 elements. They contain a type, length, and value.
 
-Senders MUST NOT repeat the same parameter type in a message unless the 
-parameter definition explicitly allows multiple instances of that type to 
+Senders MUST NOT repeat the same parameter type in a message unless the
+parameter definition explicitly allows multiple instances of that type to
 be sent in a single message. Receivers SHOULD check that there are no
 unauthorized duplicate parameters and close the session as a
 'Protocol Violation' if found.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -633,6 +633,8 @@ code, as defined below:
 |------|---------------------------|
 | 0x12 | Data Stream Timeout       |
 |------|---------------------------|
+| 0x13 | Max Auth Limit Exceeded   |
+|------|---------------------------|
 
 * No Error: The session is being terminated without an error.
 
@@ -668,6 +670,9 @@ code, as defined below:
   stream. If an endpoint times out waiting for a new object header on an
   open subgroup stream, it MAY send a STOP_SENDING on that stream or
   terminate the subscription.
+
+* Max Auth Limit Exceeded - the Session limit {{max-auth-size}} of all registered
+  Authorization tokens has been exceeded.
 
 An endpoint MAY choose to treat a subscription or request specific error as a
 session error under certain circumstances, closing the entire session in
@@ -1261,6 +1266,10 @@ serialization, referencing an alias which has not been registered, attempting to
 register an alias which has been previously registered, or providing a token value
 which does not match the declared Token type.
 
+Retiring an alais and token that was previously used to authorize a message has no
+retroactive effect on the original authorization, nor does it prevent that same token
+being re-registered as long as that registration is performed with a new alias.
+
 Clients SHOULD only register tokens which they intend to re-use during the session.
 Client SHOULD retire previously registered tokens once their utility has passed.
 
@@ -1268,7 +1277,7 @@ By registering a Token, the client is requiring the receiver to store the Token 
 and Token Value until they are retired, or the Session ends. The receiver can protect
 its resources by sending a SETUP parameter defining the MAX_AUTH_SIZE {{max-auth-size}}
 limit it is willing to accept. If a registration is attempted which would cause this
-limit to be exceeded, the receiver MUST reject that message with a
+limit to be exceeded, the receiver MUST termiate the Session with a
 "Max Auth Limit Exceeded" error.
 
 
@@ -1724,8 +1733,6 @@ as defined below:
 |------|---------------------------|
 | 0x7  | Invalid Auth Token        |
 |------|---------------------------|
-| 0x8  | Max Auth Limit Exceeded   |
-|------|---------------------------|
 
 * Internal Error - An implementation specific or generic error occurred.
 
@@ -1750,10 +1757,6 @@ as defined below:
   be due to invalid serialization, referencing an alias which has not been
   registered, or attempting to register an alias which has been previously
   registered.
-
-* Max Auth Limit Exceeded - the size of the Authorization token supplied
-  causes the Session limit of all registered Authorization tokens to be
-
 
 ## SUBSCRIBE_UPDATE {#message-subscribe-update}
 
@@ -2212,9 +2215,7 @@ as defined below:
 |------|---------------------------|
 | 0x7  | Invalid Auth Token        |
 |------|---------------------------|
-| 0x8  | Max Auth Limit Exceeded   |
-|------|---------------------------|
-| 0x9  | Invalid Subscribe ID      |
+| 0x8  | Invalid Subscribe ID      |
 |------|---------------------------|
 
 * Internal Error - An implementation specific or generic error occurred.
@@ -2236,10 +2237,6 @@ as defined below:
   and object for the track, or the track has not published any objects.
 
 * Invalid Auth Token - An invalid Authorization token was provided.
-
-* Max Auth Limit Exceeded - the size of the Authorization token supplied
-  causes the Session limit of all registered Authorization tokens to be
-  exceeded.
 
 * Invalid Subscribe ID - The joining Fetch referenced a Subscribe ID that did
   not belong to an active Subscription.
@@ -2429,8 +2426,6 @@ below:
 |------|---------------------------|
 | 0x7  | Invalid Auth Token        |
 |------|---------------------------|
-| 0x8  | Max Auth Limit Exceeded   |
-|------|---------------------------|
 
 * Internal Error - An implementation specific or generic error occurred.
 
@@ -2445,10 +2440,6 @@ below:
 * Uninterested - The namespace is not of interest to the endpoint.
 
 * Invalid Auth Token - An invalid Authorization token was provided.
-
-* Max Auth Limit Exceeded - the size of the Authorization token supplied
-  causes the Session limit of all registered Authorization tokens to be
-  exceeded.
 
 
 ## UNANNOUNCE {#message-unannounce}
@@ -2613,8 +2604,6 @@ as defined below:
 |------|---------------------------|
 | 0x7  | Invalid Auth Token        |
 |------|---------------------------|
-| 0x8  | Max Auth Limit Exceeded   |
-|------|---------------------------|
 
 * Internal Error - An implementation specific or generic error occurred.
 
@@ -2633,12 +2622,6 @@ as defined below:
   SUBSCRIBE_ANNOUNCES in the same session.
 
 * Invalid Auth Token - An invalid Authorization token was provided.
-
-* Max Auth Limit Exceeded - the size of the Authorization token supplied
-  causes the Session limit of all registered Authorization tokens to be
-  exceeded.
-
-
 
 ## UNSUBSCRIBE_ANNOUNCES {#message-unsub-ann}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1258,8 +1258,8 @@ TOKEN {
 
 * Token Alias - a session-specific integer identifier that references a Token
   Value. The Token Alias MUST be unique within the Session. Once a Token Alias has
-  been registered, it cannot be re-registered within the Session. Use of the Token
-  Alias is optional.
+  been registered, it cannot be re-registered within the Session without first
+  being deleted. Use of the Token Alias is optional.
 
 * Token Value - the payload of the Token. The contents and serialization of this
   payload are defined by the Token Type.
@@ -1270,9 +1270,11 @@ serialization, referencing an alias which has not been registered, attempting to
 register an alias which has been previously registered, or providing a token value
 which does not match the declared Token type.
 
-Retiring an alais and token that was previously used to authorize a message has no
-retroactive effect on the original authorization, nor does it prevent that same token
-being re-registered as long as that registration is performed with a new alias.
+Using an Alias to refer to a previously registered Token Value is for efficiency
+only and has the same effect as if the Token Value was included directly.
+Retiring an Alias that was previously used to authorize a message has no
+retroactive effect on the original authorization, nor does it prevent that same
+Token Value being re-registered.
 
 Clients SHOULD only register tokens which they intend to re-use during the session.
 Client SHOULD retire previously registered tokens once their utility has passed.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -976,9 +976,9 @@ When a relay receives an incoming ANNOUNCE for a given namespace, for
 each active upstream subscription that matches that namespace, it SHOULD send a
 SUBSCRIBE to the publisher that sent the ANNOUNCE.
 
-Object headers carry a short hop-by-hop `Track ` that maps to
+Object headers carry a short hop-by-hop `Track Alias` that maps to
 the Full Track Name (see {{message-subscribe-ok}}). Relays use the
-`Track ` of an incoming Object to identify its track and find
+`Track Alias` of an incoming Object to identify its track and find
 the active subscribers for that track. Relays MUST forward Objects to
 matching subscribers in accordance to each subscription's priority, group order,
 and delivery timeout.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1182,7 +1182,7 @@ set as if no AUTHORIZATION INFO had been included.
 
 If an AUTHORIZATION ALIAS is associated with an AUTHORIZATION INFO instance and then
 later in the same session associated with a different AUTHORIZATION INFO instance,
-the receiver SHOULD update the AUTHORIZATION ALIAS to reference the last provided
+the receiver MUST update the AUTHORIZATION ALIAS to reference the last provided
 AUTHORIZATION INFO instance.
 
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -687,7 +687,7 @@ code, as defined below:
   terminate the subscription.
 
 * Auth Token Cache Overflow - the Session limit {{max-auth-token-cache-size}} of
-  all registered Authorization tokens has been exceeded.
+  the size of all registered Authorization tokens has been exceeded.
 
 * Duplicate Auth Token Alias - Authorization Token attempted to register an
   alias that was in use (see {{authorization-token}}).
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -959,9 +959,9 @@ When a relay receives an incoming ANNOUNCE for a given namespace, for
 each active upstream subscription that matches that namespace, it SHOULD send a
 SUBSCRIBE to the publisher that sent the ANNOUNCE.
 
-Object headers carry a short hop-by-hop `Track Alias` that maps to
+Object headers carry a short hop-by-hop `Track ` that maps to
 the Full Track Name (see {{message-subscribe-ok}}). Relays use the
-`Track Alias` of an incoming Object to identify its track and find
+`Track ` of an incoming Object to identify its track and find
 the active subscribers for that track. Relays MUST forward Objects to
 matching subscribers in accordance to each subscription's priority, group order,
 and delivery timeout.
@@ -1144,16 +1144,81 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
 Note that since Setup parameters use a separate namespace, it is impossible for
 these parameters to appear in Setup messages.
 
-#### AUTHORIZATION INFO {#authorization-info}
+#### AUTHORIZATION TOKEN {#authorization-token}
 
-AUTHORIZATION INFO parameter (Parameter Type 0x02) identifies a track's
+AUTHORIZATION TOKEN parameter (Parameter Type 0x01) identifies a track's
 authorization information in a SUBSCRIBE, SUBSCRIBE_ANNOUNCES, ANNOUNCE
 or FETCH message. This parameter is populated for cases where the authorization
 is required at the track level.
 
+The TOKEN value is a structured object containing an optional session-specific
+alias. The Alias allows the client to reference a previously transmitted TOKEN
+in future messages. The TOKEN value is serialized as follows:
+
+~~~
+TOKEN {
+  Token Type (i),
+  Alias Type (i),
+  [Token Alias (i),]
+  [Token Value (..)]
+}
+~~~
+{: #moq-token format title="AUTHORIZATION TOKEN value"}
+
+* Token Type  - a numeric identifier for the type of Token payload being
+  transmitted. This type is defined by the IANA table "MOQT Auth Token Type". See
+  {{iana}}. Type 0 is reserved to indicate that the type is not defined in the
+  table and must be negotiated out-of-band between client and receiver.
+
+* Alias Type - an integer defining both the serialization and the processing
+  behavior of the receiver. This Alias type has the following code points:
+
+|------|-------------------------------------------------------------------------|
+| Code | Serialization and behavior                                              |
+|-----:|:------------------------------------------------------------------------|
+| 0x0  | There is an Alias but no Value. This Alias and the Token Value it was   |
+|      | previously associated with MUST be retired. Retiring removes them from  |
+|      | the pool of actively registered tokens.                                 |
+|------|-------------------------------------------------------------------------|
+| 0x1  | There is an Alias and a Value. This Alias MUST be associated  with the  |
+|      | Token Value for the duration of the Session. This action is termed      |     
+|      | "registering" the Token.                                                |
+|------|-------------------------------------------------------------------------|
+| 0x2  | There is an Alias and no Value. Use the Token Value previously          |
+|      | registered with this Alias.                                             |
+|------|-------------------------------------------------------------------------|
+| 0x3  | There is no Alias and there is a Value. Use the Token Value as provided.|
+|      | The Token Value may be discarded after processing.                      |
+|------|-------------------------------------------------------------------------|
+
+* Token Alias  - a session-specific integer identifier that references a Token
+  Value. The Token Alias MUST be unique within the Session. Once a Token Alias has
+  been registered, it cannot be re-registered within the Session. Use of the Token
+  Alias is optional.
+
+* Token Value - the payload of the Token. The contents and serialization of this
+  payload are defined by the Token Type.
+
+The receiver of a message containing an invalid AUTHORIZATION TOKEN parameter MUST
+reject that message with an "Invalid Auth Token" error. This may be due to invalid
+serialization, referencing an alias which has not been registered, attempting to
+register an alias which has been previously registered, or providing a token value
+which does not match the declared Token type.
+
+Clients SHOULD only register tokens which they intend to re-use during the session.
+Client SHOULD retire previously registered tokens once their utility has passed.
+
+By registering a Token, the client is requiring the receiver to store the Token Alias
+and Token Value until they are retired, or the Session ends. The receiver can protect
+its resources by sending a SETUP parameter defining the MAX_AUTH_SIZE {{max-auth-size}}
+limit it is willing to accept. If a registration is attempted which would cause this
+limit to be exceeded, the receiver MUST reject that message with a
+"Max Auth Limit Exceeded" error.
+
+
 #### DELIVERY TIMEOUT Parameter {#delivery-timeout}
 
-The DELIVERY TIMEOUT parameter (Parameter Type 0x03) MAY appear in a
+The DELIVERY TIMEOUT parameter (Parameter Type 0x02) MAY appear in a
 SUBSCRIBE, SUBSCRIBE_OK, or a SUBSCRIBE_UDPATE message.  It is the duration in
 milliseconds the relay SHOULD continue to attempt forwarding Objects after
 they have been received.  The start time for the timeout is based on when the
@@ -1195,25 +1260,6 @@ beginning of the Object was received.  This means Objects earlier in a
 multi-object stream will expire earlier than Objects later in the stream. Once
 Objects have expired from cache, their state becomes unknown, and a relay that
 handles a downstream request that includes those Objects re-requests them.
-
-#### AUTHORIZATION ALIAS {#authorization-alias}
-
-AUTHORIZATION ALIAS parameter (Parameter Type 0x06) identifies a varint alias by
-which an AUTHORIZATION INFO (Parameter Type 0x02) can be referenced within the
-current session. Transmission of both AUTHORIZATION ALIAS and AUTHORIZATION INFO
-parameters in the same parameter set is an instruction to the receiver to associate
-the AUTHORIZATION ALIAS with the AUTHORIZATION INFO. Receivers processing future
-parameters sets in which only the AUTHORIZATION ALIAS is sent MUST process the
-parameter set as if the AUTHORIZATION INFO had been retransmitted.
-
-If an AUTHORIZATION ALIAS is received in a message without that alias first having been
-associated with an AUTHORIZATION INFO in that session, then the receiver MUST reject
-that message with an "Invalid Auth alias" error.
-
-If an AUTHORIZATION ALIAS is associated with an AUTHORIZATION INFO instance and then
-later in the same session associated with a different AUTHORIZATION INFO instance,
-the receiver MUST update the AUTHORIZATION ALIAS to reference the last provided
-AUTHORIZATION INFO instance.
 
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
 
@@ -1291,6 +1337,15 @@ the `query` portion of the URI to the parameter.
 The MAX_SUBSCRIBE_ID parameter (Parameter Type 0x02) communicates an initial
 value for the Maximum Subscribe ID to the receiving subscriber. The default
 value is 0, so if not specified, the peer MUST NOT create subscriptions.
+
+#### MAX_AUTH_SIZE {#max-auth-size}
+
+The MAX_AUTH_SIZE parameter (Parameter Type 0x04) communicates the maximum
+size in bytes of all actively registered Authorization tokens that the server
+is willing to store per Session. This parameter is optional. The total size is
+calculated as the sum of the size of all registered tokens (Alias Type value of
+0x01 - see {{moq-token}})  minus the size for all deregistered tokens
+(Alias Type value of 0x00), since Session initiation.
 
 ## GOAWAY {#message-goaway}
 
@@ -1592,7 +1647,9 @@ as defined below:
 |------|---------------------------|
 | 0x6  | Retry Track Alias         |
 |------|---------------------------|
-| 0x7  | Invalid Auth alias        |
+| 0x7  | Invalid Auth Token        |
+|------|---------------------------|
+| 0x8  | Max Auth Limit Exceeded   |
 |------|---------------------------|
 
 * Internal Error - An implementation specific or generic error occurred.
@@ -1614,8 +1671,13 @@ as defined below:
 * Retry Track Alias - The publisher requires the subscriber to use the given
   Track Alias when subscribing.
 
-* Invalid Auth Alias - An Authorization Alias was provided which cannot
-  be associated with an Authorization Info instance.
+* Invalid Auth Token - An invalid Authorization token was provided. This may
+  be due to invalid serialization, referencing an alias which has not been
+  registered, or attempting to register an alias which has been previously
+  registered.
+
+* Max Auth Limit Exceeded - the size of the Authorization token supplied
+  causes the Session limit of all registered Authorization tokens to be
 
 
 ## SUBSCRIBE_UPDATE {#message-subscribe-update}
@@ -2055,7 +2117,9 @@ as defined below:
 |------|---------------------------|
 | 0x6  | No Objects                |
 |------|---------------------------|
-| 0x7  | Invalid Auth alias        |
+| 0x7  | Invalid Auth Token        |
+|------|---------------------------|
+| 0x8  | Max Auth Limit Exceeded   |
 |------|---------------------------|
 
 * Internal Error - An implementation specific or generic error occurred.
@@ -2076,8 +2140,11 @@ as defined below:
 * No Objects - The beginning of the requested range is after the latest group
   and object for the track, or the track has not published any objects.
 
-* Invalid Auth Alias - An Authorization Alias was provided which cannot
-  be associated with an Authorization Info instance.
+* Invalid Auth Token - An invalid Authorization token was provided.
+
+* Max Auth Limit Exceeded - the size of the Authorization token supplied
+  causes the Session limit of all registered Authorization tokens to be
+  exceeded.
 
 
 ## FETCH_CANCEL {#message-fetch-cancel}
@@ -2247,7 +2314,9 @@ below:
 |------|---------------------------|
 | 0x4  | Uninterested              |
 |------|---------------------------|
-| 0x5  | Invalid Auth alias        |
+| 0x7  | Invalid Auth Token        |
+|------|---------------------------|
+| 0x8  | Max Auth Limit Exceeded   |
 |------|---------------------------|
 
 * Internal Error - An implementation specific or generic error occurred.
@@ -2262,8 +2331,11 @@ below:
 
 * Uninterested - The namespace is not of interest to the endpoint.
 
-* Invalid Auth Alias - An Authorization Alias was provided which cannot
-  be associated with an Authorization Info instance.
+* Invalid Auth Token - An invalid Authorization token was provided.
+
+* Max Auth Limit Exceeded - the size of the Authorization token supplied
+  causes the Session limit of all registered Authorization tokens to be
+  exceeded.
 
 
 ## UNANNOUNCE {#message-unannounce}
@@ -2421,7 +2493,9 @@ as defined below:
 |------|---------------------------|
 | 0x5  | Namespace Prefix Overlap  |
 |------|---------------------------|
-| 0x6  | Invalid Auth alias        |
+| 0x7  | Invalid Auth Token        |
+|------|---------------------------|
+| 0x8  | Max Auth Limit Exceeded   |
 |------|---------------------------|
 
 * Internal Error - An implementation specific or generic error occurred.
@@ -2440,8 +2514,12 @@ as defined below:
 * Namespace Prefix Overlap - The namespace prefix overlaps with another
   SUBSCRIBE_ANNOUNCES in the same session.
 
-* Invalid Auth Alias - An Authorization Alias was provided which cannot
-  be associated with an Authorizaiton Info instance.
+* Invalid Auth Token - An invalid Authorization token was provided.
+
+* Max Auth Limit Exceeded - the size of the Authorization token supplied
+  causes the Session limit of all registered Authorization tokens to be
+  exceeded.
+
 
 
 ## UNSUBSCRIBE_ANNOUNCES {#message-unsub-ann}
@@ -2938,6 +3016,7 @@ TODO: fill out currently missing registries:
   standards utilization where space is a premium, 64 - 16383 for
   standards utilization where space is less of a concern, and 16384 and
   above for first-come-first-served non-standardization usage.
+* MOQT Auth Token Type
 
 TODO: register the URI scheme and the ALPN and grease the Extension types
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1120,9 +1120,11 @@ Protocol Violation.
 Some messages include a Parameters field that encode optional message
 elements. They contain a type, length, and value.
 
-Senders MUST NOT repeat the same parameter type in a message. Receivers
-SHOULD check that there are no duplicate parameters and close the session
-as a 'Protocol Violation' if found.
+Senders MUST NOT repeat the same parameter type in a message unless the 
+parameter definition explicitly allows multiple instances of that type to 
+be sent in a single message. Receivers SHOULD check that there are no
+unauthorized duplicate parameters and close the session as a
+'Protocol Violation' if found.
 
 Receivers ignore unrecognized parameters.
 
@@ -1167,6 +1169,8 @@ AUTHORIZATION TOKEN parameter (Parameter Type 0x01) identifies a track's
 authorization information in a SUBSCRIBE, SUBSCRIBE_ANNOUNCES, ANNOUNCE
 or FETCH message. This parameter is populated for cases where the authorization
 is required at the track level.
+
+The AUTHORIZATION TOKEN parameter MAY be repeated within a message.
 
 The TOKEN value is a structured object containing an optional session-specific
 alias. The Alias allows the client to reference a previously transmitted TOKEN
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1177,7 +1177,7 @@ parameters sets in which only the AUTHORIZATION ALIAS is sent MUST process the
 parameter set as if the AUTHORIZATION INFO had been retransmitted.
 
 If an AUTHORIZATION ALIAS is received without first having been associated with an
-AUTHORIZATION INFO in that session, then the receiver SHALL process the parameter
+AUTHORIZATION INFO in that session, then the receiver SHOULD process the parameter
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1166,6 +1166,25 @@ means Objects earlier in a multi-object stream will expire earlier than Objects
...
+parameter set as if the AUTHORIZATION INFO had been retransmitted.
+
+If an AUTHORIZATION ALIAS is received without first having been associated with an
+AUTHORIZATION INFO in that session, then the receiver SHALL process the parameter
+set as if no AUTHORIZATION INFO had been included.
+
+If an AUTHORIZATION ALIAS is associated with an AUTHORIZATION INFO instance and then
+later in the same session associated with a different AUTHORIZATION INFO instance,
+the receiver SHALL update the AUTHORIZATION ALIAS to reference the last provided
+AUTHORIZATION INFO instance.
```
wilaw, line 1186: [Restoring prior comment chain from Suhas]
[suhasHere](https://github.com/suhasHere) [yesterday](https://github.com/moq-wg/moq-transport/pull/756#discussion_r1977932280)
I think we Should go stronger here and say terminate the connection.

@[wilaw](https://github.com/wilaw) wilaw [3 hours ago](https://github.com/moq-wg/moq-transport/pull/756#discussion_r1979438150)
Session termination seems a bit dramatic, given that auth params are optional. By supplying an alias without a reference, the client is functionally no different to one that has not provided a token and hence the server should treat them identically.


@[suhasHere](https://github.com/suhasHere) suhasHere [2 hours ago](https://github.com/moq-wg/moq-transport/pull/756#discussion_r1979671108) • 
Typically In security , it is recommended to be clear with the intent.
No auth parameters implies , there is nothing to authorize against.
Having authz parameters implies there is something we need to check before allowing the action.
But this one is in-between and my preference is to keep behavior explicit and we should disallow it.

wilaw:
Restoring Individual Comment from @afrind 

Would it be better to define a structure for Auth Info that has an optional alias? If we allow sending > 1 AUTH (https://github.com/moq-wg/moq-transport/issues/722), then there's no safe way to correlate them.

wilaw:
> Would it be better to define a structure for Auth Info that has an optional alias? If we allow sending > 1 AUTH (https://github.com/moq-wg/moq-transport/issues/722), then there's no safe way to correlate them.

Since parameters have a serialization order, one alternative option is to specify that an AUTHORIZATION ALIAS refers to the AUTHORIZATION INFO which it follows. This would allow multiple auth blobs and aliases to be communicated in one message.

draft-ietf-moq-transport.md
```diff
@@ -1166,6 +1166,25 @@ means Objects earlier in a multi-object stream will expire earlier than Objects
...
+AUTHORIZATION INFO (Parameter Type 0x02) can be referenced within the current
+session. Transmission of both AUTHORIZATION ALIAS and AUTHORIZATION INFO parameters
+in the same parameter set is an instruction to the receiver to associate the
+AUTHORIZATION ALIAS with the AUTHORIZATION INFO. Receivers processing future
+parameters sets in which only the AUTHORIZATION ALIAS is sent MUST process the
+parameter set as if the AUTHORIZATION INFO had been retransmitted.
+
+If an AUTHORIZATION ALIAS is received without first having been associated with an
+AUTHORIZATION INFO in that session, then the receiver SHALL process the parameter
+set as if no AUTHORIZATION INFO had been included.
```
wilaw, line 1181: [Restoring prior comment chain from Suhas]
[suhasHere](https://github.com/suhasHere) https://github.com/moq-wg/moq-transport/pull/756#discussion_r1977932280
I think we Should go stronger here and say terminate the connection.

@[wilaw](https://github.com/wilaw) wilaw https://github.com/moq-wg/moq-transport/pull/756#discussion_r1979438150
Session termination seems a bit dramatic, given that auth params are optional. By supplying an alias without a reference, the client is functionally no different to one that has not provided a token and hence the server should treat them identically.

@[suhasHere](https://github.com/suhasHere) suhasHere https://github.com/moq-wg/moq-transport/pull/756#discussion_r1979671108 •
Typically In security , it is recommended to be clear with the intent.
No auth parameters implies , there is nothing to authorize against.
Having authz parameters implies there is something we need to check before allowing the action.
But this one is in-between and my preference is to keep behavior explicit and we should disallow it.

draft-ietf-moq-transport.md
```diff
@@ -1166,6 +1166,25 @@ means Objects earlier in a multi-object stream will expire earlier than Objects
...
+AUTHORIZATION ALIAS parameter (Parameter Type 0x05) identifies an alias by which
+AUTHORIZATION INFO (Parameter Type 0x02) can be referenced within the current
+session. Transmission of both AUTHORIZATION ALIAS and AUTHORIZATION INFO parameters
+in the same parameter set is an instruction to the receiver to associate the
+AUTHORIZATION ALIAS with the AUTHORIZATION INFO. Receivers processing future
+parameters sets in which only the AUTHORIZATION ALIAS is sent MUST process the
+parameter set as if the AUTHORIZATION INFO had been retransmitted.
+
+If an AUTHORIZATION ALIAS is received without first having been associated with an
+AUTHORIZATION INFO in that session, then the receiver SHALL process the parameter
```
ianswett, line 1180:
```suggestion
AUTHORIZATION INFO in that session, then the receiver SHOULD process the parameter
```

ianswett: Or maybe MUST?

draft-ietf-moq-transport.md
```diff
@@ -1166,6 +1166,25 @@ means Objects earlier in a multi-object stream will expire earlier than Objects
...
+set as if no AUTHORIZATION INFO had been included.
+
+If an AUTHORIZATION ALIAS is associated with an AUTHORIZATION INFO instance and then
+later in the same session associated with a different AUTHORIZATION INFO instance,
+the receiver SHALL update the AUTHORIZATION ALIAS to reference the last provided
```
ianswett, line 1185:
```suggestion
the receiver SHOULD update the AUTHORIZATION ALIAS to reference the last provided
```
Or MUST?

ianswett (commented):
I'm not an expert on this, but I think this reads clearly.  I would prefer using SHOULD or MUST instead of SHALL, though.

ianswett:
Is there an issue or issues this fixes?  I couldn't find any.

draft-ietf-moq-transport.md
```diff
@@ -1166,6 +1166,25 @@ means Objects earlier in a multi-object stream will expire earlier than Objects
...
+parameter set as if the AUTHORIZATION INFO had been retransmitted.
+
+If an AUTHORIZATION ALIAS is received without first having been associated with an
+AUTHORIZATION INFO in that session, then the receiver SHOULD process the parameter
+set as if no AUTHORIZATION INFO had been included.
+
+If an AUTHORIZATION ALIAS is associated with an AUTHORIZATION INFO instance and then
+later in the same session associated with a different AUTHORIZATION INFO instance,
+the receiver SHOULD update the AUTHORIZATION ALIAS to reference the last provided
+AUTHORIZATION INFO instance.
```
suhasHere, line 1186:
```suggestion
AUTHORIZATION INFO instance!.
```

draft-ietf-moq-transport.md
```diff
@@ -1166,6 +1166,25 @@ means Objects earlier in a multi-object stream will expire earlier than Objects
...
+
+AUTHORIZATION ALIAS parameter (Parameter Type 0x05) identifies an alias by which
+AUTHORIZATION INFO (Parameter Type 0x02) can be referenced within the current
+session. Transmission of both AUTHORIZATION ALIAS and AUTHORIZATION INFO parameters
+in the same parameter set is an instruction to the receiver to associate the
+AUTHORIZATION ALIAS with the AUTHORIZATION INFO. Receivers processing future
+parameters sets in which only the AUTHORIZATION ALIAS is sent MUST process the
+parameter set as if the AUTHORIZATION INFO had been retransmitted.
+
+If an AUTHORIZATION ALIAS is received without first having been associated with an
```
ianswett, line 1179: I wonder if this silent failure is what we want?  If someone thinks they're authenticated as bob, but they're actually anonymous, I can imagine some subtle breakages could occur.

draft-ietf-moq-transport.md
```diff
@@ -1166,6 +1166,25 @@ means Objects earlier in a multi-object stream will expire earlier than Objects
...
+AUTHORIZATION INFO in that session, then the receiver SHOULD process the parameter
+set as if no AUTHORIZATION INFO had been included.
+
+If an AUTHORIZATION ALIAS is associated with an AUTHORIZATION INFO instance and then
+later in the same session associated with a different AUTHORIZATION INFO instance,
+the receiver SHOULD update the AUTHORIZATION ALIAS to reference the last provided
```
ianswett, line 1185:
```suggestion
the receiver MUST update the AUTHORIZATION ALIAS to reference the last provided
```

ianswett (commented):
I'm not an expert on this, but I think this reads clearly.

I think we might want some text cautioning against storing an unlimited number of these aliases, since it causes the publisher to consume an unbounded amount of memory.

wilaw:
> I think we might want some text cautioning against storing an unlimited number of these aliases, since it causes the publisher to consume an unbounded amount of memory.

@Ian - the aliases are varints, so they're not the memory constraint. It would be the tokens they represent, which the relay is then forced to store for the remainder of the session.  We can set an arbitrary alias limit of 32 for example, but a single token of 1GB in size would consume more memory. So probably more reasonable to set a max total size of all tokens that should be stored per session, and then give content distributors flexibility to trade off count and size. How about 5MB limit per session? We can start off low and raise it once auth systems are deployed and we have some idea of token size & count in production.

fluffy:
How long are these cached for ?

draft-ietf-moq-transport.md
suhasHere: This goes to my comment: https://github.com/moq-wg/moq-transport/pull/756#discussion_r1977932280

I agree with @ianswett that we should be explicit. In security, explicit posture is always the right design pattern.

draft-ietf-moq-transport.md
ianswett: I'd rather not terminate the connection unless there's a contract about exactly how many auth aliases the peer needs to support.

draft-ietf-moq-transport.md
ianswett: Yes, please revise the PR whenever you have time.  You can mention it at today's presentation, but I suspect people will prefer failure over undefined behavior.

ianswett:
> > I think we might want some text cautioning against storing an unlimited number of these aliases, since it causes the publisher to consume an unbounded amount of memory.
> 
> @ianswett - the aliases are varints, so they're not the memory constraint. It would be the tokens they represent, which the relay is then forced to store for the remainder of the session. We can set an arbitrary alias limit of 32 for example, but a single token of 1GB in size would consume more memory. So probably more reasonable to set a max total size of all tokens that should be stored per session, and then give content distributors flexibility to trade off count and size. How about 5MB limit per session? We can start off low and raise it once auth systems are deployed and we have some idea of token size & count in production.

Yes, you described it better than I did.  And yes, it's not really the number of aliases, but the total memory consumed by the aliases + tokens.

I think it's worth bringing this to the WG.  I would suggest a parameter in SETUP, which is similar to what HTTP/3 does for headers: https://datatracker.ietf.org/doc/html/rfc9114#header-size-constraints

wilaw:
>  I would suggest a parameter in SETUP

The problem with that approach is that the client is not making this token itself. It is given it ahead of attempting the connection. So its not like the client can adjust what token it sends after the result of the SETUP handshake. The entity that needs to know the max token size is the back-office system that created the token and it is not party to this handshake.

It might be more useful that if the client sends a token which is too large, that a 'Token limit exceeded" error gets thrown and this error carries an attribute which is the max limit being enforced by that relay. The client can then report that problem to its token issuer. 

I think the MOQT spec could define a max token size that SHOULD be supported, but allowing relays to vary that. For example, a CDN may wish to support a 5MB limit since they are dealing with millions of tokens, while a research relay on a private network might be happy with a 100MB limit.

draft-ietf-moq-transport.md
```diff
@@ -1166,6 +1166,25 @@ means Objects earlier in a multi-object stream will expire earlier than Objects
 later in the stream. Once Objects have expired, their state becomes unknown, and
 a relay that handles a subscription that includes those Objects re-requests them.
 
+#### AUTHORIZATION ALIAS {#authorization-alias}
+
+AUTHORIZATION ALIAS parameter (Parameter Type 0x05) identifies an alias by which
```
ianswett, line 1171: I think this alias is intended to be a varint, but I'm not sure?  I think making it a varint would be sensible, since an arbitrary length string or sequence of bytes is yet another thing to worry about the length of.

wilaw: Shall I change it to type 0x06 then, which is even and therefore carrying a varint payload? What about the resolution of https://github.com/moq-wg/moq-transport/issues/794

ianswett: I think so.  Describing it as a varint would be helpful as well, since currently you don't say whether it's a varint or something else.

I'd like to land #761 soonish because I think it's a huge improvement, and then we can debate details later.

draft-ietf-moq-transport.md
wilaw: Commit https://github.com/moq-wg/moq-transport/pull/760/commits/bb424eb771b65c33a215f7cd1875eaead64f9540 updates the behavior to be 

> If an AUTHORIZATION ALIAS is received in a message without that alias first having been associated with an AUTHORIZATION INFO in that session, then the receiver MUST reject that message with an "Invalid Auth alias" error.

draft-ietf-moq-transport.md
wilaw: Commit https://github.com/moq-wg/moq-transport/commit/bb424eb771b65c33a215f7cd1875eaead64f9540 changes the type to 0x06 and explicitly mentions varint as the payload type.

draft-ietf-moq-transport.md
```diff
@@ -1536,6 +1555,8 @@ as defined below:
 |------|---------------------------|
 | 0x6  | Retry Track Alias         |
 |------|---------------------------|
+| 0x7  | Invalid Auth alias        |
```
afrind, line 1558: If the same code is common to multiple messages, can we give it a consistent number?  I realized this will renumber some other error codes but that's probably fine.

afrind:
Can you explain the rules if I get a second AUTH ALIAS with a different token?  Is that an error or does it replace a previous value.  If it replaces a previous value, what happens to existing operations that used the alias?

draft-ietf-moq-transport.md
```diff
@@ -1166,6 +1166,25 @@ means Objects earlier in a multi-object stream will expire earlier than Objects
 later in the stream. Once Objects have expired, their state becomes unknown, and
 a relay that handles a subscription that includes those Objects re-requests them.
 
+#### AUTHORIZATION ALIAS {#authorization-alias}
+
+AUTHORIZATION ALIAS parameter (Parameter Type 0x06) identifies a varint alias by
+which an AUTHORIZATION INFO (Parameter Type 0x02) can be referenced within the
+current session. Transmission of both AUTHORIZATION ALIAS and AUTHORIZATION INFO
+parameters in the same parameter set is an instruction to the receiver to associate
+the AUTHORIZATION ALIAS with the AUTHORIZATION INFO. Receivers processing future
```
suhasHere, line 1175: I am not sure if this works if there are multiple authorization parameters in a control message . I feel we should allow for such feature.

wilaw: Our current rules prohibit repeating a parameter type in a message. 

If we change that, then we can look at https://github.com/moq-wg/moq-transport/pull/760#issuecomment-2698385821 for a solution.

wilaw:
> Can you explain the rules if I get a second AUTH ALIAS with a different token? Is that an error or does it replace a previous value. If it replaces a previous value, what happens to existing operations that used the alias?

The PR currently says that the auth info that the alias references gets updated. 

> If an AUTHORIZATION ALIAS is associated with an AUTHORIZATION INFO instance and then
> later in the same session associated with a different AUTHORIZATION INFO instance,
> the receiver MUST update the AUTHORIZATION ALIAS to reference the last provided
> AUTHORIZATION INFO instance.

Operations that used the prior alias remain unaffected, because the the alias is just a pointer to an auth info instance.  The alias is only applied when the auth info is processed. It is a wireline convenience to avoid resending the entire auth info.

afrind:
> The alias is only applied when the auth info is processed.

If the claim is bound to an IP, and the IP changes, is the old or new token used?

draft-ietf-moq-transport.md
suhasHere: I am in supportive of allowing it, It makes more sense for authorization parameter to have more than one value. 

My proposal is to allow each parameter to define if more than one value is allowed or not.

afrind:
Did this PR lose removing the restriction to ASCII in AUTHORIZATION INFO or is that a separate change now?

wilaw:
> Did this PR lose removing the restriction to ASCII in AUTHORIZATION INFO or is that a separate change now?

Yes, it lost it when I rebased for some reason. See https://github.com/moq-wg/moq-transport/pull/756/files line 1127. 
I wanted to add that change back now, but the ASCII text is removed from my patch https://github.com/wilaw/moq-transport/tree/patch-5. I can do it as a simple change afterwards in a separate PR.

afrind:
> > Did this PR lose removing the restriction to ASCII in AUTHORIZATION INFO or is that a separate change now?
> 
> Yes, it lost it when I rebased for some reason. See https://github.com/moq-wg/moq-transport/pull/756/files line 1127. I wanted to add that change back now, but the ASCII text is removed from my patch https://github.com/wilaw/moq-transport/tree/patch-5. I can do it as a simple change afterwards in a separate PR.

It was removed in 7240d9f0e0aa

wilaw:
If the claim is bound to an IP, and the IP changes, is the old or new token used?

The claims in the token and when they are evaluated will be defined by an external draft (such as https://github.com/wilaw/CAT-4-MOQT. This PR is concerned only with efficiently transmitting the token from the client to the server.

I opened https://github.com/wilaw/CAT-4-MOQT/issues/7 against CAT-4-MOQT to address your question.

afrind:
> The claims in the token and when they are evaluated will be defined by an external draft (such as https://github.com/wilaw/CAT-4-MOQT. This PR is concerned only with efficiently transmitting the token from the client to the server.

The intersection with ALIAS that I'm trying to understand is:

SUBSCRIBE track=t1, token=X, alias=1; token bound to IP
Token X evaluated, access ok
SUBSCRIBE_OK

SUBSCRIBE track=t2, alias=1;
Token X evaluated, access ok
SUBSCRIBE_OK

SUBSCRIBE track=t3, token=Y, alias=1; token bound to IP
Token Y evaluated, access ok
SUBSCRIBE_OK

IP changes, token re-evaluated for all subscriptions.  Is the subscription for t1 validating token X or token Y?

This might be less ambiguous if we make AUTH_ALIAS implicit, with 0 meaning the last transmitted AUTH_INFO, and N meaning the Nth previous one.  I also agree with Ian that the receiver needs bounded state, so at some point old AUTH_INFOs lose their ALIAS.

And I just reinvented HPACK.

wilaw:
> Is the subscription for t1 validating token X or token Y?

It's evaluating token Y, because that's the last token assigned to that alias. 

> This might be less ambiguous if we make AUTH_ALIAS implicit, with 0 meaning the last transmitted AUTH_INFO, and N meaning the Nth previous one.

I don't get the unambiguous part. The alias is just an index into a store of tokens on the server side. If you don't like the reassignment capability, then we can just prohibit the reassignment of an alias, so that there is a unique relationship between each alias and token instance. Reassignment can throw an "Invalid auth alias" error.

afrind:
> I don't get the unambiguous part.

Yeah I think reassignment is what makes the ambiguity, but I do like the idea of a FIFO token store.

wilaw:
Summarizing open issues for this PR
1. Should we [wrap auth alias and auth info together ](https://github.com/moq-wg/moq-transport/pull/760#issuecomment-2698384726)in a structure to allow sending multiple Auth Infos

fluffy:
I don't care if we have multuple auth tokens in one parameters or if we have multiuple paramaters, but one way or another, we need to have multiple tokens. And the alias needs to point at a single token

fluffy:
WIll need to deal with the issue that one token say it is authorized, and the other token says it is not.

wilaw:
Actions from interim meeting
1. Create setup param for max allowed auth info size. 
2. Create wrapper for alias + info. 
3. Prohibit reassignment. 
4. Need way  for retirement.

draft-ietf-moq-transport.md
```diff
@@ -1161,16 +1161,81 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
+its resources by sending a SETUP parameter defining the MAX_AUTH_SIZE {{max-auth-size}}
+limit it is willing to accept. If a registration is attempted which would cause this
+limit to be exceeded, the receiver MUST reject that message with a
+"Max Auth Limit Exceeded" error.
+
+
 #### DELIVERY TIMEOUT Parameter {#delivery-timeout}
 
-The DELIVERY TIMEOUT parameter (Parameter Type 0x03) MAY appear in a
+The DELIVERY TIMEOUT parameter (Parameter Type 0x02) MAY appear in a
```
afrind, line 1242: Is this renumbering from a different PR?

draft-ietf-moq-transport.md
```diff
@@ -1291,6 +1356,15 @@ The MAX_SUBSCRIBE_ID parameter (Parameter Type 0x02) communicates an initial
 value for the Maximum Subscribe ID to the receiving subscriber. The default
 value is 0, so if not specified, the peer MUST NOT create subscriptions.
 
+#### MAX_AUTH_SIZE {#max-auth-size}
+
+The MAX_AUTH_SIZE parameter (Parameter Type 0x04) communicates the maximum
+size in bytes of all actively registered Authorization tokens that the server
+is willing to store per Session. This parameter is optional. The total size is
```
afrind, line 1367: If it's optional, indicate if the default is 0 (no aliases allowed) or something else.

draft-ietf-moq-transport.md
```diff
@@ -1291,6 +1356,15 @@ The MAX_SUBSCRIBE_ID parameter (Parameter Type 0x02) communicates an initial
...
+calculated as the sum of the size of all registered tokens (Alias Type value of
+0x01 - see {{moq-token}})  minus the size for all deregistered tokens
+(Alias Type value of 0x00), since Session initiation.
```
afrind, line 1370:
```suggestion
calculated as the sum of the size of all currently registered Token Values.
```

afrind: I'll note that some fixed overhead per token might also be nice, to prevent the peer registering many 1 byte tokens.

draft-ietf-moq-transport.md
```diff
@@ -1593,6 +1667,10 @@ as defined below:
 |------|---------------------------|
 | 0x6  | Retry Track Alias         |
 |------|---------------------------|
+| 0x7  | Invalid Auth Token        |
```
afrind, line 1674: I'd move this up to 0x4 and shift the others down, since all the common error codes go first.  Though now that my impl is trying to support multiple versions, shifting error codes in each version could be a huge pain.

draft-ietf-moq-transport.md
```diff
@@ -1593,6 +1667,10 @@ as defined below:
...
+|------|---------------------------|
+| 0x8  | Max Auth Limit Exceeded   |
```
afrind, line 1676: Make this a session error rather than a per-request error.

draft-ietf-moq-transport.md
wilaw: No, I updated the renumbering to match #761 so that this type is even to support the varint value.

draft-ietf-moq-transport.md
wilaw: I'd prefer a default of unlimited. Setting it to 0 blocks the use of alias for auth which is counter-productive for the performance of the protocol. Setting it to unlimited allows aliases to work by default which improves performance. The onus is on the relay to protect itself by setting a reasonable limit.

draft-ietf-moq-transport.md
wilaw: Added 8 bytes of fixed overhead per token, with https://github.com/moq-wg/moq-transport/pull/760/commits/95359821825f824876898e49fbdf169d92a97b6a

draft-ietf-moq-transport.md
wilaw: Fixed with commit https://github.com/moq-wg/moq-transport/pull/760/commits/95359821825f824876898e49fbdf169d92a97b6a

draft-ietf-moq-transport.md
wilaw: I've already moved them twice as other PRs insert new values. I rather save the reordering for an editorial pass at the end of draft development.

draft-ietf-moq-transport.md
afrind: The onus is on the sender to track how many bytes are in the token cache and not overflow it.  If the sender doesn't do this, random requests will fail and it needs to add fallback/deletion logic when its completely unnecessary.

The logic should be something like this:

```
subscribe(track, token_type, token):
  params = []
  auto alias = token_cache.get_alias_for(token)
  if alias is None:
    if length(token) < token_cache.get_max():
      while not token_cache.has_room(token):
        params.append(AUTH_TOKEN, {
            token_type=0, alias_type=DELETE, alias=token_cache.evict_one(), value=None})
      alias = token_cache.insert(token)
      alias_type = REGISTER
  else:
    alias_type = USING_ALIAS
    token = None
     
  params.append(AUTH_TOKEN, {
    token_type=token_type, alias_type=alias_type, alias=alias, value=token})
```

This is guaranteed not to fail with "Max Auth Limit Exceeded".

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,83 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
 Note that since Setup parameters use a separate namespace, it is impossible for
 these parameters to appear in Setup messages.
 
+#### AUTHORIZATION TOKEN {#authorization-token}
+
+AUTHORIZATION TOKEN parameter (Parameter Type 0x01) identifies a track's
+authorization information in a SUBSCRIBE, SUBSCRIBE_ANNOUNCES, ANNOUNCE
+or FETCH message. This parameter is populated for cases where the authorization
+is required at the track level.
```
afrind, line 1211:
```suggestion
AUTHORIZATION TOKEN parameter (Parameter Type 0x01) identifies a track's
authorization information in a SUBSCRIBE, SUBSCRIBE_ANNOUNCES, ANNOUNCE
TRACK_STATUS_REQUEST or FETCH message. This parameter is populated for
cases where the authorization is required at the track level.
```

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,83 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
+TOKEN {
+  Token Type (i),
+  Alias Type (i),
+  [Token Alias (i),]
+  [Token Value (..)]
+}
+~~~
+{: #moq-token format title="AUTHORIZATION TOKEN value"}
+
+* Token Type  - a numeric identifier for the type of Token payload being
```
afrind, line 1229:
```suggestion
* Token Type - a numeric identifier for the type of Token payload being
```

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,83 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
+| 0x1  | There is an Alias and a Value. This Alias MUST be associated  with the  |
+|      | Token Value for the duration of the Session. This action is termed      |
+|      | "registering" the Token.                                                |
+|------|-------------------------------------------------------------------------|
+| 0x2  | There is an Alias and no Value. Use the Token Value previously          |
+|      | registered with this Alias.                                             |
+|------|-------------------------------------------------------------------------|
+| 0x3  | There is no Alias and there is a Value. Use the Token Value as provided.|
+|      | The Token Value may be discarded after processing.                      |
+|------|-------------------------------------------------------------------------|
```
afrind, line 1253:
```suggestion
|------|------------|------------------------------------------------------|
| Code | Name       | Serialization and behavior                           |
|-----:|:-----------|------------------------------------------------------|
| 0x0  | DELETE     | There is an Alias but no Value. This Alias and the   |
|      |            | Token Value it was previously associated with MUST be|
|      |            | retired. Retiring removes them from the pool of      |
|      |            | actively registered tokens.                          |
|------|------------|------------------------------------------------------|
| 0x1  | REGISTER   | There is an Alias and a Value. This Alias MUST be    |
|      |            | associated with the Token Value for the duration of  |
|      |            | the Session. This action is termed "registering" the |
|      |            | Token.                                               |
|------|------------|------------------------------------------------------|
| 0x2  | USE_ALIAS  | There is an Alias and no Value. Use the Token Value  |
|      |            | previously registered with this Alias.               |
|------|------------|------------------------------------------------------|
| 0x3  | USE_VALUE  | There is no Alias and there is a Value. Use the Token|
|      |            | Value as provided. The Token Value may be discarded  |
|      |            | after processing.                                    |
|------|------------|------------------------------------------------------|

```

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,83 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
+
+* Token Alias  - a session-specific integer identifier that references a Token
```
afrind, line 1255:
```suggestion
* Token Alias - a session-specific integer identifier that references a Token
```

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,83 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
+  Value. The Token Alias MUST be unique within the Session. Once a Token Alias has
+  been registered, it cannot be re-registered within the Session. Use of the Token
+  Alias is optional.
```
afrind, line 1258:
```suggestion
  been registered, it cannot be re-registered within the Session without first
  being deleted. Use of the Token Alias is optional.
```

afrind (approved):
I reformatted the table and gave logical names to the Alias Type, and made a few other minor suggestions.

Otherwise lgtm

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,89 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
 these parameters to appear in Setup messages.
 
+#### AUTHORIZATION TOKEN {#authorization-token}
+
+AUTHORIZATION TOKEN parameter (Parameter Type 0x01) identifies a track's
+authorization information in a SUBSCRIBE, SUBSCRIBE_ANNOUNCES, ANNOUNCE
+TRACK_STATUS_REQUEST or FETCH message. This parameter is populated for
+cases where the authorization is required at the track level.
+
+The AUTHORIZATION TOKEN parameter MAY be repeated within a message.
```
suhasHere, line 1213:
```suggestion
The AUTHORIZATION TOKEN parameter MAY be repeated within a message as long as they aren't duplicates, otherwise the receiver MUST close the session with "Protocol Violation".
```

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,89 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
+* Alias Type - an integer defining both the serialization and the processing
+  behavior of the receiver. This Alias type has the following code points:
+
+|------|------------|------------------------------------------------------|
+| Code | Name       | Serialization and behavior                           |
+|-----:|:-----------|------------------------------------------------------|
+| 0x0  | DELETE     | There is an Alias but no Value. This Alias and the   |
+|      |            | Token Value it was previously associated with MUST be|
+|      |            | retired. Retiring removes them from the pool of      |
+|      |            | actively registered tokens.                          |
```
suhasHere, line 1243: This makes sense. However,  I wonder how can we enforce that the server did remove the mapping. 
Is the idea to remove the entry on the server to reuse the alias ? If so, this seems slightly complicated or not needed. Instead if the requirement is to help server memory, I doubt if this helps much. Since, if the server is constrained by memory to store tokens, it might close the session.

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,89 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
+|------|------------|------------------------------------------------------|
+| 0x1  | REGISTER   | There is an Alias and a Value. This Alias MUST be    |
+|      |            | associated with the Token Value for the duration of  |
+|      |            | the Session. This action is termed "registering" the |
+|      |            | Token.                                               |
+|------|------------|------------------------------------------------------|
+| 0x2  | USE_ALIAS  | There is an Alias and no Value. Use the Token Value  |
+|      |            | previously registered with this Alias.               |
```
suhasHere, line 1251:
```suggestion
|      |            | previously registered with this Alias.   A client uses this when sending further control messages after initially registering the token that still is valid, to avoid resending the Token Value each time.       |
```

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,89 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
+#### AUTHORIZATION TOKEN {#authorization-token}
+
+AUTHORIZATION TOKEN parameter (Parameter Type 0x01) identifies a track's
+authorization information in a SUBSCRIBE, SUBSCRIBE_ANNOUNCES, ANNOUNCE
+TRACK_STATUS_REQUEST or FETCH message. This parameter is populated for
+cases where the authorization is required at the track level.
+
+The AUTHORIZATION TOKEN parameter MAY be repeated within a message.
+
+The TOKEN value is a structured object containing an optional session-specific
```
suhasHere, line 1215:
```suggestion
The TOKEN value is a structured object as defined below:
```

suhasHere, line 1215:
```suggestion
The TOKEN value is a structured object as defined below:
```

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,89 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
+|      |            | the Session. This action is termed "registering" the |
+|      |            | Token.                                               |
+|------|------------|------------------------------------------------------|
+| 0x2  | USE_ALIAS  | There is an Alias and no Value. Use the Token Value  |
+|      |            | previously registered with this Alias.               |
+|------|------------|------------------------------------------------------|
+| 0x3  | USE_VALUE  | There is no Alias and there is a Value. Use the Token|
+|      |            | Value as provided. The Token Value may be discarded  |
+|      |            | after processing.                                    |
+|------|------------|------------------------------------------------------|
```
suhasHere, line 1256: Wondering if we need this special case at all ? Why can't always have token mapped to an alias.

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,89 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
+  being deleted. Use of the Token Alias is optional.
+
+* Token Value - the payload of the Token. The contents and serialization of this
+  payload are defined by the Token Type.
+
+The receiver of a message containing an invalid AUTHORIZATION TOKEN parameter MUST
+reject that message with an "Invalid Auth Token" error. This may be due to invalid
+serialization, referencing an alias which has not been registered, attempting to
+register an alias which has been previously registered, or providing a token value
+which does not match the declared Token type.
```
suhasHere, line 1271: We should also add the case of token that failed authorization to the list.

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,89 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
+
+Using an Alias to refer to a previously registered Token Value is for efficiency
+only and has the same effect as if the Token Value was included directly.
+Retiring an Alias that was previously used to authorize a message has no
+retroactive effect on the original authorization, nor does it prevent that same
+Token Value being re-registered.
+
+Clients SHOULD only register tokens which they intend to re-use during the session.
+Client SHOULD retire previously registered tokens once their utility has passed.
```
suhasHere, line 1280: Token's expiry should be the way to retire a token and I am not aware of use-cases where retiring by remove a token typically works.  Love to learn more about the use-case for this .

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,89 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
+
+By registering a Token, the client is requiring the receiver to store the Token Alias
+and Token Value until they are retired, or the Session ends. The receiver can protect
```
suhasHere, line 1283:
```suggestion
and Token Value until they are expired, retired, or the Session ends. The receiver can protect
```

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,89 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
+its resources by sending a SETUP parameter defining the MAX_AUTH_SIZE {{max-auth-size}}
```
suhasHere, line 1284:
```suggestion
its resources by sending a SETUP parameter defining the TOTAL_TOKEN_SIZE_IN_BYTES{total-token-size-in-bytes}}
```

draft-ietf-moq-transport.md
```diff
@@ -1328,6 +1418,19 @@ The MAX_SUBSCRIBE_ID parameter (Parameter Type 0x02) communicates an initial
...
+#### MAX_AUTH_SIZE {#max-auth-size}
+
+The MAX_AUTH_SIZE parameter (Parameter Type 0x04) communicates the maximum
+size in bytes of all actively registered Authorization tokens that the server
+is willing to store per Session. This parameter is optional. The default value
+is 0 which prohibits the use of token aliases.
+
+The token size is calculated as 8 bytes + the size of the Token value
+(see {{moq-token}}). The total size as restricted by the MAX_AUTH_SIZE parameter
+is calculated as the sum of the token sizes for all registered tokens (Alias Type
```
suhasHere, line 1430:
```suggestion
is calculated as the sum of the token sizes for all actively registered tokens (Alias Type
```

draft-ietf-moq-transport.md
```diff
@@ -1654,6 +1759,10 @@ as defined below:
 * Retry Track Alias - The publisher requires the subscriber to use the given
   Track Alias when subscribing.
 
+* Invalid Auth Token - An invalid Authorization token was provided. This may
+  be due to invalid serialization, referencing an alias which has not been
+  registered, or attempting to register an alias which has been previously
+  registered.
```
suhasHere, line 1765: We need to add token expiry to the list too.

draft-ietf-moq-transport.md
```diff
@@ -1634,6 +1737,8 @@ as defined below:
 |------|---------------------------|
 | 0x6  | Retry Track Alias         |
 |------|---------------------------|
+| 0x7  | Invalid Auth Token        |
```
suhasHere, line 1740:
```suggestion
| 0x7  | Malformed Auth Token        |
```

draft-ietf-moq-transport.md
afrind: Would it be sufficient to state that the Token Type MUST be unique?

draft-ietf-moq-transport.md
afrind: > Is the idea to remove the entry on the server to reuse the alias

Because the receiver limits the space available for tokens to be reused, it needs a mechanism to delete/retire old ones so it can use new ones.  It does not have to reuse the alias value, but it is allowed.

> if the server is constrained by memory to store tokens, it might close the session.

The goal is predictable and interoperable behavior between implementations.  You declare the amount of memory you are willing to allocate to token storage, and the subscriber is responsible for staying under the limit.  I think the model is very similar to HTTP header compression and we have a lot of experience with that.

draft-ietf-moq-transport.md
afrind: A single token larger than the receiver's token storage cannot be registered.

draft-ietf-moq-transport.md
afrind: There's already Unauthorized code for that in each response.  The cases listed here can all be Session errors as well.  Given then seriousness of getting Auth mixed up, that might be safer.

draft-ietf-moq-transport.md
afrind: MoQT has no notion of token expiry, tokens are just blobs.  Do you think we should pull that into the MoQT Token structure?

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,89 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
+  {{iana}}. Type 0 is reserved to indicate that the type is not defined in the
+  table and must be negotiated out-of-band between client and receiver.
+
+* Alias Type - an integer defining both the serialization and the processing
+  behavior of the receiver. This Alias type has the following code points:
+
+|------|------------|------------------------------------------------------|
+| Code | Name       | Serialization and behavior                           |
+|-----:|:-----------|------------------------------------------------------|
+| 0x0  | DELETE     | There is an Alias but no Value. This Alias and the   |
```
sharmafb, line 1240: I don't know if I'm missing something, but the deletion semantics seem a little odd to me. Let's say a SUBSCRIBE message introduces an authorization token, and the client later wants to retire that token. Is it true then, that they'd have to piggy-back the deletion on a future message like a SUBSCRIBE or SUBSCRIBE_ANNOUNCES? Why not have deletion just be a separate control message, rather than having this piggy-backing logic?

sharmafb (approved):

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,89 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
...
+
+The AUTHORIZATION TOKEN parameter MAY be repeated within a message.
+
+The TOKEN value is a structured object containing an optional session-specific
+alias. The Alias allows the client to reference a previously transmitted TOKEN
+in future messages. The TOKEN value is serialized as follows:
+
+~~~
+TOKEN {
+  Token Type (i),
```
MikeBishop, line 1221: What's the scenario where you would declare a different Token Type when using the alias subsequently? I'd think it's more natural to say that the Token Type and Token Value appear or are omitted together.

MikeBishop (approved):
I think this is largely reasonable. Does the token's permission get re-evaluated each time because what it authorizes might change on the back-end, or does the server do a single access check of the token and then remember what it grants access to? The difference is that the former requires storing the literal blob (whose size the client knows, as it sent it) while the latter requires storing information which the server extracted or looked up and whose size the client cannot gauge.

The discussion about removing expired tokens might lead to the latter in at least some cases, where a server forgets the blob and merely remembers the alias maps to an expired token. If either is a valid implementation, it might be worth discussing the trade-off.

afrind:
> I think this is largely reasonable. Does the token's permission get re-evaluated each time because what it authorizes might change on the back-end, or does the server do a single access check of the token and then remember what it grants access to? The difference is that the former requires storing the literal blob (whose size the client knows, as it sent it) while the latter requires storing information which the server extracted or looked up and whose size the client cannot gauge.

This is a good point. Since the definitions of the tokens are outside the scope of this document, I'm not sure we know.  Some systems might require storing the original token blob for re-evaluation (perhaps a URL that needs to be queried), while others might contain claims which could be stored which could be smaller, or larger if tokens are compressed.  In the absence of better mechanism to know how much storage the receiver is committing to, the length of the blob seems like a fine proxy?

> The discussion about removing expired tokens might lead to the latter in at least some cases, where a server forgets the blob and merely remembers the alias maps to an expired token. If either is a valid implementation, it might be worth discussing the trade-off.

I think either choice (storing the blob or storing the granted permissions) allows the receiver to clear the stored information if it detects expiration.  I think the PR can reflect something along these lines, but at least for now still require explicit deletion of the Alias by the sender.

draft-ietf-moq-transport.md
```diff
@@ -1171,9 +1176,11 @@ Protocol Violation.
...
 elements.
 
-Senders MUST NOT repeat the same parameter type in a message. Receivers
-SHOULD check that there are no duplicate parameters and close the session
-as a 'Protocol Violation' if found.
+Senders MUST NOT repeat the same parameter type in a message unless the
+parameter definition explicitly allows multiple instances of that type to
+be sent in a single message. Receivers SHOULD check that there are no
+unauthorized duplicate parameters and close the session as a
+'Protocol Violation' if found.
```
martinduke, line 1183: Maybe it's worth clarifying that receivers MUST NOT check for duplicates on unknown parameters?

draft-ietf-moq-transport.md
```diff
@@ -1196,6 +1203,89 @@ it can appear. If it appears in some other type of message, it MUST be ignored.
 Note that since Setup parameters use a separate namespace, it is impossible for
 these parameters to appear in Setup messages.
 
+#### AUTHORIZATION TOKEN {#authorization-token}
+
+AUTHORIZATION TOKEN parameter (Parameter Type 0x01) identifies a track's
```
martinduke, line 1208:
```suggestion
The AUTHORIZATION TOKEN parameter (Parameter Type 0x01) identifies a track's
```

martinduke (approved):
IIUC the AUTHORIZATION_INFO parameter (8.1.1.2) should be deleted as part of this PR.

draft-ietf-moq-transport.md
fluffy: agree

fluffy (approved):
few fix up but overall looks good to me

draft-ietf-moq-transport.md
suhasHere: I don't think so. I was trying to differentiate the case when token is malformed due to serde error vs when token failed to authorize. Since we have authz error, this looks fine .

suhasHere:
LGTM with few suggested fixes

draft-ietf-moq-transport.md
afrind: I added an error code for Expired, and explanatory text saying that receiver doesn't need to keep state after it detects an expiration, other than that Alias=X is expired.

draft-ietf-moq-transport.md
```diff
@@ -682,6 +686,12 @@ code, as defined below:
   open subgroup stream, it MAY send a STOP_SENDING on that stream or
   terminate the subscription.
 
+* Auth Token Cache Overflow - the Session limit {{max-auth-token-cache-size}} of
```
ianswett, line 689:
```suggestion
* Auth Token Cache Overflow - the Session limit {{max-auth-token-cache-size}} on the size of
```

ianswett (approved):
One small suggestion, but LG

draft-ietf-moq-transport.md
wilaw: > Token's expiry should be the way to retire a token and I am not aware of use-cases where retiring by remove a token typically works. Love to learn more about the use-case for this .

A player may be a given a token by its CMS to allow it to subscribe to a long running live stream for 20min. Every 20min it is given a new token for an additional 20min. This strategy is often used to force pirates to roll their token sharing at a higher rate. So the player would want to retire the old token when providing the new token, so that it doesn't run in to the MAX_AUTH_TOKEN_CACHE_SIZE limit.

draft-ietf-moq-transport.md
```diff
@@ -2192,8 +2335,16 @@ as defined below:
...
 
-* Invalid Subscribe ID - The joining Fetch referenced a Subscribe ID that did
-  not belong to an active Subscription.
+* Invalid Joining Request ID - The joining Fetch referenced a Request ID that
+  did not belong to an active Subscription.
+
+* Malformed Auth Token - Invalid Auth Token serialization during registration
+  (see {{authorization-token}}).
+
+* Unknown Auth Token Alias - Authroization Token refers to an alias that is
```
wilaw, line 2344:
```suggestion
* Unknown Auth Token Alias - Authorization Token refers to an alias that is
```

draft-ietf-moq-transport.md
```diff
@@ -2395,6 +2554,14 @@ below:
 
 * Uninterested - The namespace is not of interest to the endpoint.
 
+* Malformed Auth Token - Invalid Auth Token serialization during registration
+  (see {{authorization-token}}).
+
+* Unknown Auth Token Alias - Authroization Token refers to an alias that is
```
wilaw, line 2560:
```suggestion
* Unknown Auth Token Alias - Authorization Token refers to an alias that is
```

draft-ietf-moq-transport.md
```diff
@@ -2578,6 +2751,14 @@ as defined below:
 * Namespace Prefix Overlap - The namespace prefix overlaps with another
   SUBSCRIBE_ANNOUNCES in the same session.
 
+* Malformed Auth Token - Invalid Auth Token serialization during registration
+  (see {{authorization-token}}).
+
+* Unknown Auth Token Alias - Authroization Token refers to an alias that is
```
wilaw, line 2757:
```suggestion
* Unknown Auth Token Alias - Authorization Token refers to an alias that is
```

Issue #273, opened 2023-10-16
Nothing good comes out of a reason phrase


fluffy:
Reasons phrases either have useful information in them, in which case it should not be in a reason phrase, it should be in a well defined field in the protocol. Or they do not have useful information in which case they should not be there. 

I think we should get rid of reason phrases - if we need to describe more information about why an error happened, we should put that data in well defined fields.

afrind:
Individual Comment: 

I find value in reason phrases.  I agree that it's better to have specific error information in the error code, but often implementations have multiple reasons to send the same code, and adding the phrase enhances debugabiltity.  This is especially true once the error codes have frozen.  One implementation may find an error condition where none of the existing reason codes fit, so they are left with sending "UNKNOWN" or "PROTOCOL ERROR".  

HTTP removed per message reason phrases in H2.  This was good since they were always the same and redundant with the codes they had, or enough implementations assumed they were they wouldn't get proxied.  Note though that the "Proxy-Status" header has brought back an extensible mechanism for communicating additional context when something goes wrong.  I still miss the debug data in an H2 GOAWAY which was removed in H3. We log these strings and can group by them to find interesting interop bugs.

If nothing else, I think we should keep reason phrases around in moq for a while, because it's going to help interop testing. If we feel like we've nailed the error codes later on, we can yank them closer to the end.

ianswett:
I agree that at least for now, these are useful.  When developing a new extension/optimization/etc, sometimes these can really help debugging the change.

kixelated:
I think the reasons are useful for early interop. Nobody will instrument each error, nor do we have defined error codes for all permutations, so likely you'll have a big try/catch statement followed with `code=500 reason=error.messsage`.

kixelated:
In QUIC, `CONNECTION_CLOSE` has a code/reason while `STOP_SENDING/RESET_STREAM` only has a code.

I like it and we should copy it; remove `reason` from `ANNOUNCE_*` and `SUBSCRIBE_*`.

martinduke:
+1 for keeping these at least until interop has matured.

alvestrand:
Seems like a rehash of the debate we have every time we design an error field.
My usual distillation is:
- If the application that gets the error can do something sensible based on it, it needs to be a structured field.
- If it's useful for debugging the problem, it's an unstructured text field (and that's OK).

[...]

PR #761, opened 2025-03-05
Harmonizing serialization of Parameters and Extensions.

### LINKED ISSUE #672: Unify the TLV Schemes
Parameters have long had a TLV scheme where there was always a length field. I seem to remember that the value might be a varint, which would make length redundant, but I can't find that anywhere.

Extensions use a different scheme where even types are varints and odd types are blobs.

If we could just agree how to do TLVs, it would save some complicated and finicky code.

wilaw:
This PR applies extends serialization of Extensions (in which even types carry a varint payload and odd types carry a Length/Value payload) to parameters. The serialization is defined first for parameters and then referenced for Extensions.

The DELIVERY TIMEOUT parameter changes from a type of 0x03 to 0x02 as it carries a varint payload. 

The AUTHORIZATION INFO parameter changes from 0x02to 0x03 as it carries a TL payload. 

Renumbers the CLIENT_SETUP and SERVER_SETUP messages to make supporting both serializations easier.

No changes in type required for the SETUP parameters. 

Fixes #672


draft-ietf-moq-transport.md
```diff
@@ -562,7 +562,7 @@ code, as defined below:
 |------|---------------------------|
 | 0x4  | Duplicate Track Alias     |
 |------|---------------------------|
-| 0x5  | Parameter Length Mismatch |
+| 0x5  | Parameter Value Mismatch |
```
ianswett, line 565: I like Parameter Length Mismatch more, even with this change.

ianswett (approved):
Thanks for writing this up, it looks very good.

draft-ietf-moq-transport.md
ianswett: Looking at the text below again, I think there are two potential issues.  One is that you try to read a varint or variable length field and run out of bytes.  The other is that the value read doesn't match the valid values for the parameter or extension.

I think the first is a length issue, though calling it a mismatch might be misleading?

The latter I might call "Invalid Parameter Value" or "Parameter Decoding Error" or ?

draft-ietf-moq-transport.md
```diff
@@ -1072,41 +1072,45 @@ length of the message content, the receiver MUST close the session.
...
-elements. They contain a type, length, and value.
+elements. They contain a type, an optional length, and a value.
 
 Senders MUST NOT repeat the same parameter type in a message. Receivers
 SHOULD check that there are no duplicate parameters and close the session
 as a 'Protocol Violation' if found.
 
 Receivers ignore unrecognized parameters.
 
+Parameters MUST be serialized in ascending Parameter Type order.
```
LPardue, line 1083: Or else what? (I supect you want to say "the receiver MUST terminate the session with error code 'Parameter Value Mismatch'.")

draft-ietf-moq-transport.md
wilaw: Can we find a name that is broad enough to capture the 3 error conditions under which we night want to throw:
1. The varint defining the type is not well formatted
2. The length declared does not match the payload size
3. The extension is not well formatted for its type (even or odd)
4. The parameters have not been serialized in ascending order. 

How about "Parameter Formatting Error" ? 

or do we want to throw different errors for the different conditions?

afrind: Who said #273

afrind:
Individual Comment:

If we are changing the format of *SETUP, can we also assign new code points as well?  moxygen is now capable of negotiating the moq version and we plan to support multiple draft versions simultaneously, but SETUP is how we pick the version.

wilaw:
> what is the rationale for requiring ascending order? 

@gmarzot Because for compression, the Parameter Type is delta encoded. The first Parameter Type in the message is encoded with an absolute value and all subsequent Parameter Types are encoded as a delta to the prior Parameter Type. Since we only allow positive values for Type, we must necessarily enforce an ascending order.

gmarzot:
> > what is the rationale for requiring ascending order?
> 
> @gmarzot Because for compression, the Parameter Type is delta encoded.

I see, and this applies equally to object extensions which may be added and removed .. i was thinking ordering makes changes to extensions harder, but any change will affect object serialization. this seems to imply in practical relay implementations that it can be more efficient to cache object metadata separately from the payload.

this is a separate issue, but i am thinking to propose extensions be added to SUBGROUP_HEADER as a way to set set group-wide properties. This provides a mechanism (e.g., NO_OBJECT_EXT(0x20): 1)  to completely skip object level extension footprint including length.

draft-ietf-moq-transport.md
ianswett: I'm fine with "Parameter Formatting Error" or "Parameter Parsing Error"?

draft-ietf-moq-transport.md
ianswett: I actually think with this encoding, a MUST isn't necessary because you can't do anything else.

```suggestion
Parameters are serialized in ascending Parameter Type order using a delta encoding.
```

draft-ietf-moq-transport.md
```diff
@@ -1072,41 +1072,45 @@ length of the message content, the receiver MUST close the session.
...
 
-Parameter Type is an integer that indicates the semantic meaning of the
-parameter. Setup message parameters use a namespace that is constant across all
-MoQ Transport versions. All other messages use a version-specific namespace. For
-example, the integer '1' can refer to different parameters for Setup messages
-and for all other message types.
-
-SETUP message parameter types are defined in {{setup-params}}. Version-
-specific parameter types are defined in {{version-specific-params}}.
+Parameter Type is an unsigned integer that indicates the semantic meaning of the
```
suhasHere, line 1097: should we say variable length integer instead?

draft-ietf-moq-transport.md
fluffy, line 1083: Depending on how integrrity works, you might not want this. If some of them are end to end protected and some are not, you can probably not do this. I don't see the reason to do this.

suhasHere: I am not really sure if we need the serializaiton requirement and/or the delta encoding .. I haven;t thought, but I am not sure how would this work when we register things with IANA where the assigned number space may not fit with the proposed way of encoding.

fluffy: I'm very concerned with delta encoding for this. That would mean a relay that add a extension would have to rewrite all the following extensions and seem like it probably won't save much space.

draft-ietf-moq-transport.md
```diff
@@ -1072,41 +1072,45 @@ length of the message content, the receiver MUST close the session.
...
+versions. All other messages use a version-specific namespace. For example, the
+integer '1' can refer to different parameters for Setup messages and for all other
+message types. SETUP message parameter types are defined in {{setup-params}}.
+Version-specific parameter types are defined in {{version-specific-params}}.
 
-The Parameter Length field of the String Parameter encodes the length
-of the Parameter Value field in bytes.
+Parameter Values: even Parameter Types are followed by a single varint encoded
+value. Odd Parameter Types are followed by the Parameter Value length in bytes
+followed by the Parameter Value.
```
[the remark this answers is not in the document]
suhasHere: Even/Odd idea is cool , but I wonder this is overly optimizing somehting that is in control parameters.

fluffy (changes requested):
I have not really thought deeply about this but I don't think this is the time to sort this out. We clearly need a larger pass a bit later to unify error codes and such and we should deal with this then.  I would rather come back to this later.

draft-ietf-moq-transport.md
```diff
@@ -1156,6 +1153,13 @@ Publishers SHOULD consider whether the entire Object is likely to be delivered
 before sending any data for that Object, taking into account priorities,
 congestion control, and any other relevant information.
 
+#### AUTHORIZATION INFO {#authorization-info}
+
+AUTHORIZATION INFO parameter (Parameter Type 0x03) identifies a track's
+authorization information in a SUBSCRIBE, SUBSCRIBE_ANNOUNCES, ANNOUNCE
+or FETCH message. This parameter is populated for cases where the authorization
+is required at the track level. The value is an ASCII string.
```
suhasHere, line 1161: s/ASCII String/ array of bytes

draft-ietf-moq-transport.md
ianswett: It's easier to validate a parameter if it's all in one place, rather than one instance at the beginning and another at the end, based on my experience with HTTP headers.  HTTP headers are strings, so it's not as trivial to enforce ordering, but parameters have assigned varint codepoints, so ordering is obvious.

draft-ietf-moq-transport.md
suhasHere, line 1161:
```suggestion
is required at the track level. The value is array of bytes.
```

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
ianswett: It's nice to have a single parser for both, though.

fluffy:
I have thought about this more and do now want to require extensions to be ordered and also nor deleted encoded. I think that will mess up things we want to do with E2EE security. The case I worry about is where a relay adds an extension header.

mzanaty:
Wire efficiency is more critical for the data plane object messages than control messages. I strongly feel issue #794 is the right path for object extensions. If parameters or other control TLVs want to follow it, great, but if not, extensions should not follow the design of parameters. It makes no sense for large values (>8 bytes) to save a single byte of length at the expense of drastically reducing the small type space in half. Reserve the efficient, small type space with no length overhead for small values (<8 bytes).

fluffy (changes requested):
The delta encoding and requiring in order breaks E2EE and I'm strongly opposed to that part of this PR. I want to do the E2EE much like we did it for extentions in webrtc as that had a lot of thought put into and worked well.

afrind:
@fluffy Can you write down the requirements on extensions to make e2ee work with our extension model, maybe in PR in the extension section?  For example, if the extension order is important, and can't be modified by relays, that needs to be normative.  Right now the draft just says MUST forward, MAY add/modify/remove, but it sounds like there are more specific requirements.

wilaw:
> The delta encoding and requiring in order breaks E2EE 

What is the use-case for encrypting object extensions? Anything generated by the original publisher and intended for the final subscriber (i.e end-to-end) should go in the object payload, where we already have a clean architectural model for encryption. In fact I would use the need for encryption as the litmus test as to whether something should be carried in an object extension versus an object payload.

fluffy:
So here is requirements and then I will describe solution stolen from WebRTC>

Requirements when using E2E Encryptions (note many things will not use E2E Encryption):

* the whole payload is E2E encrypted - no more, no less. This makes it easier to reason about security proofs. 

* by default, MoQ headers extensions are *not* part of the E2E "extra authenticated data"

* it is possible to mark *some* of the moq headers extensions  as E2E authenticated. Relays can read these headers but can not change or remove them. 

* this does not increase bandwidth usage when it is not used 

The solution I would propose is roughly this:

There is a header extension called "start of authenticated headers". All the headers extension *after* that header are including in the E2E "extra authenticated data"  

Relays can not modify, or remove headers that follow this one but they can add, change, remove header extensions before the "start of authenticated headers".

draft-ietf-moq-transport.md
wilaw: Changed error text to "parameter Formatting Error" with commit https://github.com/moq-wg/moq-transport/pull/761/commits/8e8257db9adc4d51d83b07e15191181d19bdf7da

draft-ietf-moq-transport.md
```diff
@@ -1072,41 +1072,44 @@ length of the message content, the receiver MUST close the session.
...
 The format of Parameters is as follows:
 
 ~~~
 Parameter {
   Parameter Type (i),
-  Parameter Length (i),
-  Parameter Value (..),
+  [Parameter Value (i)]
+  [Parameter Length (i),
+   Parameter Value (..)]
```
afrind, line 1092: See the way that I changed Extension definition to not require multi-line [] optional notation.

draft-ietf-moq-transport.md
afrind: +1 I thought we removed ASCII string already, did it find it's way back.

draft-ietf-moq-transport.md
afrind: Does the ascending order requirement break what @fluffy is trying to do with authenticated extensions?

afrind (commented):
I think a rebase is needed.

draft-ietf-moq-transport.md
wilaw: Removed the requirement to serialize in ascending type order with commit https://github.com/moq-wg/moq-transport/pull/761/commits/17db4c969a0a46a309ca6b960d1b207da9030ea8.

It is not needed since we removed the delta encoding. Parameters can now be serialized in any type order.

wilaw: Commits https://github.com/moq-wg/moq-transport/pull/761/commits/8e8257db9adc4d51d83b07e15191181d19bdf7da and https://github.com/moq-wg/moq-transport/pull/761/commits/17db4c969a0a46a309ca6b960d1b207da9030ea8  removed delta encoding and requirement for serializing in asecnding order.

draft-ietf-moq-transport.md
wilaw: I rebased with https://github.com/moq-wg/moq-transport/pull/761/commits/75498823b4e13e3b97b27ba1ca21a43f09ad6aa7 and that restored the removal.

wilaw: The value is defined by the serialization, so no need to repeat here.

draft-ietf-moq-transport.md
wilaw: Resolved with https://github.com/moq-wg/moq-transport/pull/761/commits/75498823b4e13e3b97b27ba1ca21a43f09ad6aa7

fluffy:
Comment on Mo's comment. I think we could help resolve this with wherever we define this, we say it is optimized for data plane, and control plan just uses the same thing.

wilaw:
Action from interim
1. Move serialization definition to Common Wire Format  - Location - and then refer to it from Parameters and Extensions section.

draft-ietf-moq-transport.md
```diff
@@ -1038,9 +1038,9 @@ MOQT Control Message {
 |-------|-----------------------------------------------------|
 | ID    | Messages                                            |
 |------:|:----------------------------------------------------|
-| 0x40  | CLIENT_SETUP ({{message-setup}})                    |
+| 0x50  | CLIENT_SETUP ({{message-setup}})                    |
```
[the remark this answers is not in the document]
afrind: I asked Will to renumber the SETUP messages so you can write an implementation that supports both draft <= 10 and > 10.  If we kept the code point the same but changed the format of the message that negotiates version, I don't know how to do that.

wilaw:
With commit https://github.com/moq-wg/moq-transport/pull/761/commits/758984147993f43388a1d69f97d90a44b76e17a0 Parameters and extensions are now defined in terms of a new generic data structure called Key-Value-Pair.

A note was added that this structure operates in both the data and control planes but is optimized for the data plane. We can apply that optimization in a separate PR once we decide what it is, but for now the definition is cleanly in one place.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
ianswett: Thanks!

ianswett (approved):

afrind (approved):
lgtm

afrind:
Since this is non-obvious and the comment has been hidden:

This PR changes the format of SETUP, which is used to negotiate versions, so I asked Will to rotate the SETUP code points.  This is required if an implementation wants to support both 10 and 10+, but will be removed in the before we ship.

draft-ietf-moq-transport.md
```diff
@@ -296,6 +296,36 @@ Location A < Location B iff
...
+  Type (i),
+  [Length (i),]
+  Value (..)
+}
+~~~
+{: #moq-key-value-pair format title="MOQT Key-Value-Pair"}
+
+* Type: an unsigned integer, encoded as a varint, identifying the
+  type of the value and also the subsequent serialization.
+* Length: Only present when Type is odd. Species the length of the
```
suhasHere, line 320:
```suggestion
* Length: Only present when Type is odd. Specifies the length of the
```

draft-ietf-moq-transport.md
```diff
@@ -296,6 +296,36 @@ Location A < Location B iff
...
 
+### Key-Value-Pair Structure
+
+Key-Value-Pair is a flexible structure designed to carry key/value
+pairs in which the key is a variable length integer and the value
+is either a variable length integer or a byte field of arbitrary
+length.
+
+Key-Value-Pair is used in both the data plane and control plane, but
+is optimized for use in the data plane.
```
suhasHere, line 307: I think this may be misleading to say its optimized for data plane. I wonder if we should remove that statement for now and resolve https://github.com/moq-wg/moq-transport/issues/794

afrind: I view the statement as the intention of the wg moving forward and we should leave it in place in this PR.

mzanaty:
LGTM. We can address even/odd separately in #794 or another issue.

mzanaty (approved):

draft-ietf-moq-transport.md
fluffy: I thought this would help address Mo point that people reading it later would not forget that from a design point of view what we trying to do.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
fluffy: I do not think we should be doing this type of backwards compatibility at this point. If you want to do something that support 10 and 11, use the ALPN to sort it out. I'd rather just note things and breaking changes.  This just makes the draft more confusing to people reading it and adds to cruft code.

draft-ietf-moq-transport.md
```diff
@@ -296,6 +296,36 @@ Location A < Location B iff
...
+  Type (i),
+  [Length (i),]
+  Value (..)
+}
+~~~
+{: #moq-key-value-pair format title="MOQT Key-Value-Pair"}
+
+* Type: an unsigned integer, encoded as a varint, identifying the
+  type of the value and also the subsequent serialization.
+* Length: Only present when Type is odd. Specifies the length of the
```
fluffy, line 320: I don't like the even / odd way to split this. I would rather see us defined a range of numbers that have no length and range that has a length. This allows us to adjust those two ranges to not be the same size. We have such a limited space for short codes that I don't want to waste them.

fluffy: My suggestions would 0 to 10, no length, 53 to 63 has length, and 11 to 52 reserved and we can sort out how to allocate that as we get spec gets closer to RFC and we understand the usage patters better.

draft-ietf-moq-transport.md
```diff
@@ -1208,6 +1219,13 @@ Publishers SHOULD consider whether the entire Object is likely to be delivered
 before sending any data for that Object, taking into account priorities,
 congestion control, and any other relevant information.
 
+#### AUTHORIZATION INFO {#authorization-info}
```
fluffy, line 1222: I don't think section should be in the same PR

fluffy (changes requested):

draft-ietf-moq-transport.md
afrind: We're not making a decision about even/odd in this PR, we are deciding to use the *same format* to serialize Parameters and Extensions.  Please add your comments to #794.

draft-ietf-moq-transport.md
afrind: AUTH INFO has to change it's code point in this PR (from 2 to 3).  Otherwise it's just moving text.  Do you want to keep AUTH INFO before delivery timeout?  Nothing else changed.

afrind:
Chairs have indicated this is ready to merge.

PR #768, opened 2025-03-08
make syntax consistent

suhasHere:
Since content exists applies to GroupId/ObjectId together, this PR makes syntax consistent


ianswett (approved):
Thanks for catching this!

PR #774, opened 2025-03-15
Allow multiple SUBSCRIBEs if they don't overlap

### LINKED ISSUE #649: Relays should never use EndGroup with a SUBSCRIBE
If a relay is subscribing on behalf of clients that used an EndGroup, it should ignore those EndGroups and make the subscribe open-ended.

This is because if a SUBSCRIBE is up to Group 3, and then another client asks for up to Group 4, the relay can't have multiple SUBSCRIBEs for one track and SUBSCRIBE_UPDATE can't extend the EndGroup. so the relay's only choice would be to UNSUBSCRIBE and then SUBSCRIBE again right away, which may involve a FETCH to cover gaps, etc.

At the very least, the relay section should provide this advice.

Or maybe we need to re-think the SUBSCRIBE_UPDATE rules?

### LINKED ISSUE #770: Not being able to have multiple subscriptions is a challenge for ABR
Currently, you can SUBSCRIBE_UPDATE, but only narrow the subscription.  This is to avoid raciness, since in the worst case you receive a few Objects that you no longer want, but otherwise the behavior is not racy.

Multiple subscriptions are not allowed, but there needs to be some way to say "Actually, I want one more group from this track" when doing ABR switches, because if the next quality up takes longer to arrive for whatever reason, you'll want to stay on the current track.  There are likely other use cases when this is very useful.

Cancelling a subscription and re-subscribing in sequence has even more raciness issues than trying to do SUBSCRIBE_UPDATE.

The easiest solution is to allow multiple subscriptions to the same track.  If we wanted to be restrictive, we could say that subscribers SHOULD not have multiple subscriptions that overlap.

ianswett:
Loosens a previous restriction.  Does not change the wire format.  If two subscriptions are using the same track alias, then given they don't overlap, the subscriber should know what Object is associated with each SUBSCRIBE.

Fixes #770 

I believe it fixes #649 as well?

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1335,7 +1335,7 @@ is blocked. More on Subscribe ID in {{message-subscribe-req}}.
 ## SUBSCRIBE {#message-subscribe-req}
 
 A subscription causes the publisher to send newly published objects for a track.
-A subscriber MUST NOT make multiple active subscriptions for a track within a
+A subscriber MUST NOT make overlapping active subscriptions for a track within a
 single session and publishers SHOULD treat this as a protocol violation.
 
 **Filter Types**
@@ -1557,9 +1557,9 @@ expected that it will always succeed and the worst outcome is that it is not
 processed promptly and some extra objects from the existing subscription are
 delivered.
 
-Unlike a new subscription, SUBSCRIBE_UPDATE can not cause an Object to be
-delivered multiple times.  Like SUBSCRIBE, EndGroup MUST specify the
-same or a later object than StartGroup and StartObject.
+SUBSCRIBE_UPDATE does not cause an Object to be delivered multiple times.
+Like SUBSCRIBE, EndGroup MUST specify the same or a later Object than
+StartGroup and StartObject.
 
 The format of SUBSCRIBE_UPDATE is as follows:
 


afrind:
Individual Comment:

It would be good to document the use case for this in an issue.

It's annoying that subscribers need to be aware of and apply filters now.  Previously this was entirely a publisher side construct, at least in my implementation.

ianswett:
I tried to document the use case in #770, and this would also enable an easy fix to #649

An alternative would be to allow SUBSCRIBE_UPDATE to increase the EndGroup, which could be racy and error-prone?

ianswett:
I'm going to park this until changes relating to replacing Track Alias with Subscriber ID are complete, because I think it'll turn out that allowing SUBSCRIBE_UPDATE to increase the EndGroup will end up being preferable, even if it's a bit racy.

fluffy (changes requested):
Agree with this but think we need to 

1) define overlapping 

2) make clear that objects MAY have the same track-alias even though they came in different subscriptions.

ianswett:
Closing this for now, since the group was supportive of widening subscriptions with SUBSCRIBE_UPDATE instead and adding a SUBSCRIBE_UPDATE_OK or equivalent message.

PR #793, opened 2025-03-19
Groups might not be published in ascending order

### LINKED ISSUE #777: When can a FETCH be served from Objects retrieved via SUBSCRIBE?
This was briefly discussed in Denver and I believe in some other issues, and is related to #715.

There are cases when Groups will arrive out of order, even if they were created in order.  One example I have in mind is an original publisher who wants to keep within a certain boundary of live, and may have to delay sending some Groups of a higher bitrate stream until later when it has more bandwidth.

Assuming there is no Delivery Timeout, the Groups will eventually be sent from the Original Publisher, but may be quite delayed.  For cases when one wants the VoD copy to have the highest fidelity possible (ie: a video call being recorded), this seems useful.

This example is quite a bit worse than the 'this group was slightly delayed due to packet loss' example we've discussed before, but they're instances of the same issue.

I think the answer is that you can't answer the Fetch entirely from cache if there are any gaps in the Group ID space, because you don't know if you just haven't received the Group yet unless an explicit Group Does Not Exist was published and cached.

A related concern is that all of these Fetches might need to go all the way back to the original publisher unless we're careful about Fetch/request coalescing.  If that does happen, does the Fetch hang indefinitely for the Groups the Original Publisher hasn't had the bandwidth to send?

ianswett:
Even if they're produced in ascending order.

Addresses half of #777
Clarifies #715

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -388,9 +388,8 @@ Order is Descending.  Due to network reordering and the partial reliability
 features of MoQT, Groups can always be received out of order.
 
 As a result, subscribers cannot infer the existence of a Group until an object in
-the Group is received.
-This can create gaps in the MoQ cache that MUST be filled by
-doing a Fetch upstream if a Fetch is received that requests those Groups.
+the Group is received. This can create gaps in the MoQ cache that MUST be filled
+by doing a Fetch upstream if a Fetch is received that requests those Groups.
 
 Subscribers to tracks which do not follow this requirement SHOULD NOT
 use range filters which span multiple groups in FETCH or SUBSCRIBE. SUBSCRIBE and
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -389,8 +389,8 @@ Group Order.
 
 As a result, Subscribers will not know the status of a Group within a SUBSCRIBE's
 Group range, even if there is no Delivery Timeout for the Subscription, until
-either the Group is received or a 'Group Does Not Exist' status is received
-for the Group.  This can create gaps in the MoQ cache that MUST be filled by
+either the Group is received or a 'Group Does Not Exist' status is received.
+This can create gaps in the MoQ cache that MUST be filled by
 doing a Fetch upstream if a Fetch is received that requests those Groups.
 
 Subscribers to tracks which do not follow this requirement SHOULD NOT
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -407,8 +407,8 @@ Order is Descending.  Due to network reordering and the partial reliability
 features of MoQT, Groups can always be received out of order.
 
 As a result, subscribers cannot infer the existence of a Group until an object in
-the Group is received. This can create gaps in the MoQ cache that MUST be filled
-by doing a Fetch upstream if a Fetch is received that requests those Groups.
+the Group is received. This can create gaps in a cache that can be filled
+by doing a Fetch upstream, if necessary.
 
 Subscribers to tracks which do not follow this requirement SHOULD NOT
 use range filters which span multiple groups in FETCH or SUBSCRIBE. SUBSCRIBE and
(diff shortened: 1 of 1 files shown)


suhasHere (approved):
this looks good and clarifies the gaps that can occur when dsc oder is selected.

On a diff note, we have to make a choice on if dsc order is really needed. Other than that,  this PR looks good. Thanks Ian

draft-ietf-moq-transport.md
```diff
@@ -379,8 +379,21 @@ subscriber that does not want to receive the entire track can opt to receive onl
...
+Within a track, the original publisher SHOULD publish Group IDs which increase
+with time. There are cases when even if the Groups are produced in increasing order,
+they might not be sent to the MoQ network or various Subscribers in the network
+in that order. One example is if the original publisher is publishing in descending
+Group Order.
+
+As a result, Subscribers will not know the status of a Group within a SUBSCRIBE's
+Group range, even if there is no Delivery Timeout for the Subscription, until
+either the Group is received or a 'Group Does Not Exist' status is received
+for the Group.  This can create gaps in the MoQ cache that MUST be filled by
```
ianswett, line 393:
```suggestion
either the Group is received or a 'Group Does Not Exist' status is received.
This can create gaps in the MoQ cache that MUST be filled by
```

draft-ietf-moq-transport.md
```diff
@@ -379,8 +379,21 @@ subscriber that does not want to receive the entire track can opt to receive onl
...
+in that order. One example is if the original publisher is publishing in descending
+Group Order.
+
+As a result, Subscribers will not know the status of a Group within a SUBSCRIBE's
+Group range, even if there is no Delivery Timeout for the Subscription, until
+either the Group is received or a 'Group Does Not Exist' status is received.
+This can create gaps in the MoQ cache that MUST be filled by
+doing a Fetch upstream if a Fetch is received that requests those Groups.
+
+Subscribers to tracks which do not follow this requirement SHOULD NOT
```
ianswett, line 396: I'm not sure we need this paragraph, or possibly it deserves a rewrite?

afrind: Maybe something like this:

Applications that cannot produce groups with IDs that increase with time are limited to the subset of MoQT that does not numerically compare group IDs.  Subscribers to these tracks SHOULD NOT use range filters which span multiple groups in FETCH or SUBSCRIBE.

ianswett: I took your suggestion with a minor rewrite.

draft-ietf-moq-transport.md
```diff
@@ -379,13 +379,28 @@ subscriber that does not want to receive the entire track can opt to receive onl
...
+
+As a result, subscribers cannot infer the existence of a Group until an object in
+the Group is received. This can create gaps in the MoQ cache that MUST be filled
+by doing a Fetch upstream if a Fetch is received that requests those Groups.
+
+Subscribers to tracks which do not follow this requirement SHOULD NOT
 use range filters which span multiple groups in FETCH or SUBSCRIBE. SUBSCRIBE and
 FETCH delivery use Group Order, so a FETCH cannot deliver Groups out of order
 and a subscription could have unexpected delivery order if Group IDs do not increase
 with time.
```
afrind, line 398: Maybe this should be merged with the paragraph below, the first sentence redundant now, and the second sentence doesn't make sense.

draft-ietf-moq-transport.md
```diff
@@ -379,13 +379,28 @@ subscriber that does not want to receive the entire track can opt to receive onl
...
+
+Within a track, the original publisher SHOULD publish Group IDs which increase
+with time. In some cases, Groups will be produced in increasing order, but sent
+to subscribers in a different order, for example when the subscription's Group
+Order is Descending.  Due to network reordering and the partial reliability
+features of MoQT, Groups can always be received out of order.
+
+As a result, subscribers cannot infer the existence of a Group until an object in
+the Group is received. This can create gaps in the MoQ cache that MUST be filled
+by doing a Fetch upstream if a Fetch is received that requests those Groups.
```
afrind, line 392:
```suggestion
the Group is received. This can create gaps in a cache that can be filled
by doing a Fetch upstream, if necessary.
```

PR #806, opened 2025-03-20
Don't propagate Objects or Messages that close the Session

### LINKED ISSUE #775: Don't propagate messages that close the session
This may be obvious, but I think it's worth calling out that it's important not to continue propagating any message that causes the session to be closed.

I realized this when reviewing #755

ianswett:
Fixes #775

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -322,7 +322,9 @@ underlying encoding, compression, any end-to-end encryption, or
 authentication. A relay MUST NOT combine, split, or otherwise modify object
 payloads.
 
-Objects within a group are ordered numerically by their Object ID.
+Objects within a Group are ordered numerically by their Object ID. The first
+Object in a Group has an Object ID of 0, and IDs increase sequentially for
+each Object within the Group.
 
 ## Subgroups {#model-subgroup}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1069,6 +1069,11 @@ Control messages have a length to make parsing easier, but no control
 messages are intended to be ignored. If the length does not match the
 length of the message content, the receiver MUST close the session.
 
+If the session is closed due to an unknown or invalid control message
+or Object, the receiver MUST NOT continue to propagate that message
+or Object, because it enables a single session to close unrelated
```
ianswett, line 1074:
```suggestion
or Object, because it would enable a single session to close unrelated
```

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -1069,6 +1069,11 @@ Control messages have a length to make parsing easier, but no control
 messages are intended to be ignored. If the length does not match the
 length of the message content, the receiver MUST close the session.
 
+If the session is closed due to an unknown or invalid control message
+or Object, the receiver MUST NOT continue to propagate that message
+or Object, because it would enable a single session to close unrelated
+sessions.
+
```
afrind, line 1076: I think this might read better in the relay section, and use "relay" rather than receiver.

afrind (approved):

Issue #712, opened 2025-02-25
Resource attacks via Track Name(space)


LPardue:
I didn't put much thought into this but it seems like the following aspects of the protocol design put it at risk of attacks:

* Track names can be up to 2^62 - 1 bytes long.
* Track namespace (prefix) tuple fields can each be up to 2^62 -1 bytes long
* Track namespace can hold up to 32 entries
* Track namespace prefix can hold up to 32 entries. 
* IIUC Track namespace prefix is used to match against N (what is limit?) namespaces (based on my reading of https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#section-7.13, where it states "f the publisher is a relay that has received ANNOUNCE messages for namespaces ("example.com", "meeting=123", "participant=100") and ("example.com", "meeting=123", "participant=200"), a SUBSCRIBE_ANNOUNCES for ("example.com", "meeting=123") would match both."

For example, I could simply announce N huge track namespaces of 32 * 2^62 - 1 to consume memory. Or I could send N namepsaces with carefully selected tuple fields, and then throw a worst case track namespace prefix to trigger CPU consumption on the receiver as it attempts to do matching. 

I tried looking for bounds on this data but didn't see any. I didn't see any security considerations. 

Do we _really_ need to support such humungous byte sequences? Other protocols (or their implementations) tend to apply finite limits to mitigate such problems. If we don't say anything in the document, we risk interoperability problems when implementations add their own hardening. For example, this sort of thing has affected HTTP headers for a long time and shows little sign of going away (see examples in https://www.rfc-editor.org/rfc/rfc9110.html#fields.limits and https://www.rfc-editor.org/rfc/rfc9114.html#section-7.2.4.1-2.2.1)

afrind:
See also #473

wilaw:
+1 for limiting the size of track Names and Namespace tuples. They are intended to be an addressing scheme. To that end, here are two options for rational upper limits:

1. The upper end of addressability might be to insert a GUID into each Namespace field and into the name. A GUID (Globally Unique Identifier) is typically represented as a 128-bit (16-byte) value, commonly formatted as a 36-character string, including hyphens. Encoding with UTF-8 would take 36 bytes, raw binary would be less as its just a 128-bit number. That gives us a max size of (32 + 1)*36 = 1188 bytes. 

[...]

PR #810, opened 2025-03-21
Set a max length for a Full Track Name

afrind:
Fixes: #712

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -408,6 +408,10 @@ Violation.
 Track Name is a sequence of bytes that identifies an individual track within the
 namespace.
 
+The maxmimum total length of a Full Track Name is 8,192 bytes.  If an endpoint
+receives a Full Track Name exceeding this length, it MUST close the session with
+a Protocol Violation.
+
 In this specification, both the Track Namespace tuple fields and the Track Name
 are not constrained to a specific encoding. They carry a sequence of bytes and
 comparison between two Track Namespace tuple fields or Track Names is done by


fluffy:
This seems very large to me. Any particular reason for this number ? I am concerned about DDOS.

wilaw:
> I'll note that URLs on the internet seem to work into the 2k ish range

That's 2k characters, which if you UTF-8 encode will have variable length:
```
1 byte (8 bits): For ASCII characters (U+0000 to U+007F)
2 bytes (16 bits): For characters in the range U+0080 to U+07FF
3 bytes (24 bits): For characters in the range U+0800 to U+FFFF (most common for many non-Latin scripts)
4 bytes (32 bits): For characters in the range U+10000 to U+10FFFF (used for emojis, rare symbols, and some Asian characters)
```

I'm OK with a lower limit, we could go to 4kB, which would give you approximately 2k characters with 2 byte encoding or 1.3k characters with 3 byte encoding.

The use-case for longer URLs is baking access tokens into the URL path.

draft-ietf-moq-transport.md
```diff
@@ -408,6 +408,10 @@ Violation.
 Track Name is a sequence of bytes that identifies an individual track within the
 namespace.
 
+The maxmimum total length of a Full Track Name is 8,192 bytes.  If an endpoint
```
martinduke, line 411:
```suggestion
The maximum total length of a Full Track Name is 8,192 bytes.  If an endpoint
```

martinduke, line 411: Is the "total length" the encoded length, or the sum of the length field of each element in the tuple?

ianswett (approved):
I suspect this is still a bit larger than we need, but not problematically so.

draft-ietf-moq-transport.md
```diff
@@ -408,6 +408,11 @@ Violation.
 Track Name is a sequence of bytes that identifies an individual track within the
 namespace.
 
+The maximum total length of a Full Track Name is 4,196 bytes, computed as the
```
vasilvv, line 411: That's a really weird number.  Did you mean 4096?

vasilvv (approved):

suhasHere (approved):

Sy0307:
Any limits of max length of track namespace? I searched but not found yet. Track namespace can be very large too?

Issue #554, opened 2024-09-28
Can we make the control message length field 16 bits instead of varint?


afrind:
It's very annoying to write a message serializer for the current messages.  Doing it in two-passes (one to compute the length, one to serialize) is one option.  Using some kind of buffer chaining is another.  But do we need to?  The control messages aren't large, so I don't think we need more than 64k for any single message.  They also aren't so frequent that we need to shave a length byte for 64 byte messages.  Using a fixed size length allows you to pin a placeholder, serialize the message, and write the actual length back.

afrind:
FWIW - right now I'm just adding 16 bits of space and always encoding a 2 byte varint, and restricting my control msgs to 14 bits.

ianswett:
We could, but I'm not sure there's an advantage over your approach of only using 14 bits?  For those who care about saving an extra byte, at least for some messages that are always short, they can use the 1 byte encoding.

Issue #473, opened 2024-06-24
Should there be size limits for messages and strings?


huitema:
The messages defined in MoQ Transport include variable length components:

* name space,
* track name,
* setup path,
* go away URI,
* authorization info,
* number of supported versions,
* number of setup or subscribe parameters.

These components have no specified maximum size, which opens an avenue for DOS attacks: start sending a message with components of arbitrary length, up to 2^62, and watch as the other endpoint allocates memory to hold the message. Application developers are very likely to mitigate the DOS attack by setting some limits, either to the global size of messages or to the size of individual components. We may expect different implementations to pick different limits, with a fairly good potential of causing interoperability issues.

Maybe we should specify some limits.

TimEvens:
+1 to having some limits. I think we should have some defaults in the standard and then negotiate those during client/server setup.

LPardue:
Guidance on limits is good.

However, limit negotiation doesn't tend to work well when intermediaries enter the mix. We've seen this in HTTP/2 and HTTP/3 with the max field section size parameter. If a proxy is given something from the origin that exceeds the client's limit, its not really advisable to silently drop it.

fluffy:
All implementations end up with limits and having them all be different in different implementations with no way to say they were exceeded is a nightmare. 

I would argue on limits for everything and sending protocol error and closing the connection if limits are exceeded.

huitema:
If limit negotiation is known to not work, then we should have the limits expressed in the base spec.

TimEvens:
Currently, I've mapped the termination reason (https://datatracker.ietf.org/doc/html/draft-ietf-moq-transport-04#section-3.5) as the application reason code (https://github.com/private-octopus/picoquic/blob/master/picoquic/picoquic.h#L780) when closing a connection. 

It's starting to look like this is not right. If we terminate by protocol volition (e.g., 0x3) then the remote side is left unaware of the specific protocol violation. Was it malformed, was it caused by invalid state machine, was it a timeout, was it a size that exceed, ??? If it was a size limit, then what size was the problem and is there suggested value to use? How would the remote side know what to do, other than repeatedly reconnect that results in only the same violation over and over again?

huitema:
The error code in the application-specific variant of CONNECTION_CLOSE" is defined as "codes defined by the application protocol". So it is completely legit to put the specific MoQ code it there. You may also want to use a reason phrase if you want to facilitate debugging.

afrind:
[...]

PR #811, opened 2025-03-21
Limit all the things

afrind:
Created explicit limits for all variable length things in MoQT.

1. Control Messages have a max length of 2^16-1, changed the Length from varint to uint16.
2. Changed number of parameters from varint to uint8, so a max of 256 params per control message.
3. Changed number of supported versions from varint to uint8, so a max of 256 versions to offer.
4. Set reason phrase max to 1024 bytes.
5. Set max length of all extensions in streams and fetch to 2^16-1 (datagrams are limited to MTU, obvs).

Any length violation => close the session.

I'm open to different limit values, particularly for control messages if there's a compelling argument, though I'll point out that 64k is about the maximum size of an HTTP header block you can get across the internet.

Fixes: #473
Fixes: #554

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1002,11 +1002,13 @@ formatted as follows:
 ~~~
 MOQT Control Message {
   Message Type (i),
-  Message Length (i),
+  Message Length (16),
   Message Payload (..),
 }
 ~~~
 {: #moq-transport-message-format title="MOQT Message"}
+
+The following Message Types are defined:
 
 |-------|-----------------------------------------------------|
 | ID    | Messages                                            |
@@ -1101,7 +1103,9 @@ SETUP message parameter types are defined in {{setup-params}}. Version-
 specific parameter types are defined in {{version-specific-params}}.
 
 The Parameter Length field encodes the length of the Parameter Value field in
-bytes.
+bytes. The maximum length of a parameter is 2^16-1 bytes.  If an endpoint
+receives a length larger than the maximum, it MUST close the session with a
+Protocol Violation.
 
 Each parameter description will indicate the data type in the Parameter Value
 field. If a receiver understands a parameter type, and the parameter length
@@ -1182,7 +1186,7 @@ The wire format of the Setup messages are as follows:
 CLIENT_SETUP Message {
   Type (i) = 0x40,
   Length (i),
-  Number of Supported Versions (i),
+  Number of Supported Versions (8),
   Supported Version (i) ...,
   Number of Parameters (i) ...,
   Setup Parameters (..) ...,
@@ -1207,7 +1211,7 @@ This version of the specification is identified by the number 0x00000001.
 Versions with the most significant 16 bits of the version number cleared are
 reserved for use in future IETF consensus documents.
 
-The client offers the list of the protocol versions it supports; the
+The client offers a list of up to 256 protocol versions it supports; the
 server MUST reply with one of the versions offered by the client. If the
 server does not support any of the versions offered by the client, or
 the client receives a server version that it did not offer, the
@@ -1273,7 +1277,9 @@ GOAWAY Message {
   connect to continue this session.  The client MUST use this URI for the new
   session if provided. If the URI is zero bytes long, the client can reuse the
   current URI is reused instead. The new session URI SHOULD use the same scheme
-  as the current URL to ensure compatibility.
+  as the current URL to ensure compatibility.  The maxmimum length of the New
+  Session URI is 8,192 bytes.  If an endpoint receives a length exceeding the
+  maximum, it MUST close the session with a Protocol Violation.
 
   If a server receives a GOAWAY with a non-zero New Session URI Length it MUST
   terminate the session with a Protocol Violation.
@@ -1386,7 +1392,7 @@ SUBSCRIBE Message {
   [StartGroup (i),
    StartObject (i)],
   [EndGroup (i)],
-  Number of Parameters (i),
+  Number of Parameters (8),
   Subscribe Parameters (..) ...
 }
 ~~~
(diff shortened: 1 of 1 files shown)


LPardue (approved):
Lgtm

Land it and we can always change limits later down the line

fluffy (commented):
My preference would be to havre al the limits be in one section near the end which just listed them all in one section. Otherwise we end up with them repeating multiple time in spec. 

Changing the encoding should be a bit separate form the limits. I want to think very carefully about anytime we are making things bigger by going from (i) to (16). We can still have (i) and a limit of 5000. 

If we have a bunch of PR that break things on the wire. I would rather group all thoses PR together.

afrind:
> My preference would be to havre al the limits be in one section near the end which just listed them all in one section. Otherwise we end up with them repeating multiple time in spec.

I think I like having them where the fields are defined even if there's some redundancy.  This makes it easier for implementers since the value is right there, and (maybe) makes it less likely to forget something in a bottom section.  Paging co-editor @ianswett.

> Changing the encoding should be a bit separate form the limits.

That's fair - I was trying to be simple, for example, there's no need for text about what happens when control messages are > 64k, since you literally can't send one.  I also think there's an implementation hazard when sending lengths that an implementation might pass directly to vector.resize(), so having a byte there gives me a warm fuzzy (eg: num params or num versions).

> I want to think very carefully about anytime we are making things bigger by going from (i) to (16). 

I only did this for control messages, which should be pretty infrequent.  If you are sending enough control messages to care about this byte, you likely have other problems.

> If we have a bunch of PR that break things on the wire. I would rather group all thoses PR together.

At this point every draft has been wire breaking, except eg: 9/10.  Once we stabilize a bit more, that makes sense.

fluffy:
On where to put them in the doc ... clearly I don't really care and glad with whatever gets decided on that

TimEvens:
Does message type still need to be varint? Now that control is moving to fixed 2 byte length, if message type was also changed to either 1 byte or 2 bytes, then we would be able to have predictable control start headers (aka common header for control message is type(8), length(16)...

draft-ietf-moq-transport.md
```diff
@@ -1002,12 +1002,14 @@ formatted as follows:
 ~~~
 MOQT Control Message {
   Message Type (i),
```
suhasHere, line 1004:
```suggestion
  Message Type (8),
```

draft-ietf-moq-transport.md
```diff
@@ -1182,7 +1186,7 @@ The wire format of the Setup messages are as follows:
 CLIENT_SETUP Message {
   Type (i) = 0x40,
   Length (i),
-  Number of Supported Versions (i),
+  Number of Supported Versions (8),
```
suhasHere, line 1189: thought . if we had it varint, it will allow 62 versions and I am unable to see a case where 256 versions will need to be supported .

My preference is to keep it varint for now and revisit if this needs to be more flexible

TimEvens: Why not uint8 with range of ..6

ianswett: I think the plan is to move to alpn style version negotiation, so this is transitional.

ianswett (commented):
I'd be inclined to not bother with changing the version negotiation bits, since I think they're being replaced by ALPN anyway.

I'm not sure if 255 parameters is enough, but I'll let others comment.

draft-ietf-moq-transport.md
```diff
@@ -1572,7 +1580,7 @@ SUBSCRIBE_UPDATE Message {
   StartObject (i),
   EndGroup (i),
   Subscriber Priority (8),
-  Number of Parameters (i),
+  Number of Parameters (8),
```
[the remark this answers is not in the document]
afrind: Since we only have like 4 parameter types right now I'm not that worried.  I thought about also leaving this as a varint, since the PR restricts control messages to 64k anyways, but it's a bit of a hazard that some implementations might pass this straight to vector.resize().  I suppose I can limit it to 32k, since every param must be at least two bytes.

fluffy:
Gratuitous comment that I think the correct max length for the reason phrase is 0 :-)  Any extra than that and people use it to avoid correct useful error codes. But I think we have a separate issue for that.

draft-ietf-moq-transport.md
gloinul: This appears an unnecessary limiting as it is in the setup message that is exchanged once per session. Even if I agree that 256 simultanous versions being supported should go likely cover all cases.

draft-ietf-moq-transport.md
```diff
@@ -1451,7 +1457,7 @@ SUBSCRIBE Message {
   Filter Type (i),
   [Start (Location)],
   [EndGroup (i)],
-  Number of Parameters (i),
+  Number of Parameters (8),
```
gloinul, line 1460: Although unlikely to be a limit in reality, this is one place where a limit could create issues. But, I think it is hard to forsee if this might not run into future problems. The number of parameters is not what is going to be the issue, it is the total size of all parameters I would guess.

gloinul (commented):
I think this all looks ok except for where I commented. I think the number of parameters needs one more round of thinking. And restriciting things in setup appears to be the wrong place.

draft-ietf-moq-transport.md
afrind: Let's handle changing message type in a separate issue or PR, since it impacts extensibility, rather than implementation limits.

draft-ietf-moq-transport.md
afrind: I removed this change, since we are moving version negotiation anyways.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
afrind: I removed the restriction to the number of parameters and put this back to varint.  I suppose (16) would have been fine too.

ianswett (approved):

Issue #481, opened 2024-07-08
Error codes for streams being reset


vasilvv:
We currently do not define error codes with which streams are supposed to be reset.

I can think of at least four of those:

1. GENERIC -- for unknown reasons
2. SUBSCRIPTION_GONE -- the subscription associated with the stream has been terminated.
3. TIMEOUT -- for stream timeouts (possibly defer until we have a PR ready for that).
4. STREAM_LIMIT_EXCEEDED -- the associated subscription has too many streams associated with it, so we discarded the lowest priority one (probably should be a part of #462).

fluffy:
+1 

The only thing I would suggest is git rid of GENERIC and make error codes for whatever needs one.

TimEvens:
We need to discuss stream transition and if we use FIN or RESET for that. Stream transition is when we are moving to a new group.  Maybe we have an error code for transition to new stream, but right now we use zero for the app code reason.  This was an interop issue as some folks assumed, because the draft-05 doesn't indicate otherwise, that upon transition to a new group you close by FIN. That doesn't work because stuff gets stuck inflight with FIN.  RESET allows us to immediately move to a new group, which might be to specifically mitigate some head of line blocking issue caused by say retransmissions, ... 

Regardless, the above reads as informational.  We need to indicate that it's not just informational. If you receive a timeout, shouldn't you clean up the state of the subscription?

vasilvv:
> [...] upon transition to a new group you close by FIN. That doesn't work because stuff gets stuck inflight with FIN.

It is possible to reset a stream that has already been closed with a FIN (in fact, WebTransport WG went out of its way to support that).  That's why the per-stream timeouts (like the ones defined in #449) should be kept around even after FIN is sent.

>  If you receive a timeout, shouldn't you clean up the state of the subscription?

The TIMEOUT I was referencing above is timeout for a single group/object, not necessarily the track as a whole.

afrind:
Individual Comment:

> If you receive a timeout, shouldn't you clean up the state of the subscription

If I squint I think Victor was suggesting the TIMEOUT code would be sent on a stream where you started sending an object but then you passed the delivery timeout (which is in the process of being defined)?  I don't think resetting any particular stream should affect subscription state, except perhaps stream-per-track implicitly?

Also, is there a different set of codes we need for STOP_SENDING?

TimEvens:
>  The TIMEOUT I was referencing above is timeout for a single group/object, not necessarily the track as a whole.

[...]

PR #812, opened 2025-03-21
Add error codes for resetting data streams

afrind:
I selfishly pulled the list of codes I use in moxygen, open to add more now if someone has a suggestion.

Fixes: #481

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2638,7 +2638,9 @@ Internal Error:
 Cancelled:
 
 : The subscriber requested cancellation via UNSUBSCRIBE, FETCH_CANCEL or
-STOP_SENDING
+STOP_SENDING, or the publisher ended the subscription, in which case
+SUBSCRIBE_DONE ({{message-subscribe-done}}) will have a more detailed
+status code.
 
 Delivery Timeout:
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2637,7 +2637,7 @@ Internal Error:
 
 Cancelled:
 
-: The receiver requested cancellation via UNSUBSCRIBE, FETCH_CANCEL or
+: The subscriber requested cancellation via UNSUBSCRIBE, FETCH_CANCEL or
 STOP_SENDING
 
 Delivery Timeout:
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2616,7 +2616,7 @@ RESET_STREAM and STOP_SENDING on SUBSCRIBE data streams have no impact on other
 Subgroups in the Group or the subscription, although applications might cancel all
 Subgroups in a Group at once.
 
-The application SHOULD use a relevant error code in RESET_STREAM or 
+The application SHOULD use a relevant error code in RESET_STREAM or
 RESET_STREAM_AT, as defined below:
 
 |------|---------------------------|
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2616,6 +2616,39 @@ RESET_STREAM and STOP_SENDING on SUBSCRIBE data streams have no impact on other
 Subgroups in the Group or the subscription, although applications might cancel all
 Subgroups in a Group at once.
 
+The application SHOULD use a relevant error code in RESET_STREAM or 
+RESET_STREAM_AT, as defined below:
+
+|------|---------------------------|
+| Code | Reason                    |
+|-----:|:--------------------------|
+| 0x0  | Internal Error            |
+|------|---------------------------|
+| 0x1  | Cancelled                 |
+|------|---------------------------|
+| 0x2  | Delivery Timeout          |
+|------|---------------------------|
+| 0x3  | Session Closed            |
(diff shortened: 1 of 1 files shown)


ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -2616,6 +2616,39 @@ RESET_STREAM and STOP_SENDING on SUBSCRIBE data streams have no impact on other
...
+|------|---------------------------|
+
+Internal Error:
+
+: An implementation specific error
+
+Cancelled:
+
+: The receiver requested cancellation via UNSUBSCRIBE, FETCH_CANCEL or
+STOP_SENDING
```
vasilvv, line 2641: Are there cases in which the publisher can unilaterally close subscription?  Because in those cases, we'd need to close streams too, and "cancelled" does not quite fit in well

afrind: @vasilvv : CANCELLED seems fine to me in this case - what is the receiver going to do differently with a different code.  That said, tell me what you want and we can add it.  PUBLISHER_CANCELLED and SUBSCRIBER_CANCELLED?

vasilvv: It's actually a bit of the opposite case: I want a catch-all "subscription is gone so I am removing this data stream", since it does add complexity to track the reason why it was gone.

draft-ietf-moq-transport.md
```diff
@@ -2616,6 +2616,39 @@ RESET_STREAM and STOP_SENDING on SUBSCRIBE data streams have no impact on other
...
+| 0x3  | Session Closed            |
+|------|---------------------------|
+
+Internal Error:
+
+: An implementation specific error
+
+Cancelled:
+
+: The receiver requested cancellation via UNSUBSCRIBE, FETCH_CANCEL or
```
afrind, line 2640:
```suggestion
: The subscriber requested cancellation via UNSUBSCRIBE, FETCH_CANCEL or
```

draft-ietf-moq-transport.md
afrind: @suhasHere : I think you mean subscriber.  eg: this is what the publisher includes in the stream reset if the subscriber sends UNSUBSCRIBE or FETCH_CANCEL on the control stream or STOP_SENDING on a DATA stream.

draft-ietf-moq-transport.md
```diff
@@ -2616,6 +2616,41 @@ RESET_STREAM and STOP_SENDING on SUBSCRIBE data streams have no impact on other
...
+status code.
+
+Delivery Timeout:
+
+: The DELIVERY TIMEOUT {{delivery-timeout}} was exceeded for this
+stream
+
+Session Closed:
+
+: The publisher session is being closed
```
suhasHere, line 2652: wondering, in what circumstances Cancelled and/or Session Closed will be sent by the publisher

afrind: For Cancelled, it's in the text - either the subscriber asked for it to be cancelled, or the publisher is ending the subscription for some other reason, specified in SUBSCRIBE_DONE.

> The subscriber requested cancellation via UNSUBSCRIBE, FETCH_CANCEL or
STOP_SENDING, or the publisher ended the subscription, in which case
SUBSCRIBE_DONE ({{message-subscribe-done}}) will have a more detailed
status code.

For Session Closed, my implementation sends this when the MoQ session is closed with open streams.  Might happen during shutdown or some other error condition is detected (eg: Protocol Violation on control stream).

fluffy:
I think we need some discussion around this and I think there are a bunch of bigger design questions around this. 

If there is an internal error, I think we need to close the connection not the stream or we can lead to significant security problem. 

It is not clear to me which things should reset vs close a stream. I'd like to see a bit more on where we do each one.

100% support error codes like this to show why a reset happened but the issue here is what things cause reset vs somthing else.

afrind:
@fluffy 

> If there is an internal error, I think we need to close the connection not the stream 

Internal error is really there as a catch-all and future proofing -- I know i want to kill this stream and I need to put *some* value.  We can bikeshed the name but without this, people will just make up their own reset codes and that doesn't really help anyone.  

One example in my implementation - if the publisher throws an exception while handling a FETCH request, but has already opened the fetch data stream, I reset it with INTERNAL_ERROR.

Most other examples happen when a publisher violates a rule when publishing - eg: tries to send an object ID smaller than the previous one on a stream.  I could be amenable to ending the _subscription_ if this happens, but I really don't want to kill the _session_, which can be shared by multiple subscriptions.

> It is not clear to me which things should reset vs close a stream.

This has a whole section in the current document (https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#name-closing-subgroup-streams).  FIN on a subgroup stream has a specific meaning: "I know I sent you the last object in this subgroup".  If you want to get rid of a stream for _any other reason_ - you have to use reset.

> I think we need some discussion around this and I think there are a bunch of bigger design questions around this.
> ... the issue here is what things cause reset vs somthing else.

Please file issues for what you think needs discussion if they are not already filed. This PR is just defining _some_ error codes and they are all based on a real implementation.

vasilvv (approved):

Issue #569, opened 2024-09-30
Handling of unknown/invalid/or otherwise bad PATHs


LPardue:
There should be more text about how to do error-handling of values sent in the PATH setup parameter. For example, if a client sends "moq://example.com/foobar" (because it copied the entire URI, not just the `path-abempty` part) the server clearly needs to reject with an error. What should that be? A protocol violation and so a CONNECTION_CLOSE?

I'd expect there to be similarities, where possible, to how HTTP would handle bad paths.

PR #814, opened 2025-03-21
Close on bad PATH

afrind:
Creates Invalid Path and Malformed Path error codes.

Fixes: #569

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -566,6 +566,8 @@ code, as defined below:
 |------|---------------------------|
 | 0x6  | Too Many Subscribes       |
 |------|---------------------------|
+| 0x7  | Invalid Path              |
+|------|---------------------------|
 | 0x10 | GOAWAY Timeout            |
 |------|---------------------------|
 | 0x11 | Control Message Timeout   |
@@ -1230,14 +1232,15 @@ The PATH parameter (Parameter Type 0x01) allows the client to specify the path
 of the MoQ URI when using native QUIC ({{QUIC}}).  It MUST NOT be used by
 the server, or when WebTransport is used.  If the peer receives a PATH
 parameter from the server, or when WebTransport is used, it MUST close
-the connection. It follows the URI formatting rules {{!RFC3986}}.
+the session with Invalid Path. It follows the URI formatting rules
+{{!RFC3986}}.
 
 When connecting to a server using a URI with the "moqt" scheme, the
 client MUST set the PATH parameter to the `path-abempty` portion of the
 URI; if `query` is present, the client MUST concatenate `?`, followed by
 the `query` portion of the URI to the parameter.  If the server receives an
-improperly formatted PATH parameter, it MUST close the session with a Protocol
-Violation.
+improperly formatted PATH parameter, it MUST close the session with Invalid
+Path.
 
 #### MAX_SUBSCRIBE_ID {#max-subscribe-id}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1235,7 +1235,9 @@ the connection. It follows the URI formatting rules {{!RFC3986}}.
 When connecting to a server using a URI with the "moqt" scheme, the
 client MUST set the PATH parameter to the `path-abempty` portion of the
 URI; if `query` is present, the client MUST concatenate `?`, followed by
-the `query` portion of the URI to the parameter.
+the `query` portion of the URI to the parameter.  If the server receives an
+improperly formatted PATH parameter, it MUST close the session with a Protocol
+Violation.
```
fluffy, line 1240: I would prefer we got an error that indicated the path was bad, not just a generic error.

fluffy (changes requested):

draft-ietf-moq-transport.md
```diff
@@ -1230,12 +1232,15 @@ The PATH parameter (Parameter Type 0x01) allows the client to specify the path
 of the MoQ URI when using native QUIC ({{QUIC}}).  It MUST NOT be used by
 the server, or when WebTransport is used.  If the peer receives a PATH
 parameter from the server, or when WebTransport is used, it MUST close
-the connection. It follows the URI formatting rules {{!RFC3986}}.
+the session with Invalid Path. It follows the URI formatting rules
+{{!RFC3986}}.
```
ianswett, line 1236:
```suggestion
the session with Invalid Path. The PATH parameter follows the URI formatting
rules {{!RFC3986}}.
```

ianswett (approved):

afrind:
Cullen's comments are addressed so bypassing -1

vasilvv (approved):

vasilvv (changes requested):
Actually, reading closer, I am not sure this addresses #569.

It talks about improperly formatted paths, but doesn't define what it is (nor I believe there is a universally accepted definition of what an invalid URL is, which is bad for a normative MUST), nor it talks about valid URLs for which server has no mapping associated.

fluffy (approved):
LGTM

draft-ietf-moq-transport.md
```diff
@@ -1264,12 +1266,15 @@ The PATH parameter (Parameter Type 0x01) allows the client to specify the path
 of the MoQ URI when using native QUIC ({{QUIC}}).  It MUST NOT be used by
 the server, or when WebTransport is used.  If the peer receives a PATH
 parameter from the server, or when WebTransport is used, it MUST close
```
suhasHere, line 1268:
```suggestion
parameter from the server, or when WebTransport is used, or the the one server doesn't support, it MUST close
```

suhasHere (approved):
minor nit. Feel free to ignore it as you see fit.

draft-ietf-moq-transport.md
afrind: @suhasHere I think you are suggesting that the endpoint return error if it receives a path that is properly formatted but not the one the server wanted.  Maybe we want a different error code for that?  Invalid Path is sort of like HTTP 400 (bad request) and maybe it's useful to differentiate 404 (no such endpoint)?

vasilvv (approved):

draft-ietf-moq-transport.md
```diff
@@ -627,6 +627,10 @@ code, as defined below:
 |------|---------------------------|
 | 0x7  | Too Many Subscribes       |
 |------|---------------------------|
+| 0x8  | Invalid Path              |
+|------|---------------------------|
+| 0x9  | Malformed Path            |
```
afrind, line 632: Can you add an english description below the table with the other codes?

Issue #659, opened 2025-01-30
How does priority work for FETCH streams?


martinduke:
Section 5 is clearly written to just consider SUBSCRIBE. It suggests that prioritization between tracks with the same pub and sub priorities is  meaningless. Since individual groups and object IDs don't map to FETCH streams, that seems fine.

FETCH contains a subscriber priority, but each object can (in theory) have its own publisher priority. So are FETCH streams only ordered by subscriber priority or is there some f(object publisher priorities)?

I imagine we don't want to go back to the days of Max(all outstanding objects).

martinduke:
One solution would be to use the subscriber priority, and have all other fields assume lowest-priority

wilaw:
A publisher priority makes sense for the live head of tracks. Once an object falls away from that, into the realm where it can be cached, delivered faster than its encoded bitrate and retrieved via FETCH, then I think the notion of a publisher priority lacks meaning. It is no longer a live flow of data, it is the read-response from a relay cache query, in which case the SUBSCRIBER priority makes a lot more sense for prioritization. We should add normative text that the original PUBLISHER priority is ignored when prioritizing between FETCHes to the same subscriber.

fluffy:
FWIW ... I don't recall us discussing this and coming to this conclusion. What there discussion at some point ?

That said it seems OK to me.

PR #816, opened 2025-03-21
Include FETCH responses in priority scheduling algorithm

afrind:
Fixes: #659

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -800,6 +800,10 @@ the objects SHOULD be selected as follows:
    with delivery preference Subgroup), or **the lowest Object ID** (for tracks
    with delivery preference Datagram) is sent first.
 
+When objects are delivered via FETCH, only the subscriber priority is
+considered.  If a FETCH response stream has the same subscriber priority as
+another schedulable object, the order is implementation defined.
+
 This algorithm does not provide a well-defined ordering for objects that belong
 to different subscriptions, but have the same subscriber and publisher
 priority.  The ordering in those cases is implementation-defined, though the
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -802,9 +802,9 @@ the objects SHOULD be selected as follows:
    with delivery preference Subgroup), or **the lowest Object ID** (for tracks
    with delivery preference Datagram) is sent first.
 
-When objects are delivered via FETCH, only the subscriber priority is
-considered.  If a FETCH response stream has the same subscriber priority as
-another schedulable object, the order is implementation defined.
+When Objects are delivered via FETCH, only the subscriber priority is
+considered. If a FETCH response has the same subscriber priority as
+another schedulable Object, the order is implementation specific.
 
 This algorithm does not provide a well-defined ordering for objects that belong
 to different subscriptions, but have the same subscriber and publisher
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -753,29 +753,34 @@ congestion.
 MoQT maintains priorities between different _schedulable objects_.
 A schedulable object in MoQT is either:
 
-1. An object that belongs to a subgroup where that object would be the next
-   object to be sent in that subgroup.
-1. An object that belongs to a track with delivery preference Datagram.
+1. An object in response to a SUBSCRIBE that belongs to a subgroup where
+   that object would be the next object to be sent in that subgroup.
+1. An object in response to a SUBSCRIBE that belongs to a track with
+   delivery preference Datagram.
+1. An object in response to a FETCH where that object would be the next
+   object in the response.
 
-Since a single subgroup or datagram has a single publisher priority, it can be
-useful to conceptualize this process as scheduling subgroups or datagrams
-instead of individual objects on them.
+A single subgroup or datagram has a single publisher priority. Within a
+response to SUBSCRIBE, it can be useful to conceptualize this process as
+scheduling subgroups or datagrams instead of individual objects on them.
+FETCH responses however can contain objects with different publisher
+priorities.
 
 A _priority number_ is an unsigned integer with a value between 0 and 255.
 A lower priority number indicates higher priority; the highest priority is 0.
 
 _Subscriber Priority_ is a priority number associated with an individual
-subscription.  It is specified in the SUBSCRIBE message, and can be updated via
-SUBSCRIBE_UPDATE message.  The subscriber priority of an individual schedulable
-object is the subscriber priority of the subscription that caused that object
-to be sent. When subscriber priority is changed, a best effort SHOULD be made
-to apply the change to all objects that have not been sent, but it is
+request.  It is specified in the SUBSCRIBE or FETCH message, and can be
+updated via SUBSCRIBE_UPDATE message.  The subscriber priority of an individual
+schedulable object is the subscriber priority of the request that caused that
+object to be sent. When subscriber priority is changed, a best effort SHOULD be
+made to apply the change to all objects that have not been sent, but it is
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -802,6 +802,10 @@ the objects SHOULD be selected as follows:
    with delivery preference Subgroup), or **the lowest Object ID** (for tracks
    with delivery preference Datagram) is sent first.
 
+When objects are delivered via FETCH, only the subscriber priority is
+considered.  If a FETCH response stream has the same subscriber priority as
+another schedulable object, the order is implementation defined.
```
ianswett, line 807:
```suggestion
When Objects are delivered via FETCH, only the subscriber priority is
considered. If a FETCH response has the same subscriber priority as
another schedulable Object, the transmission order is implementation
specific.
```

ianswett (approved):
LG, though I made a suggestion.

draft-ietf-moq-transport.md
afrind: Happy to change this, though the next paragraph has this:

> This algorithm does not provide a well-defined ordering for objects that belong to different subscriptions, but have the same subscriber and publisher priority.  The ordering in those cases is implementation-defined,

afrind: @ianswett do you want me to take your suggestion in light of the subsequent paragraph?

martinduke (approved):

draft-ietf-moq-transport.md
ianswett: I like the wording "implementation specific" better than implementation defined, but it's a nit for sure.

My suggestion of adding "transmission" is probably not helpful given how the rest is worded, so I'll remove that change.

fluffy (changes requested):
This seems like a bad idea. It does not deal with issue like how a proxy that aggregated two downstream fetches with different priorities turn into one upstream fetch.  It makes me have different code for priorities depending on what requested them. 

Whats the argument for doing this ? I think I could be convinced either way but seems like this needs to at least explain how aggregation and such would work and why we would do this.

afrind:
> Whats the argument for doing this?

A FETCH response can have a multitude of different publisher priorities in each object transmitted and the draft doesn't explain how to use that information, if at all.  Given the difficulties in reprioritizing the response FETCH stream as objects are transmitted / retransmited, leaving this to implementations seemed like the most straightforward.  Do you have a different suggestion?

> It does not deal with issue like how a proxy that aggregated two downstream fetches with different priorities turn into one upstream fetch.

Great point, though I don't immediately have a recommendation.  For subscribe, we recommend relays set the same subscriber priority and default to the publisher, but it's really not clear what the right answer is for FETCH.  Since FETCH is in the past, the Publisher priority in each object when it was published might no longer be appropriate to prioritize against ongoing SUBSCRIBE In the same session.  Open to suggestions.

draft-ietf-moq-transport.md
```diff
@@ -802,6 +802,10 @@ the objects SHOULD be selected as follows:
    with delivery preference Subgroup), or **the lowest Object ID** (for tracks
    with delivery preference Datagram) is sent first.
 
+When Objects are delivered via FETCH, only the subscriber priority is
+considered. If a FETCH response has the same subscriber priority as
+another schedulable Object, the order is implementation specific.
```
sharmafb, line 807: Do we want to make the publisher priority optional in the subgroup header, since it's irrelevant for fetches?

afrind: I think priority is well baked into subgroups - though if we get deep into bit packing the data plane, I can imagine trying to get clever.

sharmafb (approved):

martinduke (approved):

afrind:
Based on the discussion at the 4/16 interim, the biggest concern was less about tracks that have multiple priorities, but respecting publisher priority during fetch response when all objects have the same publisher priority.  Therefore we should continue to take publisher priority into account for FETCH response.

afrind:
The content of this PR has changed entirely, please re-review.

draft-ietf-moq-transport.md
```diff
@@ -753,29 +753,34 @@ congestion.
...
 
-1. An object that belongs to a subgroup where that object would be the next
-   object to be sent in that subgroup.
-1. An object that belongs to a track with delivery preference Datagram.
-
-Since a single subgroup or datagram has a single publisher priority, it can be
-useful to conceptualize this process as scheduling subgroups or datagrams
-instead of individual objects on them.
+1. An object in response to a SUBSCRIBE that belongs to a subgroup where
+   that object would be the next object to be sent in that subgroup.
```
ianswett, line 757:
```suggestion
   that object is the next object in that subgroup.
```

draft-ietf-moq-transport.md
```diff
@@ -753,29 +753,34 @@ congestion.
...
+1. An object in response to a SUBSCRIBE that belongs to a track with
+   delivery preference Datagram.
+1. An object in response to a FETCH where that object would be the next
+   object in the response.
```
ianswett, line 761:
```suggestion
2. An object in response to a SUBSCRIBE that belongs to a track with
   delivery preference Datagram.
3. An object in response to a FETCH where that object is the next
   object in the Track.
```

draft-ietf-moq-transport.md
```diff
@@ -793,19 +798,25 @@ the objects SHOULD be selected as follows:
    the one with **the highest subscriber priority** is sent first.
 1. If two objects have the same subscriber priority, but a different publisher
    priority, the one with **the highest publisher priority** is sent first.
-1. If two objects have the same subscriber and publisher priority, but belong
-   to two different groups of the same track,
+1. If two objects in response to the same request have the same subscriber
```
ianswett, line 801:
```suggestion
2. If two objects in response to the same request have the same subscriber
```

draft-ietf-moq-transport.md
```diff
@@ -793,19 +798,25 @@ the objects SHOULD be selected as follows:
...
+   and publisher priority, but belong to two different groups of the same track,
    **the group order** of the associated subscription is used to
    decide the one that is sent first.
-1. If two objects belong to the same group of the same track,
-   the one with **the lowest Subgroup ID** (for tracks
+1. If two objects in response to the same request belong to the same group of
```
ianswett, line 805:
```suggestion
3. If two objects in response to the same request belong to the same group of
```

ianswett (approved):
Some editorial suggestions, but LGTM

draft-ietf-moq-transport.md
afrind: Track is ambiguous for FETCH, since there can be > 1 FETCH for the same track at the same time.

martinduke (approved):
New version LGTM. I have no opinion on the underlying approaches, but this description seems clear, and the implementability issues are fairly clear in this text.

gloinul (approved):

englishm (approved):

draft-ietf-moq-transport.md
```diff
@@ -885,19 +890,25 @@ the objects SHOULD be selected as follows:
...
-1. If two objects have the same subscriber and publisher priority, but belong
-   to two different groups of the same track,
+2. If two objects in response to the same request have the same subscriber
+   and publisher priority, but belong to two different groups of the same track,
    **the group order** of the associated subscription is used to
    decide the one that is sent first.
-1. If two objects belong to the same group of the same track,
-   the one with **the lowest Subgroup ID** (for tracks
+3. If two objects in response to the same request belong to the same group of
+   the same track, the one with **the lowest Subgroup ID** (for tracks
```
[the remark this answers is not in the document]
afrind: I recognize those two overlap but I don't think it reads too awkwardly so I'm going to leave it

Issue #664, opened 2025-02-03
What are default parameters for SUBSCRIBE_UPDATE?


martinduke:
If a parameter is not present in SUBSCRIBE_UPDATE, does that imply the value in the original SUBSCRIBE, or the protocol default?

suhasHere:
Since subscribe update is updating a subscription, if any parameters are omitted, they match the original subscription request.  If it helps, we can add a statement to mean this in the draft

martinduke:
yes, that would help.

afrind:
Individual Comment:

How do you remove a parameter that was set (eg: subscribe contained DELIVERY_TIMEOUT but I want to remove it)?

PR #817, opened 2025-03-21
Default param values for SUBSCRIBE_UPDATE

afrind:
Fixes: #664

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1561,8 +1561,8 @@ Unlike a new subscription, SUBSCRIBE_UPDATE can not cause an Object to be
 delivered multiple times.  Like SUBSCRIBE, EndGroup MUST specify the
 same or a later object than StartGroup and StartObject.
 
-If a parameter was included in SUBSCRIBE and is not present in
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -1561,6 +1561,10 @@ Unlike a new subscription, SUBSCRIBE_UPDATE can not cause an Object to be
 delivered multiple times.  Like SUBSCRIBE, EndGroup MUST specify the
 same or a later object than StartGroup and StartObject.
 
+If a parameter was included in SUBSCRIBE and is not present in
```
suhasHere, line 1564:
```suggestion
If a parameter is included in `SUBSCRIBE` but not in `SUBSCRIBE_UPDATE`, its value remains unchanged. Parameters cannot be removed from a subscription.```

Comment (suhasHere) [inline draft-ietf-moq-transport.md:1564]:
```suggestion
If a parameter is included in `SUBSCRIBE` but not in `SUBSCRIBE_UPDATE`, its value 
remains unchanged. Parameters cannot be removed from a subscription.
```

ianswett (approved):

Issue #674, opened 2025-02-16
Error code tables need the equivalent of HTTP 400


afrind:
We need a code that indicates the request failed because it was malformed (eg: duplicate subscribe ID)

fluffy:
agree and I would prefer errors that were specific about how it was malformed

afrind:
I checked all my error conditions in moxygen, and the only one not covered by the existing draft and #822 is when a client subscribes itself via a relay it has announced to, but I may open a separate issue for that.

PR #822, opened 2025-03-22
Add error codes for Duplicate and Invalid Subscribe ID

afrind:
Duplicate Subscribe ID is a session error, to mirror Duplicate Track Alias.

Invalid Subscribe ID is a FETCH_ERROR (not a session error) for Joining fetch if the dependent ID does not exist.  Since the client pipelines Subscribe+Joining Fetch, we don't want to close the session if the Subscribe fails and invalidates the ID.

Fixes: #674

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1812,7 +1812,7 @@ FETCH Message {
    StartObject (i),
    EndGroup (i),
    EndObject (i),]
-  [ Subscribe ID (i),
+  [ Joining Subscribe ID (i),
     Joining Start (i),]
   Number of Parameters (i),
   Parameters (..) ...
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -562,11 +562,13 @@ code, as defined below:
 |------|---------------------------|
 | 0x3  | Protocol Violation        |
 |------|---------------------------|
-| 0x4  | Duplicate Track Alias     |
+| 0x4  | Duplicate Subscribe ID    |
 |------|---------------------------|
-| 0x5  | Parameter Length Mismatch |
+| 0x5  | Duplicate Track Alias     |
 |------|---------------------------|
-| 0x6  | Too Many Subscribes       |
+| 0x6  | Parameter Length Mismatch |
+|------|---------------------------|
+| 0x7  | Too Many Subscribes       |
 |------|---------------------------|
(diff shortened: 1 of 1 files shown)


ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -2005,6 +2016,9 @@ as defined below:
 * No Objects - The beginning of the requested range is after the latest group
   and object for the track, or the track has not published any objects.
 
+* Invalid Subscribe ID - The joining Fetch referenced a Subscribe ID that did
+  not belong to an active Subscription.
```
suhasHere, line 2020: Design question and I am not sure of the answer yet. I can open an issue if we think it is useful to discuss further.

If a subscriber uses 2 QUIC sessions to the same relay and in one session he does Subscribe and in another session he does Joining Fetch. How do we define such behavior ? 

I suspect we allow such behavior for Subscribe and Standalone fetch ?

afrind: Joining fetch is defined to only work in the same session.  I thought this language was in the original PR but seems to not be there.  I filed #868

afrind: > I suspect we allow such behavior for Subscribe and Standalone fetch ?

This works because the standalone fetch uses a Full Track Name, so you can SUBSCRIBE in one session and FETCH in another.

vasilvv (approved):

draft-ietf-moq-transport.md
```diff
@@ -1817,7 +1824,8 @@ Fields common to all Fetch Types:
 
 * Subscribe ID: The Subscribe ID identifies a given fetch request. Subscribe ID
 is a variable length integer that MUST be unique and monotonically increasing
-within a session.
+within a session.  If an endpoint receives a FETCH with a Subscribe ID that is
+already in use, it MUST close the session with Duplicate Subscribe ID.
```
suhasHere, line 1828:
```suggestion
already in use, it MUST close the session with Duplicate Subscribe ID (unless the operation is Joining Fetch)
```

suhasHere (approved):
one minor nit. Please see if it makes sense.

draft-ietf-moq-transport.md
afrind: Ah, the problem here is that both fields are called Subscribe ID.  I fixed that in a different PR somewhere but let me fix it here too.

Issue #716, opened 2025-02-25
Serving FETCH from partial cache is risky


martinduke:
We specifically allow objects before FETCH_OK to allow early service from cache.

4.3 says A SUBSCRIBE_ERROR or FETCH_ERROR indicates no objects will be delivered, and both endpoints can immediately destroy relevant state. Objects MUST NOT be sent for requests that end with an error.

Say a relay has observed objects 1 and 3 and cached them. It then receives a FETCH for 1 to 3. It sends Object 1 and tries to FETCH and receives FETCH_ERROR. What then? Does it Reset the data stream?

afrind:
Individual Comment:

I filed a similar / related issue: https://github.com/moq-wg/moq-transport/issues/729

I'm hoping with joining fetch we can remove the object fields from FETCH_OK so a cache can send it immediately.

CC: fetch enthusiast @englishm

fluffy:
I don't see a problem with removing the "Largest Group ID" and "Largest Object ID" from FETCH_OK. Apps can use track status to get that. 

I think any time something is sending a FETCH data and can no longer fine the next object, resetting the stream sounds like the right thing to do.

fluffy:
I'm not sure there is anything "risky" about this, we just need to say what happens

afrind:
Individual Comment:

> I don't see a problem with removing the "Largest Group ID" and "Largest Object ID" from FETCH_OK. Apps can use track status to get that.

I remembered why these are needed.  All gaps between the last object delivered in fetch to the end of the range are "NOT_EXISTS".  If the live head is *before* the end of the range, we need the FETCH_OK values so we don't mark future objects NOT_EXISTS by mistake.

I think that means for tracks not currently live at the relay, one can send the FETCH_OK immediately if the end is smaller than the largest cached object.  However if the end is {10,10} and the largest cached is {8,7}, the relay must delay the FETCH_OK until it gets an upstream FETCH that covers {10,10}, or updates the live head to something between {8,7} and {10,10}.

If I start serving objects from my cache from the beginning of the range while the upstream fetch is happening, I may get a FETCH_ERROR from upstream, but the current rules prohibit sending either FETCH_OK or FETCH_ERROR.

I can see a couple ways out:

1) allow FETCH_ERROR to come after we start serving FETCH objects, it's weird but not sure there's harm?
2) remove the values from FETCH_OK and add a new Object Status like "Largest Known" that can be the last object in a Fetch response

PR #823, opened 2025-03-22
Allow FETCH_ERROR after sending objects

afrind:
Fixes: #716

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1714,10 +1714,12 @@ SUBSCRIBE_DONE, as defined below:
 
 ## FETCH {#message-fetch}
 
-A subscriber issues a FETCH to a publisher to request a range of already published
-objects within a track. The publisher responding to a FETCH is responsible for retrieving
-all available Objects. If there are gaps between Objects, the publisher omits them from the
-fetch response. All omitted objects have status Object Does Not Exist.
+A subscriber issues a FETCH to a publisher to request a range of already
+published objects within a track. The publisher responding to a FETCH is
+responsible for delivering all available Objects in the requested range in the
+requested order. The Objects in the response are delivered on a single
+unidirectional stream. Any Objects not included in the response do not exist
+(eg: they implicitly have status `Object Does Not Exist`).
 
 **Fetch Types**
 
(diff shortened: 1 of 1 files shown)


ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1740,10 +1745,11 @@ Type Latest Object.
 A Fetch Type other than 0x1, 0x2 or 0x3 MUST be treated as an error.
 
 A publisher responds to a FETCH request with either a FETCH_OK or a FETCH_ERROR
-message.  If it responds with FETCH_OK, the publisher creates a new unidirectional
-stream that is used to send the Objects.  A relay MAY start sending objects immediately
-in response to a FETCH, even if sending the FETCH_OK takes longer because it requires
-going upstream to populate the latest object.
+message.  The publisher creates a new unidirectional stream that is used to send the
+Objects.  A relay MAY start sending objects immediately in response to a FETCH, even
```
martinduke, line 1749: A few things here:

1) It's worth pointing out that the FETCH stream must be in ID order, so it has to pause on a gap.

2) Should we specify actions on upstream FETCH error? What if a relay has 1 to 10 in cache, and sends 1-5 when it gets FETCH_ERROR. Can it send up to 2, or Reset immediately, potentially mid-object?

3) Given that caches will generally contain later data instead of earlier, I question the utility of all this.

afrind: 1. Sure, we can add that text
2. I tried to leave it vague about timing, but I can make it explicit this is the relay's choice - eg: it can send 1-5, then FETCH_ERROR, then RESET_STREAM_AT if it wants, or it can just RESET_STREAM as soon as it gets the error.
3. Having implemented this cache, the most sensible algorithm dictates it works this way -- you keep sending the in-order objects from your cache until you get a miss.  It's actually harder to find the miss and go upstream before sending anything.

draft-ietf-moq-transport.md
```diff
@@ -1746,10 +1748,16 @@ A Fetch Type other than 0x1, 0x2 or 0x3 MUST be treated as an error.
...
-FETCH_ERROR and can reset the unidirectional stream.
+Objects.  The FETCH_OK or FETCH_ERROR can come at any time relative to object
+delivery.
+
+A relay that has cached objects from the beginning of the range MAY start
+sending objects immediately in response to a FETCH.  If it encounters an object
+in the requested range that is not cached and has unknown status, the relay MUST
+pause subsequent delivery until it has confirmed the object's status upstream.
+If the upstream FETCH fails, the relay sends a FETCH_ERROR and can reset the
+unidirectional stream.  It can chose to do so immediately or wait until the
```
martinduke, line 1759:
```suggestion
unidirectional stream.  It can choose to do so immediately or wait until the
```

martinduke (approved):

draft-ietf-moq-transport.md
```diff
@@ -1714,10 +1714,12 @@ SUBSCRIBE_DONE, as defined below:
...
-A subscriber issues a FETCH to a publisher to request a range of already published
-objects within a track. The publisher responding to a FETCH is responsible for retrieving
-all available Objects. If there are gaps between Objects, the publisher omits them from the
-fetch response. All omitted objects have status Object Does Not Exist.
+A subscriber issues a FETCH to a publisher to request a range of already
+published objects within a track. The publisher responding to a FETCH is
+responsible for delivering all available Objects in the requested range in the
+requested order. The Objects in the response are delivered on a single
+unidirectional stream. Any Objects not included in the response do not exist
+(eg: they implicitly have status `Object Does Not Exist`).
```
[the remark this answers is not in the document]
afrind: Added some text for Ascending order but describing how to find the Descending gaps makes my head hurt.

englishm (approved):

vasilvv (approved):

Issue #729, opened 2025-02-26
How to populate Largest Group/Object in FETCH_OK


afrind:
The text says it 'largest available' and I recall from our discussion in Boston that these values were supposed to be set to the live head.  But what happens when the track is not live, and not yet ended either?  My first implementation of a fetch cache sends FETCH_OK

1) immediately if the track has a active upstream subscription
2) after the first gap-filling upstream FETCH returns with a new value of largest
3) after I've served the entire range from cache, which is weird.

Am I doing it wrong?  It seems like it would be better to get the FETCH_OK fired off as soon as possible.  Maybe these values in FETCH_OK are obsolete with joining FETCH?

If I delay FETCH_OK, it introduces a hazard where I've sent cached objects but hit an error, and sort of have to send an FETCH_ERROR, but this is prohibited here:

> Objects MUST NOT be sent for requests that end with an error.

PR #824, opened 2025-03-22
Rename FETCH_OK "largest" group/object to "end" and explain

afrind:
Fixes: #729

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1870,7 +1870,7 @@ Joining Start value.
 
 A publisher sends a FETCH_OK control message in response to successful fetches.
 A publisher MAY send Objects in response to a FETCH before the FETCH_OK message is sent,
-but the FETCH_OK MUST NOT be sent until the latest group and object are known.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1937,13 +1937,15 @@ Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
...
+the End Group ID and Object Id indicate the last Object in the track,
 0 if not.
 
-* Largest Group ID: The largest Group ID available for this track.
-
-* Largest Object ID: The largest Object ID available within the largest Group ID
-for this track.
+* End Group ID/End Object ID: The largest object covered by the FETCH response.
+  This is the minimum of the {EndGroup,EndObject} specified in FETCH and the
+  largest known {group,object}.  If the track is currently live at a relay,
```
martinduke, line 1945: Is "live" a defined term? I guess a relay would do this if it has an upstream SUBSCRIBE for which it has received an object or SUBSCRIBE_OK.

martinduke (approved):

Issue #728, opened 2025-02-26
Resolving FETCH for distributed publishers of the same track


afrind:
If not all publishers who ANNOUNCE are authoritative for the entire track, relays need to send FETCH to all publishers.  However, that creates problems:

1. No single upstream FETCH response is authoritative.  The upstream responses must be merged to correctly populate the cache and produce a correct downstream response.
2. If one of the publishers is temporarily disconnected from the relay when a FETCH arrives, groups it produced will be marked NOT_EXISTS, and the response is incorrect.

Saying applications with distributed publishers won't use FETCH is problematic
1.  This is the same as saying "Don't use a cache", since only FETCH retrieves from a cache
2. Joining Fetch is a potentially common operation
3. Nothing stops a malicious/buggy client from sending a FETCH

It might be better to prohibit distributed publishing on a single track.

There are many degrees of freedom currently in MoQT - is there an example of an application that cannot replace distributed publishing on the same track with each publisher publishing a single track (perhaps in a common namespace), or another MoQT feature (eg: SUBSCRIBE_ANNOUNCES)?

suhasHere:
> If not all publishers who ANNOUNCE are authoritative for the entire track,
 IIUC all announces needs to be authorized regardless of it comes from one publisher or multiple publisher ..

afrind:
> IIUC all announces needs to be authorized regardless of it comes from one publisher or multiple publisher ..

They are all authorized.  I meant eg "publisher A is authoritative for even groups and publisher B is authoritative for odd groups"

fluffy:
I think the premise that "If not all publishers who ANNOUNCE are authoritative for the entire track" is off, every original publisher is authoritative for the whole track. Any one of them could provide an any group/object on the track and it would be accepted.

afrind:
Individual Comment: 

>  every original publisher is authoritative for the whole track

I'm ok with allowing distributed publishing if this is true, but it's not what was being discussed in the room in Denver - eg: one publisher only being authoritative for a single group, and no knowledge of other groups.

suhasHere:
Scoping Authorization at track level for publishing works best from managing authz and token when writing to a track ( announces ).

OTOH a subscriber authorization can be filtered via start_group and other filters.

fluffy:
[...]

PR #825, opened 2025-03-22
Add FETCH_ERROR for Unknown Status in Range

afrind:
This is my proposal to handle cases where the Original Publisher receives a FETCH for a range that it can't satisfy because it has no knowledge of one or more objects requested.

Addresses part of #728

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1714,6 +1714,9 @@ objects within a track. The publisher responding to a FETCH is responsible for r
 all available Objects. If there are gaps between Objects, the publisher omits them from the
 fetch response. All omitted objects have status Object Does Not Exist.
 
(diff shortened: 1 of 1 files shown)


ianswett (commented):
Failing the entire FETCH seems overly restrictive.  What about indicating the range of Groups/Objects that are not yet available?

afrind:
> Failing the entire FETCH seems overly restrictive. What about indicating the range of Groups/Objects that are not yet available?

I agree that would be more practical, but didn't want to update FETCH_OK with a list of ranges.  Do you have a concrete proposal here?

fluffy (approved):

afrind:
Unparking this per Stockholm discussion.  See #728 for more details, but we think we want this error code in any case.

Issue #745, opened 2025-02-28
Clarify interaction of extension headers and caching


TimEvens:
Related to #741.... 

The topic has come up that we need to cache fetch streams of data objects in addition to the live stream of data objects based on subscribes. 

Our original thought was that a cache is populated upon subscribed objects being received... Now that fetches may exist outside of a subscribe...  A new requirement has come up that we cache fetched data objects.   This introduces some complexities on reconciling cache within the relay (as well as any replication or distributed cache).

The unique key constraint of `<track namespace array/tuples>, <track name>, <group-id>, <object-id>` being used by both live subscribed data and fetch response data may differ in extension headers. 

It's assumed that the relay would service the fetch request from local cache if it has the data...  but that may not be the case if the range is out side of the local cache store . Fetch would then possibly fill those gaps of data with objects, but now the extension headers are possibly different because the data objects are from a fetch and not live stream of data.  For example, extension headers may have updated headers based on the fetch request, such as timestamps to be relative to the time of the fetch, etc...

If the fetch request is for a range of 200 - 300 group-ids and the local cache only has 250-300, the missing gaps are fetched upstream toward the original publisher to get 200-249.. Those objets returned by the fetch now may have headers that don't jive with the headers 250-300. The next fetch request comes in for the same range of 200-300, which now comes from local cache, but the headers for 200-249 are in conflict with the original headers from the live stream that was cached for 250-300. 

How is this supposed to be handled?

afrind:
I think the same condition can occur without FETCH --eg: if the relay has redundant publishers for the same track, which are following different relay paths and therefore different headers.  I'm retitling the issue to address the generic relationship between headers and caching.

fluffy:
agree needs clarification but I think the solution would be very easy to just say when an object is written to cache, then the headers that were with it are also written to cache. And when an object is read from cache, so are the headers with it. If a relay receives a 2nd copy of a object that is already in cache, the new copy is discarded and cache remains the same.

suhasHere:
I will be in support for caching header extensions and those getting served when retrieved from the cache.  It might be more useful to say relays SHOULD drop if there is a duplicate entry for a object key.

TimEvens:
[...]

Issue #741, opened 2025-02-27
Computing Object Uniqueness to support Caching of Objects


TimEvens:
How are we to compute object uniqueness of objects when considering extensions headers being appended or mutated inflight? Object uniqueness as defined in https://datatracker.ietf.org/doc/html/draft-ietf-moq-transport-08#name-objects 

> [Section 2.3](https://datatracker.ietf.org/doc/html/draft-ietf-moq-transport-08#model-group) An object is uniquely identified by its track namespace, track name, group ID, and object ID, and must be an identical sequence of bytes regardless of how or where it is retrieved. An Object can become unavailable, but its contents MUST NOT change over time.

If a relay decides to add extension headers, such as relay ID, region id (geo info), trace Id, trace of relay hops, telemetry, ... the extension headers will be relative to the path and relay that the subscriber received the object. In this sense, some of the extension headers are relative to point in time transmission in forwarding.  

If we replay the data, such as suggested by https://datatracker.ietf.org/doc/html/draft-jennings-moq-file-00, we would contaminate the extension header metadata. 

If caching uniqueness is computed based on extension header + payload, the object of same `group-id` and `object-id` will appear to be different, resulting in the same track alias, group-id, and object-id being cached more than once. Is that what we want? 

If caching is only based on group-id/object-id, then regardless of extension headers and payload, the data will appear to be duplicated, resulting in discards. Is that what we want? 

Is `track-alias` and `group-id` and `object-id` only used to determine uniqueness, ignoring extension headers and content that can be mutated hop-by-hop and have point in time relavance? If so, a replay of a cached file, such as via  https://datatracker.ietf.org/doc/html/draft-jennings-moq-file-00, would not result in an update to cache causing replay data to be discarded due to same data already seen.  This would have the net effect of NO or partial replay The updated extension headers with maybe new application timing, would not take effect.  For example, if extension headers include application real-time timestamps, those values in stored cache/files would become stale and would need to be updated on replay.  If replay results in discards because of duplicate group-id/object-id, the updated timestamps would never take effect. 

Careful though is needed in caching as we could result in caches on different relays being completely different based on extension header mutations. 

Maybe extension headers are non-transitive and are not cached.  Much of this complexity goes away if we do that.  The problem is that folks are using extension headers as application payload headers, assuming that they will be transitive.

afrind:
Individual Comment:

[...]

Issue #615, opened 2024-11-14
Support wildcard/pattern subscribes


TimEvens:
There are many use-cases where we would like to subscribe to a namespace tuple prefix to receive existing and new tracks that match that subscribe prefix.  In these use-cases, the publisher knows the available tracks and does not need to rely on the subscribe full track information to choose which track to send. Instead, there is a controlled set of tracks that the publisher would send based on the subscribe match. 

> Example:

Subscribe to namespace tuple `<meeting id>/participants/primary audio/*` , name is null. 

The relay would: 
1. lookup existing published tracks that matches the subscribe, such as `1234/participants/primary audio/endpointId=456`  and start forwarding to the subscriber this content. The track alias would reflect the full track name. The subscriber would be notified via control channel of what the full track name is for the given track alias.  
2. Upon new announcement, relay would forward to publisher the subscribe prefix requested. The publisher would then choose which tracks to transmit. The publisher would send to the relay (aka subscriber) the track full name information via the control channel with a track alias along with data objects that include the track alias. The original subscriber would receive the new track alias via data objects and associate it to the control channel full name information that was sent upon initial match.

afrind:
There was a previous issue #253 for "wildcard subscribe" that got closed when SUBSCRIBE_ANNOUNCES was landed #498.  Perhaps the difference is that you expect only a single track to be forwarded per SUBSCRIBE?

Your proposal conflicts with the current design where the subscriber is in control of the Track Alias space - if the publisher or relay chooses one, it could be taken by an in-flight SUBSCRIBE.  It also suffers from the HoL blocking problem where the objects with an unknown alias arrive at the subscriber before the control message linking them.

I'm curious what you mean by "controlled tracks" - are these fixed or will the publisher choose one of several based on some application logic.

afrind:
Individual Comment:

> announces only contain namespace, sans name

This is true, though there's nothing stopping an application from defining a namespace that is sufficiently unique and either a blank or otherwise fixed track name.  See [MoQ Chat](https://github.com/afrind/draft-frindell-moq-chat/blob/main/draft-frindell-moq-chat.md).  Would this meet your use case?

> Exactly, which is mentioned above.... it's a fundamental problem with MoQT design that has subscribers defining the track alias. IMO, it needs to change.

[Chair Hat on]: 

[...]

PR #826, opened 2025-03-23
Explain interaction of caching, status and extensions

afrind:
This is the simplest proposal, but doesn't cover the case of an extension that was added by the Original Publisher that was  added/modified/removed by a relay illegally (eg: LoC metadata).  To support that, the extension mechanism needs a way to identify immutable extensions.

Also added a normative statement about forwarding the same object more than once.

Fixes: #741 
Fixes: #745

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -858,7 +858,7 @@ or MAY use them to update the cache. Implementations that update the
 cache need to protect against cache poisoning.  The only fields that can be
 updated are the following:
 
-1. Object Status can transition from any status to Object Not Exists in cases
+1. Object Status can transition from any status to Object Does Not Exist in cases
    where the object is no longer available.  Transitions between Normal, End
    of Group and End of Track are invalid.
 3. Object Header Extensions can be added, removed or updated, subject
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -861,7 +861,7 @@ updated are the following:
 1. Object Status can transition from any status to Object Not Exists in cases
    where the object is no longer available
 2. Object Header Extensions can be added, removed or updated, subject
-to the constraints of the specific header extension.
+   to the constraints of the specific header extension.
 
 An endpoint that receives a duplicate object with an illegal Object Status
 change, or a Forwarding Preference, Subgroup ID, Priority or Payload that
@@ -891,7 +891,7 @@ subscriber, dependent on the congestion response.
 
 Relays MUST be able to process objects for the same Full Track Name from
 multiple publishers and forward objects to active matching subscriptions.  The
-same object SHOULD NOT be forwarded more than once to the same subscriber.
+same object SHOULD NOT be forwarded more than once on the same subscription.
 
 A relay MUST NOT reorder or drop objects received on a multi-object stream when
 forwarding to subscribers, unless it has application specific information.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -862,8 +862,9 @@ updated are the following:
    where the object is no longer available
 2. Object Header Extensions can be added, removed or updated
 
-An endpoint that receives a duplicate object with fields that differ from a
-previous version other than these MUST close the session with a Protocol
+An endpoint that receives a duplicate object with an illegal Object Status
+change, or a Forwarding Preference, Subgroup ID, Priority or Payload that
+differ from a previous version MUST close the session with a Protocol
 Violation.
 
 ## Subscriber Interactions
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -859,14 +859,20 @@ cache need to protect against cache poisoning.  The only fields that can be
 updated are the following:
 
 1. Object Status can transition from any status to Object Not Exists in cases
-   where the object is no longer available
-2. Object Header Extensions can be added, removed or updated, subject
+   where the object is no longer available.  Transitions between Normal, End
+   of Group and End of Track are invalid.
+3. Object Header Extensions can be added, removed or updated, subject
    to the constraints of the specific header extension.
 
-An endpoint that receives a duplicate object with an illegal Object Status
+An endpoint that receives a duplicate Object with an invalid Object Status
 change, or a Forwarding Preference, Subgroup ID, Priority or Payload that
-differ from a previous version MUST close the session with a Protocol
-Violation.
+differ from a previous version MUST treat the track as Malformed.
+
+Note that due to reordering, an implementation can receive a duplicate Object
+with a status of Normal, End of Group or End of Track after receiving a
+previous status of Object Not Exists.  The endpoint SHOULD NOT cache or
+forward the duplicate object in this case.
+
 
 ## Subscriber Interactions
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -846,7 +846,25 @@ validating subscribe and publish requests at the edge of a network.
 Relays are endpoints, which means they terminate Transport Sessions in order to
 have visibility of MoQ Object metadata.
 
+## Caching Relays
+
 Relays MAY cache Objects, but are not required to.
+
+A caching relay saves Objects to its cache identified by the Object's
+Full Track Name, Group ID and Object ID.
+If multiple objects are received with the same Full Track Name,
+Group ID and Object ID, Relays MAY ignore subsequently received Objects
+or MAY use them to update the cache. Implementations that update the
+cache need to protect against cache poisoning.  The only fields that can be
+updated are the following:
+
+1. Object Status can transition from any status to Object Not Exists in cases
+   where the object is no longer available
+2. Object Header Extensions can be added, removed or updated
+
+An endpoint that receives a duplicate object with fields that differ from a
+previous version other than these MUST close the session with a Protocol
+Violation.
 
 ## Subscriber Interactions
 
@@ -869,14 +887,9 @@ subscribers for each track. Each new OBJECT belonging to the
 track within the subscription range is forwarded to each active
 subscriber, dependent on the congestion response.
 
-A caching relay saves Objects to its cache identified by the Object's
-Full Track Name, Group ID and Object ID. Relays MUST be able to
-process objects for the same Full Track Name from multiple
-publishers and forward objects to active matching subscriptions.
-If multiple objects are received with the same Full Track Name,
-Group ID and Object ID, Relays MAY ignore subsequently received Objects
-or MAY use them to update the cache. Implementations that update the
-cache need to protect against cache poisoning.
+Relays MUST be able to process objects for the same Full Track Name from
+multiple publishers and forward objects to active matching subscriptions.  The
+same object SHOULD NOT be forwarded more than once to the same subscriber.
 
 A relay MUST NOT reorder or drop objects received on a multi-object stream when
 forwarding to subscribers, unless it has application specific information.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -846,8 +846,26 @@ validating subscribe and publish requests at the edge of a network.
...
+Group ID and Object ID, Relays MAY ignore subsequently received Objects
+or MAY use them to update the cache. Implementations that update the
+cache need to protect against cache poisoning.  The only fields that can be
+updated are the following:
+
+1. Object Status can transition from any status to Object Not Exists in cases
+   where the object is no longer available
+2. Object Header Extensions can be added, removed or updated
+
+An endpoint that receives a duplicate object with fields that differ from a
```
ianswett, line 865: If I understand correctly, we're really saying that the content of an Object can never change, correct?  If so, it might be simpler to say that.

ianswett (approved):

draft-ietf-moq-transport.md
afrind: The status can also not change, say from Normal to End of Group.  Technically the subgroup and priority can't change either per this language.  I'll add it all explicitly.

draft-ietf-moq-transport.md
```diff
@@ -846,8 +846,27 @@ validating subscribe and publish requests at the edge of a network.
 Relays are endpoints, which means they terminate Transport Sessions in order to
 have visibility of MoQ Object metadata.
 
+## Caching Relays
+
 Relays MAY cache Objects, but are not required to.
 
+A caching relay saves Objects to its cache identified by the Object's
+Full Track Name, Group ID and Object ID.
```
TimEvens, line 854: To support multi-publisher, we need publisher ID as well.

draft-ietf-moq-transport.md
```diff
@@ -846,8 +846,27 @@ validating subscribe and publish requests at the edge of a network.
...
+If multiple objects are received with the same Full Track Name,
+Group ID and Object ID, Relays MAY ignore subsequently received Objects
+or MAY use them to update the cache. Implementations that update the
+cache need to protect against cache poisoning.  The only fields that can be
+updated are the following:
+
+1. Object Status can transition from any status to Object Not Exists in cases
+   where the object is no longer available
+2. Object Header Extensions can be added, removed or updated
```
TimEvens, line 863: Headers, specifically extension headers, are problem as some extension headers may be specific relay by relay (e.g., Relay 5 added headers that relay 14 never sees). Caching is a bit tricky in this sense where mutable headers are not the same relay by relay. In a distributed deep cache, the headers may not exist at all due to which relay populated the deep cache. When pulling from a deep cache (distributed, not directly on the relay), the fetch may not match what would have been there had the relay originally received the data based on subscribe. Are relays to update the extension headers upon fetch response?

draft-ietf-moq-transport.md
```diff
@@ -869,14 +888,9 @@ subscribers for each track. Each new OBJECT belonging to the
...
 
-A caching relay saves Objects to its cache identified by the Object's
-Full Track Name, Group ID and Object ID. Relays MUST be able to
-process objects for the same Full Track Name from multiple
-publishers and forward objects to active matching subscriptions.
-If multiple objects are received with the same Full Track Name,
-Group ID and Object ID, Relays MAY ignore subsequently received Objects
-or MAY use them to update the cache. Implementations that update the
-cache need to protect against cache poisoning.
+Relays MUST be able to process objects for the same Full Track Name from
```
TimEvens, line 891: publisher ID is needed to support multiple publishers to the same namespace tuple/name.

draft-ietf-moq-transport.md
```diff
@@ -846,8 +846,27 @@ validating subscribe and publish requests at the edge of a network.
...
+cache need to protect against cache poisoning.  The only fields that can be
+updated are the following:
+
+1. Object Status can transition from any status to Object Not Exists in cases
+   where the object is no longer available
+2. Object Header Extensions can be added, removed or updated
+
+An endpoint that receives a duplicate object with an illegal Object Status
+change, or a Forwarding Preference, Subgroup ID, Priority or Payload that
+differ from a previous version MUST close the session with a Protocol
```
TimEvens, line 867: At some point cache invalidation (aka purge) will be needed.  We'll need to be able to from a publisher indicate, maybe in status, that object(s) should be purged.  

In addition, time to live needs to be taken into account. What is the retention policy and how will a subscriber/fetcher know that cache will never have requested data because they are asking for data outside of the retention policy?

draft-ietf-moq-transport.md
TimEvens: Correct, one or more publishers index their group/object IDs independently without any global synchronization. This can result in publisher 1 and 2 overlapping or appearing out of order, unless we distinguish that there are two publishers with their own group/object IDs.

afrind: If you want to live on the edge with multiple publishers publishing to the same track in an uncoordinated fashion, you still need to abide but the referenced MUST.  The best option is to embed your "Publisher ID" in the track namespace or name.  I imagine with forthcoming changes addressing #615 will make that easier.

draft-ietf-moq-transport.md
suhasHere: I will not be in favor adding any text that enforces a namespace design/structure within the core transport. An application can choose to have publisherId as a tuple or use one track to have multiple publisher or have publisher come acoss multiple connections (says make before break).

draft-ietf-moq-transport.md
```diff
@@ -869,14 +888,9 @@ subscribers for each track. Each new OBJECT belonging to the
...
-Full Track Name, Group ID and Object ID. Relays MUST be able to
-process objects for the same Full Track Name from multiple
-publishers and forward objects to active matching subscriptions.
-If multiple objects are received with the same Full Track Name,
-Group ID and Object ID, Relays MAY ignore subsequently received Objects
-or MAY use them to update the cache. Implementations that update the
-cache need to protect against cache poisoning.
+Relays MUST be able to process objects for the same Full Track Name from
+multiple publishers and forward objects to active matching subscriptions.  The
+same object SHOULD NOT be forwarded more than once to the same subscriber.
```
ianswett, line 893: I think we want to say for the same subscription?  Because a relay might get an Object once via SUBSCRIBE and later via FETCH?

draft-ietf-moq-transport.md
ianswett, line 863:
```suggestion
2. Object Header Extensions can be added, removed or updated, subject
to the constraints of the specific header extension.
```

draft-ietf-moq-transport.md
```diff
@@ -846,8 +846,27 @@ validating subscribe and publish requests at the edge of a network.
...
+or MAY use them to update the cache. Implementations that update the
+cache need to protect against cache poisoning.  The only fields that can be
+updated are the following:
+
+1. Object Status can transition from any status to Object Not Exists in cases
+   where the object is no longer available
+2. Object Header Extensions can be added, removed or updated
+
+An endpoint that receives a duplicate object with an illegal Object Status
+change, or a Forwarding Preference, Subgroup ID, Priority or Payload that
```
ianswett, line 866: Are we sure we can guarantee that an illegal Object Status change can't happen in a distributed network?  Also, I'm not sure it's that valuable to make the transitions illegal?  It seems the most important bit is what I can cache. (ie: negative caching)

ianswett (commented):
LG, but some comments and suggestions

draft-ietf-moq-transport.md
```diff
@@ -869,14 +889,9 @@ subscribers for each track. Each new OBJECT belonging to the
...
-Full Track Name, Group ID and Object ID. Relays MUST be able to
-process objects for the same Full Track Name from multiple
-publishers and forward objects to active matching subscriptions.
-If multiple objects are received with the same Full Track Name,
-Group ID and Object ID, Relays MAY ignore subsequently received Objects
-or MAY use them to update the cache. Implementations that update the
-cache need to protect against cache poisoning.
+Relays MUST be able to process objects for the same Full Track Name from
+multiple publishers and forward objects to active matching subscriptions.  The
+same object SHOULD NOT be forwarded more than once to the same subscriber.
```
afrind, line 894:
```suggestion
same object SHOULD NOT be forwarded more than once on the same subscription.
```

draft-ietf-moq-transport.md
ianswett: Fair point.  I guess in that case the "Normal" Object should be ignored under the assumption that the "Does not exist" status is more recent?

afrind: So some transitions will always be illegal:  eg: Normal, End of Group, End of Track can never change amongst each other.  But I think it's fine to say if you get one after receiving Does Not Exist, you can assume Does Not Exist is more recent and ignore the new status?

afrind: I added more text explaining this, and relaxed from "session error" to Malformed track.  Please take a look.

draft-ietf-moq-transport.md
```diff
@@ -846,8 +846,34 @@ validating subscribe and publish requests at the edge of a network.
...
+## Caching Relays
+
 Relays MAY cache Objects, but are not required to.
 
+A caching relay saves Objects to its cache identified by the Object's
+Full Track Name, Group ID and Object ID.
+If multiple objects are received with the same Full Track Name,
+Group ID and Object ID, Relays MAY ignore subsequently received Objects
+or MAY use them to update the cache. Implementations that update the
+cache need to protect against cache poisoning.  The only fields that can be
```
suhasHere, line 858:
```suggestion
cache need to protect against cache poisoning.  The
same object SHOULD NOT be forwarded more than once on the same subscription.
The only fields that can be
```

draft-ietf-moq-transport.md
```diff
@@ -869,14 +895,9 @@ subscribers for each track. Each new OBJECT belonging to the
...
-Full Track Name, Group ID and Object ID. Relays MUST be able to
-process objects for the same Full Track Name from multiple
-publishers and forward objects to active matching subscriptions.
-If multiple objects are received with the same Full Track Name,
-Group ID and Object ID, Relays MAY ignore subsequently received Objects
-or MAY use them to update the cache. Implementations that update the
-cache need to protect against cache poisoning.
+Relays MUST be able to process objects for the same Full Track Name from
+multiple publishers and forward objects to active matching subscriptions.  The
+same object SHOULD NOT be forwarded more than once on the same subscription.
```
suhasHere, line 900: I would remove the last sentence and move it up.

suhasHere (approved):
Looks fine to me with small reordering of the statements

martinduke (changes requested):

draft-ietf-moq-transport.md
```diff
@@ -846,8 +846,34 @@ validating subscribe and publish requests at the edge of a network.
...
+
+1. Object Status can transition from any status to Object Not Exists in cases
+   where the object is no longer available.  Transitions between Normal, End
+   of Group and End of Track are invalid.
+3. Object Header Extensions can be added, removed or updated, subject
+   to the constraints of the specific header extension.
+
+An endpoint that receives a duplicate Object with an invalid Object Status
+change, or a Forwarding Preference, Subgroup ID, Priority or Payload that
+differ from a previous version MUST treat the track as Malformed.
```
martinduke, line 869: This seems to contradict the earlier statement that relays MAY update the cache.

afrind: There are only specific fields in a cache that can be updated without breaking the rules.  Maybe change 'MAY update the cache' to MAY update allowed fields in the cache'?

Issue #762, opened 2025-03-06
3.3. Session initialization - refers to undefined messages


gmarzot:
[3.3. ](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#section-3.3)

[Session initialization](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#name-session-initialization)
The first stream opened is a client-initiated bidirectional control stream where the endpoints exchange Setup messages ([Section 8.2](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#message-setup)). All messages defined in this draft except OBJECT and OBJECT_WITH_LENGTH are sent on the control stream after the Setup message.

suggest last sentence above remove and just say only control messages are sent on the control stream ...

PR #829, opened 2025-03-23
Remove references to old messages

afrind:
Also deleting normative statements prohibiting mixing control/data messages because this is no longer possible -- control streams and data streams have totally different formats and the type spaces overlap.

Fixes: #762


ianswett (approved):

ianswett:
Despite this removing normative language, it's normative language that the draft makes impractical to exercise, so I'd still consider this editorial.

Issue #694, opened 2025-02-24
inband content negotiation make no sense


fluffy:
Section 2.4.2 on scope says 

"MOQT does not provide any in-band content negotiation methods similar to the ones defined by HTTP ([[RFC9110](https://github.com/moq-wg/moq-transport/issues/new#RFC9110)], [Section 10](https://rfc-editor.org/rfc/rfc9110#section-10)); if," 

I think we should remove that part and just leave the 2nd half of the sentence after that.

ianswett:
I tend to agree.  If we were going to have this text, this seems like entirely the wrong place to put it.

PR #851, opened 2025-03-24
Remove reference to HTTP content negotiation

afrind:
I get where we were going with this -- eg: no Accept-Language/Content-Language nonsense, but it's not necessary.

Fixes: #694


fluffy (approved):
LGTM

PR #852, opened 2025-03-24
Remove multiline optional bracket notation

afrind:
Created a Location structure, and refactored Fetch and Extensions.

Fixes: #708

PR #853, opened 2025-03-24
Clarify Latest Object Filter

### LINKED ISSUE #758: LatestObject Subscribe filter is ambiguous.
Individual comment:

Sec 8.6 says:

> Specifies an open-ended subscription beginning from the current object of the current group. If no content has been delivered yet, the subscription starts with the first published or received group.

@fluffy says this can be interpreted two ways:

1) Objects will be forwarded if the sequence number exceeds (Largest Group ID, Largest Object ID) as reported in SUBSCRIBE_OK. I would expect a relay to just send the largest object ID it's observed on a track when composing SUBSCRIBE_OK.

2) Objects will be forwarded if they arrive after the moment of Subscribe, regardless of ID. Largest ID might be (5, 0), but if (4, 25) arrives later due to prioritization/loss/etc, the publisher will send that along too.

Some points:
- I strongly believe that we should clarify this: subscribers should have a solid expectation of what they get when they send SUBSCRIBE. 
 
- For the "regular" use case where group IDs are increasing, interpretation (2) is wasteful. In general, (4, 25) is not going to be decodable without other objects that the publisher is never going to deliver. Furthermore, corresponding Fetch streams are hard to reason about if some pre-largest-ID objects are delivered via Subscribe streams. So I definitely think we need to express "send only higher sequence numbers", alternative (1), in the SUBSCRIBE message somehow.

- It wouldn't be a ton of code to have a new filter that expresses (2). This would accommodate the random Group ID use case we talked about in Denver. It gets us away from "random group ID is something that just happens to work" to "random Group ID is a thing we are explicitly supporting with protocol codepoints." I personally would not add this complication for what I see as a marginal use case, but if we have consensus to do it, I won't lie down in the road over it.

martinduke:
Fixes #758 

Since the current Local Object definition is ambiguous, I eliminated it and replaced it with two unambiguous definition.

I am open-minded about naming.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1361,10 +1361,10 @@ SUBSCRIBE_OK will be delivered by the publisher. This is most useful where the
 Group ID implies a temporal relationship.
 
 All Objects (0x6): All objects that arrive via upstream SUBSCRIBE at, or are
-created by, the publisher after the object indicated by the Largest Group ID and
-Largest Object ID fields in the SUBSCRIBE_OK, whether or not the Location is
-higher.  This is most useful for tracks where the Group ID does not imply a
-temporal relationship.
+created by, the publisher after processing the SUBSCRIBE, whether or not the
+Location is higher than the highest observed when the SUBSCRIBE is processed.
+This is most useful for tracks where the Group ID does not imply a temporal
+relationship.
 
 A filter type other than the above MUST be treated as error.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1357,14 +1357,11 @@ Object' filter.
 
 Higher Objects (0x5): Specifies an open-ended subscription. All objects with
 a Location greater than the Largest Group ID and Latest Object ID fields in the
-SUBSCRIBE_OK will be delivered by the publisher. This is most useful where the
-Group ID implies a temporal relationship.
+SUBSCRIBE_OK will be delivered by the publisher.
 
 All Objects (0x6): All objects that arrive via upstream SUBSCRIBE at, or are
 created by, the publisher after processing the SUBSCRIBE, whether or not the
 Location is higher than the highest observed when the SUBSCRIBE is processed.
-This is most useful for tracks where the Group ID does not imply a temporal
-relationship.
 
 A filter type other than the above MUST be treated as error.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -879,7 +879,7 @@ allows relays to make only a single upstream subscription for the
 track. The published content received from the upstream subscription
 request is cached and shared among the pending subscribers.
 Because SUBSCRIBE_UPDATE only allows narrowing a subscription, relays that
-aggregate upstream subscriptions can subscribe using the Latest Object
+aggregate upstream subscriptions can subscribe using the Higher Objects
 filter to avoid churn as downstream subscribers with disparate filters
 subscribe and unsubscribe from a track.
 
@@ -1346,7 +1346,7 @@ There are 3 types of filters:
 AbsoluteStart (0x3):  Specifies an open-ended subscription beginning
 from the object identified in the StartGroup and StartObject fields. If the
 StartGroup is prior to the current group, the subscription starts at the
-beginning of the current object like the 'Latest Object' filter.
+beginning of the current object like the 'Highest Objects' filter.
 
 AbsoluteRange (0x4):  Specifies a closed subscription starting at StartObject
 in StartGroup and ending at the largest object in EndGroup.  The start and
@@ -1356,22 +1356,22 @@ subscription starts at the beginning of the current object like the 'Latest
 Object' filter.
 
 Higher Objects (0x5): Specifies an open-ended subscription. All objects with
-a sequence number higher than the Largest ID in the SUBSCRIBE_OK will be
-delivered by the publisher. This is most useful where the Group ID implies a
-temporal relationship.
+a Location greater than the Largest Group ID and Latest Object ID fields in the
+SUBSCRIBE_OK will be delivered by the publisher. This is most useful where the
+Group ID implies a temporal relationship.
 
-Later Objects (0x6): All objects that arrive at, or are created by, the
-publisher after the object indicated by the Largest ID field in the
-SUBSCRIBE_OK, whether or not the ID is higher. This is most useful for tracks
-where the Group ID does not imply a temporal relationship.
+All Objects (0x6): All objects that arrive via upstream SUBSCRIBE at, or are
+created by, the publisher after the object indicated by the Largest Group ID and
+Largest Object ID fields in the SUBSCRIBE_OK, whether or not the Location is
+higher.  This is most useful for tracks where the Group ID does not imply a
+temporal relationship.
 
 A filter type other than the above MUST be treated as error.
 
-If a subscriber wants to subscribe to Objects both before and after
-the Latest Object, it can send a SUBSCRIBE for the Latest Object
-followed by a FETCH. Depending upon the application, one might want to send
-both messages at the same time or wait for the first to return before sending
-the second.
+If a subscriber wants to subscribe to Objects that precede the Higher Objects
+filter, it can send a SUBSCRIBE for Higher Objects followed by a FETCH.
+Depending upon the application, one might want to send both messages at the same
+time or wait for the first to return before sending the second.
 
 The format of SUBSCRIBE is as follows:
 
@@ -1778,7 +1778,7 @@ A Subscriber can use a Joining Fetch to, for example, fill a playback buffer wit
 certain number of groups prior to the live edge of a track.
 
 A Joining Fetch is only permitted when the associated Subscribe has the Filter
-Type Latest Object.
+Type Higher Objects.
 
 A Fetch Type other than 0x1, 0x2 or 0x3 MUST be treated as an error.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1343,10 +1343,6 @@ the publisher to identify which objects need to be delivered.
 
 There are 3 types of filters:
 
-Latest Object (0x2): Specifies an open-ended subscription beginning from
-the current object of the current group.  If no content has been delivered yet,
-the subscription starts with the first published or received group.
-
 AbsoluteStart (0x3):  Specifies an open-ended subscription beginning
 from the object identified in the StartGroup and StartObject fields. If the
 StartGroup is prior to the current group, the subscription starts at the
@@ -1358,6 +1354,16 @@ end of the range are inclusive.  EndGroup MUST specify the same or a later
 group than StartGroup. If the StartGroup is prior to the current group, the
 subscription starts at the beginning of the current object like the 'Latest
 Object' filter.
+
+Higher Objects (0x5): Specifies an open-ended subscriptions. All objects with
+a Group and Object ID higher than the Largest ID in the SUBSCRIBE_OK will be
+delivered by the subscriber. This is most useful where the Group ID implies a
+temporal relationship.
+
+Later Objects (0x6): All objects that arrive at, or are created by, the
+publisher after the object indicated by the Largest ID field in the
+SUBSCRIBE_OK, whether or not the ID is higher. This is most useful for tracks
+where the Group ID does not imply a temporal relationship.
 
 A filter type other than the above MUST be treated as error.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1355,9 +1355,9 @@ group than StartGroup. If the StartGroup is prior to the current group, the
 subscription starts at the beginning of the current object like the 'Latest
 Object' filter.
 
-Higher Objects (0x5): Specifies an open-ended subscriptions. All objects with
-a Group and Object ID higher than the Largest ID in the SUBSCRIBE_OK will be
-delivered by the subscriber. This is most useful where the Group ID implies a
+Higher Objects (0x5): Specifies an open-ended subscription. All objects with
+a sequence number higher than the Largest ID in the SUBSCRIBE_OK will be
+delivered by the publisher. This is most useful where the Group ID implies a
 temporal relationship.
 
 Later Objects (0x6): All objects that arrive at, or are created by, the
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1346,7 +1346,7 @@ There are 3 types of filters:
 AbsoluteStart (0x3):  Specifies an open-ended subscription beginning
 from the object identified in the StartGroup and StartObject fields. If the
 StartGroup is prior to the current group, the subscription starts at the
-beginning of the current object like the 'Highest Objects' filter.
+beginning of the current object like the 'Higher Objects' filter.
 
 AbsoluteRange (0x4):  Specifies a closed subscription starting at StartObject
 in StartGroup and ending at the largest object in EndGroup.  The start and


draft-ietf-moq-transport.md
```diff
@@ -1359,6 +1355,16 @@ group than StartGroup. If the StartGroup is prior to the current group, the
 subscription starts at the beginning of the current object like the 'Latest
 Object' filter.
 
+Higher Objects (0x5): Specifies an open-ended subscription. All objects with
+a sequence number higher than the Largest ID in the SUBSCRIBE_OK will be
```
afrind, line 1359: We don't define a sequence number, so can you talk about this in terms of the {group,object} pair?  In #852 I added a Location structure and defined its < operator if that helps.

draft-ietf-moq-transport.md
```diff
@@ -1359,6 +1355,16 @@ group than StartGroup. If the StartGroup is prior to the current group, the
...
+delivered by the publisher. This is most useful where the Group ID implies a
+temporal relationship.
```
afrind, line 1361: I'm not sure we need to comment on utility?  I think one key difference is that when using Joining Fetch, only Higher will guarantee you will not get the same object via FETCH and SUBSCRIBE.

draft-ietf-moq-transport.md
```diff
@@ -1359,6 +1355,16 @@ group than StartGroup. If the StartGroup is prior to the current group, the
...
+
+Later Objects (0x6): All objects that arrive at, or are created by, the
```
afrind, line 1363: I would maybe call this Any Objects or All Objects?  Later and Higher seem easy to confuse.

draft-ietf-moq-transport.md
```diff
@@ -1359,6 +1355,16 @@ group than StartGroup. If the StartGroup is prior to the current group, the
...
+publisher after the object indicated by the Largest ID field in the
```
afrind, line 1364: There's no Largest ID, use Latest Group/Latest Object.

draft-ietf-moq-transport.md
```diff
@@ -1359,6 +1355,16 @@ group than StartGroup. If the StartGroup is prior to the current group, the
...
+SUBSCRIBE_OK, whether or not the ID is higher. This is most useful for tracks
+where the Group ID does not imply a temporal relationship.
```
afrind, line 1366: Same comment on utility - though another key difference is that this filter might be useful when there is re-ordering (eg: if the relay has seen 2-2 because 2-1 was temporarily delayed, you will still get 2-1)

afrind:
You may need to scan the document for places where Latest Object is referenced.  I think at least Joining Fetch mentions it.

wilaw:
This still doesn't address the case where the relay has received a fraction of an Object (i.e at least the first byte but not the last byte). What does it return as Largest Object in the SUBSCRIBE_OK? Does it return the fractional Object, or the last complete Object it received? I feel we should make this clear for relay implementers.

draft-ietf-moq-transport.md
martinduke: If you really don't like it, I can delete it, but I think it's better not to be obtuse about what these are for.

draft-ietf-moq-transport.md
martinduke: See above. In the additional case you suggest, using this filter is very janky compared to Joining Fetch -- I wouldn't recommend it in this case.

I also changed it to make it clear that stuff that arrives by Upstream FETCH is not included in this filter.

martinduke:
> > This still doesn't address the case where the relay has received a fraction of an Object (i.e at least the first byte but not the last byte). What does it return as Largest Object in the SUBSCRIBE_OK? Does it return the fractional Object, or the last complete Object it received? I feel we should make this clear for relay implementers.
> 
> This is tracked in your issue #592 and we'll make a separate PR soon.

Indeed, this PR references the Largest Group/Object fields in SUBSCRIBE_OK, so I prefer that the in-progress issue be addressed there.

draft-ietf-moq-transport.md
```diff
@@ -1359,13 +1355,23 @@ group than StartGroup. If the StartGroup is prior to the current group, the
...
 
+Higher Objects (0x5): Specifies an open-ended subscription. All objects with
+a Location greater than the Largest Group ID and Latest Object ID fields in the
+SUBSCRIBE_OK will be delivered by the publisher. This is most useful where the
+Group ID implies a temporal relationship.
+
+All Objects (0x6): All objects that arrive via upstream SUBSCRIBE at, or are
+created by, the publisher after the object indicated by the Largest Group ID and
+Largest Object ID fields in the SUBSCRIBE_OK, whether or not the Location is
+higher.  This is most useful for tracks where the Group ID does not imply a
```
afrind, line 1366: Maybe change it to:

All objects that arrive via upstream SUBSCRIBE at, or are created by, the publisher after processing the SUBSCRIBE, whether or not the Location is higher.

And I still assert All Objects has some utility in applications even with temporal groups.

martinduke: the Location is higher than what? I added another clause.

I don't see the use case. But you're the editor, feel free to add more text.

fluffy (changes requested):
This is not at all a clarification. This is a fundamental chance to core part of the draft and I would think the chairs are failing to acknowledge we had consensus on that.  

I think you should redo this PR as a new filter time based on higher object but also keep the latest object and make it clear that is is object that arive after this point in time are forwarded on the subscribe.

martinduke:
> This is not at all a clarification. This is a fundamental chance to core part of the draft and I would think the chairs are failing to acknowledge we had consensus on that.
> 
> I think you should redo this PR as a new filter time based on higher object but also keep the latest object and make it clear that is is object that arive after this point in time are forwarded on the subscribe.

As we discussed in Denver, you and I read the same text and drew radically different conclusions as to what it meant. To me, that requires clarification.

Is the "AllObjects" filter not describing the behavior you want? Or is just that you want to retain the LatestObject name?

afrind:
> This is not at all a clarification. This is a fundamental chance to core part of the draft and I would think the chairs are failing to acknowledge we had consensus on that.

Different people in the working group were looking at the same text (for LatestObject) and interpreting it differently. If anything, the current draft supports the "Higher" interpretation of Latest Object.  

> Latest Object (0x2): Specifies an open-ended subscription **beginning from the current object of the current group.** 

> Each new Object belonging to the track **within the subscription range** is forwarded to each active subscriber

> If a subscriber wants to subscribe to Objects **both before and after the Latest Object**, it can send a SUBSCRIBE for the Latest Object followed by a FETCH

I don't see anything in this PR that is a fundamental change.  The goal is to end up with two different filter types that each have an unambiguous meaning.  I'm not sure why the name or code point matters as long as applications have a mechanism to get the behavior they want.

draft-ietf-moq-transport.md
```diff
@@ -1359,13 +1355,23 @@ group than StartGroup. If the StartGroup is prior to the current group, the
...
+Higher Objects (0x5): Specifies an open-ended subscription. All objects with
+a Location greater than the Largest Group ID and Latest Object ID fields in the
+SUBSCRIBE_OK will be delivered by the publisher. This is most useful where the
+Group ID implies a temporal relationship.
+
+All Objects (0x6): All objects that arrive via upstream SUBSCRIBE at, or are
+created by, the publisher after processing the SUBSCRIBE, whether or not the
+Location is higher than the highest observed when the SUBSCRIBE is processed.
+This is most useful for tracks where the Group ID does not imply a temporal
+relationship.
```
suhasHere, line 1367: Subscription filters should be agnostic of group structure. This seems to be adding application semantics and feels fragile to me. 

I am bit unsure on how this filter is addressing the original linked issue

draft-ietf-moq-transport.md
```diff
@@ -1359,13 +1355,23 @@ group than StartGroup. If the StartGroup is prior to the current group, the
 subscription starts at the beginning of the current object like the 'Latest
 Object' filter.
 
+Higher Objects (0x5): Specifies an open-ended subscription. All objects with
+a Location greater than the Largest Group ID and Latest Object ID fields in the
+SUBSCRIBE_OK will be delivered by the publisher. This is most useful where the
+Group ID implies a temporal relationship.
```
suhasHere, line 1361: I am not comfortable of adding group relationship in these filters.

suhasHere (commented):
I see the intent for resolving Will's original issue on clarifying the interpretation. But I think this PR is doing more than just that.  Wondering if we can split into 2 PRs , if needed. 

I don't see a reason to express group structures in the filters though.

draft-ietf-moq-transport.md
```diff
@@ -1359,13 +1355,23 @@ group than StartGroup. If the StartGroup is prior to the current group, the
 subscription starts at the beginning of the current object like the 'Latest
 Object' filter.
 
+Higher Objects (0x5): Specifies an open-ended subscription. All objects with
+a Location greater than the Largest Group ID and Latest Object ID fields in the
```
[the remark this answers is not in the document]
martinduke: It's defined in a different PR. It's just the (group, object) sequence number.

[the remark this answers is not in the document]
martinduke: #852

draft-ietf-moq-transport.md
martinduke: The original issue is that people read the LatestObject definition and had two different interpretations. Both interpretations serve valid use cases. So now there are two precise filter definitions, one for each use case.

martinduke:
> I see the intent for resolving Will's original issue on clarifying the interpretation. But I think this PR is doing more than just that. Wondering if we can split into 2 PRs , if needed.
> 
> I don't see a reason to express group structures in the filters though.

Which use case should I delete from the PR? I think Higher Objects is critical to keep.

afrind:
@suhasHere 

> I am not comfortable of adding group relationship in these filters.

I don't understand the discomfort. LatestObject already has a group relationship:

> Latest Object (0x2): Specifies an open-ended subscription **beginning from the current object of the current group.**

It sounds like you want a filter that just returns any object that arrives or is created after subscribe.  That is the purpose of the "All Objects" filter.  Or do you want something else? 

> I see the intent for resolving Will's original issue on clarifying the interpretation. But I think this PR is doing more than just that. Wondering if we can split into 2 PRs , if needed.

This isn't intended to resolve Will's issue at all.  This is addressing #758 (LatestObject subscribe filter is ambiguous).

> I don't see a reason to express group structures in the filters though.

Have a read through the linked issue and let us know if you have a concrete suggestion to move forward.  I think this PR is in the right direction, since it offers both behaviors that have been requested by the working group.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
afrind: Moved it to #858

suhasHere:
> I don't understand the discomfort. LatestObject already has a group relationship:
> 
> >

I was referring to the sentence that was indicating if a group implies a temporal relationship or not. I don't think that matters for filters.

suhasHere:
> Have a read through the linked issue and let us know if you have a concrete suggestion to move forward. I think this PR is in the right direction, since it offers both behaviors that have been requested by the working group.

I did comment on the original issue where out of order groups can still happen even when groups are temporal and sequential. The AllObjects filter will not address the flow IIUC. 

 I think this fix needs to address that flow and perhaps clarify:

- what does it really mean to include Group/Object info in Subscribe Ok
- Does those are just hints and receivers should handle if things on wire doesn't exactly match (esp when Dsc order was set by the original pub or upstream relay)

martinduke:
> > Have a read through the linked issue and let us know if you have a concrete suggestion to move forward. I think this PR is in the right direction, since it offers both behaviors that have been requested by the working group.
> 
> I did comment on the original issue where out of order groups can still happen even when groups are temporal and sequential. The AllObjects filter will not address the flow IIUC.
> 
> I think this fix needs to address that flow and perhaps clarify:
> 
> * what does it really mean to include Group/Object info in Subscribe Ok
> * Does those are just hints and receivers should handle if things on wire doesn't exactly match (esp when Dsc order was set by the original pub or upstream relay)

The fact that objects can arrive out of order does not, in any way, imply that receivers should expect senders to forward objects that are outside the subscription range defined by the SUBSCRIBE_OK.

Neither does Descending Order.

Maybe you want that use case, and that's fine. That's what the AnyObjects filter is for.

afrind:
> I was referring to the sentence that was indicating if a group implies a temporal relationship or not. I don't think that matters for filters.

+1 to that.  We can just describe the properties of how the filters work.

> what does it really mean to include Group/Object info in Subscribe Ok

I do think this clarification is useful. 

For Highest - this is the exclusive start of your range: you will only receive objects with Location > than the value in OK
For all other filter types - this is informational about the state of the relay and has no bearing on what you will receive

suhasHere:
> The fact that objects can arrive out of order does not, in any way, imply that receivers should expect senders to forward objects that are outside the subscription range defined by the SUBSCRIBE_OK.
> 
> Neither does Descending Order.

The problem here is group is arriving out of order. If we can get rid of descending order for Subscriber, almost all the confusion here is gone IIUC.

suhasHere:
> For Highest - this is the exclusive start of your range: you will only receive objects with Location > than the value in OK

I am still not sure how would this satisfy descending group order issue . How would a end subscriber even know to use Highest vs AllObjects when the groups are all sequential but publisher is sending it in dsc order. Are we saying that , if one asks for Highest and if the groups follow descending order, then the end subscribers should expect not to receive any objects from groups that are behind the one listed in the SubscribeOk. If so, then probably we need to add a note to that extent for dsc group order delivery.

martinduke:
> > The fact that objects can arrive out of order does not, in any way, imply that receivers should expect senders to forward objects that are outside the subscription range defined by the SUBSCRIBE_OK.
> > Neither does Descending Order.
> 
> The problem here is group is arriving out of order. If we can get rid of descending order for Subscriber, almost all the confusion here is gone IIUC.

Descending order is irrelevant to this discussion. The issue is what filter the sender applies to group/object IDs for a subscription.

We all agree, I hope, that for Absolute Range the sender does not send objects < Start, whether the order is Ascending or Descending, and whether or not objects arrive out of order.

The HigherObjects filter is one interpretation of the current LatestObject text. A HigherObject filter is equivalent to an AbsoluteStart filter with a start value of (largest group, largest_id + 1) and behaves in exactly the same way.

An AllObjects filter conforms to the other interpretation. There is no limit on forwarded IDs.

martinduke:
> > For Highest - this is the exclusive start of your range: you will only receive objects with Location > than the value in OK
> 
> I am still not sure how would this satisfy descending group order issue . How would a end subscriber even know to use Highest vs AllObjects when the groups are all sequential but publisher is sending it in dsc order. Are we saying that , if one asks for Highest and if the groups follow descending order, then the end subscribers should expect not to receive any objects from groups that are behind the one listed in the SubscribeOk. If so, then probably we need to add a note to that extent for dsc group order delivery.

It was well-understood by many people who read the LatestObject text that the group in the Subscribe OK was the left limit of what you would receive, whether descending or ascending. The point of descending is not to get stuff deep in the past, it is to get the highest Group IDs first. For descending, old groups are especially irrelevant, so I must not understand the nature of this objection.

Moreover, you are asking for a clarifying note. That is the entire purpose of this PR!

afrind:
Is All Objects the same as AbsoluteStart={0,0} ?

martinduke:
> Is All Objects the same as AbsoluteStart={0,0} ?

Sort of? Except that everything except AllObjects uses Location is a proxy for time. AllObjects uses actual time of arrival. So as long as a (0,0) filter is only applied to new arriving objects, then yes.

fluffy (changes requested):
This makes breaking changes to Latest Object. I disagree that was ever ambiguous and certainly don't see a reason to make breaking changes to it. Can you separate this into two PR. One that proposes the new higher objects filter and another that clarifies the latest object.  

If a publisher is publishing object  1,2,3,4,5,6, and  a relay has already received objects 1,2,3, and then the subscribe what whatever the name of latest object/all objects is, I want the subscribe to get 4,5,6 and not get 1,2,3 which is why I think all objects is a bad name as that would imply it would get 1,2,3 as well as 4,5,6. SO Like the name latest object more than higher objet.

draft-ietf-moq-transport.md
```diff
@@ -1366,29 +1366,32 @@ the publisher to identify which objects need to be delivered.
...
 AbsoluteRange (0x4):  Specifies a closed subscription starting at `Start`
 and ending at the largest object in EndGroup.  The start and
 end of the range are inclusive.  EndGroup MUST specify the same or a later
 group than specified in `start`. If the start group is prior to the current
 group, the subscription starts at the beginning of the current object like
-the 'Latest Object' filter.
+the 'Higher Objects' filter.
+
+Higher Objects (0x5): Specifies an open-ended subscription. All objects with
+a Location greater than the Largest Group ID and Latest Object ID fields in the
```
fluffy, line 1382: I think this is the wrong definition and will result in really hard to resolve race conditions. I think you should take it off the highest group / object the relay has received, not what was in the Subscribe OK. I'm speaking from implementation experience here.  Think hard about the case of a non  cacheing relay that got the object refereed to the the subscribe OK multiple object before it got the OK

vasilvv: Would that actually work, given that joining FETCH relies on the SUBSCRIBE_OK-provided values?

afrind:
Addressed by #927

PR #856, opened 2025-03-25
Restore reference to HTTP content negotiation

vasilvv:
Elaborate on it, and add more normative text that follows from it.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -433,6 +433,10 @@ Because the tuple of Track Namespace and Track Name are unique within an
 MOQT scope, they can be used as a cache key for the track.
 If, at a given moment in time, two tracks within the same scope contain
 different data, they MUST have different names and/or namespaces.
+MOQT provides subscribers with the ability to alter the specific manner in
+which tracks are delivered via Subscribe Parameters, but the actual content of
+the tracks does not depend on those parameters; this is in contrast to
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1442,6 +1446,10 @@ If a publisher cannot satisfy the requested start or end or if the end has
 already been published it SHOULD send a SUBSCRIBE_ERROR with code 'Invalid Range'.
 A publisher MUST NOT send objects from outside the requested start and end.
 
+Subscribe Parameters MUST NOT cause the publisher to alter the payload of the
+objects it sends, as that would violate the track uniqueness guarantee
+described in {{track-scope}}.
```
afrind, line 1451: Do we need a similar statement in the FETCH section?

ianswett: I think this would be better in the "Retrieving Tracks with Subscribe and Fetch" section.  Admittedly, that's true of quite a bit of text in this section, but that way we don't have to duplicate it for Fetch.

fluffy (approved):
LGTM

draft-ietf-moq-transport.md
ianswett: I moved it up and mentioned FETCH as well.

ianswett (approved):

suhasHere (approved):

Issue #708, opened 2025-02-24
Use of square brackets  to indicate optional groups of things is pushing the convention


LPardue:
The document gives a non-normative summary of the RFC 9000 notation it imports, which omits details. In RFC 9000 the notation is defined as

> Packet and frame diagrams in this document use a custom format. The purpose of this format is to summarize, not define, protocol elements. Prose defines the complete semantics and details of structures.Complex fields are named and then followed by a list of fields surrounded by a pair of matching braces. Each field in this list is separated by commas.Individual fields include length information, plus indications about fixed value, optionality, or repetitions. Individual fields use the following notational conventions, with all lengths in bits:

Whereas MoQT does things like:

```
FETCH Message {
  Type (i) = 0x16,
  Length (i),
  Subscribe ID (i),
  Subscriber Priority (8),
  Group Order (8),
  Fetch Type (i),
  [Track Namespace (tuple),
   Track Name Length (i),
   Track Name (..),
   StartGroup (i),
   StartObject (i),
   EndGroup (i),
   EndObject (i),]
  [Joining Subscribe ID (i),
   Preceding Group Offset (i),]
  Number of Parameters (i),
  Parameters (..) ...
}
```

where there are two optional groups of things. 

Possible ways to improve the situation:

1. Explicity extend the convention - state what it means to put square brackets around a set of fields. Including whether the last element should have a trailing comma inside the brackets or not. 
2. Convert these groups into structure. For instance, the QUIC ACK frame has an optional `ECN Counts` field, a structure https://datatracker.ietf.org/doc/html/rfc9000#figure-25.
3. Convert each individual field into being option. This seems like the worst possible approach since it would be easy to get confused what field belongs to what logical group.

PR #857, opened 2025-03-25
Remove multiline optional notation from Extensions

afrind:
Addresses part of #708


martinduke (approved):
This makes it clearer.

PR #866, opened 2025-03-25
Minor section 5 edits

martinduke:
Replaced a pronoun with an explicit reference for clarity.

Replaced subset with superset, because I'm fairly sure that's a typo.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -699,8 +699,8 @@ SHOULD close the session with a protocol error if it detects receiving more than
 one.
 
 The receiver of a SUBSCRIBE_ANNOUNCES_OK or SUBSCRIBE_ANNOUNCES_ERROR ought to
-forward the result to the application, so that the application can make decisions
-about further publishers to contact.
+forward the result to the application, so the application can decide which other
+publishers to contact.
 
 An UNSUBSCRIBE_ANNOUNCES withdraws a previous SUBSCRIBE_ANNOUNCES. It does
 not prohibit the receiver (publisher) from sending further ANNOUNCE messages.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -691,7 +691,7 @@ The syntax of these messages is described in {{message}}.
 If the subscriber is aware of a namespace of interest, it can send
 SUBSCRIBE_ANNOUNCES to publishers/relays it has established a session with. The
 recipient of this message will send any relevant ANNOUNCE or UNANNOUNCE messages
-for that namespace, or subset of that namespace.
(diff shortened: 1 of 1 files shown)


afrind:
is a/b a subset or superset of a/b/c?

draft-ietf-moq-transport.md
```diff
@@ -699,8 +699,8 @@ SHOULD close the session with a protocol error if it detects receiving more than
 one.
 
 The receiver of a SUBSCRIBE_ANNOUNCES_OK or SUBSCRIBE_ANNOUNCES_ERROR ought to
-forward the result to the application, so that it can make decisions about
-further publishers to contact.
+forward the result to the application, so that the application can make decisions
+about further publishers to contact.
```
ianswett, line 703:
```suggestion
forward the result to the application, so the application can decide which other
publishers to contact.
```

ianswett (approved):

martinduke:
> is a/b a subset or superset of a/b/c?

A superset.

This making my head hurt, but in the section above it says "The
recipient of this message will send any relevant ANNOUNCE or UNANNOUNCE messages
for that namespace, or subset of that namespace."

So a SUBSCRIBE_ANNOUNCES for a/b will get ANNOUNCE for a/b/c, which is a subset as used there.

In the sentence I edited is the inverse of that. I guess you could say a/b is a subset, but then you have to change the other paragraph.

afrind:
> > is a/b a subset or superset of a/b/c?
> 
> A superset.

Maybe subset and superset are ambiguous terms here.  Maybe use "more generic" and "more specific"?

draft-ietf-moq-transport.md
```diff
@@ -715,7 +715,7 @@ an ANNOUNCE for it.
 If a publisher is authoritative for a given namespace, or is a relay that has
 received an authorized ANNOUNCE for that namespace from an upstream publisher,
 it MUST send an ANNOUNCE to any subscriber that has subscribed to ANNOUNCE for
-that namespace, or a subset of that namespace. A publisher MAY send the ANNOUNCE
+that namespace, or a superset of that namespace. A publisher MAY send the ANNOUNCE
```
ianswett, line 718: I thought we decided that you only get namespaces directly under the one you SUBSCRIBE_ANNOUNCE to, and it's not recursive?  Or am I misremembering?

Also, this paragraph is one giant sentence that's difficult to read.

ianswett (approved):

draft-ietf-moq-transport.md
martinduke: I think SUBSCRIBE/FETCH are routed by direct match only, but ANNOUNCES are not.

I don't know how to fix the sentence.

Issue #840, opened 2025-03-24
ANNOUNCE is not routing


vasilvv:
Draft says:

> An ANNOUNCE indicates to the subscriber where to route a SUBSCRIBE or FETCH for that namespace.

I think we should not use the word "route" anywhere, since MoQT does not do any routing.  An ANNOUNCE indicates that the sender has tracks for the specified namespace available; what the peer does with it is application-specific, unless the recipient is a relay.

PR #890, opened 2025-03-27
Remove the word "routing" from Namespace Discovery section

afrind:
Fixes: #840

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -674,7 +674,7 @@ both endpoints can immediately destroy relevant state. Objects MUST NOT be sent
 for requests that end with an error.
 
 
-# Namespace Discovery and Routing Subscriptions {#track-discovery}
+# Namespace Discovery {#track-discovery}
 
 Discovery of MoQT servers is always done out-of-band. Namespace discovery can be
 done in the context of an established MoQT session.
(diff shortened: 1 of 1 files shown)


vasilvv (approved):

fluffy (changes requested):
This is too vague. It needs to say the relay networks needs to send the SUBSCRIBE to any matching ANNOUCE and then stop doing that after the UNANNOUCE.

afrind:
> This is too vague. It needs to say the relay networks needs to send the SUBSCRIBE to any matching ANNOUCE 

This is already there in the relay section:

> When a relay receives an incoming SUBSCRIBE request that triggers an upstream subscription, it SHOULD send a SUBSCRIBE request to each publisher that has announced the subscription's namespace

> and then stop doing that after the UNANNOUCE

We have these two statements:

> When a publisher wants to stop new subscriptions for an announced namespace it sends an UNANNOUNCE. 

> An UNANNOUNCE message withdraws a previous ANNOUNCE, although it is not a protocol error for the subscriber to send a SUBSCRIBE or FETCH message after receiving an UNANNOUNCE.

I believe the second clause is there because of races can happen between UNANNOUNCE and SUBSCRIBE/FETCH.  

I think that's pretty clear.  We could perhaps add a "SHOULD NOT send SUBSCRIBE after UNANNOUNCE" in there, but this PR is just editorial removing the word "routing".  Please file an issue if you believe normative changes are required.

Issue #818, opened 2025-03-21
Max_cache_duration is an confusing name for things that don't cache


suhasHere:
the idea behind "max_cache_duration" was a publisher picked time for object's lifetime, more in the lines of TTL for an object. 

I wonder if we should instead call it as Object_TTL instead ?

afrind:
The current definition of MAX_CACHE_DURATION seems to intersect with DELIVERY_TIMEOUT.  In my head (which is not what's in the draft) is MAX_CACHE_DURATION is something that only impacts a FETCH cache, and is precisely what's on the tin - the maximum amount of time to keep an object in a FETCH cache.

Does that definition work better than the current one, which doesn't even mention caching?

PR #891, opened 2025-03-27
Clarify MAX_CACHE_DURATION

afrind:
Fixes: #818


afrind:
I think this is purely editorial?  I added the following words:

"an object can be served from a cache"

and changed

"subscription" to "downstream request"

@suhasHere why you feel like this is a Design change?

ianswett (approved):

suhasHere:
@afrind my apologies , i accidentally labeled it as design. I have one question though?

Who is setting this or does moq provides a way to set this in any control or object messages ?

afrind:
> Who is setting this or does moq provides a way to set this in any control or object messages ?

#813 adds the text about where you can put this, it comes from the publisher.

Issue #804, opened 2025-03-20
Correlation of control messages


gloinul:
So the subscribe and fetch message control messages are all correlated using subscription ID. However, the ANNOUNCE and SUBSCRIBE_ANNOUNCE related are correlated based on track name space. To me this indicates that we might end up with uncertainties about messages. If an endpoint would send two messages before they receive the response message there could be uncertainty to which message the response belongs to. I think there are hidden assumptions here, maybe based on in-order delivery, but are we certain that all implementation may not have different processing. For example receiving a SUBSCRIBE_ANNOUNCES request, forward that upstream before responding, and then getting a second one with an error in, then the response to the error one could be delivered first.

I also get the impression that forcing strict order between request and responses will delay control messages unnecessary. 

So my ask wouldn't a request ID be a clearer handling of this, or do other believe using  track name space is sufficient robust?

ianswett:
TRACK_STATUS has the same issue as ANNOUNCE FYI

gloinul:
I think this applies to all messages that aren't SUBSCRIBE or FETC.

suhasHere:
my proposal is to add request_id/transaction_id to all the control messages.  Fetch might also fall in the same category. not too sure about Subscribe though.

ianswett:
We could also put these on their own streams, as we've discussed before, particularly for FETCH.

afrind:
As noted, SUBSCRIBE and FETCH have ids to correlate already.  That leaves

ANNOUNCE family
SUBSCRIBE_ANNOUNCES family
TRACK_STATUS family

It's easy enough to add a control message identifier to correlate, say, ANNOUNCE_OK with the ANNOUNCE.  I don't think we should allow more than one concurrent announcement per namespace though.  eg:

P->S: ANNOUNCE x
S->P: ANNOUNCE_OK x
P->S: ANNOUNCE x  // this replaces/updates the first announce.

In other words UNANNOUNCE and ANNOUNCE_CANCEL would still use namespace rather than an ID.

The same should be true for SUBSCRIBE_ANNOUNCES, eg: UNSUBSCRIBE_ANNOUNCES only references the namespace.

PR #898, opened 2025-03-30
Use a control message identifier to correlate requests and responses

afrind:
This is more wire efficient and less ambiguous

Fixes: #804

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1103,7 +1103,7 @@ Protocol Violation.
 Some messages contain a control message ID that correlates requests and
 responses.  There are independent control message IDs for each endpoint. The
 client's control message ID starts at 0 and the server's control message ID
-starts at 1.  The control message ID increments by 2 with ANNOUNCE, 
+starts at 1.  The control message ID increments by 2 with ANNOUNCE,
 SUBSCRIBE_ANNOUNCES or TRACK_STATUS.
 
 ## Parameters {#params}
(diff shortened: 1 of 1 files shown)


ianswett (commented):
This PR doesn't add any limits on the control message ID, so there's no protection against rapid reset type attacks as you noted.

I think I'd be simpler to rename Subscribe ID to Control Message ID, and since Subscribe ID already has a MAX mechanism, we get limits on outstanding control messages for free.

draft-ietf-moq-transport.md
```diff
@@ -1098,6 +1098,14 @@ Payload, which is defined by each message type.  If the length does not match
 the length of the Message Payload, the receiver MUST close the session with
 Protocol Violation.
 
+## Control Message ID
+
+Some messages contain a control message ID that correlates requests and
+responses.  There are independent control message IDs for each endpoint. The
+client's control message ID starts at 0 and the server's control message ID
+starts at 1.  The control message ID increments by 2 with ANNOUNCE,
+SUBSCRIBE_ANNOUNCES or TRACK_STATUS.
```
suhasHere, line 1107: Should we define a rule for computing the control message ID ? 
I may be wrong, do we support server sending its own control message today ? It does respond to one, but will send its own though right ?

afrind: Sure, a relay sends ANNOUNCE and SUBSCRIBE today.

afrind:
#905 is my preferred alternate to this PR

ianswett:
Based on feedback, let's proceed with #905

Issue #838, opened 2025-03-24
Data Stream Timeout


vasilvv:
Draft says:

> Data Stream Timeout: The session was closed because the peer took too long to send data expected on an open Data Stream [Section 9](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#data-streams). This includes fields of a stream header or an object header within a data stream. If an endpoint times out waiting for a new object header on an open subgroup stream, it MAY send a STOP_SENDING on that stream, terminate the subscription, or close the session with an error.

I think it's extremely odd for this to cause a fatal connection error?  Especially given that this is can easily occur naturally with due to the publisher being flow control blocked.

afrind:
In HTTP/3 I think there's language that basically says any stream error can be elevated to a session error by an implementation if it feels like it.  Maybe we should add a comment to that effect, but remove the mention of session error here.

afrind:
From RFC 9114:

> An endpoint MAY choose to treat a stream error as a connection error under certain circumstances, closing the entire connection in response to a condition on a single stream. Implementations need to consider the impact on outstanding requests before making this choice.

LPardue:
Yes. IMO Implementations have the right to close anything, at any time, for any reason. The H3 text is pretty decent at articulating that.

PR #901, opened 2025-03-30
Implementations can close the session on any error

afrind:
This mirrors language from HTTP/3, because implementations gonna do.  There is also no recourse for the peer if the other endpoint closes the session.

Fixes: #838

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -618,8 +618,14 @@ code, as defined below:
   long to send data expected on an open Data Stream {{data-streams}}.  This
   includes fields of a stream header or an object header within a data
   stream. If an endpoint times out waiting for a new object header on an
-  open subgroup stream, it MAY send a STOP_SENDING on that stream, terminate
-  the subscription, or close the session with an error.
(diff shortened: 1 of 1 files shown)


vasilvv (approved):

ianswett (approved):

vasilvv (approved):

suhasHere (approved):

fluffy (changes requested):
I don't like this - it just leads to really hard to test system when ever relay treats an error differently. Lets be clear for any given error if what happens and not just be "do whatever. We have done that in the spec so far/.

LPardue:
Disagree. You can't take away an implementation right to close things when it wants. No matter what language is in the spec, there will be good reasons for an endpoint to stop things and not tell the peer why. 

Deployments need to be anti fragile. Errors or abrupt termination will happen and they need to be geared up for that to happen.

gloinul:
As a chair I judge this as being ok to merge. @fluffy I can't see any proposal for what change would solve your concern. I think Lucas comment is right, and I would note that is just documenting the reality. But in most cases the peer will provide error codes and indicate what the issue is.

PR #904, opened 2025-04-02
Clarify TRACK_STATUS codes

afrind:
I made a small change for 0x01 to mean "NOT FOUND" rather than "NOT EXISTS", since a relay may not be able to determine the track status if there is no current publisher for the namespace.  We can add a separate code for NOT_EXISTS, which sounds cacheable.

Fixes: #495

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2110,28 +2110,36 @@ The 'Status Code' field provides additional information about the status of the
 track. It MUST hold one of the following values. Any other value is a malformed
 message.
 
-0x00: The track is in progress, and subsequent fields contain the highest group
-and object ID for that track.
+0x00 (IN PROGRESS): The track is in progress.  Either the publisher is the
+Original Publisher, or a relay with an active subscription for this track.  The
+`Largest` field contains the largest group and object ID available at the
+endpoint.
 
-0x01: The track does not exist. Subsequent fields MUST be zero, and any other
-value is a malformed message.
+0x01 (NOT FOUND): The track does not exist or could not be located by a relay.
+The `Largest` field must have group and object set to 0.  If an endpoint
+receives a non-zero value for either field it MUST close the session with a 
+`Protocol Violation`.
 
-0x02: The track has not yet begun. Subsequent fields MUST be zero. Any other
-value is a malformed message.
+0x02 (NOT STARTED): The track has not yet begun. The Largest field must have
+group and object set to 0.  If an endpoint receives a non-zero value for either
+field it MUST close the session with a `Protocol Violation`.
 
-0x03: The track has finished, so there is no "live edge." Subsequent fields
-contain the highest Group and object ID known.
+0x03 (FINISHED): The track has finished.  The `Largest` field contains the
+highest Group and Object ID in the track.
 
-0x04: The publisher is a relay that cannot obtain the current track status from
-upstream. Subsequent fields contain the largest group and object ID known.
+0x04 (PARTIAL INFORMATION): The publisher is a relay with prior knowledge of
+this track (for example it has at least one object cached), but does not have
+an active subscription, and cannot obtain the current track status from
+upstream. The `Largest` field contains the largest group and object ID known
+to the relay.
 
-Any other value in the Status Code field is a malformed message.
+If an endpoint receives any other value in the Status Code field it MUST close
+the session with a `Protocol Violation`.
 
 The `Largest` field represents the largest Object location observed by the
 Publisher for an active subscription. If the publisher is a relay without an
 active subscription, it SHOULD send a TRACK_STATUS_REQUEST upstream or MAY
-subscribe to the track, to obtain the same information. If neither is possible,
-it should return the best available information with status code 0x04.
+subscribe to the track, to obtain the same information.
 
 The receiver of multiple TRACK_STATUS messages for a track uses the information
 from the latest arriving message, as they are delivered in order on a single


ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -2106,32 +2106,38 @@ TRACK_STATUS Message {
...
-The 'Status Code' field provides additional information about the status of the
-track. It MUST hold one of the following values. Any other value is a malformed
-message.
+The 'Status Code' field provides information about the status of the track. It
+MUST hold one of the following values. If an endpoint receives any other value
+in the Status Code field it MUST close the session with a `Protocol Violation`.
 
-0x00: The track is in progress, and subsequent fields contain the highest group
-and object ID for that track.
+0x00 (IN PROGRESS): The track is in progress.  Either the publisher is the
```
vasilvv, line 2113: Do those need to be in all-caps?

draft-ietf-moq-transport.md
```diff
@@ -2106,32 +2106,38 @@ TRACK_STATUS Message {
...
+Original Publisher, or a relay with an active subscription for this track.  The
```
vasilvv, line 2114: `active subscription` -> `active upstream subscription`

draft-ietf-moq-transport.md
```diff
@@ -2106,32 +2106,38 @@ TRACK_STATUS Message {
...
-0x02: The track has not yet begun. Subsequent fields MUST be zero. Any other
-value is a malformed message.
+0x02 (NOT STARTED): The track has not yet begun. The Largest field must have
+group and object set to 0.  If an endpoint receives a non-zero value for either
+field it MUST close the session with a `Protocol Violation`.
 
-0x03: The track has finished, so there is no "live edge." Subsequent fields
-contain the highest Group and object ID known.
+0x03 (FINISHED): The track has finished.  The `Largest` field contains the
+largest Group and Object ID in the track.
```
vasilvv, line 2128: What happens when a track is closed with no objects ever published on it?

(actual question I had to answer when implementing this)

afrind: @vasilvv : do you mean SUBSCRIBE_OK(ContentExists=false) + SUBSCRIBE_DONE (TRACK_ENDED) and no objects in between?  I guess lets add another status code like "Empty Track"?

draft-ietf-moq-transport.md
afrind, line 2114:
```suggestion
Original Publisher, or a relay with an active upstream subscription for this track.  The
```

draft-ietf-moq-transport.md
vasilvv: Yeah, either "Empty Track", or some other guidance on how to act in this situation.

draft-ietf-moq-transport.md
```diff
@@ -2106,32 +2106,43 @@ TRACK_STATUS Message {
...
-
-Any other value in the Status Code field is a malformed message.
+The 'Status Code' field provides information about the status of the track. It
+MUST hold one of the following values. If an endpoint receives any other value
+in the Status Code field it MUST close the session with a `Protocol Violation`.
+
+0x00 (In Progress): The track is in progress.  Either the publisher is the
+Original Publisher, or a relay with an active upstream subscription for this
+track.  The `Largest` field contains the largest group and object ID available
+at the endpoint.
```
suhasHere, line 2116: nit s/endpoint/publisher

suhasHere, line 2116:
```suggestion
at the publisher
```

draft-ietf-moq-transport.md
suhasHere: My inclination is not to add one more for Empty track , rather expand on Track Finished and clarify it can be reported in these 2 situations. Since accepting subscribe and not sending anything and ending the track is a base case for finished.

draft-ietf-moq-transport.md
```diff
@@ -2106,32 +2106,43 @@ TRACK_STATUS Message {
...
+0x04 (Partial Information): The publisher is a relay with prior knowledge of
+this track (for example, it has at least one object cached), but does not have
+an active subscription, and cannot obtain the current track status from
+upstream. The `Largest` field contains the largest group and object ID known
+to the relay.
+
+0x05 (Empty Track): The track is complete and is known to have no objects.
+The `Largest` field must have group and object set to 0.  If an endpoint
+receives a non-zero value for either field it MUST close the session with
+a `Protocol Violation`.
```
suhasHere, line 2139: please see my comments above. thanks

draft-ietf-moq-transport.md
```diff
@@ -2106,32 +2106,43 @@ TRACK_STATUS Message {
...
+0x02 (Not Started): The track has not yet begun. The `Largest` field must have
+group and object set to 0.  If an endpoint receives a non-zero value for either
+field it MUST close the session with a `Protocol Violation`.
+
+0x03 (Finished): The track has finished.  The `Largest` field contains the
+largest Group and Object ID in the track.
+
+0x04 (Partial Information): The publisher is a relay with prior knowledge of
+this track (for example, it has at least one object cached), but does not have
+an active subscription, and cannot obtain the current track status from
```
suhasHere, line 2132:
```suggestion
an active subscription, and cannot obtain the current status of the track from
```

draft-ietf-moq-transport.md
```diff
@@ -2106,32 +2106,43 @@ TRACK_STATUS Message {
...
+receives a non-zero value for either field it MUST close the session with
+a `Protocol Violation`.
 
 The `Largest` field represents the largest Object location observed by the
 Publisher for an active subscription. If the publisher is a relay without an
 active subscription, it SHOULD send a TRACK_STATUS_REQUEST upstream or MAY
-subscribe to the track, to obtain the same information. If neither is possible,
-it should return the best available information with status code 0x04.
+subscribe to the track to obtain the same information before sending
+TRACK_STATUS.
```
suhasHere, line 2145:
```suggestion
TRACK_STATUS. Alternatively, a relay publisher MAT respond with TRACK_STATUS having `Status` of 0x04 (Partial Information), wherever possible
```

suhasHere (commented):
Few editorial suggestions. Otherwise Looks good to me.

draft-ietf-moq-transport.md
fluffy: Is it clear how long the relay keeps this info before it just become a 01 error ?

draft-ietf-moq-transport.md
```diff
@@ -2106,32 +2106,44 @@ TRACK_STATUS Message {
...
+0x04 (Partial Information): The publisher is a relay with prior knowledge of
+this track (for example, it has at least one object cached), but does not have
+an active upstream subscription, and cannot obtain the current status of the
+track from upstream. The `Largest` field contains the largest group and object
+ID known to the relay.
+
+0x05 (Empty Track): The track is complete and is known to have no objects.
+The `Largest` field must have group and object set to 0.  If an endpoint
+receives a non-zero value for either field it MUST close the session with
+a `Protocol Violation`.
```
fluffy, line 2139: I'm not follow why we can not use a 03 for empty track

fluffy:
Few comments inline but I can live with this. Not sure I am see the use case that need it but whatever.

suhasHere (approved):

afrind:
FWIW - I took a stab a [complete rewrite of TRACK_STATUS](https://github.com/moq-wg/moq-transport/compare/main...afrind-track-status-rewrite) - basically to make two response messages (OK/ERROR) which very closely match what would come in SUBSCRIBE_OK/_ERROR.

draft-ietf-moq-transport.md
DanielFay22: I think the dedicated Empty code makes sense. The track is known to exist (at least at that moment), so 0x01 doesn't make sense, and 0x03 communicates an object that does not exist and will fail if a subscriber attempts to retrieve it.

DanielFay22 (approved):

PR #914, opened 2025-04-07
Proposing `BATCH_SUBSCRIBE`

Johan511:
Often multiple channels of media such as subtitles, audio, base layer / enhancement layer video are subscribed to simultaneously.

It would make sense that they are all within the same track namespace

Consider the following object model

```
very_long_namespace_prefix/media_id/subtitles/eng
very_long_namespace_prefix/media_id/audio/eng
very_long_namespace_prefix/media_id/base_layer_video
very_long_namespace_prefix/media_id/enhancement_layer_video
``` 

Currently this would require 4 different subscribe messages to be sent, each with the same track namespace prefix of  "`very_long_namespace_prefix/media_id`"

We propose a simple fix to this redundancy with the `BATCH_SUBSCRIBE` message to subscribe to multiple tracks using a single message.

```
BATCH_SUBSCRIBE Message {
  Type (i) = 0x6,
  Length (i),
  Track Namespace Prefix (tuple),
  Number of Subscriptions (i),
  (
    Subscribe ID (i),
    Track Alias (i),
    Track Namespace Suffix (tuple),
    Track Name Length (i),
    Track Name (..),
    Subscriber Priority (8),
    Group Order (8),
    Filter Type (i),
    [StartGroup (i),
     StartObject (i)],
    [EndGroup (i)],
    Number of Parameters (i),
    Subscribe Parameters (..) ...
  ) ...
}
```

This also provides the implementation an hint that these subscriptions are related and possibly optimise delivery accordingly.

Answers to some possible questions
1. **Why not update the existing `SUBSCRIBE` message?** : Updating the existing subscribe message adds an overhead of 1-2 bytes in situations where no batched subscriptions are being made. I do not have a strong opinion on whether the 1-2 bytes are significant

2. **How do we handle the numerous interactions associated with `SUBSCRIBE`?** : The receiver of the `BATCH_SUBSCRIBE` must treat it to be identical to receiving multiple `SUBSCRIBE` messages, this means it should reply to it using multiple `SUBSCRIBE_OK`/`SUBSCRIBE_ERROR` messages and `SUBSCRIBE_DONE` when the subscription is fulfilled, allow updates or cancels on each of the subscriptions individually using `SUBSCRIBE_UPDATE` or `UNSUBSCRIBE`. This removes much of the complexity associated with handling `BATCH_SUBSCRIBE` separately. 

3. **There are already talks of wildcard subscription, why do we need this?** : Assuming wildcard subscriptions to mean all tracks within a namespace; this might seldom be the right solution. Consider the above mentioned case, a wildcard subscription to all tracks within `very_long_namespace_prefix/media_id` would mean the client receives all subtitles and audio they are authorized to, this might include languages the client is not interested in receiving and data which would end up being discarded by the client. There might also be multiple post processed versions of the for censorship, better experience for specially-abled, etc.,

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1077,6 +1077,8 @@ MOQT Control Message {
 |-------|-----------------------------------------------------|
 | 0x5   | SUBSCRIBE_ERROR ({{message-subscribe-error}})       |
 |-------|-----------------------------------------------------|
+| 0x6   | BATCH_SUBSCRIBE ({{message-batch-subscribe}})       |
+|-------|-----------------------------------------------------|
 | 0xA   | UNSUBSCRIBE ({{message-unsubscribe}})               |
 |-------|-----------------------------------------------------|
 | 0x2   | SUBSCRIBE_UPDATE ({{message-subscribe-update}})     |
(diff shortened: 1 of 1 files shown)


ianswett:
Thanks for your thorough thoughts, but I believe this is more appropriate for an issue than a PR.  In particular, I believe there is some overlap with the Wildcard subscribe issue #615

Johan511:
Raised a PR instead of an issue because I had an concrete idea of how the change in the draft would look like too.

I will raise an issue tagging this PR

As mentioned, in the description
There are use cases which aren't aligned with wildcard subscription.

Most propositions for wildcard subscribe are for track discovery and a general assumption that all tracks discovered would be used.

PS: We assume out of band knowledge of object hierarchy

fluffy:
I think this would need to deal with ways of updating or ending the multiple subscribes. I question of it the saving in bandwidth gets us that much from just doing multiple subscribes. Would be interested it see how this lines up with wildcard subscribe stuff.

Johan511:
> I think this would need to deal with ways of updating or ending the multiple subscribes. 

Does this answer the concern?

> How do we handle the numerous interactions associated with SUBSCRIBE? : The receiver of the BATCH_SUBSCRIBE must treat it to be identical to receiving multiple SUBSCRIBE messages, this means it should reply to it using multiple SUBSCRIBE_OK/SUBSCRIBE_ERROR messages and SUBSCRIBE_DONE when the subscription is fulfilled, allow updates or cancels on each of the subscriptions individually using SUBSCRIBE_UPDATE or UNSUBSCRIBE. This removes much of the complexity associated with handling BATCH_SUBSCRIBE separately.

Saving bandwidth is definitely one of the main goals, It will definitely use less bandwidth than sending multiple subscribe messages as we are only batching the subscribe messages together, everything else related to the subscribe messages (such as SUBSCRIBE_OK / SUBSCRRIBE_ERROR / ...) remain the same

> Would be interested it see how this lines up with wildcard subscribe stuff.

The PR description talks about this briefly

> There are already talks of wildcard subscription, why do we need this? : Assuming wildcard subscriptions to mean all tracks within a namespace; this might seldom be the right solution. Consider the above mentioned case, a wildcard subscription to all tracks within very_long_namespace_prefix/media_id would mean the client receives all subtitles and audio they are authorized to, this might include languages the client is not interested in receiving and data which would end up being discarded by the client. There might also be multiple post processed versions of the media for censored content, region local content, better experience for specially-abled, etc.,

Also, I believe this had been discussed in a different elsewhere that wildcard SUBSCRIBE has more of a POST semantic where the server is pushing data to the client based on hints (the wildcard subscription) given by the client.

IMHO
Wildcard subscribe will be much easier to reason out once [PUBLISH] (or some message type with a POST semantic)(https://datatracker.ietf.org/meeting/122/materials/slides-122-moq-publish-and-subscribe-to-a-namespace-00.pdf) as proposed by @ianswett @fluffy gets solidified

fluffy:
on the question about updating / ending ... yes, I am clear on how that works now and don't see a problem with it. Thanks.

Johan511:
Closing PR along with the issue

Issue #916, opened 2025-04-08
Can we remove the Length field in Client_Setup and Server_Setup?


Sy0307:
I encountered some issues when trying to be compatible with older versions of MOQ (05 and earlier). In other control flow messages, I can try to determine the currently negotiated version through the session, and then perform different parsing accordingly. Although this is complex, it can still be done. However, it's difficult for me to do this in Client_Setup because the first field I parse is Length. In versions before Draft-05, we parsed 'Number of Supported Versions'. We also cannot obtain version information during the negotiation phase, which leads to irreparable backward incompatibility here.

Can we remove the setting of the Length field here? (In fact, I found that many MOQ implementations do not rely on the Length field here.)

Sy0307:
I do not think so. 0x50 stands for STREAM_HEADER_TRACK in the draft-04 and before. Maybe other code point can be better? Removing version negotiation could result in larger compatibility problems.

afrind:
Why are you trying to support older moq draft versions?  In general, maintaining compatibility with older draft versions is a non-goal.

afrind:
> 0x50 stands for STREAM_HEADER_TRACK in the draft-04 and before

STREAM_HEADER_TRACK and SETUP can't appear in the same type of stream (uni vs bidi).

>  At the very least, ensure we don't have code point conflicts with past versions. 

We can change the new setup points for draft-11.  It's not easy from the draft to see all the old points we used.  However, be aware that #881 and #499 are coming eventually.

> We previously used an MOQ based on an old draft for the media transfer function, and it's currently live. 

Do you have an upgrade path for the live code to get more modern draft versions, or you plan to support these draft versions indefinitely?  Do you need to support *all* draft versions?

Can you look over the diff between the editor's copy and draft-10, to see if there's any other minor tweaks we can make to improve draft-10/11 compat?

Sy0307:
> STREAM_HEADER_TRACK and SETUP can't appear in the same type of stream (uni vs bidi).

I know that STREAM_HEADER_TRACK and SETUP exist in different types of streams, but I hope to avoid them as much as possible. After all, we still have some spare numbers we can use.

> However, be aware that https://github.com/moq-wg/moq-transport/issues/881 and https://github.com/moq-wg/moq-transport/pull/499 are coming eventually.

I just read a bit, and I think https://github.com/moq-wg/moq-transport/pull/499 is also a good idea. It will take some time, but it can maybe still maintain compatibility.

[...]

PR #917, opened 2025-04-08
Change new SETUP points to 0x20/21

afrind:
Removes conflict with "STREAM_HEADER_TRACK" in drafts 02-05.  I didn't specifically reserve this code point, since STREAM_HEADER_TRACK was not legal on a control stream in those draft versions.  See #916.

Reserved code point 0x1 that was used for draft-00 SETUP


vasilvv (approved):
Not sure it will help much if we swap tag and length, but why not.

afrind:
> Not sure it will help much if we swap tag and length, but why not.

Maybe we should hold that off until we can use ALPN VN?

ianswett (approved):

afrind:
Since we just landed the new code points and no one has had time to use them yet, I'm marking this editorial.

Issue #731, opened 2025-02-26
Have each Group have a pointer to the prior Group ID to assist with intentionally sparse IDs


wilaw:
[Proposed by Zafer Gurel] For publishers publishing sparse Group IDs, it would assist relays in knowing that a gap has been intentionally created, since relays are required to attempt to fill such gaps when responding to a FETCH request. 

Consider a relay receiving groups 1,2,3,5. It caches those groups. A later FETCH then requests the range 1..5.

If the relay has no information about the gap, it must make an upstream FETCH (4).

If however Group 5 carried a header indicating that the prior Group was 4, then the relay does not need to make any upstream FETCH as it knows the gap is intentional. 

This efficiency gain may seem trivial for one missing object, if however a publisher has produced 10,000 sparse gaps then this efficiency gain becomes meaningful for a relay operator. 

This indicator of the prior group need only exist in the stream header so it will not materially effect wire size.

ianswett:
From a functional perspective, I believe this is equivalent to Missing Group in Object Status: https://www.ietf.org/archive/id/draft-ietf-moq-transport-08.html#section-8.1.1.1

However, it appears we got rid of Missing Group at some point, and now we only have Missing Object, but I can't remember why?

afrind:
Wait where did GROUP_NOT_EXISTS go?

afrind:
https://github.com/moq-wg/moq-transport/pull/509 removed it

afrind:
Individual Comment:

+1 This could be useful.  It may need to be an optional field, since I can imagine cases where even the original publisher doesn't know (crash and restart?).  There's also not a super convenient place to serialize this right now?  In SUBGROUP_HEADER?

gwendalsimon:
Yes, I wanted to mention it in the last interim, GROUP_NOT_EXIST is missing --- or MISSING_GROUP does not exist :-).

I like this idea, which seems better than the "track flag" since it offers flexibility. It would typically handle a track being set to generate sequential +1 group_id (so easy to cache) but where publisher reconnections can happen, so gaps may happen, even if rare. It would be much harder to deal in the flag implementation.

The Previous_Group_ID (or any other name) may be put in the object 0 of a new group.

afrind:
Thoughts on use an extension header to carry this?  eg:

Previous Group Extension

Type: 0xTBD(even)
Value: Varint containing the ID of the previous published group in this track
Added By: Original Publisher only
Relays may not remove or modify this extension
Must be cached
[...]

PR #922, opened 2025-04-11
Add a mechanism for communicating the size of a Group ID gap

afrind:
Fixes: #731

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3021,11 +3021,11 @@ Prior Group ID Gap (Extension Header Type 0x40) is a variable length integer
 containing the number of Groups prior to the current Group that do not and will
 never exist. For example, if the Original Publisher published an Object in Group
 7 and knows it will never publish any Objects in Group 8 or Group 9, it can
-include Prior Group ID Gap = 2 in any Object in Group 10.  A track with a Group
-that contains more than one Object with different values for Prior Group ID Gap,
+include Prior Group ID Gap = 2 in any number of Objects in Group 10, as it sees fit.  A track with a Group
+that contains more than one Object with different values for Prior Group ID Gap or
 has a Prior Group ID Gap larger than the Group ID is considered malformed.  If an
 endpoint receives an Object with a Group ID within a previously communicated gap
-is also treats the track as malformed.
+it also treats the track as malformed.
 
 This extension is optional, as publishers might not know the prior gap gize. If
 Prior Group ID Gap is not present, the receiver cannot infer any information
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3028,7 +3028,7 @@ MUST be ignored.
 
 This extension is not mandatory, as publishers might not know the Gap Size. This
 extension can be added by the Original Publisher, but cannot be added by relays.
-This extension cannot be removed.
+This extension cannot be modified or removed.
 
 # Security Considerations {#security}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3023,7 +3023,7 @@ never exist. For example, if the Original Publisher published an Object in Group
 7 and knows it will never publish any Objects in Group 8 or Group 9, it can
 include Group ID Gap Size = 2 in any Object in Group 10.  A Group that contains
 more than one Object with different values for Group ID Gap Size is considered
-malformed.  If this extensions is present in an Object in Group 0, the value
+malformed.  If this extension is present in an Object in Group 0, the value
 MUST be ignored.
 
 This extension is not mandatory, as publishers might not know the Gap Size. This
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3027,8 +3027,8 @@ malformed.  If this extension is present in an Object in Group 0, the value
 MUST be ignored.
 
 This extension is not mandatory, as publishers might not know the Gap Size. This
-extension can be added by the Original Publisher, but cannot be added by relays.
-This extension cannot be modified or removed.
+extension can be added by the Original Publisher, but MUST NOT be added by
+relays. This extension MUST NOT be modified or removed.
 
 # Security Considerations {#security}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3021,15 +3021,16 @@ Prior Group ID Gap (Extension Header Type 0x40) is a variable length integer
 containing the number of Groups prior to the current Group that do not and will
 never exist. For example, if the Original Publisher published an Object in Group
 7 and knows it will never publish any Objects in Group 8 or Group 9, it can
-include Prior Group ID Gap = 2 in any number of Objects in Group 10, as it sees fit.  A track with a Group
-that contains more than one Object with different values for Prior Group ID Gap or
-has a Prior Group ID Gap larger than the Group ID is considered malformed.  If an
-endpoint receives an Object with a Group ID within a previously communicated gap
-it also treats the track as malformed.
+include Prior Group ID Gap = 2 in any number of Objects in Group 10, as it sees
+fit.  A track with a Group that contains more than one Object with different
+values for Prior Group ID Gap or has a Prior Group ID Gap larger than the Group
+ID is considered malformed.  If an endpoint receives an Object with a Group ID
+within a previously communicated gap it also treats the track as malformed.
 
-This extension is optional, as publishers might not know the prior gap gize. If
-Prior Group ID Gap is not present, the receiver cannot infer any information
-about the existence of prior groups (see {{group-ordering}}).
+This extension is optional, as publishers might not know the prior gap gize, or
+there may not be a gap. If Prior Group ID Gap is not present, the receiver
+cannot infer any information about the existence of prior groups (see
+{{group-ordering}}).
 
 This extension can be added by the Original Publisher, but MUST NOT be added by
 relays. This extension MUST NOT be modified or removed.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3011,6 +3011,24 @@ SUBGROUP_HEADER {
 
 ~~~
 
+# Extension Headers
+
+The following Object Extension Headers are defined in MoQT.
+
+## Group ID Gap Size
+
+Group ID Gap Size (Extension Header Type 0x40) is a variable length integer
+containing the number of Groups prior to the current Group that do not and will
+never exist. For example, if the Original Publisher published an Object in Group
+7 and knows it will never publish any Objects in Group 8 or Group 9, it can
+include Group ID Gap Size = 2 in any Object in Group 10.  A Group that contains
+more than one Object with different values for Group ID Gap Size is considered
+malformed.  If this extensions is present in an Object in Group 0, the value
+MUST be ignored.
+
+This extension is not mandatory, as publishers might not know the Gap Size. This
+extension can be added by the Original Publisher, but cannot be added by relays.
+This extension cannot be removed.
 
 # Security Considerations {#security}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3022,11 +3022,11 @@ containing the number of Groups prior to the current Group that do not and will
 never exist. For example, if the Original Publisher published an Object in Group
 7 and knows it will never publish any Objects in Group 8 or Group 9, it can
 include Group ID Gap Size = 2 in any Object in Group 10.  A Group that contains
-more than one Object with different values for Group ID Gap Size is considered
-malformed.  If this extension is present in an Object in Group 0, the value
-MUST be ignored.
+more than one Object with different values for Group ID Gap Size, has a
+Group ID Gap Size larger than or equal to the Group ID, or has an Object present
+in a Group ID within a gap is considered malformed.
 
-This extension is not mandatory, as publishers might not know the Gap Size. This
+This extension is optional, as publishers might not know the Gap Size. This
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -3011,6 +3011,24 @@ SUBGROUP_HEADER {
...
+never exist. For example, if the Original Publisher published an Object in Group
+7 and knows it will never publish any Objects in Group 8 or Group 9, it can
+include Group ID Gap Size = 2 in any Object in Group 10.  A Group that contains
+more than one Object with different values for Group ID Gap Size is considered
+malformed.  If this extensions is present in an Object in Group 0, the value
+MUST be ignored.
+
+This extension is not mandatory, as publishers might not know the Gap Size. This
+extension can be added by the Original Publisher, but cannot be added by relays.
+This extension cannot be removed.
```
afrind, line 3031:
```suggestion
This extension cannot be modified or removed.
```

draft-ietf-moq-transport.md
```diff
@@ -3011,6 +3011,24 @@ SUBGROUP_HEADER {
...
+
+## Group ID Gap Size
+
+Group ID Gap Size (Extension Header Type 0x40) is a variable length integer
+containing the number of Groups prior to the current Group that do not and will
+never exist. For example, if the Original Publisher published an Object in Group
+7 and knows it will never publish any Objects in Group 8 or Group 9, it can
+include Group ID Gap Size = 2 in any Object in Group 10.  A Group that contains
+more than one Object with different values for Group ID Gap Size is considered
+malformed.  If this extensions is present in an Object in Group 0, the value
```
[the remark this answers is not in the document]
zafergurel: A solution to this may be as follows:
`If there are more than one Object with different values for Group ID Gap size, the Object with the lowest ID is used to set Group ID Gap Size.`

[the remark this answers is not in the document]
afrind: > How should the receiver react to a malformed group? 

I'll file a separate issue.  There are lots of ways a track or group can be malformed today.

draft-ietf-moq-transport.md
```diff
@@ -3011,6 +3011,24 @@ SUBGROUP_HEADER {
 
 ~~~
 
+# Extension Headers
+
+The following Object Extension Headers are defined in MoQT.
```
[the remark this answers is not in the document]
afrind: The motivation for using Extension Headers is largely about keeping the SUBGROUP_HEADER cleaner, but could also be spelled that way.

> be updated if MOQT is also going to start defining extensions?

Let's gather more feedback here, but if we define some "core extensions" that are in moqt we will update that.

draft-ietf-moq-transport.md
```diff
@@ -3011,6 +3011,24 @@ SUBGROUP_HEADER {
...
+never exist. For example, if the Original Publisher published an Object in Group
+7 and knows it will never publish any Objects in Group 8 or Group 9, it can
+include Group ID Gap Size = 2 in any Object in Group 10.  A Group that contains
+more than one Object with different values for Group ID Gap Size is considered
+malformed.  If this extension is present in an Object in Group 0, the value
+MUST be ignored.
+
+This extension is not mandatory, as publishers might not know the Gap Size. This
+extension can be added by the Original Publisher, but cannot be added by relays.
+This extension cannot be modified or removed.
```
afrind, line 3031:
```suggestion
extension can be added by the Original Publisher, but MUST NOT be added by
relays. This extension MUST NOT be modified or removed.
```

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
ianswett: We don't expect gaps to be on most Objects, so a header seems appropriate

draft-ietf-moq-transport.md
```diff
@@ -3011,6 +3011,24 @@ SUBGROUP_HEADER {
...
+
+## Group ID Gap Size
+
+Group ID Gap Size (Extension Header Type 0x40) is a variable length integer
+containing the number of Groups prior to the current Group that do not and will
+never exist. For example, if the Original Publisher published an Object in Group
+7 and knows it will never publish any Objects in Group 8 or Group 9, it can
+include Group ID Gap Size = 2 in any Object in Group 10.  A Group that contains
+more than one Object with different values for Group ID Gap Size is considered
+malformed.  If this extension is present in an Object in Group 0, the value
```
ianswett, line 3026: As discussed, please also discuss the case when the Gap value is larger than the Group ID

ianswett (commented):
Some small comments

draft-ietf-moq-transport.md
```diff
@@ -3011,6 +3011,24 @@ SUBGROUP_HEADER {
...
+MUST be ignored.
```
afrind, line 3027:
```suggestion
include Group ID Gap Size = 2 in any Object in Group 10.  A Group that contains
more than one Object with different values for Group ID Gap Size, has a
Group ID Gap Size larger than or equal to the Group ID, or has an Object present
in a Group ID within a gap is considered malformed.
```

draft-ietf-moq-transport.md
```diff
@@ -3011,6 +3011,24 @@ SUBGROUP_HEADER {
...
+
+This extension is not mandatory, as publishers might not know the Gap Size. This
```
afrind, line 3029:
```suggestion
This extension is optional, as publishers might not know the Gap Size. This
```

draft-ietf-moq-transport.md
```diff
@@ -3011,6 +3011,24 @@ SUBGROUP_HEADER {
...
+## Group ID Gap Size
+
+Group ID Gap Size (Extension Header Type 0x40) is a variable length integer
+containing the number of Groups prior to the current Group that do not and will
+never exist. For example, if the Original Publisher published an Object in Group
+7 and knows it will never publish any Objects in Group 8 or Group 9, it can
+include Group ID Gap Size = 2 in any Object in Group 10.  A Group that contains
+more than one Object with different values for Group ID Gap Size, has a
+Group ID Gap Size larger than or equal to the Group ID, or has an Object present
+in a Group ID within a gap is considered malformed.
```
suhasHere, line 3027: I am not sure I fully parse the last sentence. May be an example might help here.

suhasHere (changes requested):
I have few questions on the proposal here on

1.  how can this be enforced if the original publisher marks some cases where there was a gap and some places doesn't do so ? 

2. If this extension is omitted, what is the expectation at the subscriber ? Does this mean that groups are linearly and sequentially incrementing or is it that the subscriber can't assume group structure information?

draft-ietf-moq-transport.md
afrind: > has an Object present in a Group ID within a gap is considered malformed.

This part?  What I mean is: I receive Object {10,0} with Group ID Gap Size=1, then I receive Object {9,0}, but the publisher told me Group ID 9 does not exist.

afrind:
> 1. how can this be enforced if the original publisher marks some cases where there was a gap and some places doesn't do so ?

The only thing that can be _enforced_ is that if the publisher says there's a gap, then it can't publish anything in the gap range.  This is a helpful hint to caches and receivers that they don't need to bother using FETCH to fill a gap, because there are no objects in it.

> 2. If this extension is omitted, what is the expectation at the subscriber ? Does this mean that groups are linearly and sequentially incrementing or is it that the subscriber can't assume group structure information?

I intended it to mean "the subscriber can't assume group structure information".

draft-ietf-moq-transport.md
ianswett: It's quite a long sentence, so how about splitting it up? I think I'd split up the "Has an Object present ..." clause because it's a fairly different case from the first two.

Also, can the gap not be equal to the Group ID?  ie: Can Group 2 not have a Gap Size of 1?

draft-ietf-moq-transport.md
```diff
@@ -3011,6 +3011,24 @@ SUBGROUP_HEADER {
 
 ~~~
 
+# Extension Headers
+
+The following Object Extension Headers are defined in MoQT.
+
+## Group ID Gap Size
```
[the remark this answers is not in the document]
ianswett: Or maybe "Prior Group Gap"?  Size is somewhat implied.

ianswett (approved):

suhasHere:
> I intended it to mean "the subscriber can't assume group structure information".

I think we should make it explicit given our consensus from Denver interim

draft-ietf-moq-transport.md
afrind: I think you're right, equal is ok but it can't be greater than the group ID:

Group 1 can have a Gap size 1 (eg, Group 0 does not exist)

draft-ietf-moq-transport.md
```diff
@@ -3011,6 +3011,28 @@ SUBGROUP_HEADER {
...
+
+The following Object Extension Headers are defined in MoQT.
+
+## Prior Group ID Gap
+
+Prior Group ID Gap (Extension Header Type 0x40) is a variable length integer
+containing the number of Groups prior to the current Group that do not and will
+never exist. For example, if the Original Publisher published an Object in Group
+7 and knows it will never publish any Objects in Group 8 or Group 9, it can
+include Prior Group ID Gap = 2 in any Object in Group 10.  A track with a Group
```
suhasHere, line 3024:
```suggestion
include Prior Group ID Gap = 2 in any number of Objects in Group 10, as it sees fit.  A track with a Group
```

draft-ietf-moq-transport.md
```diff
@@ -3011,6 +3011,28 @@ SUBGROUP_HEADER {
...
+that contains more than one Object with different values for Prior Group ID Gap,
```
suhasHere, line 3025:
```suggestion
that contains more than one Object with different values for Prior Group ID Gap or
```

draft-ietf-moq-transport.md
```diff
@@ -3011,6 +3011,28 @@ SUBGROUP_HEADER {
...
+has a Prior Group ID Gap larger than the Group ID is considered malformed.  If an
+endpoint receives an Object with a Group ID within a previously communicated gap
+is also treats the track as malformed.
```
suhasHere, line 3028:
```suggestion
treats the track as malformed.
```

draft-ietf-moq-transport.md
```diff
@@ -3011,6 +3011,28 @@ SUBGROUP_HEADER {
...
+
+This extension is optional, as publishers might not know the prior gap gize. If
+Prior Group ID Gap is not present, the receiver cannot infer any information
+about the existence of prior groups (see {{group-ordering}}).
+
+This extension can be added by the Original Publisher, but MUST NOT be added by
+relays. This extension MUST NOT be modified or removed.
```
suhasHere, line 3035: I get the intention. There is no easy to enforce this unless there is a way to verify the integrity between original publisher and the end subscriber.

suhasHere (changes requested):
Thanks for writing this proposal and I feel is heading in the right direction. 

I do have big comments and some minor nits (added as Pr comments). 

I think a publisher has 2 options:
1. Send End of Group object for each such missing groups
2. Use  the gap proposal 
I suspect original publisher do (1) when they clearly know due to congestion they are unable to publish an entire group. Ex is one group per audio object flows.
Publishers use (2), when they know the gap is big and they don't want to fill in end of group objects as (2) is more better option. The proposal should add text identifying both.

Canoninical object definition needs to be updated to reflect the gap information.

Also, this PR should add some notes on caching relays when they see objects with Gap size.

draft-ietf-moq-transport.md
afrind, line 3028:
```suggestion
it also treats the track as malformed.
```

draft-ietf-moq-transport.md
afrind: That will be true of every extension defined in MoQT, right?

draft-ietf-moq-transport.md
```diff
@@ -3011,6 +3011,28 @@ SUBGROUP_HEADER {
...
+
+The following Object Extension Headers are defined in MoQT.
+
+## Prior Group ID Gap
+
+Prior Group ID Gap (Extension Header Type 0x40) is a variable length integer
+containing the number of Groups prior to the current Group that do not and will
+never exist. For example, if the Original Publisher published an Object in Group
+7 and knows it will never publish any Objects in Group 8 or Group 9, it can
+include Prior Group ID Gap = 2 in any number of Objects in Group 10, as it sees fit.  A track with a Group
```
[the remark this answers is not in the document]
afrind: That's true.  It's also true for other object Statuses that might be more important like End of Group/End of Track.

sharmafb (approved):

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
suhasHere: I am fine with extension given its applies to specific flows or conditions. Having said that, I am ok with anything that works for streams as well as datagrams and not bloat the latter, where not needed.

zafergurel (approved):

suhasHere (approved):
LGTM

draft-ietf-moq-transport.md
```diff
@@ -3011,6 +3011,28 @@ SUBGROUP_HEADER {
...
+that contains more than one Object with different values for Prior Group ID Gap or
+has a Prior Group ID Gap larger than the Group ID is considered malformed.  If an
+endpoint receives an Object with a Group ID within a previously communicated gap
+it also treats the track as malformed.
+
+This extension is optional, as publishers might not know the prior gap gize. If
```
ianswett, line 3030:
```suggestion
This extension is optional, as publishers might not know the prior gap gize or there might not be a gap. If
```

ianswett (approved):

Issue #909, opened 2025-04-02
Wire efficiency for Subgroup ID


afrind:
Subgroup ID can be optimized away in the most common cases:

Subgroup-per-group => subgroup ID=0
Subgroup-per-object => subgroup ID=object ID

We can use a couple bits from the stream header type to indicate this.  With #821 we use the LSB to determine if extension header length is present.  We can use the next two bits:

0x2: Subgroup ID present
0x4: Subgroup ID=Object ID

There's only 3 valid states, if 0x2=1 then 0x4=1 is a protocol error.

Issue #923, opened 2025-04-11
How to deal with malformed tracks?


afrind:
There are currently several ways a group or track can be malformed by the publisher:

1) Subgroup with decreasing or duplicate object ID
2) Subgroup FIN in the middle of an object
3) A duplicate subgroup ID with object IDs higher than received on the previously FIN'd subgroup
4) Objects in a Group following "End of Group"
5) Objects or Groups in a Track following "End of Track"
6) The same object produced twice with different content or other immutable properties
7) mixing Subgroup and Datagram in the same track

There may be more...

We should add a section that mentions these conditions, and tell relays and consumers what to do if they receive a malformed track

ianswett:
I added some text in #806 but I agree we probably need more.

afrind:
HTTP/3 has the following section: https://datatracker.ietf.org/doc/html/rfc9114#section-4.1.2

We have the beginnings of "relays/proxies/intermediaries MUST NOT forward malformed tracks".  The other bit that HTTP has is that if you receive something malformed, you reset the stream.  What's the MoQ equivalent?

Issue #610, opened 2024-11-08
Handling subgroup discontinuity at a relay


afrind:
This came from the comments on #587 

The context is that a track has a subgroup with only even numbered objects, eg: 2, 4, 6, 8.

> sending 2, 4, 8 = illegal (in order, but with an omitted object in the middle.)

If a relay gets a subgroup with objects `2, 4, 6, 8`, I expect it can send one of the following

```
Nothing
2
2, 4
2, 4, 6
2, 4, 6, 8
```

However, consider this case:

```
          /--- relay 1 ---\
Publisher                  relay3 ---> Subscriber 
          \--- relay 2 ---/
```

For a particular track/group/subgroup (eg: group=77, subgroup=11) relay3 gets the following from relay1:

```
Control Stream: <GOAWAY>
Subgroup Stream: 2, 4<RELIABLE_RESET>
```
as well as a new ANNOUNCE from relay2.  

relay3 starts making a new subscription to relay2, but doesn't close anything downstream.  

relay3 issues SUBSCRIBE(latest object) to relay2, and gets a stream with just `8` for group=77 subgroup=11.
If it gets any non-sequential object (even 6) it has no idea if it missed any objects.

Would the relay be allowed to send `2, 4, 8` on the subgroup stream downstream?  If not, then what is it supposed to do?

* FETCH to fill in the gap?  
* Reset all the downstream subgroup streams?
  * If so, can it open a new subgroup stream containing just `8` or stop trying to forward this subgroup?
* Something else?

TimEvens:
Your scenario is pretty specific on the mobility/resume use-case where the publisher moves from relay 1 to relay 2. Is that correct? 

Mobility/resume/reuse does need to be handled, which I don't see in the current MoQT draft.  Effort is needed to define how to resume and properly handle GOAWAY to migrate/transition existing  tracks from one to another. 

Regarding peering between relays, such as relay 1 and relay 2 to relay 3... fetch should not be used by relays unless the subscriber requested it.  IMO, fetch should be focused on subscriber specific uses  and not be used to globally sync/replicate caches. If we want to define a cache sync protocol, then that's a different discussion... for example, we really need to look at distributed centralized caches that support relays that are behind load balancers and anycast (load balancers). 

As I would implement the above, the following would happen:

1. Publisher would make a new connection to relay 2
2. Publisher would undergo the normal connect/subscribe/announce states for the tracks. 
[...]

PR #924, opened 2025-04-11
Save a byte when subgroup ID is a common value

afrind:
Use the Type field to encode Subgroup ID=0 (eg stream-per-group) or Subgroup ID=Object ID (eg stream-per-object).

Fixes: #909

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2552,7 +2552,9 @@ the type of the stream in question.
 |-------------|-------------------------------------------------|
 | ID          | Type                                            |
 |------------:|:------------------------------------------------|
-| 0x08-0x09   | SUBGROUP_HEADER  ({{subgroup-header}})          |
+| 0x08-0x0D   | SUBGROUP_HEADER  ({{subgroup-header}})          |
+|-------------|-------------------------------------------------|
+| 0x0E-0xF    | RESERVED                                        |
 |-------------|-------------------------------------------------|
 | 0x05        | FETCH_HEADER  ({{fetch-header}})                |
 |-------------|-------------------------------------------------|
@@ -2772,7 +2774,7 @@ SUBGROUP_HEADER {
   Type (i),
   Track Alias (i),
   Group ID (i),
-  Subgroup ID (i),
+  [Subgroup ID (i),]
   Publisher Priority (8),
 }
 ~~~
@@ -2781,11 +2783,26 @@ SUBGROUP_HEADER {
 All Objects received on a stream opened with `SUBGROUP_HEADER` have an
 `Object Forwarding Preference` = `Subgroup`.
 
-The Type field takes the form 0b0000100X (or the set of values from 0x08 to
-0x09). The LSB determines if the Extensions Headers Length is present in Objects
-in this subgroup.  When it is 0, Extensions Headers Length is not present and all
-Objects have no extensions.  When it is 1, Extension Headers Length is present in
-all Objects in this subgroup.
+The Type field takes the form 0b00001XXX (or the set of values from 0x08 to
+0x0D). The three low-order bits of the frame type determine the fields that
+are present in the frame:
+
+The EXT bit (0x01) determines if the Extensions Headers Length is present in
+Objects in this subgroup.  When it is 0, Extensions Headers Length is not
+present and all Objects have no extensions.  When it is 1, Extension Headers
+Length is present in all Objects in this subgroup.
+
+The HAS_SUBGROUP_ID  bit (0x02) determines if the Subgroup ID field is present.
+When it is 0, there is no explicit Subgroup ID field and the SUBGROUP_ID bit
+determines the Subgroup ID.  When it is 1, the Subgroup ID field is present.
+
+The SUBGROUP_ID bit (0x4) determines the implicit Subgroup ID when the
+HAS_SUBGROUP_ID bit is 0.  When the SUBGROUP_ID bit is 0, the Subgroup ID is 0.
+When the SUBGROUP_ID bit is 1, the Subgroup ID is the Object ID of the first
+object transmitted in this subgroup.
+
+It is not valid to send HAS_SUBGROUP_ID=1 and SUBGROUP_ID=1.  Those values of
+type (0x0E-0x0F) are reserved and have no meaning in MoQT.
 
 To send an Object with `Object Forwarding Preference` = `Subgroup`, find the open
 stream that is associated with the subscription, `Group ID` and `Subgroup ID`,


draft-ietf-moq-transport.md
```diff
@@ -2781,11 +2783,26 @@ SUBGROUP_HEADER {
...
+The Type field takes the form 0b00001XXX (or the set of values from 0x08 to
+0x0D). The three low-order bits of the frame type determine the fields that
+are present in the frame:
+
+The EXT bit (0x01) determines if the Extensions Headers Length is present in
+Objects in this subgroup.  When it is 0, Extensions Headers Length is not
+present and all Objects have no extensions.  When it is 1, Extension Headers
+Length is present in all Objects in this subgroup.
+
+The HAS_SUBGROUP_ID  bit (0x02) determines if the Subgroup ID field is present.
```
afrind, line 2795:
```suggestion
The HAS_SUBGROUP_ID bit (0x02) determines if the Subgroup ID field is present.
```

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -2781,11 +2783,26 @@ SUBGROUP_HEADER {
...
-Objects have no extensions.  When it is 1, Extension Headers Length is present in
-all Objects in this subgroup.
+The Type field takes the form 0b00001XXX (or the set of values from 0x08 to
+0x0D). The three low-order bits of the frame type determine the fields that
+are present in the frame:
+
+The EXT bit (0x01) determines if the Extensions Headers Length is present in
+Objects in this subgroup.  When it is 0, Extensions Headers Length is not
+present and all Objects have no extensions.  When it is 1, Extension Headers
+Length is present in all Objects in this subgroup.
```
suhasHere, line 2793: how does it work in case where

1. Few objects have extensions , but few don't . Say key_frame being presenting in only one object vs others
2. Few extensions are type|value type, but some ore type|len|value type ..

draft-ietf-moq-transport.md
```diff
@@ -2781,11 +2783,26 @@ SUBGROUP_HEADER {
...
-The Type field takes the form 0b0000100X (or the set of values from 0x08 to
-0x09). The LSB determines if the Extensions Headers Length is present in Objects
-in this subgroup.  When it is 0, Extensions Headers Length is not present and all
-Objects have no extensions.  When it is 1, Extension Headers Length is present in
-all Objects in this subgroup.
+The Type field takes the form 0b00001XXX (or the set of values from 0x08 to
+0x0D). The three low-order bits of the frame type determine the fields that
+are present in the frame:
+
+The EXT bit (0x01) determines if the Extensions Headers Length is present in
```
suhasHere, line 2790:
```suggestion
The EXT bit (0x001) determines if the Extensions Headers Length is present in
```

draft-ietf-moq-transport.md
```diff
@@ -2781,11 +2783,26 @@ SUBGROUP_HEADER {
...
+Objects in this subgroup.  When it is 0, Extensions Headers Length is not
+present and all Objects have no extensions.  When it is 1, Extension Headers
+Length is present in all Objects in this subgroup.
+
+The HAS_SUBGROUP_ID bit (0x02) determines if the Subgroup ID field is present.
```
suhasHere, line 2795:
```suggestion
The HAS_SUBGROUP_ID bit (0x010) determines if the Subgroup ID field is present.
```

draft-ietf-moq-transport.md
```diff
@@ -2781,11 +2783,26 @@ SUBGROUP_HEADER {
...
+When it is 0, there is no explicit Subgroup ID field and the SUBGROUP_ID bit
+determines the Subgroup ID.  When it is 1, the Subgroup ID field is present.
+
+The SUBGROUP_ID bit (0x4) determines the implicit Subgroup ID when the
```
suhasHere, line 2799:
```suggestion
The SUBGROUP_ID bit (0x100) determines the implicit Subgroup ID when the
```

draft-ietf-moq-transport.md
```diff
@@ -2781,11 +2783,26 @@ SUBGROUP_HEADER {
...
+HAS_SUBGROUP_ID bit is 0.  When the SUBGROUP_ID bit is 0, the Subgroup ID is 0.
+When the SUBGROUP_ID bit is 1, the Subgroup ID is the Object ID of the first
+object transmitted in this subgroup.
+
+It is not valid to send HAS_SUBGROUP_ID=1 and SUBGROUP_ID=1.  Those values of
```
suhasHere, line 2804: what should the receiver do in such cases?

draft-ietf-moq-transport.md
```diff
@@ -2781,11 +2783,26 @@ SUBGROUP_HEADER {
...
+Length is present in all Objects in this subgroup.
+
+The HAS_SUBGROUP_ID bit (0x02) determines if the Subgroup ID field is present.
+When it is 0, there is no explicit Subgroup ID field and the SUBGROUP_ID bit
+determines the Subgroup ID.  When it is 1, the Subgroup ID field is present.
+
+The SUBGROUP_ID bit (0x4) determines the implicit Subgroup ID when the
+HAS_SUBGROUP_ID bit is 0.  When the SUBGROUP_ID bit is 0, the Subgroup ID is 0.
+When the SUBGROUP_ID bit is 1, the Subgroup ID is the Object ID of the first
+object transmitted in this subgroup.
```
suhasHere, line 2802: it is also not valid to have SUBGROUP_ID set to 0 in multiple sub-groups within a group mode. Since one will end up having 2 subgroups with same SubGroupID.

draft-ietf-moq-transport.md
afrind: > Few objects have extensions , but few don't
> Few extensions are type|value type, but some ore type|len|value type .

This is how it works today.

All objects have extension length.  The ones with extensions set a non-zero length, followed by a serialized block of the specified length in bytes.

draft-ietf-moq-transport.md
afrind: See #923 which is compiling a list of all the ways a publisher can make a malformed track.

Transmitting the same subgroup ID twice is _not necessarily_ malformed.  The publisher can reset a subgroup stream and later open another stream with same subgroup ID.  

It can happen when a relay loses an upstream publisher in the middle of a subgroup and resets the downstream.  Then the publisher comes back but is still publishing objects in the same group/subgroup.  See #610.

draft-ietf-moq-transport.md
afrind: The same thing the receiver does today when receiving a unidirectional stream with unknown type.  I guess we don't say what to do?  moxygen closes the session: https://github.com/facebookexperimental/moxygen/blob/main/moxygen/MoQCodec.cpp#L133-L136.

martinduke (approved):

draft-ietf-moq-transport.md
suhasHere: This will not allow some objects to have extensions and some don't. I think that's a miss and we should allow it

suhasHere (changes requested):

draft-ietf-moq-transport.md
suhasHere: I did notice #923  which is useful. I do think, we can clarify in this PR that there can't be 2 active subgroups within the same track.

draft-ietf-moq-transport.md
```diff
@@ -2781,11 +2783,26 @@ SUBGROUP_HEADER {
...
+Length is present in all Objects in this subgroup.
+
+The HAS_SUBGROUP_ID bit (0x010) determines if the Subgroup ID field is present.
+When it is 0, there is no explicit Subgroup ID field and the SUBGROUP_ID bit
+determines the Subgroup ID.  When it is 1, the Subgroup ID field is present.
+
+The SUBGROUP_ID bit (0x100) determines the implicit Subgroup ID when the
+HAS_SUBGROUP_ID bit is 0.  When the SUBGROUP_ID bit is 0, the Subgroup ID is 0.
+When the SUBGROUP_ID bit is 1, the Subgroup ID is the Object ID of the first
+object transmitted in this subgroup.
```
suhasHere, line 2802: Also I wonder if we should flip the decision.

SubGroupId bit is 1 , use the SubGroupId in the header 
SubGroupId bit is 0 , use the objectId of the first object that is observed on the stream at the subscriber

draft-ietf-moq-transport.md
afrind: > This will not allow some objects to have extensions and some don't. I think that's a miss and we should allow it

You are right and I agree, but I don't know how to do that with the current serialization.  After the SUBGROUP_HEADER, there's only { Object ID, [Extensions], Length, Payload }.  We could try to get cute and borrow a bit from Object ID?

draft-ietf-moq-transport.md
afrind: We had this text in draft-02, but lost it in -06, I think:

> A sender SHOULD NOT open more than one multi-object stream at a time with the same stream header message type and fields. 

I'll add back some flavor of that.

draft-ietf-moq-transport.md
afrind: > SubGroupId bit is 1 , use the SubGroupId in the header
> SubGroupId bit is 0 , use the objectId of the first object that is observed on the stream at the subscriber

I think you are mixing the two bits, this PR has:

HAS_SUBGROUP_ID | SUBGROUP_ID | Subgroup ID
-|-|-
0 | 0 | 0
0 | 1 | Object ID of first Object
1 | 0 | Subgroup ID Field
_1_ | _1_ | _Reserved Code Point_

draft-ietf-moq-transport.md
afrind: There's this:

> Objects from the same Subgroup MUST NOT be sent on different streams, unless one of the streams was reset prematurely, or upstream conditions have forced objects from a Subgroup to be sent out of Object ID order.

So there actually is a case where you can have two active streams for the same subgroup - some kind of reconnect case where objects from the first publish attempt are delayed.

Publisher Session 1 -> Relay: {g=10, sg=0, o=0} (packet lost)
Publisher Starts make-before-break, opens Session 2
Publisher -> Relay Session 1: {g=10, sg=0, o=1} (HoL blocked at relay)
Publisher -> Relay Session 2: {g=10, sg=0, o=1} -> opens a stream to downstream subscriber for 10/0
Publisher Session 1 -> Relay: {g=10, sg=0, o=0} (retransmit) -> have to open a second subgroup downstream

draft-ietf-moq-transport.md
suhasHere: This confusion is caused since we had a bit in the type field for extensions. If we leave the extensions to the object scoped, then this will not be a problem.

One proposal i can think is extend the type bits to specify 3 values
- all objects scoped rules 
- per object scoped rule

XY 
00 - undefined
01 - each object has to be looked at to determine  ( this is same as it is defined today)
10 - all objects have no extensions
11 - all objects have extensions

afrind: 01 and 11 are the same - if 01 means "same as today" - then you have to look at the Extension Length (in bytes).  If non-zero, there are extensions and if 0, there are not.  11 is "all objects have a non-zero extension length" - this doesn't save any bytes on the wire.

If you want to save the Extension Length byte in some subgroup objects, you need to use bits from either Object ID or Payload Length to cue the presence of Extension Length.  In any case, it's complex enough I think it should be a separate PR.

suhasHere: I meant to say you use 4 bits instead of 3 bits in the type
Field

This giving 2 bits for extensions logic

The way we have the Pr ,
Removes a functionality that was working

suhasHere: I meant to say you use 4 bits instead of 3 bits in the type
Field

This giving 2 bits for extensions logic

The way we have the Pr ,
Removes a functionality that was working

afrind: > Removes a functionality that was working

That is 100% not the intent.  This PR does not change extensions _at all_ - it changes _subgroup id_.  Maybe you are referring to #821?  But that PR _also_ did not change the function of extensions.

In draft-10 - _every object_ has Extension Headers Length, in bytes, even if it is 0.  With 821, the publisher can _optionally_ remove this byte if it knows there will be no extensions in _any objects_ in the subgroup.

ianswett: > 
> In draft-10 - _every object_ has Extension Headers Length, in bytes, even if it is 0. With 821, the publisher can _optionally_ remove this byte if it knows there will be no extensions in _any objects_ in the subgroup.

Thanks, I think this is a clean way of thinking about it.

sharmafb (approved):

draft-ietf-moq-transport.md
suhasHere: thanks @ianswett  and @afrind  . I was indeed mixing up few issues here.

suhasHere (approved):

gmarzot:
Doesn't that control exist already to the extent that each object can
express whether it has extensions or not, and so the publisher could send
them out that way (first object only) w/o subgroup signalling?

also, and not to be a broken record (at least i am consistent) - adding a
type to SUBGROUP_HEADER where the subgroup header itself can have
extensions that apply to the entire group/subgroup. This aligns fairly well
with GOP-specific metadata and has the added benefit of being attached to
an essential element in the stream rather than a "special" object which may
be dropped or filtered(?) and the metadata is lost.

fluffy (commented):
UHHH .... part of me likes this and it might be right thing, but I think I would rather looks at getting rid of bits in a comprehensive look at all the bits in the message.

afrind:
There's been consistent feedback that the wire format is fine but it would be better to list explicit code points for specific purposes/encodings so we don't waste any bits in the type field.

ianswett (approved):
Thanks, this is definitely a simpler read

Issue #899, opened 2025-03-30
Do we need to allow multiple ranges in a FETCH request


afrind:
Consider this case:

A track has two subgroups, base layer, with the even objects and enhancement layer, with odd objects.
Because of delivery timeout, a caching relay ends up with only the base layer (objects=0, 2, 4, 6, ...)
A client sends a FETCH for the whole group

This will generate one FETCH _per object_ from the relay to the upstream.

A solution would be to amend FETCH to have N ranges, eg:

```
Range {
  Start (Location),
  End (Location)
}

FETCH Message {
  ...
  Num Ranges (i),
  [ Requested Range (Range) ]+
  ...
}
```

I think the response is unambiguous(?), all objects after the End returned in FETCH_OK are implicitly not available yet.

gwendalsimon:
It is a good idea. 

What about the response to it? In your example, I would assume FETCH_ERROR No Object for every range (1-1, 3-3, 5-5, ...).

What happens in another case:
Subscriber A subscribed to the (gapped) Track from 14 to 27. Relay cached all groups it dealt with: 14-17-22-23-24-27.
Subscriber B subscribed to the Track from 49 to 53, Relay cached all groups it dealt with: 49-50-52-53.
Relay receives FETCH 20-56 from Subscriber C.

The Relay would send upstream FETCH [ 20-21, 25-26, 28-48- 51-51]
Which would generate
FETCH_ERROR No Object 20-21
probably a FETCH_OK for the range between 27 and 49 and some transmitted group
FETCH_ERROR No Object 51-51

We may generate one FETCH response (either ERROR or OK) per range. Or we may group FETCH responses too, but the number of ranges in the first response may not be equal to the number of ranges in the initial FETCH.

wilaw:
In principle I like the idea. One issue to address is scale.  think multiple ranges works fine with examples of 3 gaps. But imagine the same situation of odd numbers where your range is 0 - 10000. Now you want a FETCH with 5,000 ranges in the message. We would need an upper limit of how many ranges can be specified in one message.

Since FETCH responses are sent down a single stream, the results don't need to be grouped. There may be multiple sparse errors for missing objects. I thought that we weren't explicitly declaring missing objects. The FETCH response would close and any objects not returned were ipso facto missing?

afrind:
> What about the response to it? In your example, I would assume FETCH_ERROR No Object for every range (1-1, 3-3, 5-5, ...).

[...]

Issue #907, opened 2025-04-02
gaps management in Joining FETCH


gwendalsimon:
Gaps in GroupID may require clarification for JONING_FETCH.

### Relative Joining FETCH

The draft says that the "_Subscriber can use a Joining Fetch to, for example, fill a playback buffer with a certain number of groups prior to the live edge of a track._" Hence I understand that the "_relative starting offset_" (simply offset) is the number of Objects that the Subscriber wants to gets from the FETCH.

The Section 8.12.1 explains how to calculate the range from the Offset. It is said "_Fetch StartGroup: Subscribe Largest Group - Joining start_". (By the way "Joining start" is not defined but I guess it is the offset). Using a subtraction between integers here does not take into account the case with gaps. If we want the offset to be a number of Objects, we must define Fetch StartGroup as being the Group ID _i_ such that the number of Objects in the range [i, Subscribe Largest Group] is exactly equal to the offset.

#### Problem
It is not only an editorial problem. Imagine a Relay without cache who has an ongoing Subscribe (for any other subscriber). It receives a new SUBSCRIBE + a Joining_FETCH with offset=10. It cannot upstream the Joining_FETCH to the Publisher since it does not have any Subscribe ID to attach to the Joining_FETCH. The Relay should instead sends a Standalone FETCH upstream, but since it does not know whether gaps exist, it cannot know how to set Start_Group.

### Absolute Joining FETCH

Absolute Joining FETCH has been designed to enable SWITCH by letting the Subscriber decide the exact _transition group_ at which the switch should happen. The Subscriber knows the groupID it is currently downloading (say _x_). It may know the latest groupID at the Publisher (say _y_). It chooses the transition groupID (say _z_) between _x_ and _y_, based on multiple criteria: whether it wants the transition to happen fast, whether it can take the risk to not have the data on time wrt playback speed, whether some events are planned.

With gaps, the Subscriber is blind on the status in the range _[x,y]_. It does not know how many groups are in the range and which groupID exist. It is thus hard to take any decision. Consequences include pause during switch due to bad choice. An example is shown in #902.

ianswett:
An application needs to have some reasonable expectations of when there might be gaps in Group ID.  I'd expect most applications would have few to no gaps, in which case Joining Fetch is ok.

If you're doing unusual group numbering, hopefully the Subscriber knows that.

afrind:
[...]

PR #925, opened 2025-04-11
Allow multiple ranges in FETCH

afrind:
Fixes: #899

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1928,8 +1928,8 @@ using the following structues:
 
 ~~~
 Delta Location {
-  Group Delta (i), 
-  Object Delta (i) 
+  Group Delta (i),
+  Object Delta (i)
 }
 ~~~
 


draft-ietf-moq-transport.md
```diff
@@ -1910,17 +1913,57 @@ delivery.
...
+  End (Delta Location)
+}
+~~~
+
+A Delta Range contains a Start and End.  Start is the delta from the previous
+range End, or from Group=0, Object=0 for the first range.  The decoded range
+MUST contain at least one object, otherwise the endpoint returns FETCH_ERROR
+with `Invalid Range`.
+
+If the End Location Object in a range is 0 after decoding, the entire End Group
```
ianswett, line 1959: End Group?

```suggestion
If the End Location Object in a range is 0 after decoding, the entire Group
```

draft-ietf-moq-transport.md
```diff
@@ -1935,10 +1978,8 @@ FETCH Message {
   [Track Namespace (tuple),
    Track Name Length (i),
    Track Name (..),
-   StartGroup (i),
-   StartObject (i),
-   EndGroup (i),
-   EndObject (i),]
+   Number of Ranges (i),
+   Delta Range (..) ]
```
ianswett, line 1982: Did you miss a square bracket or were you intending something else?

```suggestion
   [ Delta Range (..) ]
```

draft-ietf-moq-transport.md
```diff
@@ -1895,6 +1896,8 @@ Absolute Joining Fetch (0x3) : Identical to a Relative Joining Fetch except that
 StartGroup is determined by an absolute Group value rather than a relative offset to
 the subscription.
 
+Joining Fetch types only request a single range.
```
ianswett, line 1899: Aside: Joining Fetch is getting awkward.  I think we're close to the point where we should just fold it back into SUBSCRIBE

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1910,17 +1913,57 @@ delivery.
...
+Delta Range {
+  Start (Delta Location),
+  End (Delta Location)
+}
+~~~
+
+A Delta Range contains a Start and End.  Start is the delta from the previous
+range End, or from Group=0, Object=0 for the first range.  The decoded range
+MUST contain at least one object, otherwise the endpoint returns FETCH_ERROR
+with `Invalid Range`.
```
gwendalsimon, line 1957: What is the motivation for delta encoding instead of a list of static ranges? I would prefer static ranges for several reasons: 
First is practical: When the Relay reads the cache, it is easy to identify the gaps, and to state [ Start=(group1,object1), End=(group2, object2) ] although delta requires checking the End of the previous range and subtracting.
Second is editorial: I found it hard to read. It is not a good sign when pseudocode is needed.
Third is about precision. The Relay does not know the number of objects in a group that it misses. We would need another value to say that the Relay requests all objects of a group (maybe -1?).
(I may completely misunderstand this part)

afrind: The purpose of delta encoding the ranges is to enforce that the ranges are in-order and end is > start for each range.  This is only delta encoded on the wire, it should be trivial to convert the delta encoded request to the equivalent request with absolute ranges.

draft-ietf-moq-transport.md
```diff
@@ -2024,18 +2065,18 @@ error code 'No Objects'.
 
 The publisher receiving a Relative Joining Fetch computes the range as follows:
 
-* Fetch StartGroup: Subscribe Largest Group - Joining start
-* Fetch StartObject: 0
-* Fetch EndGroup: Subscribe Largest Group
-* Fetch EndObject: Subscribe Largest Object
+* Start Location Group: Subscribe Largest Group - Joining Start
```
gwendalsimon, line 2068: If we want to include gaps, then we would need to say that the Fetch StartGroup must be computed such that there exists exactly _Joining Start_ objects between _Subscribe Largest Group_ and _Fetch StartGroup_, i.e. the cardinality of the range [ Fetch StartGroup, Subscribe Largest Group ] is equal to Joining Start. I know it makes the formulation less simple, sorry. 

Another option: we stick to this simple subtraction, but then the motivation for Relative Joining Fetch must be updated because it currently says that "A Subscriber can use a Joining Fetch to, for example, fill a playback buffer with a certain number of groups prior to the live edge of a track." I believe this sentence reflects the consensus on this offset though, so I would prefer the cardinality definition of Fetch StartGroup.

afrind: Your comment doesn't seem related to the thesis of this PR - you may want to add it to #907?

I prefer the simple subtraction here, and would rather massage the motivation to match the case that gaps can exist in the group space.

draft-ietf-moq-transport.md
gwendalsimon: I know it is trivial to convert absolute gaps into delta encoded ones at the requester side, and reconverted it the other way around at the server side... but it is extra-computation on both sides, for the sake of enforcing rules. Wouldn't it be easier to stick to absolute range and say something like: "if the Start of a given range is smaller than the End of a prior range in the list of ranges, then FETCH_ERROR with Invalid Range"?

draft-ietf-moq-transport.md
martinduke: Individual comment: I don't think we should fold it into subscribe, but I *do* think this strengthens the case for it to be a different message type.

martinduke:
> I don't think I like this. Let me explain ...
> 
> We get into the issue of errors for the separate ranges.
> 
> I don't see how this help. We can pipeline all the fetch at the same time. The amount of bandwidth saved in irrelevant.
> 
> This adds complexity to the draft that can be left till later. I'm not aware of a use case we can not do without this.

Individual comment:
I agree with fluffy. This seems like a fair amount of code complexity that does not save us an RTT or anything else that really matters, unless stream or request ID flow control is strict. Arguably, forcing the use of a request ID per range limits the ability to DoS a publisher with a huge set of ranges.

This also makes control messages arbitrarily long, which I don't like.

That said, I'm not going to lie down in the road over this. I agree the odd/even case in the issue that motivates this is super-annoying.

fluffy (changes requested):
I don't think we should do this now.  Looking at profiling of code, I can not see how a request for 5 ranges is going to be faster in any way that matters that 5 fetch requests sent at same time. This is adding to the complexity of fetch, we have a way to do it now, and it does not seem clear it is needed.

afrind:
>  This is adding to the complexity of fetch, we have a way to do it now,

+1  FETCH is already very complicated and I'm not sure I want to update my implementation to do this.

CC: @gwendalsimon who was interested?

gwendalsimon:
OK. No worries.

afrind:
No one is clamoring for a solution to this.  I parked the issue and closing this PR.

Issue #534, opened 2024-09-20
Allow padding streams


vasilvv:
Since we currently don't allow unknown frame types and unknown stream types, there is no way to inject padding on the application level.  I believe it would be useful to allow for at least padding unidirectional streams (i.e. streams where receiver ignores everything past the stream type), those could be used for probing (e.g. sending a unidirectional stream at the lowest priority to determine if the connection could sustain more application data).

ianswett:
As a publisher, one can publish a track like what you're describing, but it sounds like you want it built into the transport.

Understanding how it might be used would be helpful.  Is the subscriber trying to see if it has more bandwidth so it can increase quality?  If so, why not just subscribe to the higher quality at a lower priority?

suhasHere:
Agree with Ian . We already do a mix of these things today ( hi/low tracks, low priority probe-like stuff) .. May be I am lacking to understand the requirement here ?

vasilvv:
> Understanding how it might be used would be helpful. Is the subscriber trying to see if it has more bandwidth so it can increase quality? If so, why not just subscribe to the higher quality at a lower priority?

Those are valid if you have one publisher and a lot of subscriber, but for 1:1 scenarios, you would typically have the sender produce only one video stream (and then adjust its bitrate unilaterally), so that's not really an option.

ianswett:
For 1:1 scenarios where the sender is only producing a single stream, the publisher/sender can decide to probe for more bandwidth unilaterally.  The simplest but effective approach is to just retransmit data you've already sent, aka world's simplest FEC.

afrind:
There was a discussion in the slack channel about this issue, which AI has summarized thus:

The discussion revolves around ensuring the congestion controller isn't app-limited in BBR (Bottleneck Bandwidth and Round-trip propagation time) congestion control algorithm. @kixelated mentions sending QUIC PADDING or PING frames periodically, specifically during the BBR ProbeBW_UP phase, to fill the congestion window until a target bitrate is achieved.

For WebTransport clients, creating a "padding" stream with a high sendOrder (lowest priority) is suggested. However, timing the padding with ProbeBW_UP (~1/8th of the time) is crucial for it to be effective with BBR. @vasilvv mentions BBRv1 adjusts cycles depending on whether you're app-limited or not, and ensuring the probe is at least 8 RTTs long can also help.

[...]

PR #926, opened 2025-04-13
Add editorial text on congestion control

### LINKED ISSUE #37: Dealing with Congestion Control
There were few discussions that discussed on dealing with congestion and needing more thinking around decisions around how and when to drop things.

ianswett:
Fixes #37

Loosely based on @kixelated comment on the issue.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -734,8 +734,8 @@ many MOQT applications will typically be limited by the available data to send, 
 not the congestion controller. Many congestion control algorithms
 only increase the congestion window or bandwidth estimate if fully utilized. This
 combination can lead to underestimating the available network bandwidth. As a result,
-applications might need to ensure the congestion controller is not app-limited for at
-least a full round trip every so often to ensure the full available bandwidth can be
+applications might need to periodically ensure the congestion controller is not
+app-limited for at least a full round trip to ensure the available bandwidth can be
 measured.
 
 ### Consistent Throughput
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -729,7 +729,7 @@ no packet loss.
 ### Application-Limited
 
 The average bitrate for latency sensitive content needs to be less than the available
-bandwidth of the network, otherwise data will be queued and/or dropped. As such,
+bandwidth, otherwise data will be queued and/or dropped. As such,
 many MOQT applications will typically be limited by the available data to send, and
 not the congestion controller. Many congestion control algorithms
 only increase the congestion window or bandwidth estimate if fully utilized. This
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -721,8 +721,8 @@ built on top of MOQT.
 ### Bufferbloat
 
 Traditional AIMD congestion controllers (ex. CUBIC {{?RFC9438}} and Reno {{?RFC6582}})
-are prone to Bufferbloat. Bufferbloat is when elements along the path build up
-a substantial queue or packets, commonly more than doubling the round trip time.
+are prone to Bufferbloat. Bufferbloat occurs when elements along the path build up
+a substantial queue of packets, commonly more than doubling the round trip time.
 These queued packets cause head-of-line blocking and latency, even when there is
 no packet loss.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -720,17 +720,18 @@ built on top of MOQT.
 
 ### Bufferbloat
 
-Traditional AIMD congestion controllers (ex. CUBIC and Reno) are prone to Bufferbloat.
-Bufferbloat is when elements along the path build up a substantial queue or packets,
-commonly more than doubling the round trip time. These queued packets cause
-head-of-line blocking and latency, even when there is no packet loss.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -712,6 +712,40 @@ and announcements. The client can choose to delay closing the session if it
...
+
+MOQT does not specify a congestion controller, but there are important attributes
+to consider when selecting a congestion controller for use with an application
+built on top of MOQT.
+
+### Bufferbloat
+
+Traditional AIMD congestion controllers (ex. CUBIC {{?RFC9438}} and Reno {{?RFC6582}})
+are prone to Bufferbloat. Bufferbloat is when elements along the path build up
+a substantial queue or packets, commonly more than doubling the round trip time.
```
afrind, line 725:
```suggestion
are prone to Bufferbloat. Bufferbloat occurs when elements along the path build up
a substantial queue of packets, commonly more than doubling the round trip time.
```

draft-ietf-moq-transport.md
```diff
@@ -712,6 +712,40 @@ and announcements. The client can choose to delay closing the session if it
...
+
+The average bitrate for latency sensitive content needs to be less than the available
+bandwidth of the network, otherwise data will be queued and/or dropped. As such,
+many MOQT applications will typically be limited by the available data to send, and
+not the congestion controller. Many congestion control algorithms
+only increase the congestion window or bandwidth estimate if fully utilized. This
+combination can lead to underestimating the available network bandwidth. As a result,
+applications might need to ensure the congestion controller is not app-limited for at
+least a full round trip every so often to ensure the full available bandwidth can be
+measured.
```
afrind, line 739: Does MoQT need a mechanism to supply this data when the app is not (eg padding streams?)

ianswett: I'd rather not, because you can always request a higher quality at a low priority if there is one.

If there's not a higher quality, then you don't really need to discover more bandwidth.

Typically, I-Frames supply enough data to make the sender non app-limited for long enough to avoid completely under-utilizing the connection.

Senders control their congestion control, so client mechanisms to try to get their senders to unstick their congestion controllers are a bit scary.  This does remind me to read: https://datatracker.ietf.org/doc/draft-ietf-ccwg-ratelimited-increase/

afrind: See also #534

afrind (approved):
lgtm but it would be good for a few more folks to have a look

draft-ietf-moq-transport.md
suhasHere: Tending to agree with @ianswett  .  Artificial load is always going to be very application specific and may or may not always help (from our BBR experimenations).

suhasHere (approved):

fluffy (approved):
I think this is good but I would not call the Application Limited part as an editorial comment.

draft-ietf-moq-transport.md
```diff
@@ -712,6 +712,40 @@ and announcements. The client can choose to delay closing the session if it
...
+
+### Bufferbloat
+
+Traditional AIMD congestion controllers (ex. CUBIC {{?RFC9438}} and Reno {{?RFC6582}})
+are prone to Bufferbloat. Bufferbloat occurs when elements along the path build up
+a substantial queue of packets, commonly more than doubling the round trip time.
+These queued packets cause head-of-line blocking and latency, even when there is
+no packet loss.
+
+### Application-Limited
```
vasilvv, line 729: I feel like this session should probably also mention ABR somehow.

draft-ietf-moq-transport.md
```diff
@@ -712,6 +712,40 @@ and announcements. The client can choose to delay closing the session if it
...
+
+The average bitrate for latency sensitive content needs to be less than the available
+bandwidth of the network, otherwise data will be queued and/or dropped. As such,
+many MOQT applications will typically be limited by the available data to send, and
+not the congestion controller. Many congestion control algorithms
+only increase the congestion window or bandwidth estimate if fully utilized. This
+combination can lead to underestimating the available network bandwidth. As a result,
+applications might need to ensure the congestion controller is not app-limited for at
+least a full round trip every so often to ensure the full available bandwidth can be
```
ianswett, line 738:
```suggestion
applications might need to periodically ensure the congestion controller is not
app-limited for at least a full round trip to ensure the available bandwidth can be
```

draft-ietf-moq-transport.md
ianswett: Possibly so, but I don't have any text in mind at the moment.  PRs welcome, though.

draft-ietf-moq-transport.md
```diff
@@ -712,6 +712,40 @@ and announcements. The client can choose to delay closing the session if it
...
+Traditional AIMD congestion controllers (ex. CUBIC {{?RFC9438}} and Reno {{?RFC6582}})
+are prone to Bufferbloat. Bufferbloat occurs when elements along the path build up
+a substantial queue of packets, commonly more than doubling the round trip time.
+These queued packets cause head-of-line blocking and latency, even when there is
+no packet loss.
+
+### Application-Limited
+
+The average bitrate for latency sensitive content needs to be less than the available
+bandwidth of the network, otherwise data will be queued and/or dropped. As such,
```
ianswett, line 732:
```suggestion
bandwidth, otherwise data will be queued and/or dropped. As such,
```

Issue #892, opened 2025-03-28
unordered Group ID brings confusion here and there


gwendalsimon:
The Original Publisher may now create groups such that group_ID does not increase with the time of production. In Section 2.3: "_Within a track, the original publisher SHOULD produce Group IDs which increase with time._" A recent added sentence is "_SUBSCRIBE and FETCH delivery use Group Order, so a FETCH cannot deliver Groups out of order and a subscription could have unexpected delivery order if Group IDs do not increase with time._" I am not sure what we can do with these track where group ID do not increase if neither FETCH nor SUBSCRIBE work properly. Anyway, this sentence does not address all the problems in the current spec.

Section 2.3 Groups. "_Subscribers to tracks which do not follow this requirement SHOULD NOT use range filters..._" The Subscriber cannot know whether the track follow strictly increasing group ID or not, it is thus impossible for the Subscriber to know whether it can use ranger filters or not.

Section 6.2 Priority: "_If two objects have the same subscriber and publisher priority, but belong to two different groups of the same track, the group order of the associated subscription is used to decide the one that is sent first._" The Relay cannot use the Group ID to check the order. It should know the order by another mean.

Section 8.6 SUBSCRIBE. The "Filter Types" part should be re-written. For example "_Absolute Start [...] If the start group is prior to the current group, the subscription starts at the beginning of the current object_" can hardly be implemented. The Relay has to ask the Publisher whether, in its full history, there existed a group with this ID. If so, then the Relay should go with beginning of the current object. If not, the Relay should wait for this group to eventually be generated. It is one of the many inconsistencies.

Section 8.6 SUBSCRIBE "_A publisher MUST NOT send objects from outside the requested start and end._" This cannot be implemented.

Section 8.9 SUBSCRIBE_UPDATE "_The end Group MUST NOT increase_" This cannot be implemented. Another sentence "_Like SUBSCRIBE, EndGroup MUST specify the same or a later object than StartGroup and StartObject._" requires a Publisher to check whether the specified end group is later or not. 

Section 9.1.1.1. The bullet "_0x4 End of Track_" must be rewritten since it refers to comparisons of groupID.

For all these group orders problem, one solution could be a specific message to query order to the original Publisher. Something like "is_prior(groupID_a, groupID_b): return TRUE if groupID_a was prior to groupID_b, else FALSE". But it is obviously quite an overhead.

[...]

PR #927, opened 2025-04-14
Clarify how SUBSCRIBE filters work, allow Absolute start before Latest

### LINKED ISSUE #758: LatestObject Subscribe filter is ambiguous.

afrind:
This is an alternate to #853, attempting to resolve #758.

Fixes: #758
Closes #726

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1421,7 +1421,7 @@ AbsoluteRange (0x4):  The filer Start Location and End Group are specified
 explicitly in the SUBSCRIBE message. The `Start` specified in the SUBSCRIBE
 message MAY be less than the `Largest Object` observed at the publisher. If the
 specified `EndGroup` is the same group specified in `Start`, the remainder of
-that Group passes the filter. `EndGroup` MUST specify the same or a later Group
+that Group passes the filter. `EndGroup` MUST specify the same or a larger Group
 than specified in `Start`.
 
 A filter type other than the above MUST be treated as error.
@@ -1648,7 +1648,7 @@ delivered.
 
 Unlike a new subscription, SUBSCRIBE_UPDATE can not cause an Object to be
 delivered multiple times.  Like SUBSCRIBE, EndGroup MUST specify the
-same or a later object than StartGroup and StartObject.
+same or a larger Object than StartGroup and StartObject.
 
 If a parameter included in `SUBSCRIBE` is not present in
 `SUBSCRIBE_UPDATE`, its value remains unchanged.  There is no mechanism to
@@ -1861,7 +1861,7 @@ The Object Forwarding Preference does not apply to fetches.
 
 Fetch specifies an inclusive range of Objects starting at StartObject
 in StartGroup and ending at EndObject in EndGroup. EndGroup and EndObject MUST
-specify the same or a later object than StartGroup and StartObject.
+specify the same or a larger Object than StartGroup and StartObject.
 
 The format of FETCH is as follows:
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1861,7 +1861,7 @@ The Object Forwarding Preference does not apply to fetches.
 
 Fetch specifies an inclusive range of Objects starting at StartObject
 in StartGroup and ending at EndObject in EndGroup. EndGroup and EndObject MUST
-specify the same or a larger Object than StartGroup and StartObject.
+specify the same or a larger Location than StartGroup and StartObject.
 
 The format of FETCH is as follows:
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1414,7 +1414,7 @@ Latest Object filter.
 AbsoluteStart (0x3):  The Start Location is specified explicitly in the message.
 The Start Location MAY be less than the `Largest Object`. There is no End Group
 - the subscription is open ended.  To receive any object that is published or
-is received after this subscription is processed, a subscriber can use a an
+is received after this subscription is processed, a subscriber can use an
 AbsoluteStart filter with Start Location = {0, 0}.
 
 AbsoluteRange (0x4):  The Start Location and End Group are specified explicitly
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1410,8 +1410,8 @@ prioritization, relays can receive objects with Locations smaller than `Largest
 Object` after the SUBSCRIBE is processed, but these objects do not pass the
 Latest Object filter.
 
-AbsoluteStart (0x3):  The Start Location is specified explicitly in the message.
-The Start Location MAY be less than the `Largest Object`. There is no End Group
+AbsoluteStart (0x3):  The Start Location is specified explicitly in the Subscribe message.
+The Start Location specified in the Subscribe message MAY be less than the `Largest Object` observed at the publisher. There is no End Group
 - the subscription is open ended.  To receive any object that is published or
 is received after this subscription is processed, a subscriber can use an
 AbsoluteStart filter with Start Location = {0, 0}.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1393,7 +1393,7 @@ the publisher to identify which objects need to be delivered.
 
 All filters have a Start Location and an optional End Group.  Only objects
 published or received via a subscription having Locations greater than or
-equal to start and strictly less than than or within the End Group (when
+equal to Start and strictly less than or within the End Group (when
 present) pass the filter.
 
 The `Largest Object` is defined to be the object with the largest Location
@@ -1421,15 +1421,12 @@ AbsoluteRange (0x4):  The Start Location and End Group are specified explicitly
 in the message. The Start Location MAY be less than the `Largest Object`. If End
 Group is the same group specified in Start Location, the remainder of that group
 passes the filter.  EndGroup MUST specify the same or a later group than
-specified in `Start`. 
+specified in `Start`.
 
 A filter type other than the above MUST be treated as error.
 
-Because objects can be reordered on the network, a SUBSCRIBE with the Latest
-Object filter might begin at 
-
-Subscribe only delivers newly published or received Objects.  Objects from
-the past are retrieved using FETCH ({{message-fetch}}).
+Subscribe only delivers newly published or received Objects.  Objects from the
+past are retrieved using FETCH ({{message-fetch}}).
 
 A Subscription can also request a publisher to not forward Objects for a given
 track by setting the `Forward` field to 0. This allows the publisher or relay
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1402,25 +1402,27 @@ processing the SUBSCRIBE message.
 
 There are 3 types of filters:
 
-Latest Object (0x2): The Start Location is `{Largest Object.Group, Largest Object.Object + 1}`. `Largest Object` is
-communicated in SUBSCRIBE_OK. If no content has been delivered yet, the Start
-Location is {0, 0}.  There is no End Group - the
-subscription is open ended.  Note that due to network reordering or
-prioritization, relays can receive objects with Locations smaller than `Largest
-Object` after the SUBSCRIBE is processed, but these objects do not pass the
-Latest Object filter.
+Latest Object (0x2): The filter Start Location is `{Largest Object.Group,
+Largest Object.Object + 1}` and `Largest Object` is communicated in
+SUBSCRIBE_OK. If no content has been delivered yet, the filter Start Location is
+{0, 0}. There is no End Group - the subscription is open ended.  Note that due
+to network reordering or prioritization, relays can receive Objects with
+Locations smaller than  `Largest Object` after the SUBSCRIBE is processed, but
+these Objects do not pass the Latest Object filter.
 
-AbsoluteStart (0x3):  The Start Location is specified explicitly in the Subscribe message.
-The Start Location specified in the Subscribe message MAY be less than the `Largest Object` observed at the publisher. There is no End Group
-- the subscription is open ended.  To receive any object that is published or
-is received after this subscription is processed, a subscriber can use an
-AbsoluteStart filter with Start Location = {0, 0}.
+AbsoluteStart (0x3):  The filter Start Location is specified explicitly in the
+SUBSCRIBE message. The `Start` specified in the SUBSCRIBE message MAY be less
+than the `Largest Object` observed at the publisher. There is no End Group - the
+subscription is open ended.  To receive any Object that is published or is
+received after this subscription is processed, a subscriber can use an
+AbsoluteStart filter with `Start` = {0, 0}.
 
-AbsoluteRange (0x4):  The Start Location and End Group are specified explicitly
-in the message. The Start Location MAY be less than the `Largest Object`. If End
-Group is the same group specified in Start Location, the remainder of that group
-passes the filter.  EndGroup MUST specify the same or a later group than
-specified in `Start`.
+AbsoluteRange (0x4):  The filer Start Location and End Group are specified
+explicitly in the SUBSCRIBE message. The `Start` specified in the SUBSCRIBE
+message MAY be less than the `Largest Object` observed at the publisher. If the
+specified `EndGroup` is the same group specified in `Start`, the remainder of
+that Group passes the filter. `EndGroup` MUST specify the same or a later Group
+than specified in `Start`.
 
 A filter type other than the above MUST be treated as error.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1393,7 +1393,7 @@ the publisher to identify which objects need to be delivered.
 
 All filters have a Start Location and an optional End Group.  Only objects
 published or received via a subscription having Locations greater than or
-equal to Start and strictly less than or within the End Group (when
+equal to Start and strictly less than or equal to the End Group (when
 present) pass the filter.
 
 The `Largest Object` is defined to be the object with the largest Location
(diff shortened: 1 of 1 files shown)


martinduke (approved):
This is fine.

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,42 @@ single session and publishers SHOULD treat this as a protocol violation.
...
-AbsoluteRange (0x4):  Specifies a closed subscription starting at `Start`
-and ending at the largest object in EndGroup.  The start and
-end of the range are inclusive.  EndGroup MUST specify the same or a later
-group than specified in `start`. If the start group is prior to the current
-group, the subscription starts at the beginning of the current object like
-the 'Latest Object' filter.
+Latest Object (0x2): The Start Location is the next object published or received
+via subscription with a Location larger than the `Largest Object`, which is
+communicated in SUBSCRIBE_OK. If no content has been delivered yet, the Start
+Location is the first published or received group.  There is no End Group - the
```
afrind, line 1408: > If no content has been delivered yet, the Start Location is the first published or received group.

This is an old rule, but I think this makes it challenging to update this subscription range later, since you might accidentally move it backwards.  Should we change this to say Start Location is {0,0} when Content Exists=0?

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,42 @@ single session and publishers SHOULD treat this as a protocol violation.
...
+subscription is open ended.  Note that due to network reordering or
+prioritization, relays can receive objects with Locations smaller than `Largest
+Object` after the SUBSCRIBE is processed, but these objects do not pass the
+Latest Object filter.
```
gwendalsimon, line 1412: Along with the clarified definition of Locations (which is based on comparisons between _integer_ values of groupID and objectID), it means that, in [this example of unordered groupID](https://github.com/moq-wg/moq-transport/issues/892#issuecomment-2763852591), the Relay would implement the third interpretation (forward only the objects whose Location is greater than the "Start Location"). Hence, it would be a filter for the incoming objects received from the upstream subscribe. It requires the Relay to store the Start Location for every subscriber but this is fine.

Dealing with a Subscribe Latest Object (17,0), the Relay who receives from upstream (19,0), (8, 2), (32,0) and (1,1) in this temporal order, will only forward (19,0) and (32,0) downstream.

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,42 @@ single session and publishers SHOULD treat this as a protocol violation.
 The subscriber specifies a filter on the subscription to allow
 the publisher to identify which objects need to be delivered.
 
-There are 3 types of filters:
+All filters have a Start Location and an optional End Group.  Only objects
+published or received via a subscription having Locations greater than or
+equal to Start and strictly less than or within the End Group (when
```
ianswett, line 1396: Unless there's a reason you're using within, how about equal to?

```suggestion
equal to Start and strictly less than or equal to the End Group (when
```

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,42 @@ single session and publishers SHOULD treat this as a protocol violation.
...
-and ending at the largest object in EndGroup.  The start and
-end of the range are inclusive.  EndGroup MUST specify the same or a later
-group than specified in `start`. If the start group is prior to the current
-group, the subscription starts at the beginning of the current object like
-the 'Latest Object' filter.
+Latest Object (0x2): The Start Location is the next object published or received
+via subscription with a Location larger than the `Largest Object`, which is
+communicated in SUBSCRIBE_OK. If no content has been delivered yet, the Start
+Location is the first published or received group.  There is no End Group - the
+subscription is open ended.  Note that due to network reordering or
```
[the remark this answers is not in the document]
afrind: Only Latest Object uses Largest Object to create the filter, see the end of the sentence:

> but these objects do not pass the Latest Object filter.

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,42 @@ single session and publishers SHOULD treat this as a protocol violation.
...
+The Start Location MAY be less than the `Largest Object`. There is no End Group
+- the subscription is open ended.  To receive any object that is published or
+is received after this subscription is processed, a subscriber can use an
+AbsoluteStart filter with Start Location = {0, 0}.
+
+AbsoluteRange (0x4):  The Start Location and End Group are specified explicitly
+in the message. The Start Location MAY be less than the `Largest Object`. If End
+Group is the same group specified in Start Location, the remainder of that group
+passes the filter.  EndGroup MUST specify the same or a later group than
+specified in `Start`.
```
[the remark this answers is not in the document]
afrind: You dropped the "MAY be less than Largest Object", I thought that was useful.

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,42 @@ single session and publishers SHOULD treat this as a protocol violation.
...
 
-AbsoluteRange (0x4):  Specifies a closed subscription starting at `Start`
-and ending at the largest object in EndGroup.  The start and
-end of the range are inclusive.  EndGroup MUST specify the same or a later
-group than specified in `start`. If the start group is prior to the current
-group, the subscription starts at the beginning of the current object like
-the 'Latest Object' filter.
+Latest Object (0x2): The Start Location is the next object published or received
+via subscription with a Location larger than the `Largest Object`, which is
+communicated in SUBSCRIBE_OK. If no content has been delivered yet, the Start
```
afrind, line 1407:
```suggestion
Latest Object (0x2): The Start Location is `{Largest Object.Group, Largest Object.Object + 1}`. `Largest Object` is
communicated in SUBSCRIBE_OK. If no content has been delivered yet, the Start
```

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,41 @@ single session and publishers SHOULD treat this as a protocol violation.
...
 
-AbsoluteRange (0x4):  Specifies a closed subscription starting at `Start`
-and ending at the largest object in EndGroup.  The start and
-end of the range are inclusive.  EndGroup MUST specify the same or a later
-group than specified in `start`. If the start group is prior to the current
-group, the subscription starts at the beginning of the current object like
-the 'Latest Object' filter.
+Latest Object (0x2): The Start Location is `{Largest Object.Group, Largest Object.Object + 1}`. `Largest Object` is
+communicated in SUBSCRIBE_OK. If no content has been delivered yet, the Start
+Location is the first published or received group.  There is no End Group - the
```
afrind, line 1407:
```suggestion
Location is {0, 0}.  There is no End Group - the
```

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,41 @@ single session and publishers SHOULD treat this as a protocol violation.
 The subscriber specifies a filter on the subscription to allow
 the publisher to identify which objects need to be delivered.
 
-There are 3 types of filters:
+All filters have a Start Location and an optional End Group.  Only objects
```
suhasHere, line 1394: wonder why is it EndGroup and not End Location?

afrind: #593 removed End Object from Absolute Range.

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,41 @@ single session and publishers SHOULD treat this as a protocol violation.
...
+published or received via a subscription having Locations greater than or
+equal to Start and strictly less than or equal to the End Group (when
+present) pass the filter.
 
-Latest Object (0x2): Specifies an open-ended subscription beginning from
-the current object of the current group.  If no content has been delivered yet,
-the subscription starts with the first published or received group.
+The `Largest Object` is defined to be the object with the largest Location
+({{location-structure}}) in the track from the perspective of the endpoint
+processing the SUBSCRIBE message.
```
suhasHere, line 1401: since the term "LargestObject" is used in multiple places, may be it will benefit from a sub-section so that all the usages refer to the same defintion. It need not be in this PR and we can create a new issue to track, if it helps.

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,41 @@ single session and publishers SHOULD treat this as a protocol violation.
...
+There are 3 types of filters:
 
-AbsoluteRange (0x4):  Specifies a closed subscription starting at `Start`
-and ending at the largest object in EndGroup.  The start and
-end of the range are inclusive.  EndGroup MUST specify the same or a later
-group than specified in `start`. If the start group is prior to the current
-group, the subscription starts at the beginning of the current object like
-the 'Latest Object' filter.
+Latest Object (0x2): The Start Location is `{Largest Object.Group, Largest Object.Object + 1}`. `Largest Object` is
+communicated in SUBSCRIBE_OK. If no content has been delivered yet, the Start
```
suhasHere, line 1406: So that i understand correctly, this means that the current object will never be delivered with filter type `Latest`and the receiver has to wait for the next object to arrive (which may never arrive ?)

I am not sure if we want the above.

afrind: I think that can be up to the implementation.  If it's halfway through 5-2 and has the first part of it cached, it can return Largest Object 5-1 in Subscribe OK, and 5-2 is included in Latest.  I was planning to make the explicit in a follow up to this PR if we can get agreement on the general principles here.

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,41 @@ single session and publishers SHOULD treat this as a protocol violation.
...
+Location is {0, 0}.  There is no End Group - the
```
suhasHere, line 1407:
```suggestion
Location is undefined.  There is no End Group - the
```

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,41 @@ single session and publishers SHOULD treat this as a protocol violation.
...
+subscription is open ended.  Note that due to network reordering or
+prioritization, relays can receive objects with Locations smaller than `Largest
+Object` after the SUBSCRIBE is processed, but these objects do not pass the
+Latest Object filter.
+
+AbsoluteStart (0x3):  The Start Location is specified explicitly in the message.
```
suhasHere, line 1413:
```suggestion
AbsoluteStart (0x3):  The Start Location is specified explicitly in the Subscribe message.
```

draft-ietf-moq-transport.md
suhasHere: One thing i got confused when reading this was, there is a Start (Location) field in the Subscribe message as well. The use of Start Location in this PR refers to publisher's view of filter's start location and not the field in the subscribe message. 

I feel it will help clarify or disambiguate the field vs state in the code on the publisher.

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,41 @@ single session and publishers SHOULD treat this as a protocol violation.
...
+The Start Location MAY be less than the `Largest Object`. There is no End Group
```
suhasHere, line 1414:
```suggestion
The Start Location specified in the Subscribe message MAY be less than the `Largest Object` observed at the publisher. There is no End Group
```

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,41 @@ single session and publishers SHOULD treat this as a protocol violation.
...
+- the subscription is open ended.  To receive any object that is published or
+is received after this subscription is processed, a subscriber can use an
+AbsoluteStart filter with Start Location = {0, 0}.
```
suhasHere, line 1417: Just to confirm this is a way of supporting all objects ?

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,41 @@ single session and publishers SHOULD treat this as a protocol violation.
 The subscriber specifies a filter on the subscription to allow
 the publisher to identify which objects need to be delivered.
 
-There are 3 types of filters:
+All filters have a Start Location and an optional End Group.  Only objects
+published or received via a subscription having Locations greater than or
+equal to Start and strictly less than or equal to the End Group (when
```
suhasHere, line 1396:
```suggestion
equal to Start (unless Start Location is {0,0} ) and strictly less than or equal to the End Group, (when
```

suhasHere (commented):
Thanks for this PR. This seems to be clarifying. I did had few clarificatin questions and some design comments too. Please let me know your thoughts. Thanks

draft-ietf-moq-transport.md
afrind: I don't understand why {0,0} would be an exception?

draft-ietf-moq-transport.md
afrind: Good Point.  I'll disambiguate this.

draft-ietf-moq-transport.md
afrind: Yes, all Objects published or received via subscription pass a filer that starts at {0,0} and has no End Group.

draft-ietf-moq-transport.md
suhasHere: Hmm . I missed this somehow. Thanks for sharing the PR. That PR does say this "Specifies a closed subscription starting at StartObject
in StartGroup and ending at EndObject in EndGroup."

I wonder if removing the EndObject was resolving any issue. Sorry for missing it earlier.

draft-ietf-moq-transport.md
suhasHere: since it represents AllObjects and that would allow unordered groupIds to work which can in turn send groups smaller than the largest.

afrind: Since there are no negative Group IDs, {0,0} is not a special case.  All Objects with Location >= Start pass the filter.

draft-ietf-moq-transport.md
afrind: @ianswett do you have the linked issue for that?

Sy0307:
Maybe adding some cases like subgroup can be clearer to readers?

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,43 @@ single session and publishers SHOULD treat this as a protocol violation.
...
+subscription is open ended.  To receive any Object that is published or is
+received after this subscription is processed, a subscriber can use an
+AbsoluteStart filter with `Start` = {0, 0}.
+
+AbsoluteRange (0x4):  The filer Start Location and End Group are specified
+explicitly in the SUBSCRIBE message. The `Start` specified in the SUBSCRIBE
+message MAY be less than the `Largest Object` observed at the publisher. If the
+specified `EndGroup` is the same group specified in `Start`, the remainder of
+that Group passes the filter. `EndGroup` MUST specify the same or a later Group
+than specified in `Start`.
```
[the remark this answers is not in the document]
afrind: I updated later to larger here and in Subscribe Update.  I think the intent is input validation on the message, so it's clear that larger was the intent.  There's only more more place we use the word _later_ ambiguously, which is in DELIVERY TIMEOUT.  I think we should handle it separately.

sharmafb (approved):

draft-ietf-moq-transport.md
```diff
@@ -1391,31 +1391,43 @@ single session and publishers SHOULD treat this as a protocol violation.
...
+SUBSCRIBE_OK. If no content has been delivered yet, the filter Start Location is
+{0, 0}. There is no End Group - the subscription is open ended.  Note that due
+to network reordering or prioritization, relays can receive Objects with
+Locations smaller than  `Largest Object` after the SUBSCRIBE is processed, but
+these Objects do not pass the Latest Object filter.
+
+AbsoluteStart (0x3):  The filter Start Location is specified explicitly in the
+SUBSCRIBE message. The `Start` specified in the SUBSCRIBE message MAY be less
+than the `Largest Object` observed at the publisher. There is no End Group - the
+subscription is open ended.  To receive any Object that is published or is
```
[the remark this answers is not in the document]
afrind: Yes, every object published or received via subscription will pass a filter with Absolute Start {0,0}, because object.location >= filter.startLocation.

I can change the text to `all Objects that are published or are received...` if that is clearer.

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1636,7 +1648,7 @@ delivered.
 
 Unlike a new subscription, SUBSCRIBE_UPDATE can not cause an Object to be
 delivered multiple times.  Like SUBSCRIBE, EndGroup MUST specify the
-same or a later object than StartGroup and StartObject.
+same or a larger Object than StartGroup and StartObject.
```
englishm, line 1651:
```suggestion
same or a larger Object id than StartGroup and StartObject.
```

consider clarifying that this is not an object with a larger payload length

Issue #931, opened 2025-04-18
FETCH edge cases


martinduke:
1) What if FETCH_OK::end < FETCH::start? 

Is this FETCH_ERROR::invalid range or FETCH_OK with an empty data stream? (and if the latter, what would you put in largest_id, especially if no content exists?)

2) What is FETCH_OK::end if all objects are in a gap? Is it the end of the range, to go with an empty stream?

afrind:
> What if FETCH_OK::end < FETCH::start?

Do you mean FETCH::start > Publisher.Largest Object?  I [currently](https://github.com/facebookexperimental/moxygen/blob/main/moxygen/samples/date/MoQDateServer.cpp#L194-L198) treat that as INVALID_RANGE.  FETCH_OK::end < FETCH::start should maybe be a protocol violation?

> What is FETCH_OK::end if all objects are in a gap? Is it the end of the range, to go with an empty stream?

I think the answer is FETCH_ERROR / NO_OBJECTS.  This is a little awkward, since it's not a true "failure", but opening an empty data stream seems pointless.  Otherwise it would be FETCH::end and an empty data stream.

martinduke:
That's fine with me. Let's clarify this in the spec.

afrind:
The spec has this for the first one:

> If StartGroup/StartObject is greater than the latest published Object group, the publisher MUST return FETCH_ERROR with error code 'No Objects'.

So I added the thing you actually said, which is if the response is invalid.

martinduke:
LGTM

PR #936, opened 2025-04-21
Clarify FETCH_OK fields

afrind:
Also close the session if the publisher returns a bad value in FETCH_OK.

Fixes: #931

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2049,8 +2049,9 @@ A publisher MUST send fetched groups in the determined group order, either
 ascending or descending. Within each group, objects are sent in Object ID order;
 subgroup ID is not used for ordering.
 
-If StartGroup/StartObject is greater than the latest published Object group,
-the publisher MUST return FETCH_ERROR with error code 'No Objects'.
+If StartGroup/StartObject is greater than `Largest Object`
+({{message-subscribe-req}}), or all requested Objects have status `Object Does
+Not Exist`, the publisher MUST return FETCH_ERROR with error code 'No Objects'.
 
 ### Calculating the Range of a Relative Joining Fetch
 
@@ -2120,7 +2121,8 @@ the End Group ID and Object Id indicate the last Object in the track,
   track, the largest known {group,object} at the relay is used.  For tracks
   with a requested end larger than what is cached without an active
   subscription, the relay makes an upstream request in order to satisfy the
-  FETCH.
+  FETCH.  If End is smaller than the Start Location in the corresponding
+  FETCH the receiver MUST close the session with `Protocol Violation`
 
 * Subscribe Parameters: The parameters are defined in {{version-specific-params}}.
 
@@ -3010,6 +3012,10 @@ The Object Status field is only sent if the Object Payload Length is zero.
 
 The Subgroup ID field of an object with a Forwarding Preference of "Datagram"
 (see {{object-properties}}) is set to the Object ID.
+
+A valid FETCH stream always contains at least one Object.  If a subscriber
+receives a FIN on a FETCH stream without any Objects, it MUST close the session
+with a `Protocol Violation`.
 
 ## Examples
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2049,7 +2049,6 @@ A publisher MUST send fetched groups in the determined group order, either
 ascending or descending. Within each group, objects are sent in Object ID order;
 subgroup ID is not used for ordering.
 
-If the StartGroup/StartObject is greater than `Largest Object`
 If StartGroup/StartObject is greater than `Largest Object`
 ({{message-subscribe-req}}), or all requested Objects have status `Object Does
 Not Exist`, the publisher MUST return FETCH_ERROR with error code 'No Objects'.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2328,9 +2328,10 @@ Values of 0x0 and those larger than 0x2 are a protocol error.
    - Otherwise, End Location is {Fetch.End Group, Fetch.End Object}
 
   If the relay is subscribed to the track, it uses its knowledge of the largest
-  {Group, Object} to set End Location.  If if is not subscribed and the requested End
-  Location exceeds its cached data, the relay makes an upstream request to
-  complete the FETCH, and uses the upstream response to set End Location.
+  {Group, Object} to set End Location.  If if is not subscribed and the
+  requested End Location exceeds its cached data, the relay makes an upstream
+  request to complete the FETCH, and uses the upstream response to set End
+  Location.
 
   If End is smaller than the Start Location in the corresponding FETCH the
   receiver MUST close the session with `Protocol Violation`
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2116,13 +2116,17 @@ the End Group ID and Object Id indicate the last Object in the track,
 0 if not.
 
 * End: The largest object covered by the FETCH response.
-  This is the minimum of the {EndGroup,EndObject} specified in FETCH and the
-  largest known {group,object}.  If the relay is currently subscribed to the
-  track, the largest known {group,object} at the relay is used.  For tracks
-  with a requested end larger than what is cached without an active
-  subscription, the relay makes an upstream request in order to satisfy the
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2120,7 +2121,8 @@ the End Group ID and Object Id indicate the last Object in the track,
   track, the largest known {group,object} at the relay is used.  For tracks
   with a requested end larger than what is cached without an active
   subscription, the relay makes an upstream request in order to satisfy the
-  FETCH.
+  FETCH.  If End is smaller than the Start Location in the corresponding
```
ianswett, line 2124: Note: The preceding text for End is quite difficult to read.  I'm not sure if you want to rewrite it now or we should do it in a different PR?

ianswett (approved):

draft-ietf-moq-transport.md
suhasHere: +1 , would be nice to fix in this PR if possible as they are related

suhasHere (approved):
Looks fine , with the nit

martinduke (approved):

draft-ietf-moq-transport.md
```diff
@@ -2115,12 +2116,17 @@ the End Group ID and Object Id indicate the last Object in the track,
...
 * End: The largest object covered by the FETCH response.
-  This is the minimum of the {EndGroup,EndObject} specified in FETCH and the
-  largest known {group,object}.  If the relay is currently subscribed to the
-  track, the largest known {group,object} at the relay is used.  For tracks
-  with a requested end larger than what is cached without an active
-  subscription, the relay makes an upstream request in order to satisfy the
-  FETCH.
+  The End value is determined by taking the minimum of two values:
+   - The {EndGroup, EndObject} specified in the FETCH request.
+   - The largest known {Group, Object} available.
```
afrind, line 2121: I realized the description of "End of Track" above is a little off now.  I think it should be 1 if End is the final object in the track.

Should I update it in this PR?

draft-ietf-moq-transport.md
```diff
@@ -2110,17 +2111,21 @@ FETCH_OK
...
+  the End Location is the final Object in the Track, 0 if not.
 
 * End: The largest object covered by the FETCH response.
-  This is the minimum of the {EndGroup,EndObject} specified in FETCH and the
-  largest known {group,object}.  If the relay is currently subscribed to the
-  track, the largest known {group,object} at the relay is used.  For tracks
-  with a requested end larger than what is cached without an active
-  subscription, the relay makes an upstream request in order to satisfy the
-  FETCH.
+  The End value is determined by taking the minimum of two values:
```
afrind, line 2118: This is still ambiguous whether {group,object + 1} is used in FETCH_OK.  In my implementation, I think it's easiest to use + 1, and I _think_ End.object == 0 is required to correctly denote the end of the group when there is no END_OF_GROUP marker

```
S->P: FETCH start=10,0 end=10,0  // requesting all of group 10
P->S: FETCH_OK, 10,0
objects 10,0 - 10,10             // 10,10 is the last object in the group
```
If FETCH_OK had 10/10 (actual end) it is ambiguous with the case where 10/10 is the Largest object.

fluffy (changes requested):
This make my code for handling the no objects case totally different than the handling one object. I would rather get rid of of the no objects error and say if there are no objects, the fetch stream will have no objects.

afrind:
> This make my code for handling the no objects case totally different than the handling one object. I would rather get rid of of the no objects error and say if there are no objects, the fetch stream will have no objects.

I'd be ok with that though it's a bit wasteful to open the stream in this case.  In any case, I don't want to disallow empty FETCH streams because it means delaying sending OK or ERROR until I know I have at least one object to send.

draft-ietf-moq-transport.md
```diff
@@ -3282,6 +3287,10 @@ The Object Status field is only sent if the Object Payload Length is zero.
 The Subgroup ID field of an object with a Forwarding Preference of "Datagram"
 (see {{object-properties}}) is set to the Object ID.
 
+A valid FETCH stream always contains at least one Object.  If a subscriber
+receives a FIN on a FETCH stream without any Objects, it MUST close the session
+with a `Protocol Violation`.
+
```
afrind, line 3293:
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -2315,17 +2316,21 @@ FETCH_OK Message {
...
-  This is the minimum of the {End Group,End Object} specified in FETCH and the
-  largest known {group,object}.  If the relay is currently subscribed to the
-  track, the largest known {group,object} at the relay is used.  For tracks
-  with a requested end larger than what is cached without an active
-  subscription, the relay makes an upstream request in order to satisfy the
-  FETCH.
+  The End value is determined by taking the minimum of two values:
+   - The {End Group, End Object} specified in the FETCH request.
+   - The largest known, possibly final, {Group, Object} available
+
```
afrind, line 2326:
```suggestion
  The End Location is determined as follows:
   - If End Object in the FETCH request was 0 and the response covers the last
     Object in the Group, End Location is {Fetch.End Group, 0}
   - If the requested FETCH End was beyond the Largest known (possibly final)
     Object, End Location is {Largest.Group, Largest.Object + 1}
   - Otherwise, End Location is {Fetch.End Group, Fetch.End Object} 

```

draft-ietf-moq-transport.md
```diff
@@ -2315,17 +2316,24 @@ FETCH_OK Message {
...
-  with a requested end larger than what is cached without an active
-  subscription, the relay makes an upstream request in order to satisfy the
-  FETCH.
+  The End Location is determined as follows:
+   - If End Object in the FETCH request was 0 and the response covers the last
+     Object in the Group, End Location is {Fetch.End Group, 0}
+   - If the requested FETCH End was beyond the Largest known (possibly final)
+     Object, End Location is {Largest.Group, Largest.Object + 1}
+   - Otherwise, End Location is {Fetch.End Group, Fetch.End Object} 
+
```
afrind, line 2329: If the track is over, the final object is 5, 3, and the requested end is 5,0, then this text allows both:

`End=5, 0` and `End of Track=1, End=5, 4` as valid FETCH_OK.  I'd like to make only a single way to encode that, without making the rules too complicated.

draft-ietf-moq-transport.md
```diff
@@ -2315,17 +2316,24 @@ FETCH_OK Message {
...
-  FETCH.
+  The End Location is determined as follows:
+   - If End Object in the FETCH request was 0 and the response covers the last
+     Object in the Group, End Location is {Fetch.End Group, 0}
+   - If the requested FETCH End was beyond the Largest known (possibly final)
+     Object, End Location is {Largest.Group, Largest.Object + 1}
+   - Otherwise, End Location is {Fetch.End Group, Fetch.End Object}
+
+  If the relay is subscribed to the track, it uses its knowledge of the largest
+  {Group, Object} to set End.  If if is not subscribed and the requested End
```
ianswett, line 2331:
```suggestion
  {Group, Object} to set End Location.  If if is not subscribed and the requested End
```

ianswett (approved):
Two nits, but LG

draft-ietf-moq-transport.md
```diff
@@ -2315,17 +2316,25 @@ FETCH_OK Message {
...
-  track, the largest known {group,object} at the relay is used.  For tracks
-  with a requested end larger than what is cached without an active
-  subscription, the relay makes an upstream request in order to satisfy the
-  FETCH.
+  The End Location is determined as follows:
+   - If End Object in the FETCH request was 0 and the response covers the last
+     Object in the Group, End Location is {Fetch.End Group, 0}
+   - If the requested FETCH End was beyond the Largest known (possibly final)
+     Object, End Location is {Largest.Group, Largest.Object + 1}
+   - Otherwise, End Location is {Fetch.End Group, Fetch.End Object}
```
afrind, line 2328:
```suggestion
   - If the requested FETCH End Location was beyond the Largest known (possibly
     final) Object, End Location is {Largest.Group, Largest.Object + 1}
   - If End Location.Object in the FETCH request was 0 and the response covers
     the last Object in the Group, End Location is {Fetch.End Location.Group, 0}
   - Otherwise, End Location is Fetch.End Location
```

draft-ietf-moq-transport.md
afrind: I reordered the rules to prefer "End of Track=1, End Location=5,4" interpretation.

PR #937, opened 2025-04-21
Clarify how Latest Object works when the Largest Object is incomplete

afrind:
Fixes: #592

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1494,15 +1494,15 @@ There are 3 types of filters:
 Latest Object (0x2): The filter Start Location is `{Largest Object.Group,
 Largest Object.Object + 1}` and `Largest Object` is communicated in
 SUBSCRIBE_OK. If no content has been delivered yet, the filter Start Location is
-{0, 0}. There is no End Group - the subscription is open ended.  
-If an implementation has published or received the beginning of a new `Largest
-Object` but not the end, it can choose either this "current" Object or the
-previous `Largest Object` to create this filter.  If it chooses the previous
-value, the current Object passes the Latest Object filter, and it MUST be
-delivered starting from the beginning. Note that due to network reordering or
-prioritization, relays can receive Objects with Locations smaller than
-`Largest Object` after the SUBSCRIBE is processed, but these Objects do not pass
-the Latest Object filter.
+{0, 0}. There is no End Group - the subscription is open ended. If an
+implementation has published or received the beginning of a new `Largest Object`
+but not the end, it can choose either this "current" Object or the previous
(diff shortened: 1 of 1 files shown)


ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1494,10 +1494,15 @@ There are 3 types of filters:
 Latest Object (0x2): The filter Start Location is `{Largest Object.Group,
 Largest Object.Object + 1}` and `Largest Object` is communicated in
 SUBSCRIBE_OK. If no content has been delivered yet, the filter Start Location is
-{0, 0}. There is no End Group - the subscription is open ended.  Note that due
-to network reordering or prioritization, relays can receive Objects with
-Locations smaller than  `Largest Object` after the SUBSCRIBE is processed, but
-these Objects do not pass the Latest Object filter.
+{0, 0}. There is no End Group - the subscription is open ended. If an
+implementation has published or received the beginning of a new `Largest Object`
+but not the end, it can choose either this "current" Object or the previous
```
gloinul, line 1499: Is it really smart to have an either or definition here. Isn't it better to have a single definition. I understand that there are a timing aspect here. But, wouldn't it be clearer to define this to be either largest fully received object, or largest that has started to be received, and not make it either or.

ianswett:
At the virtual interim, there was interest in increasing the largest Object value as soon as the beginning of that Object arrives and not giving publishers the option of whether to deliver an in progress Object.

Also, there was interest in making sure all the cases largest Object is used in the draft are clear and more of them may need updating.

draft-ietf-moq-transport.md
```diff
@@ -1494,10 +1494,15 @@ There are 3 types of filters:
...
+`Largest Object` to create this filter.  If it chooses the previous value, the
+current Object passes the Latest Object filter, and it MUST be delivered
+starting from the beginning. Note that due to network reordering or
+prioritization, relays can receive Objects with Locations smaller than `Largest
+Object` after the SUBSCRIBE is processed, but these Objects do not pass the
+Latest Object filter.
```
afrind, line 1507:
```suggestion
{0, 0}. There is no End Group - the subscription is open ended.  Note that due
to network reordering or prioritization, relays can receive Objects with
Locations smaller than  `Largest Object` after the SUBSCRIBE is processed, but
these Objects do not pass the Latest Object filter.
```

afrind:
@suhasHere @vasilvv I rewrote this based on feedback during the interim which is that "largest" gets bumped as soon as possible.  Please take another look.

vasilvv (approved):

suhasHere (approved):

ianswett (approved):

Issue #889, opened 2025-03-27
Ranged DNE Objects


martinduke:
Individual comment:

It is helpful to subscriber/relay accounting to not have gaps of omission in group and object IDs (at least when group IDs increase with time). Hence, the Group Does Not Exist and Object DNE object status. These are useful even if not required.

However, a potential obstacle to this is the overhead of sending potentially very large numbers of Does Not Exist status objects.

A fix for this would be to allow range in these messages, i.e. the status code would be followed by a varint specifying the number of object IDs that are missing. It could go in the same subgroup as the preceding object ID.

For example, an original publisher could produce:
Group 3, Subgroup 0, Object 0.
Group 3, Subgroup 0, Object 1, ObjectDoesNotExist, Range: 1000, FIN
Group 3, Subgroup 1, Object 1001.

Since we have accepted that group IDs not increase with time, we can't make this inclusion of status objects mandatory for Groups. However, since object IDs *do* increase with time, we could do so for Object Does Not Exist, which I would support. However, the mandate question is separate from allowing the encoding.

An immediately apparent use case is Group IDs that are timestamps, where gaps can be arbitrarily large.

afrind:
FETCH fixes this problem because a gap of arbitrary size is encoded implicitly. Because of DELIVERY_TIMEOUT, putting these gap messages in subscriptions might have limited utility - you might not get the gap message.

gwendalsimon:
The Section 9.1.1.1. about Object Status would require some extra-information. Is it mandatory for the Original Publisher to generate all these objects with 0 payload and only Object_Status? Or is it at its discretion?

Should the 0-payload Objects be in a Subgroup? especially the DNE Objects which may be the only one in the Group.

It is said that Object with DNE "_SHOULD be cached_", I don't know how the Range DNE Objects would be cached in that case.

Can we say both DNE and End_of_Group? It is said that Object_Status is an "enumeration" but the format of Object header is not specified.

afrind:
See also #731 / #922, which could communicate a large group gap with a few bytes.

martinduke:
> FETCH fixes this problem because a gap of arbitrary size is encoded implicitly. Because of DELIVERY_TIMEOUT, putting these gap messages in subscriptions might have limited utility - you might not get the gap message.

Well, that depends on what the

afrind:
#922 has landed, do we want the equivalent but for Object IDs (Prior Object Gap)?

martinduke:
[...]

PR #938, opened 2025-04-21
Explain malformed tracks and what to do with them

afrind:
Fixes: #923

PR #939, opened 2025-04-22
Prior Object ID Gap Extension header

afrind:
Similar to Prior Group ID Gap.  I rewrote the text on the malformed conditions, and also changed the code points to be in the 1 byte range.

Fixes: #889

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3598,11 +3598,12 @@ The following Object Extension Headers are defined in MOQT.
 
 Prior Group ID Gap (Extension Header Type 0x3C) is a variable length integer
 containing the number of Groups prior to the current Group that do not and will
-never exist. For example, if the Original Publisher is publishing an Object in
-Group 7 and knows it will never publish any Objects in Group 8 or Group 9, it
-can include Prior Group ID Gap = 2 in any number of Objects in Group 10, as it
-sees fit.  A Track is considered malformed (see {{malformed-tracks}} if any of
-the following conditions are detected:
+never exist. This is equivalent to receiving an `End of Group` status with
+Object ID 0 for each skipped Group. For example, if the Original Publisher is
+publishing an Object in Group 7 and knows it will never publish any Objects in
+Group 8 or Group 9, it can include Prior Group ID Gap = 2 in any number of
+Objects in Group 10, as it sees fit.  A Track is considered malformed (see
+{{malformed-tracks}} if any of the following conditions are detected:
 
  * An Object contains more than one instance of Prior Group ID Gap
  * A Group contains more than one Object with different values for Prior Group
@@ -3625,10 +3626,12 @@ relays. This extension MUST NOT be modified or removed.
 
 Prior Object ID Gap (Extension Header Type 0x3E) is a variable length integer
 containing the number of Objects prior to the current Object that do not and
-will never exist. For example, if the Original Publisher is publishing Object 10
-in Group 3 and knows it will never publish Objects 8 or 9 in this Group, it can
-include Prior Object ID Gap = 2.  A Track is considered malformed (see
-{{malformed-tracks}} if any of the following conditions are detected:
+will never exist. This is equivalent to receiving an `Object Does Not Exist`
+status for each skipped Object ID. For example, if the Original Publisher is
+publishing Object 10 in Group 3 and knows it will never publish Objects 8 or 9
+in this Group, it can include Prior Object ID Gap = 2.  A Track is considered
+malformed (see {{malformed-tracks}} if any of the following conditions are
+detected:
 
  * An Object contains more than one instance of Prior Object ID Gap
  * An Object has a Prior Object ID Gap larger than the Object ID


ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -3077,15 +3077,22 @@ The following Object Extension Headers are defined in MoQT.
...
 7 and knows it will never publish any Objects in Group 8 or Group 9, it can
 include Prior Group ID Gap = 2 in any number of Objects in Group 10, as it sees
-fit.  A track with a Group that contains more than one Object with different
-values for Prior Group ID Gap or has a Prior Group ID Gap larger than the Group
-ID is considered malformed.  If an endpoint receives an Object with a Group ID
-within a previously communicated gap it also treats the track as malformed.
+fit.  A Track is considered malformed if any of the following conditions are
+detected:
+
+ * An Object contains more than one instance of Prior Group ID Gap
```
gloinul, line 3088: Is this overlapping with #938?

draft-ietf-moq-transport.md
```diff
@@ -3095,6 +3102,30 @@ cannot infer any information about the existence of prior groups (see
 This extension can be added by the Original Publisher, but MUST NOT be added by
 relays. This extension MUST NOT be modified or removed.
 
+## Prior Object ID Gap
+
+Prior Object ID Gap (Extension Header Type 0x3E) is a variable length integer
+containing the number of Objects prior to the current Object that do not and
+will never exist. For example, if the Original Publisher published Object 10 in
```
gloinul, line 3109: Wrong tempus here? Isn't the Prior Object ID Gap extension header related to Object 10, and thus it should say, when publishing object 10.

draft-ietf-moq-transport.md
```diff
@@ -3095,6 +3102,30 @@ cannot infer any information about the existence of prior groups (see
...
+Group 3 and knows it will never publish Objects 8 or 9 in this Group, it can
+include Prior Object ID Gap = 2.  A Track is considered malformed if any of the
+following conditions are detected:
+
+ * An Object contains more than one instance of Prior Object ID Gap
```
gloinul, line 3114: Should this be collected together with the ones in #938

ianswett:
We discussed this at the interim, and there might be some tricky aspects here since a Group can have multiple Subgroups.

ianswett (commented):
Let's think about this one more.  Even with subgroups, I think it's likely ok, but I haven't thought of all the cases.

draft-ietf-moq-transport.md
```diff
@@ -3077,15 +3077,22 @@ The following Object Extension Headers are defined in MoQT.
...
+detected:
+
+ * An Object contains more than one instance of Prior Group ID Gap
+ * A Group contains more than one Object with different values for Prior Group
+    ID Gap
+ * An Object has a Prior Group ID Gap larger than the Group ID
+ * An endpoint receives an Object with a Prior Group ID Gap covering an Object
+   it previously received
+ * An endpoint receives an Object with a Group ID within a previously
+   communicated gap
```
fluffy, line 3095: The malformed stuff should be in it's own PR - Can we remove from here.

fluffy (approved):
LGTM me but I think some malformed stuff may have leaked into this

draft-ietf-moq-transport.md
afrind: I don't think this malformed text should be moved because this PR is introducing new ways a track can be malformed.

afrind:
We may need to split this into "Prior Object ID Gap [Subgroup]" and "Prior Object ID Gap [Group]"

afrind:
If we use delta encoding for object IDs on the same subgroup, then it will make it clear the the Prior Object ID Gap extension is across the entire track.  Parking this until we fix #1029.

suhasHere (approved):

draft-ietf-moq-transport.md
afrind: For now, I'm including the malformed track conditions introduced by extensions in the sections where the extensions are defined, in some ways to create a model for extensions defined outside of the base draft that might also introduce malformed track conditions.

draft-ietf-moq-transport.md
afrind: See above.  For now, I think no.

afrind (commented):
I've updated this now that

1) malformed tracks PR has landed
2) object IDs are now delta encoded in subgroup streams

Ready for another round of review

draft-ietf-moq-transport.md
```diff
@@ -3592,16 +3596,22 @@ The following Object Extension Headers are defined in MOQT.
 
 ## Prior Group ID Gap
 
-Prior Group ID Gap (Extension Header Type 0x40) is a variable length integer
+Prior Group ID Gap (Extension Header Type 0x3C) is a variable length integer
```
ianswett, line 3599: Why did you move this codepoint?

draft-ietf-moq-transport.md
```diff
@@ -3592,16 +3596,22 @@ The following Object Extension Headers are defined in MOQT.
...
-ID is considered malformed.  If an endpoint receives an Object with a Group ID
-within a previously communicated gap it also treats the track as malformed
-(see {{malformed-tracks}}.
+never exist. For example, if the Original Publisher is publishing an Object in
+Group 7 and knows it will never publish any Objects in Group 8 or Group 9, it
+can include Prior Group ID Gap = 2 in any number of Objects in Group 10, as it
+sees fit.  A Track is considered malformed (see {{malformed-tracks}} if any of
+the following conditions are detected:
+
+ * An Object contains more than one instance of Prior Group ID Gap
```
ianswett, line 3608: This is a bit redundant, because we say that unless specified, it's illegal for a header to appear twice in an Object.

draft-ietf-moq-transport.md
```diff
@@ -3592,16 +3596,22 @@ The following Object Extension Headers are defined in MOQT.
...
 containing the number of Groups prior to the current Group that do not and will
-never exist. For example, if the Original Publisher published an Object in Group
-7 and knows it will never publish any Objects in Group 8 or Group 9, it can
-include Prior Group ID Gap = 2 in any number of Objects in Group 10, as it sees
-fit.  A track with a Group that contains more than one Object with different
-values for Prior Group ID Gap or has a Prior Group ID Gap larger than the Group
-ID is considered malformed.  If an endpoint receives an Object with a Group ID
-within a previously communicated gap it also treats the track as malformed
-(see {{malformed-tracks}}.
+never exist. For example, if the Original Publisher is publishing an Object in
```
ianswett: Can/should we add some text about how this gap SHOULD be placed in the Object that is most likely to be delivered, commonly the first Object of the Group?  Or something similar?

draft-ietf-moq-transport.md
```diff
@@ -3611,6 +3621,30 @@ cannot infer any information about the existence of prior groups (see
 This extension can be added by the Original Publisher, but MUST NOT be added by
 relays. This extension MUST NOT be modified or removed.
 
+## Prior Object ID Gap
+
+Prior Object ID Gap (Extension Header Type 0x3E) is a variable length integer
```
ianswett, line 3627: Should we say something about how receiving this or the other header is equivalent to receiving Object Status of Missing for the Groups/Objects?

draft-ietf-moq-transport.md
afrind: To keep it in the 1 byte range (0x40 is 64 which is the smallest 2 byte varint, for now)

draft-ietf-moq-transport.md
afrind: Not necessarily, because the gap could exist on any object in the subgroup?

draft-ietf-moq-transport.md
afrind: We do?  I don't see it here: https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#name-object-extension-header

draft-ietf-moq-transport.md
afrind: Sure

afrind: I did this but it's slightly awkward for Group, since we killed Group Does Not Exist.

draft-ietf-moq-transport.md
ianswett: Apparently I was thinking of Parameters: Senders MUST NOT repeat the same parameter type in a message unless the parameter definition explicitly allows multiple instances of that type to be sent in a single message.

I think we should be consistent between the two, otherwise it adds unnecessary confusion.

ianswett: I opened #1121

draft-ietf-moq-transport.md
ianswett: Agreed, but it's fine for now.

Issue #940, opened 2025-04-23
Is No Objects the correct error code when FETCH start > Largest Object?


afrind:
> If StartGroup/StartObject is greater than the latest published Object group, the publisher MUST return FETCH_ERROR with error code 'No Objects'.

If the Largest Object is (10,10) and I make a FETCH for (10,11), and the publisher returns NO_OBJECTS, then I should be able to cache Object Does Not Exist for (10,11).  This doesn't seem quite right, since the publisher can publish (10,11) later.

I can't remember our rationale for returning NO_OBJECTS past the known Largest Object, but it seems like a different error code - either `Invalid Range` or a new code like `Future Range` would be better.

There's a sub-case which is when the track is _complete_ and the FETCH start is past the end of the track - No Objects is also bad here because I need to cache that, it would be better to error the request in some way.

Issue #844, opened 2025-03-24
Improve discussion of caching


vasilvv:
We mention caching multiple times, but never elaborate how it works, and in what context.  We should probably have a dedicated section for that.

From my understanding, SUBSCRIBE and FETCH can both cache-fill, but only FETCH uses cache.  Is that correct?

We assert that once in the cache, the object never changes, but we also talk about how relays MAY "update the cache", which does not appear to be consistent with immutability.

afrind:
> SUBSCRIBE and FETCH can both cache-fill, but only FETCH uses cache. Is that correct?

This is my interpretation.

> MAY "update the cache", which does not appear to be consistent with immutability.

I created a caching section in that talks about what can be updated in #745.

gwendalsimon:
> > SUBSCRIBE and FETCH can both cache-fill, but only FETCH uses cache. Is that correct?
> 
> This is my interpretation.

If, among multiple concurrent SUBSCRIBEs, one of them is far behind live, then the Relay can use cached Objects to serve the late SUBSCRIBE. 

For SUBSCRIBE at Relay side, I see two designs:
- either the Relay stores the Object it receives from Publisher in a memory until this Object has been sent to _all_ Subscribers (including the late ones). Let's call it a "Subscribe_memory". 
- or the Relay stores the Objects it receives from Publisher directly in the cache. It then serves all "late" Subscribers from the cache.

In the first case, the Relay needs to create two copies of the Object: one in the "Subscribe_memory" and one in the Cache for future FETCH. The second option may be quicker (it depends on the I/O of the cache vs I/O of the "Subscribe_memory").

afrind:
In my implementation, all object payloads are refcounted so there's no difference between memory in the cache and memory queued for transmission, but I get that not all implementations are like that.

I'm trying to envision how to implement SUBSCRIBE as pulling from a cache - do you queue the group ID/object ID for transmission, and do a cache lookup when it's time to send it?  What happens if the cache has evicted that item?

PR #946, opened 2025-04-24
Expand cases where FETCH returns Invalid Range

afrind:
This changes the FETCH_ERROR code when Start is beyond Largest Object, beyong the end of a completed track, or no objects have been published.

Fixes: #940

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2088,7 +2088,8 @@ stream indicate objects that do not exist (eg: they implicitly have status
 between the first requested object and the first object in the stream; between
 objects in the stream; and between the last object in the stream and the Largest
 Group/Object indicated in FETCH_OK, so long as the fetch stream is terminated by
-a FIN.
+a FIN.  If no Objects exist in the requested range, the publisher returns
+FETCH_ERROR with code `No Objects`.
 
 **Fetch Types**
 
@@ -2217,7 +2218,7 @@ ascending or descending. Within each group, objects are sent in Object ID order;
 subgroup ID is not used for ordering.
 
 If Start Group/Start Object is greater than the latest published Object group,
-the publisher MUST return FETCH_ERROR with error code 'No Objects'.
+the publisher MUST return FETCH_ERROR with error code 'Invalid Range'.
 
 ### Calculating the Range of a Relative Joining Fetch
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2358,10 +2359,11 @@ as defined below:
...
 * Track Does Not Exist - The requested track is not available at the publisher.
 
-* Invalid Range - The end of the requested range is earlier than the beginning.
+* Invalid Range - The end of the requested range is earlier than the beginning,
+  the start of the requested range is beyond the Largest Object, or the track
+  has not published any Objects yet.
 
-* No Objects - The beginning of the requested range is after the latest group
-  and object for the track, or the track has not published any objects.
+* No Objects - No Objects exist between the requested Start and End Locations.
```
ianswett, line 2366: Q: Is this a cacheable response?  ie: If a downstream subsscriber issues the same FETCH, can I respond with No Objects?

ianswett (approved):

draft-ietf-moq-transport.md
afrind: We have this for Object Status:

> 0x1 := Indicates Object does not exist. Indicates that this object does not exist at any publisher and it will not be published in the future. This SHOULD be cached.

Since "No Objects" is shorthand for "all objects in a range have status 'Object Does Not Exist'", yes, this is cacheable.  If not then there would be DoS concerns for any track with a gap.  We might need to add a bit more text about caching see #844.

draft-ietf-moq-transport.md
```diff
@@ -2358,10 +2359,11 @@ as defined below:
 
 * Track Does Not Exist - The requested track is not available at the publisher.
 
-* Invalid Range - The end of the requested range is earlier than the beginning.
+* Invalid Range - The end of the requested range is earlier than the beginning,
+  the start of the requested range is beyond the Largest Object, or the track
+  has not published any Objects yet.
```
suhasHere, line 2364: nit, may be we need to reword this as it is ambiguous as stated

afrind: I'm not sure which case is ambiguous

draft-ietf-moq-transport.md
```diff
@@ -2358,10 +2359,11 @@ as defined below:
 
 * Track Does Not Exist - The requested track is not available at the publisher.
 
-* Invalid Range - The end of the requested range is earlier than the beginning.
+* Invalid Range - The end of the requested range is earlier than the beginning,
```
gwendalsimon, line 2362: _smaller_ rather than _earlier_?

Issue #942, opened 2025-04-23
Clarify that subscriptions with Forward=0 still get SUBSCRIBE_DONE


afrind:
> A Subscription can also request a publisher to not forward Objects for a given track by setting the Forward field to 0. 

It's kind of implied that only _objects_ are withheld, but I think an explicit callout that SUBSCRIBE_DONE is forwarded even when forward=0.

PR #947, opened 2025-04-24
Clarify Forward=0 does not affect SUBSCRIBE_DONE

afrind:
Fixes: #942

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1674,14 +1674,15 @@ Subscribe only delivers newly published or received Objects.  Objects from the
 past are retrieved using FETCH ({{message-fetch}}).
 
 A Subscription can also request a publisher to not forward Objects for a given
-track by setting the `Forward` field to 0. This allows the publisher or relay
-to prepare to serve the subscription in advance, reducing the time to
-receive objects in the future. Relays SHOULD set the `Forward` flag to 1 if a
-new subscription needs to be sent upstream, regardless of the value of the
-`Forward` field from the downstream subscription. Subscriptions that are not
-forwarded consume resources from the publisher, so a publisher might
-deprioritize, reject, or close those subscriptions to ensure other
-subscriptions can be delivered.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1674,14 +1674,15 @@ Subscribe only delivers newly published or received Objects.  Objects from the
...
 A Subscription can also request a publisher to not forward Objects for a given
-track by setting the `Forward` field to 0. This allows the publisher or relay
-to prepare to serve the subscription in advance, reducing the time to
-receive objects in the future. Relays SHOULD set the `Forward` flag to 1 if a
-new subscription needs to be sent upstream, regardless of the value of the
-`Forward` field from the downstream subscription. Subscriptions that are not
-forwarded consume resources from the publisher, so a publisher might
-deprioritize, reject, or close those subscriptions to ensure other
-subscriptions can be delivered.
+track by setting the `Forward` field to 0. The SUBCRIBE_DONE
```
ianswett, line 1677: nit: I'd probably stick this at the end of the paragraph, since it seems sort of obvious.  I mean, Group Order doesn't change SUBSCRIBE_DONE either

ianswett (approved):
This is ok, but I don't see why it's necessary?  We don't explicitly say that Group Order doesn't change SUBCRIBE_DONE, but I think it doesn't?

draft-ietf-moq-transport.md
```diff
@@ -1674,14 +1674,15 @@ Subscribe only delivers newly published or received Objects.  Objects from the
...
+({{message-subscribe-done}}) control message is not affected by this flag.
```
ianswett, line 1678:
```suggestion
track by setting the `Forward` field to 0. Control messages, such as SUBCRIBE_DONE
({{message-subscribe-done}}) are still sent.
```

ianswett (approved):

suhasHere (approved):

Issue #943, opened 2025-04-23
Relays need to reset open streams for subscriptions with forward=0 when receiving an object


afrind:
Consider this case:

S->P: SUBSCRIBE, forward=1
P->S: SUB_OK
P->S: Object (g=0, sg=0, o=0)
S->P: SUBSCRIBE_UPDATE forward=0
P: receives g=0, sg=0, o=1
S->P: SUBSCRIBE_UPDATE forward=1
P->S: Object (g=0, sg=0, o=2)

The receiver will get o=0, o=2 on the same stream, which is not allowed.

Instead, the relay/publisher MUST reset the open subgroups on paused subscriptions when a new subgroup object arrives.

PR #948, opened 2025-04-24
Explain when a relay can reuse or must reset a subgroup

afrind:
Fixes: #610 
Fixes: #943

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3134,7 +3134,7 @@ with a reliable_size equal to the length of the stream header. This explicitly
 tells the receiver there is an unsent Subgroup.
 
 A relay MUST NOT forward an Object on an existing Subgroup stream unless it is
-the next Object in that Subgroup.  A relay knows that that an Object is the next
+the next Object in that Subgroup.  A relay knows that an Object is the next
 Object in the Subgroup if
  * the Object ID is one greater than the previous Object sent on this Subgroup
    stream.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3133,6 +3133,19 @@ a SUBGROUP_HEADER on a new stream, with no objects, and then send RESET_STREAM_A
 with a reliable_size equal to the length of the stream header. This explicitly
 tells the receiver there is an unsent Subgroup.
 
+A relay MUST NOT forward an Object on an existing Subgroup stream unless it is
+the next Object in that Subgroup.  A relay can know that that an Object is next
+Object in the Subgroup if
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -3133,6 +3133,19 @@ a SUBGROUP_HEADER on a new stream, with no objects, and then send RESET_STREAM_A
 with a reliable_size equal to the length of the stream header. This explicitly
 tells the receiver there is an unsent Subgroup.
 
+A relay MUST NOT forward an Object on an existing Subgroup stream unless it is
+the next Object in that Subgroup.  A relay can know that that an Object is next
+Object in the Subgroup if
```
ianswett, line 3138:
```suggestion
the next Object in that Subgroup.  A relay knows that that an Object is the next
Object in the Subgroup if
```

afrind:
> This LG, but I realize this logic is somewhat complex.

In practice, I think relays will do the easy thing -- forward if it came on the same stream, if it's the next sequential ID, or there's a gap with Prior Object ID Gap.  Otherwise reset.  No one is going to keep score of all objects going by.

draft-ietf-moq-transport.md
```diff
@@ -3133,6 +3133,19 @@ a SUBGROUP_HEADER on a new stream, with no objects, and then send RESET_STREAM_A
...
 
+A relay MUST NOT forward an Object on an existing Subgroup stream unless it is
+the next Object in that Subgroup.  A relay knows that that an Object is the next
+Object in the Subgroup if
+ * the Object ID is one greater than the previous Object sent on this Subgroup
+   stream.
+ * the Object was received on the same upstream Subgroup stream as the
+   previously sent Object, with no other Objects in between.
+ * it knows all Object IDs between the current and previous Object IDs
+   on the Subgroup stream belong to different Subgroups or do not exist.
```
suhasHere, line 3144: May be we should add the fetch case too here ?

draft-ietf-moq-transport.md
```diff
@@ -3133,6 +3133,19 @@ a SUBGROUP_HEADER on a new stream, with no objects, and then send RESET_STREAM_A
 with a reliable_size equal to the length of the stream header. This explicitly
 tells the receiver there is an unsent Subgroup.
 
+A relay MUST NOT forward an Object on an existing Subgroup stream unless it is
+the next Object in that Subgroup.  A relay knows that that an Object is the next
+Object in the Subgroup if
```
suhasHere, line 3138: IIUC , these are OR conditions and not AND conditions. If so, can we clarify by adding "or" at the end of each option

martinduke (approved):
This is gross, but I agree it is correct.

PR #949, opened 2025-04-25
Fetch Object serialization optimization

### LINKED ISSUE #660: Fetch Object serialization isn't efficient
Objects in a Fetch stream have to be in increasing Group and Object ID.

Currently it's:
```
{
  Group ID (i),
  Subgroup ID (i),
  Object ID (i),
  Publisher Priority (8),
  Object Payload Length (i),
  [Object Status (i)],
  Object Payload (..),
}
```

I would expect Group ID to very often be the same as the previous Object and Object ID to be one larger than the previous Object.

The simplest option might be to delta encode Group ID and Object ID. Anytime Group ID increases, the Object ID is an absolute value and not relative.

Additionally or alternatively, one could have a codepoint for "The Group ID is the same, the Subgroup ID is the same, and Object ID increases by 1."

I'm also happy to park this for now, but we should probably improve it at some point.

wilaw:
This PR adds a 1-byte Serialization field to the FETCH object definition and defines 4 code points: 

0x0 - status quo - Group/Subgroup and Object ID fields are present. This would be used at the start of a track and adds 1 byte to the current serialization.

0x1 - The Group ID and Subgroup ID of this object are identical to those of the prior object and Object ID increments by 1. The Group ID, Subgroup ID and Object ID fields are not present.  This code point would most commonly occur in a FETCH response and would save 2 bytes per object.

0x2 - The Group ID field exists and Subgroup ID and Object ID are both 0. The SubGroup ID and Object ID fields are not present.  This code point would be used at the start of each new Group and saves 1 byte per object.

0x3 - The Group ID of this object is identical to that of the prior Object. Subgroup ID increments by 1 and Object ID increments by 1. The Subgroup ID and Object ID fields are not present. This code point would be used at the start of a new Subgroup and would save 1 byte per object. 

Fixes #660

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3221,9 +3221,10 @@ Each object sent on a fetch stream after the FETCH_HEADER has the following form
 
 ~~~
 {
-  Group ID (i),
-  Subgroup ID (i),
-  Object ID (i),
+  Serialization (8),
+  [Group ID (i),]
+  [Subgroup ID (i),]
+  [Object ID (i),]
   Publisher Priority (8),
   Extension Headers Length (i),
   [Extension headers (...)],
@@ -3233,6 +3234,26 @@ Each object sent on a fetch stream after the FETCH_HEADER has the following form
 }
 ~~~
 {: #object-fetch-format title="MOQT Fetch Object Fields"}
+
+The Serialization field defines the serialization of the object. The
+following code points are defined:
+
+|------|--------------------------------------------------------------------|
+| Code | Description                                                        |
+|-----:|:-------------------------------------------------------------------|
+| 0x0  | The Group ID, Subgroup ID and Object ID fields are present         |
+|------|--------------------------------------------------------------------|
+| 0x1  | The Group ID and Subgroup ID of this object are identical to those |
+|      | of the prior object and Object ID increments by 1. The Group ID,   |
+|      | Subgroup ID and Object ID fields are not present.                  |
+|------|--------------------------------------------------------------------|
+| 0x2  | The Group ID field exists and Subgroup ID and Object ID are both 0.|
+|      | The Subgroup ID and Object ID fields are not present.              |
+|------|--------------------------------------------------------------------|
+| 0x3  | The Group ID of this object is identical to that of the prior      |
+|      | Object. Subgroup ID increments by 1 and Object ID is 0. The        |
+|      | Subgroup ID and Object ID fields are not present.                  |
+|------|--------------------------------------------------------------------|
 
 The Object Status field is only sent if the Object Payload Length is zero.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3251,8 +3251,8 @@ following code points are defined:
 |      | The Subgroup ID and Object ID fields are not present.              |
 |------|--------------------------------------------------------------------|
 | 0x3  | The Group ID of this object is identical to that of the prior      |
-|      | Object. Subgroup ID increments by 1 and Object ID is 0. The        |
-|      | Subgroup ID and Object ID fields are not present.                  |
+|      | Object. Subgroup ID increments by 1 and Object ID increments by 1. |
+|      | The Subgroup ID and Object ID fields are not present.              |
 |------|--------------------------------------------------------------------|
 
 The Object Status field is only sent if the Object Payload Length is zero.


afrind (commented):
Thanks for the proposal, I like this direction.

suhasHere (commented):
I feel we don't need a type field to define the possible combinations but may be use the field present in the header or take it from the context.

wilaw:
> I feel we don't need a type field to define the possible combinations but may be use the field present in the header or take it from the context.

The FETCH response is a single stream. If you placed the 'type' in the only header, how could you differentiate Group boundaries (where Group ID increments and Object ID resets to 0) from non-Group boundaries (where GroupID is the same and Object ID increments?)

draft-ietf-moq-transport.md
```diff
@@ -3234,6 +3235,26 @@ Each object sent on a fetch stream after the FETCH_HEADER has the following form
 ~~~
 {: #object-fetch-format title="MOQT Fetch Object Fields"}
 
+The Serialization field defines the serialization of the object. The
+following code points are defined:
+
+|------|--------------------------------------------------------------------|
+| Code | Description                                                        |
```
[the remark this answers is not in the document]
wilaw: I like the idea of merging the Extension bit into the Serialization byte. 

> One potentially painful serialization is N subgroups and the object IDs rotate amongst them 

That would be its own punishment. None-the-less, it is still better than the status quo, which is to always repeat all 3 IDS. We can invent ever more codes for ever more inefficient schemes, but we should probably design the optimization for the majority use cases and let the more exotic ones be slightly more verbose.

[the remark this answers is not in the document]
wilaw: @afrind    Commit https://github.com/moq-wg/moq-transport/pull/949/commits/90f67b11663ff8ce14b78f26c0f2db8ce09075a6 reformats as a table and adds the new code points to allow for removing priority and extension length bytes.

draft-ietf-moq-transport.md
```diff
@@ -3276,6 +3277,37 @@ Each object sent on a fetch stream after the FETCH_HEADER has the following form
...
 {: #object-fetch-format title="MOQT Fetch Object Fields"}
 
+The Serialization field defines the serialization of the object. The
+following code points are defined:
+
+|------|----------|-------------|--------------|-----------|-------------------|
+| Code | Group ID	| Subgroup ID	| Object ID	   | Publisher | Extension Headers |
+|      |         	|             |              | Priority	 | Length            |
+|------|----------|-------------|--------------|-----------|-------------------|
+| 0x0	 | Present  | Present	    | Present	     | Present   | Present           |
```
ianswett: Nit: Present and Previous look similar at a glance.  I'd prefer Yes like Alan did in #1055.

You could use Prev and Prev+1 to make the table a bit narrower as well, since it looks like it might be over 80 characters?

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
afrind: Oops missed your comment from April

> let the more exotic ones be slightly more verbose.

I think increasing object IDs within a group amongst 2 or more subgroups may actually be fairly common -- eg it matches the layered encoding schemes folks have been discussing.  The other option for handling Object IDs across multiple subgroups is something like `sg=0` has object IDs `0...Q-1`, `sg=1` had object IDs `Q...2*Q - 1`, and so on.

draft-ietf-moq-transport.md
afrind: Something appears wonky with the table when I do "View File" the column headers are on the wrong columns.

wilaw: Commit https://github.com/moq-wg/moq-transport/pull/949/commits/39602c3e8233585b2706f3e7ce42413c1a3fddc5 makes table text clearer and more precise and also attempts to fix the column formatting problem.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
wilaw: @afrind We have used up our 8 options for a 1 byte serialization field. How would you propose adding in the use case of layered coding where objects IDs increase across multiple subgroups? By removing 0c6 and 0x7 and replacing them with a hard-coded case of odd/even across two subgroups called 0 and 1 ?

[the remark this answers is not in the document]
afrind: There are 256 options for an 8 byte value. Let's keep the encodings we have and add a couple more?

draft-ietf-moq-transport.md
```diff
@@ -3589,6 +3590,35 @@ Each object sent on a fetch stream after the FETCH_HEADER has the following form
...
+| 0x3	 | Yes      | 0        | 0         | Yes       | 0              |
+|------|----------|----------|-----------|-----------|----------------|
+| 0x4	 | Prior    | Prior    | Prior + 1 | Prior     | Yes            |
+|------|----------|----------|-----------|-----------|----------------|
+| 0x5	 | Prior    | Prior    | Prior + 1 | Prior     | 0              |
+|------|----------|----------|-----------|-----------|----------------|
+| 0x6	 | Prior    | Prior +1 | Prior + 1 | Yes       | Yes            |
+|------|----------|----------|-----------|-----------|----------------|
+| 0x7	 | Prior    | Prior +1 | Prior + 1 | Yes       | 0              |
+|------|----------|----------|-----------|-----------|----------------|
```
```suggestion
|------|----------|----------|-----------|-----------|----------------|
| 0x8	 | Prior    | 0 | Prior + 1 | Yes       | Yes              |
|------|----------|----------|-----------|-----------|----------------|
| 0x9	 | Prior    | 0 | Prior + 1 | Yes       | 0              |
|------|----------|----------|-----------|-----------|----------------|
```

afrind:
I left a suggestion for 2 additional serializations that will further compress groups with multiple subgroups.

We maybe could squeeze out publisher priority in a couple cases.  We only need it for the first object in a subgroup.

vasilvv:
This sounds a bit too complex, and it's not entirely clear to me as to how much we gain on average (given that we add one byte that's guaranteed to be present).  I feel like this needs more discussion / analysis.

cc @martinduke

wilaw:
> it's not entirely clear to me as to how much we gain on average (given that we add one byte that's guaranteed to be present).

An estimate of the savings is given in the intro to the PR, which I repeat below. The most common encoding would result in a  net 2 byte savings per object. 

> 0x0 - status quo - Group/Subgroup and Object ID fields are present. This would be used at the start of a track and adds 1 byte to the current serialization.
> 
> 0x1 - The Group ID and Subgroup ID of this object are identical to those of the prior object and Object ID increments by 1. The Group ID, Subgroup ID and Object ID fields are not present. This code point would most commonly occur in a FETCH response and would save 2 bytes per object.
> 
> 0x2 - The Group ID field exists and Subgroup ID and Object ID are both 0. The SubGroup ID and Object ID fields are not present. This code point would be used at the start of each new Group and saves 1 byte per object.
> 
> 0x3 - The Group ID of this object is identical to that of the prior Object. Subgroup ID increments by 1 and Object ID increments by 1. The Subgroup ID and Object ID fields are not present. This code point would be used at the start of a new Subgroup and would save 1 byte per object.

martinduke:
I like the idea of compression. From an implementation perspective, it would be great if we can map this to bits, rather than having 10 codewords whose capabilities have to be matched to a clumsy table, so that every check for "what element is next" requires multiple boolean checks against the integer. This is definitely an aesthetics/complexity complaint rather than a real performance one.

I would propose:
0x01 and 0x02: subgroup -- 4 permutations: 0, Prior, Prior+1, present
0x04: object -- present or prior+1. -- the "0" encoding saves at most one byte per group, which IMO is not worth the complexity.
0x08: Group --  prior or present.
0x10: publisher priority -- present or not. I know that the table is trying to enforce rules about when the publisher priority can actually differ, but that is futile. It's impossible to enforce this with encoding -- I still need to explicitly check, because the peer can make all the fields explicit and still send a subgroup with two different priorities. This is a false economy.
0x20: Extension header

I don't care about the ordering, except that subgroup should be the two LSBs.

wilaw:
@martin - I like the bitmasking approach. Before I implement as a patch on this PR, do we have consensus to make this change? It would look something like that the tables below:

The two least significant bits (LSBs) of the serialization byte form a two-bit field that defines the state of the subgroup. To extract this value, a bitwise AND operation with the mask 0x03 MUST be performed.

| Bitmask Result (Byte & 0x03) | Meaning |
| ----------- | ----------- |
| 0x00 | Subgroup ID is zero |
| 0x01 | Subgroup ID is the same as the Subgroup ID of the prior Object |
| 0x02 | Subgroup ID is one more than the Subgroup ID of the prior Object |
| 0x03|  The Subgroup ID field is present |

The following table defines additional individual flags within the serialization byte. Each flag is an independent boolean value, where a set bit (1) indicates the corresponding condition is true.

| Bitmask | Condition if set | Condition if not set |
| ----------- | ----------- | -------|
|0x04 | Object ID field is present | Object ID is one more than the Object ID of the prior Object |
|0x08 | Group ID field is present | Group ID is identical to the Group ID of the prior Object |
|0x10 | Priority field is present | Priority field is not present |
|0x20 | Extension header field is present | Extension header field is not present |

yekuiwang:
Related to #1153: The Subgroup ID field of an object with Forwarding Preference of "Datagram" is set to the Object ID. Use of the Forwarding Preference of "Datagram" does not seem to be a corner use case. Would it be good to consider adding another short-cut signalling of Subgroup ID, in which case its value is the same as the Object ID? 

I will try to figure out a way for this if there is interest of doing it.

yekuiwang:
> The following table defines additional individual flags within the serialization byte. Each flag is an independent boolean value, where a set bit (1) indicates the corresponding condition is true.
> 
> Bitmask	Condition if set	Condition if not set
> 0x04	Object ID field is present	Object ID is one more than the Object ID of the prior Object
> 0x08	Group ID field is present	Group ID is identical to the Group ID of the prior Object
> 0x10	Priority field is present	Priority field is not present
> 0x20	Extension header field is present	Extension header field is not present

@wilaw @afrind In the above "Priority field is not present" should be updated to "Priority field is not present and the value is  identical to the Priority of the prior Object with the same Subgroup ID"?

afrind:
> 0x10: publisher priority -- present or not. I know that the table is trying to enforce rules about when the publisher priority can actually differ, but that is futile. It's impossible to enforce this with encoding -- I still need to explicitly check, because the peer can make all the fields explicit and still send a subgroup with two different priorities. This is a false economy.

What is the value of the publisher priority if the value is not present?  Some previous value from the same subgroup, and it's malformed if it's the first object?

fluffy:
I like the direction of the bitmask approach will has above but our syntax for defining the messages on the wire explicitly support bitfields notation and I think we could use that.

afrind:
> What is the value of the publisher priority if the value is not present? Some previous value from the same subgroup, and it's malformed if it's the first object?

I realized for datagram tracks with one object per group, this still leaves us with 1 byte per datagram for priority.  So maybe "not present" means: previous in subgroup, if first in subgroup, previous in FETCH response.

> 0x04: object -- present or prior+1. -- the "0" encoding saves at most one byte per group, which IMO is not worth the complexity.

Along the same line, if datagrams are one object per group, that's a lot of bytes to encode 0.  Should we use one of the two remaining bits?

martinduke:
> > 0x10: publisher priority -- present or not. I know that the table is trying to enforce rules about when the publisher priority can actually differ, but that is futile. It's impossible to enforce this with encoding -- I still need to explicitly check, because the peer can make all the fields explicit and still send a subgroup with two different priorities. This is a false economy.
> 
> What is the value of the publisher priority if the value is not present? Some previous value from the same subgroup, and it's malformed if it's the first object?

As in the actual PR, if not present it's as the previous object. If the first object, I'm fine with either calling it parse error or using some default.

martinduke:
> > What is the value of the publisher priority if the value is not present? Some previous value from the same subgroup, and it's malformed if it's the first object?
> 
> I realized for datagram tracks with one object per group, this still leaves us with 1 byte per datagram for priority. So maybe "not present" means: previous in subgroup, if first in subgroup, previous in FETCH response.
> 
> > 0x04: object -- present or prior+1. -- the "0" encoding saves at most one byte per group, which IMO is not worth the complexity.
> 
> Along the same line, if datagrams are one object per group, that's a lot of bytes to encode 0. Should we use one of the two remaining bits?

The PR just says "prior object", which ought to be clarified. I took it to mean the previous one in the Fetch stream, but I don't feel strongly.

afrind:
Discussed in 9/10 virtual interim

1. We'll go with "priority not present = priority from previous object in subgroup if present, otherwise previous object in FETCH response".  

Upon reflection, it does put the burden on a fetch decoder of a map from subgroup ID -> priority which can only be cleared when the group changes.  A malicious publisher can force the subscriber into keeping effectively infinite state.  When I write this up, I'll put a bound on the maximum number of subgroups per group that can use this compression flag (256?).

2. There wasn't sufficient interest in adding a bit for Object ID=0.  Suhas said they do FETCH datagram Tracks with one object per group, but the extra byte is not significant in a FETCH response.  This leaves 2 bits easily accessible for future versions.

draft-ietf-moq-transport.md
```diff
@@ -3584,6 +3585,39 @@ Each object sent on a fetch stream after the FETCH_HEADER has the following form
 ~~~
 {: #object-fetch-format title="MOQT Fetch Object Fields"}
 
+The Serialization field defines the serialization of the object. The
+following code points are defined:
+
+| Code | Group ID | Subgroup<br>ID | Object<br>ID | Publisher<br>Priority | Extension<br>Headers<br>Length |
```
ianswett: Nit: How about just calling it Extensions in the table, esp since it'll get renamed soon anyway.

```suggestion
| Code | Group ID | Subgroup<br>ID | Object<br>ID | Publisher<br>Priority | Extensions |
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,69 @@ FETCH_HEADER {
 {: #fetch-header-format title="MOQT FETCH_HEADER"}
 
 
-Each object sent on a fetch stream after the FETCH_HEADER has the following format:
+Each Object sent on a FETCH stream after the FETCH_HEADER has the following format:
```
```suggestion
Each Object sent on a FETCH stream after the FETCH_HEADER has the following
format:
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,69 @@ FETCH_HEADER {
...
+  [Publisher Priority (8),]
+  [Extensions (..)],
   Object Payload Length (i),
   [Object Status (i),]
   [Object Payload (..),]
 }
 ~~~
 {: #object-fetch-format title="MOQT Fetch Object Fields"}
 
+The Serialization Flags field defines the serialization of the Object.
```
```suggestion
The Serialization Flags field defines the serialization of the Object.
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,69 @@ FETCH_HEADER {
...
+The Serialization Flags field defines the serialization of the Object. 
+
+The two least significant bits (LSBs) of the Serialization Flags form a two-bit
+field that defines the encoding of the Subgroup.  To extract this value, the
+Subscriber performs a bitwise AND operation with the mask 0x03.
+
+Bitmask Result (Byte & 0x03)	| Meaning
+0x00	| Subgroup ID is zero
+0x01	| Subgroup ID is the same as the Subgroup ID of the prior Object
+0x02	| Subgroup ID is one more than the Subgroup ID of the prior Object
```
```suggestion
0x01	| Subgroup ID is the prior Object's Subgroup ID
0x02	| Subgroup ID is the prior Object's Subgroup ID plus one
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,69 @@ FETCH_HEADER {
...
+that would cause the size of the Subscriber's map to exceed 64 entries.
+
 The Extensions structure is defined in {{object-extensions}}.
 
-The Object Status field is only sent if the Object Payload Length is zero.
+The Object Status field is only present if the Object Payload Length is zero.
 
-The Subgroup ID field of an object with a Forwarding Preference of "Datagram"
-(see {{object-properties}}) is set to the Object ID.
+When encoding an Objects with a Forwarding Preference of "Datagram" (see
```
```suggestion
When encoding an Object with a Forwarding Preference of "Datagram" (see
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,69 @@ FETCH_HEADER {
...
+the corresponding condition is true.
+
+Bitmask |	Condition if set	| Condition if not set
+0x04	| Object ID field is present	| Object ID is the prior Object's ID plus one
+0x08	| Group ID field is present |	Group ID is the prior Object's Group ID
+0x10	| Priority field is present	| See below
+0x20	| Extensions field is present |	Extensions field is not present
+0x40 | Undefined | Undefined
+0x80 | Undefined | Undefined
+
```
afrind, line 3705:
```suggestion

If the first Object in the FETCH response uses a flag that references the prior
Object, the Subscriber MUST close the session with a `PROTOCOL_VIOLATION`.
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,72 @@ FETCH_HEADER {
...
+If the first Object in the FETCH response uses a flag that references the prior
+Object, the Subscriber MUST close the session with a `PROTOCOL_VIOLATION`.
+When the Priority field is not present, the Publisher Priority of the Object
```
```suggestion

When the Priority field is not present, the Publisher Priority of the Object
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,73 @@ FETCH_HEADER {
...
+is determined as follows:
+
+1. If the Object is the first in the FETCH response, the Publisher Prioirty is
+128.
+2. If the Object is the first Object in a Group and Subgroup included in the
+FETCH response, the Publisher Priority is the same as the previous Object in the
+response.
+3. Otherwise the Object has the same Publisher Priority as the first Object in
+the Group and Subgroup included in the FETCH response.
```
ianswett: I'm confused about how you're using "Group and Subgroup" in these two bullets.  A Object might be the first Object in a Subgroup, but not the first Object in a Group.

afrind: I meant {group ID, subgroup ID}, but since we say in Object Model that subgroups are scoped to group, maybe that's implicit?

```suggestion
2. Otherwise, if the Object is the first Object in its Subgroup included in the
FETCH response, the Publisher Priority is the same as the previous Object in the
response.
3. Otherwise the Object has the same Publisher Priority as the first Object in
its Subgroup included in the FETCH response.
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,73 @@ FETCH_HEADER {
...
-  Subgroup ID (i),
-  Object ID (i),
-  Publisher Priority (8),
-  Extensions (..),
+  Serialization Flags (8),
+  [Group ID (i),]
+  [Subgroup ID (i),]
+  [Object ID (i),]
+  [Publisher Priority (8),]
+  [Extensions (..)],
```
yekuiwang: For consistency, ',' before ']'
```suggestion
  [Extensions (..),]
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,73 @@ FETCH_HEADER {
...
+The two least significant bits (LSBs) of the Serialization Flags form a two-bit
+field that defines the encoding of the Subgroup.  To extract this value, the
+Subscriber performs a bitwise AND operation with the mask 0x03.
+
+Bitmask Result (Byte & 0x03)	| Meaning
+0x00	| Subgroup ID is zero
+0x01	| Subgroup ID is the prior Object's Subgroup ID
+0x02	| Subgroup ID is the prior Object's Subgroup ID plus one
+0x03	| The Subgroup ID field is present
+
```
yekuiwang, line 3692: See below, we have "When encoding an Object with a Forwarding Preference of "Datagram" (see {{object-properties}}), the Publisher treats it as having a Subgroup ID equal to the Object ID." Why don't we use three LSBs for Subgroup ID and add a value to indicate Subgroup ID not present and is the Group ID?

draft-ietf-moq-transport.md
yekuiwang: Suggest to change "If" in step 2 to "Otherwise, if"

draft-ietf-moq-transport.md
afrind: I thought about this too.  So long as the objects in a datagram track use sequential object IDs, then 0x02 will give the same compression.  If the object's don't, then the 0x4 bit will be set and the object ID itself is present, which might be enough of a deterrent against choosing object IDs like that?

Using a third bit for "subgroup=object id" creates some invalid code points, and in practice wouldn't get much use.  It also removes a bit we could use in future versions to compress something else.  That said, I'm persuadable if others are in favor.

ianswett (approved):

draft-ietf-moq-transport.md
yekuiwang: Fair! Let's see whether anyone else has an opinion on this or not.

BTW, what were the use cases that drove the inclusion of the Datagram object transport mode in MOQT? I learned recently that low-bitrate audio is one. Were there any other use cases that would most likely use Datagrams?

afrind: I think audio is the main driver.  I think there are some open issues that may be a better place to discuss.

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,73 @@ FETCH_HEADER {
...
 ~~~
 {: #object-fetch-format title="MOQT Fetch Object Fields"}
 
+The Serialization Flags field defines the serialization of the Object.
+
+The two least significant bits (LSBs) of the Serialization Flags form a two-bit
+field that defines the encoding of the Subgroup.  To extract this value, the
+Subscriber performs a bitwise AND operation with the mask 0x03.
+
+Bitmask Result (Byte & 0x03)	| Meaning
```
```suggestion
Bitmask Result (Serialization Flags & 0x03)	| Meaning
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,73 @@ FETCH_HEADER {
...
+0x00	| Subgroup ID is zero
+0x01	| Subgroup ID is the prior Object's Subgroup ID
+0x02	| Subgroup ID is the prior Object's Subgroup ID plus one
+0x03	| The Subgroup ID field is present
+
+The following table defines additional flags within the Serialization Flags
+field. Each flag is an independent boolean value, where a set bit (1) indicates
+the corresponding condition is true.
+
+Bitmask |	Condition if set	| Condition if not set
```
```suggestion
Bitmask | Condition if set | Condition if not set
--------|------------------|---------------------
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,73 @@ FETCH_HEADER {
...
+0x04	| Object ID field is present	| Object ID is the prior Object's ID plus one
+0x08	| Group ID field is present |	Group ID is the prior Object's Group ID
+0x10	| Priority field is present	| See below
+0x20	| Extensions field is present |	Extensions field is not present
+0x40 | Undefined | Undefined
+0x80 | Undefined | Undefined
+
+If the first Object in the FETCH response uses a flag that references the prior
+Object, the Subscriber MUST close the session with a `PROTOCOL_VIOLATION`.
```
```suggestion
If the first Object in the FETCH response uses a flag that references the prior
Object (0x04) or prior Group (0x08), the Subscriber MUST close the session 
with a `PROTOCOL_VIOLATION`.
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,73 @@ FETCH_HEADER {
...
+128.
+2. Otherwise, if the Object is the first Object in its Subgroup included in the
+FETCH response, the Publisher Priority is the same as the previous Object in the
+response.
+3. Otherwise the Object has the same Publisher Priority as the first Object in
+its Subgroup included in the FETCH response.
+
+To decode a FETCH response, a Subscriber maintains a map of priorities for each
+Subgroup in the current Group.  The Publisher MUST NOT encode a FETCH response
+that would cause the size of the Subscriber's map to exceed 64 entries.
```
suhasHere: Does this imply the publisher can change the priority of a cached object when sending fetch response  rather than reflecting what is the canonical object stored in the cache  OR may be I am not clear of the condition this statement is making ?

suhasHere (commented):
@afrind  this is looking good. I do have few comments and clarification questions . please let me know. Thanks

ianswett (approved):

draft-ietf-moq-transport.md
afrind: Both 0x4 and 0x8 reference properties of the prior Object, as do 0x1 and 0x2, so I think the original text is fine.

draft-ietf-moq-transport.md
afrind: No, the publisher can't _change_ the priority.  But if there are more than 64 subgroups in a group in a fetch response, the priority must be encoded explicitly using "Priority Present".  I'll add this text.

```suggestion
Subgroup in the current Group.  The Publisher MUST NOT encode a FETCH response
that would cause the size of the Subscriber's map to exceed 64 entries. The
Publisher can only set the Priority field flag (0x10) to 0 for the first 64
Subgroups in a Group in the response.
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,76 @@ FETCH_HEADER {
...
+FETCH response, the Publisher Priority is the same as the previous Object in the
+response.
+3. Otherwise the Object has the same Publisher Priority as the first Object in
+its Subgroup included in the FETCH response.
+
+To decode a FETCH response, a Subscriber maintains a map of priorities for each
+Subgroup in the current Group.  The Publisher MUST NOT encode a FETCH response
+that would cause the size of the Subscriber's map to exceed 64 entries. The
+Publisher can only set the Priority field flag (0x10) to 0 for the first 64
+Subgroups in a Group in the response.
```
```suggestion
that would cause the size of the Subscriber's map to exceed 64 entries.  In any
given Group in the response, the Publisher can only set the Priority field flag
(0x10) to 0 for the first 64 Subgroups.  All subsequent Subgroups in the Group
MUST set the Priority field flag to 1.
```

afrind: Ugh even this is incorrect -- the limitation is only on objects in the subgroup after the first, which is hard to say.

draft-ietf-moq-transport.md
yekuiwang: Is it clearly said anywhere that the publisher can't change the priority of a cached object?

draft-ietf-moq-transport.md
yekuiwang: What exactly is in a map of priorities for a Subgroup? How to describe this 64-entries constraint depends on that.

draft-ietf-moq-transport.md
afrind: In 8.1 Caching Relays: 

> An endpoint that receives a duplicate Object with an invalid Object Status change, or with a different Forwarding Preference, Subgroup ID, Priority or Payload MUST treat the track as Malformed.

draft-ietf-moq-transport.md
afrind: You can think of it as a dictionary from subgroup ID to priority, with maximum size 64.  The subscriber needs to push an entry into this map when it sees the first object from the subgroup and determines it's priority, so that the subscriber can infer the correct Priority of subsequent objects with Priority not present.

And in writing that, I realized that the proposed text _is_ correct. Since the subscriber has no way of knowing if there will be subsequent objects in the subgroup, it has to push an entry into the map.  Since one might send many DATAGRAMs in a Group, and in FETCH response they have unique subgroups, maybe we need to allow for a larger map?

draft-ietf-moq-transport.md
yekuiwang: > In 8.1 Caching Relays:
> 
> > An endpoint that receives a duplicate Object with an invalid Object Status change, or with a different Forwarding Preference, Subgroup ID, Priority or Payload MUST treat the track as Malformed.

But that does not disallow a Relay receiving Objects of a Track from upstream, then send Objects of the Track downstream and change the Priority of some Objects while sending downstream, right?

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,76 @@ FETCH_HEADER {
...
+1. If the Object is the first in the FETCH response, the Publisher Prioirty is
+128.
+2. Otherwise, if the Object is the first Object in its Subgroup included in the
+FETCH response, the Publisher Priority is the same as the previous Object in the
+response.
+3. Otherwise the Object has the same Publisher Priority as the first Object in
+its Subgroup included in the FETCH response.
+
+To decode a FETCH response, a Subscriber maintains a map of priorities for each
+Subgroup in the current Group.  The Publisher MUST NOT encode a FETCH response
```
```suggestion
To decode a FETCH response, a Subscriber maintains a map of priority-of-first-object-in-subgroup for Subgroups in the current Group.  The Publisher MUST NOT encode a FETCH response

File: draft-ietf-moq-transport.md (hunk 18 above)
Comment (yekuiwang) [reply draft-ietf-moq-transport.md]: I see, then the wording should be slightly updated (I made a suggestion), as the existing wording seems saying a list of priority values for one Subgroup.

> Since one might send many DATAGRAMs in a Group, and in FETCH response they have unique subgroups, maybe we need to allow for a larger map?

First of all, I don't fully understand why Objects received from Datagram need to have Subgroup ID equal to Group ID, as when such Objects are fetched they have to be in the same stream anyway regardless of the number of Subgroups.

Secondly, if Subgroup IDs are equal to Group IDs, as we discussed earlier, it is pretty easy to compress using the indication that the priority is not present but equal to that of the previous Object plus 1, therefore, only one entry is needed (and even when there are a couple gaps then only a couple of more entries are needed). Therefore, I think 64 is good enough.

Comment (yekuiwang) [reply draft-ietf-moq-transport.md]: > Secondly, if Subgroup IDs are equal to Group IDs, as we discussed earlier, it is pretty easy to compress using the indication that priority is not present but equal to that of the previous Object plus 1, therefore, only one entry is needed (and even when there are a couple gaps then only a couple of more entries are needed). Therefore, I think 64 is good enough.

Aha, then some rewording is needed, as the map does not have to be continuous regarding Subgroup IDs, meaning e.g., the table could be { (SG0, 0), (SG101, 1)} for the case when there is only case of Group ID cap just prior to the Object with Object ID 101 (note the context herein is Subgroup ID equal to Object ID.

Comment (yekuiwang) [inline draft-ietf-moq-transport.md:3724]:
```suggestion
that would cause the size of the Subscriber's map to exceed 64 entries.

Comment (yekuiwang) [reply draft-ietf-moq-transport.md]: I think just removing the sentence "The Publisher can only set the Priority field flag (0x10) to 0 for the first 64 Subgroups in a Group in the response." would be fine, then.

Comment (yekuiwang) [reply draft-ietf-moq-transport.md:3724]: I think just then removing the sentence "The Publisher can only set the Priority field flag (0x10) to 0 for the first 64 Subgroups in a Group in the response." would work.

Comment (yekuiwang) [reply draft-ietf-moq-transport.md]: Ahh, sorry, I misunderstood the semantics (mixed Subgroup ID signalling and priority signalling): there is no such short cut signalling of priority: the priority is not present but equal to that of the previous Object plus 1. Therefore forget about my above comments regarding that 64 is sufficiently large for the map (but not the one below).

Comment (yekuiwang) [reply draft-ietf-moq-transport.md]: On the other hand, for Objects received from Datagrams, since Subgroup ID is Group ID, any Object is the first Object in its Subgroup, while this map is only useful for the case when "the Object has the same Publisher Priority as the first Object in its Subgroup included in the FETCH response", thus the map is not needed herein anyway. So still, 64 is sufficiently large for the map :-)

Comment (yekuiwang) [reply draft-ietf-moq-transport.md]: > All subsequent Subgroups in the Group MUST set the Priority field flag to 1.

This sentence is not correct, as the following could/would still be used with the Priority field flag equal to 0 without the need of the map:

> 2. Otherwise, if the Object is the first Object in its Subgroup included in the
> FETCH response, the Publisher Priority is the same as the previous Object in the
> response.

Comment (afrind) [reply draft-ietf-moq-transport.md]: I like the compression properties of your suggestion, but it makes the decoder rules even a little more complicated.

```
if object.group_id != previous_group_id:
  subgroup_priorities.clear()
previous_group_id = object.group_id

should_insert = True
if (serialization_byte & 0x10) == 0:
  if previous_object_priority:
    priority = subgroup_priorities[object.subgroup_id]
    if priority is None:
      priority = previous_object_priority
    else:
      should_insert = len(subgroup_priorities) < 64
  else:
    priority = 128
else:
  priority = decode_priority(input)

previous_object_priority = priority
if should_insert:
  subgroup_priorities[object.subgroup_id] = priority
```

Comment (yekuiwang) [reply draft-ietf-moq-transport.md]: @afrind Per your code above, if an Object of the n-th Subgroup with n< 64 has priority = 0, then the next Object would have priority = 128.  So I guess "if previous_object_priority:" should be "if previous_object_priority == -1:" if the initial value in the code was set to -1. Of course the initial value in the coded for previous_group_id also needs to be set properly.

A question: for the first Object of a Group, in your understanding, does it always have to have the priority field present? Your code seems implying that, but I cannot find that from the text. Probably we need to add a constraint that, at least for tracks received from Subgroup streams, for the first Object of a Group, the value of the Serialization Flags shall indicate the presence of all the three fields of Group ID, Object ID, and Publisher Priority. However, for tracks received from Datagrams, do we have to do the same?

If we don't specify the same constraint for tracks received from Datagrams (as currently is the case in the text of this PR), then it is possible for first Object of a Group to have the priority field not present, and if so then the current semantics below applies (as for fetching of tracks received from Datagrams, Subgroup ID is set to Object ID, thus each Object is in its own Subgroup and each Object is always "the first Object in its Subgroup"): 

> 2. Otherwise, if the Object is the first Object in its Subgroup included in the
> FETCH response, the Publisher Priority is the same as the previous Object in the
> response.

In other words, I was not suggesting a compression, but was just pointing out what is allowed by the current text of this PR.

File: draft-ietf-moq-transport.md [hunk 20]
```diff
@@ -3660,28 +3660,76 @@ FETCH_HEADER {
...
+The Serialization Flags field defines the serialization of the Object.
+
+The two least significant bits (LSBs) of the Serialization Flags form a two-bit
+field that defines the encoding of the Subgroup.  To extract this value, the
+Subscriber performs a bitwise AND operation with the mask 0x03.
+
+Bitmask Result (Serialization Flags & 0x03)	| Meaning
+0x00	| Subgroup ID is zero
+0x01	| Subgroup ID is the prior Object's Subgroup ID
+0x02	| Subgroup ID is the prior Object's Subgroup ID plus one
```
vasilvv, line 3690: Why plus one?  How common is this pattern?

draft-ietf-moq-transport.md
afrind: Discussed with Ian and Suhas and decided to not add Subgroup ID=Object ID bit for now.

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,76 @@ FETCH_HEADER {
...
+the corresponding condition is true.
+
+Bitmask | Condition if set | Condition if not set
+--------|------------------|---------------------
+0x04	| Object ID field is present	| Object ID is the prior Object's ID plus one
+0x08	| Group ID field is present |	Group ID is the prior Object's Group ID
+0x10	| Priority field is present	| See below
+0x20	| Extensions field is present |	Extensions field is not present
+0x40 | Undefined | Undefined
+0x80 | Undefined | Undefined
```
vasilvv: Please say something about what the applications are supposed to set those to / do with the values.

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,76 @@ FETCH_HEADER {
...
+
+If the first Object in the FETCH response uses a flag that references the prior
+Object, the Subscriber MUST close the session with a `PROTOCOL_VIOLATION`.
```
```suggestion
If the first Object in the FETCH response uses a flag that references fields in
the prior Object, the Subscriber MUST close the session with a
`PROTOCOL_VIOLATION`.
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,76 @@ FETCH_HEADER {
...
+
+When the Priority field is not present, the Publisher Priority of the Object
+is determined as follows:
+
+1. If the Object is the first in the FETCH response, the Publisher Prioirty is
```
```suggestion
1. If the Object is the first in the FETCH response, the Publisher Priority is
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,76 @@ FETCH_HEADER {
...
+128.
+2. Otherwise, if the Object is the first Object in its Subgroup included in the
+FETCH response, the Publisher Priority is the same as the previous Object in the
+response.
+3. Otherwise the Object has the same Publisher Priority as the first Object in
+its Subgroup included in the FETCH response.
+
+To decode a FETCH response, a Subscriber maintains a map of priorities for each
+Subgroup in the current Group.  The Publisher MUST NOT encode a FETCH response
+that would cause the size of the Subscriber's map to exceed 64 entries. The
```
vasilvv: If my reading of this paragraph is correct, this basically limits any individual group to 64 subgroups.

afrind: Not quite. We are debating how to handle this - either use "previous object" or the publisher has to explicitly encode the priority.

draft-ietf-moq-transport.md
afrind: if you have two subgroups 0 and 1 with even/odd object IDs, then FETCH will alternate between them.

draft-ietf-moq-transport.md
```suggestion
0x40 | `PROTOCOL_VIOLATION` | N/A
0x80 | `PROTOCOL_VIOLATION` | N/A
```

draft-ietf-moq-transport.md
```diff
@@ -3660,28 +3660,76 @@ FETCH_HEADER {
...
+0x00	| Subgroup ID is zero
+0x01	| Subgroup ID is the prior Object's Subgroup ID
+0x02	| Subgroup ID is the prior Object's Subgroup ID plus one
+0x03	| The Subgroup ID field is present
+
+The following table defines additional flags within the Serialization Flags
+field. Each flag is an independent boolean value, where a set bit (1) indicates
+the corresponding condition is true.
+
+Bitmask | Condition if set | Condition if not set
```
```suggestion
Bitmask | Condition if set | Condition if not set (0)
```

PR #956, opened 2025-04-27
Address section 9 edits and add common definition for reason phrase

### LINKED ISSUE #791: Format for Reason Phrase
A number of message errors allows for a reason phrase. What I see the reason phrase is underspecified as it appears to only be specified like this:

Reason Phrase: Provides the reason for subscription error.

So is this an UTF-8 string, and is it null terminated or not. It doesn't need to as we have a reason phrase length field. 

I would also like to point out that one probably should be explicit if the reason phrase length is in bytes, which is what I expect, and not characters which requires parsing to jump over.

### LINKED ISSUE #868: Joining Fetch Subscribe ID must refer to subscription in the current session
https://github.com/moq-wg/moq-transport/pull/822#discussion_r2014072806

### LINKED ISSUE #887: Minor Section 9 Issues
9.

"Every Track has a single 'Object Forwarding Preference' and the Original Publisher MUST NOT mix different forwarding preferences within a single track. If a subscriber receives different forwarding preferences for a track, it SHOULD close the session with an error of 'Protocol Violation'."

This is weirdly phrased because subscribers don't really "receive different forwarding preferences." How about changing the second sentence to

"If a subscriber receives objects via both stream and datagram in response to a SUBSCRIBE, or via Datagram in response to a FETCH, it SHOULD close the session with an error of 'Protocol Violation."

9.1.1

"An Object MUST be sent according to its Object Forwarding Preference, described below."

Prepend, "When in response to a SUBSCRIBE, ..."

9.1.1.1

"Though some status information could be inferred from QUIC stream state, that information is not reliable and cacheable."

I don't think this statement is broadly true. Gaps in FETCH streams imply the object DNE." 9.4.3 is an extended discussion of what stream state allows the receiver to infer about object status.

suhasHere:
Fixes #887 
Fixes #868 
Fixes #791

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2651,7 +2651,7 @@ ANNOUNCE_CANCEL Message {
   Length (i),
   Track Namespace (tuple),
   Error Code (i),
-  Erorr Reason (Reason Phrase),
+  Error Reason (Reason Phrase),
 }
 ~~~
 {: #moq-transport-announce-cancel-format title="MOQT ANNOUNCE_CANCEL Message"}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2631,8 +2631,7 @@ ANNOUNCE_CANCEL Message {
   Length (i),
   Track Namespace (tuple),
   Error Code (i),
-  Reason Phrase Length (i),
-  Reason Phrase Length (..),
+  Erorr Reason (Reason Phrase),
 }
 ~~~
 {: #moq-transport-announce-cancel-format title="MOQT ANNOUNCE_CANCEL Message"}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2353,31 +2353,31 @@ FETCH_ERROR Message {
 The application SHOULD use a relevant error code in FETCH_ERROR,
 as defined below:
 
-|------|---------------------------|
-| Code | Reason                    |
-|-----:|:--------------------------|
-| 0x0  | Internal Error            |
-|------|---------------------------|
-| 0x1  | Unauthorized              |
-|------|---------------------------|
-| 0x2  | Timeout                   |
-|------|---------------------------|
-| 0x3  | Not Supported             |
-|------|---------------------------|
-| 0x4  | Track Does Not Exist      |
-|------|---------------------------|
-| 0x5  | Invalid Range             |
-|------|---------------------------|
-| 0x6  | No Objects                |
-|------|---------------------------|
-| 0x7  | Invalid Joining Request ID|
-|------|---------------------------|
-| 0x10 | Malformed Auth Token      |
-|------|---------------------------|
-| 0x11 | Unknown Auth Token Alias  |
-|------|---------------------------|
-| 0x12 | Expired Auth Token        |
-|------|---------------------------|
+|------|------------------------------|
+| Code | Reason                       |
+|-----:|:-----------------------------|
+| 0x0  | Internal Error               |
+|------|------------------------------|
+| 0x1  | Unauthorized                 |
+|------|------------------------------|
+| 0x2  | Timeout                      |
+|------|------------------------------|
+| 0x3  | Not Supported                |
+|------|------------------------------|
+| 0x4  | Track Does Not Exist         |
+|------|------------------------------|
+| 0x5  | Invalid Range                |
+|------|------------------------------|
+| 0x6  | No Objects                   |
+|------|------------------------------|
+| 0x7  | Invalid Joining Subscribe ID |
+|------|------------------------------|
+| 0x10 | Malformed Auth Token         |
+|------|------------------------------|
+| 0x11 | Unknown Auth Token Alias     |
+|------|------------------------------|
+| 0x12 | Expired Auth Token           |
+|------|------------------------------|
 
 * Internal Error - An implementation specific or generic error occurred.
 
@@ -2397,7 +2397,7 @@ as defined below:
 * No Objects - The beginning of the requested range is after the latest group
   and object for the track, or the track has not published any objects.
 
-* Invalid Joining Request ID - The joining Fetch referenced a Request ID that
+* Invalid Joining Subscribe ID - The joining Fetch referenced a Request ID that
   did not belong to an active Subscription.
 
 * Malformed Auth Token - Invalid Auth Token serialization during registration
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -342,14 +342,14 @@ Reason Phrase {
 ~~~
 
 * Reason Phrase Length: A variable-length integer specifying the length of the
-reason phrase in bytes. The  reason phrase length has a maximum length of
-1024 bytes. If an endpoint recieves a length exceeding the maximum, it MUST
-close the session with a Protocol Violation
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -329,6 +329,28 @@ Length/Value does not match the serialization defined by that Type,
...
+
+* Reason Phrase Length: A variable-length integer specifying the length of the
+reason phrase in bytes. The  reason phrase length has a maximum length of
+1024 bytes. If an endpoint recieves a length exceeding the maximum, it MUST
+close the session with a Protocol Violation
+
+* Reason Phrase Value: Additional diagnostic information about the error condition.
+The reason phrase value is encoded as UTF-8 string and does not carry information,
+such as language tags, that would aid comprehension by  any entity other than
+the one that created the text.
```
afrind, line 352:
```suggestion
* Reason Phrase Length: A variable-length integer specifying the length of the
  reason phrase in bytes. The reason phrase length has a maximum length of
  1024 bytes. If an endpoint receives a length exceeding the maximum, it MUST
  close the session with a Protocol Violation

* Reason Phrase Value: Additional diagnostic information about the error condition.
  The reason phrase value is encoded as UTF-8 string and does not carry information,
  such as language tags, that would aid comprehension by any entity other than
  the one that created the text.
```

draft-ietf-moq-transport.md
```diff
@@ -2196,10 +2214,10 @@ requested.
...
 Fields present only for Relative Fetch (0x2) and Absolute Fetch (0x3):
 
-* Joining Request ID: The Request ID of the existing subscription to be
-joined. If a publisher receives a Joining Fetch with a Joining Request ID that
-does not correspond to an existing Subscribe, it MUST respond with a Fetch
-Error Invalid Joining Request ID.
+* Joining Subscribe ID: The Subscribe ID of the existing subscription to be
+  joined. If a publisher receives a Joining Fetch with a Subscribe ID that does
+  not correspond to an existing Subscribe in the same session, it MUST respond
+  with a Fetch Error with code Invalid Subscribe ID.
```
afrind, line 2220: Subscribe ID is no longer a term in the document, but we can call this Joining Subscribe ID -- if you are changing the error code name please also update the section in FETCH_ERROR.

```suggestion
* Joining Subscribe ID: The Request ID of the existing subscription to be
  joined. If a publisher receives a Joining Fetch with a Request ID that does
  not correspond to an existing Subscribe in the same session, it MUST respond
  with a Fetch Error with code Invalid Joining Subscribe ID.
```

draft-ietf-moq-transport.md
```diff
@@ -2839,10 +2845,10 @@ an Object it never requested, it SHOULD terminate the session with a protocol
...
 so the session MUST NOT be teriminated in that case.
 
-Every Track has a single 'Object Forwarding Preference' and the Original
-Publisher MUST NOT mix different forwarding preferences within a single track.
-If a subscriber receives different forwarding preferences for a track, it
-SHOULD close the session with an error of 'Protocol Violation'.
+Every Track has a single 'Object Forwarding Preference'. If a subscriber
+receives objects via both stream and datagram in response to a SUBSCRIBE, or
+via Datagram in response to a FETCH, it SHOULD close the session with an error
+of 'Protocol Violation."
```
afrind, line 2851: I forgot we had this text.  Changing the normative language here makes this PR no longer editorial, let's make the change in #938 instead

draft-ietf-moq-transport.md
```diff
@@ -2865,7 +2871,7 @@ A canonical MoQ Object has the following information:
 the Object {{priorities}}.
 
 * Object Forwarding Preference: An enumeration indicating how a publisher sends
-an object. The preferences are Subgroup and Datagram.  An Object
+an object. The preferences are Subgroup and Datagram.  When publishing, an Object
```
afrind, line 2874: When is an object sent other than when it is published? Fetch response?

suhasHere: yeah . this was just to clarify the context. I am fine either ways.

draft-ietf-moq-transport.md
```diff
@@ -329,6 +329,28 @@ Length/Value does not match the serialization defined by that Type,
...
+~~~
+Reason Phrase {
+  Reason Phrase Length (i),
+  Reason Phrase Value (..)
+}
+~~~
+
+* Reason Phrase Length: A variable-length integer specifying the length of the
+  reason phrase in bytes. The reason phrase length has a maximum length of
+  1024 bytes. If an endpoint receives a length exceeding the maximum, it MUST
```
mengelbart, line 346: If the length is max 1024 bytes, can we use a 2 byte length field instead of a variable length integer? (I see we had this restriction before, and changing the field here would make this non-editorial. Maybe we do it in a different PR?)

mengelbart (approved):

draft-ietf-moq-transport.md
ianswett: I think we discussed this before and decided it'd be best to do it in a separate PR.

draft-ietf-moq-transport.md
afrind, line 2851:
```suggestion
Every Track has a single 'Object Forwarding Preference' and the Original
Publisher MUST NOT mix different forwarding preferences within a single track.
If a subscriber receives Objects via both Subgroup streams and Datagrams in
response to a SUBSCRIBE, it SHOULD close the session with an error of 'Protocol Violation."
```

draft-ietf-moq-transport.md
```diff
@@ -2865,7 +2871,7 @@ A canonical MoQ Object has the following information:
 the Object {{priorities}}.
 
 * Object Forwarding Preference: An enumeration indicating how a publisher sends
-an object. The preferences are Subgroup and Datagram.  An Object
+an object. The preferences are Subgroup and Datagram.  When publishing, an Object
 MUST be sent according to its `Object Forwarding Preference`, described below.
```
afrind, line 2875:
```suggestion
an object. The preferences are Subgroup and Datagram.  When in response to a
SUBSCRIBE, an Object MUST be sent according to its `Object Forwarding
Preference`, described below.
```

draft-ietf-moq-transport.md
```diff
@@ -2841,8 +2847,8 @@ so the session MUST NOT be teriminated in that case.
 
 Every Track has a single 'Object Forwarding Preference' and the Original
 Publisher MUST NOT mix different forwarding preferences within a single track.
-If a subscriber receives different forwarding preferences for a track, it
-SHOULD close the session with an error of 'Protocol Violation'.
+If a subscriber receives Objects via both Subgroup streams and Datagrams in
+response to a SUBSCRIBE, it SHOULD close the session with an error of 'Protocol Violation."
```
afrind, line 2851:
```suggestion
response to a SUBSCRIBE, it SHOULD close the session with an error of 'Protocol
Violation'
```

afrind (approved):

draft-ietf-moq-transport.md
```diff
@@ -2639,8 +2651,7 @@ ANNOUNCE_CANCEL Message {
   Length (i),
   Track Namespace (tuple),
   Error Code (i),
-  Reason Phrase Length (i),
-  Reason Phrase (..),
+  Erorr Reason (Reason Phrase),
```
ianswett, line 2654:
```suggestion
  Error Reason (Reason Phrase),
```

PR #965, opened 2025-04-28
Remove TRACK_STATUS

### LINKED ISSUE #895: TRACK_STATUS 0x04
The draft defines the following TRACK_STATUS:

> 0x04: The publisher is a relay that cannot obtain the current track status from upstream. Subsequent fields contain the largest group and object ID known.

What are the circumstances under which this can happen?  Is this so that relay can immediately answer without going upstream (see #882)?  Is that to distinguish original publisher denying existence vs relay having no associated original publisher?

ianswett:
Equivalent functionality can be achieved with SUBSCRIBE or FETCH and we keep forgetting to update TRACK_STATUS.

Possibly it's worth editorial text on how to do the equivalent functionality with SUBSCRIBE or FETCH?

Fixes #895

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1256,10 +1256,6 @@ The following Message Types are defined:
 |-------|-----------------------------------------------------|
 | 0x17  | FETCH_CANCEL ({{message-fetch-cancel}})             |
 |-------|-----------------------------------------------------|
-| 0xD   | TRACK_STATUS_REQUEST ({{message-track-status-req}}) |
-|-------|-----------------------------------------------------|
-| 0xE   | TRACK_STATUS ({{message-track-status}})             |
-|-------|-----------------------------------------------------|
 | 0x6   | ANNOUNCE  ({{message-announce}})                    |
 |-------|-----------------------------------------------------|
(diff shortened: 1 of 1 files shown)


afrind:
The one difference between SUBSCRIBE, forward=0 and TRACK_STATUS_REQUEST is that the subscriber still needs to UNSUBSCRIBE, which is annoying.  I do like the idea of making the relationship between TRACK_STATUS_REQUEST and SUBSCRIBE identical to HTTP HEAD and GET - the response is the same but without the bytes.

wilaw:
SUBSCRIBE, forward=0, wait for SUBSCRIBE_OK, UNSUBSCRIBE is a messy and hacky mechanism to retrieve information about the live edge of a track. It forces media objects to move (or begin moving) over the wire, perhaps over many hops, just in order to return metadata about the track , which could otherwise be exchanged via messages. We should retain TRACK_STATUS as defined today.

ianswett:
I agree that relying on Forward is not ideal, I was thinking of sending a SUBSCRIBE or FETCH from (0, 0) to (0, 0) or some other empty range?

afrind:
Discussed in Stockholm - we're going to keep it and make it better but not remove it

PR #966, opened 2025-04-28
Unify parameter parsing by always using byte length

ianswett:
Instead of the number of parameters, as are used in some control messages.

I thought the parsers were already the same, but it turns out some spots use number of parameters and Objects use total length.  I don't think that difference was intentional, but let me know if it was.

This was intended as an editorial change, but it isn't as currently written.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -337,6 +337,7 @@ length of the pairs in bytes.
 Key-Value-Pairs {
   Length (i),
   Pair (Key-Value)...,
+}
 ~~~
 {: #moq-key-value-pairs format title="MOQT Key-Value-Pairs"}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1506,14 +1506,14 @@ CLIENT_SETUP Message {
   Length (i),
   Number of Supported Versions (i),
   Supported Versions (i) ...,
-  Setup Parameters (..),
+  Setup Parameters (KeyValuePairs),
 }
 
 SERVER_SETUP Message {
   Type (i) = 0x21,
   Length (i),
   Selected Version (i),
-  Setup Parameters (..),
+  Setup Parameters (KeyValuePairs),
 }
 ~~~
 {: #moq-transport-setup-format title="MOQT Setup Messages"}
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1742,8 +1752,7 @@ SUBSCRIBE Message {
   Filter Type (i),
   [Start Location (Location)],
   [End Group (i)],
-  Number of Parameters (i),
-  Subscribe Parameters (..) ...
+  Subscribe Parameters (KeyValuePairs)
```
ianswett, line 1755:
```suggestion
  Parameters (KeyValuePairs)
```

draft-ietf-moq-transport.md
```diff
@@ -329,6 +330,17 @@ Length/Value does not match the serialization defined by that Type,
 the receiver MUST terminate the session with error code 'Key-Value
 Formatting Error'.
 
+
+Key-Value-Pairs are a sequence of Key-Values preceded by a total
+length of the pairs in bytes.
+~~~
+Key-Value-Pairs {
+  Length (i),
+  Pair (Key-Value)...,
```
ianswett, line 339:
```suggestion
  Pair (Key-Value)...,
}
```

martinduke:
Individual comment:

Everything used to be NumParams/NumExtensions. We changed it for extensions to a byte count because relays, at the time, might just want to pull the whole blob and pass it up without parsing.

I'm not sure if that purpose has survived putting in relay-useful data like object ID gaps.

In the case of params, we have message lengths and params are always at the end, so I don't know that we need any field at all.

I don't have a firm position on this, but we should think about / discuss it more.

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -329,6 +330,18 @@ Length/Value does not match the serialization defined by that Type,
...
```
fluffy, line 339: Uh, Is "(Key-Value)..." valid syntax for this ? I wish it was but I don't think it is.

fluffy (changes requested):
I don't want to loose the ability to do extension on data plane that don't have an explicit  length field because the length is known from the type. I think this is critical to keep the data plan bandwidth optimized.

afrind:
Discussed in Stockholm - not changing this now

PR #972, opened 2025-04-29
Use two byte int in all control message definitions

mengelbart:


sharmafb (approved):

fluffy:
I don't have any objection to this but why would we do this ?

afrind (approved):

PR #975, opened 2025-04-30
rename Type to Format for SUBGROUP_HEADER

suhasHere:
This is purely editorial PR that renames Type to Config for identifying different configurations of SubGroup Streams. 

If we feel this is a bad idea, I am totally ok for this PR to be closed.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3063,7 +3063,7 @@ and the subgroup indicated by 'Group ID' and `Subgroup ID`.
 
 ~~~
 SUBGROUP_HEADER {
-  Type (i),
+  Config (i),
   Track Alias (i),
   Group ID (i),
   [Subgroup ID (i),]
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -3094,12 +3094,12 @@ There are 6 defined Type values for SUBGROUP_HEADER:
 | 0x0D | Yes           | N/A             | Yes        |
 |------|---------------|-----------------|------------|
 
-For Type values where Subgroup ID Field Present is No, there is no explicit
-Subgroup ID field in the header and the Subgroup ID is either 0 (for Types
-0x08-09) or the Object ID of the first object transmitted in this subgroup
-(for Types 0x0A-0B).
+For Confif values where Subgroup ID Field Present is No, there is no explicit
```
mengelbart, line 3097:
```suggestion
For Config values where Subgroup ID Field Present is No, there is no explicit
```

suhasHere:
> I don't love "Config" as a name for this field but I can see that Type s perhaps a little messy. Maybe Format?

Makes sense. updated PR.

ianswett (commented):
I'm fine with Type, but I can why you might want to try to rename it.  "Object Encoding", "Elided Object Fields", "Object Format" would work for me.  "Stream Format" or "Subgroup Format" are also ok, but I think I prefer something referring to Objects, since this value is changing the encoding of Objects on the wire for this stream.

suhasHere:
> I'm fine with Type, but I can why you might want to try to rename it.  "Object Encoding", "Elided Object Fields", "Object Format" would work for me.  "Stream Format" or "Subgroup Format" are also ok, but I think I prefer something referring to Objects, since this value is changing the encoding of Objects on the wire for this stream.

I am also ok with SubGroupFormat or ObjectFormat

afrind:
Closing, sinceFormat is not substantially clearer than Type and Type is used elsewhere to differentiate the function of streams.  However, I think we can add editorial text explaining that the Type field in a unidirectional stream indicates both the Type of stream (Fetch/Subgroup) as well as formatting / compression options.

PR #977, opened 2025-05-01
Move Track Alias from SUBSCRIBE to SUBSCRIBE_OK

afrind:
* Explain it a little better
* Add normative relay handling
* Mention buffering

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2867,13 +2867,6 @@ To optimize wire efficiency, Subgroups and Datagrams refer to a track by a
 numeric identifier, rather than the Full Track Name.  Track Alias is chosen by
 the publisher and included in SUBSCRIBE_OK ({{message-subscribe-ok}}.
 
-Though a Track Alias is only required to be unique within a single session,
-Publishers MAY set Track Alias to a more unique (possibly globally unique)
-identifier for the Track.  Relays SHOULD assign the Track Alias received in an
-upstream SUBSCRIBE_OK to downstream subscriptions for the same track. Since
-subscribers can request Tracks from uncoordinated publishers through a single
-relay session, it is not always possible to reuse the upstream Track Alias.
-
 ## Objects {#message-object}
 
 An Object contains a range of contiguous bytes from the
@@ -3068,9 +3061,10 @@ ID` in the SUBGROUP_HEADER.
 
 If an endpoint receives a subgroup with an unknown Track Alias, it MAY abandon
 the stream, or choose to buffer it for a brief period to handle reordering with
-the control message that establishes the Track Alias.  The endpoint SHOULD NOT
-release stream flow control beyond the SUBGROUP_HEADER until the Track Alias has
-been established.  TODO: talk about possible deadlocks.
+the control message that establishes the Track Alias.  The endpoint MAY withhold
+stream flow control beyond the SUBGROUP_HEADER until the Track Alias has been
+established.  To prevent deadlocks, the publisher MUST allocate connection flow
+control to the control stream before allocating it any data streams.
 
 ~~~
 SUBGROUP_HEADER {
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1140,7 +1140,7 @@ each active upstream subscription that matches that namespace, it SHOULD send a
 SUBSCRIBE to the publisher that sent the ANNOUNCE.
 
 Relays use the Track Alias ({{track-alias}}) of an incoming Object to identify
-its track and find the active subscribers for that track. Relays MUST forward
+its track and find the active subscribers. Relays MUST forward
 Objects to matching subscribers in accordance to each subscription's priority,
 group order, and delivery timeout.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2972,16 +2972,25 @@ Object Extension Headers are serialized as Key-Value-Pairs {{moq-key-value-pair}
 Header types are registered in the IANA table 'MOQ Extension Headers'.
 See {{iana}}.
 
-## Object Datagram {#object-datagram}
+## Datagrams
+
+A single object can be conveyed in a datagram.  The Track Alias field
+{{track-alias}} indicates the track this Datagram belongs to.  If an endpoint
+receives a datagram with an unknown Track Alias, it MAY drop the datagram or
+choose to buffer it for a brief period to handle reordering with the control
+message that establishes the Track Alias.
+
+An Object received in an `OBJECT_DATAGRAM` or `OBJECT_DATAGRAM_STATUS` message
+has an `Object Forwarding Preference` = `Datagram`. To send an Object with
+`Object Forwarding Preference` = `Datagram`, determine the length of the header
+and payload and send the Object as datagram. In certain scenarios where the
+object size can be larger than maximum datagram size for the session, the Object
+will be dropped.
+
+
+### Object Datagram {#object-datagram}
 
 An `OBJECT_DATAGRAM` carries a single object in a datagram.
-
-An Object received in an `OBJECT_DATAGRAM` message has an `Object
-Forwarding Preference` = `Datagram`. To send an Object with `Object
-Forwarding Preference` = `Datagram`, determine the length of the header and
-payload and send the Object as datagram. In certain scenarios where the object
-size can be larger than maximum datagram size for the session, the Object
-will be dropped.
 
 ~~~
 OBJECT_DATAGRAM {
@@ -3006,7 +3015,7 @@ Violation.
 There is no explicit length field.  The entirety of the transport datagram
 following Publisher Priority contains the Object Payload.
 
-## Object Datagram Status {#object-datagram-status}
+### Object Datagram Status {#object-datagram-status}
 
 An `OBJECT_DATAGRAM_STATUS` is similar to OBJECT_DATAGRAM except it
 conveys an Object Status and has no payload.
@@ -3030,11 +3039,6 @@ The Type field takes the form 0b0000001X (or the set of values from 0x02 to
 Extension headers are present. If an endpoint receives a datagram with Type
 0x03 and Extension Headers Length is 0, it MUST close the session with Protocol
 Violation.
-
-* Track Alias: The Track Alias ({{track-alias}} indicating the track this
-  Datagram belongs to.  If an endpoint receives a datagram with an unknown Track
-  Alias, it MAY drop the datagram or choose to buffer it for a brief period to
-  handle reordering with the control message that establishes the Track Alias.
 
 ## Streams
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1139,12 +1139,10 @@ When a relay receives an incoming ANNOUNCE for a given namespace, for
 each active upstream subscription that matches that namespace, it SHOULD send a
 SUBSCRIBE to the publisher that sent the ANNOUNCE.
 
-Object headers carry a short hop-by-hop `Track Alias` that maps to
-the Full Track Name (see {{message-subscribe-ok}}). Relays use the
-`Track Alias` of an incoming Object to identify its track and find
-the active subscribers for that track. Relays MUST forward Objects to
-matching subscribers in accordance to each subscription's priority, group order,
-and delivery timeout.
+Relays use the Track Alias ({{track-alias}}) of an incoming Object to identify
+its track and find the active subscribers for that track. Relays MUST forward
+Objects to matching subscribers in accordance to each subscription's priority,
+group order, and delivery timeout.
 
 If an upstream session is closed due to an unknown or invalid control message
 or Object, the relay MUST NOT continue to propagate that message or Object
@@ -1732,7 +1730,6 @@ SUBSCRIBE Message {
   Type (i) = 0x3,
   Length (i),
   Request ID (i),
-  Track Alias (i),
   Track Namespace (tuple),
   Track Name Length (i),
   Track Name (..),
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1139,12 +1139,10 @@ When a relay receives an incoming ANNOUNCE for a given namespace, for
...
 SUBSCRIBE to the publisher that sent the ANNOUNCE.
 
-Object headers carry a short hop-by-hop `Track Alias` that maps to
-the Full Track Name (see {{message-subscribe-ok}}). Relays use the
-`Track Alias` of an incoming Object to identify its track and find
-the active subscribers for that track. Relays MUST forward Objects to
-matching subscribers in accordance to each subscription's priority, group order,
-and delivery timeout.
+Relays use the Track Alias ({{track-alias}}) of an incoming Object to identify
+its track and find the active subscribers for that track. Relays MUST forward
```
ianswett, line 1143:
```suggestion
its track and find the active subscribers. Relays MUST forward
```

draft-ietf-moq-transport.md
```diff
@@ -2874,6 +2861,19 @@ If a subscriber receives Objects via both Subgroup streams and Datagrams in
 response to a SUBSCRIBE, it SHOULD close the session with an error of 'Protocol
 Violation'
 
+## Track Alias
```
suhasHere, line 2864:
```suggestion
## Track Alias {#track-alias}
```

afrind: Through some dark magic, an anchor with the name #foo-bar is automatically created for a section titled Foo Bar.  So this isn't necessary.

draft-ietf-moq-transport.md
```diff
@@ -1804,6 +1795,7 @@ SUBSCRIBE_OK Message {
   Type (i) = 0x4,
   Length (i),
   Request ID (i),
+  Track Alias (i),
```
ianswett, line 1798: I think we should rename this at some point, because it's really an alas for the subscription, but it's ok for now.

draft-ietf-moq-transport.md
```diff
@@ -1817,6 +1809,12 @@ SUBSCRIBE_OK Message {
 * Request ID: The Request ID of the SUBSCRIBE this message is replying to
   {{message-subscribe-req}}.
 
+* Track Alias: The identifer used for this track in Subgroups or Datagrams (see
+  {{track-alias}}). The same Track Alias MUST NOT be used to refer to two
+  different Tracks simultaneously. If a subscriber receives a SUBSCRIBE_OK that
+  uses the same Track Alias as a different track with an active subscription, it
+  MUST close the session with error 'Duplicate Track Alias'.
```
ianswett, line 1816:
```suggestion
  different Tracks (and therefore Subscriptions) simultaneously. If a subscriber receives a SUBSCRIBE_OK that
  uses the same Track Alias as a different track with an active subscription, it
  MUST close the session with error 'Duplicate Track Alias'.
```

draft-ietf-moq-transport.md
```diff
@@ -3057,9 +3062,15 @@ effect on outstanding subscriptions.
...
-and the subgroup indicated by 'Group ID' and `Subgroup ID`.
+All Objects on a Subgroup stream belong to the track identified by `Track Alias`
+(see {{track-alias}}) and the Subgroup indicated by 'Group ID' and `Subgroup
+ID` in the SUBGROUP_HEADER.
+
+If an endpoint receives a subgroup with an unknown Track Alias, it MAY abandon
+the stream, or choose to buffer it for a brief period to handle reordering with
+the control message that establishes the Track Alias.  The endpoint SHOULD NOT
+release stream flow control beyond the SUBGROUP_HEADER until the Track Alias has
+been established.  TODO: talk about possible deadlocks.
```
ianswett, line 3073: Is this because the application layer can't really use this subgroup data yet?  If so, maybe say why?

ianswett (approved):
This is certainly simpler and can save some bytes in some cases.

draft-ietf-moq-transport.md
fluffy: > I think we should rename this at some point, because it's really an alas for the subscription, but it's ok for now.

I don't care if we rename it but it's not really an alias for the subscribe but a local scoped map to the full track name. Depending on other PR, publish would also use this ID. Something might make one subscribe to a track, end it, then make new subscribe to same track, Those two subscribes might on some implementaions get the same track alias for both of them because they were both to the same track.  

I fine with renaming it at some point to make it clear this is not the same thing as track alias in earlier version of the track, but I don't want to treat it like it is the same things as subscribe id.

draft-ietf-moq-transport.md
```diff
@@ -2874,6 +2861,19 @@ If a subscriber receives Objects via both Subgroup streams and Datagrams in
...
 
+## Track Alias
+
+To optimize wire efficiency, Subgroups and Datagrams refer to a track by a
+numeric identifier, rather than the Full Track Name.  Track Alias is chosen by
+the publisher and included in SUBSCRIBE_OK ({{message-subscribe-ok}}.
+
+Though a Track Alias is only required to be unique within a single session,
+Publishers MAY set Track Alias to a more unique (possibly globally unique)
+identifier for the Track.  Relays SHOULD assign the Track Alias received in an
```
fluffy, line 2872: Nope, this is total non starter for me, if we are going to have this SHOULD, we need to go back to also having the track error. This is simply not going to work. 

Imagine the case where for meetings the client all use 1 for their first track, then 2, etc. Nearly every alias will need to be changed yet putting this SHOULD here is going to make applications think they can assume t not be changed.

draft-ietf-moq-transport.md
```diff
@@ -3038,6 +3038,11 @@ Extension headers are present. If an endpoint receives a datagram with Type
 0x03 and Extension Headers Length is 0, it MUST close the session with Protocol
 Violation.
 
+* Track Alias: The Track Alias ({{track-alias}} indicating the track this
+  Datagram belongs to.  If an endpoint receives a datagram with an unknown Track
+  Alias, it MAY drop the datagram or choose to buffer it for a brief period to
+  handle reordering with the control message that establishes the Track Alias.
```
fluffy, line 3044: Nit ... this seems like the the wrong place to put the "drop of buffer" text but I agree with the text.

draft-ietf-moq-transport.md
```diff
@@ -3057,9 +3062,15 @@ effect on outstanding subscriptions.
...
-When a stream begins with `SUBGROUP_HEADER`, all Objects on the stream
-belong to the track requested in the Subscribe message identified by `Track Alias`
-and the subgroup indicated by 'Group ID' and `Subgroup ID`.
+All Objects on a Subgroup stream belong to the track identified by `Track Alias`
+(see {{track-alias}}) and the Subgroup indicated by 'Group ID' and `Subgroup
+ID` in the SUBGROUP_HEADER.
+
+If an endpoint receives a subgroup with an unknown Track Alias, it MAY abandon
+the stream, or choose to buffer it for a brief period to handle reordering with
+the control message that establishes the Track Alias.  The endpoint SHOULD NOT
```
fluffy, line 3071: I think you could make this a MAY but I strongly object to it being a SHOULD - this data arrived because the subscriber asked for this track. I see no reason not to deliver it. Doing this SHOULD will delay it getting there and make the user experience worse. Sure there is some DDOS buffer limit of what the client may stop but that has to do with DOS stuff that we can deal with across all the stuff.

fluffy (changes requested):
I'm still trying to decide if we need the error to use a different alias or not.

draft-ietf-moq-transport.md
afrind: I was trying to leave open the possibility that this identifier gets bound to the track, as Cullen says.  There's also cases if we allow multiple concurrent subscriptions to the same track with non-overlapping filters.

draft-ietf-moq-transport.md
afrind: This problem exists already with subscriber chosen alias and RETRY.  The two original publishers both selected Alias=1 on different sessions, but the downstream subscriber can only have a single track with Alias=1.

Would this work any better for you?

> Relays MUST assign the Track Alias received in SUBSCRIBE_OK to downstream subscriptions to the same Track, unless another Track is already using that Alias.

draft-ietf-moq-transport.md
afrind: I sort of agree but not sure where else to put it.

draft-ietf-moq-transport.md
afrind: > I think you could make this a MAY but I strongly object to it being a SHOULD

I'm confused.  It is a MAY:

> it MAY abandon the stream, or choose to buffer it for a brief period

What are you asking for?  MUST buffer data with unknown identifiers?  I'm not sure that's any better because how much and for how long? No one has infinite memory, and it could be an attack so I don't see how we can make buffering everything forever a requirement.  If a subscriber's buffer gets filled up and it drops some stuff, it can always fetch it.  Applications that care will set buffer limits that don't impact performance.

> this data arrived because the subscriber asked for this track

Data with an unknown identifier could be due to a buggy or malicious peer.

draft-ietf-moq-transport.md
afrind: If the connection flow control is used up by streams with an unknown identifier it could block control messages that establish the indentifiers.

draft-ietf-moq-transport.md
vasilvv: There is also a bit of a tension here with the "default case": when there are no global IDs, the relay would want to always use the lowest unused track alias for the connection.

draft-ietf-moq-transport.md
ianswett: I think either case can be useful.  I think "Publish ID" would be good, given it's obvious and not used elsewhere.

draft-ietf-moq-transport.md
ianswett: Agreed that this is probably the wrong spot for this.  How about: https://github.com/moq-wg/moq-transport/blob/main/draft-ietf-moq-transport.md#delivery-timeout-parameter-delivery-timeout

Alternatively. we could add a section for Datagrams adjacent to Streams?

ianswett:
Discussed in Stockholm, there was wide interest in making this always publisher chosen, but people weren't sure about MUST vs SHOULD and @fluffy thought SHOULD was not useful and instead of SHOULD, it'd be better to say nothing.

@ianswett suggested adding a param to SETUP to indicate if the Publisher ID/Track Alias should be preserved in that Session.

draft-ietf-moq-transport.md
fluffy: after discussion of this ... I am OK with just going to we don't say anything about what relays do with the numbers (and of course they can not conflict)

draft-ietf-moq-transport.md
afrind: Let's paint the possible renaming bikeshed later.

draft-ietf-moq-transport.md
afrind: I'm being kind of intentionally vague here until we completely nail "multiple concurrent subscriptions".  Per last week, it seems like we're leaning "no concurrent subscriptions to same track, allow update to widen filter, publisher acknowledgement to subscribe_update".

The text I have is valid either way.

draft-ietf-moq-transport.md
afrind: Chatted offline - I was the confused one, Cullen was referring to releasing flow control.  I'll make this a MAY.

draft-ietf-moq-transport.md
```diff
@@ -2874,6 +2861,19 @@ If a subscriber receives Objects via both Subgroup streams and Datagrams in
...
+numeric identifier, rather than the Full Track Name.  Track Alias is chosen by
+the publisher and included in SUBSCRIBE_OK ({{message-subscribe-ok}}.
+
+Though a Track Alias is only required to be unique within a single session,
+Publishers MAY set Track Alias to a more unique (possibly globally unique)
+identifier for the Track.  Relays SHOULD assign the Track Alias received in an
+upstream SUBSCRIBE_OK to downstream subscriptions for the same track. Since
+subscribers can request Tracks from uncoordinated publishers through a single
+relay session, it is not always possible to reuse the upstream Track Alias.
+
```
afrind, line 2876:
```suggestion
```

draft-ietf-moq-transport.md
afrind, line 3073:
```suggestion
the control message that establishes the Track Alias.  The endpoint MAY withhold
stream flow control beyond the SUBGROUP_HEADER until the Track Alias has been
established.  To prevent deadlocks, the publisher MUST allocate connection flow
control to the control stream before allocating it any data streams.
```

draft-ietf-moq-transport.md
afrind: I removed the normative relay text.

draft-ietf-moq-transport.md
afrind: I took your second suggestion and made a top level datagram section

afrind:
@vasilvv @suhasHere any additional feedback required for your approval?

draft-ietf-moq-transport.md
suhasHere: If a subscribers have multiple subscriptions to same track and if they overlap, the subscriber will indeed get objects sent multiple times for the overlapped region of the subscriptions. I think it is a fine behavior and believe that the core transport shouldn't block concurrent subscriptions to a given track.

fluffy (approved):

vasilvv (approved):

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -3057,9 +3059,16 @@ effect on outstanding subscriptions.
...
+All Objects on a Subgroup stream belong to the track identified by `Track Alias`
+(see {{track-alias}}) and the Subgroup indicated by 'Group ID' and `Subgroup
+ID` in the SUBGROUP_HEADER.
+
+If an endpoint receives a subgroup with an unknown Track Alias, it MAY abandon
+the stream, or choose to buffer it for a brief period to handle reordering with
+the control message that establishes the Track Alias.  The endpoint MAY withhold
+stream flow control beyond the SUBGROUP_HEADER until the Track Alias has been
+established.  To prevent deadlocks, the publisher MUST allocate connection flow
+control to the control stream before allocating it any data streams.
```
afrind, line 3071:
```suggestion
established.  To prevent deadlocks, the publisher MUST allocate connection flow
control to the control stream before allocating it any data streams. Otherwise,
a receiver might wait for a control message containing a Track Alias to release
flow control, while the sender waits for flow control to send the message.
```

PR #987, opened 2025-05-02
Fetch should use Joining Request ID instead of subscribe ID

TimEvens:


afrind:
This was intentional based on @suhasHere  feedback -- the ID must be the Request ID of a Subscribe

TimEvens:
> This was intentional based on @suhasHere  feedback -- the ID must be the Request ID of a Subscribe

Is fetch joining not referring to an existing request id?  I don't see any issue calling it what it is, request id. The verbiage and invalid error name should change to reflect that.

Referring to it in code and docs as subscribe id is wrong IMHO.

mengelbart (approved):

suhasHere:
> > This was intentional based on @suhasHere  feedback -- the ID must be the Request ID of a Subscribe
> 
> Is fetch joining not referring to an existing request id? I don't see any issue calling it what it is, request id. The statement, only needs to be changed to use request Id. I just updated that. I also updated the invalid joining error.
> 
> Referring to it in code and docs as request ID is consistent and intuitive. It supports the existing data types.

Yes they are indeed request Ids , however, for joining fetch it makes it very clear to call it as SubscribeID to reflect the semantic purpose

TimEvens:
> Yes they are indeed request Ids , however, for joining fetch it makes it very clear to call it as SubscribeID to reflect the semantic purpose

@suhasHere , by that logic it should be renamed to `subscribe_request_id`

Can you explain why this **MUST** be called out differently than the other messages that do the same, such as  `SUBSCRIBE_UPDDATE { ..., Request Id (i), ... }` is defined as follows:

```
Request ID: The Request ID of the SUBSCRIBE ([Section 8.7](https://datatracker.ietf.org/doc/html/draft-ietf-moq-transport-11#message-subscribe-req)) this message is updating. 
This MUST match an existing Request ID.
```

Subscribe update is not the only one that uses a request Id that refers to an existing subscribe request Id.

afrind (approved):

ianswett (approved):

PR #992, opened 2025-05-05
Reference concept of "time" in 2.3.1

alvestrand:
The concept of "time" elsewhere in the document refers to protocol event time. In this particular section, it refers to media time. This should be called out.


afrind (approved):

suhasHere (approved):

ianswett (commented):
I'm wondering if you're getting at a related point, which is that Groups and Objects might be produced in temporal order, but not arrive at various points in the MoQ network in the same order?

fluffy (approved):
LGTM

Issue #778, opened 2025-03-16
Video ingestion has extra latency and is awkward


ianswett:
I believe the intended flow for ingestion today is a client connects, then sends an ANNOUNCE for the namespace it wants to publish to, then the server sends a SUBSCRIBE for a track.

This takes an extra RTT at best over the client just beginning to publish once it connects.

How does the server know which track name(s) to subscribe to?

It seems like a simpler model would be similar to HTTP POST, where the client connects and starts publishing whenever it wants to, and the server can cancel the PUBLISH at any time if it doesn't want to receive the data.

As implied, this requires a new verb, ie: PUBLISH to allow the publisher to initiate publication of a Track.

afrind:
Individual Comment:

The awkwardness of "publisher speaks first" is not limited to video ingestion either.  We are investigating moqt to carry existing relay-to-relay realtime traffic and would likely use something like PUBLISH if it existed.

gmarzot:
+1 to ingest ease / PUBLISH. making it easy to drive video into the mesh from a lot of different types of sources is a good thing. WHIP/WHEP comes to mind.

afrind:
Discussed in Stockholm - the direction of the publish PR is likely to resolve this.

Issue #765, opened 2025-03-06
Should we revisit relay routing on exact namespace match?


afrind:
The current design is:

1. Publishers ANNOUNCE individual namespaces to the relay
2. SUBSCRIBE/FETCH has a namespace field
3. The relay compares these field and routes to sessions that match exactly

An alternate design is prefix match on either the namespace only, or the namespace and the name.  We previously decided not to prefix match because there was no safe way to do this before we tupleized namespace.

See related #764

suhasHere:
+1 on revisiting this. Relays need to implement prefix match for  Authz purposes.

afrind:
> Relays need to implement prefix match for Authz purposes.

Can you say more about how this is an authorization requirement?

suhasHere:
For authorization, the claim in the token will typically be a prefix of the actual namespace or ftn that will be provided in the corresponding control message ( for ex: slide 2 in https://datatracker.ietf.org/meeting/interim-2025-moq-01/materials/slides-interim-2025-moq-01-sessa-authorization-in-moq-00)

Relays will do prefix match to see if the announced namespace is covered by the claim, for example

afrind:
Individual Comment:

> Relays will do prefix match to see if the announced namespace is covered by the claim, for example

I think what the relay does is separable into two steps:

1) check if the user authorized is for the action+track
2) routing upstream SUBSCRIBE/FETCH if needed, based on matching logic 

We could decide to use prefix matching in auth but exact matching for routing.  I'm not trying to argue against prefix based routing, but I don't think it's a hard dependency.

PR #995, opened 2025-05-05
Add a PUBLISH message for publisher initiated subscriptions

afrind:
Also changes relay matching from exact namespace match to namespace tuple prefix match.

Note: this includes PR #977 - please review that first

Fixes: #615
Fixes: #765
Fixes: #778

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -803,8 +803,8 @@ State is 0, the publisher does not send objects for the subscription.  If the
 Forward State is 1, the publisher sends objects.  The initiator of the
 subscription sets the initial Forward State in either PUBLISH or SUBSCRIBE.  The
 sender of PUBLISH_OK can update the Forward State based on its preference.  Once
-established, the subscriber can change the Forward State by sending
-SUBSCRIBE_UPDATE.
+the subscription is established, the subscriber can update the Forward State by
+sending SUBSCRIBE_UPDATE.
 
 Either endpoint can initiate a subscription to a track without exchanging any
 prior messages other than SETUP.  Relays MUST NOT send any PUBLISH messages
@@ -883,8 +883,8 @@ publishers to contact, if any.
 
 An UNSUBSCRIBE_ANNOUNCES withdraws a previous SUBSCRIBE_ANNOUNCES. It does not
 prohibit original publishers from sending further ANNOUNCE or PUBLISH messages,
-but relays MUST not send a PUBLISH message to a client without knowing the
-client is interested in and authorized to receive the content.
+but relays MUST not send any further PUBLISH messages to a client without
+knowing the client is interested in and authorized to receive the content.
 
 ## Announcements
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -529,7 +529,7 @@ The subscriber either accepts or rejects the subscription using PUBLISH_OK or
 PUBLISH_ERROR.  A subscriber initiates a subscription to a track by sending the
 SUBSCRIBE message.  The publisher either accepts or rejects the subscription
 using SUBSCRIBE_OK or SUBSCRIBE_ERROR.  Once either of these sequences is
-complete, the subscription can be updated by the subscriber using
+successful, the subscription can be updated by the subscriber using
 SUBSCRIBE_UPDATE, terminated by the subscriber using UNSUBSCRIBE, or terminated
 by the publisher using SUBSCRIBE_DONE.
 
@@ -883,8 +883,8 @@ forward the result to the application, so the application can decide which other
 publishers to contact, if any.
 
 An UNSUBSCRIBE_ANNOUNCES withdraws a previous SUBSCRIBE_ANNOUNCES. It does not
-prohibit the publishers from sending further ANNOUNCE or PUBLISH messages, but
-it does prevent relays from sending them.
+prohibit original publishers from sending further ANNOUNCE or PUBLISH messages,
+but it does prevent relays from sending them.
 
 ## Announcements
 
@@ -1169,8 +1169,8 @@ subscription in Forward State=0, it SHOULD send a SUBSCRIBE_UDPATE with
 Forward=1 to all publishers.
 
 When a relay receives an incoming PUBLISH message, it SHOULD send a PUBLISH
-request to each subscriber that has subscribed to the track's namespace or
-prefix thereof.
+request to each subscriber that has subscribed (via SUBSCRIBE_ANNOUNCES)
+to the track's namespace or prefix thereof.
 
 When a relay receives an incoming ANNOUNCE for a given namespace, for
 each active upstream subscription that matches that namespace, it SHOULD send a
@@ -2227,7 +2227,8 @@ PUBLISH_OK Message
 
 * Subscriber Priority: The Subscriber Priority for this subscription.
 
-* Group Order: The Group Order preference for this subscription.
+* Group Order: The Group Order preference for this subscription.  This
+overwrites the GroupOrder specified PUBLISH.
 
 * Filter Type, Start, End Group: See {{message-subscribe-req}}.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3242,11 +3242,6 @@ Extension headers are present. If an endpoint receives a datagram with Type
 0x03 and Extension Headers Length is 0, it MUST close the session with Protocol
 Violation.
 
-* Track Alias: The Track Alias ({{track-alias}} indicating the track this
-  Datagram belongs to.  If an endpoint receives a datagram with an unknown Track
-  Alias, it MAY drop the datagram or choose to buffer it for a brief period to
-  handle reordering with the control message that establishes the Track Alias.
-
 ## Streams
 
 When objects are sent on streams, the stream begins with a Subgroup Header
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1145,7 +1145,7 @@ new tracks.
 
 2. Send an ANNOUNCE message for a Track Namespace to the relay. This enables the
 relay to send SUBSCRIBE messages for Tracks in this Namespace in response to
-Consumer SUBSCRIBE messages.
+received SUBSCRIBE messages.
 
 Relays MUST verify that publishers are authorized to publish the set of tracks
 whose Track Namespace matches the namespace in
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -891,7 +891,7 @@ publishers to contact, if any.
 
 An UNSUBSCRIBE_ANNOUNCES withdraws a previous SUBSCRIBE_ANNOUNCES. It does not
 prohibit original publishers from sending further ANNOUNCE or PUBLISH messages,
-but relays MUST not send any further PUBLISH messages to a client without
+but relays MUST NOT send any further PUBLISH messages to a client without
 knowing the client is interested in and authorized to receive the content.
 
 ## Announcements
@@ -1147,10 +1147,10 @@ new tracks.
 relay to send SUBSCRIBE messages for Tracks in this Namespace in response to
 Consumer SUBSCRIBE messages.
 
-Relays MUST verify that publishers are authorized to publish the content
-associated with the set of tracks whose Track Namespace matches the namespace in
-an ANNOUNCE, or the Full Track Name in PUBLISH. Where the authorization and
-identification of the publisher occurs depends on the way the relay is managed
+Relays MUST verify that publishers are authorized to publish the set of tracks
+whose Track Namespace matches the namespace in
+an ANNOUNCE, or the Full Track Name in PUBLISH. The authorization and
+identification of the publisher depends on the way the relay is managed
 and is application specific.
 
 A Relay can receive announcements for the same Track Namespace or PUBLISH
@@ -2248,7 +2248,7 @@ overwrites the GroupOrder specified PUBLISH.
 
 ## PUBLISH_ERROR {#message-publish-error}
 
-The subscriber sends an PUBLISH_ERROR control message for to reject
+The subscriber sends an PUBLISH_ERROR control message to reject
 a subscription initiated by PUBLISH.
 
 ~~~
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1129,11 +1129,15 @@ to the old relay can be stopped with an UNSUBSCRIBE.
 
 ## Publisher Interactions
 
-There are two ways to publish through a relay - sending a PUBLISH messages for a
-specific Track or sending an ANNOUNCE message for a Track Namespace and waiting
-for SUBSCRIBE.  Relays MAY respond to PUBLISH with PUBLISH_OK in Forward State=0
-until there are known subscribers for new tracks. ANNOUNCE enables the relay to
-know which publisher to forward a SUBSCRIBE to.
+There are two ways to publish through a relay:
+
+1. Send a PUBLISH message for a specific Track to the relay. The relay MAY
+respond with PUBLISH_OK in Forward State=0 until there are known subscribers for
+new tracks.
+
+2. Send an ANNOUNCE message for a Track Namespace to the relay. This enables the
+relay to send SUBSCRIBE messages for Tracks in this Namespace in response to
+Consumer SUBSCRIBE messages.
 
 Relays MUST verify that publishers are authorized to publish the content
 associated with the set of tracks whose Track Namespace matches the namespace in
@@ -2246,8 +2250,7 @@ PUBLISH_ERROR Message
   Length (i),
   Request ID (i),
   Error Code (i),
-  Reason Phrase Length (i),
-  Reason Phrase (..),
+  Error Reason (Reason Phrase),
 }
 ~~~
 {: #moq-transport-publish-error format title="MOQT PUBLISH_ERROR Message"}
@@ -2257,7 +2260,7 @@ PUBLISH_ERROR Message
 
 * Error Code: Identifies an integer error code for failure.
 
-* Reason Phrase: Provides the reason for error.
+* Error Reason: Provides the reason for subscription error. See {{reason-phrase}}.
 
 The application SHOULD use a relevant error code in PUBLISH_ERROR, as defined
 below:
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1160,8 +1160,9 @@ determine which publishers receive a SUBSCRIBE or which subscribers receive a
 PUBLISH. For example, a SUBSCRIBE namespace=(foo,bar), track=x message will be
 forwarded to the sessions that sent ANNOUNCE namespace=(foo) and ANNOUNCE
 namespace=(foo, bar) respectively, but not one that sent ANNOUNCE
-namespace=(foobar).  Relays MUST forward to all publishers or subscribers that
-have a namespace prefix match.  TODO: do we need to say more about this.
+namespace=(foobar).  Relays MUST forward SUBSCRIBE messages to all publishers
+and ANNOUNCE and PUBLISH messages to all subscribers that have a namespace
+prefix match.
 
 When a relay receives an incoming SUBSCRIBE that triggers an upstream
 subscription, it SHOULD send a SUBSCRIBE request to each publisher that has
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1163,8 +1163,8 @@ in a namespace it previously responded ANNOUNCE_OK to by sending an
 ANNOUNCE_CANCEL.
 
 A relay manages sessions from multiple publishers and subscribers, connecting
-them based on the track namespace or track name.  Prefix matching is used to
-determine which publishers receive a SUBSCRIBE or which subscribers receive a
+them based on the Track Namespace or Full Track Name.  Prefix matching is used
+to determine which publishers receive a SUBSCRIBE or which subscribers receive a
 PUBLISH. For example, a SUBSCRIBE namespace=(foo,bar), track=x message will be
 forwarded to the sessions that sent ANNOUNCE namespace=(foo) and ANNOUNCE
 namespace=(foo, bar) respectively, but not one that sent ANNOUNCE
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -541,10 +541,12 @@ sender of PUBLISH_OK can update the Forward State based on its preference.  Once
 established, the subscriber can change the Forward State by sending
 SUBSCRIBE_UPDATE.
 
-Either endpoint can initiate a subscription to a track without exhanging any
-prior messages other than SETUP.  Relays however do not send any PUBLISH
-messages without first receiving a SUBSCRIBE_ANNOUNCES to a relevant track
-namespace or namespace prefix.
+Either endpoint can initiate a subscription to a track without exchanging any
+prior messages other than SETUP.  Relays MUST NOT send any PUBLISH messages
+without knowing the client is interested in and authorized to receive the
+content. The communication of intent and authorization can be accomplished by
+the client sending SUBSCRIBE_ANNOUNCES, or conveyed in other mechanisms out of
+band.
 
 # Sessions {#session}
 
@@ -884,7 +886,8 @@ publishers to contact, if any.
 
 An UNSUBSCRIBE_ANNOUNCES withdraws a previous SUBSCRIBE_ANNOUNCES. It does not
 prohibit original publishers from sending further ANNOUNCE or PUBLISH messages,
-but it does prevent relays from sending them.
+but relays MUST not send a PUBLISH message to a client without knowing the
+client is interested in and authorized to receive the content.
 
 ## Announcements
 
@@ -1160,7 +1163,7 @@ namespace=(foo, bar) respectively, but not one that sent ANNOUNCE
 namespace=(foobar).  Relays MUST forward to all publishers or subscribers that
 have a namespace prefix match.  TODO: do we need to say more about this.
 
-When a relay receives an incoming SUBSCRIBE request that triggers an upstream
+When a relay receives an incoming SUBSCRIBE that triggers an upstream
 subscription, it SHOULD send a SUBSCRIBE request to each publisher that has
 announced the subscription's namespace or prefix thereof, unless it already has
 an active subscription for the Objects requested by the incoming SUBSCRIBE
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -784,20 +784,44 @@ in order to obtain an accurate minimum RTT. Similarly, Reno halves it's congesti
...
+Forward State is 1, the publisher sends objects.  The initiator of the
+subscription sets the initial Forward State in either PUBLISH or SUBSCRIBE.  The
+sender of PUBLISH_OK can update the Forward State based on its preference.  Once
+established, the subscriber can change the Forward State by sending
+SUBSCRIBE_UPDATE.
+
+Either endpoint can initiate a subscription to a track without exchanging any
+prior messages other than SETUP.  Relays MUST NOT send any PUBLISH messages
+without knowing the client is interested in and authorized to receive the
+content. The communication of intent and authorization can be accomplished by
```
wilaw, line 812:
```
Relays MUST NOT send any PUBLISH messages
without knowing the client is interested in and authorized to receive the
content. 
```
This is unenforceable and should not be normative. Bad relays will do this and we should provide mechanisms for the receivers to protect themselves. Additionally, Publish should carry a token to enforce authorization.

ianswett: Agreed, given this is unenforceable, I don't think we can go stronger than SHOULD

ianswett:
Notes from Stockholm discussion:
- Make sure know how PUBLISH works with multiple relays and the Forward bit
- 765: Remove the change to do prefix namespace matching instead of exact namespace matching and put it into another PR
- 764: We're ok with no longer merging Track Namespace and Name, partially because it is odd to be able to ANNOUNCE and PUBLISH on the same tuple.
- Merging the verbs: Being able to set the forward bit on SUBSCRIBE_ANNOUNCES is important to @ianswett but trying to get SUBSCRIBE to do what SUBSCRIBE_ANNOUNCES is doing is awkward, as @afrind found out when trying to write the PR multiple times.  Similarly, subscriber priority is likely useful in wildcard subscribe (SUBSCRIBE_ANNOUNCES)

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2145,149 @@ SUBSCRIBE_DONE, as defined below:
...
+* Request ID: The Request ID of the PUBLISH this message is replying to
+  {{message-publish}}.
+
+* Forward: The Forward State for this subscription, either 0 (don't
+  forward) or 1 (forward).
+
+* Subscriber Priority: The Subscriber Priority for this subscription.
+
+* Group Order: The Group Order preference for this subscription.  This
+overwrites the GroupOrder specified PUBLISH.
```
sharmafb, line 2231: If a subscriber receives a PUBLISH with GroupOrder == Ascending, but overrides it to Descending, what does the peer do it it is unwilling or unable to serve in that order? Does it just immediately send a SUBSCRIBE_DONE with an error code?

sharmafb:
Does this really solve the additional RTT issue mentioned in #778? My understanding is that the publisher would have to wait till it receives a PUBLISH_OK before it starts sending any data.

draft-ietf-moq-transport.md
afrind: Group Order in a subscription is only input to prioritization, not object production, so I don't think there's a case where the publisher is unwilling/unable.

afrind:
> Does this really solve the additional RTT issue mentioned in #778? My understanding is that the publisher would have to wait till it receives a PUBLISH_OK before it starts sending any data.

The publisher is allowed to begin sending data before OK, though we should probably say that explicitly.

draft-ietf-moq-transport.md
```diff
@@ -784,20 +784,44 @@ in order to obtain an accurate minimum RTT. Similarly, Reno halves it's congesti
...
+subscription sets the initial Forward State in either PUBLISH or SUBSCRIBE.  The
+sender of PUBLISH_OK can update the Forward State based on its preference.  Once
+established, the subscriber can change the Forward State by sending
+SUBSCRIBE_UPDATE.
+
+Either endpoint can initiate a subscription to a track without exchanging any
+prior messages other than SETUP.  Relays MUST NOT send any PUBLISH messages
+without knowing the client is interested in and authorized to receive the
+content. The communication of intent and authorization can be accomplished by
+the client sending SUBSCRIBE_ANNOUNCES, or conveyed in other mechanisms out of
```
suhasHere, line 813:
```suggestion
the client sending SUBSCRIBE_ANNOUNCES/SUBSCRIBE message, or conveyed in other mechanisms out of
```

draft-ietf-moq-transport.md
```diff
@@ -857,15 +881,17 @@ The receiver of a SUBSCRIBE_ANNOUNCES_OK or SUBSCRIBE_ANNOUNCES_ERROR ought to
 forward the result to the application, so the application can decide which other
 publishers to contact, if any.
 
-An UNSUBSCRIBE_ANNOUNCES withdraws a previous SUBSCRIBE_ANNOUNCES. It does
-not prohibit the receiver (publisher) from sending further ANNOUNCE messages.
+An UNSUBSCRIBE_ANNOUNCES withdraws a previous SUBSCRIBE_ANNOUNCES. It does not
+prohibit original publishers from sending further ANNOUNCE or PUBLISH messages,
+but relays MUST not send a PUBLISH message to a client without knowing the
```
suhasHere, line 886:
```suggestion
but relays MUST not send any subsequent PUBLISH message to a client without knowing the
```

draft-ietf-moq-transport.md
```diff
@@ -1102,49 +1129,59 @@ to the old relay can be stopped with an UNSUBSCRIBE.
...
 
-Publishing through the relay starts with publisher sending ANNOUNCE
-control message with a `Track Namespace` ({{model-track}}).
-The ANNOUNCE enables the relay to know which publisher to forward a
-SUBSCRIBE to.
+There are two ways to publish through a relay - sending a PUBLISH messages for a
+specific Track or sending an ANNOUNCE message for a Track Namespace and waiting
+for SUBSCRIBE.  Relays MAY respond to PUBLISH with PUBLISH_OK in Forward State=0
+until there are known subscribers for new tracks. ANNOUNCE enables the relay to
+know which publisher to forward a SUBSCRIBE to.
```
suhasHere, line 1136:
```suggestion
There are two ways to publish through a relay - sending a PUBLISH messages for a
specific Track or sending an ANNOUNCE message for a Track Namespace and waiting
for SUBSCRIBE.  Relays MAY respond to PUBLISH with PUBLISH_OK in Forward State=0
until there are known subscribers for new tracks. ANNOUNCE enables the relay to
know which publisher to forward a SUBSCRIBE to.```

There are two methods to publish through a relay:
1. Sending a PUBLISH message for a specific track. Relays MAY respond to PUBLISH with PUBLISH_OK with Forward State=0 until subscribers for new tracks are identified.
2. Sending an ANNOUNCE message for a track namespace and waiting for SUBSCRIBE(s) to the tracks in that namespace. This allows the relay to determine which publisher(s) to forward the SUBSCRIBE(s) to.

File: draft-ietf-moq-transport.md [hunk 6]
```diff
@@ -1102,49 +1129,59 @@ to the old relay can be stopped with an UNSUBSCRIBE.
...
-Relays MUST verify that publishers are authorized to publish
-the content associated with the set of
-tracks whose Track Namespace matches the announced namespace. Where the
-authorization and identification of the publisher occurs depends on the way the
-relay is managed and is application specific.
+Relays MUST verify that publishers are authorized to publish the content
+associated with the set of tracks whose Track Namespace matches the namespace in
+an ANNOUNCE, or the Full Track Name in PUBLISH. Where the authorization and
+identification of the publisher occurs depends on the way the relay is managed
+and is application specific.
```
suhasHere, line 1142:
```suggestion
Relays MUST verify that publishers are authorized to publish the content
associated with the set of tracks whose Track Namespace matches the namespace in
an ANNOUNCE, or the Full Track Name in PUBLISH. Where the authorization and
identification of the publisher occurs depends on the way the relay is managed
and is application specific.
```
There are two methods to publish through a relay:

1. Sending a PUBLISH message for a specific track. Relays MAY respond to PUBLISH with 
PUBLISH_OK and Forward State=0, until subscribers for new tracks are identified.

2. Sending an ANNOUNCE message for a track namespace and waiting for SUBSCRIBE(s) to the track within the announced track namespace. This allows the relay to determine which publisher(s) to forward the SUBSCRIBE(s) to.

draft-ietf-moq-transport.md
suhasHere, line 1136:
```suggestion
There are two methods to publish through a relay:

1. Sending a PUBLISH message for a specific track. Relays MAY respond to PUBLISH with
PUBLISH_OK and Forward State=0, until subscribers for new tracks are identified.

2.Sending an ANNOUNCE message for a track namespace and waiting for SUBSCRIBE(s) to the track within the announced track namespace. This allows the relay to determine which publisher(s) to forward the SUBSCRIBE(s) to.```

File: draft-ietf-moq-transport.md [hunk 7]
```diff
@@ -784,20 +784,44 @@ in order to obtain an accurate minimum RTT. Similarly, Reno halves it's congesti
...
+SUBSCRIBE_UPDATE, terminated by the subscriber using UNSUBSCRIBE, or terminated
+by the publisher using SUBSCRIBE_DONE.
+
+All subscriptions have a Forward State which is either 0 or 1.  If the Forward
+State is 0, the publisher does not send objects for the subscription.  If the
+Forward State is 1, the publisher sends objects.  The initiator of the
+subscription sets the initial Forward State in either PUBLISH or SUBSCRIBE.  The
+sender of PUBLISH_OK can update the Forward State based on its preference.  Once
+established, the subscriber can change the Forward State by sending
+SUBSCRIBE_UPDATE.
```
suhasHere, line 807:
```suggestion
sender of PUBLISH_OK can update the Forward State based on its preference.  Once
the subscription is established, the subscriber can update the Forward State by sending
SUBSCRIBE_UPDATE.
```

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2145,149 @@ SUBSCRIBE_DONE, as defined below:
...
+* Track Alias: The identifer used for this track in Subgroups or Datagrams (see
+  {{track-alias}}). The same Track Alias MUST NOT be used to refer to two
+  different Tracks simultaneously. If a subscriber receives a PUBLISH that
+  uses the same Track Alias as a different track with an active subscription, it
+  MUST close the session with error 'Duplicate Track Alias'.
+
+* ContentExists: 1 if an object has been published on this track, 0 if not.
+If 0, then the Largest Group ID and Largest Object ID fields will not be
+present. Any other value is a protocol error and MUST terminate the
+session with a Protocol Violation ({{session-termination}}).
```
suhasHere, line 2187: I am not sure if we need ContentExists for publish message, since the intent of publish message is to publish content.

suhasHere (approved):
Minor nits, Otherwise LGTM

draft-ietf-moq-transport.md
fluffy: Not so much about the issue here but I strongly object to this idea that things that are not enforceable can not be normative. We need to tell relays what they can and can not do. This is also very testable and observable by a client.

draft-ietf-moq-transport.md
afrind: I don't think we want to send PUBLISH in response to SUBSCRIBE.  We would send SUBSCRIBE_OK plus track data.

draft-ietf-moq-transport.md
```diff
@@ -857,15 +881,17 @@ The receiver of a SUBSCRIBE_ANNOUNCES_OK or SUBSCRIBE_ANNOUNCES_ERROR ought to
 forward the result to the application, so the application can decide which other
 publishers to contact, if any.
 
-An UNSUBSCRIBE_ANNOUNCES withdraws a previous SUBSCRIBE_ANNOUNCES. It does
-not prohibit the receiver (publisher) from sending further ANNOUNCE messages.
+An UNSUBSCRIBE_ANNOUNCES withdraws a previous SUBSCRIBE_ANNOUNCES. It does not
+prohibit original publishers from sending further ANNOUNCE or PUBLISH messages,
+but relays MUST not send a PUBLISH message to a client without knowing the
+client is interested in and authorized to receive the content.
```
afrind, line 887:
```suggestion
but relays MUST not send any further PUBLISH messages to a client without
knowing the client is interested in and authorized to receive the content.
```

draft-ietf-moq-transport.md
afrind: I think we do.  I imagine in most cases, Content Exists = 0 (hence no Largest Location yet).  However, if a publisher is starting in the middle of a track (reconnect?) then Content Exists = 1 and Largest Location will be present.

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2145,149 @@ SUBSCRIBE_DONE, as defined below:
...
+
+~~~
+PUBLISH_ERROR Message
+{
+  Type (i) = 0x8,
+  Length (i),
+  Request ID (i),
+  Error Code (i),
+  Reason Phrase Length (i),
+  Reason Phrase (..),
```
afrind, line 2250:
```suggestion
  Error Reason (Reason Phrase),
```

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2145,149 @@ SUBSCRIBE_DONE, as defined below:
...
+}
+~~~
+{: #moq-transport-publish-error format title="MOQT PUBLISH_ERROR Message"}
+
+* Request ID: The Request ID of the PUBLISH this message is replying to
+  {{message-publish}}.
+
+* Error Code: Identifies an integer error code for failure.
+
+* Reason Phrase: Provides the reason for error.
```
afrind, line 2260:
```suggestion
* Error Reason: Provides the reason for subscription error. See {{reason-phrase}}.
```

draft-ietf-moq-transport.md
```diff
@@ -3049,6 +3242,11 @@ Extension headers are present. If an endpoint receives a datagram with Type
 0x03 and Extension Headers Length is 0, it MUST close the session with Protocol
 Violation.
 
+* Track Alias: The Track Alias ({{track-alias}} indicating the track this
+  Datagram belongs to.  If an endpoint receives a datagram with an unknown Track
+  Alias, it MAY drop the datagram or choose to buffer it for a brief period to
+  handle reordering with the control message that establishes the Track Alias.
+
```
afrind, line 3249:
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -3277,9 +3267,12 @@ ID` in the SUBGROUP_HEADER.
 
 If an endpoint receives a subgroup with an unknown Track Alias, it MAY abandon
 the stream, or choose to buffer it for a brief period to handle reordering with
-the control message that establishes the Track Alias.  The endpoint SHOULD NOT
-release stream flow control beyond the SUBGROUP_HEADER until the Track Alias has
-been established.  TODO: talk about possible deadlocks.
+the control message that establishes the Track Alias.  The endpoint MAY withhold
+stream flow control beyond the SUBGROUP_HEADER until the Track Alias has been
+established.  To prevent deadlocks, the publisher MUST allocate connection flow
```
fluffy, line 3272: not a request for any changes but just a question .... in other places we said 

```
Given the critical nature of control messages and their relatively
small size, the control stream SHOULD be prioritized higher than all
subscribed Objects.
```

Does that prioritize higher imply this flow control as well ? I sort of feel like all of this should be covered in the {#session-init} or some other place.

afrind: This text was part of #977, not this PR.

> Does that prioritize higher imply this flow control as well ?

I don't think it's guaranteed, though it's quite sensible.  SHOULD vs MUST also needs to be resolved.

>  sort of feel like all of this should be covered in the {#session-init} or some other place.

I added it with Track Alias since that introduced the deadlock condition, but we can move it around.

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -1110,47 +1137,63 @@ to the old relay can be stopped with an UNSUBSCRIBE.
...
+
+When a relay receives an incoming PUBLISH message, it SHOULD send a PUBLISH
+request to each subscriber that has subscribed (via SUBSCRIBE_ANNOUNCES)
+to the track's namespace or prefix thereof.
 
 When a relay receives an incoming ANNOUNCE for a given namespace, for
 each active upstream subscription that matches that namespace, it SHOULD send a
-SUBSCRIBE to the publisher that sent the ANNOUNCE.
+SUBSCRIBE to the publisher that sent the ANNOUNCE.  When it receives an incoming
+PUBLISH message for a track that has active subscribers, it SHOULD respond
```
ianswett, line 1190: By active, we mean that at least one subscription has a Forward State=1?

draft-ietf-moq-transport.md
```diff
@@ -2108,6 +2157,148 @@ SUBSCRIBE_DONE, as defined below:
...
+
+* Request ID: The Request ID of the PUBLISH this message is replying to
+  {{message-publish}}.
+
+* Forward: The Forward State for this subscription, either 0 (don't
+  forward) or 1 (forward).
+
+* Subscriber Priority: The Subscriber Priority for this subscription.
+
+* Group Order: The Group Order preference for this subscription.  This
```
ianswett, line 2242: Future-looking comment: This seems like it'd be better as a param, because I would think that typically you wouldn't want to change GroupOrder?

draft-ietf-moq-transport.md
```diff
@@ -2108,6 +2157,148 @@ SUBSCRIBE_DONE, as defined below:
...
+overwrites the GroupOrder specified PUBLISH.
+
+* Filter Type, Start, End Group: See {{message-subscribe-req}}.
+
+* Parameters: Parameters associated with this message.
+
+## PUBLISH_ERROR {#message-publish-error}
+
+The subscriber sends an PUBLISH_ERROR control message for to reject
```
ianswett, line 2251:
```suggestion
The subscriber sends an PUBLISH_ERROR control message to reject
```

ianswett (approved):

vasilvv (approved):

afrind (commented):
> I'm mildly surprised to see we're confident enough to say that a relay SHOULD simultaneously stream duplicate objects on a track from every available publisher. I'm not going to lie down in the road over it, but it seems really profligate with bandwidth.

This isn't different from the current specified behavior for SUBSCRIBE with multiple publishers announcing the same namespace, right?

draft-ietf-moq-transport.md
```diff
@@ -2108,6 +2157,148 @@ SUBSCRIBE_DONE, as defined below:
...
+* Track Namespace: Identifies a track's namespace as defined in ({{track-name}})
+
+* Track Name: Identifies the track name as defined in ({{track-name}}).
+
+* Track Alias: The identifer used for this track in Subgroups or Datagrams (see
+  {{track-alias}}). The same Track Alias MUST NOT be used to refer to two
+  different Tracks simultaneously. If a subscriber receives a PUBLISH that
+  uses the same Track Alias as a different track with an active subscription, it
+  MUST close the session with error 'Duplicate Track Alias'.
+
```
martinduke, line 2195: Group Order is missing from this text.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
martinduke: This whole time I assumed (based on the word "request") that the Group Order in SUBSCRIBE was advisory, but I see now how one could read it as mandatory.

I agree that the current PR is consistent with the ID that the subscriber always chooses. I can't say if that's wise or not.

draft-ietf-moq-transport.md
```diff
@@ -2696,24 +2887,25 @@ Violation.
...
 
 A subscriber cannot make overlapping namespace subscriptions on a single
-session.  Within a session, if a publisher receives a SUBSCRIBE_ANNOUNCES
-with a Track Namespace Prefix that is a prefix of an earlier
-SUBSCRIBE_ANNOUNCES or vice versa, it MUST respond with
-SUBSCRIBE_ANNOUNCES_ERROR, with error code Namespace Prefix Overlap.
+session.  Within a session, if a publisher receives a SUBSCRIBE_ANNOUNCES with a
+Track Namespace Prefix that is a prefix of an earlier SUBSCRIBE_ANNOUNCES or
+vice versa, it MUST respond with SUBSCRIBE_ANNOUNCES_ERROR, with error code
+Namespace Prefix Overlap.
```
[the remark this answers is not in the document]
martinduke: I see that now. Meh. I don't have a suggested rewording except to just do another sentence that just states the reverse.

Maybe it's just a common sense reading, but I'd prefer it if it was an earlier, **active** SUBSCRIBE_ANNOUNCES rather than just an earlier one.

martinduke:
> > I'm mildly surprised to see we're confident enough to say that a relay SHOULD simultaneously stream duplicate objects on a track from every available publisher. I'm not going to lie down in the road over it, but it seems really profligate with bandwidth.
> 
> This isn't different from the current specified behavior for SUBSCRIBE with multiple publishers announcing the same namespace, right?

> 
"A subscriber MAY send a SUBSCRIBE or FETCH for a track to any publisher. If it has accepted an ANNOUNCE with a namespace that exactly matches the namespace for that track, it SHOULD only request it from the senders of those ANNOUNCE messages."

If you mean this text, I didn't mean it to say "subscribe to all of them," I meant, 'pick as many as you want, but from the set of announcers.' Once again, longstanding text is ambiguous...

Again, this strikes me as potentially very silly advice, but if the real-time and CDN guys tell me this is how it gotta be, I can live with it.

draft-ietf-moq-transport.md
```diff
@@ -1110,47 +1137,63 @@ to the old relay can be stopped with an UNSUBSCRIBE.
...
+new tracks.
 
-Relays MUST verify that publishers are authorized to publish
-the content associated with the set of
-tracks whose Track Namespace matches the announced namespace. Where the
-authorization and identification of the publisher occurs depends on the way the
-relay is managed and is application specific.
+2. Send an ANNOUNCE message for a Track Namespace to the relay. This enables the
+relay to send SUBSCRIBE messages for Tracks in this Namespace in response to
+received SUBSCRIBE messages.
```
suhasHere: Send an ANNOUNCE message for a Track Namespace to the relay. This enables the
relay to send SUBSCRIBE messages **to the publishers** for Tracks in this Namespace in response to the received SUBSCRIBE messages.

draft-ietf-moq-transport.md
```diff
@@ -1110,47 +1137,63 @@ to the old relay can be stopped with an UNSUBSCRIBE.
...
+to determine which publishers receive a SUBSCRIBE or which subscribers receive a
+PUBLISH. For example, a SUBSCRIBE namespace=(foo,bar), track=x message will be
+forwarded to the sessions that sent ANNOUNCE namespace=(foo) and ANNOUNCE
+namespace=(foo, bar) respectively, but not one that sent ANNOUNCE
+namespace=(foobar).  Relays MUST forward SUBSCRIBE messages to all publishers
+and ANNOUNCE and PUBLISH messages to all subscribers that have a namespace
+prefix match.
+
+When a relay receives an incoming SUBSCRIBE that triggers an upstream
+subscription, it SHOULD send a SUBSCRIBE request to each publisher that has
```
suhasHere, line 1221:
```suggestion
subscription, it MUST send a SUBSCRIBE request to each publisher that has
```

draft-ietf-moq-transport.md
```diff
@@ -1110,47 +1137,63 @@ to the old relay can be stopped with an UNSUBSCRIBE.
...
+announced the subscription's namespace or prefix thereof, unless it already has
+an active subscription for the Objects requested by the incoming SUBSCRIBE
+request from all available publishers.  If it already has a matching upstream
+subscription in Forward State=0, it SHOULD send a SUBSCRIBE_UDPATE with
```
suhasHere, line 1225:
```suggestion
subscription in Forward State=0, it MUST send a SUBSCRIBE_UDPATE with
```

draft-ietf-moq-transport.md
```diff
@@ -1110,47 +1137,63 @@ to the old relay can be stopped with an UNSUBSCRIBE.
...
+Forward=1 to all publishers.
+
+When a relay receives an incoming PUBLISH message, it SHOULD send a PUBLISH
```
```suggestion
When a relay receives an incoming PUBLISH message, it MUST send a PUBLISH
```

draft-ietf-moq-transport.md
```diff
@@ -1110,47 +1137,63 @@ to the old relay can be stopped with an UNSUBSCRIBE.
...
+request to each subscriber that has subscribed (via SUBSCRIBE_ANNOUNCES)
+to the track's namespace or prefix thereof.
 
 When a relay receives an incoming ANNOUNCE for a given namespace, for
 each active upstream subscription that matches that namespace, it SHOULD send a
```
suhasHere, line 1233:
```suggestion
each active upstream subscription that matches that namespace, it MUST send a
```

draft-ietf-moq-transport.md
```diff
@@ -1110,47 +1137,63 @@ to the old relay can be stopped with an UNSUBSCRIBE.
...
-SUBSCRIBE to the publisher that sent the ANNOUNCE.
+SUBSCRIBE to the publisher that sent the ANNOUNCE.  When it receives an incoming
+PUBLISH message for a track that has active subscribers, it SHOULD respond
```
suhasHere, line 1235:
```suggestion
PUBLISH message for a track that has active subscribers, it MUST respond
```

draft-ietf-moq-transport.md
```suggestion
relay to send SUBSCRIBE messages to publishers for Tracks in this Namespace in
response to received SUBSCRIBE messages.
```

draft-ietf-moq-transport.md
afrind: This SHOULD was there before.

draft-ietf-moq-transport.md
afrind: This is for consistency with how we define Forward for SUBSCRIBE - I think it says the relay SHOULD send Forward=1 upstream.

draft-ietf-moq-transport.md
suhasHere: I understand that and it will be nice if we can update it

draft-ietf-moq-transport.md
suhasHere: works for me to leave as-is.

draft-ietf-moq-transport.md
suhasHere: I think this will be needed to ensure the behavior across multiple relay implementations.  Open to hear thoughts if we think otherwise too.

draft-ietf-moq-transport.md
afrind: Keeping forward state normative language matching SUBSCRIBE here

draft-ietf-moq-transport.md
afrind: I filed #1037 so we can track that separately from this PR, which is introducing PUBLISH mechanics.

afrind:
Discussed in 6/4 interim.  Decision is to move forward with existing normative multi-publisher language and resolve in #1037.

Issue #508, opened 2024-08-23
Simplify naming by removing the distinction between Track Namespace and Track Name


wilaw:
Alan's PR https://github.com/moq-wg/moq-transport/pull/498 introduced a N-tuple construct for namespace, in which the namespace is made up of an ordered sequence of between 1 - 32 byte arrays. This allows us to construct namespaces in the form

`example.com` `meeting` `abc123` `alice`

Alice would publish tracks under this namespace called audio|video|catalog leading to the final track identifiers being

`example.com` `meeting` `abc123` `alice` `audio`
`example.com` `meeting` `abc123` `alice` `video`
`example.com` `meeting` `abc123` `alice` `catalog`

At this point, the distinction between the track name and last stanza of the namespace is arbitrary and I would argue, unnecessary.

We could remove the separate concepts of TRACK-NAMESPACE and TRACK-NAME and instead coalesce this down to a single construct of TRACK-NAME, which would hold the same definition as a bounded N-tuple. So under the simplified scheme, the TRACK-NAMES would be 

`example.com` `meeting` `abc123` `alice` `audio`
`example.com` `meeting` `abc123` `alice` `video`
`example.com` `meeting` `abc123` `alice` `catalog`

Alice as the publisher would ANNOUNCE the _minimum prefix_ which uniquely describes describes the content she is responsible for, in this case, she would announce

`example.com` `meeting` `abc123` `alice`

This schema then allows for easy aggregate subscription. We could remove the need for SUBSCRIBE-NAMESPACE. For example, a subscriber wanting to receive everything produced by Alice would SUBSCRIBE  to 

`example.com` `meeting` `abc123` `alice`

Someone wanting to receive all feeds from meeting abc123 would SUBSCRIBE to

`example.com` `meeting` `abc123`

If we standardize some means of communicating a null (*) stanza, and if different publishers produced

`example.com` `meeting` `abc123`
`statistics.com` `metrics` `abc123`

the someone wanting both the meeting and metrics for abc123 could SUBSCRIBE to 

`    ` `    ` `abc123`

suhasHere:
I generally like the suggestion here except the  last <null> <null> abc123.  If we allow nulls or (.*) regex parsing in between , it just makes things pretty slow.  We can start with prefix matching to begin with,

alvestrand:
Having complete names from the name root to the tree node you want to point
to is usually a Good Thing (experience with DNS searchpaths).
[...]

PR #1008, opened 2025-05-07
Clarify FTN construction

### LINKED ISSUE #1005: Disallow a namespace and track name with the same tuple set
Having a `Full Track Name = (a, b), c` should mean that there cannot be a Track Namespace `(a, b, c)`.

Related to #764

suhasHere:
Fixes #1005

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -488,6 +488,7 @@ Violation.
 Track Name is a sequence of bytes that identifies an individual track within the
 namespace.
 
+
 The maximum total length of a Full Track Name is 4,096 bytes, computed as the
 sum of the lengths of each Track Namespace tuple field and the Track Name length
 field.  If an endpoint receives a Full Track Name exceeding this length, it MUST
@@ -500,6 +501,10 @@ exact comparison of the bytes. Specifications that use MoQ Transport may
 constrain the information in these fields, for example by restricting them to
 UTF-8. Any specification that does needs to specify the canonicalization into
 the bytes in the Track Namespace or Track Name such that exact comparison works.
+
+Repeating the bytes corresponding to tuples between the Track Namespace 
+and the Track Name is an invalid construction. Endpoints encountering such 
+a construction MUST terminate the session with a 'Protocol Violation' error.
 
 ### Scope {#track-scope}
 


draft-ietf-moq-transport.md
```diff
@@ -501,6 +502,10 @@ constrain the information in these fields, for example by restricting them to
 UTF-8. Any specification that does needs to specify the canonicalization into
 the bytes in the Track Namespace or Track Name such that exact comparison works.
 
+Repeating the bytes corresponding to tuples between the Track Namespace 
+and the Track Name is an invalid construction. Endpoints encountering such 
+a construction MUST terminate the session with a 'Protocol Violation' error.
+
```
afrind, line 508:
```suggestion
Repeating the bytes corresponding to tuples between the Track Namespace
and the Track Name is an invalid construction. Endpoints encountering such
a construction MUST terminate the session with a 'Protocol Violation' error.

```

draft-ietf-moq-transport.md
```diff
@@ -488,6 +488,7 @@ Violation.
 Track Name is a sequence of bytes that identifies an individual track within the
 namespace.
 
+
```
afrind, line 491:
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -501,6 +501,10 @@ constrain the information in these fields, for example by restricting them to
 UTF-8. Any specification that does needs to specify the canonicalization into
 the bytes in the Track Namespace or Track Name such that exact comparison works.
 
+Repeating the bytes corresponding to tuples between the Track Namespace
+and the Track Name is an invalid construction. Endpoints encountering such
+a construction MUST terminate the session with a 'Protocol Violation' error.
```
fluffy, line 506: Uh, I think I know what you are getting at but this seems pretty confusing. How about 

Publishers MUST NOT create a tracks names that are the same as namespace tuples inside the same namespace.

fluffy (changes requested):

fluffy:
And to be clear, I agree with the design here - just hard to figure out how to phrase the text

draft-ietf-moq-transport.md
afrind: Or something like this?

The combined tuple consisting of a Full Track Name's Namespace tuple fields and its Track Name MUST NOT also be a Track Namespace or a prefix of a Track Namespace.

wilaw:
> From draft-11 - A MOQT scope is a set of servers (as identified by their connection URIs) for which the tuple of Track Name and Track Namespace are guaranteed to be unique and identify a specific track

The fact that we concatenate namespace and name down to a common tuple , and that we need to consider this PR in order to disambiguate the two, suggests we should re-evaluate [Simplify naming by removing the distinction between Track Namespace and Track Name](https://github.com/moq-wg/moq-transport/issues/508#top). The primary argument for closing that was the complexity of wildcard subscriptions. But that was prior to the PUBLISH workflow being considered. If a client wants to subscribe to all of example.com, and they have the permissions to, then let them. Relays never need to know the differentiation of namespace and names as they only ever operate on the concatenation of the two. ANNOUNCE is naturally wildcard-enabled, as is SUBSCRIBE_ANNOUNCES.

afrind (approved):

draft-ietf-moq-transport.md
```diff
@@ -512,6 +512,11 @@ constrain the information in these fields, for example by restricting them to
 UTF-8. Any specification that does needs to specify the canonicalization into
 the bytes in the Track Namespace or Track Name such that exact comparison works.
 
+To avoid confusion and potential ambiguity, the combined tuple consisting of a
+Full Track Name's Namespace tuple fields and its Track Name MUST NOT also be a
+Track Namespace or a prefix of a Track Namespace. Endpoints encountering such a
+construction MUST terminate the session with a 'Protocol Violation' error.
```
ianswett, line 561: I'm not sure this will be trivial to check for all implementations, so I'd be inclined towards SHOULD.

suhasHere:
@afrind @ianswett wondering if this is ready to merge . Please let me know . Thanks

afrind:
> wondering if this is ready to merge . Please let me know . Thanks

This PR lacks sufficient reviews/approvals to merge.  Our process requires 3 approvals for Editorial and Minor Design, or all Editors and Named Authors for Design.

suhasHere:
Makes sense . How should we encourage reviews or feedback.

draft-ietf-moq-transport.md
ianswett: I guess I'm ok with the "encountering" bit, since if one doesn't encounter it you don't need to terminate the connection.

ianswett (approved):

vasilvv:
It's not entirely clear to me as to how exactly this requirement is supposed to be enforced.  If I am a random client trying to publish `(a, b, c, d)`, and `(a, b, c)` already exists, the relay would just reset the connection as a result?  This seems a bit odd.

wilaw:
I agree with Victor's comment. I don't think a combined tuple should ever be produced and then compared. If we allow that, then we are de facto making name the last entry in the namespace tuple. 

The namespace and name should be compared separately:

NS[a,b,c] N[d] is then not equivalent to NS[a,b,c,d] N[]

afrind:
Victor's example doesn't show the problem we're trying to solve

> If I am a random client trying to publish (a, b, c, d), and (a, b, c) already exists

It should be

ns=(a, b, c), n=d and ns=(a, b, c, d) already exists.

Will is correct that there likely isn't a _functional_ problem here, eg: 

> NS[a,b,c] N[d] is then not equivalent to NS[a,b,c,d] N[]

This is correct.  SUBSCRIBE and PUBLISH have Namespace and Track separately, SUBSCRIBE_ANNOUNCES and PUBLISH contain only Namespace.

I think it's safe to say that applications probably shouldn't define namespaces and names that overlap like this -- but it sounds like folks would prefer we not make it an error if they do?

martinduke:
As an individual, I agree with those who say this is more trouble than it's worth. It's a bit of code, a bit of text, an unclear enforcement mechanism, and something that I don't see as an actual problem that will arise. Why should MoQT care if an application deploys a confusing namespace hierarchy?

afrind:
Discussed in the interim and decided to close without action.

Issue #1009, opened 2025-05-07
End of Group marking


fluffy:
We need to be able to do this with a flag, and we also need to be able to it with option empty object. 

The empty object is for when you don't know when encorder will sipit out new i frame 

The flag is good when you have one audio object per group.

afrind:
For DATAGRAM, it's easy enough to use one of the Type bits.  We could use a Type bit for SUBGROUP_HEADER also, meaning "final object before FIN (if there is one) is EOG".  Works for single-object streams and single-stream-groups.  If you are using multiple subgroups-per-group and multiple objects-per-subgroup, you just have to send a full EOG object.  If a stream has "Final Object EOG" set but is reset instead of FIN, you don't know the EOG.

martinduke:
Related to #877, so not too late

fluffy:
I think there are at least two use cases that are important here. 

The first is when the publishers will know the object is the last and the other is when it does not know. 

last object known case: For example, when doing audio with 1 object per group, the publisher knows the object it is sending is the last one. I think it has to be able to indicate this in a bit efficient way to make low bitrate audio work. 

last object not known case: For example, when going stream per I-frame, a video encoder outputs blocks of bits and they get sent as object, but after they are sent, the video codec starts a new i-frame that starts a new group. The last object in the previous group has already been sent. There has to be some way to send some additionally info that indicates what the last object in the group was even though that will likely arrive later than the last object in the group. 

I think there are many ways that we could solve both the above. The property I want in a solution is that it can work for datagrams not just streams ( will be unreliable of course ), will be in the data plane not the control plane because there will be lots of these and want to minimize impact on relay performance, and finally will keep the bitrate low particularly for low bitrate audio.

PR #1011, opened 2025-05-08
Allow End of Group to be signaled in a non-zero length object

afrind:
This works best for datagrams, groups with a single subgroup, or subgroups with a single object.  Multi-subgroup groups with multiple objects per subgroup will probably continue to signal End Of Group with a zero-length Status object.

Fixes: #1009

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2854,9 +2854,9 @@ the datagram.
 |-----------|---------------------------------------------------|
 | ID        | Type                                              |
 |----------:|:--------------------------------------------------|
-| 0x00-0x01 | OBJECT_DATAGRAM ({{object-datagram}})             |
+| 0x00-0x03 | OBJECT_DATAGRAM ({{object-datagram}})             |
 |-----------|---------------------------------------------------|
-| 0x02-0x03 | OBJECT_DATAGRAM_STATUS ({{object-datagram}})      |
+| 0x04-0x05 | OBJECT_DATAGRAM_STATUS ({{object-datagram}})      |
 |-----------|---------------------------------------------------|
 
 An endpoint that receives an unknown stream or datagram type MUST close the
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2992,7 +2995,7 @@ will be dropped.
 
 ~~~
 OBJECT_DATAGRAM {
-  Type (i),
+  Type (i) = 0x0-0x4,
```
fluffy, line 2998: stylistic suggestion ... why don't you make type be defined as 4 bit reserved value of 0, followed by 4 one bit fields that are the flags. I realize this there are a few points in the table that are not used that your current encoding disallows but you still need to deal with an error when you get that value.  It just seems like it would be easier to describe and code.    

That said, I don't deeply care about the how we spell this and like that this is very compact. From a design point of view, I think does what is needed.

ianswett (approved):

suhasHere (approved):

vasilvv (approved):

martinduke:
I am considering an issue to back out this PR. This is adding a bunch of code complexity to save a few bytes per group.

Allowing EOG + Payload has several problems:
(1) It can't be efficiently expressed in a FETCH stream.
(2) When I receive a complete object on a EOG stream, I don't know the status until I receive either another byte or a FIN (and thus can't deliver it to the application unless I invent an API to revise the status after delivery)
(3) @afrind points out new malformed track definitions in #1093

Having the EOG flag create an implied object (with no extension headers) fixes these but has its own problems
(1) An original publisher and a relay cannot open a stream of this type unless it knows a priori that it will not add an extension to the EOG object. This is a weird and subtle thing.
(2) It is unclear what a FETCH stream should do if the last object has ID VARINT62_MAX, so an Original publisher has to commit to not mint this Object ID at the start of the stream.

In either case, an OP can't initiate the stream type unless it knows at the start that this subgroup will contain the largest normal Object ID.

There are several implied commitments that amount to a pretty big foot gun.  A good MoQT implementation would check that applications aren't violating all of these commitments, or just never express a stream using the EOG bit.

Issue #1010, opened 2025-05-07
In publish, need equivalent of subscribe expires


fluffy:
When A sends a publish to a relay, there needs to be a way for relay to say when A needs to redo the auth for the Publish. ( and there needs to be a way to do that like PUBLISH_UPDATE )

afrind:
The requirement is to notify the sender of an authentication token an idea of how long they have to renew the auth before the subscription is terminated.  Since auth is carried in SUBSCRIBE, PUBLISH then the expires needs to go in SUBSCRIBE_OK and PUBLISH_OK, and also why we need PUBLISH_UPDATE.

See also #879, I think `expires` is a good candidate for a parameter.

afrind:
Is it correct to say that ANNOUNCE_OK and SUBSCRIBE_ANNOUNCES_OK should also have expires?  It seems like it's relevant to all of them.

gwendalsimon:
I am a bit surprised by this issue. 

If I understand correctly: The sender of an auth token sends a token (that it got from a third party), and the server processes the token, extracts the expiration time, and includes the expiration time in the confirmation. Am I correct?

It seems odd: One of the reasons tokens are encrypted is that the issuer does not want anyone on the path to know the claims.

It is annoying in practice: When a CDN edge server acts as Relay, the token validation system usually runs in a separate module. The criteria to analyze the token are out of the Relay's knowledge. We would need to create APIs between Relay and "token validation module" to collect some information from the token claims. 

Why would we restrict this mechanism (token information sharing) to the "expiration" claim, which is only one of the claims in the token?

Note that, in CAT, "_catr_" is a claim that allows a token validation system to trigger some actions when a token is close to expire. It runs at the server-side (in the module that processes the token), which is the natural behavior. Possible actions includes the generation of a new token or re-routing this client to the token issuer. Rather than an "expires" parameter of OK messages, I would consider embedding some MOQT actions in the catr's claim of CAT.

PR #1012, opened 2025-05-08
Make expires a parameter

afrind:
This makes it easily optional and added to all possible messages

Fixes: #1010 
Fixes: #1026

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1741,8 +1741,12 @@ The EXPIRES parameter (Parameter Type 0x8) MAY appear in SUBSCRIBE_OK, PUBLISH
 or PUBLISH_OK (TOOD: or REQUEST_OK).  It is a variable length integer encoding
 the time in milliseconds after which the sender of the parameter will terminate
 the subscription.  The receiver of the parameter can extend the subscription by
-sending a SUBSCRIBE_UPDATE (TODO: SUBSCRIPTION_UPDATE).  This value is advisory
-and the sender can terminate the subscription prior to or after the expiry time.
+sending a SUBSCRIBE_UPDATE (TODO: SUBSCRIPTION_UPDATE). If the receiver of the
+parameter has one or more updated AUTHORIZATION_TOKENs, it SHOULD include those
+in the SUBSCRIBE_UPDATE. Relays that send this parameter and applications that
+receive it MAY introduce jitter to prevent many endpoints from updating
+simultaneously. This value is advisory and the sender can terminate the
+subscription prior to or after the expiry time.
 
 If the EXPIRES parameter is 0 or is not present in a message, the subscription
 does not expire or expires at an unknown time.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1485,6 +1485,8 @@ integer encoding the time in milliseconds after which the sender will terminate
 the subscription or announcement.  This value is advisory and can terminate the
 subscription or announcement prior to or after the expiry time.
 
+If the EXPIRES parameter is 0 or is not present in a message, the subscription
+or announcement does not expire or expires at an unknown time.
 
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1476,6 +1476,16 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+
+#### EXPIRES Parameter {#expires}
+
+The EXPIRES parameter (Parameter Type 0x8) MAY appear in SUBSCRIBE_OK,
+SUBSCRIBE_ANNOUNCES_OK, ANNOUNCE_OK or TRACK_STATUS_OK.  It is a variable length
+integer encoding the time in milliseconds after which the sender will terminate
+the subscription or announcement.  This value is advisory and can terminate the
+subscription or announcement prior to or after the expiry time.
+
+
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
 
 The `CLIENT_SETUP` and `SERVER_SETUP` messages are the first messages exchanged
@@ -1804,7 +1814,6 @@ SUBSCRIBE_OK Message {
   Type (i) = 0x4,
   Length (16),
   Request ID (i),
-  Expires (i),
   Group Order (8),
   Content Exists (8),
   [Largest Location (Location)],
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1476,6 +1476,16 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
 re-requests them.
 
+
+#### EXPIRES Parameter {#expires}
+
+The EXPIRES parameter (Parameter Type 0x8) MAY appear in SUBSCRIBE_OK,
+SUBSCRIBE_ANNOUNCES_OK, ANNOUNCE_OK or TRACK_STATUS_OK.  It is a variable length
+integer encoding the time in milliseconds after which the sender will terminate
+the subscription or announcement.  This value is advisory and can terminate the
+subscription or announcement prior to or after the expiry time.
```
suhasHere, line 1486: There needs to be a default value to implementations to know when the parameter is omitted.

suhasHere (commented):
Expires seems to be a mandatory information and not optional parameter. Either we specify defaults or make it a field .

draft-ietf-moq-transport.md
```diff
@@ -1476,6 +1476,16 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+
+
```
afrind, line 1488:
```suggestion
If the EXPIRES parameter is 0 or is not present in a message, the subscription
or announcement does not expire or expires at an unknown time.

```

vasilvv (approved):

afrind:
@suhasHere are you good with the default specified?

draft-ietf-moq-transport.md
```diff
@@ -1476,6 +1476,18 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+
+#### EXPIRES Parameter {#expires}
+
+The EXPIRES parameter (Parameter Type 0x8) MAY appear in SUBSCRIBE_OK,
+SUBSCRIBE_ANNOUNCES_OK, ANNOUNCE_OK or TRACK_STATUS_OK.  It is a variable length
```
suhasHere, line 1483: I don't see why we need expires for TrackStatus or Announce.

suhasHere (commented):
I am not sure proposal of expires for anything other than Subscribe or probably Subscribe_Announces

draft-ietf-moq-transport.md
afrind: @suhasHere 

TRACK_STATUS_OK is there because it mirrors SUBSCRIBE_OK.

ANNOUNCE is there because the purpose of expires is to indicate to the peer when the state established by the request will be terminated.  ANNOUNCE creates state so it seems logical to me -- also the parameter is optional: if you never expire announces, don't send it.

fluffy (changes requested):
This is way too vague. I think we have to get very explicit what this is for. We need text that the client has to renew before the expiry time. We need to talk about how much before and randomness so they don't all happen at the same time.  This should not be optional. It should not default to infinity. We need to talk about after the time is exceeded, relay can just remove the subscribe state with no signalling to the subscriber.  Termination before this require signaling to the subscriber. I view this as mandatory not advisory - relays need to be able to count on subscfribers doing it or we can't make an authorization refresh scheme that works.

afrind:
> This is way too vague. I think we have to get very explicit what this is for. We need text that the client has to renew before the expiry time. We need to talk about how much before and randomness so they don't all happen at the same time. This should not be optional. It should not default to infinity. We need to talk about after the time is exceeded, relay can just remove the subscribe state with no signalling to the subscriber. Termination before this require signaling to the subscriber. I view this as mandatory not advisory - relays need to be able to count on subscfribers doing it or we can't make an authorization refresh scheme that works.

@fluffy Much of what you are suggesting applies to the draft-10 `expires` text also, since this PR only moves it, accepts in multiple messages, and assigns it a default value when absent.  I think the 0 value is the only sensible default if we make the parameter optional.  I filed #1026 to track your new requirements.  We can park this until that is resolved.

draft-ietf-moq-transport.md
ianswett: I don't think I know what Expires does in TRACK_STATUS_OK so I'd be inclined to restrict it to the first 3.

ianswett (commented):
LG, but I have no idea what Expires does in TRACK_STATUS_OK

gwendalsimon:
I am very not convinced by this parameter. I strongly call for more discussion on what we want to achieve here.

First we must disambiguate:
- **token re-validation time**: Today's CDN edge servers delivering ABR do not revalidate the claims of an already validated token because, by nature, requests arrive with a token every x seconds. In the case of MoQ, we must revalidate the claims because the token is processed only at the connection and some characteristics of the client may change. Some ideas for notifying revalidation time are discussed [here](https://github.com/wilaw/CAT-4-MOQT/pull/9#discussion_r2073679318)
- **token expiration time**: It is an optional claim in CAT. Expiration time usually serves two purposes: 1/ right management (this content is available just for a given time); and 2/ prevent CDN leeching by forcing clients to re-authenticate. In MoQ, the second motivation is much less significant because the tokenized connection is established between client and server, so token replay is much easier to prevent.

Forcing a token renewal just for the sake of revalidating it is a particularly complex implementation (to not say a bad idea), especially with respect to how challenging token renewal is in practice.

Second, many content providers do not use expiration time in tokens.
- actually many content providers do not implement tokens at all. Public TV broadcasters for example do not need to protect content. The more viewers the better.
- some content providers serve content infinitely, for example a TV screen in an airport plays 24/7/365 and fixed video surveillance. We want a token identifier for various reasons at connection (analytics, 

Making expires a mandatory parameter is quite a forceful push. I am opposed to force everyone to use parameters just because a subset of cases need it.

Third, the token issuer and the token edge validator may want to keep their conversation (via token) secret. The Client and the Relay are regarded as possible leaks. The least information is revealed, the best. Encrypting tokens serves this purpose, Expiration time and (possible) re-validation time are crucial information, which the content protection scheme may want to keep encrypted.

draft-ietf-moq-transport.md
afrind: > TRACK_STATUS_OK is there because it mirrors SUBSCRIBE_OK.

eg: if SUBSCRIBE_OK would have returned it then TRACK_STATUS_OK returns it too, right?

ianswett: But TRACK_STATUS_OK is the end of the transaction, whereas SUBSCRIBE_OK is not.

ianswett (approved):
Minus the oddity of TRACK_STATUS_OK, I think this is a good change and I'd like to keep this separate from the requirements of Expires itself if we can?

draft-ietf-moq-transport.md
afrind: I'm going to remove the Namespace related messages pending #1267

suhasHere (approved):

Issue #997, opened 2025-05-05
Editorial: "AUTHORIZATION TOKEN" vs "TOKEN"


yekuiwang:
Sec. 8.2.1.1 includes the following:

> The AUTHORIZATION TOKEN parameter MAY be repeated within a message.
> 
> The TOKEN value is a structured object containing an optional session-specific alias. The Alias allows the client to reference a previously transmitted TOKEN in future messages. The TOKEN value is serialized as follows:
> 
> TOKEN {
>   Alias Type (i),
>   [Token Alias (i),]
>   [Token Type (i),]
>   [Token Value (..)]
> }
> Figure 4: AUTHORIZATION TOKEN value

Would it be better, for consistency purpose, to replace "TOKEN" in lines other than the top line and the bottom line with "AUTHORIZATION TOKEN"? I.e., change the above to be as follows:

> The AUTHORIZATION TOKEN parameter MAY be repeated within a message.
> 
> The AUTHORIZATION TOKEN value is a structured object containing an optional session-specific alias. The Alias allows the client to reference a previously transmitted AUTHORIZATION TOKEN in future messages. The AUTHORIZATION TOKEN value is serialized as follows:
> 
> AUTHORIZATION TOKEN {
>   Alias Type (i),
>   [Token Alias (i),]
>   [Token Type (i),]
>   [Token Value (..)]
> }
> Figure 4: AUTHORIZATION TOKEN value

afrind:
Generally structs are Camel Case but the parameters are ALL_CAPS.  Rename the structure "Token" or "Authorization Token"

Issue #984, opened 2025-05-02
Problems with the Max Auth Token Cache Size description


martinduke:
Sec 8.2.1.1.

> Any message carrying an AUTHORIZATION TOKEN with Alias Type REGISTER that does not result in Malformed Auth Token MUST effect the token registration, even if the message fails for other reasons, including Unauthorized. This allows senders to pipeline messages that refer to previously registered tokens.

first of all, s/effect/affect.

<incorrect issue deleted>

I don't know what it means to register a token you won't honor (i.e. do I have to store the value?). 

I would not be sad if we decided this was too much to take on and punted MOQPACK to a protocol extension.

See below for other minor issues.

martinduke:
Furthermore, if we're counting bytes, then it should be 8B for the alias + 8B for the token_type + token size, not token_size + 8

martinduke:
Oof:
Sec 8.3.2.3:

> The token size is calculated as 8 bytes + the size of the Token value (see [Figure 4](https://www.ietf.org/archive/id/draft-ietf-moq-transport-11.html#moq-token)).

If I go to Figure 4, it describes a structure in which one field is "Token Value", but the *title* of the figure is also 'Figure 4. AUTHORIZATION TOKEN value]. I assume you mean the length of the field, not the struct, but the nomenclature could stand to be cleaned up here.

martinduke:
I withdraw the original core of the issue, because overflow became a session error and so storing it doesn't matter. But the other notes apply.

afrind:
> first of all, s/effect/affect.

The intent was to use effect as a verb "meaning to cause something to happen or to bring about a result", but the subject of the sentence is the token, which doesn't effect anything. Editors will address.

> I don't know what it means to register a token you won't honor (i.e. do I have to store the value?).

The intent that so long as the registration is not a session error, then a subsequent message using the alias is also not a session error.  Yes, you have to store the value.  The key example is eg: an expired token, a token with an invalid signature, or a token that does not authorize the operation carrying it.  Yes, MoQT needs to store such tokens until deleted by the sender.

> I would not be sad if we decided this was too much to take on and punted MOQPACK to a protocol extension.

@wilaw who was a primary advocate.  Note that it is already _entirely optional_ both for the sender and receiver.  The sender can use `USE_VALUE` and the receiver can set the max size to 0.

> Furthermore, if we're counting bytes, then it should be 8B for the alias + 8B for the token_type + token size, not token_size + 8

[...]

Issue #970, opened 2025-04-29
Remove Duplicate Token Alias Error


suhasHere:
Indicating that an alias is a duplicate is giving away more information to the requestor about the token aliases in use. How about we just have Invalid Token Alias as the common error

Issue #967, opened 2025-04-28
Are Authorization Aliases unidirectional?


martinduke:
Sec 8.2.1.1 sayeth:

> The TOKEN value is a structured object containing an optional session-specific alias. The Alias allows the client to reference a previously transmitted TOKEN in future messages. 

and later

> The Token Alias MUST be unique within the Session.

But 'client' is neither here nor there. A client can be a subscriber or a publisher, or both. Even if single-role, subscribers use tokens for SUBSCRIBE, FETCH, AND TRACK_STATUS_REQUEST. Publishers use it for ANNOUNCE. So clearly, both endpoints can use tags.

So either
1) the tables are separate for each direction. This certainly allows asymmetric tag cache sizes to coexist easily.
2) the tables are common to both directions. In which case, I think the tag IDs need even/odd separation to avoid collisions in-flight, and the cache size has to be min(client, server)

I think I prefer (1) but don't really care.

ianswett:
My understanding is that they were separate for each direction (1).  This makes sense for all the auth situations I can think of.

martinduke:
There are other references to 'client' in the section, which should be replaced as well.

PR #1013, opened 2025-05-08
Allow AUTHORIZATION TOKEN parameter in CLIENT_SETUP and other Auth fixes

afrind:
* Updating the description of the Unauthorized Session Error Code
* Adding other Session Error Codes that can be used by Auth
* Make Unknown Token Alias a session error and remove from message error tables
* Clarify various error conditions
* Replaced "client" with "sender"
* Consistent capitalization of Alias and session
* TOKEN -> Token structure

Fixes: #721
Fixes: #839
Fixes: #964 
Fixes: #967
Fixes: #970 
Fixes: #984
Fixes: #997

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1446,9 +1446,9 @@ Token {
 
 
 * Token Alias - a Session-specific integer identifier that references a Token
-  Value. There are separate Alias spaces for the client and server (eg: they can
-  each register Alias=1). Once a Token Alias has been registered, it cannot be
-  re-registered by the same sender in the Session without first being
+  Value. There are separate Alias spaces for the client and server (e.g.: they
+  can each register Alias=1). Once a Token Alias has been registered, it cannot
+  be re-registered by the same sender in the Session without first being
   deleted. Use of the Token Alias is optional.
 
 * Token Type - a numeric identifier for the type of Token payload being
@@ -1477,7 +1477,8 @@ in the token cache, even if the message fails for other reasons, including
 previously registered tokens without potentially terminating the entire Session.
 A receiver MAY store an error code (eg: Unauthorized or Malformed Auth Token) in
 place of the Token Type and Token Alias if any future message referencing the
-Token Alias will result in that error.
+Token Alias will result in that error. The size of a registered cache entry
+includes the length of the Token Value, regardless of whether it is stored.
 
 If a receiver detects that an authorization token has expired, it MUST retain
 the registered Alias until it is deleted by the sender, though it MAY discard
@@ -1654,7 +1655,7 @@ See {{authorization-token}}.  The endpoint can specify one or more tokens in
 CLIENT_SETUP or SERVER_SETUP that the peer can use to authorize MOQT session
 establishment.
 
-If a server receives an AUTHORIZATION TOKEN parmeter in CLIENT_SETUP with Alias
+If a server receives an AUTHORIZATION TOKEN parameter in CLIENT_SETUP with Alias
 Type REGISTER_TOKEN that exceeds its MAX_AUTH_TOKEN_CACHE_SIZE, it MUST NOT fail
 the session with `Auth Token Cache Overflow`.  Instead, it MUST treat the
 parameter as Alias Type USE_VALUE.  A client MUST handle registration failures
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -668,13 +668,18 @@ code, as defined below:
 |------|---------------------------|
 | 0x15 | Version Negotiation Failed|
 |------|---------------------------|
+| 0x16 | Malformed Auth Token      |
+|------|---------------------------|
+| 0x17 | Unknown Auth Token Alias  |
+|------|---------------------------|
+| 0x18 | Expired Auth Token        |
+|------|---------------------------|
 
 * No Error: The session is being terminated without an error.
 
 * Internal Error: An implementation specific error occurred.
 
-* Unauthorized: The endpoint breached an agreement, which MAY have been
- pre-negotiated by the application.
+* Unauthorized: The client is not authorized to establish a session.
 
 * Protocol Violation: The remote endpoint performed an action that was
   disallowed by the specification.
@@ -714,10 +719,18 @@ code, as defined below:
   the size of all registered Authorization tokens has been exceeded.
 
 * Duplicate Auth Token Alias - Authorization Token attempted to register an
-  alias that was in use (see {{authorization-token}}).
+  Alias that was in use (see {{authorization-token}}).
 
 * Version Negotiation Failed: The client didn't offer a version supported
   by the server.
+
+* Malformed Auth Token - Invalid Auth Token serialization during registration
+  (see {{authorization-token}}).
+
+* Unknown Auth Token Alias - Authorization Token refers to an Alias that is
+  not registered (see {{authorization-token}}).
+
+* Expired Auth Token - Authorization token has expired {{authorization-token}}).
 
 An endpoint MAY choose to treat a subscription or request specific error as a
 session error under certain circumstances, closing the entire session in
@@ -1333,26 +1346,27 @@ these parameters to appear in Setup messages.
 
 #### AUTHORIZATION TOKEN {#authorization-token}
 
-The AUTHORIZATION TOKEN parameter (Parameter Type 0x01) identifies a track's
-authorization information in a SUBSCRIBE, SUBSCRIBE_ANNOUNCES, ANNOUNCE
-TRACK_STATUS_REQUEST or FETCH message. This parameter is populated for
-cases where the authorization is required at the track or namespace level.
+The AUTHORIZATION TOKEN parameter (Parameter Type 0x03) MAY appeat in a
+CLIENT_SETUP, SUBSCRIBE, SUBSCRIBE_ANNOUNCES, ANNOUNCE TRACK_STATUS_REQUEST or
+FETCH message. This parameter converys information to authorize the sender to
+perform the operation carrying the parameter.
 
 The AUTHORIZATION TOKEN parameter MAY be repeated within a message.
 
-The TOKEN value is a structured object containing an optional session-specific
-alias. The Alias allows the client to reference a previously transmitted TOKEN
-in future messages. The TOKEN value is serialized as follows:
+The parameter value is a Token structure containing an optional session-specific
+Alias. The Alias allows the sender to reference a previously transmitted Token
+Type and Token Value in future messages. The Token structure is serialized as
+follows:
 
 ~~~
-TOKEN {
+Token {
   Alias Type (i),
   [Token Alias (i),]
   [Token Type (i),]
   [Token Value (..)]
 }
 ~~~
-{: #moq-token format title="AUTHORIZATION TOKEN value"}
+{: #moq-token format title="Token structure"}
 
 * Alias Type - an integer defining both the serialization and the processing
   behavior of the receiver. This Alias type has the following code points:
@@ -1367,7 +1381,7 @@ TOKEN {
 |------|------------|------------------------------------------------------|
 | 0x1  | REGISTER   | There is an Alias, a Type and a Value. This Alias    |
 |      |            | MUST be associated with the Token Value for the      |
-|      |            | duration of the Session or it is deleted. This action|
+|      |            | duration of the session or it is deleted. This action|
 |      |            | is termed "registering" the Token.                   |
 |------|------------|------------------------------------------------------|
 | 0x2  | USE_ALIAS  | There is an Alias but no Type or Value. Use the Token|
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -714,11 +719,19 @@ code, as defined below:
...
+  Alias that was in use (see {{authorization-token}}).
 
 * Version Negotiation Failed: The client didn't offer a version supported
   by the server.
 
+* Malformed Auth Token - Invalid Auth Token serialization during registration
+  (see {{authorization-token}}).
+
+* Unknown Auth Token Alias - Authorization Token refers to an Alias that is
+  not registered (see {{authorization-token}}).
```
suhasHere, line 731:
```suggestion
* Unknown Auth Token Alias - No registered token found for the provided  Alias (see {{authorization-token}}).
```

draft-ietf-moq-transport.md
```diff
@@ -1333,26 +1346,27 @@ these parameters to appear in Setup messages.
 
 #### AUTHORIZATION TOKEN {#authorization-token}
 
-The AUTHORIZATION TOKEN parameter (Parameter Type 0x01) identifies a track's
-authorization information in a SUBSCRIBE, SUBSCRIBE_ANNOUNCES, ANNOUNCE
-TRACK_STATUS_REQUEST or FETCH message. This parameter is populated for
-cases where the authorization is required at the track or namespace level.
+The AUTHORIZATION TOKEN parameter (Parameter Type 0x03) MAY appeat in a
```
suhasHere, line 1349:
```suggestion
The AUTHORIZATION TOKEN parameter (Parameter Type 0x03) MAY appear in a
```

draft-ietf-moq-transport.md
```diff
@@ -1333,26 +1346,27 @@ these parameters to appear in Setup messages.
...
+CLIENT_SETUP, SUBSCRIBE, SUBSCRIBE_ANNOUNCES, ANNOUNCE TRACK_STATUS_REQUEST or
```
suhasHere, line 1350:
```suggestion
CLIENT_SETUP, SUBSCRIBE, SUBSCRIBE_ANNOUNCES, ANNOUNCE, TRACK_STATUS_REQUEST or
```

draft-ietf-moq-transport.md
```diff
@@ -1333,26 +1346,27 @@ these parameters to appear in Setup messages.
...
+FETCH message. This parameter converys information to authorize the sender to
```
suhasHere, line 1351:
```suggestion
FETCH message. This parameter conveys information to authorize the sender to
```

draft-ietf-moq-transport.md
```diff
@@ -1380,52 +1394,59 @@ TOKEN {
...
-not currently registered MUST reject the message with `Unknown Auth Token
-Alias`. The receiver of a message attempting to register an alias which is
-already registered MUST close the session with `Duplicate Auth Token Alias`.
-
-Any message carrying an AUTHORIZATION TOKEN with Alias Type REGISTER that does
-not result in `Malformed Auth Token` MUST effect the token registration, even
-if the message fails for other reasons, including `Unauthorized`.  This allows
-senders to pipeline messages that refer to previously registered tokens.
+If the Token structure cannot be decoded, the receiver MUST close the session
+with Key-Value Formatting error.  The receiver of a message attempting to
```
suhasHere, line 1411: where is this error defined ?

afrind: It's Session error 0x6 (https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#name-termination)

draft-ietf-moq-transport.md
```diff
@@ -1333,26 +1346,27 @@ these parameters to appear in Setup messages.
 
 #### AUTHORIZATION TOKEN {#authorization-token}
 
-The AUTHORIZATION TOKEN parameter (Parameter Type 0x01) identifies a track's
-authorization information in a SUBSCRIBE, SUBSCRIBE_ANNOUNCES, ANNOUNCE
-TRACK_STATUS_REQUEST or FETCH message. This parameter is populated for
-cases where the authorization is required at the track or namespace level.
+The AUTHORIZATION TOKEN parameter (Parameter Type 0x03) MAY appear in a
+CLIENT_SETUP, SUBSCRIBE, SUBSCRIBE_ANNOUNCES, ANNOUNCE, TRACK_STATUS_REQUEST or
```
ianswett, line 1350: Can it not appear in SERVER_SETUP?  If I connect to a server to Upload video, I would want some sort of assurance it was the server I intended. Traditional TLS SNI is likely sufficient, but might there be use for further auth?

draft-ietf-moq-transport.md
```diff
@@ -1367,7 +1381,7 @@ TOKEN {
 |------|------------|------------------------------------------------------|
 | 0x1  | REGISTER   | There is an Alias, a Type and a Value. This Alias    |
 |      |            | MUST be associated with the Token Value for the      |
-|      |            | duration of the Session or it is deleted. This action|
+|      |            | duration of the session or it is deleted. This action|
```
ianswett, line 1384: nit: Why wouldn't Session be capitalized?

draft-ietf-moq-transport.md
```diff
@@ -1380,52 +1394,59 @@ TOKEN {
...
-Clients SHOULD only register tokens which they intend to re-use during the session.
-Client SHOULD retire previously registered tokens once their utility has passed.
+Senders of tokens SHOULD only register tokens which they intend to re-use during
+the session and SHOULD retire previously registered tokens once their utility
+has passed.
 
-By registering a Token, the client is requiring the receiver to store the Token
-Alias and Token Value until they are retired, or the Session ends. The receiver
+By registering a Token, the sender is requiring the receiver to store the Token
+Alias and Token Value until they are delete, or the session ends. The receiver
```
ianswett, line 1445:
```suggestion
Alias and Token Value until they are deleted, or the session ends. The receiver
```

ianswett (commented):
Some small comments, but mostly LG

draft-ietf-moq-transport.md
```diff
@@ -1380,52 +1394,59 @@ TOKEN {
...
+The receiver of a message containing a well-formed Token structure but otherwise
+invalid AUTHORIZATION TOKEN parameter MUST reject that message with an
+`Malformed Auth Token` error.
+
+The receiver of a message carrying an AUTHORIZATION TOKEN with Alias Type
+REGISTER that does not result in a session error MUST register the Token Alias,
+Type and Value in the token cache, even if the message fails for other reasons,
+including `Unauthorized`.  This allows senders to pipeline messages that refer
+to previously registered tokens without potentially terminating the entire
+session.
```
suhasHere, line 1426: I am not sure if we want to register an Unauthorized token.

suhasHere (approved):
One clarification question, otherwise LGTM

draft-ietf-moq-transport.md
ianswett: Similar to an expired token, I think the idea is you'd store that there was an alias, but you wouldn't need to store the token itself since it's useless?

draft-ietf-moq-transport.md
afrind: Oh I was trying to improve consistency and it's more lower than upper case right now.  Do we like Session better?

draft-ietf-moq-transport.md
afrind: > I am not sure if we want to register an Unauthorized token.

> you'd store that there was an alias, but you wouldn't need to store the token itself since it's useless

Yes this is fine.  If I send you an invalid token with alias=7, you can just mark alias=7 results in Unauthorized.

Let me put it a different way -- if the message carrying the register does not result in a session error, then subsequent usage of the alias MUST NOT result in a _session_ error either.  Would updating the explanation make you more comfortable @suhasHere?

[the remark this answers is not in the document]
afrind: Yes. In that case storing the value could be optimized away as well.

draft-ietf-moq-transport.md
```diff
@@ -1560,13 +1581,22 @@ value is 0, so if not specified, the peer MUST NOT send requests.
...
+sizes for all deregistered tokens (Alias Type value of 0x00), since Session
+initiation.
+
+#### AUTHORIZATION TOKEN {#setup-auth-token}
+
+See {{authorization-token}}.  The client can specify one or more tokens in
+CLIENT_SETUP that the server can use to authorize the client to establish an
+MOQT session.  If the server receives a MAX_AUTH_TOKEN_CACHE_SIZE parameter
+after an AUTHORIZATION_TOKEN in CLIENT_SETUP, it MUST close the session with a
+Protocol Violation.
```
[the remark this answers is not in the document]
afrind: Derp you are right.  I confused myself thinking that the MAX_TOKEN_CACHE_SIZE param controlled the cache for tokens registered by AUTH_TOKEN params.

An unfortunate side-effect is that because the default max size is 0, the client can't register any tokens in SETUP (but the server can, which makes less sense).  Since we think optionality of this feature is important for servers, the only fix may be to remember params for 0-RTT resume like HTTP?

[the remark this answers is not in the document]
afrind: Not being able to register a token from CLIENT_SETUP means in cases where I got a single token that authorizes both connecting and say, subscribe, I always have to send it twice :(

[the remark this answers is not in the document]
martinduke: I'm not suggesting a protocol change. I'm wondering if any real publisher would expect the same token for both session and subscription establishment.

I honestly don't know the answer.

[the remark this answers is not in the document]
afrind: > I would expect a single composite token being used for connection and actions

@wilaw Given that, what would you like to see here?  Ideally the client could use REGISTER_TOKEN in CLIENT_SETUP, and USE_ALIAS in subsequent requests.  However, the default toke cache size is 0, so the best you can do now is sending USE_TOKEN in CLIENT_SETUP and then REGISTER_TOKEN in the first control message after SERVER_SETUP.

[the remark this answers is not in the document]
wilaw: @afrind - good point. It would be ideal if the alias could be registered on the very first CLIENT-SETUP and then for all other actions you could just pass the alias back. To enable this, I'd like to see the default token cache size to be set to some reasonable non-zero level, to hold at least one composite token. 5MB?

draft-ietf-moq-transport.md
afrind: s/session/Session/

draft-ietf-moq-transport.md
afrind: Added text indicating that the cache can store Token Alias + Error in cases when all future uses of the token would result in that error.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
afrind: Removed the nonsense about param ordering and added special handling of CLIENT_SETUP for optimistic registration.

draft-ietf-moq-transport.md
```diff
@@ -1648,11 +1650,16 @@ initiation.
...
-after an AUTHORIZATION_TOKEN in CLIENT_SETUP, it MUST close the session with a
-Protocol Violation.
+See {{authorization-token}}.  The endpoint can specify one or more tokens in
+CLIENT_SETUP or SERVER_SETUP that the peer can use to authorize 
+MOQT session establishment.
+
+If a server receives an AUTHORIZATION TOKEN parmeter in CLIENT_SETUP with Alias
+Type REGISTER_TOKEN that exceeds its MAX_AUTH_TOKEN_CACHE_SIZE, it MUST NOT fail
+the session with `Auth Token Cache Overflow`.  Instead, it MUST treat the
+parameter as Alias Type USE_VALUE.  A client MUST handle registration failures
```
suhasHere, line 1661: I don't want to block this. But I still don't see a value or the idea behind using an alias if the cache limit is already exceeded. May be I am missing something here.

afrind: The scenario is that there is a single token for session establishment and all allowed actions.  It would be ideal to transmit in CLIENT_SETUP and reference the Alias thereafter.  Without some special treatment however, the client would never be able to REGISTER in CLIENT_SETUP, because the default cache size is 0.

The idea is: allow the clientto REGISTER blind in CLIENT_SETUP, but the server doesn't treat it like an error if the token is too large or it doesn't support token caching.  The server is _not_ obligated to cache tokens larger than its max.

suhasHere (approved):

vasilvv (approved):

draft-ietf-moq-transport.md
```diff
@@ -1431,49 +1445,58 @@ TOKEN {
 |------|------------|------------------------------------------------------|
 
 
-* Token Alias - a session-specific integer identifier that references a Token
-  Value. The Token Alias MUST be unique within the Session. Once a Token Alias has
-  been registered, it cannot be re-registered within the Session without first
-  being deleted. Use of the Token Alias is optional.
+* Token Alias - a Session-specific integer identifier that references a Token
+  Value. There are separate Alias spaces for the client and server (eg: they can
```
martinduke: nit: e.g.

draft-ietf-moq-transport.md
```diff
@@ -1616,13 +1639,27 @@ value is 0, so if not specified, the peer MUST NOT send requests.
...
+sizes for all deregistered tokens (Alias Type value of 0x00), since Session
+initiation.
+
+#### AUTHORIZATION TOKEN {#setup-auth-token}
+
+See {{authorization-token}}.  The endpoint can specify one or more tokens in
+CLIENT_SETUP or SERVER_SETUP that the peer can use to authorize MOQT session
+establishment.
+
+If a server receives an AUTHORIZATION TOKEN parmeter in CLIENT_SETUP with Alias
```
martinduke: nit: parameter

draft-ietf-moq-transport.md
```diff
@@ -1431,49 +1445,58 @@ TOKEN {
...
 
-* Token Alias - a session-specific integer identifier that references a Token
-  Value. The Token Alias MUST be unique within the Session. Once a Token Alias has
-  been registered, it cannot be re-registered within the Session without first
-  being deleted. Use of the Token Alias is optional.
+* Token Alias - a Session-specific integer identifier that references a Token
+  Value. There are separate Alias spaces for the client and server (eg: they can
+  each register Alias=1). Once a Token Alias has been registered, it cannot be
+  re-registered by the same sender in the Session without first being
+  deleted. Use of the Token Alias is optional.
```
```suggestion
* Token Alias - a Session-specific integer identifier that references a Token
  Value. There are separate Alias spaces for the client and server (e.g.: they
  can each register Alias=1). Once a Token Alias has been registered, it cannot
  be re-registered by the same sender in the Session without first being
  deleted. Use of the Token Alias is optional.
```

draft-ietf-moq-transport.md
```diff
@@ -1616,13 +1639,27 @@ value is 0, so if not specified, the peer MUST NOT send requests.
...
+initiation.
+
+#### AUTHORIZATION TOKEN {#setup-auth-token}
+
+See {{authorization-token}}.  The endpoint can specify one or more tokens in
+CLIENT_SETUP or SERVER_SETUP that the peer can use to authorize MOQT session
+establishment.
+
+If a server receives an AUTHORIZATION TOKEN parmeter in CLIENT_SETUP with Alias
+Type REGISTER_TOKEN that exceeds its MAX_AUTH_TOKEN_CACHE_SIZE, it MUST NOT fail
```
```suggestion
If a server receives an AUTHORIZATION TOKEN parameter in CLIENT_SETUP with Alias
Type REGISTER_TOKEN that exceeds its MAX_AUTH_TOKEN_CACHE_SIZE, it MUST NOT fail
```

draft-ietf-moq-transport.md
```diff
@@ -1431,49 +1445,58 @@ TOKEN {
...
+`Malformed Auth Token` error.
+
+The receiver of a message carrying an AUTHORIZATION TOKEN with Alias Type
+REGISTER that does not result in a Session error MUST register the Token Alias,
+in the token cache, even if the message fails for other reasons, including
+`Unauthorized`.  This allows senders to pipeline messages that refer to
+previously registered tokens without potentially terminating the entire Session.
+A receiver MAY store an error code (eg: Unauthorized or Malformed Auth Token) in
+place of the Token Type and Token Alias if any future message referencing the
+Token Alias will result in that error.
```
```suggestion
Token Alias will result in that error. The size of a registered cache entry
includes the length of the Token Value, regardless of whether it is stored.
```

PR #1014, opened 2025-05-08
Clarify Too Far Behind and DELIVERY TIMEOUT

afrind:
DELIVERY TIMEOUT is not a guarantee

Fixes: #715

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1439,18 +1439,20 @@ timeout is based on when the beginning of the Object is received, and does
 not depend upon the forwarding preference. There is no explicit signal that
 an Object was not sent because the delivery timeout was exceeded.
 
-If both the subscriber and publisher specify the parameter, they use the min of the
-two values for the subscription.  The publisher SHOULD always specify the value
-received from an upstream subscription when there is one, and nothing otherwise.
-If an earlier Object arrives later than subsequent Objects, relays can consider
-the receipt time as that of the next later Object, with the assumption that the
-Object's data was reordered.
+If both the subscriber and publisher specify the parameter, they use the min of
+the two values for the subscription.  The publisher SHOULD always specify the
+value received from an upstream subscription when there is one, and nothing
+otherwise.  If an earlier Object arrives later than subsequent Objects, relays
+can consider the receipt time as that of the next later Object, with the
+assumption that the Object's data was reordered.
 
-If neither the subscriber or publisher specify DELIVERY TIMEOUT, all Objects
-in the track matching the subscription filter are delivered as indicated by
-their Group Order and Priority.  If a subscriber exceeds the publisher's
-resource limits by failing to consume objects at a sufficient rate, the
-publisher MAY terminate the subscription with error 'Too Far Behind'.
+Publishers can, at their discretion, discontinue forwarding Objects earlier than
+the negotiated DELIVERY TIMEOUT.  However, if neither the subscriber or
+publisher specify DELIVERY TIMEOUT, all Objects in the track matching the
+subscription filter are delivered as indicated by their Group Order and
+Priority.  If a subscriber fails to consume Objects at a sufficient rate,
+causing the publisher to exceed its resource limits, the publisher MAY terminate
+the subscription with error 'Too Far Behind'.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1439,18 +1439,20 @@ timeout is based on when the beginning of the Object is received, and does
...
-publisher MAY terminate the subscription with error 'Too Far Behind'.
+If both the subscriber and publisher specify the parameter, they use the min of
+the two values for the subscription.  The publisher SHOULD always specify the
+value received from an upstream subscription when there is one, and nothing
+otherwise.  If an earlier Object arrives later than subsequent Objects, relays
+can consider the receipt time as that of the next later Object, with the
+assumption that the Object's data was reordered.
+
+Publishers can, at their discretion, discontinue forwarding Objects earlier than
+the negotiated DELIVERY TIMEOUT.  However, if neither the subscriber or
```
suhasHere, line 1450: I see the intent here, but wondering if we can add more information explaining the publisher's discretion statement?

suhasHere (approved):
Nit , otherwise looks fine to me

draft-ietf-moq-transport.md
afrind: @suhasHere do you have a specific suggestion?  I suppose the primary case in mind is resource constraints but that is just one example.

afrind:
@vasilvv is this sufficiently addressing your concern?

draft-ietf-moq-transport.md
fluffy: They can't just randomly drop stuff, they have to close the steam or signal they are doing that. Think of the example where a relay send object 1,2, and 4 on a fetch stream and just decided to drop 3.  I think this is the wrong place to talk about how relays deal with overload.

fluffy (changes requested):
I don't think this is correct. It is also really hard to figure out what actually changed as the diff seems huge.

draft-ietf-moq-transport.md
```diff
@@ -1454,11 +1454,13 @@ otherwise.  If an Object with a smaller ID arrives later than subsequent
...
-their Group Order and Priority.  If a subscriber exceeds the publisher's
-resource limits by failing to consume objects at a sufficient rate, the
-publisher MAY terminate the subscription with error 'Too Far Behind'.
+Publishers can, at their discretion, discontinue forwarding Objects earlier than
+the negotiated DELIVERY TIMEOUT.  However, if neither the subscriber or
+publisher specify DELIVERY TIMEOUT, all Objects in the track matching the
+subscription filter are delivered as indicated by their Group Order and
+Priority.  If a subscriber fails to consume Objects at a sufficient rate,
+causing the publisher to exceed its resource limits, the publisher MAY terminate
+the subscription with error 'Too Far Behind'.
```
afrind, line 1463:
```suggestion
Publishers can, at their discretion, discontinue forwarding Objects earlier than
the negotiated DELIVERY TIMEOUT, subject to stream closure and ordering
constraints described in {{closing-subgroup-streams}}.  However, if neither the
subscriber or publisher specify DELIVERY TIMEOUT, all Objects in the track
matching the subscription filter are delivered as indicated by their Group Order
and Priority.  If a subscriber fails to consume Objects at a sufficient rate,
causing the publisher to exceed its resource limits, the publisher MAY terminate
the subscription with error 'Too Far Behind'.
```

draft-ietf-moq-transport.md
afrind: @fluffy you are correct that random object drops are not allowed.  I added a note and linked to the section describing those rules.

>  I think this is the wrong place to talk about how relays deal with overload.

This is describing what delivery timeout is and how to process subscriptions with delivery timeouts, so it seems relevant here, at least for now.

afrind:
> I don't think this is correct. It is also really hard to figure out what actually changed as the diff seems huge.

The original PR included a drive-by reflow of the previous paragraph, but that got handled in a different PR, which has merged, so this is smaller now.  Only the modified paragraph has reflowed, but I only added the first sentence, and the word "However".

fluffy (approved):

fluffy:
thanks - that makes more sense now

ianswett (approved):

Issue #1002, opened 2025-05-06
Naming nit in Track status message


fluffy:
This got changed in -11

I think that "Largest Location (Location)" should be called "Largest Object (Location)"

PR #1015, opened 2025-05-08
Make TRACK_STATUS mirror SUBSCRIBE

afrind:
I added a parameter (TRACK_STATUS_RESPONDER) to allow for cases where a relay has cached information about a track and doesn't want to or can't establish a subscription or send TRACK_STATUS upstream.

Fixes: #895
Fixes: #1002

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2509,12 +2509,12 @@ The TRACK_STATUS message format is identical to the SUBSCRIBE message
 ({{message-subscribe-req}}).
 
 The receiver of a TRACK_STATUS message treats it identically as if it had
-received a SUBSCRIBE message, except it does not create subscription state or
-send any Objects.  Relays without an active subscription MAY forward
-TRACK_STATUS to one or more publishers, or MAY initiate a subscription as
-described in {{publisher-interactions}} to determine the response. The publisher
-does not send SUBSCRIBE_DONE for this request, and the subscriber cannot send
-SUBSCRIBE_UPDATE or UNSUBSCRIBE.
+received a SUBSCRIBE message, except it does not create downstream subscription
+state or send any Objects.  Relays without an active subscription MAY forward
+TRACK_STATUS to one or more publishers, or MAY initiate a subscription (subject
+to authorization) as described in {{publisher-interactions}} to determine the
+response. The publisher does not send SUBSCRIBE_DONE for this request, and the
+subscriber cannot send SUBSCRIBE_UPDATE or UNSUBSCRIBE.
 
 ## TRACK_STATUS_OK {#message-track-status-ok}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1256,9 +1256,11 @@ The following Message Types are defined:
 |-------|-----------------------------------------------------|
 | 0x17  | FETCH_CANCEL ({{message-fetch-cancel}})             |
 |-------|-----------------------------------------------------|
-| 0xD   | TRACK_STATUS_REQUEST ({{message-track-status-req}}) |
+| 0xD   | TRACK_STATUS ({{message-track-status}})             |
 |-------|-----------------------------------------------------|
-| 0xE   | TRACK_STATUS ({{message-track-status}})             |
+| 0xE   | TRACK_STATUS_OK ({{message-track-status-ok}}        |
+|-------|-----------------------------------------------------|
+| 0xF   | TRACK_STATUS_ERROR ({{message-track-status-error}}) |
 |-------|-----------------------------------------------------|
 | 0x6   | ANNOUNCE  ({{message-announce}})                    |
 |-------|-----------------------------------------------------|
@@ -1335,8 +1337,8 @@ these parameters to appear in Setup messages.
 
 The AUTHORIZATION TOKEN parameter (Parameter Type 0x01) identifies a track's
 authorization information in a SUBSCRIBE, SUBSCRIBE_ANNOUNCES, ANNOUNCE
-TRACK_STATUS_REQUEST or FETCH message. This parameter is populated for
-cases where the authorization is required at the track or namespace level.
+TRACK_STATUS or FETCH message. This parameter is populated for cases where the
+authorization is required at the track or namespace level.
 
 The AUTHORIZATION TOKEN parameter MAY be repeated within a message.
 
@@ -1475,6 +1477,18 @@ earlier in a multi-object stream will expire earlier than Objects later in the
 stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
+
+#### TRACK STATUS RESPONDER
+
+The TRACK_STATUS_RESPONDER parameter (Parameter Type 0x6) indicates the entity
+that sent a TRACK_STATUS_OK or TRACK_STATUS_ERROR.  The allowed values are:
+
+0x0: Original Publisher
+0x1: Relay with Active Subscription
+0x2: Relay without Active Subscription
+
+If a subscriber receives any other value for this header it MUST close the
+session with error `Protocol Violation`.
 
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
 
@@ -1625,9 +1639,9 @@ MAX_REQUEST_ID Message {
 
 * Request ID: The new Maximum Request ID for the session. If a Request ID equal
   or larger than this is received by the endpoint that sent the MAX_REQUEST_ID
-  in any request message (ANNOUNCE, FETCH, SUBSCRIBE, SUBSCRIBE_ANNOUNCES
-  or TRACK_STATUS_REQUEST), the endpoint MUST close the session with an error
-  of 'Too Many Requests'.
+  in any request message (ANNOUNCE, FETCH, SUBSCRIBE, SUBSCRIBE_ANNOUNCES or
+  TRACK_STATUS), the endpoint MUST close the session with an error of 'Too Many
+  Requests'.
 
 MAX_REQUEST_ID is similar to MAX_STREAMS in ({{?RFC9000, Section 4.6}}), and
 similar considerations apply when deciding how often to send MAX_REQUEST_ID.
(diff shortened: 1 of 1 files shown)


sharmafb:
Do we want the TRACK_STATUS_RESPONDER parameter to be optional, or to be populated in all cases? If it's the latter (which is what the spec makes it sound like), should we have this be a part of the format of the TRACK_STATUS_OK and TRACK_STATUS_ERROR messages, rather than having it as a parameter that may or may not be present?

afrind:
> Do we want the TRACK_STATUS_RESPONDER parameter to be optional, or to be populated in all cases? If it's the latter (which is what the spec makes it sound like), should we have this be a part of the format of the TRACK_STATUS_OK and TRACK_STATUS_ERROR messages, rather than having it as a parameter that may or may not be present?

Good question - I think there's a lot of value in keeping the responses identical, so I'd prefer we just say that if the param is not present, it defaults to Original Publisher.

ianswett (commented):
LG, though I'm unclear what the purpose of the RESPONDER param is.

draft-ietf-moq-transport.md
```diff
@@ -1476,6 +1478,18 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+#### TRACK STATUS RESPONDER
```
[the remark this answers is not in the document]
afrind: I was trying to preserve the "status=0x4" case in the previous text -- eg: 

1. The track was live, relay cached some of it, publisher disconnected
2. Relay receives TRACK_STATUS

The relay doesn't want to or cannot forward the request upstream, but can give some useful info.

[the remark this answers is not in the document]
suhasHere: I feel this can be a separate PR or I don't feel this is really needed either ?

ianswett (approved):
I don't have a strong preference on the RESPONDER param, but it feels more like an error case that should be symmetric with SUBSCRIBE?

A new error code to put into TRACK_STATUS_ERROR (or SUBSCRIBE) might make more sense?

ianswett (commented):
I think a new error code for the _ERROR messages might be a better approach if we want to indicate "No Upstream", "Unroutable", etc

suhasHere (changes requested):
Should we revisit track responder idea in a different PR or do we need it at all ??

afrind:
> Should we revisit track responder idea in a different PR or do we need it at all ??

@suhasHere can you reply to my comments [here](https://github.com/moq-wg/moq-transport/pull/1015#issuecomment-2877760360) regarding the use case for the functionality you are proposing we remove?

fluffy (changes requested):

fluffy:
I see very little value of unifying subscribe ok and track status and I see a few strong downsides.  The main one is need to track alias. This forces a client that scans a ton of track info to pick a track to burn up a ton of small track alias resulting in more messages. There are lots of other fields that don't seem to make sense and I suspect over time we want to extent track alias differently than subscribe.

draft-ietf-moq-transport.md
```diff
@@ -2433,87 +2447,43 @@ FETCH_CANCEL Message {
...
 
-* Request ID: The Request ID of the TRACK_STATUS_REQUEST this message is
-  replying to {{message-track-status}}.
+The TRACK_STATUS message format is identical to the SUBSCRIBE message
+({{message-subscribe-req}}).
 
-* Status Code: Provides additional information about the status of the
-track. It MUST hold one of the following values. Any other value is a malformed
-message.
+The receiver of a TRACK_STATUS message treats it identically as if it had
```
martinduke, line 2458: As written, I believe this implies the receiver, if a relay, would create an upstream subscription?

draft-ietf-moq-transport.md
```diff
@@ -2433,87 +2447,43 @@ FETCH_CANCEL Message {
...
-  Parameters (..) ...,
-}
-~~~
-{: #moq-track-status-format title="MOQT TRACK_STATUS Message"}
+A potential subscriber sends a 'TRACK_STATUS' message on the control
+stream to obtain information about the current status of a given track.
 
-* Request ID: The Request ID of the TRACK_STATUS_REQUEST this message is
-  replying to {{message-track-status}}.
+The TRACK_STATUS message format is identical to the SUBSCRIBE message
```
martinduke, line 2455: I understand the desire for simplicity, but what is the purpose of the priority, group order, and forward fields?

fluffy:
On the call today people said the track alias could be zero but I'm not seeing that here

draft-ietf-moq-transport.md
martinduke: OK, this offends my bit-optimizing sensibilities, but because it simplifies my implementation I can live with it.

draft-ietf-moq-transport.md
```diff
@@ -2498,87 +2500,45 @@ FETCH_CANCEL Message {
...
+({{message-subscribe-req}}).
 
-* Status Code: Provides additional information about the status of the
-track. It MUST hold one of the following values. Any other value is a malformed
-message.
+The receiver of a TRACK_STATUS message treats it identically as if it had
+received a SUBSCRIBE message, except it does not create subscription state or
+send any Objects.  Relays without an active subscription MAY forward
+TRACK_STATUS to one or more publishers, or MAY initiate a subscription as
+described in {{publisher-interactions}} to determine the response. The publisher
```
suhasHere: Where is the section {{publisher-interactions}} defined ?

draft-ietf-moq-transport.md
```diff
@@ -2498,87 +2500,45 @@ FETCH_CANCEL Message {
...
+The TRACK_STATUS message format is identical to the SUBSCRIBE message
+({{message-subscribe-req}}).
 
-* Status Code: Provides additional information about the status of the
-track. It MUST hold one of the following values. Any other value is a malformed
-message.
+The receiver of a TRACK_STATUS message treats it identically as if it had
+received a SUBSCRIBE message, except it does not create subscription state or
+send any Objects.  Relays without an active subscription MAY forward
+TRACK_STATUS to one or more publishers, or MAY initiate a subscription as
```
suhasHere: I wonder if a relay should be allowed to initiate a subscription on behalf of the subscriber. This assumes the authz scopes for TRACK_STATUS and SUBSCRIBE cover exactly the same claims, which may or may not be true.

afrind: The relay's ability to establish an upstream subscription is always subject to authentication (mentioned elsewhere).  Still, I think we want to permit the relay to subscribe, otherwise the upstream could get DoS'd with TRACK_STATUS. This could be for cases where the client -> relay auth is separate from relay -> publisher auth or other out-of-band auth schemes.

I can add something like:

> MAY initiate a subscription **(subject to authorization)** as described in Publisher Interactions.

draft-ietf-moq-transport.md
afrind: It's in the relay section (https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#name-publisher-interactions)

suhasHere: ah ! I see we are missing markdown section marking for the same.

draft-ietf-moq-transport.md
```diff
@@ -2498,87 +2500,45 @@ FETCH_CANCEL Message {
...
 
-0x01: The track does not exist. Subsequent fields MUST be zero, and any other
-value is a malformed message.
+The publisher sends a TRACK_STATUS_OK control message in response
+to a successful TRACK_STATUS message.
 
-0x02: The track has not yet begun. Subsequent fields MUST be zero. Any other
-value is a malformed message.
+The TRACK_STATUS_OK message format is identical to the SUBSCRIBE_OK message
+({{message-subscribe-ok}}).
```
suhasHere, line 2752: Shouldn't we add an explicit requirement on track_alias here and what is expected for track status responses.

draft-ietf-moq-transport.md
suhasHere: As much as this feels ok, I do feel this will turn into Subscription on every implementation and having to set the right track aliases otherwise, isn't it ?
I see the point on DOS, however I wonder if it gets handled differently ?

afrind: The publisher chooses the Track Alias now so I'm not sure how having the relay issue a SUBSCRIBE in response to TRACK_STATUS has any impact.  Did I miss something?

draft-ietf-moq-transport.md
afrind: There's an implicit anchor for "Section Title" {{section-title}}.  You only need a specific anchor if you want a different/shorter one.

draft-ietf-moq-transport.md
afrind: It's in the next sentence: "setting Track Alias to 0"

draft-ietf-moq-transport.md
suhasHere: There seems to be a conflict here between these 2 statements on Subscription State creation

1. "except it does not create subscription state or send any Objects"

2. "Relays without an active subscription MAY forward
TRACK_STATUS to one or more publishers, or MAY initiate a subscription "

In most implementations, initiating a subscription creates subscription state.  May be we should just say the "Relays without an active subscription MAY forward TRACK_STATUS to one or more publishers "

suhasHere (approved):
Thanks @afrind  . The clarification looks fine.

draft-ietf-moq-transport.md
```diff
@@ -2498,87 +2500,45 @@ FETCH_CANCEL Message {
...
 
-0x03: The track has finished, so there is no "live edge." Subsequent fields
-contain the highest Group and object ID known.
+The publisher populates the fields of TRACK_STATUS_OK exactly as it would have
+populated a SUBSCRIBE_OK, setting Track Alias to 0. It is not considered an error
```
martinduke, line 2755: What goes in the Group Order field?

draft-ietf-moq-transport.md
```diff
@@ -2498,87 +2500,45 @@ FETCH_CANCEL Message {
...
-  Parameters (..) ...,
-}
-~~~
-{: #moq-track-status-format title="MOQT TRACK_STATUS Message"}
+A potential subscriber sends a 'TRACK_STATUS' message on the control
+stream to obtain information about the current status of a given track.
 
-* Request ID: The Request ID of the TRACK_STATUS_REQUEST this message is
-  replying to {{message-track-status}}.
+The TRACK_STATUS message format is identical to the SUBSCRIBE message
```
martinduke, line 2735: What about the Group Order, Forward, and Filter Type fields? ISTM any filter type that has explicit Locations is just a waste of bits.

afrind: The subscriber chooses whatever it wants in those fields, and the response will take that into account. For example, Group Order=Default in the request means the response will show the publisher's preference.  Filter may be uninteresting, but it may be a parameter soon?

draft-ietf-moq-transport.md
afrind: The same thing it would have put in SUBSCRIBE_OK?

vasilvv (approved):

afrind:
I believe @fluffy's -1 came from an earlier version of this PR where it was unclear what was happening with Track Alias.  Hopefully now this is clear.

Issue #549, opened 2024-09-25
Better compression of VarInts


fluffy:
A one byte VarInt can only 64 values but most variable length encodings could express twice that many values. 

The bandwidth of MoQ could be reduced, helping reduce global warming, by a better VarInt design for MoQ.

kixelated:
QUIC varints use 2 bits in the first byte to express if the number is 1, 2, 4, or 8 bytes long.

Protobuf varints use 1 bit *per byte* to express if the number continues to the next byte. This is probably similar to what you're referring to.

QUIC is usually more efficient when encoding numbers greater than 2^14. Protobuf is only more efficient encoding numbers between 2^6 and 2^7.

Regardless, you already need a QUIC library for MoQ and it often comes with a VarInt encoder/decoder for free.

LPardue:
So if I understand the problem statement, the issue is that reserving the 2 MSBs means that a value of 64-127 needs 2 bytes rather than 1 (and values 16384-32767 need 3 bytes rather than 2 etc). 

Is there any idea how much this would affect MOQT traffic? If its isolated to a few message types, rather than defining another new primitive variable-length integer, we could consider structured variable-length messages. I.e. a VAR_FOO message would have two (or more) variants that support different length thresholds

8-bit version
```
VAR_FOO Message {
Type (i) = 0XF00,
Length (i),
Field that is commonly 0..255 (8)
}
```

16-bit version
```
VAR_FOO Message {
Type (i) = 0XF01,
Length (i),
Field that is commonly 256..65535 (16)
}
```

fluffy:
If you only had one varint in the messages, yah, that would work. But does not really scale very well as you get more.

LPardue:
Agree it doesn't scale well. I don't understand the concrete cases where 1 byte per field really matters that much. What sorts of MoQ things are significantly affected?

ianswett:
Given we get to choose what types fit into a single byte, and we have a number of things like Object length that will typically need 2+ bytes, I'm not sure how many cases there are where the current approach is suboptimal?

fluffy:
I think these need to be discussed now and pushing it to later is just trying to veto the issue by pushing it to a point that is too late to do.

mzanaty:
For control messages, I would argue aligning with QUIC varint is best even if suboptimal.

[...]

Issue #548, opened 2024-09-25
Support 64 bit Var Int values 


fluffy:
To facilitate mapping object ID to existing 64 bit is not possible.

LPardue:
What shape of solution is acceptable here? QUIC varints are limited to 62 bits (64  minus 2 reserved for the length indicator). Is the suggestion to have 125 bits (128 minus 3) and require implementations to handle this as 128 bits?

Or maybe just "extend" the varint by add a new MSB field to messages that require full range of 64 bits e.g

```
MSB (2)
Unused (6)
BigVal (i)
```

Then a parser, which already represents varints as u64, can just combine the two fields to create the actual value?

fluffy:
That's a bit ugly but I think it meets the requirements.

kixelated:
...are you using a u64 timestamp as an object ID or something? And those high two bits are mandatory to support or is this a theoretical overflow in a billion years sort of thing?

But I do agree that encoder overflows above 2^62 are annoying to deal with. However, a VarInt that can represent larger values is equally annoying because of decoder overflows. Only a VarInt encoding that represents exactly 64 bits would be an improvement IMO but the math doesn't quite work out.

But Cullen it continues to feel like you're abusing Object ID with such huge numbers and gaps. Can you put this u64 inside of the payload instead of the MoqTransport header?

afrind:
We discussed in the meeting today and agreed this needs motivating use cases

fluffy:
I think these need to be discussed now and pushing it to later is just trying to veto the issue by pushing it to a point that is too late to do.

afrind:
> this needs motivating use cases

@fluffy Can you add these to kick off discussion?  I assume some existing system has 64 bit identifiers and it's unclear if it's safe to cram them into, say, moqt group ID which might truncate them to 62 bits?

Here's my QPACK style proposal

1 byte = 0 + 7 bits [128]
2 bytes = 10 + 14 bits [16kb]
4 bytes = 110 + 29 bits [524MB]
8 bytes = 1110 + 60 [huge]
9 bytes = 11110000 + 64 [full 64 bits]

gloinul:
[...]

PR #1016, opened 2025-05-08
New Varint Encoding

afrind:
Fixes: #548
Fixes: #549

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -243,11 +243,6 @@ x (L):
 
 : Indicates that x is L bits long
 
-x (i):
-
-: Indicates that x holds an integer value using the variable-length
-  encoding as described in ({{?RFC9000, Section 16}})
-
 x (..):
 
 : Indicates that x can be any length including zero bits long.  Values
@@ -261,6 +256,13 @@ x (L) ...:
 
 : Indicates that x is repeated zero or more times and that each instance
   has a length of L
+
+This document redfines the following RFC9000 syntax:
+
+x (i):
+
+: Indicates that x holds an integer value using the variable-length
+  encoding as described in {{variable-length-integers}}.
 
 This document extends the RFC9000 syntax and with the additional field types:
 
@@ -276,9 +278,43 @@ x (tuple):
   as described in ({{?RFC9000, Section 16}}), followed by that many variable
   length tuple fields, each of which are encoded as (b) above.
 
+
+### Variable-Length Integers
+
+RFC9000 variable-length integers use a two-bit length prefix in the first byte
+to determine the encoded length.  This limits the number of encodings to 4,
+reduces the value range of 1-byte encodings to 0-63, and prevents encoding full
+64-bit numbers.  This encoding is not suitable for MoQT.
+
+The variable-length integer encoding reserves the most one to four significant
+bits of the first byte to encode the length of the integer encoding in
+bytes. The integer value is encoded on the remaining bits, in network byte
+order.
+
+Integers are encoded in 1, 2, 4, 8, or 9 bytes and can encode 7-, 14-, 29-, 60-,
+or 64-bit values, respectively. The following table summarizes the encoding
+properties.
+
+|--------------|----------------|-------------|---------------|
+| Leading Bits | Length (bytes) | Usable Bits | Range         |
+|--------------|----------------|-------------|---------------|
+| 0            | 1              | 7           | 0-127         |
+|--------------|----------------|-------------|---------------|
+| 10           | 2              | 14          | 0-16,383      |
+|--------------|----------------|-------------|---------------|
+| 110          | 4              | 29          | 0-536,870,911 |
+|--------------|----------------|-------------|---------------|
+| 1110         | 8              | 60          | 0-2^60-1      |
+|--------------|----------------|-------------|---------------|
+| 1111XXXX     | 9              | 64          | 0-2^64-1      |
+|--------------|----------------|-------------|---------------|
+{: format title="Summary of Integer Encodings"}
+
 To reduce unnecessary use of bandwidth, variable length integers SHOULD
 be encoded using the least number of bytes possible to represent the
 required value.
+
+A sample decoding algorithm is in {{sample-decoding}}.
 
 ### Location Structure
 
@@ -3442,6 +3478,31 @@ document:
 - Will Law
 
 --- back
+
+# Sample Variable-Length Integer Decoding {#sample-decoding}
+
+The function ReadVarint takes a single argument -- a sequence of bytes, which
+can be read in network byte order.
+
+~~~
+ReadVarint(data):
+  lengths = [ 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 4, 4, 8, 9 ]
+  masks = [ 0x7f, 0x7f, 0x7f, 0x7f, 0x7f, 0x7f, 0x7f, 0x7f,
+            0x3f, 0x3f, 0x3f, 0x3f, 0x1f, 0x1f, 0x0f, 0x00 ]
+  // The length of variable-length integers is encoded in the
+  // first one to four bits of the first byte.
+  v = data.next_byte()
+  prefix = v >> 4
+  length = lengths[prefix]
+  mask = masks[prefix]
+
+  // Once the length is known, remove these bits and read any
+  // remaining bytes.
+  v = v & mask
+  repeat length-1 times:
+    v = (v << 8) + data.next_byte()
+  return v
+~~~
 
 # Change Log
 


draft-ietf-moq-transport.md
```diff
@@ -276,10 +278,44 @@ x (tuple):
   as described in ({{?RFC9000, Section 16}}), followed by that many variable
   length tuple fields, each of which are encoded as (b) above.
 
+
+### Variable-Length Integers
+
+RFC9000 variable-length integers use a two-bit length prefix in the first byte
```
LPardue, line 284: I understand why you put this text here but it seems superfluous to talk about something that is not suitable for MoQT. Consider simply stating MoQT's requirements or abilities and then state how the encoding supports thst on the wire.

draft-ietf-moq-transport.md
```diff
@@ -3443,6 +3479,31 @@ document:
 
 --- back
 
+# Sample Variable-Length Integer Decoding {#sample-decoding}
+
+The function ReadVarint takes a single argument -- a sequence of bytes, which
```
LPardue, line 3484: The RFC editor gets twitchy about pseudo code. Since this emulates RFC 9000's section, I suggest more closely copying the prose a markdown behind https://datatracker.ietf.org/doc/html/rfc9000#section-a.1

draft-ietf-moq-transport.md
```diff
@@ -3443,6 +3479,31 @@ document:
...
+  mask = masks[prefix]
+
+  // Once the length is known, remove these bits and read any
+  // remaining bytes.
+  v = v & mask
+  repeat length-1 times:
+    v = (v << 8) + data.next_byte()
+  return v
+~~~
+
```
LPardue, line 3506: Rfc 9000 provides a concrete example, that would be nice here as a simple test vector

LPardue:
For a blanket change, I think you'll need to sanity check all uses. For instance, in [SUBSCRIBE_DONE](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#figure-14) the Stream Count field would allow expressing a larger stream count than the underlying transport can support . That's easily fixable by stating the logical limit of the field and what to do if you receive a bogus value, but it's an example of a new interop issue this change introduces.

draft-ietf-moq-transport.md
suhasHere: @afrind  It would also be nice to add a encode algorithm as well.

draft-ietf-moq-transport.md
```diff
@@ -276,10 +278,44 @@ x (tuple):
   as described in ({{?RFC9000, Section 16}}), followed by that many variable
   length tuple fields, each of which are encoded as (b) above.
 
+
+### Variable-Length Integers
+
+RFC9000 variable-length integers use a two-bit length prefix in the first byte
+to determine the encoded length.  This limits the number of encodings to 4,
+reduces the value range of 1-byte encodings to 0-63, and prevents encoding full
+64-bit numbers.  This encoding is not suitable for MoQT.
```
afrind, line 287:
```suggestion
MoQT requires a variable-length integer encoding with the following properties:

1. The encoded length can be determined from the first encoded byte.
2. The range of 1 byte values is as large as possible.
3. All 64 bit numbers can be encoded.
```

draft-ietf-moq-transport.md
```diff
@@ -3443,6 +3479,31 @@ document:
 
 --- back
 
+# Sample Variable-Length Integer Decoding {#sample-decoding}
+
+The function ReadVarint takes a single argument -- a sequence of bytes, which
+can be read in network byte order.
+
+~~~
```
afrind, line 3487:
```suggestion
~~~pseudocode
```

draft-ietf-moq-transport.md
afrind, line 3506:
```suggestion
For example, the nine-byte sequence 0xf0ffffffffffffffff decodes to the decimal
value 18,446,744,073,709,551,615; the eight-byte sequence 0xe2197c5eff14e88c
decodes to the decimal value 151,288,809,941,952,652; the four-byte sequence
0xdd7f3e7d decodes to 494,878,333; the two-byte sequence 0xbbbd decodes to
15,293; and the single byte 0x25 decodes to 37 (as does the two-byte sequence
0x8025).

```

draft-ietf-moq-transport.md
```diff
@@ -3443,6 +3479,31 @@ document:
 
 --- back
 
+# Sample Variable-Length Integer Decoding {#sample-decoding}
```
afrind, line 3482:
```suggestion
# Sample Variable-Length Integer Encoding and Decoding {#sample-varint}

The WriteVarint function takes two parameters, a 64 bit integer and an
output buffer with an append operation.

~~~pseudocode
Function WriteVarint(value, output):
    if value <= 127:
        output.append(value, 1)
    else if value <= 16383:
        output.append(networkByteOrder(uint16_t(value | 0x8000)), 2)
    else if value <= 536870911:
        output.append(networkByteOrder(uint32_t(value | 0xC0000000)), 4)
    else if value <= 1152921504606846975:
        output.append(networkByteOrder(uint64_t(value | 0xE000000000000000)), 8)
    else:
        output.append(0xF0, 1)
        output.append(networkByteOrder(value), 8)
~~~
        
```

draft-ietf-moq-transport.md
```diff
@@ -276,10 +278,44 @@ x (tuple):
...
+|--------------|----------------|-------------|---------------|
+| 1111XXXX     | 9              | 64          | 0-2^64-1      |
+|--------------|----------------|-------------|---------------|
+{: format title="Summary of Integer Encodings"}
+
 To reduce unnecessary use of bandwidth, variable length integers SHOULD
 be encoded using the least number of bytes possible to represent the
 required value.
 
+A sample decoding algorithm is in {{sample-decoding}}.
```
afrind, line 317:
```suggestion
A sample encoding and decoding algorithm is in {{sample-varint}}.
```

draft-ietf-moq-transport.md
```diff
@@ -276,10 +278,45 @@ x (tuple):
...
+MoQT requires a variable-length integer encoding with the following properties:
+
+1. The encoded length can be determined from the first encoded byte.
+2. The range of 1 byte values is as large as possible.
+3. All 64 bit numbers can be encoded.
+
+The variable-length integer encoding reserves the most one to four significant
+bits of the first byte to encode the length of the integer encoding in
+bytes. The integer value is encoded on the remaining bits, in network byte
+order.
```
suhasHere, line 293:
```suggestion
The variable-length integer encoding uses the most significant one to four 
bits of the first byte to indicate the length of the encoding in bytes. The remaining bits 
represent the integer value, encoded in network byte order.```

File: draft-ietf-moq-transport.md [hunk 9]
```diff
@@ -262,6 +257,13 @@ x (L) ...:
 : Indicates that x is repeated zero or more times and that each instance
   has a length of L
 
+This document redfines the following RFC9000 syntax:
+
+x (i):
```
vasilvv, line 262: Can we change `(i)` to avoid confusion with QUIC?

draft-ietf-moq-transport.md
```diff
@@ -276,10 +278,44 @@ x (tuple):
...
+|--------------|----------------|-------------|---------------|
+| 0            | 1              | 7           | 0-127         |
+|--------------|----------------|-------------|---------------|
+| 10           | 2              | 14          | 0-16,383      |
+|--------------|----------------|-------------|---------------|
+| 110          | 4              | 29          | 0-536,870,911 |
+|--------------|----------------|-------------|---------------|
+| 1110         | 8              | 60          | 0-2^60-1      |
+|--------------|----------------|-------------|---------------|
+| 1111XXXX     | 9              | 64          | 0-2^64-1      |
```
vasilvv, line 309: I have no idea what this actually means.  What are the `XXXX`?

Do we expect to revisit this scheme later?  If the goal is to make "the range of 1 byte values is as large as possible", it would make more sense to just allocate 0x00~0xf8 values for those, but then we basically don't have two-byte varints.

draft-ietf-moq-transport.md
```diff
@@ -276,10 +278,44 @@ x (tuple):
   as described in ({{?RFC9000, Section 16}}), followed by that many variable
   length tuple fields, each of which are encoded as (b) above.
 
+
+### Variable-Length Integers
```
ianswett, line 282: Can we move this earlier?  Because currently it's forward referenced.

draft-ietf-moq-transport.md
```diff
@@ -2060,18 +2096,18 @@ SUBSCRIBE_DONE Message {
...
-subscription than specified in Stream Count, it MAY close the session with a
-Protocol Violation.
+  opened for this subscription.  This helps the subscriber know if it has
+  received all of the data published in this subscription by comparing the
+  number of streams received.  The subscriber can immediately remove all
+  subscription state once the same number of streams have been processed.  If
+  the track had Forwarding Preference = Datagram, the publisher MUST set Stream
+  Count to 0.  If the publisher is unable to set Stream Count to the exact
+  number of streams opened for the subscription, it MUST set Stream Count to a
+  value greater than or equal to 2^60. Subscribers SHOULD use a timeout or other
```
ianswett, line 2106: Why 2^60?

ianswett (commented):
This LG, a few comments.  I think reusing 'i' from QUIC as a varint could be very confusing.

draft-ietf-moq-transport.md
afrind: > I have no idea what this actually means. What are the XXXX?

The 9 byte encoding uses 4 bits for length with 4 reserved bits.

> Do we expect to revisit this scheme later?

No.  If we make the switch let's go from QUIC to MoQ varints once and never look back.  Now is the time to propose new schemes.

> allocate 0x00~0xf8 values for those, but then we basically don't have two-byte varints.

To clarify, you if the first byte is > 0xf8 the length is `value - 0xf8 + 1`?  Technically you have 2 byte varints but they only encode 7 values that can't be encoded in 1 byte.  These things are all about tradeoffs and I'm agnostic if the group prefers 248 1-byte values, at the expense of more bytes for larger values.  It might be bad for shortish payload lengths (eg: 300-1200 bytes, though datagrams don't have a length field).

draft-ietf-moq-transport.md
```diff
@@ -276,10 +278,44 @@ x (tuple):
...
+|--------------|----------------|-------------|---------------|
+| 110          | 4              | 29          | 0-536,870,911 |
+|--------------|----------------|-------------|---------------|
+| 1110         | 8              | 60          | 0-2^60-1      |
+|--------------|----------------|-------------|---------------|
+| 1111XXXX     | 9              | 64          | 0-2^64-1      |
+|--------------|----------------|-------------|---------------|
+{: format title="Summary of Integer Encodings"}
+
 To reduce unnecessary use of bandwidth, variable length integers SHOULD
```
afrind, line 313:
```suggestion
The four least significant bits of the first byte are reserved in 9-byte
encodings.

To reduce unnecessary use of bandwidth, variable length integers SHOULD
```

draft-ietf-moq-transport.md
afrind: The maximum number of peer initiated uni streams in a quic connection is 2^60.  Sadly it doesn't fit in an 8 byte int, which can only encode up to 2^60-1.

draft-ietf-moq-transport.md
afrind: Yes, but let's sort the other comments first.

draft-ietf-moq-transport.md
```diff
@@ -276,10 +278,44 @@ x (tuple):
...
+
+### Variable-Length Integers
+
+MoQT requires a variable-length integer encoding with the following properties:
+
+1. The encoded length can be determined from the first encoded byte.
+2. The range of 1 byte values is as large as possible.
+3. All 64 bit numbers can be encoded.
+
+The variable-length integer encoding uses the most significant one to four
```
afrind, line 290:
```suggestion
The variable-length integer encoding uses the most significant one to four
```

draft-ietf-moq-transport.md
ianswett: Thanks, of course.

draft-ietf-moq-transport.md
```diff
@@ -262,6 +257,13 @@ x (L) ...:
 : Indicates that x is repeated zero or more times and that each instance
   has a length of L
 
+This document redfines the following RFC9000 syntax:
```
ianswett, line 260:
```suggestion
This document redefines the following RFC9000 syntax:
```

ianswett (approved):
LG and I'm ok with using (i) here the more I think about it.  Because clearly this protocol will in no way work if you try to use QUIC style varints.

draft-ietf-moq-transport.md
suhasHere: I am also fine with x(i) as it is redefinition.  If not, x(V) seems fine to me

suhasHere (approved):

draft-ietf-moq-transport.md
LPardue: I posted somewhere already but I can't find it, so I want to repeat it here. I hate the idea of redefinition, it provides no value and only a source of eternal confusion. Please just pick a different identifier like `mi` or whatever. 

In isolation you think you have no issues. When I want to e.g. draw a diagram that renders how a MoQ control message is sent/received on an HTTP/3 stream I'm going to want to draw something like

```
STREAM Frame {
  Type (i) = 0x08..0x0f,
  Stream ID (i),
  [Offset (i)],
  [Length (i)],
  GOAWAY Message {
    Type (i) = 0x10,
    Length (16),
    New Session URI Length (i),
    New Session URI (..),
  }
```
see the ambiguity?

afrind: I guess I'm sold by @LPardue's example.

draft-ietf-moq-transport.md
afrind: @fluffy sounds good to me.  Would you have it be an error to receive 11110001-11111111?  I don't want to check those bits in my decoder.

draft-ietf-moq-transport.md
```diff
@@ -3443,6 +3482,58 @@ document:
 
 --- back
 
+# Sample Variable-Length Integer Encoding and Decoding {#sample-varint}
+
```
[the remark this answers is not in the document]
afrind: I'm mirroring RFC 9000 here, but I don't have a strong opinion.

draft-ietf-moq-transport.md
fluffy: I don't care if the receiver never checks them, I just care that senders set them all zero and that future extensions can negotiate using them. 

That said, it really does not seem a big deal to check and it's going to be easier to debug if you get a great reason phrase explaining what is going on when they are not valid.

draft-ietf-moq-transport.md
suhasHere: +1 ...  how about we define  x(V64) to be very explicit about the intent of the type ?

@LPardue @afrind @fluffy   ??

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
ianswett: This is in an appendix, correct?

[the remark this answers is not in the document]
afrind: yes, this in an appendix

draft-ietf-moq-transport.md
```diff
@@ -3443,6 +3482,58 @@ document:
...
+The WriteVarint function takes two parameters, a 64 bit integer and an
+output buffer with an append operation.
+
+~~~pseudocode
```
fluffy: Im very against putting code like this in this draft. I'm fine with another draft about implementation advice but we need to keep this draft as short as possible.

draft-ietf-moq-transport.md
```diff
@@ -262,6 +257,13 @@ x (L) ...:
 : Indicates that x is repeated zero or more times and that each instance
   has a length of L
 
+This document redefines the following RFC9000 syntax:
+
+x (vi64):
```
fluffy: I think this should just be (i) not (vi64)

draft-ietf-moq-transport.md
fluffy: Still want the change to 1111XXXX to 11110000 and leave the other values as reserved for future extensions.

fluffy:
any update on this one ?

afrind:
> any update on this one ?

1) We can't land this change until ALPN version negotiation (#499) lands, otherwise this would break version negotiation.  That in turn is blocked on Chrome support for WT-AvailableProtocols, which has just reached Chrome canary.  I expect we can resolve this blocker by Montreal.  Hopefully FireFox and Safari will quickly follow, but I don't think we should wait for them.

2) There are still some issues to discuss here.  First, we've not had a discussion of Victor's alternate design mentioned above or a similar one from MT on the linked issue.  Second, there's the other outstanding feedback here - what to do about the trailing 4 bits when using 64-bit varints, and whether we should include pseudocode.  We can add time to resolve these in Toronto.

afrind:
Discussed in Toronto:

- No one felt strongly one way or the other about having pseudocode in the doc.  I'm happy to remove it for now.
- We should require the tailing bits of 1111XXXX to be 0000.  This will allow future extensibility.
- Not a clear signal about whether we should pursue an alternate encoding.  There wasn't a specific counterproposal that was easy to digest, and we also lack production data about histograms of the values that matter.
- At least one person felt we shouldn't be inventing a new encoding, but should use some existing encoding, although to our knowledge there is not an existing varint encoding with the tradeoffs and properties we want.

afrind:
I believe the counterproposal is:

250 1 byte values, 256 2 byte values, 2^16 3 byte values, etc.

I think this is a worse tradeoff than this PR:

128 1 byte values, 2^14 2 byte values.

For, say, datagram Object IDs in a group where the group is 30s long and there are 50 objects/second, we will more values between 250 and 2^14 than between between 128 and 256.

draft-ietf-moq-transport.md
afrind: I had it as (i), and @LPardue made a convincing argument why not --

if you are trying to show the encoding of multiple protocol layers, then the MOQT (i) and the H3/QUIC (i) difference is not visible.

draft-ietf-moq-transport.md
```diff
@@ -269,20 +264,67 @@ x (..):
 x (L) ...:
 
 : Indicates that x is repeated zero or more times and that each instance
-  has a length of L.
+  has a length of L
```
```suggestion
  has a length of L.
```

draft-ietf-moq-transport.md
```diff
@@ -271,18 +266,65 @@ x (L) ...:
...
+Integers are encoded in 1, 2, 4, 8, or 9 bytes and can encode 7-, 14-, 29-, 60-,
+or 64-bit values, respectively. The following table summarizes the encoding
+properties.
+
+|--------------|----------------|-------------|---------------|
+| Leading Bits | Length (bytes) | Usable Bits | Range         |
+|--------------|----------------|-------------|---------------|
+| 0            | 1              | 7           | 0-127         |
+|--------------|----------------|-------------|---------------|
+| 10           | 2              | 14          | 0-16,383      |
```
ianswett: When I read this at first the comma threw me off, and I thought you were saying you could encode 0-16 as well as the number 383.  But probably other people won't be confused.

afrind: Haha.  Watch for my next April 1 RFC "A complex varint encoding for 4 bit numbers and a single small prime"

afrind: QUIC seems to spell out the entire max of the range without commas: https://datatracker.ietf.org/doc/html/rfc9000#name-variable-length-integer-enc

ianswett: Thanks for updating the PR, this is definitely less confusing for me.

ianswett (approved):

afrind:
This needs a rebase/merge and update new varint fields that have been added.
Victor wants to add 6 byte lengths for time-stampy things

draft-ietf-moq-transport.md
```diff
@@ -245,18 +245,68 @@ This document uses stream management terms described in {{?RFC9000, Section
...
-required value.
+### Variable-Length Integers
+
+MoQT requires a variable-length integer encoding with the following properties:
+
+1. The encoded length can be determined from the first encoded byte.
+2. The range of 1 byte values is as large as possible.
+3. All 64 bit numbers can be encoded.
+
+The variable-length integer encoding uses the most significant one to four
```
```suggestion
The variable-length integer encoding uses the most significant one to five
```

afrind:
@vasilvv found this format, which is used in MKV: https://www.rfc-editor.org/rfc/rfc8794.html#name-variable-size-integer

This is identical to the "9 sizes" encoding discussed here, with 0's and 1's flipped.  We could just adopt that scheme.

fluffy:
On call today, I think we came to encode in a way that can use bit count to get number of additional bytes and have at least one value of that protocol error and might be usable for future protocol extensions.

draft-ietf-moq-transport.md
```diff
@@ -245,18 +245,79 @@ This document uses stream management terms described in {{?RFC9000, Section
...
+| 10           | 2              | 14          | 0-16383                |
+|--------------|----------------|-------------|------------------------|
+| 110          | 3              | 21          | 0-2097151              |
+|--------------|----------------|-------------|------------------------|
+| 1110         | 4              | 28          | 0-268435455            |
+|--------------|----------------|-------------|------------------------|
+| 11110        | 5              | 35          | 0-34359738367          |
+|--------------|----------------|-------------|------------------------|
+| 111110       | 6              | 42          | 0-4398046511103        |
+|--------------|----------------|-------------|------------------------|
```
vasilvv, line 278: nit: add an entry here to emphasize that this is where 7 would be if we've not removed it.

draft-ietf-moq-transport.md
```diff
@@ -245,18 +245,79 @@ This document uses stream management terms described in {{?RFC9000, Section
...
+{: format title="Example Integer Encodings"}
+
+11111100 is an invalid code point.  An endpoint that receives this value MUST
+close the session with a `PROTOCOL_VIOLATION`.
+
+To reduce unnecessary use of bandwidth, variable length integers SHOULD be
+encoded using the least number of bytes possible to represent the required
+value.
+
+x (vi64):
```
vasilvv, line 307: FWIW, I still have a mild preference for `v64`.

vasilvv (approved):

draft-ietf-moq-transport.md
suhasHere: somehow I remember to remove entry for 5 as well from the last interim ?

draft-ietf-moq-transport.md
```diff
@@ -245,18 +245,79 @@ This document uses stream management terms described in {{?RFC9000, Section
...
+| 0xbbbd               | 15,293                     |
+| 0xdd7f3e7d           | 494,878,333                |
+| 0xfaa1a0e403d8       | 2,893,212,287,960          |
+| 0xfefa318fa8e3ca11   | 70,423,237,261,249,041     |
+| 0xffffffffffffffffff | 18,446,744,073,709,551,615 |
+|----------------------|----------------------------|
+{: format title="Example Integer Encodings"}
+
+11111100 is an invalid code point.  An endpoint that receives this value MUST
+close the session with a `PROTOCOL_VIOLATION`.
```
martinthomson, line 301: So if you receive 0b1111_110x you abort?  Weird.  Now your integer decoder - whereever it is invoked - has to be wired into error handling.  Not how I'd have done it.  Better to have the 7 byte encoding than to impose that on implementations.

draft-ietf-moq-transport.md
martinthomson, line 307:
```suggestion
An extension to the packet structure notation uses `vi64` to describe this type.

x (vi64):
```

draft-ietf-moq-transport.md
vasilvv: You already have to deal with that when dealing with truncated varints, though?

martinthomson: That's a very different code path.  Most of the cases involve blocking when we don't have enough data.  Stream termination from one of those blocking states causes an error, but the code that handles the parsing doesn't directly deal with that.

Issue #882, opened 2025-03-27
Can a relay send SUBSCRIBE_OK before talking to the upstream?


vasilvv:
I recall that at some point the answer was "yes" (IIRC we decided against putting delivery order into SUBSCRIBE_OK on those grounds).  Then we shifted away from that with fields like ContentExists in SUBSCRIBE_OK, but the text still says that the relay "SHOULD withhold sending its own SUBSCRIBE_OK until receiving one from upstream"; "SHOULD" here implies that it is not necessary, though it's not clear to me as to where the values for Largest Group ID and such would come in this case.  We should probably be more clear as to what behavior we want.

ianswett:
I thought we settled on the idea that you could send Objects before a SUBSCRIBE_OK, but you should go upstream unless you already had a subscription upstream.

wilaw:
The relay MUST go upstream first unless it already has an open subscription to the track and it has a prior SUBSCRIBE_OK for that open subscription. 

Receiving a prior SUBSCRIBE_OK without an active subscription is insufficient, as the track may have changed or terminated since that information was received.

fluffy:
This is all key to the proposal that Ian and I presented at last meeting and we got agreement on. Lets wait for Ian's PR which will clarify all of this.

afrind:
@fluffy I'm worried we're not talking about the same things.  The PR set being made to address PUBLISH does not say anything about "pending subscribes".  I recall you had some slides about that but skipped them during the IETF 122 session, am I misremembering?  The PUBLISH proposal talked about wildcard SUBSCRIBE which would solicit future PUBLISH messages - is that the part you are talking about?  If not, can you summarize what your expectations are?  Is it related to what @suhasHere is describing in #807?

CC: @ianswett 

For this issue specifically, I prefer MUST go upstream before SUBSCRIBE_OK.  If PUBLISH doesn't address the 807 use case, I have an alternate proposal there - a new message for SUBSCRIBE_PENDING.

afrind:
Discussed in Stockholm - with PUBLISH, the use case for "pending" subscribe will be handled by SUBSCRIBE_ANNOUNCES, so we can change this SHOULD to a MUST.

PR #1017, opened 2025-05-08
Relays MUST have upstream subscription to send SUBSCRIBE_OK

afrind:
"Pending" behavior will be addressed by PUBLISH

Fixes: #882

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1060,11 +1060,10 @@ information can be part of subscription request itself or part of the
 encompassing session. The specifics of how a relay authorizes a user are outside
 the scope of this specification.
 
-The relay will have to send an upstream SUBSCRIBE and/or FETCH if it does not
-have all the objects in the FETCH, or is not currently subscribed to the full
-requested range. In this case, it SHOULD withhold sending its own SUBSCRIBE_OK
-until receiving one from upstream. It MUST withhold FETCH_OK until receiving
-one from upstream.
+The relay MUST have an established upstream subscription before sending
+SUBSCRIBE_OK in response to a downstream SUBSCRIBE.  If a relay does not have
+sufficient information to send a FETCH_OK immediately in response to a FETCH, it
+MUST withhold sending FETCH_OK until it does.
 
 For successful subscriptions, the publisher maintains a list of
 subscribers for each track. Each new Object belonging to the


draft-ietf-moq-transport.md
```diff
@@ -1060,11 +1060,10 @@ information can be part of subscription request itself or part of the
...
 
-The relay will have to send an upstream SUBSCRIBE and/or FETCH if it does not
-have all the objects in the FETCH, or is not currently subscribed to the full
-requested range. In this case, it SHOULD withhold sending its own SUBSCRIBE_OK
-until receiving one from upstream. It MUST withhold FETCH_OK until receiving
-one from upstream.
+The relay MUST have an established upstream subscription before sending
+SUBSCRIBE_OK in response to a downstream SUBSCRIBE.  If a relay does not have
+sufficient information to send a FETCH_OK immediately in response to a FETCH, it
+MUST withhold sending FETCH_OK until it does.
```
suhasHere, line 1066: I am not sure if you mean SUBSCRIBE_OK . Looks like 2 paras are mixed up?

afrind: I think it says what I want.  A relay must have an established upstream subscription in order to send SUBSCRIBE_OK downstream.  I reworded the SUBSCRIBE and FETCH separately cases since they are somewhat different now.

ianswett (approved):

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -1060,11 +1060,10 @@ information can be part of subscription request itself or part of the
 encompassing session. The specifics of how a relay authorizes a user are outside
 the scope of this specification.
 
-The relay will have to send an upstream SUBSCRIBE and/or FETCH if it does not
-have all the objects in the FETCH, or is not currently subscribed to the full
-requested range. In this case, it SHOULD withhold sending its own SUBSCRIBE_OK
-until receiving one from upstream. It MUST withhold FETCH_OK until receiving
-one from upstream.
+The relay MUST have an established upstream subscription before sending
```
fluffy, line 1063: I think this might need a tiny bit of additional clarification on what "established" means - something like this does not mean it has received any objects for the subscription, just that it knows the track exists and has enough information to send the OK.

fluffy (changes requested):
I'm pretty good with this but just marked it RFC because I think it needs a bit o clarification but agree with the design.

draft-ietf-moq-transport.md
afrind: @fluffy The PUBLISH PR is adding more clarity around what an established subscription is (eg: SUBSCRIBE+OK or PUBLISH+OK).

vasilvv (approved):

Issue #1001, opened 2025-05-06
Missing definition of stream. Confusing usage


michalhosna:
Section 2. Object Data Model talks about "stream"

> This allows the client to join at the logical points where they can get the information to start decoding the **stream**. 

Which, from context, talks about a media stream?

Then Section 2.2. Subgroups mention different streams

> Objects in a subgroup have a dependency and priority relationship consistent with sharing a **stream** and are sent on a single **stream** whenever possible. A Group is delivered using at least as many **streams** as there are Subgroups, typically with a one-to-one mapping between Subgroups and **streams**

Which probably talks about QUIC/WebTransport streams?

In my humble opinion, this usage of the word stream may be confusing, especially for new people reading the document top-to-bottom.

afrind:
Thanks for pointing this out - agree we should try not to use "stream" in its media context where avoidable.  Feel free to submit a PR if you feel comfortable, or we will try to address.

michalhosna:
> Feel free to submit a PR if you feel comfortable

Not quite comfortable, but I’ll give it a shot — one has to start somewhere. At the very least, we’ll have something concrete to bikeshed about.
If it’s not any good, we can always throw it out. Hopefully without adding any unnecessary burden.

afrind:
I think this was addressed by #1020

PR #1020, opened 2025-05-09
Minimal Stream definition

michalhosna:
I believe this is the minimal change needed to address #1001.

I’m unsure about the preferred ordering of terms. Is there any guidance on that, or on other editorial details of this kind?

---

Section 9. Talks about “Data streams”. And it is referenced in a couple of places as “data stream”.  Without a prior definition and a differentiation of “Data stream” and “streams”. 
I don’t know the best way forward there. If even a tangible difference between Data stream and stream exists in the context of this document, or what the original reason for the “data” stream term was.

Note. [RTP over QUIC](https://www.ietf.org/archive/id/draft-ietf-avtcore-rtp-over-quic-14.html#section-2-11.18.1) also “struggles” with the usage and definition of the word stream.

This is my first PR here. I am sorry if it ends up being more of a burden than a help. I know this borders on bikeshedding, but hopefully it nudges things a little forward and improves clarity for new readers.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -369,7 +369,7 @@ track to allow the subscriber to pick the appropriate resolution given
 the display environment and available bandwidth. Each "group of pictures"
 in a video is sent as a group because the first frame is needed to
 decode later frames. This allows the client to join at the logical points
-where they can get the information to start decoding the stream.
+where they can get the information to start decoding the media.
 The temporal layers are sent as separate sub groups to allow the
 priority mechanism to favour the base layer when there is not enough
 bandwidth to send both the base and enhancement layers. Each frame of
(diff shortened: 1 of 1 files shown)


ianswett (approved):
Looks good, though it might be nice to stop using "Data Stream" as well because as you pointed out, I don't think there's a difference?  Unless we want to define Data Stream as a Stream with Objects on it, opposed to the Control Stream?

afrind (approved):

michalhosna:
> Looks good, though it might be nice to stop using "Data Stream" as well because as you pointed out, I don't think there's a difference?

I’ll try to put together a more substantial proposal later today or early next week.

Would it make sense to open that as a new PR? This one could be merged as a small improvement on its own.
Or do you generally prefer larger PRs that group related changes together, rather than smaller atomic ones? In which case I can just append it here?

afrind:
> Would it make sense to open that as a new PR? 

Yes, we went ahead and merged this.

PR #1024, opened 2025-05-12
Remove some temporal language

afrind:
Addresses parts of #892 

latest -> largest
before -> smaller location
later/after -> larger location

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1922,18 +1922,18 @@ as defined below:
 
 ## SUBSCRIBE_UPDATE {#message-subscribe-update}
 
-A subscriber issues a SUBSCRIBE_UPDATE to a publisher to request a change to an
-existing subscription. Subscriptions can only become more narrow, not wider,
-because an attempt to widen a subscription could fail. If Objects with Locations
-smaller than the Start Location of the current subscription are needed, a FETCH
-might be able to retrieve them. The Start Location MUST NOT decrease and when it
-increases, there is no guarantee that a publisher will not have already sent
-Objects with IDs smaller the new Start Location.  The End Group MUST NOT
-increase and when it decreases, there is no guarantee that a publisher will not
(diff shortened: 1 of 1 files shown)


sharmafb (approved):

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1922,18 +1922,18 @@ as defined below:
...
+because an attempt to widen a subscription could fail. If Objects with Locations
+smaller than the Start Location of the current subscription are needed, a FETCH
+might be able to retrieve them. The Start Location MUST NOT decrease and when it
+increases, there is no guarantee that a publisher will not have already sent
+Objects with IDs smaller the new Start Location.  The End Group MUST NOT
+increase and when it decreases, there is no guarantee that a publisher will not
+have already sent Objects with Locations larger than the new End Location. A
+publisher SHOULD close the Session as a 'Protocol Violation' if the
+SUBSCRIBE_UPDATE violates either rule or if the subscriber specifies a Request
+ID that has not existed within the Session.
```
suhasHere, line 1936:
```suggestion
A subscriber sends a SUBSCRIBE_UPDATE to a publisher to request a modification to an existing 
subscription. Subscriptions can only be narrowed, not widened, as widening could result in failure. 
If Objects with Locations earlier than the current subscription's Start Location are required, a 
FETCH may be used to retrieve them. The Start Location MUST NOT decrease, and if it increases, 
there is no guarantee that the publisher has not already sent Objects with Locations smaller than the 
new Start Location. Similarly, the End Group MUST NOT increase, and if it decreases, there is no 
guarantee that the publisher has not already sent Objects with Locations larger than the new End Location. 
A publisher MUST terminate the session with a 'Protocol Violation' if the SUBSCRIBE_UPDATE violates 
these rules or if the subscriber specifies a request ID that has not existed within the Session.
```

suhasHere (approved):
minor nits. Otherwise LGTM.

PR #1025, opened 2025-05-14
add marker for end to end header authentication

fluffy:
This add an header extension that indicates extensions things after this can be end to end authenticated ( not encrypted ).  

This PR is to set up for work in ObjCrypto draft that provided the end to end encryption. This is becoming an implementation blocker for us as we start working through the end to end security stuff.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3366,6 +3366,36 @@ cannot infer any information about the existence of prior groups (see
 This extension can be added by the Original Publisher, but MUST NOT be added by
 relays. This extension MUST NOT be modified or removed.
 
+## Immutable Extensions Marker
+
+The "Immutable Extension Marker" (Extension Header Type 0xA) serves as a
+boundary indicating the start of immutable header extensions.  The
+primary purpose of this marker is so that end to end encryption schemes
+know that any header extensions after and including this marker can be
+end to end authenticated.
+
+Header extensions appearing after this marker MUST NOT be modified or
+removed by relays and MUST be cached if the object is
+cached. Conversely, header extensions preceding this marker MAY be
+modified or removed by relays, as specified by their individual
+specifications. The Immutable Extension Marker is assigned Type 0xA and
+represents a boolean value.  It's absence implies that none of the
+header extensions should be end to end authenticated.
+
+The following figure shows an example Object structure with a
+combination of mutable and immutable extensions and end to end encrypted
+extensions in the object payload.
+
+~~~
++------+------------+-----------+-------------+------------+--------+
+|Header| Mutable    | Marker    | Immutable   | Encrypted  | Object |
+|      | Extensions | Extension | Extensions  | Extensions | Data   |
++------+------------+-----------+-------------+------------+--------+
+                   Object Header                    Object Payload
+<---------------------------------------------><-------------------->
+~~~
+
+
 # Security Considerations {#security}
 
 TODO: Expand this section, including subscriptions.


fluffy:
This is similar to rfc6904 in webrtc

suhasHere (approved):

afrind (commented):
Directionally this seems ok to me.  Added some comments.

draft-ietf-moq-transport.md
```diff
@@ -3366,6 +3366,36 @@ cannot infer any information about the existence of prior groups (see
 This extension can be added by the Original Publisher, but MUST NOT be added by
 relays. This extension MUST NOT be modified or removed.
 
+## Immutable Extensions Marker
+
+The "Immutable Extension Marker" (Extension Header Type 0xA) serves as a
+boundary indicating the start of immutable header extensions.  The
+primary purpose of this marker is so that end to end encryption schemes
```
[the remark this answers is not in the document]
fluffy: probably better to say AEAD schemes (which is what IETF requires) and that would be less confusing.

draft-ietf-moq-transport.md
```diff
@@ -3366,6 +3366,36 @@ cannot infer any information about the existence of prior groups (see
...
+know that any header extensions after and including this marker can be
+end to end authenticated.
+
+Header extensions appearing after this marker MUST NOT be modified or
+removed by relays and MUST be cached if the object is
```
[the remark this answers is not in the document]
fluffy: yes, that was implied - glad to add explicit text on that

draft-ietf-moq-transport.md
```diff
@@ -3366,6 +3366,36 @@ cannot infer any information about the existence of prior groups (see
...
+cached. Conversely, header extensions preceding this marker MAY be
+modified or removed by relays, as specified by their individual
+specifications. The Immutable Extension Marker is assigned Type 0xA and
+represents a boolean value.  It's absence implies that none of the
+header extensions should be end to end authenticated.
```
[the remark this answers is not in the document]
fluffy: 100% agree with not using lower case version of should / must. 

how about change "should" to "not meant to be"

draft-ietf-moq-transport.md
```diff
@@ -3366,6 +3366,36 @@ cannot infer any information about the existence of prior groups (see
...
+
+The following figure shows an example Object structure with a
+combination of mutable and immutable extensions and end to end encrypted
+extensions in the object payload.
+
+~~~
++------+------------+-----------+-------------+------------+--------+
+|Header| Mutable    | Marker    | Immutable   | Encrypted  | Object |
+|      | Extensions | Extension | Extensions  | Extensions | Data   |
++------+------------+-----------+-------------+------------+--------+
```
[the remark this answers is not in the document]
fluffy: Agree on need to find a way to not have confusing terminology but Loc has a need for encrypted extension and I want to show that those fit into the Object Payload part as this has been a topic of discussion in WG and it is key design thing from the amount of overhead we end up adding.

draft-ietf-moq-transport.md
```diff
@@ -3366,6 +3366,36 @@ cannot infer any information about the existence of prior groups (see
...
+primary purpose of this marker is so that end to end encryption schemes
+know that any header extensions after and including this marker can be
+end to end authenticated.
+
+Header extensions appearing after this marker MUST NOT be modified or
+removed by relays and MUST be cached if the object is
+cached. Conversely, header extensions preceding this marker MAY be
+modified or removed by relays, as specified by their individual
+specifications. The Immutable Extension Marker is assigned Type 0xA and
+represents a boolean value.  It's absence implies that none of the
```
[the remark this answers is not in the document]
afrind: I think we're waiting on a PR for extension types/ranges etc from @mzanaty

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
afrind: I think it's fine to say that the purpose of the immutability marker is to facilitate end-to-end authentication of _relay visible_ extensions.  But a streaming format can also define end-to-end authentication of _non-relay visible_ metadata within its Object Payload definition.  The text you have implies that the only way to do auth is by making the metadata explicitly relay visible via extensions.

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
afrind: LoC can define the contents of its payload however it likes.  MoQT though shouldn't make any assumptions or assertions about payload contents.  Maybe in the example divide Object Payload into "Metadata" and "Object Data" and indicate both Metadata and Object Data can be encrypted?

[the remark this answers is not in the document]
afrind: I made a PR against Secure Objects to explain "private extensions" that are structured metadata from a Secure Objects POV but serialized and encrypted with the payload: https://github.com/suhasHere/moq-secure-objects/pull/25.

Since LoC doesn't have a hard dep on Secure Object (eg: moq-mi sends unencrypted LoC objects), I would recommend updating LoC to explain which extensions defined there are public, immutable and private, both when Secure Objects is used and when it is not.

draft-ietf-moq-transport.md
```diff
@@ -3366,6 +3366,36 @@ cannot infer any information about the existence of prior groups (see
 This extension can be added by the Original Publisher, but MUST NOT be added by
 relays. This extension MUST NOT be modified or removed.
 
+## Immutable Extensions Marker
+
+The "Immutable Extension Marker" (Extension Header Type 0xA) serves as a
+boundary indicating the start of immutable header extensions.  The
+primary purpose of this marker is so that end to end encryption schemes
+know that any header extensions after and including this marker can be
+end to end authenticated.
```
```suggestion
The "Immutable Extension Marker" (Extension Header Type 0xA) serves as a
boundary indicating the start of immutable header extensions.
```

draft-ietf-moq-transport.md
```diff
@@ -3366,6 +3366,36 @@ cannot infer any information about the existence of prior groups (see
...
+combination of mutable and immutable extensions and end to end encrypted
+extensions in the object payload.
+
+~~~
++------+------------+-----------+-------------+------------+--------+
+|Header| Mutable    | Marker    | Immutable   | Encrypted  | Object |
+|      | Extensions | Extension | Extensions  | Extensions | Data   |
++------+------------+-----------+-------------+------------+--------+
+                   Object Header                    Object Payload
+<---------------------------------------------><-------------------->
```
```suggestion
The following figure shows an example Object structure with a combination of
mutable and immutable extensions and end to end encrypted metadata in the object
payload.

~~~
                   Object Header                          Object Payload
<------------------------------------------------------> <------------->
+--------+------------+--------+------------+-----------+--------------+
| Object | Mutable    | Marker | Immutable  | [Payload] | App Metadata |
| Fields | Extensions |        | Extensions | [Length ] | App Payload  |
+--------+------------+--------+------------+-----------+--------------+
                       xxxxxxxxxxxxxxxxxxxxx              xxxxxxxxxxxxx
                                                          yyyyyyyyyyyyy
x = e2e Authenticated Data
y = e2e Encrypted Data
```

draft-ietf-moq-transport.md
```suggestion
specifications. The Immutable Extension Marker is assigned Type 0xA and is a
variable length integer that MUST have value 1.  Any other value results in
closing the session with a `PROTOCOL_VIOLATION`.
```

afrind: Just until we have a way to define true boolean extensions.

draft-ietf-moq-transport.md
```diff
@@ -3366,6 +3366,36 @@ cannot infer any information about the existence of prior groups (see
...
+The "Immutable Extension Marker" (Extension Header Type 0xA) serves as a
+boundary indicating the start of immutable header extensions.  The
+primary purpose of this marker is so that end to end encryption schemes
+know that any header extensions after and including this marker can be
+end to end authenticated.
+
+Header extensions appearing after this marker MUST NOT be modified or
+removed by relays and MUST be cached if the object is
+cached. Conversely, header extensions preceding this marker MAY be
+modified or removed by relays, as specified by their individual
```
```suggestion
Relays MUST NOT add, modify, remove or reorder extensions after the Immutable
Extension Marker.  Relays MUST cache all immutable header extensions if the
object is cached. Conversely, header extensions preceding this marker MAY be
modified or removed by relays, as specified by their individual
```

afrind (commented):
Some concrete suggestions:

1. Use an integer type, since we don't have boolean types yet (we can update once we do)
2. Further restrict what happens after the marker - no adding or reordering
3. Updated the example.  I used "application metadata or extensions" in text and "App Metadata" in the table.

draft-ietf-moq-transport.md
suhasHere: @afrind  Even the immutable header extension is part of AAD scheme ( as it was in the original text)

afrind: MOQT does not define any end-to-end authentication or encryption schemes (Secure Object does).  I recommend representing that in the example below by including `xxxx` under the marker as well, and building the requirement that the marker extension be included on my [Secure Objects PR](https://github.com/suhasHere/moq-secure-objects/pull/25).

draft-ietf-moq-transport.md
fluffy: So this is good but I think we also need to have e2e encrypted object headers.

fluffy: I'm going to commit this and add a bit more

fluffy:
@afrind - thanks for all the updates. I merged all your change and changed slightly to indicate payload length was also part of what was authenticated as the security needs that.

fluffy:
@afrind  - can you re-review this one - thanks

suhasHere:
@fluffy, @afrind  
the latest updates LGTM

draft-ietf-moq-transport.md
```diff
@@ -3366,6 +3366,37 @@ cannot infer any information about the existence of prior groups (see
...
+~~~
+                   Object Header                          Object Payload
+<------------------------------------------------------> <---------------->
++--------+------------+--------+------------+-----------+-----------------+
+| Object | Mutable    | Marker | Immutable  | [Payload] | Enc Obj Headers |
+| Fields | Extensions |        | Extensions | [Length ] | App Payload     |
++--------+------------+--------+------------+-----------+-----------------+
+                       xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+                                                          yyyyyyyyyyyyyyyy
+x = e2e Authenticated Data
```
afrind, line 3679: For my own understanding -- the [Payload Length] piece is _implicitly_ rather than explicitly included in AEAD, is that right?  It seems like you don't know that length in advance - it's an output from AEAD.

draft-ietf-moq-transport.md
```diff
@@ -3366,6 +3366,37 @@ cannot infer any information about the existence of prior groups (see
...
+
+The following figure shows an example Object structure with a combination of
+mutable and immutable extensions and end to end encrypted metadata in the object
+payload.
+
+~~~
+                   Object Header                          Object Payload
+<------------------------------------------------------> <---------------->
++--------+------------+--------+------------+-----------+-----------------+
+| Object | Mutable    | Marker | Immutable  | [Payload] | Enc Obj Headers |
```
afrind: > Enc Obj Headers

This reads a little strange to me since the yyyyy is meant to show what is encrypted (eg the headers AND the App payload).

But I don't want to faceplant on the example.  I think we're talking about the same thing.

afrind (approved):

afrind:
This is an implementation detail, but from an API perspective it seems like you want to pass Mutable and Immutable extensions into/out-of your MOQT impl in two different maps, and the marker only appears on the wire, rather than in APIs.

draft-ietf-moq-transport.md
fluffy: Good question ... No, you know how much the crypto will increase the size based on what crypto algorithm is being used so you do know the length before you do the encryption step. If you look at lots of AEAD code they allocate the output buffer before calling encrypt based on size of input buffer and based on what crypto was chosen. The security problem is that some ciphers modes might allow you to truncate the stream by changing the length so for some ciphers it is probably best practice to authenticate the length to stop some weird attacks that cut and paste of encrypted data. For some modes this is not needed as auth would fail even without this but I think  it is a security good practice and needed for some cases.

draft-ietf-moq-transport.md
fluffy: yah, I know what you mean - I'm trying to seprate the headers that defined by MoQ from the headers are defined by an applications that sits on top of MoQ and are not part of the MoQ header registry.

draft-ietf-moq-transport.md
afrind: Ok, just note 1) the length doesn't always appear on the wire in that spot (datagrams use the enclosing dg length - other fields) and 2) I don't see how that length is included explicitly as _input_ anywhere in the Secure Objects draft.

draft-ietf-moq-transport.md
```diff
@@ -3366,6 +3366,37 @@ cannot infer any information about the existence of prior groups (see
...
+Relays MUST NOT add, modify, remove or reorder extensions after the Immutable
+Extension Marker.  Relays MUST cache all immutable header extensions if the
+object is cached. Conversely, header extensions preceding this marker MAY be
+modified or removed by relays, as specified by their individual
+specifications. The Immutable Extension Marker is assigned Type 0xA and is a
+variable length integer that MUST have value 1.  Any other value results in
+closing the session with a `PROTOCOL_VIOLATION`.
+
+The following figure shows an example Object structure with a combination of
+mutable and immutable extensions and end to end encrypted metadata in the object
```
```suggestion
mutable and immutable extensions and end to end encrypted metadata in the Object
```

draft-ietf-moq-transport.md
afrind: > I'm trying to seprate the headers that defined by MoQ from the headers are defined by an applications that sits on top of MoQ and are not part of the MoQ header registry

Maybe after this:

> The following figure shows an example Object structure with a combination of
mutable and immutable extensions and end to end encrypted metadata in the object
payload.

Add: 

The metadata can be structured (e.g. as key/value pairs) which is encoded and prepended to the Application Payload prior to encryption.

```
| Encoded Metadata |
| App Payload      |
 yyyyyyyyyyyyyyyyyy
```

vasilvv:
I wonder if instead of making this a marker we should make this an extension where the payload is the immutable extensions?  If the total length of immutable extensions is below 64 bytes, this would be length-neutral with respect to the current definition (since it already requires a `1` that does not do anything).

afrind:
> I wonder if instead of making this a marker we should make this an extension where the payload is the immutable extensions? If the total length of immutable extensions is below 64 bytes, this would be length-neutral with respect to the current definition (since it already requires a `1` that does not do anything).

I think I like that better.  It's much easier semantically to say "you can't change the content of this extension" than "you can't add, remove, modify or reorder these".  It may also make the layering with secure object more clear, since one of the inputs is basically the buffer containing the serialized immutable extensions, which would be the value of the this parameter.  I think if the length of your immutable extensions exceeds 64 (or 128 after new varints) then you probably don't care about one more byte in the object?

afrind:
Victor's proposal also gives us more future flexibility.

fluffy:
I like victors suggestion. The relay still has to part the stuff in the extension to see if there is anything it wants. So to take this one step further, I think we should have two extensions Immutable and ImmutableNotVisable - the relay can ignore everything in the ImmutableNotVisable but would still need to parse the stuff in the Immutable. 

We will need text that these can not contain either of them ( we don't want recursive containers ).  In the case where the payloads are above 64 bytes, the messages are already large enough the 1 byte increase does not matter so seems fine from space point of view.

afrind:
> I like victors suggestion. The relay still has to part the stuff in the extension to see if there is anything it wants. So to take this one step further, I think we should have two extensions Immutable and ImmutableNotVisable - the relay can ignore everything in the ImmutableNotVisable but would still need to parse the stuff in the Immutable.

You are saying we need to add an extension whose content is "don't look at this"?  I assume these are for holding things are are serialized and then encrypted into a ciphertext buffer? What is the value in doing this over https://github.com/suhasHere/moq-secure-objects/pull/25 where the private extensions are within the MOQT payload? It would seem to make Secure Object slightly more complicated.

> We will need text that these can not contain either of them ( we don't want recursive containers ). 

+1 excellent point

fluffy:
Ignore E2EE for a second - most deployments will not use E2EE. We will have options that are only meant for ends and not the relays. The relays need to have a way to know they can skip past those and ignore them as this will speed things up. 

E2EE can use this too but relays don't implement the E2EE RFC, they would implement the base spec. So whatever relays need to do to support E2EE sort of needs to be in the base spec. The stuff that goes in the object is defined by the applications that use the object but E2EE needs to be defined outside of the objects.

afrind:
> Ignore E2EE for a second - most deployments will not use E2EE. We will have options that are only meant for ends and not the relays. The relays need to have a way to know they can skip past those and ignore them as this will speed things up.

There's already a marker indicating what a relay can skip over, it's the MOQT Payload length.  Also, any application can define an extension, which is also skipped over.

> E2EE can use this too but relays don't implement the E2EE RFC, they would implement the base spec. So whatever relays need to do to support E2EE sort of needs to be in the base spec.

We've already achieved that - private extensions can be serialized and encrypted along with the rest of the application payload inside the MOQT payload, or placed in an extension defined by the E2EE spec.

> The stuff that goes in the object is defined by the applications that use the object but E2EE needs to be defined outside of the objects.

I really don't understand _why_ this needs to be outside of the MOQT payload, or how that provides any value.  Isn't it also more efficient to encrypt/decrypt the serialized private extensions along with the application payload, rather than in two separate blobs?  I would really appreciate comments on my secure objects PR, or an alternate PR that explains how these two are working together.

draft-ietf-moq-transport.md
```diff
@@ -3637,6 +3637,36 @@ cannot infer any information about the existence of prior groups (see
...
++--------+------------+--------+------------+-----------+-----------------+
+| Object | Mutable    | Marker | Immutable  | [Payload] | Enc Obj Headers |
+| Fields | Extensions |        | Extensions | [Length ] | App Payload     |
++--------+------------+--------+------------+-----------+-----------------+
+                       xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+                                                          yyyyyyyyyyyyyyyy
+x = e2e Authenticated Data
+y = e2e Encrypted Data
+~~~
+
```
```suggestion
## Immutable Extensions

The Immutable Extensions (Extension Header Type 0xB) contains a sequence of
Key-Value-Pairs (see {{moq-key-value-pair}}) which are also Object Extension
Headers of the Object.

~~~
Immutable Extensions {
  Type (0xB),
  Length (i),
  Key-Value-Pair (..) ...
}
~~~

This extension can be added by the Original Publisher, but MUST NOT be added by
Relays. This extension MUST NOT be modified or removed. Relays MUST cache this
extension if the Object is cached.  Relays are permitted to decode and view
these extensions.

A Track is considered malformed (see {{malformed-tracks}}) if any of the
following conditions are detected:

 * An Object contains an Immutable Extensions header that contains another
   Immutable Extensions key
 * A Key-Value-Pair cannot be parsed

The following figure shows an example Object structure with a combination of
mutable and immutable extensions and end to end encrypted metadata in the Object
payload.

~~~
                   Object Header                      Object Payload
<------------------------------------------------> <------------------->
+--------+-------+------------+-------+-----------+--------------------+
| Object | Ext 1 | Immutable  | Ext N | [Payload] | Private Extensions |
| Fields |       | Extensions |       | [Length]  | App Payload        |
+--------+-------+------------+-------+-----------+--------------------+
                  xxxxxxxxxxxx                     xxxxxxxxxxxxxxxxxxxx
                                                   yyyyyyyyyyyyyyyyyyyy
x = e2e Authenticated Data
y = e2e Encrypted Data
~~~

```

draft-ietf-moq-transport.md
```diff
@@ -3637,6 +3637,49 @@ cannot infer any information about the existence of prior groups (see
...
+                   Object Header                      Object Payload
+<------------------------------------------------> <------------------->
++--------+-------+------------+-------+-----------+--------------------+
+| Object | Ext 1 | Immutable  | Ext N | [Payload] | Private Extensions |
+| Fields |       | Extensions |       | [Length]  | App Payload        |
++--------+-------+------------+-------+-----------+--------------------+
+                  xxxxxxxxxxxx                     xxxxxxxxxxxxxxxxxxxx
+                                                   yyyyyyyyyyyyyyyyyyyy
+x = e2e Authenticated Data
+y = e2e Encrypted Data
```
suhasHere, line 3680:
```suggestion
y = e2e Encrypted Data
EXT 1 and EXT N can be modified or removed by Relays
```

suhasHere (approved):
This is looking good

draft-ietf-moq-transport.md
```diff
@@ -3637,6 +3637,49 @@ cannot infer any information about the existence of prior groups (see
...
+Immutable Extensions {
+  Type (0xB),
+  Length (i),
+  Key-Value-Pair (..) ...
+}
+~~~
+
+This extension can be added by the Original Publisher, but MUST NOT be added by
+Relays. This extension MUST NOT be modified or removed. Relays MUST cache this
+extension if the Object is cached.  Relays are permitted to decode and view
```
suhasHere, line 3656:
```suggestion
extension if the Object is cached.  Relays MAY decode and view these extensions, however
```

draft-ietf-moq-transport.md
```diff
@@ -3637,6 +3637,49 @@ cannot infer any information about the existence of prior groups (see
...
+these extensions.
```
```suggestion
This extension can be added by the Original Publisher, but MUST NOT be added by
Relays. This extension MUST NOT be modified or removed. Relays MUST cache this
extension if the Object is cached and MUST forward this extension if the
enclosing Object is forwarded. Relays MAY decode and view these extensions.
```

draft-ietf-moq-transport.md
```diff
@@ -3637,6 +3637,50 @@ cannot infer any information about the existence of prior groups (see
 This extension can be added by the Original Publisher, but MUST NOT be added by
 relays. This extension MUST NOT be modified or removed.
 
+## Immutable Extensions
```
ianswett, line 3640: Nit: Why do we call them extensions in some cases and headers in others?  I realize header is used by HTTP, but I still think it's better than extension header.

draft-ietf-moq-transport.md
```diff
@@ -3637,6 +3637,50 @@ cannot infer any information about the existence of prior groups (see
...
+  Key-Value-Pair (..) ...
+}
+~~~
+
+This extension can be added by the Original Publisher, but MUST NOT be added by
+Relays. This extension MUST NOT be modified or removed. Relays MUST cache this
+extension if the Object is cached and MUST forward this extension if the
+enclosing Object is forwarded. Relays MAY decode and view these extensions.
+
+A Track is considered malformed (see {{malformed-tracks}}) if any of the
```
ianswett, line 3659: Nit: I think it's cleaner to talk about malformed Objects and then say any Track that contains malformed Objects is malformed?  Also, what about simple corruption?  Maybe that's a case we can only handle with encryption and authentication of Objects?

afrind: Nothing we can do about corruption without authentication.  For now, the language about "Track is malformed if an object contains..." is consistent with the other sections.

draft-ietf-moq-transport.md
afrind: Filed #1181

vasilvv (approved):

vasilvv:
One thing we should probably mention is that relays should actually look into immutable extensions by default.  I.e. if I put the gap indication into immutable block, it should read it.

PR #1027, opened 2025-05-16
Reset stream clarification when setting SUBSCRIBE_UPDATE to false

sharmafb:
There's some ambiguity as to whether the publisher should send a FIN or RESET the stream when forward is set to false, causing no more objects to be delivered in a subgroup. I'm attempting to clarify it here.


ianswett (commented):
I'm not sure this is correct in all cases.  The forward flag changes what is sent, but the intent is that FIN and RESET convey information about the Group structure.

PR #1030, opened 2025-05-23
Editorial changes for issues #959, #983, #985, #988, #998, #1000, and #1021

### LINKED ISSUE #959: Editorial: "MOQT" vs "MoQT", and "a MOQT" vs "an MOQT"
There are 80 instances of "MOQT" vs 7 instances of "MoQT" (these 7 instances are in Sections 5, 6.1, 6.2, and 9.4.3). I guess it'd be better to change those 7 instances of "MoQT" to "MOQT" to be consistent.

There are 6 instances of "a MOQT" vs 3 instances of "an MOQT". I guess it'd be better to change those 6 instances of "a MOQT" to "an MOQT" to be consistent.

### LINKED ISSUE #983: On Considerations for Setting Priorities
The first paragraph in Section 6.3 is as follows:

Relays SHOULD respect the subscriber and original publisher's priorities. Relays can receive subscriptions with conflicting subscriber priorities or Group Order preferences. Relays SHOULD NOT directly use Subscriber Priority or Group Order from incoming subscriptions for upstream subscriptions. Relays use of these fields for upstream subscriptions can be based on factors specific to it, such as the popularity of the content or policy, or relays can specify the same value for all upstream subscriptions.

The first sentence (Relays SHOULD respect the subscriber and original publisher's priorities.) and the third sentence (Relays SHOULD NOT directly use Subscriber Priority or Group Order from incoming subscriptions for upstream subscriptions.) appear, to me, a little contradicting with each other. Is it possible to rephrase to avoid such seemingly contradiction, or it is totally fine with everyone (thus my feeling is simply not correct)?

### LINKED ISSUE #985: Inconsistency on relay modifying extension headers in different sections
Sec. 9.1.1.2 contains the following sentence: "If supported by the relay and subject to the processing rules specified in the definition of the extension, Extension Headers MAY be modified, added, removed, and/or cached by relays."

However, in Sec. 7.4, it is said that "A relay MUST NOT modify Object properties when forwarding.", while in Sec. 9.1.1 (titled Object Properties), one of the listed Object properties is Object Extensions, which is defined as a sequence of Object Extension Headers. 

Therefore, some changes need to be made to avoid this inconsistency. Sorry that I am not familiar with all involved aspects enough to confidently suggest a modification, thus I am just reporting the issue without a suggestion to resolve it.

### LINKED ISSUE #988: A bunch of minor and seemingly obvious editorial improvements
As I continue reading the draft, I have gathered a bunch of minor and seemingly obvious editorial improvements, and counting. Following @afrind's suggestion, I could directly file a PR for each of these editorial improvements. However, it'd be much easier and more convenient to bundle these collectively in one PR. Thus, I am posting what I have so far below, just in case someone objects to one or more of them, then those that are not agreeable can be excluded from the PR. Any opinion on any of the suggestions is appreciated.

Here is the list of my minor and seemingly obvious editorial improvements so far. The changed parts are highlighted in **bold fonts**.

a.	2.1: "All objects belong to a group, indicating ordering and potential dependencies. Section 2.3" -> "All objects belong to a group, indicating ordering and potential dependencies **(see Section 2.3)**."

b.	3.1.3: "The syntax of the Connection URL and the associated connection setup procedures are specific to the underlying transport protocol usage Section 3." -> "The syntax of the Connection URL and the associated connection setup procedures are specific to the underlying transport protocol usage **(see Section 3)**."

c.	3.2: "It also includes the union of all Setup Parameters Section 8.2.2 required for a handshake by any of those versions." -> "It also includes the union of all Setup Parameters **(see Section 8.2.2)** required for a handshake by any of those versions."

d.	3.4: "The application MAY use any error message and SHOULD use a relevant code, as defined below:" -> "**When terminating the Transport Session, t**he application MAY use any error message and SHOULD use a relevant code, as defined below:"

e.	3.4, 8.4: "equal or larger than" -> "**equal to** or larger than".

f.	3.4: "The session was closed because the peer took too long to send data expected on an open Data Stream Section 9." -> "The session was closed because the peer took too long to send data expected on an open Data Stream **(Section 9)**."

g.	6.2: "If two objects have a different subscriber priority associated with them, the one with the highest subscriber priority is sent first." -> "If two objects have **different subscriber priorities** associated with them, the one with the highest subscriber priority is sent first."

h.	6.2: "If two objects have the same subscriber priority, but a different publisher priority, the one with the highest publisher priority is sent first." -> "If two objects have the same subscriber priority, but **different publisher priorities**, the one with the highest publisher priority is sent first."

i.	6.3: "Relays use of these fields for upstream subscriptions can be ..." -> "**Relays' use** of these fields for upstream subscriptions can be ..."

j.	7.3.1: "This section describes behavior that a publisher MAY choose to implement to allow for a better users experience when switching between networks, such as WiFi to Cellular or vice versa." -> "This section describes **a behavior** that a publisher MAY choose to implement to allow for a better **user experience** when switching between networks, such as WiFi to Cellular or vice versa."

k.	7.3.1: "Once the subscriptions have migrated over to session on the new network, the publisher can stop publishing objects on the old network." -> "Once the subscriptions have migrated over to **the session** on the new network, the publisher can stop publishing objects on the old network."

l.	7.3.2: "This section describes behavior that a publisher MAY choose to implement to allow for a better user experience when a relay sends them a GOAWAY." -> " This section describes **a behavior** that a publisher MAY choose to implement to allow for a better user experience when a relay sends them a GOAWAY."

m.	8.2: "Some messages include a Parameters field that encode optional message elements." -> "Some messages include a Parameters field that **encodes** optional message elements."

n.	8.2: "The number of parameters in a message is not specifically limited, but the total length of a control message is limited to 2^16-1." -> "The number of parameters in a message is not specifically limited, but the total length of a control message is limited to **2^16-1 bytes**."

### LINKED ISSUE #998: On the PATH parameter
Sec. 8.3.2.1 includes the following:

> It MUST NOT be used by the server, or when WebTransport is used. If the peer receives a PATH parameter from the server, when WebTransport is used, or one the server does not support, it MUST close the session with Invalid Path.

The second sentence has some issues. Was the intention to have the sentence to say something as follows?

> If the peer receives a PATH parameter from the server, or when WebTransport is used, it MUST close the session with Invalid Path.

If not, what is the intention? In any case, the phrase "or one the server does not support" has multiple issues, grammatically and on technical consistency with the preceding sentence.

### LINKED ISSUE #1000: 2nd bunch of minor and seemingly obvious editorial improvements
o.	8.2: "Parameters are serialized as Key-Value-Pairs Figure 2." -> "Parameters are serialized as Key-Value-Pairs **(see Figure 2)**."

p.	8.2.1: "If it appears in some other type of message, it MUST be ignored." -> "If it appears in some other **types of messages**, it MUST be ignored."

q.	8.2.1.1: "Token Type - a numeric identifier for the type of Token payload being transmitted. This type is defined by the IANA table "MOQT Auth Token Type". See Section 12." -> "Token Type - a numeric identifier for the type of Token payload being transmitted. This type is defined by **the IANA table "MOQT Auth Token Type" (see Section 12)**."

r.	8.2.1.1: "... though it MAY discard other state associated with the token that is no longer needed." -> "... though it MAY discard other **states** associated with the token that is no longer needed."

s.	8.2.1.1: "The receiver can protect its resources by sending a SETUP parameter defining the MAX_AUTH_TOKEN_CACHE_SIZE Section 8.3.2.3 limit it is willing to accept." -> "The receiver can protect its resources by sending a SETUP parameter defining **the MAX_AUTH_TOKEN_CACHE_SIZE limit (Section 8.3.2.3)** it is willing to accept."

t.	8.2.1.2: "The DELIVERY TIMEOUT parameter (Parameter Type 0x02) MAY appear in a TRACK_STATUS, SUBSCRIBE, SUBSCRIBE_OK, or a SUBSCRIBE_UDPATE message." -> "The DELIVERY TIMEOUT parameter (Parameter Type 0x02) MAY appear in a TRACK_STATUS, SUBSCRIBE, **SUBSCRIBE_OK or SUBSCRIBE_UDPATE** message."

### LINKED ISSUE #1021: Semantics of Request ID in the MAX_REQUEST_ID message
Sorry if this has already been discussed and concluded.

In Sec. 8.5, the semantics of Request ID in the MAX_REQUEST_ID message is as follows:

> Request ID: The new Maximum Request ID for the session. If a Request ID equal or larger than this is received by the endpoint that sent the MAX_REQUEST_ID in any request message (ANNOUNCE, FETCH, SUBSCRIBE, SUBSCRIBE_ANNOUNCES or TRACK_STATUS_REQUEST), the endpoint MUST close the session with an error of 'Too Many Requests'.

The first sentence says that this is the maximum Request ID, i.e., the maximum value is allowed. However, the second sentence says that the maximum value is not allowed, Therefore, the two sentences are not entire consistent with each other.

Either the first sentence and probably also the name of the message should be changed to be aligned with the sentence. Or, the second sentence is changed to be as follows (changing the sentence is probably simpler from the text point of view):

> If a **Request ID larger than this** is received by the endpoint that sent the MAX_REQUEST_ID in any request message (ANNOUNCE, FETCH, SUBSCRIBE, SUBSCRIBE_ANNOUNCES or TRACK_STATUS_REQUEST), the endpoint MUST close the session with an error of 'Too Many Requests'.

yekuiwang:
This PR contains editorial changes for multiple issues.

Fixes #959
Fixes #983
Fixes #985
Fixes #988
Fixes #998
Fixes #1000 
Fixes #1021

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -468,11 +468,11 @@ the Group is received. This can create gaps in a cache that can be filled
 by doing a Fetch upstream, if necessary.
 
 Applications that cannot produce Group IDs that increase with time are limited
-to the subset of MOQT that does not compare group IDs. Subscribers to these Tracks
-SHOULD NOT use range filters which span multiple Groups in FETCH or SUBSCRIBE.
-SUBSCRIBE and FETCH delivery use Group Order, so a FETCH cannot deliver Groups
-out of order and a subscription could have unexpected delivery order if Group IDs
-do not increase with time.
+to the subset of MOQT that does not compare group IDs. Subscribers to these
+Tracks SHOULD NOT use range filters which span multiple Groups in FETCH or
+SUBSCRIBE.  SUBSCRIBE and FETCH delivery use Group Order, so a FETCH cannot
+deliver Groups out of order and a subscription could have unexpected delivery
+order if Group IDs do not increase with time.
 
 ## Track {#model-track}
 
@@ -504,10 +504,10 @@ close the session with a Protocol Violation.
 In this specification, both the Track Namespace tuple fields and the Track Name
 are not constrained to a specific encoding. They carry a sequence of bytes and
 comparison between two Track Namespace tuple fields or Track Names is done by
-exact comparison of the bytes. Specifications that use MOQT may
-constrain the information in these fields, for example by restricting them to
-UTF-8. Any specification that does needs to specify the canonicalization into
-the bytes in the Track Namespace or Track Name such that exact comparison works.
+exact comparison of the bytes. Specifications that use MOQT may constrain the
+information in these fields, for example by restricting them to UTF-8. Any
+specification that does needs to specify the canonicalization into the bytes in
+the Track Namespace or Track Name such that exact comparison works.
 
 ### Scope {#track-scope}
 
@@ -638,8 +638,8 @@ is used, the session is closed using the CONNECTION\_CLOSE frame
 closed using the CLOSE\_WEBTRANSPORT\_SESSION capsule ({{WebTransport,
 Section 5}}).
 
-When terminating the Transport Session, the application MAY use any error message and SHOULD use a relevant
-code, as defined below:
+When terminating the Session, the application MAY use any error message
+and SHOULD use a relevant code, as defined below:
 
 |------|---------------------------|
 | Code | Reason                    |
@@ -711,12 +711,12 @@ code, as defined below:
 * Control Message Timeout: The session was closed because the peer took too
   long to respond to a control message.
 
-* Data Stream Timeout: The session was closed because the peer took too
-  long to send data expected on an open Data Stream (see {{data-streams}}).  This
+* Data Stream Timeout: The session was closed because the peer took too long to
+  send data expected on an open Data Stream (see {{data-streams}}).  This
   includes fields of a stream header or an object header within a data
-  stream. If an endpoint times out waiting for a new object header on an
-  open subgroup stream, it MAY send a STOP_SENDING on that stream or
-  terminate the subscription.
+  stream. If an endpoint times out waiting for a new object header on an open
+  subgroup stream, it MAY send a STOP_SENDING on that stream or terminate the
+  subscription.
 
 * Auth Token Cache Overflow - the Session limit {{max-auth-token-cache-size}} of
   the size of all registered Authorization tokens has been exceeded.
@@ -995,13 +995,13 @@ subscribed Objects.
 
 ## Considerations for Setting Priorities
 
-For downstream subscriptions, relays SHOULD respect the subscriber and original publisher's priorities.
-Relays can receive subscriptions with conflicting subscriber priorities
-or Group Order preferences.  Relays SHOULD NOT directly use Subscriber Priority
-or Group Order from incoming subscriptions for upstream subscriptions. Relays'
-use of these fields for upstream subscriptions can be based on factors specific
-to it, such as the popularity of the content or policy, or relays can specify
-the same value for all upstream subscriptions.
+For downstream subscriptions, relays SHOULD respect the subscriber and original
+publisher's priorities.  Relays can receive subscriptions with conflicting
+subscriber priorities or Group Order preferences.  Relays SHOULD NOT directly
+use Subscriber Priority or Group Order from incoming subscriptions for upstream
+subscriptions. Relays' use of these fields for upstream subscriptions can be
+based on factors specific to it, such as the popularity of the content or
+policy, or relays can specify the same value for all upstream subscriptions.
 
 MoQ Sessions can span multiple namespaces, and priorities might not
 be coordinated across namespaces.  The subscriber's priority is
@@ -1190,7 +1190,8 @@ the announcement and subscription to the old relay can be stopped.
 
 MOQT encodes the delivery information via Object headers
 ({{message-object}}).  A relay MUST NOT modify Object properties when
-forwarding, except for Object Extension Headers as specified in {{object-extensions}}.
+forwarding, except for Object Extension Headers as specified in
+{{object-extensions}}.
 
 A relay MUST treat the object payload as opaque.  A relay MUST NOT
 combine, split, or otherwise modify object payloads.  A relay SHOULD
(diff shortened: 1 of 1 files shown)


yekuiwang:
@afrind @suhasHere @vasilvv  @ianswett There is one failing check. I took a look by clicking the above "[Update Editor's Copy / Update Editor's Copy (pull_request)](https://github.com/moq-wg/moq-transport/actions/runs/15219350745/job/42811900408?pr=1030)", but could not figure out what's exactly wrong. Still a lot to learn about making changes to the draft text and making PRs. Any help is greatly appreciated.

draft-ietf-moq-transport.md
```diff
@@ -1438,7 +1438,7 @@ Overflow` error.
 #### DELIVERY TIMEOUT Parameter {#delivery-timeout}
 
 The DELIVERY TIMEOUT parameter (Parameter Type 0x02) MAY appear in a
-TRACK_STATUS, SUBSCRIBE, SUBSCRIBE_OK, or a SUBSCRIBE_UDPATE message.
+TRACK_STATUS, SUBSCRIBE, SUBSCRIBE_OK or SUBSCRIBE_UDPATE message.
```
DanielFay22, line 1441: I prefer including the Oxford comma, I don't know if there's an accepted IETF styling for this.

draft-ietf-moq-transport.md
```diff
@@ -1547,9 +1547,10 @@ identified as 0xff00000D.
 
 The PATH parameter (Parameter Type 0x01) allows the client to specify the path
 of the MoQ URI when using native QUIC ({{QUIC}}).  It MUST NOT be used by
-the server, or when WebTransport is used.  If the peer receives a PATH
-parameter from the server, when WebTransport is used, or one the server
-does not support, it MUST close the session with Invalid Path.
+the server, or when WebTransport is used.  When a PATH parameter is received 
+from a server, or when a PATH parameter is received while WebTransport is used, 
+or when a PATH parameter is received by a server but the server does not 
+support the specified path, the session MUST be closed with Invalid Path.
```
DanielFay22, line 1553: I think existing wording is a little unclear, but prefer just a small change over suggested change:
```suggestion
the server, or when WebTransport is used.  If the peer receives a PATH
parameter from the server when WebTransport is used, or receives a 
PATH parameter the server does not support, it MUST close the session 
with Invalid Path.
```

draft-ietf-moq-transport.md
```diff
@@ -160,7 +160,7 @@ The following terms are used with the first letter capitalized.
 
 Application:
 
-: The entity using MoQT to transmit and receive data.
+: The entity using MOQT to transmit and receive data.
```
DanielFay22, line 163: We should decide whether `MoQT` or `MOQT` is the preferred way to write the acronym. I have been writing `MoQT` this whole time, but both are used inconsistently in the draft, so we probably just need to settle on one.

draft-ietf-moq-transport.md
```diff
@@ -86,7 +86,7 @@ discovery and subscription.
 
 * {{model}} describes the data model employed by MOQT.
 
-* {{session}} covers aspects of setting up a MOQT session.
+* {{session}} covers aspects of setting up an MOQT session.
```
DanielFay22, line 89: Use of "a" vs "an" depends on whether you read "MoQT" as "mock-T" or "M-O-Q-T". Similar to "MoQT" vs "MOQT" we should decide a consistent language for the draft and stick with it. I'm inclined to go with "an" (accept changes in this PR) but don't feel that strongly.

draft-ietf-moq-transport.md
ianswett: I don't think there is, but I think the drafts I've been on recently have also preferred the Oxford comma.

ianswett (approved):
Thanks for the PR, most of this looks good.  I think a few lines are too long now, so you might have to reflow the text.

ianswett:
Thanks for the PR, I also changed the description text slightly so when this PR lands, it'll close the editorial issues you referenced.

draft-ietf-moq-transport.md
afrind: I'm ambivalent on a vs an, other than we should be consistent.

afrind (approved):

afrind:
> afrind It seems that you have helped trimming spaces and reflowing (thanks so much), and therefore there is no further action from myself needed for this PR. Please let me know otherwise.

@yekuiwang - yes I fixed it up.  Thanks for your contribution.  I will merge this.  We can come back to a / an moqt if needed.

Issue #1028, opened 2025-05-20
Protocol needs a better way to support applications that use dynamic sized groups


suhasHere:
Currently,  if the group sizes are not static or not small sized or  not known in advance and if subscriber wants to join a mid group the only option available is to fetch the entire group from the beginning of the group 
 
 This has limitation that the data needs to available in the caches to be useful , the more longer the group duration is, the lesser chances are to find them in the cache. Otoh, if the caches are too deep, it will be hard to sync up to live edge in time, especially when dealing with realtime sessions.

afrind:
> join a mid group the only option available is to fetch the entire group from the beginning of the group

Isn't the the intentional design choice of MoQT, since it defines groups as the join points?  It sounds like the application you describe is using groups to denote something else. Can you give more detail on how such an application is leveraging groups, and why they can't be shorter to allow easier joining?

> I have a rough proposal on how to address the above issue and can propose the same

Please share here or on the list?

suhasHere:
thanks @afrind and @gloinul .. I now see my wordings were not clear. 

I was referring to conferencing use-cases where something like FIR request will trigger publishers to trigger a new i-frame . This is because in conferencing applications GOPs are not of static size and IDRs are generated only on request.  This is done so not to clobber network with IDRs, unless it is needed especially when publishing multiple streams.  

When applied within the MOQ context and when having long lived groups, subscribers joining will not always be able to get all the data from cache ( say subscriber joining around 30s  or later into a meeting). Also Subscribers can't predict when a new GOP will be generated either. 

The proposal here is to add equivalent of FIR but working for MOQ

ianswett:
The use case you're expressing is one we want to support, but I don't think it should require changes to MoQ Transport.

I believe you could create a track for only these FIR frames and use SUBSCRIBE and/or FETCH to get the most recent one when joining.  That avoids these i-frames being sent to everyone.

suhasHere:
@ianswett  I did think about using fetch, it may work just fine. 

 I had few follow up questions to get your thoughts

- The most recent FIR may not be in the cache &
- We need a way to trigger original publisher to start a new group

afrind:
> We need a way to trigger original publisher to start a new group

Can you use the PUBLISHcoming back from the relay in response to a SUBSCRIBE_ANNOUNCES as a signal?  eg:

Alice joins first:

C->R: SUBSCRIBE_ANNOUNCES(meeting id)
[...]

PR #1031, opened 2025-05-24
Add Group Sync message to trigger a new group from the Publisher

suhasHere:
Addresses #1028 

Love to gather feedback on the proposal. The proposal attempts to keep these highlevel goals in mind

 - Keep signaling minimal to trigger the action of generating the new group
 - Leverage MoQ Relay's ability to aggregate requests and reusing caching
 - Keep demand on publishers to minimum when multiple subscribers need similar functionality

This message addresses subscriber rejoins, mid group joins and conferencing feature of speaker changes which is a pretty common flow and a generic way for subscribers to trigger a new sync point.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1606,6 +1606,13 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+#### DYNAMIC GROUPS Parameter {#dynamic-groups}
+
+The DYNAMIC_GROUPS is Original Publisher only parameter (parameter type 0x20) and MAY appear
+in PUBLISH or SUBSCRIBE_OK.  Values larger than 1 are a Protocol Violation.  When the value is
+1, it indicates that the subscriber can send a NEW_GROUP messsage to request a
+new synchronization point for the track.
+
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
 
 The `CLIENT_SETUP` and `SERVER_SETUP` messages are the first messages exchanged
@@ -3121,20 +3128,24 @@ UNSUBSCRIBE_ANNOUNCES Message {
 ## NEW GROUP  {#message-new-group}
 
 Subscribers issue a NEW_GROUP message to request a new synchronization point
-(new Group) from the publisher. This message is sent after a subscription to a
+(new Group) from original publisher. This message is sent after a subscription to a
 track has been successfully established. The NEW_GROUP message identifies the
-Largest Group observed by the subscriber in the track and requests the publisher
-to publish a new Group with a GroupID greater than the one specified in the
-message.
+Largest Group observed by the subscriber in the track and requests the
+original publisher to publish a new Group with a GroupID greater than the
+one specified in the message.
 
 Relay publishers MUST aggregate NEW_GROUP messages if the Largest Group in
 the request is less than or equal to the Largest Group in the NEW_GROUP
 message issued upstream. Otherwise, the relay publisher MUST issue a new
 upstream NEW_GROUP message.
 
-Original publishers MUST create and publish a new group in the requested track
-if no Group with a GroupID greater than or equal to the one specified in the
-most recent NEW_GROUP message has been published.
+Original Publishers that supports NEW_GROUP (via DYNAMIC_GROUPS parameter) MUST
+create and publish a new group in the requested track if no Group with GroupID
+larger than or equal to the one specified in the most recent NEW_GROUP message
+has been published. Original Publisher(s) MAY delay to generate such a group
+subject to implementation specific concerns, for example, acheiving a minimum
+duration for each Group. The Original Publisher chooses the next Group ID; there
+are no requirements that it be equal to the Group ID in the NEW_GROUP message.
 
 ~~~
 NEW_GROUP Message {
@@ -3158,7 +3169,7 @@ Parameters (..) ...,
 * Largest Group: The largest observed group in the track (either in
 SUBSCRIBE_OK or TRACK_STATUS_OK or SUBSCRIBE_UPDATE_OK).
 
-* Parameters: The parameters are d
+* Parameters: The parameters as defined in {{version-specific-params}}.
 
 # Data Streams and Datagrams {#data-streams}
 


draft-ietf-moq-transport.md
```diff
@@ -2834,6 +2836,48 @@ UNSUBSCRIBE_ANNOUNCES Message {
...
 
 
+## GROUP SYNC {#message-group-sync}
+
+Subscribers issue a GROUP_SYNC message to request a new synchronization point
+(new Group) from the publisher. This message is sent after a subscription to a
+track has been successfully established. The GROUP_SYNC message identifies the
+Largest Group observed by the subscriber in the track and requests the publisher
+to publish a new Group with a GroupID greater than the one specified in the
+message.
```
gwendalsimon, line 2846: Is it a fork, i.e., the "new groupID" contains a series of new objects that are created specifically for this subscriber, in parallel of the actual track? We will need to "reconcile" (or merge) the forks eventually, otherwise we will have tons of parallel encoding and delivery processes.

It may be quite confusing for the Relay who receives, for the same track, group with groupID=n (regular subscribers), groupID=n+1 (subscribers who have issues 1 GROUP_SYNC), ... and groupID=n+X (for those who have issued X GROUP_SYNC). Should it require an additional flag in the object to allow the Relay to identify which object is the regular one (the reference for new subscribers).

Each Publisher (the original one and the Relays) must store, for each subscriber, an "individual_groupID" and an "individual_objectID" (plus some extra info probably) and may make some processing at every object to compute the actual numbers. Do we plan to define a normative behavior for the Relay?

draft-ietf-moq-transport.md
```diff
@@ -2834,6 +2836,48 @@ UNSUBSCRIBE_ANNOUNCES Message {
 * Track Namespace Prefix: As defined in {{message-subscribe-ns}}.
 
 
+## GROUP SYNC {#message-group-sync}
```
wilaw, line 2839: NEW_GROUP would be a better name than GROUP SYNC.

draft-ietf-moq-transport.md
```diff
@@ -2834,6 +2836,48 @@ UNSUBSCRIBE_ANNOUNCES Message {
...
+
+Subscribers issue a GROUP_SYNC message to request a new synchronization point
+(new Group) from the publisher. This message is sent after a subscription to a
+track has been successfully established. The GROUP_SYNC message identifies the
+Largest Group observed by the subscriber in the track and requests the publisher
+to publish a new Group with a GroupID greater than the one specified in the
```
wilaw, line 2845: > The GROUP_SYNC message identifies the
Largest Group observed by the subscriber in the track and requests the publisher
to publish a new Group with a GroupID greater than the one specified 

Is that necessary? The publisher already knows the state of the stream and in fact may know about later groups which the subscriber does not know about. It would seem sufficient that the subscriber just has to ask for a new Group to be inserted and the original publisher can do the rest.

draft-ietf-moq-transport.md
```diff
@@ -2834,6 +2836,48 @@ UNSUBSCRIBE_ANNOUNCES Message {
...
+message.
+
+Relay publishers MUST aggregate GROUP_SYNC messages if the Largest Group in
+the request is less than or equal to the Largest Group in the GROUP_SYNC
+message issued upstream. Otherwise, the relay publisher MUST issue a new
+upstream GROUP_SYNC message.
+
+Original publishers MUST create and publish a new group in the requested track
+if no Group with a GroupID greater than or equal to the one specified in the
+most recent GROUP_SYNC message has been published.
```
wilaw, line 2855: What happens if the original publisher receives 3 GROUP_SYNC  messages 10 ms apart. Is it really going to insert keyframes that close to each other? It seems there should be some language that the publisher may choose to not insert a group if it is too close to a prior group.

wilaw (commented):
I'd like the name to be more closey aligned to the action. Better candidates would be

INSERT_NEW_GROUP
NEW_GROUP_REQUEST
NEW_GROUP

I also support Alan's suggestion of perhaps just extending SUBSCRIBE_UPDATE to add this utility.

draft-ietf-moq-transport.md
suhasHere: agreed. I can update it so in my next edits

suhasHere:
> I'd like the name to be more closey aligned to the action. Better candidates would be
> 
> INSERT_NEW_GROUP NEW_GROUP_REQUEST NEW_GROUP
> 
> I also support Alan's suggestion of perhaps just extending SUBSCRIBE_UPDATE to add this utility.

I did explore the idea of extending SUBSCRIBE_UPDATE to add this functionality ,  but ended up choosing this option since the aggregation rules are different and didn't want to have 2 different rules for the same message based on the fields.  We can definitely explore that route as well and I can take a stab at redoing it

draft-ietf-moq-transport.md
```diff
@@ -2834,6 +2836,48 @@ UNSUBSCRIBE_ANNOUNCES Message {
...
+## GROUP SYNC {#message-group-sync}
+
+Subscribers issue a GROUP_SYNC message to request a new synchronization point
+(new Group) from the publisher. This message is sent after a subscription to a
+track has been successfully established. The GROUP_SYNC message identifies the
+Largest Group observed by the subscriber in the track and requests the publisher
+to publish a new Group with a GroupID greater than the one specified in the
+message.
+
+Relay publishers MUST aggregate GROUP_SYNC messages if the Largest Group in
```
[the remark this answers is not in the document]
gloinul: I assume that this is to prevent feedback implosion. So if a relay has already sent a NEW_GROUP message upstream with a largest group value equal or larger than what is in the request it will take no upstream affect. I think what might also need to be clearer is that if the relay already have started publishing on the track a group ID larger than the requested then it can also ignore this request.

draft-ietf-moq-transport.md
gloinul: Agreed this is another effect it needs to consider. The RTP FIR mechanism (RFC 5104)[https://www.rfc-editor.org/rfc/rfc5104.html#section-3.5.1] do discuss the need to supress additional request arriving shortly after having acted on a previous request. So I think some text on this is needed.

suhasHere:
I t

> I think another aspect that is missing in here is that the usage of a mechanism to trigger a new group is highly use case dependent. Please consider if this should have some capability negotiation or indication that usage of this is allowed from the publishers perspective. To me a live streaming service that runs on periodic refresh points and targets large groups will not care about NEW_GROUP messages, so there are no point in sending them. And that should possibly be made clear already when subscribing to the track/name space.

Good point Magnus. The way I was thinking is the applications know if they are using static GOPs vs Dynamic GOPs  and sending NewGroup was the former might not be much help (it is wasteful) and such applications don't send them.

draft-ietf-moq-transport.md
suhasHere: Agreed.  We can add a recommendation to that effect

draft-ietf-moq-transport.md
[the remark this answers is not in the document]
suhasHere: Yes, the main idea is to avoid IDR flooding.   My deck below does cover the flow. capturing above recommendations and agree we should clarify that.

https://docs.google.com/presentation/d/1TkfMi22rfUbiNuAg7xrKDteZ7ZUP6OGzmjLwqZR-pPk/edit?usp=sharing does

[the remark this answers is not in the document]
suhasHere: > Can you elaborate what you mean by aggregate here? Do you mean ignore?

yes ignore for the current inflight requests.

draft-ietf-moq-transport.md
```diff
@@ -2834,6 +2836,48 @@ UNSUBSCRIBE_ANNOUNCES Message {
...
+
+* Request ID: See {{request-id}}.
+
+* Track Namespace: Identifies a track's namespace as defined in
+({{track-name}})
+
+* Track Name: Identifies the track name as defined in ({{track-name}}).
+
+* Largest Group: The largest observed group in the track (either in
+SUBSCRIBE_OK or TRACK_STATUS_OK or SUBSCRIBE_UPDATE_OK).
```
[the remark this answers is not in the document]
suhasHere: good point. I don't see why it cannot happen. I can clarify that.

draft-ietf-moq-transport.md
```diff
@@ -2834,6 +2836,48 @@ UNSUBSCRIBE_ANNOUNCES Message {
 * Track Namespace Prefix: As defined in {{message-subscribe-ns}}.
 
 
+## GROUP SYNC {#message-group-sync}
+
+Subscribers issue a GROUP_SYNC message to request a new synchronization point
```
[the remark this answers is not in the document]
suhasHere: @gwendalsimon  that's right.

draft-ietf-moq-transport.md
suhasHere: > Is it a fork, i.e., the "new groupID" contains a series of new objects that are created specifically for this subscriber, in parallel of the actual track?

Its not a fork. It is the Original Publisher starting a new group for the track and all the subscribers see the effect.

draft-ietf-moq-transport.md
suhasHere: @wilaw having the group included will allow relays to decide when not to send upstream requests.

fluffy:
Suggestion in WG call to think about:

Could this be done in an subscribe update
Do in different level

draft-ietf-moq-transport.md
```diff
@@ -3116,6 +3125,52 @@ UNSUBSCRIBE_ANNOUNCES Message {
...
 
 
+## NEW GROUP  {#message-new-group}
+
+Subscribers issue a NEW_GROUP message to request a new synchronization point
+(new Group) from original publisher. This message is sent after a subscription to a
+track has been successfully established. The NEW_GROUP message identifies the
+Largest Group observed by the subscriber in the track and requests the
+original publisher to publish a new Group with a GroupID greater than the
+one specified in the message.
```
afrind, line 3135: What value do you send in NEW_GROUP if you have observed no objects: eg a track with content-exists=0.

draft-ietf-moq-transport.md
```diff
@@ -3116,6 +3125,52 @@ UNSUBSCRIBE_ANNOUNCES Message {
 * Track Namespace Prefix: As defined in {{message-subscribe-ns}}.
 
 
+## NEW GROUP  {#message-new-group}
+
+Subscribers issue a NEW_GROUP message to request a new synchronization point
+(new Group) from original publisher. This message is sent after a subscription to a
+track has been successfully established. The NEW_GROUP message identifies the
```
afrind, line 3132: > successfully established

What if you want to indicate to the publisher a new group request along with the initial subscribe (eg pipelined)?

draft-ietf-moq-transport.md
```diff
@@ -3116,6 +3125,52 @@ UNSUBSCRIBE_ANNOUNCES Message {
...
+Largest Group observed by the subscriber in the track and requests the
+original publisher to publish a new Group with a GroupID greater than the
+one specified in the message.
+
+Relay publishers MUST aggregate NEW_GROUP messages if the Largest Group in
```
afrind, line 3137: > aggregate

I don't think aggregate is totally clear, maybe discard or ignore?

draft-ietf-moq-transport.md
```diff
@@ -3116,6 +3125,52 @@ UNSUBSCRIBE_ANNOUNCES Message {
...
+the request is less than or equal to the Largest Group in the NEW_GROUP
+message issued upstream. Otherwise, the relay publisher MUST issue a new
```
afrind, line 3139: > the Largest Group in the request is less than or equal to the Largest Group in the NEW_GROUP message issued upstream. 

I think I know what you mean but the text is unclear.  I think you want to condition the behavior on the relay having one (or more?) outstanding NEW_GROUP to the relay - something like:

If the relay has an outstanding NEW_GROUP request upstream it MUST ignore any incoming NEW_GROUP request with a value less than or equal to the outstanding request's value.

draft-ietf-moq-transport.md
```diff
@@ -3116,6 +3125,52 @@ UNSUBSCRIBE_ANNOUNCES Message {
...
+upstream NEW_GROUP message.
+
+Original Publishers that supports NEW_GROUP (via DYNAMIC_GROUPS parameter) MUST
+create and publish a new group in the requested track if no Group with GroupID
+larger than or equal to the one specified in the most recent NEW_GROUP message
```
afrind, line 3144: What is the use case for allowing more than one request at a time in the same session?  The publisher is likely to process requests as they arrive, rather than parse a bunch and mint the new group with the largest ID.

draft-ietf-moq-transport.md
suhasHere: One typically wouldn't send a new_group if track has not started. 

OTOH , if any new_group arrives at the relay for track that hasn't started , it will be a no-op ?

afrind:
In the virtual interim, folks agreed that #1060 was preferable to this, though we are also exploring a generic "FEEDBACK" message that would look like somewhat like this message.

PR #1032, opened 2025-05-26
List what parameters can be repeated in IANA table

### LINKED ISSUE #1004: Clearly indicate which parameters can have duplicates
We have a bunch of parameters like MAX_CACHE_DURATION for which we don't define if duplicates are allowed or not.

Maybe the IANA section for that should have a "duplicates allowed" for each of those.

ianswett:
When we make an IANA registry.

Fixes #1004

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3426,7 +3426,7 @@ TODO: fill out currently missing registries:
 
 * MOQT version numbers
 * Setup parameters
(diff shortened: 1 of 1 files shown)


afrind:
Is this really fixing 1004 just by adding a TODO?

ianswett:
It is clear which parameters can be repeated, so I think the only action to take for right now is to add a TODO?

afrind (approved):

PR #1033, opened 2025-05-26
Object extension headers and Datagrams

### LINKED ISSUE #785: How to learn that there are an MTU issue for Datagram
The MTU available for the Datagram is depending on the path MTU for the underlying QUIC connection between the subscriber and publisher. So how does the subscriber determine that missing objects are due to the publisher being unable to send the packets, or the packet being lost in an MTU black hole on path? Assuming that one have reasonable well working PMTUD on the QUIC connection level, the publisher will determine that it is unable to send oversized datagrams. But that information appears to be stuck on the publisher side, and no way of giving it to the subscriber. So, does MoQT need some type of error indication or high level reporting. 

There is likely some relation to #719 which could be one reasons for pushing datagrams into oversized.

ianswett:
Describe what might happen without any normative language.

Fixes #785

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2995,11 +2995,18 @@ choose to buffer it for a brief period to handle reordering with the control
 message that establishes the Track Alias.
 
 An Object received in an `OBJECT_DATAGRAM` or `OBJECT_DATAGRAM_STATUS` message
-has an `Object Forwarding Preference` = `Datagram`. To send an Object with
-`Object Forwarding Preference` = `Datagram`, determine the length of the header
-and payload and send the Object as datagram. In certain scenarios where the
-object size can be larger than maximum datagram size for the session, the Object
-will be dropped.
+has an `Object Forwarding Preference` = `Datagram`.
+
+To send an Object with `Object Forwarding Preference` = `Datagram`, determine
+the length of the header and payload and send the Object as datagram.  When the
+total size is larger than maximum datagram size for the session, the Object will
+be dropped without any explicit notification.
+
+Each session along the path between the Original Publisher and End Subscriber
+might have different maximum datagram sizes. Additionally, object extensions
+({{object-extensions}}) can be added to Objects as they pass through the MOQT
+network, increasing the size of th Object and the chances it will exceed the
+maximum datagram size of a downstream session and be dropped.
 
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2995,11 +2995,18 @@ choose to buffer it for a brief period to handle reordering with the control
...
+
+To send an Object with `Object Forwarding Preference` = `Datagram`, determine
+the length of the header and payload and send the Object as datagram.  When the
+total size is larger than maximum datagram size for the session, the Object will
+be dropped without any explicit notification.
+
+Each session along the path between the Original Publisher and End Subscriber
+might have different maximum datagram sizes. Additionally, object extensions
+({{object-extensions}}) can be added to Objects as they pass through the MOQT
+network, increasing the size of th Object and the chances it will exceed the
```
ianswett, line 3008:
```suggestion
network, increasing the size of the Object and the chances it will exceed the
```

draft-ietf-moq-transport.md
```diff
@@ -3003,11 +3003,18 @@ choose to buffer it for a brief period to handle reordering with the control
...
-will be dropped.
+has an `Object Forwarding Preference` = `Datagram`.
+
+To send an Object with `Object Forwarding Preference` = `Datagram`, determine
+the length of the header and payload and send the Object as datagram.  When the
+total size is larger than maximum datagram size for the session, the Object will
+be dropped without any explicit notification.
+
+Each session along the path between the Original Publisher and End Subscriber
+might have different maximum datagram sizes. Additionally, object extensions
```
afrind, line 3014: Aren't they "Object Extension Headers" ?

draft-ietf-moq-transport.md
```diff
@@ -2995,11 +2995,18 @@ choose to buffer it for a brief period to handle reordering with the control
...
+({{object-extensions}}) can be added to Objects as they pass through the MOQT
+network, increasing the size of the Object and the chances it will exceed the
+maximum datagram size of a downstream session and be dropped.
```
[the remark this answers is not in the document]
afrind: "SHOULD consider" and "MAY choose not to" are very weak normative statements and don't add value.  SHOULD is already kind of weak when it's "SHOULD take specific action A".

I think explaining there be dragons if relays add extensions to datagrams is sufficient.

PR #1042, opened 2025-06-17
Delta encode Object IDs within Subgroups

### LINKED ISSUE #1029: Delta encoding Object ID would prevent malformed tracks
Today Object ID is not delta encoded, allowing malformed tracks, see #938 

If the Object ID was delta encoded, it could both save bytes and prevent malformed tracks.

We discussed this before, but it seems like we're in the mode of optimizing the wire format, so maybe now is the time to make this change.

ianswett:
We've being saying we'd do this for a while, so here it is.  Let me know if you'd like me to delta encode Object IDs in FETCH responses as well?

Fixes #1029

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -519,32 +519,30 @@ to MoQT constraints. Such a Track is considered malformed.  Some example
 conditions that constitute a malformed track when detected by a receiver
 include:
 
-1. An Object is received on a Subgroup stream whose Object ID is not strictly
-   larger than the previous Object received on the same Subgroup.
-2. An Object is received in a FETCH response with the same Group as the
+1. An Object is received in a FETCH response with the same Group as the
    previous Object, but whose Object ID is not strictly larger than the previous
    object.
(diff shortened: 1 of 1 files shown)


afrind:
FETCH has a proposed scheme in #949 which includes delta compressing group and subgroup as well, which aren't needed in SUBGROUP.

draft-ietf-moq-transport.md
```diff
@@ -3463,9 +3461,14 @@ following fields.
 
 The Object Status field is only sent if the Object Payload Length is zero.
 
+The Object ID Delta + 1 is added to the previous Object ID in the Subgroup
+if there was one.  The Object ID is the Object ID Delta if it's the first
```
afrind: Maybe Subgroup _stream_?  It's possible the first object in a Subgroup stream is not the first object in the subgroup.

ianswett: Good point, updated to Subgroup stream

fluffy (approved):
LGTM

afrind (approved):

suhasHere (approved):

vasilvv (approved):

PR #1043, opened 2025-06-17
Make GROUP_ORDER a Parameter

### LINKED ISSUE #878: Make SUBSCRIBE group order into an optional parameter
Reasons:

- There is a reasonable default ("use publisher priority").
- We have situations in the draft where allowed values are `{asc, desc, default}`, and situation where those are only `{asc, desc}` (SUBSCRIBE_OK).  It would be nice to reduce to one set.

ianswett:
Fixes #878

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1613,8 +1613,8 @@ by group. See {{priorities}}. When sent by the publisher, it indicates the
 publisher's preferred order. Values larger than 0x1 are a protocol error.
 
 Group order of a subscription MUST NOT be changed, so it is invalid to use
-GROUP ORDER in SUBSCRIBE_UPDATE or PUBLISH_OK. GROUP ORDER is valid in SUBSCRIBE,
-SUBSCRIBE_OK, PUBLISH, and FETCH.
+GROUP ORDER in SUBSCRIBE_UPDATE, PUBLISH_OK, or FETCH_OK.
+GROUP ORDER is valid in SUBSCRIBE, SUBSCRIBE_OK, PUBLISH, and FETCH.
 
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1614,7 +1614,7 @@ publisher's preferred order. Values larger than 0x1 are a protocol error.
 
 Group order of a subscription MUST NOT be changed, so it is invalid to use
 GROUP ORDER in SUBSCRIBE_UPDATE or PUBLISH_OK. GROUP ORDER is valid in SUBSCRIBE,
-SUBSCRIBE_OK, PUBLISH, FETCH, and FETCH_OK.
+SUBSCRIBE_OK, PUBLISH, and FETCH.
 
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1042,9 +1042,10 @@ _Group Order_ is a property of an individual subscription.  It can be either
 'Ascending' (groups with lower group ID are sent first), or 'Descending'
 (groups with higher group ID are sent first).  The subscriber optionally
 communicates its group order preference in the SUBSCRIBE message; the
-publisher's preference is used if the subscriber did not express one (by
-setting Group Order field to value 0x0).  The group order of an existing
-subscription cannot be changed.
+publisher's preference is used if the subscriber did not specify the
+'Group Order' param. If the publisher also does not specify the group order,
+it defaults to ascending. The group order of an existing subscription cannot
+be changed.
```
afrind, line 1048:
```suggestion
it defaults to ascending.
```

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1605,18 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+#### GROUP ORDER Parameter
+
+The GROUP ORDER parameter (Parameter Type 0x06) allows the subscriber to
+request Objects be delivered in Ascending (0x0) or Descending (0x1) order
+by group. See {{priorities}}. When sent by the publisher, it indicates the
+publisher's preferred order. A value of 0x0 indicates the original Values larger
+than 0x1 are a protocol error.
```
```suggestion
publisher's preferred order. Values larger than 0x1 are a protocol error.
```

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1605,18 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+
+Group order of a subscription MUST NOT be changed, so it is invalid to use
```
afrind, line 1615: 1) Group Order is just part of the priority algorithm, so I don't see why it can't be changed - just allow it in SUBSCRIBE_UPDATE.  I guess the publisher can never change it.
2) Even if it can't be changed, it has to be allowed in PUBLISH_OK because that's how the subscriber gets input.
3) I don't think it should be allowed in FETCH_OK - only the subscriber gets a preference there.

draft-ietf-moq-transport.md
```diff
@@ -2491,11 +2481,6 @@ Fields common to all Fetch Types:
 other subscriptions or fetches in the same session. Lower numbers get higher
 priority. See {{priorities}}.
 
-* Group Order: Allows the subscriber to request Objects be delivered in
-Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
-A value of 0x0 indicates the original publisher's Group Order SHOULD be
```
afrind, line 2496: This never made any sense?  I think the subscriber should choose and the publisher should just do what the subscriber asked for.  In particular, you can't process the fetch response stream for gaps without knowing the order first.

ianswett: Are you saying that FETCH can indicate it and FETCH_OK can't change it?  If so, that makes a lot of sense.  Or are you saying that FETCH shouldn't allow specifying it?

draft-ietf-moq-transport.md
ianswett: I removed it from FETCH_OK

ianswett: I filed an issue for whether Group Order should be updatable, #1045

vasilvv (approved):

afrind:
A similar change is now in #1210, though that one does not change FETCH.

Issue #879, opened 2025-03-27
Make SUBSCRIBE filters into optional parameters


vasilvv:
Currently, SUBSCRIBE has mandatory fields specifying the parts of the track that the subscriber asks to subscribe to.  This made a lot of sense back when SUBSCRIBE could go back in the past, but I feel like this is not really consistent with our new definition that "SUBSCRIBE gives you newly published objects".

Proposal:
- By default, SUBSCRIBE is unfiltered.
- We define parameters that correspond to the currently defined filters.
- If multiple filters are specified, the object has to match all of them to be forwarded.

This naturally solves the extensibility problem (#790).

suhasHere:
I am bit unsure if this complexity is needed, esp , having to provide multiple filters. It makes API hard to reason about. Unless there is a strong use-case to have that flexibility, my preference is to go with current api format.

vasilvv:
I am not sure what you mean by complexity, since the version I have proposed seems to me less complex than what we currently have.

afrind:
Here's the current set of SUBSCRIBE params that could be optional (Full Track Name is obviously required):

```
  Subscriber Priority (8),
  Group Order (8),
  Filter Type (i),
  [Start (Location)],
  [EndGroup (i)],
```

Maybe the proposal is adding the following parameters

PRIORITY (default to 128 if not present)
GROUP_ORDER (default to ASC if not present)
START_LOCATION (defaults to the next object larger than current largest, or 0,0 if no-content)
END_GROUP (default to open ended)

I would say that it's a Session Error if any appear more than once, to avoid nonsense.  Regarding #758, I think setting START_LOCATION={0,0} is the same as #853 's All Objects, right?  

This would allow END_GROUP without START_LOCATION, which is currently not allowed.  It might be empty, but it's unambiguous.

suhasHere:
> I am not sure what you mean by complexity, since the version I have proposed seems to me less complex than what we currently have.

couple of observations. It is not less complex since either you pick the things from fields or from a param structure , you need to do the work regardless. Also adding multiple filters leads to complicated logic to root out invalid combinations and is more work than needed. I would love to understand the use-case driving this feature. May be I am missing something

afrind:
>  Also adding multiple filters leads to complicated logic to root out invalid combinations

See my proposal above, I think we can avoid any illegal combinations.  It's not that there's actually "multiple filters".

> I would love to understand the use-case driving this feature.

[...]

Issue #872, opened 2025-03-27
What if the subscriber needs an immediate SUBSCRIBE_UPDATE?


martinduke:
I cannot send SUBSCRIBE_UPDATE for a LatestObject subscription if I haven't received SUBSCRIBE_OK because I don't know what to put in the Start field to avoid a protocol error. It's possible to immediately send UNSUBSCRIBE and then another SUBSCRIBE if there are enough IDs, I suppose.

The most likely reason is a sudden change in priority. Waiting an RTT for this seems wrong, especially if the message goes all the way back to the original publisher.

Relatedly, this is super-pedantic, but it is legal to send a SUBSCRIBE with AbsoluteRange and EndGroup 2^62-1, even though it is equivalent to AbsoluteStart with no end. It is impossible to update other properties of this subscription without accidentally shortening it, unless the SUBSCRIBE_UPDATE EndGroup field is zero. Code should be smart enough to detect the case and not report it as a widening of the window.

There are many spelling changes that can fix this, but one clean way would be to make Start and EndGroup be diff encoded. This also eliminates most error cases -- no way to expand the window.

afrind:
Can we make all the fields of SUBSCRIBE_UPDATE into parameters, so you can update any of them independently?

martinduke:
> Can we make all the fields of SUBSCRIBE_UPDATE into parameters, so you can update any of them independently?

I like that solution, although it is, editorially, a challenge -- these properties would be parameters in SUBSCRIBE_UPDATE but message fields in SUBSCRIBE.

I'm also fine with diff encoding. What's positive about that is it makes it impossible to express widening of the window.

ianswett:
If we change SUBSCRIBE so the start Location is a filter that doesn't change in SUBSCRIBE_OK, then I believe this raciness is removed?

martinduke:
> If we change SUBSCRIBE so the start Location is a filter that doesn't change in SUBSCRIBE_OK, then I believe this raciness is removed?

I think so, though I would need to see a fully fleshed out proposal to see that it doesn't open other problems.

ianswett:
There is an interest at the interim in ensuring that SUBSCRIBE_UPDATE can be sent before SUBSCRIBE_OK has been received.

Key use cases are updating the priority or start location.

afrind:
Seems like we can either make filters into parameters #879 or add a filter type to SUBSCRIBE_UPDATE indicating no change to filters.

afrind:
[...]

PR #1044, opened 2025-06-17
Make SUBSCRIBER PRIORITY a Parameter

ianswett:
Note that this would be a convenient time to rename this to PRIORITY and let either peer specify it.

Fixes part of #879
Related to #872

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1608,7 +1608,8 @@ re-requests them.
 
 The SUBSCRIBER PRIORITY parameter (Parameter Type 0x08) specifies the priority
 of a subscription relative to other subscriptions in the same session.
-Lower numbers get higher priority. See {{priorities}}. The SUBSCRIBER PRIORITY
+The value is from 0 to 255 and lower numbers get higher priority.
+See {{priorities}}. Priorities above 255 are invalid. The SUBSCRIBER PRIORITY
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,13 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+#### SUBSCRIBER PRIORITY Parameter {#subscriber-priority}
+
+The SUBSCRIBER PRIORITY parameter (Parameter Type 0x08) specifies the priority
```
afrind, line 1610: We need to constrain this to 0..255 to match the old behavior, this is now encoded as a varint.

afrind (commented):
I'd be in favor of (eventually) renaming to just PRIORITY, allowing the publisher to set it in PUBLISH and SUBSCRIBE_OK, and a Subgroup/Datagram encoding that omits `publisher priority` in the data-plane, instead using the value from the control plane, or the default.

In cases where the track never changes priority, this saves 1 byte per datagram or subgroup.

vasilvv (changes requested):
This needs a default subscriber priority.

ianswett:
> This needs a default subscriber priority.

Good point, I added one.

afrind:
A similar change is now part of #1210.

PR #1046, opened 2025-06-18
Make FORWARD a parameter

ianswett:
Fixes part of #879

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1611,7 +1611,7 @@ subscription will be forwarded to the subscriber. If 0, Objects are not forwarde
 to the subscriber.  The default is 1 and values of 1 indicate Objects will be
 forwarded. Values larger than 0x01 are a protocol error and the session
 MUST be terminated with a Protocol Violation ({{session-termination}}).
-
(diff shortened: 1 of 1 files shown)


vasilvv (approved):

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,15 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+#### FORWARD
+
+The FORWARD parameter (Parameter Type 0x0A) indicates if Objects matching the
```
afrind, line 1609: You need to say which messages support it - SUBSCRIBE, SUBSCRIBE_UPDATE, PUBLISH, PUBLISH_OK?

suhasHere (approved):

afrind:
During the 8/26 interim, @fluffy pointed out that there is no "common" value of Forward, so it might make more sense to stay as a fixed field rather than a param.

afrind:
I think we should redo this PR since it's mostly small and I don't want to rebase it

PR #1047, opened 2025-06-18
Parameters on SUBSCRIBE_ANNOUNCES are subscription defaults

ianswett:
This was discussed in Stockholm.  I believe it fixes an issue, but I can't quickly find it.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2985,10 +2985,10 @@ an N of 0 or more than 32, it MUST close the session with a Protocol
 Violation.
 
 * Parameters: The parameters are defined in {{version-specific-params}}. Any
-parameter that is applicable to a subscription, such as DELIVERY TIMEOUT,
-SHOULD be applied to any subscriptions that result from the SUBSCRIBE_ANNOUNCES.
-This enables a user of SUBSCRIBE_ANNOUNCES to specify defaults for any
-resulting subscriptions.
+  parameter that is applicable to a subscription, such as DELIVERY TIMEOUT,
+  SHOULD be applied to any subscriptions that result from the
+  SUBSCRIBE_ANNOUNCES.  This enables a user of SUBSCRIBE_ANNOUNCES to specify
+  defaults for any resulting subscriptions.
 
 The publisher will respond with SUBSCRIBE_ANNOUNCES_OK or
 SUBSCRIBE_ANNOUNCES_ERROR.  If the SUBSCRIBE_ANNOUNCES is successful, the
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2984,7 +2984,11 @@ would match both.  If an endpoint receives a Track Namespace Prefix tuple with
 an N of 0 or more than 32, it MUST close the session with a Protocol
 Violation.
 
-* Parameters: The parameters are defined in {{version-specific-params}}.
+* Parameters: The parameters are defined in {{version-specific-params}}. Any
+parameter that is applicable to a subscription, such as DELIVERY TIMEOUT,
+SHOULD be applied to any subscriptions that result from the SUBSCRIBE_ANNOUNCES.
+This enables a user of SUBSCRIBE_ANNOUNCES to specify defaults for any
+resulting subscriptions.
 
 The publisher will respond with SUBSCRIBE_ANNOUNCES_OK or
 SUBSCRIBE_ANNOUNCES_ERROR.  If the SUBSCRIBE_ANNOUNCES is successful, the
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2984,7 +2984,11 @@ would match both.  If an endpoint receives a Track Namespace Prefix tuple with
 an N of 0 or more than 32, it MUST close the session with a Protocol
 Violation.
 
-* Parameters: The parameters are defined in {{version-specific-params}}.
+* Parameters: The parameters are defined in {{version-specific-params}}. Any
+parameter that is applicable to a subscription, such as DELIVERY TIMEOUT,
+SHOULD be applied to any subscriptions that result from the SUBSCRIBE_ANNOUNCES.
+This enables a user of SUBSCRIBE_ANNOUNCES to specify defaults for any
+resulting subscriptions.
```
```suggestion
* Parameters: The parameters are defined in {{version-specific-params}}. Any
  parameter that is applicable to a subscription, such as DELIVERY TIMEOUT,
  SHOULD be applied to any subscriptions that result from the
  SUBSCRIBE_ANNOUNCES.  This enables a user of SUBSCRIBE_ANNOUNCES to specify
  defaults for any resulting subscriptions.
```

afrind: Should we explain this in more detail or be more specific?  The flow is:

SUBSCRIBE_ANNOUNCES (with params)
...later...
PUBLISH <- Should assume SUBSCRIBE_ANNOUNCES params as subscriber inputs.  The Subscriber gets another chance to override in PUBLISH_OK.

ianswett: Yes, I think being more specific would be good.  Are we sure it's always clear what SUBSCRIBE_ANNOUNCES caused the PUBLISH?  It seems safer to have the PUBLISH repeat them anyway?

afrind: Yes, the PUBLISH should specify any parameters the publisher wanted mixed with the ones in a relevant SUBSCRIBE_ANNOUNCES, if present.

ianswett: SG, I updated the PR, PTAL

draft-ietf-moq-transport.md
```diff
@@ -2984,7 +2984,15 @@ would match both.  If an endpoint receives a Track Namespace Prefix tuple with
 an N of 0 or more than 32, it MUST close the session with a Protocol
 Violation.
 
-* Parameters: The parameters are defined in {{version-specific-params}}.
+* Parameters: The parameters are defined in {{version-specific-params}}. Any
+  parameter that is applicable to a subscription, such as DELIVERY TIMEOUT,
+  SHOULD be applied to any subscriptions that result from the
+  SUBSCRIBE_ANNOUNCES.  This enables a user of SUBSCRIBE_ANNOUNCES to specify
+  defaults for any resulting subscriptions. When a PUBLISH is caused by a
+  SUBSCRIBE_ANNOUNCES the publisher SHOULD copy any parameters that are
```
afrind, line 2992: I'm not sure that _copy_ is the right verb.  Some of the parameters in PUBLISH are the _publisher's_ preferences (eg: delivery timeout), and the publisher already takes the min.  However, for FORWARD or GROUP ORDER, it makes sense to use the value in SUBSCRIBE_ANNOUNCES.

afrind:
If different SUBSCRIBE_ANNOUNCES have different defaults, do we need to remove the restriction of no overlapping SUBSCRIBE_ANNOUNCES?  I'd rather _not_, since that means there could be conflicting defaults.

fluffy (commented):
This seems both too clever and too vague. I just see lots of surprises coming out of this and I'm not seeing the use case that would benefit from it. There some usage I am missing ?

afrind:
> This seems both too clever and too vague. I just see lots of surprises coming out of this and I'm not seeing the use case that would benefit from it. There some usage I am missing ?

The use case is hinting to the relay or publisher what the settings should be in matching publishes -- eg: maybe I am a Track Discovery service so always want Forward=0.  Otherwise, the publisher guesses and the subscriber overrides in SUBSCRIBE_OK, possibly getting some things it doesn't want.

draft-ietf-moq-transport.md
fluffy, line 2992: To avoid confusion, I think it would be better to explicitly which the things that are copied over.

afrind: Perhaps another place where we could have a single parameter with the encoded block of default publish parameters?

ianswett:
I'm inclined to close this for now in favor of #1220, which adds Forward to SUBSCRIBE_ANNOUNCES

ianswett:
Closing because I think this is no longer the direction we want to go in and other PRs have overtaken it.

PR #1052, opened 2025-06-20
Make Largest Location a Param

ianswett:
It's already optional, so this is simpler.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1604,6 +1604,13 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+#### LARGEST LOCATION Parameter {#largest-location}
+
+The LARGEST LOCATION parameter (Parameter type 0x05) is the location of the
+largest object available for this track encoded as a Location.
+This parameter is valid in SUBSCRIBE_OK, PUBLISH, TRACK_STATUS_OK, and
+FETCH_OK.
+
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
 
 The `CLIENT_SETUP` and `SERVER_SETUP` messages are the first messages exchanged
@@ -1945,8 +1952,6 @@ SUBSCRIBE_OK Message {
   Track Alias (i),
   Expires (i),
   Group Order (8),
-  Content Exists (8),
-  [Largest Location (Location)],
   Number of Parameters (i),
   Subscribe Parameters (..) ...
 }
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,13 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+#### LARGEST LOCATION Parameter {#largest-location}
+
+The LARGEST LOCATION parameter (Parameter type 0x05) is the location of the
+largest object available for this track encoded as a Location.
+This parameter is valid in SUBSCRIBE_OK, PUBLISH, TRACK_STATUS_OK, and
```
afrind: TRACK_STATUS_OK isn't a thing (yet).  It's just TRACK_STATUS.

afrind (commented):
Was it not in PUBLISH already?  I didn't see you remove it there.

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,13 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+FETCH_OK.
+
```
suhasHere, line 1615: I think this is missing the rules on when a sending endpoint should be setting this parameter. I think we need to bring the text similar to the effect of , if track is live ( content being published) , then .... , else ...

draft-ietf-moq-transport.md
```suggestion
This parameter is valid in SUBSCRIBE_OK, PUBLISH, TRACK_STATUS, and
```

draft-ietf-moq-transport.md
ianswett: Thanks, updated.  PTAL

ianswett:
> Was it not in PUBLISH already? I didn't see you remove it there.

It was named 'Largest', so I missed it.  I've now removed it.

draft-ietf-moq-transport.md
```diff
@@ -2780,20 +2770,20 @@ TRACK_STATUS Message {
...
+0x01: The track does not exist. The Largest Location MUST be zero, and any other
 value is a malformed message.
 
-0x02: The track has not yet begun. Subsequent fields MUST be zero. Any other
+0x02: The track has not yet begun. The Largest Location MUST be zero. Any other
 value is a malformed message.
 
-0x03: The track has finished, so there is no "live edge." Subsequent fields
-contain the highest Group and object ID known.
+0x03: The track has finished, so there is no "live edge." Largst Location
```
```suggestion
0x03: The track has finished, so there is no "live edge." Largest Location
```

suhasHere (approved):

afrind:
Going to redo this also after #1273 lands

Issue #867, opened 2025-03-25
Flow Control and Priorities


martinduke:
The priority algorithm in 6.2 may be suboptimal given certain receiver flow control states. For example:

- Streams might have different flow control credit, allowing send on a lower priority stream but not a higher one. This is most likely to happen if a frame granting stream credit is lost, or if the available data in a subgroup exceeds the initial flow control credit.

- Connection flow control might shut down delivery over streams, but cannot limit sending of datagrams.

- There may be no stream creation credit for a high-priority stream, but we can keep delivering on the low-priority one.

If we actually want low-priority stuff to be blocked in this case, then the current language is fine. Otherwise, this has to be rewritten a bit. It may be easiest to update to the definition of schedulable object in 6.1, and then the rest would be unchanged.

afrind:
> If we actually want low-priority stuff to be blocked in this case

This is the important question: if I have a high pri stream that's FC blocked and a low pri one that is not, do I send *nothing*?

I've also wondered if we need a moq specific level of flow control that is per subscription.  Per-stream is too fine and Per-connection is too coarse.

You essentially never want the sender to run out of stream credit, unless you can partition it per subscription.

fluffy:
I would rather not change any of this unless we had data to show it is a problem.

ianswett:
I'd be happy to add some text documenting this issue, but I don't think it's a 'problem' perse.

I think we might need subscription level flow control of some sort, but I'm happy to wait until we understand the problem we're solving better.

vasilvv:
> This is the important question: if I have a high pri stream that's FC blocked and a low pri one that is not, do I send nothing?

That's a good question, but I feel like this is more of a QUIC than an MoQ question.

From what I remember, our implementation uses priorities across the streams that are not flow-control blocked.

afrind:
> From what I remember, our implementation uses priorities across the streams that are not flow-control blocked.

So does ours, and until Martin asked, I never would have thought about blocking low-pri streams when high-pri streams are flow-control blocked.

> I would rather not change any of this unless we had data to show it is a problem.

The purpose of all the text in the priority section was to create predictability/interoperability for applications running through relays, and since there's ambiguity here, we should clarify the expectations.

LPardue:
[...]

PR #1054, opened 2025-06-23
Objects are not schedulable if flow-control blocked

afrind:
I intended this text to cover both 

1) stream cannot be opened
2) object doesn't fit in stream limits

Fixes: #867

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1015,8 +1015,8 @@ A schedulable object in MOQT is either:
 3. An Object in response to a FETCH where that Object is the next
    Object in the response.
 
-An Object is not schedulable if it cannot be written due to underlying
-transport flow control limits.
+An Object is not schedulable if it is known that no part of it can be written
+due to underlying transport flow control limits.
 
 A single subgroup or datagram has a single publisher priority. Within a
 response to SUBSCRIBE, it can be useful to conceptualize this process as
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1009,12 +1009,15 @@ congestion.
 MOQT maintains priorities between different _schedulable objects_.
 A schedulable object in MOQT is either:
 
-1. An object in response to a SUBSCRIBE that belongs to a subgroup where
-   that object is the next object in that subgroup.
-2. An object in response to a SUBSCRIBE that belongs to a track with
+1. An Object in response to a SUBSCRIBE that belongs to a Subgroup where
+   that Object is the next Object in that Subgroup.
+2. An Object in response to a SUBSCRIBE that belongs to a Track with
    delivery preference Datagram.
-3. An object in response to a FETCH where that object is the next
-   object in the response.
+3. An Object in response to a FETCH where that Object is the next
+   Object in the response.
+
+An Object is not schedulable if it cannot be written due to underlying
+transport flow control limits.
 
 A single subgroup or datagram has a single publisher priority. Within a
 response to SUBSCRIBE, it can be useful to conceptualize this process as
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1009,12 +1009,15 @@ congestion.
 MOQT maintains priorities between different _schedulable objects_.
 A schedulable object in MOQT is either:
 
-1. An object in response to a SUBSCRIBE that belongs to a subgroup where
-   that object is the next object in that subgroup.
-2. An object in response to a SUBSCRIBE that belongs to a track with
+1. An Object in response to a SUBSCRIBE that belongs to a Subgroup where
+   that Object is the next Object in that Subgroup.
+2. An Object in response to a SUBSCRIBE that belongs to a Track with
```
```suggestion
2. An Object in response to a subscription that belongs to a Track with
```

draft-ietf-moq-transport.md
```diff
@@ -1009,12 +1009,15 @@ congestion.
 MOQT maintains priorities between different _schedulable objects_.
 A schedulable object in MOQT is either:
 
-1. An object in response to a SUBSCRIBE that belongs to a subgroup where
-   that object is the next object in that subgroup.
-2. An object in response to a SUBSCRIBE that belongs to a track with
+1. An Object in response to a SUBSCRIBE that belongs to a Subgroup where
+   that Object is the next Object in that Subgroup.
```
```suggestion
1. The first or next Object in a Subgroup that is in response to a subscription.
```

ianswett (approved):

sharmafb (approved):

draft-ietf-moq-transport.md
```diff
@@ -1009,12 +1009,14 @@ congestion.
...
+1. The first or next Object in a Subgroup that is in response to a subscription.
+2. An Object in response to a subscription that belongs to a Track with
    delivery preference Datagram.
-3. An object in response to a FETCH where that object is the next
-   object in the response.
+3. An Object in response to a FETCH where that Object is the next
+   Object in the response.
+
+An Object is not schedulable if it cannot be written due to underlying
+transport flow control limits.
```
mengelbart: From Slack: I don't think the QUIC implementation I'm using exposes this information to my application. @afrind  suggested to add "if known", which is fine for me, but I think makes it somewhat useless to add the restriction in the first place?

Another question that came up is whether we need to add "cannot be written _entirely_". I'm not sure if that is intended here. @afrind says no, because we don't want relays to buffer large objects. Also, I guess it would allow small objects to indefinitely block larger objects. On the other hand, the restriction would only apply to stream level flow control, because as long as there is any connection level flow control quota, every object can be partially written.

fluffy: We definitely need to be able to "stream" object in the sense of sending some of them at a time otherwise each hop of the relay network introduces latency that all adds up. In implementing our initial implementation ended up in the stupidity of sending zillions of tiny portions of the message so implementations need to have some good implementation logic on how to balance thisl

draft-ietf-moq-transport.md
```diff
@@ -1009,12 +1009,14 @@ congestion.
...
-2. An object in response to a SUBSCRIBE that belongs to a track with
+1. The first or next Object in a Subgroup that is in response to a subscription.
+2. An Object in response to a subscription that belongs to a Track with
    delivery preference Datagram.
-3. An object in response to a FETCH where that object is the next
-   object in the response.
+3. An Object in response to a FETCH where that Object is the next
+   Object in the response.
+
+An Object is not schedulable if it cannot be written due to underlying
```
fluffy: perhaps change "if it cannot be written" to "if no part of of the object can be written" or something like that

draft-ietf-moq-transport.md
```suggestion
An Object is not schedulable if it is known that no part of it can be written
due to underlying transport flow control limits.
```

suhasHere (approved):

Issue #1022, opened 2025-05-11
Make object ID omittable in datagrams


vasilvv:
Given the nature of datagrams, I expect datagrams to be mostly used in group-per-object use cases, where people would typically set object ID to zero.  Since we're already taking bits from type field, and type field in datagrams is particularly uninteresting, I think it would make sense to add a bit for setting object ID to zero.

gloinul:
I thought that someone at the interim talked about actually have groups also for audio that align with the video join points. I understand that in practice on the audio track one would be able to join at any point. And the application might even use the knowledge that it can start at any object, but the relevant information is have group ID for all audio objects matching the segment. 

Having said that if there are sufficient interest in being able to set object ID to zero in datagrams it can clearly be implemented.

martinduke:
Related to #877, so not too late

fluffy:
I like the idea of having a low bandwidth way for datagrams to say object ID went up by one, or group ID went up by one. 

How does this work when datagrams are delivered out of order ?

fluffy:
+1

Issue #877, opened 2025-03-27
Pack the bits


martinduke:
Individual comment:

There are tons of variable fields in some of these messages that take up a whole byte to communicate 1-2 bits of information. These can at least be consolidated, and in many cases can be brought directly into the message type byte. This saves a bandwidth and eliminates error cases for invalid values.

Another possibility is to push more of this into parameters.

For example:
SUBSCRIBE has 3 filter types (maybe going to 5) and 3 possible Group Orders, for 9-15 possible configs. If ranges are a rare 

special case, they could become parameters.

SUBSCRIBE_OK needs 1 bit for group order and 1 for content exists.

SUBSCRIBE_UPDATE updates up to 4 properties of the subscribe (counting all parameters as one priority), but all the fields are included. Assuming one property is being updated in most messages, this could be more compact by making each property a parameter.

FETCH has three types with very different fields, as well as three values for group order, for 9 possible configurations encoded in 2 bytes.

FETCH_OK: one bit for group order and one for end of track.

martinduke:
I would not object to parking this until message formats have settled a bit, although I would be equally happy to address it now.

afrind:
I'm not sure how hard we need to work to pack control messages, but we should pack data streams as much as we can.

afrind:
Suggestion from @ianswett - Specify a default publisher priority for all datagrams/subgroups in a subscription (included in PUBLISH and SUBSCRIBE_OK) and allow the publisher to omit Priority field -- especially effective for datagram tracks that never change priority.

afrind:
Notes from #1215

In the SUBSCRIBE or TRACK_STATUS message, we have two 8-bit fields Group Order and Forward, wherein Group Order only has three valid values and Forward only has two valid values. It would be sufficient to use just one byte for the two fields, by specifying 2 bits for Group Order and 1 bit for Forward, and possibly reserve the other 5 bits for future uses.

Similarly for the following:

Group Order and Content Exists in SUBSCRIBE_OK, PUBLISH, and TRACK_STATUS_OK
Forward and Group Order in PUBLISH_OK
Group order and End Of Track in FETCH_OK

yekuiwang:
Let me also have #1212 mentioned here, which is related.

yekuiwang:
> Notes from [#1215](https://github.com/moq-wg/moq-transport/issues/1215)
> 
[...]

PR #1055, opened 2025-06-23
Use a bit in Datagram Type to convey Object ID = 0

afrind:
Fixes: #1022

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3303,7 +3303,7 @@ OBJECT_DATAGRAM {
   Type (i) = 0x0-0x4,
   Track Alias (i),
   Group ID (i),
-  Object ID (i),
+  [Object ID (i),]
   Publisher Priority (8),
   [Extension Headers Length (i),
   Extension headers (...)],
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -3303,7 +3303,7 @@ OBJECT_DATAGRAM {
   Type (i) = 0x0-0x4,
```
```suggestion
  Type (i) = 0x0-0x7,
```

draft-ietf-moq-transport.md
```diff
@@ -3300,10 +3300,10 @@ An `OBJECT_DATAGRAM` carries a single object in a datagram.
 
 ~~~
 OBJECT_DATAGRAM {
-  Type (i) = 0x0-0x4,
+  Type (i) = 0x0-0x7,
   Track Alias (i),
   Group ID (i),
-  Object ID (i),
+  [Object ID (i),]
   Publisher Priority (8),
```
ianswett, line 3293: Sidenote: Expecially for datagrams, it seems very wasteful to always specify the Publisher Priority.  Is there an issue to fix that?

ianswett (approved):

draft-ietf-moq-transport.md
afrind: I put a note in #877

suhasHere (approved):

afrind:
Dang it.  This overlaps code points with OBJECT_DATAGRAM_STATUS (0x4/0x5).

We need to move OBJECT_DATAGRAM_STATUS higher in the space (eg: 0x20..0x21).  Maybe it also makes sense to recombine the two messages, eg:

```
OBJECT_DATAGRAM {
  Type (i),
  Track Alias (i),
  Group ID (i),
  [Object ID (i),]
  Publisher Priority (8),
  [Extension Headers Length (i),
  Extension headers (...),]
  [Object Payload (..) | Object Status (i)]
}
```

Where (for example) bit 5 indicates status/payload and bit 0 still indicates extensions present.

vasilvv (approved):

draft-ietf-moq-transport.md
```diff
@@ -3322,6 +3330,10 @@ Extension Headers Length is present.  If an endpoint receives a datagram with
 Type 0x01 and Extension Headers Length is 0, it MUST close the session with
 Protocol Violation.
 
+For Type values where Object ID Present is No, Object ID is not present and the
+Object has ID zero.  When Object ID Present is Yes, Object ID is present and
+contains the Object ID.
```
ianswett: nit:
```suggestion
For Type values where Object ID Present is No, the Object ID field is omitted
and the Object ID is 0.  When Object ID Present is Yes, the Object ID field is
present and encodes the Object ID.
```

ianswett (approved):
One editorial suggestion, but feel free to ignore it.

PR #1056, opened 2025-06-23
Make default PUBLISHER PRIORITY a parameter, optional in Subgroup/Datagram

ianswett:
Adds the ability to specify the parameter in PUBLISH, SUBSCRIBE_OK, and FETCH_OK.

Putting it on every Datagram and Fetch Object seems quite wasteful for most use cases, so this allows you to only specify a different value from the overall subscription value when needed.

Inspired by #1055 
Fixes: #1202

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1056,11 +1056,9 @@ implementation dependent what happens to objects that have already been
 scheduled.
 
 _Publisher Priority_ is a priority number associated with an individual
-schedulable object.  A default can be specified in the parameters of PUBLISH,
-SUBSCRIBE_OK, or FETCH_OK. Publisher priority can also be specified in the
-Extension Headers of the first Object in a subgroup, datagrams, or any
-Object in a FETCH response. If no value is specified, publisher priority
-defaults to 128.
+schedulable object.  A default can be specified in the parameters of PUBLISH, or
+SUBSCRIBE_OK. Publisher priority can also be specified in a subgroup header or
+datagram
 
 _Group Order_ is a property of an individual subscription.  It can be either
 'Ascending' (groups with lower group ID are sent first), or 'Descending'
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1037,8 +1037,9 @@ received and possibly scheduled.
 _Publisher Priority_ is a priority number associated with an individual
 schedulable object.  A default can be specified in the parameters of PUBLISH,
 SUBSCRIBE_OK, or FETCH_OK. Publisher priority can also be specified in the
-parameters in first Object of a subgroup, datagrams, or each Object in a
-FETCH response. If no value is specified, publisher priority defaults to 128.
+Extension Headers of the first Object in a subgroup, datagrams, or any
+Object in a FETCH response. If no value is specified, publisher priority
+defaults to 128.
 
 _Group Order_ is a property of an individual subscription.  It can be either
 'Ascending' (groups with lower group ID are sent first), or 'Descending'
@@ -1612,7 +1613,8 @@ The PUBLISHER PRIORITY parameter (Parameter Type 0x0E) specifies the priority
 of a subscription relative to other subscriptions in the same session.
 The value is from 0 to 255 and lower numbers get higher priority.
 See {{priorities}}. Priorities above 255 are invalid. The PUBLISHER PRIORITY
-parameter is valid in SUBSCRIBE_OK, PUBLISH and FETCH_OK.
+parameter is valid in SUBSCRIBE_OK, PUBLISH and FETCH_OK, as well as in
+Object Extension Headers.
 
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1130,7 +1130,7 @@ scheduled.
 _Publisher Priority_ is a priority number associated with an individual
 schedulable object.  A default can be specified in the parameters of PUBLISH, or
 SUBSCRIBE_OK. Publisher priority can also be specified in a subgroup header or
-datagram
+datagram (see {{data-streams}}).
 
 _Group Order_ is a property of an individual subscription.  It can be either
 'Ascending' (groups with lower group ID are sent first), or 'Descending'
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1657,12 +1657,14 @@ handles a downstream request that includes those Objects re-requests them.
 
 #### PUBLISHER PRIORITY Parameter {#subscriber-priority}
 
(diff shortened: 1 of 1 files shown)


afrind:
> This appears to remove per-subgroup priorities altogether.

Oops.  I think the idea is to use another bit to datagram/subgroup_header "Priority Present".

ianswett:
> > This appears to remove per-subgroup priorities altogether.
> 
> Oops. I think the idea is to use another bit to datagram/subgroup_header "Priority Present".

Actually, the intent was to allow the PUBLISHER PRIORITY param to be used in Object Extension Headers, but I wrote parameters instead.  PTAL.

afrind:
> Actually, the intent was to allow the PUBLISHER PRIORITY param to be used in Object Extension Headers, but I wrote parameters instead. PTAL.

Oh you did this in a very subtle way:

> as well as in Object Extension Headers.

I think we need to explicitly define the extension header down where Prior Group ID Gap is defined if that's how you want to do it.

Whether it should be baked into the fields with a bit in the type, or an extension depends on how ofter we would expect the priority to change, the parameter is larger.  Would we also want a control-plane message to change the default for the track after publishing begins?

draft-ietf-moq-transport.md
```diff
@@ -1056,8 +1056,11 @@ implementation dependent what happens to objects that have already been
 scheduled.
 
 _Publisher Priority_ is a priority number associated with an individual
-schedulable object.  It is specified in the header of the respective subgroup or
-datagram, or in each object in a FETCH response.
+schedulable object.  A default can be specified in the parameters of PUBLISH,
+SUBSCRIBE_OK, or FETCH_OK. Publisher priority can also be specified in the
+Extension Headers of the first Object in a subgroup, datagrams, or any
+Object in a FETCH response. If no value is specified, publisher priority
+defaults to 128.
```
```suggestion
schedulable object.  A default can be specified in the parameters of PUBLISH, or
SUBSCRIBE_OK. Publisher priority can also be specified in a subgroup header or
datagram
```

yekuiwang (commented):
These changes look good to me.

draft-ietf-moq-transport.md
```diff
@@ -1652,6 +1653,17 @@ multi-object stream will expire earlier than Objects later in the stream. Once
 Objects have expired from cache, their state becomes unknown, and a relay that
 handles a downstream request that includes those Objects re-requests them.
 
+#### PUBLISHER PRIORITY Parameter {#subscriber-priority}
+
+The PUBLISHER PRIORITY parameter (Parameter Type 0x0E) specifies the priority of
+a subscription relative to other subscriptions in the same session.  The value
+is from 0 to 255 and lower numbers get higher priority.  See
+{{priorities}}. Priorities above 255 are invalid. The PUBLISHER PRIORITY
```
ianswett, line 1745: Nit, isn't it impossible to encode values larger than 255 anyway?

```suggestion
{{priorities}}. The PUBLISHER PRIORITY
```

afrind: For now, you can encode up to varint max, however, when we circle back to #1184 we might decide it's TV, in which case we can move it back to be one byte.

vasilvv (approved):
I think we chose to not do this originally since we did not want to block receiving objects on SUBSCRIBE_OK; with the new way we pick track alias, this is moot, so this change makes sense.

afrind:
> I think we chose to not do this originally since we did not want to block receiving objects on SUBSCRIBE_OK; with the new way we pick track alias, this is moot, so this change makes sense.

There is a potential race condition with the design in this PR.  If SUBSCRIBE_OK set's default publisher priority to 10 but the objects arrive first without a priority field, then they will get Priority 128.  Assuming we follow this up with a publisher message to update the default, the race condition will be there too.  The publisher can avoid it by sending objects with the priority until it is (pretty) sure that the subscriber has the new default.

The race condition is further exacerbated because a caching relay could receive the same object and compute the priority differently, which elsewhere we say is a Malformed Track.

Saving bytes in datagrams in particular seems to outweigh the possibility of a small number of misprioritized objects.  Should we add a note here acknowledging this?

vasilvv:
>  If SUBSCRIBE_OK set's default publisher priority to 10 but the objects arrive first without a priority field, then they will get Priority 128.

I don't think that can happen, since you don't know track alias before receiving SUBSCRIBE_OK, so you can't do anything with that object anyways.

> Assuming we follow this up with a publisher message to update the default, the race condition will be there too.

Okay, that is a good point.  Should we disallow that?

draft-ietf-moq-transport.md
```diff
@@ -1128,8 +1128,9 @@ implementation dependent what happens to objects that have already been
 scheduled.
 
 _Publisher Priority_ is a priority number associated with an individual
-schedulable object.  It is specified in the header of the respective subgroup or
-datagram, or in each object in a FETCH response.
+schedulable object.  A default can be specified in the parameters of PUBLISH, or
+SUBSCRIBE_OK. Publisher priority can also be specified in a subgroup header or
+datagram
```
```suggestion
datagram (see {{data-streams}}).
```

draft-ietf-moq-transport.md
```diff
@@ -3351,6 +3373,11 @@ There are 10 defined Type values for OBJECT_DATAGRAM.
   and the Object ID is 0.  When Object ID Present is Yes, the Object ID field is
   present and encodes the Object ID.
 
+* Priority Present: If Priority Present is No, Priority is not present and this
+  Object inherits the Publisher Priority specified in the control message that
+  established the subscription.  When Priority Present is Yes, the Priority field
+  is present.
```
suhasHere, line 3379:
```suggestion
  established the subscription.  When Priority Present is Yes, the Priority field
  value is present and MUST be considered for scheduling objects.
```

suhasHere (approved):
few nits, otherwise LGTM

afrind:
> I don't think that can happen, since you don't know track alias before receiving SUBSCRIBE_OK, so you can't do anything with that object anyways.

You are right.

> Okay, that is a good point. Should we disallow that?

Disallow publisher updating the default?  That would be kind of sad since you can only get the compression benefit if you never change the priority.  If we used a mechanism with an acknowledgement, it would at least give the publisher a clear signal when it could use that.  Since we don't have a mechanism yet, we can proceed with this PR for now, and decide later.

draft-ietf-moq-transport.md
afrind: There is no anchor section for subscriber-priority (yet).  It's also defined immediately above this on line 1121.  I think a reference to subgroups and datagrams could be good though.

draft-ietf-moq-transport.md
afrind: "field is present" is used for the other optional fields (Extensions, Object ID) -- we can revisit using "field value" everywhere in an editorial pass though.

> MUST be considered for scheduling objects.

I think we have sufficient description of the normative rules around priority elsewhere.

afrind (approved):

draft-ietf-moq-transport.md
suhasHere: I do agree , but adding here doesn't hurt get the context I feel

PR #1059, opened 2025-06-25
Make START LOCATION and END LOCATION params

### LINKED ISSUE #1035: Add a more explicit mechanism to express an unfiltered subscription
Today all subscriptions are filtered, and a subscriber that wants no filtering has to specify "AbsoluteStart 0/0".  This is somewhat counter-intuitive.

ianswett:
Adds the START LOCATION and END LOCATION params and updates SUBSCRIBE, SUBSCRIBE_UPDATE, PUBLISH and FETCH to use them.

Currently this allows some functionality the current draft does not, such as subscriptions that end at a specific Object rather than a Group boundary.  We could restrict that if we wanted, but it's not clear it adds any implementation complexity.

Fixes #1035

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1608,8 +1608,8 @@ re-requests them.
 
 The START LOCATION parameter (Parameter Type 0x03) indicates the beginning
 of a subscription or fetch and can be used in SUBSCRIBE, SUBSCRIBE_UPDATE,
-FETCH, PUBLISH, and PUBLISH_OK. The default for START LOCATION is the
-beginning of the Track, equivalent to {0, 0}.
+FETCH, and PUBLISH_OK. The default for START LOCATION is the beginning of
+the Track, equivalent to None (0x0).
 
 There are 6 types of filter locations and any other value MUST be treated
 as an error. Some types are only the type value and others have one or
@@ -1639,8 +1639,8 @@ two sequential varints.
 
 The END LOCATION parameter (Parameter Type 0x05) indicates the end of a
 subscription or fetch and can be used in SUBSCRIBE, SUBSCRIBE_UPDATE,
-FETCH, PUBLISH, and PUBLISH_OK. The default for END LOCATION is the
-end of the Track, equivalent to {MaxVarint, MaxVarint}.
+FETCH, and PUBLISH_OK. The default for END LOCATION is the end of the
+Track, equivalent to None (0x1).
 
 The types and formatting of END LOCATION are identical to
 START LOCATION {{start-location}} defined above.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1607,9 +1607,9 @@ re-requests them.
 #### START LOCATION Parameter {#start-location}
 
 The START LOCATION parameter (Parameter Type 0x03) indicates the beginning
-of a subscription or fetch and can be used in SUBSCRIBE, SUBSCRIBE_UPDATE,
-FETCH, and PUBLISH_OK. The default for START LOCATION is the beginning of
-the Track, equivalent to None (0x0).
+of a subscription filter or fetch and can be used in SUBSCRIBE,
+SUBSCRIBE_UPDATE, FETCH, and PUBLISH_OK. The default for START LOCATION is
+the beginning of the Track, equivalent to None (0x0).
 
 There are 6 types of filter locations and any other value MUST be treated
 as an error. Some types are only the type value and others have one or
@@ -1638,7 +1638,7 @@ two sequential varints.
 #### END LOCATION Parameter {#end-location}
 
 The END LOCATION parameter (Parameter Type 0x05) indicates the end of a
-subscription or fetch and can be used in SUBSCRIBE, SUBSCRIBE_UPDATE,
+subscription filter or fetch and can be used in SUBSCRIBE, SUBSCRIBE_UPDATE,
 FETCH, and PUBLISH_OK. The default for END LOCATION is the end of the
 Track, equivalent to None (0x1).
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1604,6 +1604,48 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+#### START LOCATION Parameter {#start-location}
+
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,48 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
 re-requests them.
 
+#### START LOCATION Parameter {#start-location}
+
+The START LOCATION parameter (Parameter Type 0x03) indicates the beginning
+of a subscription or fetch and can be used in SUBSCRIBE, SUBSCRIBE_UPDATE,
+FETCH, PUBLISH, and PUBLISH_OK. The default for START LOCATION is the
+beginning of the Track, equivalent to {0, 0}.
+
+There are 4 types of filter locations and any other value MUST be treated
```
```suggestion
There are 6 types of filter locations and any other value MUST be treated
```

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,48 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+filter Start Location is {0, 0}.
+
+Next Group (0x2): The filter Location is `{Largest Object.Group + 1,
+0}`. If no content has been delivered yet, the Location is {0, 0}.
+
+Prior Group (0x3): The filter Location is '{Largest Object.Group - numgroups}'
+where numgroups is a specified as a subsequent varint.
+
+Absolute Group (0x4): The filter Group is specified explicitly as a
+subsequent varint.
```
afrind: And Start Object is 0.

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,48 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+#### START LOCATION Parameter {#start-location}
+
+The START LOCATION parameter (Parameter Type 0x03) indicates the beginning
+of a subscription or fetch and can be used in SUBSCRIBE, SUBSCRIBE_UPDATE,
+FETCH, PUBLISH, and PUBLISH_OK. The default for START LOCATION is the
```
afrind: Why PUBLISH?

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,48 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+
+Largest Object (0x1): The filter Location is `{Largest Object.Group,
+Largest Object.Object + 1}`. If no content has been delivered yet, the
+filter Start Location is {0, 0}.
+
+Next Group (0x2): The filter Location is `{Largest Object.Group + 1,
+0}`. If no content has been delivered yet, the Location is {0, 0}.
+
+Prior Group (0x3): The filter Location is '{Largest Object.Group - numgroups}'
+where numgroups is a specified as a subsequent varint.
```
afrind, line 1631: Let's remove this - I don't think we have an equivalent, and perhaps not a use case either?

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,48 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+#### START LOCATION Parameter {#start-location}
+
+The START LOCATION parameter (Parameter Type 0x03) indicates the beginning
+of a subscription or fetch and can be used in SUBSCRIBE, SUBSCRIBE_UPDATE,
+FETCH, PUBLISH, and PUBLISH_OK. The default for START LOCATION is the
+beginning of the Track, equivalent to {0, 0}.
```
afrind: Let's make the default None = No filtering?

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,48 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+#### START LOCATION Parameter {#start-location}
+
+The START LOCATION parameter (Parameter Type 0x03) indicates the beginning
+of a subscription or fetch and can be used in SUBSCRIBE, SUBSCRIBE_UPDATE,
```
afrind: subscription _filter_

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,48 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+FETCH, PUBLISH, and PUBLISH_OK. The default for START LOCATION is the
+beginning of the Track, equivalent to {0, 0}.
+
+There are 6 types of filter locations and any other value MUST be treated
+as an error. Some types are only the type value and others have one or
+more following varints.
+
+None (0x0): Specifies no filter, so the default is used. Useful in
+SUBSCRIBE_UPDATE to remove a previously specified filter.
```
afrind, line 1621: No filter.  I think for FETCH you can say the default is 0/0

afrind (commented):
Partial review

draft-ietf-moq-transport.md
ianswett: Why not?  But I guess if we don't allow that today, then we shouldn't change that in this PR?

draft-ietf-moq-transport.md
ianswett: Relative Joining Fetch is the use case.

draft-ietf-moq-transport.md
```suggestion
subsequent varint, and the Start Object is 0.
```

vasilvv (changes requested):
I am skeptical of reusing this for FETCH.  What does it mean for joining fetch? What does it even mean in PUBLISH_OK?

This also needs more text about when the `Largest Object` is resolved.

draft-ietf-moq-transport.md
```diff
@@ -1606,6 +1606,48 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+0}`. If no content has been delivered yet, the Location is {0, 0}.
+
+Prior Group (0x3): The filter Location is '{Largest Object.Group - numgroups}'
+where numgroups is a specified as a subsequent varint.
+
+Absolute Group (0x4): The filter Group is specified explicitly as a
+subsequent varint, and the Start Object is 0.
+
+Absolute Location (0x5): The filter Location is specified explicitly by
+two sequential varints.
```
fluffy, line 1637: trivial nit but here I think you want to say that the two varints represet the Group followed by the object

afrind:
We're doing this differently, at least for now, see #1210, #1273 and #1164

Issue #1003, opened 2025-05-06
Difficult to know if/how a SUBSCRIBE_UPDATE was processed


ianswett:
Originally, SUBSCRIBE_UPDATE was designed to always succeed.  However, there are things like DELIVERY_TIMEOUT that the publisher might not be willing to honor.

Additionally, one can't tell if a SUBSCRIBE_UPDATE was processed before or after sending the SUBSCRIBE_DONE, which is potentially annoying.

My strawman proposal is a PUBLISH_UPDATE, which would allow the publisher to ACK the SUBSCRIBE_UPDATE and would also allow the publisher to unilaterally change things like track priority (if it existed).

Additionally, if a peer is going away, it might want to tell the subscriber that it's going to end the subscription early at a group in the near future, so the subscriber would know what Group to start subscribing at from a different publisher.

ianswett:
When discussing whether to allow SUBSCRIBE_UPDATE to widen a subscription, it was agreed that a SUBSCRIBE_UPDATE_OK or similar message is valuable and likely needs to be added if we're going to widen a subscription.

afrind:
I think @fluffy also mentioned SUBSCRIBE_UPDATE_ERROR, because auth can fail, though perhaps that should simply terminate your subscription with SUBSCRIBE_DONE and appropriate code?

martinduke:
This is related to #770. If #770 is rejected, then this should be labeled "Past Deadline"

suhasHere:
If the Subscribe Update fails,  say due to Auth, the publisher sends PublishDone with error indicating the reason. I am not seeing a strong reason for new response message. Am i missing something ?

martinduke:
> If the Subscribe Update fails, say due to Auth, the publisher sends PublishDone with error indicating the reason. I am not seeing a strong reason for new response message. Am i missing something ?

Yes, that would work.

This is probably a failure of my imagination, but I don't see why AUTH is even necessary in SUBSCRIBE_UPDATE. Under what circumstances would the original SUBSCRIBE have been accepted but the UPDATE is some new level of privilege. Would a certain user identity be allowed to prewarm a subscription but then not be allowed to receive it?

Issue #794, opened 2025-03-19
Object Extension Types should not rob the bottom bit for even/odd games


mzanaty:
The current design robs the least significant bit of the object extension type to signal whether the following varint is the value or a length followed by the value. The goal is wire efficiency to avoid the extra byte of length for a simple extension. However, this actually harms wire efficiency since it reduces the efficient code point space from 64 (0-63 all) to 32 (0-62 even) possible small extensions. If an extension requires a length because it can't fit in the largest varint, that means it is over 8 bytes, hence it should not be a candidate to use the small space for small extensions.

If folks agree, I will create a PR to remove the even/odd shenanigans, and define all extension types 0-63 as varint values for small extensions, and all extension types 64-maxvarint as followed by varint length followed by value for larger extensions.

afrind:
#549 is relevant.  Maybe we should resolve those first to know if the impact of using the LSB for this is indeed 64 -> 32 or 128 -> 64?

mzanaty:
Even if #549 expands 1-byte varints to 128 values, I argue we should reserve all those 128 values for small extensions that fit in a varint. But I agree if #549 lands any changes, that would impact the range we reserve here. To be clear, I want to drop the odd/even design no matter what #549 does.

vasilvv:
If the goal is to maximize one-byte space, we could define `0x00`~`0xfe` as one-byte varint extensions, and `0xff` as "long" extensions.

afrind:
#672 is also relevant if we are changing the TLV scheme here.

suhasHere:
+1 on keeping one byte number space fully available for most common header extensions.

ianswett:
I would prefer landing #761 with the even/odd scheme and then exploring whether we need something more complex, because the current divergence between the two encodings increases complexity and is suboptimal.

afrind:
> +1 on keeping one byte number space fully available for most common header extensions.

There's a hidden assumption that the most _common_ header extensions are all integers, rather than something else that needs variable length.

> I would prefer landing https://github.com/moq-wg/moq-transport/pull/761 with the even/odd scheme

+1 lets get to a unified scheme first and circle back to bitpack.

afrind:
Is there a need for an extension type that doesn't have a value (eg: a true boolean)?

afrind:
Discussed in Stockholm - @mzanaty to write a PR giving all of the 1 byte space and some of the 2 byte space to varint only, and split the remaining space between varint and length prefix.

[...]

Issue #781, opened 2025-03-17
MoQT lacks support for track level object extension


zaheduzzaman:
MoQT is not very specific about whether a object extensions are session level or track level, however, only allows extension to be negotiated as the session setup. This means the object extensions are applicable for the whole session.

However, all extensions such as metadata about tracks is not necessarily a uniform session-level property. There can be several kinds of metadata that may be needed in one track and not in another. One example is the "rendered pose" in a split rendering scenario, i.e. the server signals ( say in an RTP header extension) the pose used to render the video frame carried. Another example is the audio level header extension [RFC 6464], which indicates the audio level of the audio sample carried in the RTP packet.
 
It is unclear how such per track level object extension can be negotiated.

gmarzot:
track-scoped metadata could live here: https://github.com/moq-wg/warp-streaming-format/pull/47 .  the use case you are describing has to be on the wire?

There does seem to be a gap in granularity with only object level extensions.  Subgroup header seems a natural place to support them at little cost.

zaheduzzaman:
Putting track-scoped metadata at the streaming format could work. I am wondering how we should signal that or do we need to singal the existance of the metadata? 

Another think is, for some cases the metadata could be useful to be seen and acted upon at the relay. I guess if we put this track-scoped metadata in the WARP then the relays won't have them.

afrind:
Individual Comment:

SUBSCRIBE_OK/FETCH_OK could carry track scoped metadata, but that may require some text like "relays MUST forward unknown parameters to downstream subscribers".

afrind:
> Having a clear use case to drive this might be helpful, since that could tell us whether Parameters are good enough as is or if we need something different?

I think that's a good point.

@zaheduzzaman @fluffy @suhasHere @vasilvv : do any of you have examples of track-level extensions you would like to support?  How many are "end-to-end" (eg: only the End Consumer cares) vs something relays might act on?

afrind:
Forwarding parameters from PUBLISH may not be a great idea - maybe there should be two sets of key-value pairs in those messages.

afrind:
I wonder if the @vasilvv pattern would work - add a parameter which contains the serialized Track Level extensions.

suhasHere:
I need to think more about this
. Is the expectation that, this is set by the OP and gets sent all the way to end subscriber ?

suhasHere:
For the audio level, the OP puts it as object header extension, since it is per audio object property. That is now we have implemented it.

[...]

PR #1060, opened 2025-06-25
DYNAMIC_GROUPS and NEW_GROUP_REQUEST parameters

afrind:
This is an alternate to #1031 that uses parameters in existing control messages instead of a new message type.

1. There's a specific opt-in sent by the publisher (DYNAMIC_GROUPS)
2. Explain the relay coalescing logic for NGR

Fixes: #1028

Closes: #1219

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -988,9 +988,9 @@ session with a `PROTOCOL_VIOLATION`.
 ### Joining an Ongoing Track
 
 The MOQT Object model is designed with the concept that the beginning of a Group
-is a joint point, so in order for a subscriber to join a Track, it needs to
-request an existing or future Group.  Different applications will have different
-approaches for when to begin a new Group.
+is a join point, so in order for a subscriber to join a Track, it needs to
+request an existing Group or wait for a future Group.  Different applications
+will have different approaches for when to begin a new Group.
 
 To join a Track at a past Group, the subscriber sends a SUBSCRIBE with Filter
 Type `Largest Object` followed by a Joining FETCH (see {{joining-fetches}}) for
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1630,7 +1630,7 @@ case.
 When an Original Publisher that supports dynamic Groups receives a
 NEW_GROUP_REQUEST with a value of 0 or a value larger than the current Group,
 it SHOULD end the current Group and begin a new Group as soon as practical.  The
-Original Publisher MAY delay or ignore the NEW_GROUP_REQUEST subject to
+Original Publisher MAY delay the NEW_GROUP_REQUEST subject to
 implementation specific concerns, for example, acheiving a minimum duration for
 each Group. The Original Publisher chooses the next Group ID; there are no
 requirements that it be equal to the NEW_GROUP_REQUEST parameter value.
@@ -1649,7 +1649,7 @@ value of 0 or a value larger than the Largest Group MUST send a SUBSCRIBE_UPDATE
 including the NEW_GROUP_REQUEST to the publisher unless:
 
 1. The Track does not support dynamic Groups
-2. There is already an outstanding NEW_GROUP_REQUEST
+2. There is already an outstanding NEW_GROUP_REQUEST from this Relay
 
 If a relay receives a NEW_GROUP_REQUEST with a non-zero value less than or equal
 to the Largest Group, it does not send a NEW_GROUP_REQUEST upstream.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1637,10 +1637,6 @@ requirements that it be equal to the NEW_GROUP_REQUEST parameter value.
 
 Relay Handling:
 
-A relay can have at most one outstanding NEW_GROUP_REQUEST per Track at a time.
-After sending a NEW_GROUP_REQUEST upstream, the request is considered
-outstanding until the Largest Group increases.
-
 A relay that receives a NEW_GROUP_REQUEST for a Track without an active
 subscription MUST include the NEW_GROUP_REQUEST when subscribing upstream.
 
@@ -1649,11 +1645,14 @@ value of 0 or a value larger than the Largest Group MUST send a SUBSCRIBE_UPDATE
 including the NEW_GROUP_REQUEST to the publisher unless:
 
 1. The Track does not support dynamic Groups
-2. There is already an outstanding NEW_GROUP_REQUEST from this Relay
+2. There is already an outstanding NEW_GROUP_REQUEST from this Relay with a
+   greater or equal value
 
 If a relay receives a NEW_GROUP_REQUEST with a non-zero value less than or equal
 to the Largest Group, it does not send a NEW_GROUP_REQUEST upstream.
 
+After sending a NEW_GROUP_REQUEST upstream, the request is considered
+outstanding until the Largest Group increases.
 
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1604,6 +1604,53 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+#### DYNAMIC GROUPS Parameter {#dynamic-groups}
+
+The DYNAMIC_GROUPS parameter (parameter type 0x20) MAY appear in PUBLISH or
+SUBSCRIBE_OK.  Values larger than 1 are a Protocol Violation.  When the value is
+1, it indicates that the subscriber can send a NEW_GROUP_REQUEST parameter in
+PUBLISH_OK or SUBSCRIBE_UPDATE for this Track.
+
+#### NEW GROUP_REQUEST Parameter {#new-group-request}
+
+The NEW_GROUP_REQUEST parameter (parameter type 0x22) MAY appear in PUBLISH_OK,
+SUBSCRIBE or SUBSCRIBE_UPDATE.  It is an integer representing the largest Group
+ID in the Track known by the subscriber, plus 1. A value of 0 indicates that the
+subscriber has no Group information for the Track.  A subscriber MUST NOT send
+this parameter in PUBLISH_OK or SUBSCRIBE_UPDATE if the publisher did not
+include DYNAMIC_GROUPS=1 when establishing the subscription.  A subscriber MAY
+include this parameter in SUBSCRIBE without foreknowledge of support.  If the
+publisher does not support dynamic Groups, it ignores the parameter in that
+case.
+
+A Original Publisher that supports dynamic Groups that receives a
+NEW_GROUP_REQUEST with a value of 0 or a value larger than the current Group
+SHOULD end the current Group and begin a new Group as soon as practical.  The
+Original Publisher MAY delay or ignore the NEW_GROUP_REQUEST subject to
+implementation specific concerns, for example, acheiving a minimum duration for
+each Group. The Original Publisher chooses the next Group ID; there are no
+requirements that it be equal to the NEW_GROUP_REQUEST parameter value.
+
+Relay Handling:
+
+A relay can have at most one outstanding NEW_GROUP_REQUEST per Track at a time.
+After sending a NEW_GROUP_REQUEST upstream, the request is considered
+outstanding until the Largest Group increases.
+
+A relay that receives a NEW_GROUP_REQUEST for a Track without an active
+subscription MUST include the NEW_GROUP_REQUEST when subscribing upstream.
+
+A relay that receives a NEW_GROUP_REQUEST for an established subscription with a
+value of 0 or a value larger than the Largest Group MUST send a SUBSCRIBE_UPDATE
+including the NEW_GROUP_REQUEST to the publisher unless:
+
+1. The Track does not support dynamic Groups
+2. There is already an outstanding NEW_GROUP_REQUEST
+
+If a relay receives a NEW_GROUP_REQUEST with a non-zero value less than or equal
+to the Largest Group, it does not send a NEW_GROUP_REQUEST upstream.
+
+
 ## CLIENT_SETUP and SERVER_SETUP {#message-setup}
 
 The `CLIENT_SETUP` and `SERVER_SETUP` messages are the first messages exchanged
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1611,6 +1611,10 @@ SUBSCRIBE_OK.  Values larger than 1 are a Protocol Violation.  When the value is
 1, it indicates that the subscriber can send a NEW_GROUP_REQUEST parameter in
 PUBLISH_OK or SUBSCRIBE_UPDATE for this Track.
 
+Relays MUST preserve the value of this parameter received from an upstream
+publisher in SUBSCRIBE_OK or PUBLISH when sending these messages to downstream
+subscribers.
+
 #### NEW GROUP_REQUEST Parameter {#new-group-request}
 
 The NEW_GROUP_REQUEST parameter (parameter type 0x22) MAY appear in PUBLISH_OK,
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1010,12 +1010,14 @@ One possible subscriber pattern is to SUBSCRIBE to a Track using Filter Type
 `Largest Object` and observe the `Largest Location` in the response.  If the
 Object ID is below the application's threshold, the subscriber sends a FETCH for
 the beginning of the Group.  If the Object ID is above the threshold and the
-Track supports dynamic groups, it sends a SUBSCRIBE_UPDATE message with the
+Track supports dynamic groups, the subscriber sends a SUBSCRIBE_UPDATE message with the
 NEW_GROUP_REQUEST parameter equal to the Largest Location's Group, plus one (see
 {{new-group-request}}).
 
 Another possible subscriber pattern is to send a SUBSCRIBE with Filter Type
-`Next Group Start` and NEW_GROUP_REQUEST equal to 0.
+`Next Group Start` and NEW_GROUP_REQUEST equal to 0.  The value of
+DYNAMIC_GROUPS in SUBSCRIBE_OK will indicate if the publisher supports dynamic
+groups. A publisher that does will begin the next group as soon as practical.
 
 ## Fetch State Management
 
@@ -1796,8 +1798,9 @@ does not expire or expires at an unknown time.
 
 The DYNAMIC_GROUPS parameter (parameter type 0x20) MAY appear in PUBLISH or
 SUBSCRIBE_OK.  Values larger than 1 are a Protocol Violation.  When the value is
-1, it indicates that the subscriber can send a NEW_GROUP_REQUEST parameter in
-PUBLISH_OK or SUBSCRIBE_UPDATE for this Track.
+1, it indicates that the subscriber can request the Original Publisher to start
+a new Group by including the NEW_GROUP_REQUEST parameter in PUBLISH_OK or
+SUBSCRIBE_UPDATE for this Track.
 
 Relays MUST preserve the value of this parameter received from an upstream
 publisher in SUBSCRIBE_OK or PUBLISH when sending these messages to downstream
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -985,6 +985,38 @@ than specified in `Start`.
 An endpoint that receives a filter type other than the above MUST close the
 session with a `PROTOCOL_VIOLATION`.
 
+### Joining an Ongoing Track
+
+The MOQT Object model is designed with the concept that the beginning of a Group
+is a joint point, so in order for a subscriber to join a Track, it needs to
+request an existing or future Group.  Different applications will have different
+approaches for when to begin a new Group.
+
+To join a Track at a past Group, the subscriber sends a SUBSCRIBE with Filter
+Type `Largest Object` followed by a Joining FETCH (see {{joining-fetches}}) for
+the intended start Group, which can be relative.  To join a Track at the next
+Group, the subscriber sends a SUBSCRIBE with Filter Type `Next Group Start`.
+
+#### Dynamically Starting New Groups
+
+While some publishers will deterministically create new Groups, other
+applications might want to only begin a new Group when needed.  A subscriber
+joining a Track might detect that it is more efficient to request the Original
+Publisher create a new group than issue a Joining FETCH.  Publishers indicate a
+Track supports dynamic group creation using the DYNAMIC_GROUPS parameter
+({{dynamic-groups}}).
+
+One possible subscriber pattern is to SUBSCRIBE to a Track using Filter Type
+`Largest Object` and observe the `Largest Location` in the response.  If the
+Object ID is below the application's threshold, the subscriber sends a FETCH for
+the beginning of the Group.  If the Object ID is above the threshold and the
+Track supports dynamic groups, it sends a SUBSCRIBE_UPDATE message with the
+NEW_GROUP_REQUEST parameter equal to the Largest Location's Group, plus one (see
+{{new-group-request}}).
+
+Another possible subscriber pattern is to send a SUBSCRIBE with Filter Type
+`Next Group Start` and NEW_GROUP_REQUEST equal to 0.
+
 ## Fetch State Management
 
 The publisher MUST send exactly one FETCH_OK or REQUEST_ERROR in response to a
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1623,9 +1623,9 @@ include this parameter in SUBSCRIBE without foreknowledge of support.  If the
 publisher does not support dynamic Groups, it ignores the parameter in that
 case.
 
-A Original Publisher that supports dynamic Groups that receives a
-NEW_GROUP_REQUEST with a value of 0 or a value larger than the current Group
-SHOULD end the current Group and begin a new Group as soon as practical.  The
+When an Original Publisher that supports dynamic Groups receives a
+NEW_GROUP_REQUEST with a value of 0 or a value larger than the current Group,
+it SHOULD end the current Group and begin a new Group as soon as practical.  The
 Original Publisher MAY delay or ignore the NEW_GROUP_REQUEST subject to
 implementation specific concerns, for example, acheiving a minimum duration for
 each Group. The Original Publisher chooses the next Group ID; there are no
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1815,7 +1815,7 @@ subscriber has no Group information for the Track.  A subscriber MUST NOT send
 this parameter in PUBLISH_OK or SUBSCRIBE_UPDATE if the publisher did not
 include DYNAMIC_GROUPS=1 when establishing the subscription.  A subscriber MAY
 include this parameter in SUBSCRIBE without foreknowledge of support.  If the
-publisher does not support dynamic Groups, it ignores the parameter in that
+original publisher does not support dynamic Groups, it ignores the parameter in that
 case.
 
 When an Original Publisher that supports dynamic Groups receives a


draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,53 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+publisher does not support dynamic Groups, it ignores the parameter in that
+case.
+
+A Original Publisher that supports dynamic Groups that receives a
+NEW_GROUP_REQUEST with a value of 0 or a value larger than the current Group
+SHOULD end the current Group and begin a new Group as soon as practical.  The
+Original Publisher MAY delay or ignore the NEW_GROUP_REQUEST subject to
+implementation specific concerns, for example, acheiving a minimum duration for
+each Group. The Original Publisher chooses the next Group ID; there are no
+requirements that it be equal to the NEW_GROUP_REQUEST parameter value.
```
afrind, line 1827: I want to highlight this difference with #1031 - if a publisher is on group 99 and receives NGR=200, it does not need to make the next Group ID=200.  It didn't seem right to give the subscribers that much control over track construction?  If the subscriber is subscribed to this publisher, it can't have a largest group greater than 99, so it sent 200 because it's buggy or malicious.  Otherwise it's some kind of multi-publisher scenario, in which case I'm not sure DYNAMIC_GROUPS make much sense.

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,53 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+subscriber has no Group information for the Track.  A subscriber MUST NOT send
+this parameter in PUBLISH_OK or SUBSCRIBE_UPDATE if the publisher did not
+include DYNAMIC_GROUPS=1 when establishing the subscription.  A subscriber MAY
+include this parameter in SUBSCRIBE without foreknowledge of support.  If the
+publisher does not support dynamic Groups, it ignores the parameter in that
+case.
+
+A Original Publisher that supports dynamic Groups that receives a
+NEW_GROUP_REQUEST with a value of 0 or a value larger than the current Group
+SHOULD end the current Group and begin a new Group as soon as practical.  The
```
```suggestion
When an Original Publisher that supports dynamic Groups receives a
NEW_GROUP_REQUEST with a value of 0 or a value larger than the current Group,
it SHOULD end the current Group and begin a new Group as soon as practical.  The
```

draft-ietf-moq-transport.md
ianswett: I think it's ok to say that if a subscriber says their largest Group is X, and X has never been published, that's an error of some sort.

ianswett (approved):
Some small comments, but I think this looks good.

draft-ietf-moq-transport.md
afrind: I was going to make this an error, but I realized that the OP discovers the error, and...what...ends the subscription?  That means a single broken downstream subscriber can spoil the upstream subscription.  The relay shouldn't clamp the NGR value to its Largest+1 either, since a valid subscriber got Group 199 from relay A and then switched to relay B and could be ahead of the largest group there.

ianswett: I need to write a PR for it, but it'll soon be possible to fail a SUBSCRIBE_UPDATE, which is a better option than failing the subscription, see #1003 and #770

If we had SUBSCRIBE_UPDATE_OK or similar, that could respond with the Group ID of the next/most recent join point that a subscriber should use, which could be slightly in the past.  This still allows coalescing SUBSCRIBE_UPDATE upstream when a NGR is made, but without the complexity of a param value.

ianswett (commented):
I'm not sure we need a Group ID in the request, as I mentioned in my comment.  I think it's making the mechanism a bit more complex than it needs to be.

The idea that a new SUBSCRIBE could cause a SUBSCRIBE_UPDATE upstream is reasonable, even as the draft stands today.  However, I'm concerned that if we tried to do this as an extension, it wouldn't work, because if any hop didn't understand the param, no more updates would be sent up the chain.  This is along the lines of my original concern.  However, if we put it in MoQ Transport, it does work because everyone understands it so this change itself should be fine.

ianswett:
I'll observe that for a single hop, this is pretty much identical to a SUBSCRIBE or SUBSCRIBE_UPDATE with the Start being Next Group.

draft-ietf-moq-transport.md
afrind: If we removed the group ID from the request, the downside is that the OP may have already begun the group needed by the subscriber and it will get a new request it is unable to deduplicate.  This will cause unnecessary increase in bandwidth to generate a new i-frame (even if the OP delays some minimum).  With the ID, either the relay or the OP can optimize away NGR that have already been satisfied.

afrind:
> I'll observe that for a single hop, this is pretty much identical to a SUBSCRIBE or SUBSCRIBE_UPDATE with the Start being Next Group.

It's slightly different:

SUBSCRIBE (NextGroup) -- I'll wait for the next group
SUBSCRIBE (NewGroupRequest) -- I need you to start a new group ASAP

draft-ietf-moq-transport.md
suhasHere: > I want to highlight this difference with #1031 - if a publisher is on group 99 and receives NGR=200, it does not need to make the next Group ID=200. It didn't seem right to give the subscribers that much control over track construction? If the subscriber is subscribed to this publisher, it can't have a largest group greater than 99, so it sent 200 because it's buggy or malicious. Otherwise it's some kind of multi-publisher scenario, in which case I'm not sure DYNAMIC_GROUPS make much sense.

I thunk it is not clear in #1031. But you are right in the intention with the either of the approaches. Subscriber doesn;t control the group Id generation, but just a new sync point that publisher can do .

suhasHere: > If we removed the group ID from the request, the downside is that the OP may have already begun the group needed by the subscriber and it will get a new request it is unable to deduplicate. This will cause unnecessary increase in bandwidth to generate a new i-frame (even if the OP delays some minimum). With the ID, either the relay or the OP can optimize away NGR that have already been satisfied.

Also GroupID acts as gating points at the relay publishers too and thus naturally aggregating the new requests. This avoids maintaining timers at relays to manage sync point requests.

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,53 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+#### DYNAMIC GROUPS Parameter {#dynamic-groups}
+
+The DYNAMIC_GROUPS parameter (parameter type 0x20) MAY appear in PUBLISH or
+SUBSCRIBE_OK.  Values larger than 1 are a Protocol Violation.  When the value is
+1, it indicates that the subscriber can send a NEW_GROUP_REQUEST parameter in
+PUBLISH_OK or SUBSCRIBE_UPDATE for this Track.
```
```suggestion
PUBLISH_OK or SUBSCRIBE_UPDATE for this Track.  This parameter allows subscribers request a new synchronization point (new Group) from the publisher
```

suhasHere: One way I do feel this parameter is useful. However, I wonder if this make new_group_request not interop with the relays, since a relay-publisher can turn it off per subscription or not support dynamic groups( which is fine too).

My inclination was to have it mandatory for relays to support. Thoughts ?

afrind: @suhasHere I think you are saying a relay MUST store/forward this parameter to downstream subscribers, and implement the NGR logic?  This seems fine.

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,53 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+
```
afrind, line 1804:
```suggestion

Relays MUST preserve the value of this parameter received from an upstream
publisher in SUBSCRIBE_OK or PUBLISH when sending these messages to downstream
subscribers.

```

ianswett (commented):
As discussed today, I think we should address this use case soon, but I did come up with a different approach at the hackathon that I'll writeup.

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,57 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+Relays MUST preserve the value of this parameter received from an upstream
```
TimEvens, line 1805: Why?

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,57 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+#### DYNAMIC GROUPS Parameter {#dynamic-groups}
+
+The DYNAMIC_GROUPS parameter (parameter type 0x20) MAY appear in PUBLISH or
```
TimEvens, line 1799: Considering unknown params are ignored, why could we not just send the optional parameter `NEW_GROUP_REQUEST`?  It's a passive request to ask for a new group. It's up to the publisher (aka APP) to actually start a new group or not.  What is the use-case driving why we need to maintain yet another state if the publisher supports new group request parameter?

draft-ietf-moq-transport.md
TimEvens: It should be optional and treated more of a hint to the publisher (APP) to start a new group. It should not be a **MUST** as many app use-cases will not support that.... We have an app that uses group for example for something else.  Sending new group to this app will not result in a group change.

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,57 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+subscribers.
+
+#### NEW GROUP_REQUEST Parameter {#new-group-request}
+
+The NEW_GROUP_REQUEST parameter (parameter type 0x22) MAY appear in PUBLISH_OK,
+SUBSCRIBE or SUBSCRIBE_UPDATE.  It is an integer representing the largest Group
+ID in the Track known by the subscriber, plus 1. A value of 0 indicates that the
+subscriber has no Group information for the Track.  A subscriber MUST NOT send
+this parameter in PUBLISH_OK or SUBSCRIBE_UPDATE if the publisher did not
+include DYNAMIC_GROUPS=1 when establishing the subscription.  A subscriber MAY
```
TimEvens, line 1816: I would prefer we don't add more state unless it adds value. The value isn't articulated here why we would need this when considering unknown/supported params can be ignored. The following sentence mentions that the publisher can ignore it.

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,57 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+publisher does not support dynamic Groups, it ignores the parameter in that
+case.
+
+When an Original Publisher that supports dynamic Groups receives a
+NEW_GROUP_REQUEST with a value of 0 or a value larger than the current Group,
+it SHOULD end the current Group and begin a new Group as soon as practical.  The
+Original Publisher MAY delay or ignore the NEW_GROUP_REQUEST subject to
+implementation specific concerns, for example, acheiving a minimum duration for
+each Group. The Original Publisher chooses the next Group ID; there are no
+requirements that it be equal to the NEW_GROUP_REQUEST parameter value.
```
TimEvens, line 1827: > If we removed the group ID from the request, the downside is that the OP may have already begun the group needed by the subscriber and it will get a new request it is unable to deduplicate. This will cause unnecessary increase in bandwidth to generate a new i-frame (even if the OP delays some minimum). With the ID, either the relay or the OP can optimize away NGR that have already been satisfied.

Are you suggesting that new group could request a specific group? The text suggests otherwise as it's just a request to start a new group, whatever that new group id might be.   

Maybe you were trying to address a race condition that a new group request was received for X and then another request comes in for the same but was satisfied by the first new group request, thus the second could be suppressed... The method that I would prefer to handle this is via dampening the churn of subscribers and new group requests.  If we have 1000+ subscribers, all starting up asking for new group, we surely will consolidate that and dampen what is actually sent to the publisher. The relays know the start of each subscriber, so it knows if  that subscriber received a new group or not.  If it didn't, it's possible to re-queue the new group request.  Having the group Id in the param request requires the subscriber to know it and the relay to do some more advanced reconciliation of 1000+ subscriber new group requests with different group IDs.   This is far more complicated than it needs to be.  

We really just need a bool to request a new group with some constraints around sending new group request to the publisher so that we don't flood the publisher  and don't end up with every object being a new group.

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,57 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+
+Relay Handling:
+
+A relay can have at most one outstanding NEW_GROUP_REQUEST per Track at a time.
+After sending a NEW_GROUP_REQUEST upstream, the request is considered
+outstanding until the Largest Group increases.
```
TimEvens: Makes sense to dampen the flood of new group requests, but this method is broken when the group never changes. Based on this text, new group will never be sent again if the group doesn't increment... The APP may decide to ignore the request based on timing/whatever.  It may then never increment the group id. Had a new group request been sent again later it might have been accepted and resulted in a new group. 

```suggestion
complete. Subsequent NEW_GROUP_REQUEST received SHOULD be aggregated to a single request to the publisher that will be delayed by CLIENT_SETUP parameter TBD milliseconds from when the last request was sent. 
```

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,57 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+
+A relay that receives a NEW_GROUP_REQUEST for a Track without an active
+subscription MUST include the NEW_GROUP_REQUEST when subscribing upstream.
+
+A relay that receives a NEW_GROUP_REQUEST for an established subscription with a
+value of 0 or a value larger than the Largest Group MUST send a SUBSCRIBE_UPDATE
+including the NEW_GROUP_REQUEST to the publisher unless:
```
TimEvens, line 1836: The below list will likely only be known by the origin relay.  The downstream relays do not have direct access to the publisher and the negotiated params for that publisher connection.  This is especially the case with multi-publisher. 

The text should say origin relay does this.

draft-ietf-moq-transport.md
suhasHere: The original publisher is free to apply its logic on when and how to generate the new group. An application, say video conferencing app that supports non-static GOPs, will apply some hueristcs/batching logically at its end to decide when to generate a new group on the data stream.

draft-ietf-moq-transport.md
suhasHere: Since the groupId is latching point and avoids using timers configured across all the relays. In the current proposal, the timers or such logic is moved only to the OP.

draft-ietf-moq-transport.md
suhasHere: GroupID  is used to reduce the churn to the original publisher. A simple boolean along will not work . you need a combination of simple boolean + dampening logic ( like timers, for example). 

The goal with this proposal is to reduce signaling churn ( when requests can be latched) and to not to have arbitrary timers across the relay network.

draft-ietf-moq-transport.md
TimEvens: Recall the implementation will apply this at the publisher edge relay only. Timers for such can be negotiated via parameters for such. Not complex or extensive to implement when considering publisher edge (aka origin relay).

draft-ietf-moq-transport.md
TimEvens: Dampening IMO is required. Bool is totally fine and group id only complicates things with reconciliation between other subscriber requests. The net result is to start a new group, despite whatever the group id is. It’s a classic synchronization request to start anew.

draft-ietf-moq-transport.md
TimEvens: +1 but a little vague and open ended

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,57 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
 a relay that handles a downstream request that includes those Objects
 re-requests them.
 
+#### DYNAMIC GROUPS Parameter {#dynamic-groups}
+
+The DYNAMIC_GROUPS parameter (parameter type 0x20) MAY appear in PUBLISH or
+SUBSCRIBE_OK.  Values larger than 1 are a Protocol Violation.  When the value is
+1, it indicates that the subscriber can send a NEW_GROUP_REQUEST parameter in
```
fluffy: I much rather have a boolean type type than a one off customer boolean masquerading as as an int.

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,57 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+NEW_GROUP_REQUEST with a value of 0 or a value larger than the current Group,
+it SHOULD end the current Group and begin a new Group as soon as practical.  The
+Original Publisher MAY delay or ignore the NEW_GROUP_REQUEST subject to
+implementation specific concerns, for example, acheiving a minimum duration for
+each Group. The Original Publisher chooses the next Group ID; there are no
+requirements that it be equal to the NEW_GROUP_REQUEST parameter value.
+
+Relay Handling:
+
+A relay can have at most one outstanding NEW_GROUP_REQUEST per Track at a time.
```
fluffy: I really don't like this part. it causes race conditions where the algorithm will fail. I think it should simply be the relay only sends up if larger than any previous request.

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,57 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+After sending a NEW_GROUP_REQUEST upstream, the request is considered
+outstanding until the Largest Group increases.
+
+A relay that receives a NEW_GROUP_REQUEST for a Track without an active
+subscription MUST include the NEW_GROUP_REQUEST when subscribing upstream.
+
+A relay that receives a NEW_GROUP_REQUEST for an established subscription with a
```
fluffy, line 1834: if a newgroup of 1 has been requested, I don' think the relay should send a newgroup of 0

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,57 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+value of 0 or a value larger than the Largest Group MUST send a SUBSCRIBE_UPDATE
+including the NEW_GROUP_REQUEST to the publisher unless:
+
+1. The Track does not support dynamic Groups
+2. There is already an outstanding NEW_GROUP_REQUEST
+
+If a relay receives a NEW_GROUP_REQUEST with a non-zero value less than or equal
+to the Largest Group, it does not send a NEW_GROUP_REQUEST upstream.
```
fluffy, line 1843: Again, I think this is a bad design, is should also not send it if it has already sent a new group reques that is equal too or larger than this

fluffy (changes requested):
I don't think this is the algorithm I thought we had agreed too. I want relay to only send a new group greq is the request group is larger than anything in previous new group requests. That will give us good aggregation while at the same time not requiring strage keeping track of pending transactions that have race conditions with transaction finished but group did not change so we keep doing more requests for the same group. 

Might need some time to explain on a call

draft-ietf-moq-transport.md
fluffy: Say we have 100k people joining a company meeting in the first 2 minutes of the call. First person joins, great and publisher sends I frame as group=1. Now the next 100k-1 people are going to request a new group. Let say in the first 2 seconds , 10 k people join, and in the next 2 seconds another 10k people join and that the original publisher is configure not to do an I frame more often than every 2 seconds. 

The goal is for the relay to end up agregating most of the requests so the original publisher only gets a new group request shortly after each time it starts a new group ( every 2 seconds in this case )

draft-ietf-moq-transport.md
afrind: @fluffy @mzanaty has volunteered to write a PR addressing #794 and adding boolean types in Stockholm.  🤞

draft-ietf-moq-transport.md
afrind: This is to protect publishers that don't support the NGR from getting any useless SUBSCRIBE_UDPATES (consider live stream publisher with 1M subscribers)

draft-ietf-moq-transport.md
afrind: > Why?

This param indicates to both relays and subscribers that NGR is allowed.  It needs to propagate anywhere the Track goes.

draft-ietf-moq-transport.md
afrind: If the state you are referring to is the opt-in parameter, I think the wg felt strongly this was needed.

draft-ietf-moq-transport.md
afrind: @fluffy The value of the param is the current group + 1.  0 means "I don't know the current group" not "I want group 0"

draft-ietf-moq-transport.md
afrind: The key difference is that if a Relay forwards NGR 10, it won't forward NGR 11 until after it sees group 10 (or higher).  This prevents misbehaving clients from sending NGR 1...MAXINT, flooding the OP.

Can you explain the race conditions you see in detail?

draft-ietf-moq-transport.md
afrind: I think you are right. The OP shouldn't _ignore_ an NGR, but it can arbitrarily delay it.

draft-ietf-moq-transport.md
afrind: I don't see why only the origin relay would know this:

Since we forward "DYNAMIC_GROUPS" any relay with the track knows 1.

Any relay that sees an NGR and forwards it knows there is an NGR outstanding, even if not the origin.  I think maybe I could clarify this to 

> There is already an outstanding NEW_GROUP_REQUEST _from this Relay_

draft-ietf-moq-transport.md
afrind: You are saying Largest Group = 10 and Relay gets NGR 15, then NGR 12? I think this is covered.

This algorithm will send NGR 15, but not send NGR 12 because there's an outstanding NGR.  If the relay sends NGR=15, receives Group 11, then receives NGR=12 it *would* forward it but that seems correct, since it could be a subscriber that's seen 11 and wants something new.

The group value is only for aggregation and not a command to mint a specific group number.

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,57 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+this parameter in PUBLISH_OK or SUBSCRIBE_UPDATE if the publisher did not
+include DYNAMIC_GROUPS=1 when establishing the subscription.  A subscriber MAY
+include this parameter in SUBSCRIBE without foreknowledge of support.  If the
+publisher does not support dynamic Groups, it ignores the parameter in that
+case.
+
+When an Original Publisher that supports dynamic Groups receives a
+NEW_GROUP_REQUEST with a value of 0 or a value larger than the current Group,
+it SHOULD end the current Group and begin a new Group as soon as practical.  The
+Original Publisher MAY delay or ignore the NEW_GROUP_REQUEST subject to
```
```suggestion
Original Publisher MAY delay the NEW_GROUP_REQUEST subject to
```

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,57 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+
+A relay that receives a NEW_GROUP_REQUEST for a Track without an active
+subscription MUST include the NEW_GROUP_REQUEST when subscribing upstream.
+
+A relay that receives a NEW_GROUP_REQUEST for an established subscription with a
+value of 0 or a value larger than the Largest Group MUST send a SUBSCRIBE_UPDATE
+including the NEW_GROUP_REQUEST to the publisher unless:
+
+1. The Track does not support dynamic Groups
+2. There is already an outstanding NEW_GROUP_REQUEST
```
```suggestion
2. There is already an outstanding NEW_GROUP_REQUEST from this Relay
```

draft-ietf-moq-transport.md
gwendalsimon: No, I am strongly opposed to any message that would force an Origin Publisher to create a New Group upon request, even delayed.

It was very clear since the beginning (and we collectively agreed) that this New Group Request is a great feature for video conferencing scenario but it is dangerous in any massive live streaming use case, so it must not be a forceful message. We agreed to debate on this topic at the condition that the Origin Publisher can decide to not apply a new group.

An Origin Publisher for the Superbowl or for moon landing does not want to pay any single attention to NGR messages.

afrind: @gwendalsimon the OP has to opt-in to NGR by sending DYNAMIC_GROUPS=1.  Does that mitigate your concern?

gwendalsimon: Not really. The OP should be free to decide whether it wants to create a new group, to delay this creation, or to ignore. There are cases where it makes sense to create a group (dynamic frames, small population, many resources...) and ten minutes later, it does not make sense any longer (static frames, large population, less resources...).
Imho, subscribers are free to send a NGR, and, well, they may get a new group soon, or not, ... but they have to understand that there is no "right to get a new group". The OP has a much wider understanding of the configuration and he may not fulfill all requests.

afrind: The only way to truly "ignore" an NGR is to _never_ make a new group.  If we retain the algorithm described here, an OP that silently drops an NGR and never starts a new group might never get another NGR.  Perhaps we can leave "ignore" in as a valid action with that caveat?

We can also fix this is by relaxing the restriction at the relay (as Cullen suggests), at the expense of losing some of the protection offered by the "one NGR at a time" rule.

draft-ietf-moq-transport.md
fluffy: If the NGR 15 had completed at the point the NGR 12 happened, there would not be an outstanding and at that point I thought it would send the NGR 12. I thought that was what was being proposed here but perhaps I had it wrong. I guess I might be unclear on what outstanding transaction means

draft-ietf-moq-transport.md
afrind: @fluffy in case you missed my question here

draft-ietf-moq-transport.md
afrind: @fluffy 

> If the NGR 15 had completed at the point the NGR 12 happened, there would not be an outstanding and at that point I thought it would send the NGR 12.

This is how it works. Per my example above:

```
Relay Largest = 10
S1->R: NGR 15
R->P: NGR
P->R: Object, Group=11
S2->R: NGR 12
R->P: NGR
```

Conceptually the NGR has _no group value_ when received by the OP (relays do include it in case next upstream is a relay).  See this bit:

>  The Original Publisher chooses the next Group ID; there are no requirements that it be equal to the NEW_GROUP_REQUEST parameter value.

> I guess I might be unclear on what outstanding transaction means

The definition is on line 1641:

> After sending a NEW_GROUP_REQUEST upstream, the request is considered outstanding _until the Largest Group increases_.

Can you articulate why you think this is a bad design?  And why this is a necessary requirement?

> it should also not send it if it has already sent a new group request that is equal too or larger than this

fluffy:
@afrind  - I think I need to make some slides to explain the race condition I am concerned with. I'll make these and find some time with you

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,57 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+A relay that receives a NEW_GROUP_REQUEST for an established subscription with a
+value of 0 or a value larger than the Largest Group MUST send a SUBSCRIBE_UPDATE
+including the NEW_GROUP_REQUEST to the publisher unless:
+
+1. The Track does not support dynamic Groups
+2. There is already an outstanding NEW_GROUP_REQUEST from this Relay
+
+If a relay receives a NEW_GROUP_REQUEST with a non-zero value less than or equal
+to the Largest Group, it does not send a NEW_GROUP_REQUEST upstream.
+
```
```suggestion
A relay that receives a NEW_GROUP_REQUEST for a Track without an active
subscription MUST include the NEW_GROUP_REQUEST when subscribing upstream.

A relay that receives a NEW_GROUP_REQUEST for an established subscription with a
value of 0 or a value larger than the Largest Group MUST send a SUBSCRIBE_UPDATE
including the NEW_GROUP_REQUEST to the publisher unless:

1. The Track does not support dynamic Groups
2. There is already an outstanding NEW_GROUP_REQUEST from this Relay with a
   greater or equal value

If a relay receives a NEW_GROUP_REQUEST with a non-zero value less than or equal
to the Largest Group, it does not send a NEW_GROUP_REQUEST upstream.

After sending a NEW_GROUP_REQUEST upstream, the request is considered
outstanding until the Largest Group increases.
```

draft-ietf-moq-transport.md
afrind: Spoke offline with @fluffy.  The scenario where the "one outstanding" request does not work is as follows:


Largest Group seen in parens
```
C1 (22) -> R1 (22) -> P (24)
C2 (24) -> R2 (24) -/
```

R1 and C1 got behind.  Now C2 fails over from R2 to R1.

```
C1 (22) -> R1 (22) -> P (24)
C2 (24) -/
```

C1 sends NGR 23
C2 sends NGR 25

Under the original algorithm, the OP would discard NGR 23, because it has published 24, but R1 would discard NGR 25, because NGR 23 is outstanding (until the an object from 24 reaches R1).

The tradeoff is that a buggy or malicious client can send an unlimited number of NEW_GROUP_REQUESTs to the OP.  Cullen pointed out that a client can already send an unlimited number of FETCH requests that reach the OP as well.

suhasHere (approved):
Re-reviewed the PR and this looks good.

draft-ietf-moq-transport.md
fluffy: latest algorithm resolves my comment

draft-ietf-moq-transport.md
TimEvens: This dramatically complex inter-relay communication having to maintain state in more than one relay based on a publish that happened in the past where the relay had no subscribers.  It also is overly complex with multi-publisher where not all publishers will indicate new group or not. Tracking that state and sending to subscriber is more complex in that we have to send new group in OK to subscriber if at least one publisher supports it. When the subscriber sends it, I suspect we are to send it only to the publisher that said it would support it. This requires additional state with inter-relay protocol.

draft-ietf-moq-transport.md
TimEvens: Params are ignored if not supported. The publisher should just ignore it if it doesn't support the param in the update or subscribe.  This is redundant to that and adds complexity where it's not needed.

draft-ietf-moq-transport.md
```diff
@@ -1604,6 +1604,56 @@ stream. Once Objects have expired from cache, their state becomes unknown, and
...
+A relay that receives a NEW_GROUP_REQUEST for an established subscription with a
+value of 0 or a value larger than the Largest Group MUST send a SUBSCRIBE_UPDATE
+including the NEW_GROUP_REQUEST to the publisher unless:
+
+1. The Track does not support dynamic Groups
+2. There is already an outstanding NEW_GROUP_REQUEST from this Relay with a
+   greater or equal value
+
+If a relay receives a NEW_GROUP_REQUEST with a non-zero value less than or equal
+to the Largest Group, it does not send a NEW_GROUP_REQUEST upstream.
```
TimEvens, line 1843: In the case of peering/inter-relay, there is no ACK that new-group-request was actually sent to the publisher. `PENDING` state in this case could only be cleared if the relay receives a new group, which it may never receive if the publisher didn't get or decided to ignore it. This would result in subsequent requests for new group to be suppressed until then... which could be indefinitely.

afrind:
Thanks for your review @TimEvens, comments inline:

> 1. `NEW_GROUP_REQUEST` param can only be sent by the subscriber IF the `SUBSCRIBE_OK` indicated that it's supported.  The keyword is **MUST** but there is no recourse stated if the subscriber violates this MUST.  Does the relay close the connection with protocol violation or should it ignore it, or should it `SUBSCRIBE_ERROR` ?

Good point, until we have a way to fail SUBSCRIBE_UPDATE, the options are limited, I'd say it's a PROTOCOL_VIOLATION (close session) for now.

> 2. Value of zero is used when the subscriber doesn't know the group. Makes sense, but  is that really true? The subscriber could have had a `SUBSCRIBE_OK` with group id indicated as well could have been sent objects with group_id.  In this case the subscriber MUST know the group id, but because this PR doesn't clarify that,  the subscriber cloud always use zero. This PR does not require the subscriber that has previously been sent group_id via subscribe ok or via an object to honor that information when sending  `NEW_GROUP_REQUEST=<group_id + 1>`.

I can see it's a little nicer to include the group if you know it.  We can add some encouragement to this effect, it could be a MUST I suppose, but I'm ambivalent.  A client can always get the same effect by sending a random group ID larger than what's been delivered/known.

> 3. The way this is written, it suggests that ZERO will be sent to the publisher regardless of other subscribers that have requested specific group IDs. In inter-relay (peering), does the ZERO override sending a new group request with a value? In other words, if zero is received at the same time as other requests with group ids present, send only the zero?.?

The updated PR says you only send NGR upstream there's not already one, or if the value is greater than one already sent.  So the first 0 gets through but all other 0s are coalesced.  Specific values always override a zero.  If you are using zero properly, it's always fine to get coalesced to another NGR, since you're asking for _any_ new group.

> 4. A relay sends new group request but there is no ACK that it was actually delivered to the publisher. In addition, the publisher may have received it, but since there is no ACK with a status of try again later or ignored, the relay will no longer send new group requests to the publisher. Resulting in the publisher never starting a new group. The PR states that the publisher may delay the new group request, but it doesn't clearly state that the original publisher MUST maintain this request state because it will never receive a new request after this until it starts a new group. Suggest "**Original Publisher Handling:**"...  Some applications may not want to maintain this state and would prefer to get a request again from subscribers if it still needs it.

Yes, the OP can delay (indefinitely) but can't ignore. We may need to clarify that when the OP disconnects the relay clears the state too.  I think SUBSCRIBE_UPDATE responses are coming soon in a yet to be written PR.

> 5. Subscribe updates are not currently defined in MOQT to be queued and flooded or reconciled by the relay when peering/inter-relay communication is established. In other words, there maybe be a disconnect from the subscriber relay and the original publisher relay when new group request is sent. The PR does indicate that if the subscribe comes in with new-group request, that state would be maintained, but it does not mention anything about what happens to subscribe updates when there is this break of communication in the middle... This would result in the relay holding on to pending state that it sent new group request, but that actually was dropped in the middle with no feedback to the relay.

If I understand your scenario correctly you've got:

P -> R1 -> R2 -> S

but there's a break so either

P x R1 -> R2 -> S or
P -> R1 x R2 -> S

I think SUBSCRIBE_UPDATE responses will help here too, since R1/R2 can indicate that the update failed (because there's no OP).

> 6. There is lack of clarity mentioned for how the relay should handle multiple new group requests with different group ID's, which could be zero as well. If current group is 10 and subscriber 1 asks for 20 and subscriber 2 asks for 8 and subscribe 3 asks for 50, which new-group request group-id do we send to the original publisher?

It's pretty clear - you send if it's larger than the largest one you've sent and largest than largest group.  So in your example, it would send 20 and 50, but not 8.

> 7. Considering there is no ACK for the subscriber that new group will be handled or has been handled, the subscriber will have no choice but to continue to send new group requests when receiving objects that haven't changed to the new requested group. This would result in flood.

The relay would drop them because the value of the group would be the same.  The relay already needs to handle update floods (also, we added a separate Request ID to SUBSCRIBE_UPDATE, which will limit the flood rate to N per RTT).

We can consider Parking this until SUBSCRIBE_UPDATE responses have landed since that might make the solution more robust.

draft-ietf-moq-transport.md
afrind: Based on feedback from the working group, this is a must-have from the live-stream folks who want nothing to do with these updates.  The complexity is fairly minimal.

draft-ietf-moq-transport.md
afrind: You are missing the second half of that sentence...

> ...if the publisher did not include DYNAMIC_GROUPS=1 when establishing the subscription

The subscriber sends PUBLISH_OK in response to PUBLISH.  Yes I believe this is correct.

draft-ietf-moq-transport.md
afrind: See #781 - in effect, this parameter is part of the track state and has to be preserved.  We may end up with a slightly different design to address extensibility in which case we can migrate this parameter to that design.  Relays already have to track state about existing subscriptions so I can't believe that an additional bit is harmful.

draft-ietf-moq-transport.md
fluffy: Tim, it seems each relay would individually know this with no need to synchronize it across the relay mesh

draft-ietf-moq-transport.md
fluffy: There sort of is the ACK in the sense that all of the control messages are over a reliable transport but if some relay looses it, I agree but I don't see this as a problem. I think systems that use new groups at all are likely to pick a periodic update where they do a new group even if none were requested and it will be easy to see this problem in metrics.

draft-ietf-moq-transport.md
fluffy: I was imagining this happen on each individual relay with no sharing the info across the mesh of relays.

draft-ietf-moq-transport.md
fluffy: The key issue here is not if the original publisher supports it but if all the relays on the path support it. 

It might make more sense if we described it like publisher sets flag "FEATURE_NGR_SUPPORTED" to true and it is passed across the subscription path, any relays that does not support MUST change it to false, then subscriber can only use the feature it it is supported. It might make sense to make that a more generic mechanism that could be used to negotiate other features across the relay path.

draft-ietf-moq-transport.md
TimEvens: @fluffy , I believe that makes an assumption that the inter-relay protocol is MOQT hop-by-hop. My understanding is that MOQT is not defining inter-relay protocol.  Has that changed? 

My two-cents are that this is a publisher capability exchange for subscribers, which is the first that needs to be conveyed to every subscriber via `SUBSCRIBE_OK`.  This is new and is pretty significant. If MOQT is going to start advertising capabilities of publishers to subscribers, then wouldn't there be more in the very near future? Either by this draft or by subsequent follow-up drafts that add publisher capabilities (aka params) that need to be made available to subscribers?  I'm supported of having that as a feature, but I do feel that this isn't solving how we sustain one or more publisher capability advertisements to subscribers. How we do that also needs to take into account multi-publisher for the same track.

draft-ietf-moq-transport.md
TimEvens: Subsequent RFC/drafts to this protocol will likely add more params. Those params will end up going to the publisher, no different than new group request. Why go to this level just for new group? Sounds like the live-stream folks are interested in param/update filtering.  Basically, notify me of X but not Y.  Do not include params that I do not support.  This is classic capabilities exchange on params.

draft-ietf-moq-transport.md
TimEvens: An issue here is that the publisher may not support NGR and the subscriber is unaware of that, resulting in repeated attempts for NGR from the subscriber to no avail. This causes a lot of churn for both the relays and publisher. 

The NGR capability being advertised appears like a good idea to notify the subscriber, but IMHO, this opens a bit of pandora's box. 

NGR is just one capability, there will likely be many more capabilities that are similar to NGR over time. 

MOQT should just facilitate the exchange of capabilities of a publisher to subscribers.

If the capabilities were made available to the subscriber, then the subscriber can act accordingly. If the edge relay can provide in-depth filtering (aka support the capability inspection and enforcement), then it can filter a subscriber sending NGR or other params that shouldn't result in notifying the publisher. 

I see this as an edge relay function, which requires publisher capabilities to be known.  If we go this route, then `SUBSCRIBE_UPDATE` params  need to be marked on where they should go.   Some params are only meant for auth and to go to auth servers, others are meant for the local relay to control/influence the direct subscription delivery, some are notifications to the publisher, ...

draft-ietf-moq-transport.md
TimEvens: Talking with @suhasHere, he indicated that the intent of group ID is another form of dampening. Basically, upon NGR, the relay will go into pending state for the NGR regardless of the group ID. Additional NGRs from other subscribers will be ignored if their NGR groupId is `less than or equal` to the current NGR group ID that is already pending.  Upon receiving `group != previous group`, the relay clears the state of any NGRs that were received during the initial pending request.  Thus allowing a new NGR to start the process again.

This works when NGR GIDs are high to low, but it gets really ugly when the NGR GID is low to high. 

For example: 

Subscriber 1 NGR group_id=100 (**forwarded to publisher with GID=100**)
Subscriber 2 NGR group_id=5 (ignored, less than previous pending)
Subscriber 3 NGR group_id=90 (ignored, less than previous pending)
Subscriber 4 NGR group+id=101 (**forwarded to publisher with GID=101**)

Publisher receives NGRs for group 100 and group 101.

Having the group id introduces complexity, causing arch self inflicted pain, which Suhas has slides on to describe why subscriber 4 NGR needs to be sent. Performing this function to send multiple NGRs while one is still pending  exposes, IMHO, a major DoS issue where there are no guards around max number of NGRs from a subscriber, allowing a subscriber or many subscribers to send unlimited `NGR++`.  

For example, a for loop of group_id++ sending NGRs.  Based on the current text in the PR, those all would be forwarded to the publisher and would require state updates for each in the relay.The publisher would have to deal with the churn of a for loop of NGRs... likely causing a publisher DoS. 

Nowhere in the PR text/spec does it say that the original publisher SHOULD NOT use the group id in the request. If the publisher application actually uses the group id sent in the NGR, then it would be feasible that the publisher may ignore ones that are either invalid due to current state or way off into the future.  If the publisher ignores it for one reason or another, it would cause a DoS in that relays would get stuck in PENDING state for group change that may not happen. The publisher may simply have dropped the invalid NGR, but would have accepted another that had a reasonable GID.  For example, had `NGR group_id=5` been sent the publisher would have responded with a new group.

draft-ietf-moq-transport.md
TimEvens: I believe I'm on the same page as @gwendalsimon  in that NGR with or without GID is application/publisher specific. Timing is a factor as well as GID if GID is provided. Receiving a NGR too soon might result in no group change. Receiving a NGR for a GID that is too high or to low might also be ignored. Receiving a NGR that is in the correct range would result in actual group change. 

Having GID as part of NGR opens the door for all kinds of other uses, such as allowing subscribers to quiesce on a new group that might be +2 of HEAD.   If MOQT doesn't want the applications/publishers do to that, then it should be very clear that it MUST or SHOULD NOT do that.

draft-ietf-moq-transport.md
TimEvens: @afrind, "since we forward DYNAMIC_GROUPS" is an assumption for inter-relay communication. AFAIK, MOQT is not defining inter-relay protocol.  As mentioned in other comments, DYNAMIC_GROUPS is a publisher edge relay side capability/filter. It is also only known when a publisher connects. We support lingering subscribers, where subscribers join before the publisher. We do this of course to support meetings where publishers (aka participant of a meeting) disconnects and reconnects, which will not result in resubscribe churn. Item 2 is more challenging in terms of DoS, which I mentioned in other comments. 

Talked with Suhas about this and it's a bit more of a problem than what I originally wrote. He is supposed to do an update to this PR to fix some of the issues with ZERO value regarding when we forward NGR (updates) or not. For example, `SUBSCRIBE_UPDATE` NGR **CANNOT** have a **ZERO** value for group ID.  Only the `SUBSCRIBE` can have that. Allowing ZERO for `SUBSCRIBE_UPDATE` NGR using the text in the PR would result in applications just using ZERO all the time. After all, as an application subscriber developer, where does it say I can't use zero all the time and why would I care to use a group ID when ZERO is allowed and works to start a new group?   Per the text in the PR, this would result in flooding updates to the publisher(s) regardless of `less than or equal to last NGR`.

draft-ietf-moq-transport.md
TimEvens: Reliable transmission of control messages is fine and not an issue. The issue is about how does the subscriber know that the NGR failed and should be reattempted and how to clear the PENDING state of NGRs on relays when new group doesn't happen?  There are many scenarios where NGRs will be ignored, suppressed, or lost.  The subscriber is completely in the dark and is left only to fire off attempt after attempt to get a new group started.  Relay is also completely in the dark if the publisher got the message and what it is doing about it. The relay just gets stuck in a PENDING state that results in suppression of NGRs, which likely will be a problem.

yekuiwang:
>> There is lack of clarity mentioned for how the relay should handle multiple new group requests with different group ID's, which could be zero as well. If current group is 10 and subscriber 1 asks for 20 and subscriber 2 asks for 8 and subscribe 3 asks for 50, which new-group request group-id do we send to the original publisher?

> It's pretty clear - you send if it's larger than the largest one you've sent and largest than largest group. So in your example, it would send 20 and 50, but not 8.

Just try to understand better: If the current group is 10, to respond to valid NGRs, shouldn't it be better to respond by sending just one group and the one group to send is group 11?

draft-ietf-moq-transport.md
afrind: > Those params will end up going to the publisher, no different than new group request.

Similar to what Victor observed during the last interim, SUBSCRIBE and SUBSCRIBE_UPDATE params are all hop-by-hop, since there's no expectation that a new SUBSCRIBE will be sent past the first relay.  This PR defines special handling of NEW_GROUP_REQUEST, and uses SUBSCRIBE_UPDATE to avoid additional message machinery.  There's not a clear path for how we would add another param or message like this in subsequent MoQT versions, since it would require every hop on the path to have support for it.

I think @mzanaty and @wilaw looked hard at how to build something more generic or extensible, but ultimately did not produce a proposal.

draft-ietf-moq-transport.md
afrind: > How we do that also needs to take into account multi-publisher for the same track.

I think in this PR, if there are two publishers with different DYNAMIC_GROUPS values for the same Track, I think it still works: one of them would get the NGR and one wouldn't.  I'm not sure that's likely though in the cases we care about.

draft-ietf-moq-transport.md
afrind: > The key issue here is not if the original publisher supports it but if all the relays on the path support it.

The relay handling described here is a MUST for Relays, so all Relays on the path do support it.  It's really a question if the Original Publisher supports it.

> An issue here is that the publisher may not support NGR and the subscriber is unaware of that

That's why DYNAMIC_GROUPS needs to be conveyed in SUBSCRIBE_OK/PUBLISH.  If we feel like Params is not a good fit here, consider it a track-level Extension (see #1176).

Regarding Generic Capabilities Exchange:

I think additional Publisher -> Subscriber capabilities can be conveyed in future versions without any additional help from Relays.  However, Subscriber -> Original Publisher messages need some kind of relay support which is tricky.  I'd like to see a concrete proposal and PR to fully understand.

draft-ietf-moq-transport.md
afrind: Ok, I see the problem:  If current group is 10 and someone sends NGR=1000000, the OP ignores it and keeps sending the current group.  Then legit subscribers cannot send NGR=11.

My vision is that #1182 / #1159 will soon positively or negatively acknowledge SUBSCRIBE_UPDATE.  In that case, the OP can send a REQUEST_ERROR for invalid NGR, which could clear the state at the relay.

draft-ietf-moq-transport.md
afrind: > "since we forward DYNAMIC_GROUPS" is an assumption for inter-relay communication.

In MOQT we describe a single Relay.  If you have a mesh of Relays that act as a single Relay, then it needs to conform to the requirements in the spec.  So yes, in a way.

> MOQT is not defining inter-relay protocol

100% agree, but see above.  You can use any inter-relay protocol you want.  The requirement is if you are a Relay, you include DYNAMIC_GROUPS in the SUBSCRIBE_OK/PUBLISH if you received it.

> We support lingering subscribers, where subscribers join before the publisher.

I'm pretty sure sending the SUBSCRIBE_OK before having an upstream subscription is now disallowed.  If the subscriber shows up first it should use SUBSCRIBE_NAMESPACE, and then it will get a PUBLISH.

> Per the text in the PR, this would result in flooding updates to the publisher(s) regardless of less than or equal to last NGR.

I'm not sure restricting 0 to SUBSCRIBE will prevent a DoS of the OP, since an attacker can just send SUBSCRIBE.  But also, if SUBSCRIBE has 0 but there's already a subscription, the relay needs to send SUBSCRIBE_UPDATE with NGR=0, so I don't think we can restrict it.

draft-ietf-moq-transport.md
afrind: See above comment regarding REQUEST_OK / REQUEST_ERROR as a solution to #1003

draft-ietf-moq-transport.md
suhasHere: > Ok, I see the problem: If current group is 10 and someone sends NGR=1000000, the OP ignores it and keeps sending the current group. Then legit subscribers cannot send NGR=11.

I don't think that's true.  A new subscriber can still send 11 and it will work .

afrind: > A new subscriber can still send 11 and it will work .

How?  11 < 10000000 and the relay has not seen a new group to clear the old request.

suhasHere: The OP supporting NGR cannot ignore the request. It will send the next group based on its application logic (may be delayed) and that should clear the one for 10000000.  

Yes something like this can cause subscribers to not get new group for a while until OP sends a new group, but it cannot block things indefinitely. 

However, if OP decides to never start a new group in which case the expected behavior on the subscriber is same regardless of what group Id is carried in  NGR.  This seems like application level issue and should not to be dealt at moq level and I think i did clarify this to Tim as well.

suhasHere:
> Just try to understand better: If the current group is 10, to respond to valid NGRs, shouldn't it be better to respond by sending just one group and the one group to send is group 11?

The OP will respond to whatever is the next group.  Yes it might just send 11 and that clears all the pending state in the relays , since any new group started should clear relay NGR-pending states

draft-ietf-moq-transport.md
suhasHere: In this case a publisher saying i am capable of NGR is an optimization for the first hop and I am bit unsure of generic capabilities exchange as it defines its own sub protocol inside another protocol. Agree with @afrind a more concrete proposal/issue will help understand better

afrind:

We didn't have as much time to discuss open questions, and @englishm expressed interest in presenting in Toronto on an application rather than MoQT layer scheme.  Please continue to review, but we will keep this open at least until then.

ianswett (changes requested):
Removing my approval until we can discuss other options such as #1219

fluffy:
In the Toronto meeting, we discussed wide range of alternatives to this and ways to make this better. The general consensus at meeting seemed to be this looks like best approach and we should go with it.

ianswett (changes requested):
I am not sure this PR actually does what we'd like in some cases.  In particular, if you recently produced a Group, is there a way to say "Just start at the past Group, don't wait for a new one?"  I don't see a concept of time in this PR.  The OP can't afford to fail the SUBSCRIBE_UPDATE, because that fails the subscription (or will with new PRs).

suhasHere:
New informational text seems pretty good and provides good context.

draft-ietf-moq-transport.md
```diff
@@ -985,6 +985,38 @@ than specified in `Start`.
...
+joining a Track might detect that it is more efficient to request the Original
+Publisher create a new group than issue a Joining FETCH.  Publishers indicate a
+Track supports dynamic group creation using the DYNAMIC_GROUPS parameter
+({{dynamic-groups}}).
+
+One possible subscriber pattern is to SUBSCRIBE to a Track using Filter Type
+`Largest Object` and observe the `Largest Location` in the response.  If the
+Object ID is below the application's threshold, the subscriber sends a FETCH for
+the beginning of the Group.  If the Object ID is above the threshold and the
+Track supports dynamic groups, it sends a SUBSCRIBE_UPDATE message with the
```
```suggestion
Track supports dynamic groups, the subscriber sends a SUBSCRIBE_UPDATE message with the
```

draft-ietf-moq-transport.md
```diff
@@ -985,6 +985,38 @@ than specified in `Start`.
...
+While some publishers will deterministically create new Groups, other
+applications might want to only begin a new Group when needed.  A subscriber
+joining a Track might detect that it is more efficient to request the Original
+Publisher create a new group than issue a Joining FETCH.  Publishers indicate a
+Track supports dynamic group creation using the DYNAMIC_GROUPS parameter
+({{dynamic-groups}}).
+
+One possible subscriber pattern is to SUBSCRIBE to a Track using Filter Type
+`Largest Object` and observe the `Largest Location` in the response.  If the
+Object ID is below the application's threshold, the subscriber sends a FETCH for
```
ianswett, line 1011:
```suggestion
Object ID is below the application's threshold, the subscriber sends a joining FETCH for
```

draft-ietf-moq-transport.md
```diff
@@ -985,6 +985,38 @@ than specified in `Start`.
...
+the beginning of the Group.  If the Object ID is above the threshold and the
+Track supports dynamic groups, it sends a SUBSCRIBE_UPDATE message with the
+NEW_GROUP_REQUEST parameter equal to the Largest Location's Group, plus one (see
+{{new-group-request}}).
+
+Another possible subscriber pattern is to send a SUBSCRIBE with Filter Type
```
ianswett, line 1017: Can you add a bit more text on what this does?

draft-ietf-moq-transport.md
```diff
@@ -1760,6 +1792,56 @@ simultaneously.
...
+
+#### NEW GROUP_REQUEST Parameter {#new-group-request}
+
+The NEW_GROUP_REQUEST parameter (parameter type 0x22) MAY appear in PUBLISH_OK,
+SUBSCRIBE or SUBSCRIBE_UPDATE.  It is an integer representing the largest Group
+ID in the Track known by the subscriber, plus 1. A value of 0 indicates that the
+subscriber has no Group information for the Track.  A subscriber MUST NOT send
+this parameter in PUBLISH_OK or SUBSCRIBE_UPDATE if the publisher did not
+include DYNAMIC_GROUPS=1 when establishing the subscription.  A subscriber MAY
+include this parameter in SUBSCRIBE without foreknowledge of support.  If the
```
ianswett, line 1817: I don't think this behavior is a good idea, since it's a silent failure.

ianswett (commented):
A few more suggestions and comments.

draft-ietf-moq-transport.md
afrind: In the SUBSCRIBE/SUBSCRIBE_OK/FETCH flow -- it doesn't need to be a Joining FETCH, since you will know the correct Largest Object.  It could be a Joining FETCH if we want to document a PUBLISH/PUBLISH_OK/FETCH flow.

draft-ietf-moq-transport.md
afrind: It's not a silent failure -- SUBSCRIBE_OK will lack DYNAMIC_GROUPS, and the subscriber will know that they need to FETCH or wait.

draft-ietf-moq-transport.md
```diff
@@ -985,6 +985,38 @@ than specified in `Start`.
...
+One possible subscriber pattern is to SUBSCRIBE to a Track using Filter Type
+`Largest Object` and observe the `Largest Location` in the response.  If the
+Object ID is below the application's threshold, the subscriber sends a FETCH for
+the beginning of the Group.  If the Object ID is above the threshold and the
+Track supports dynamic groups, it sends a SUBSCRIBE_UPDATE message with the
+NEW_GROUP_REQUEST parameter equal to the Largest Location's Group, plus one (see
+{{new-group-request}}).
+
+Another possible subscriber pattern is to send a SUBSCRIBE with Filter Type
+`Next Group Start` and NEW_GROUP_REQUEST equal to 0.
```
```suggestion
Another possible subscriber pattern is to send a SUBSCRIBE with Filter Type
`Next Group Start` and NEW_GROUP_REQUEST equal to 0.  The value of
DYNAMIC_GROUPS in SUBSCRIBE_OK will indicate if the publisher supports dynamic
groups. A publisher that does will begin the next group as soon as practical.
```

draft-ietf-moq-transport.md
```diff
@@ -1760,6 +1792,56 @@ simultaneously.
 If the EXPIRES parameter is 0 or is not present in a message, the subscription
 does not expire or expires at an unknown time.
 
+#### DYNAMIC GROUPS Parameter {#dynamic-groups}
+
+The DYNAMIC_GROUPS parameter (parameter type 0x20) MAY appear in PUBLISH or
+SUBSCRIBE_OK.  Values larger than 1 are a Protocol Violation.  When the value is
+1, it indicates that the subscriber can send a NEW_GROUP_REQUEST parameter in
+PUBLISH_OK or SUBSCRIBE_UPDATE for this Track.
```
```suggestion
The DYNAMIC_GROUPS parameter (parameter type 0x20) MAY appear in PUBLISH or
SUBSCRIBE_OK.  Values larger than 1 are a Protocol Violation.  When the value is
1, it indicates that the subscriber can request the Original Publisher to start
a new Group by including the NEW_GROUP_REQUEST parameter in PUBLISH_OK or
SUBSCRIBE_UPDATE for this Track.
```

draft-ietf-moq-transport.md
ianswett: Good point.

draft-ietf-moq-transport.md
```diff
@@ -1760,6 +1794,57 @@ simultaneously.
...
+#### NEW GROUP_REQUEST Parameter {#new-group-request}
+
+The NEW_GROUP_REQUEST parameter (parameter type 0x22) MAY appear in PUBLISH_OK,
+SUBSCRIBE or SUBSCRIBE_UPDATE.  It is an integer representing the largest Group
+ID in the Track known by the subscriber, plus 1. A value of 0 indicates that the
+subscriber has no Group information for the Track.  A subscriber MUST NOT send
+this parameter in PUBLISH_OK or SUBSCRIBE_UPDATE if the publisher did not
+include DYNAMIC_GROUPS=1 when establishing the subscription.  A subscriber MAY
+include this parameter in SUBSCRIBE without foreknowledge of support.  If the
+publisher does not support dynamic Groups, it ignores the parameter in that
```
ianswett: Is this original publisher?  Because this is required for relays, correct?

```suggestion
original publisher does not support dynamic Groups, it ignores the parameter in that
```

draft-ietf-moq-transport.md
ianswett: Fair point.

draft-ietf-moq-transport.md
```diff
@@ -1760,6 +1794,57 @@ simultaneously.
...
+Relays MUST preserve the value of this parameter received from an upstream
+publisher in SUBSCRIBE_OK or PUBLISH when sending these messages to downstream
+subscribers.
+
+#### NEW GROUP_REQUEST Parameter {#new-group-request}
+
+The NEW_GROUP_REQUEST parameter (parameter type 0x22) MAY appear in PUBLISH_OK,
+SUBSCRIBE or SUBSCRIBE_UPDATE.  It is an integer representing the largest Group
+ID in the Track known by the subscriber, plus 1. A value of 0 indicates that the
+subscriber has no Group information for the Track.  A subscriber MUST NOT send
```
ianswett, line 1814: Is it valid to combine this with Forward=0?

ianswett (approved):
This isn't the most elegant mechanism, but does seem workable.

draft-ietf-moq-transport.md
```diff
@@ -1760,6 +1794,57 @@ simultaneously.
...
+this parameter in PUBLISH_OK or SUBSCRIBE_UPDATE if the publisher did not
```
ianswett, line 1815: You might send SUBSCRIBE_UPDATE before receiving SUBSCRIBE_OK, but I can infer that you mean the text to cover that case?

draft-ietf-moq-transport.md
afrind: It seems a little mean to send NGR Forward=0, but not sure if we want to make it an error.

draft-ietf-moq-transport.md
afrind: Good catch, can you suggest the change you want, if any?  Just allow it in SUBSCRIBE_UPDATE and the publisher can ignore?

draft-ietf-moq-transport.md
afrind: This text applies to any publisher (including relays).  eg: 

```
P->R: PUBLISH (DYNAMIC_GROUPS=0)
// track discovery out of band
S->R: SUBSCRIBE (NGR)
// relay ignores NGR
```

draft-ietf-moq-transport.md
ianswett: I'm ok with leaving this as is, but it is odd.

draft-ietf-moq-transport.md
ianswett: I guess my point is this is a required feature in the core draft, so it's really the original publisher that's going to support this, correct?  And anyone downstream of them is just reflecting that value?

draft-ietf-moq-transport.md
```diff
@@ -1760,6 +1794,57 @@ simultaneously.
...
+include DYNAMIC_GROUPS=1 when establishing the subscription.  A subscriber MAY
```
ianswett, line 1816:
```suggestion
include DYNAMIC_GROUPS=1 when establishing the subscription.  As such, if a
SUBSCRIBE_UPDATE is sent prior to receiving a SUBSCRIBE_OK, the SUBSCRIBE_UPDATE
MUST NOT include a NEW_GROUP_REQUEST param. A subscriber MAY
```

afrind: The publisher can't always tell if the SUBSCRIBE_UPDATE was sent before/after SUBSCRIBE_OK.  So this MUST is unenforceable, but I'm ok with that if others are?

draft-ietf-moq-transport.md
afrind: Ok, I see.

draft-ietf-moq-transport.md
```diff
@@ -985,6 +985,40 @@ than specified in `Start`.
 An endpoint that receives a filter type other than the above MUST close the
 session with a `PROTOCOL_VIOLATION`.
 
+### Joining an Ongoing Track
+
+The MOQT Object model is designed with the concept that the beginning of a Group
+is a joint point, so in order for a subscriber to join a Track, it needs to
```
vasilvv: `a join point`?

draft-ietf-moq-transport.md
```diff
@@ -985,6 +985,40 @@ than specified in `Start`.
...
+request an existing or future Group.  Different applications will have different
```
vasilvv: request an existing group or wait for a new one?

vasilvv (approved):

draft-ietf-moq-transport.md
```diff
@@ -985,6 +985,40 @@ than specified in `Start`.
...
+approaches for when to begin a new Group.
```
```suggestion
is a join point, so in order for a subscriber to join a Track, it needs to
request an existing Group or wait for a future Group.  Different applications
will have different approaches for when to begin a new Group.
```

Issue #1076, opened 2025-07-11
OBJECT_DATAGRAM_STATUS types weren't updated everywhere


martinduke:
Sec 9 says the types are 0x4 and 0x5. 

9.3.2 says they're 0x2 and 0x3, which is wrong.

PR #1082, opened 2025-07-14
OBJECT_DATAGRAM_STATUS

RichLogan:
Fixes: #1076 

I think Object Datagram Status might not have been updated with the other values moved? Added a table and altered the text to reflect the values given above. 

Maybe this is worth another issue, but I see that extensions being declared as present with a length of 0 is sometimes a protocol violation and sometimes not. (Datagram vs Stream). Not sure that's intentional?

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3330,7 +3330,7 @@ There are 2 defined Type values for OBJECT_DATAGRAM_STATUS:
 |------|------------|
 
 The LSB of the type determines if the Extensions Headers Length and Extension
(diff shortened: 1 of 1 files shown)


afrind:
> Maybe this is worth another issue, but I see that extensions being declared as present with a length of 0 is sometimes a protocol violation and sometimes not. (Datagram vs Stream). Not sure that's intentional?

It is intentional.  It's not an error on a stream to indicate extensions are present but give a size 0, because multiple objects can appear in a stream and some may have extensions while others don't.  It is an error for datagrams, because only a single object is allowed.

afrind (approved):

Issue #335, opened 2023-11-09
Unique track token


vasilvv:
(I thought I have proposed this before, but I can't find the corresponding issue)

The proposal is to have the publisher (the original one) generate a random token blob whenever it starts publishing a track. This allows us to deal with situations where the publisher for the track changes.  One such common situation is a publisher restart (related issues: #225, #204, #79): this allows us to tell apart the same binary reconnecting (meaning it's a continuation of the same stream) from the binary crashing and restarting (meaning it's a new stream that's a semantic continuation of the previous stream with the same name).

(this is in some sense similar to how HTTP uses ETag)

kixelated:
+1 This token would inform a relay if two ANNOUNCEs are cooperative and share a cache.

suhasHere:
I am but worried that we are solving a problem that isn't actually a problem or I am not seeing it clearly.

A publisher crashes/restarts and comes back - in either of these cases,  from a relay perspective there is objects for a given track with a small interruption in between. Yes the streams on which arrives changes in both the cases, but from a track and object state, the relays need to forward to matching subscribers and store the objects in the cache.

I am not sure if having this token, a relay would act differently in terms of how it caches or forwards things.

what am i missing here?

kixelated:
@suhasHere it depends on what you mean by "a publisher crashes". Do you mean:

- MoQ crashed?
- QUIC crashed?
- Encoder crashed?
- Process crashed?
- Host crashed?

The difference between them is the amount of state retained. You can resume existing subscriptions ONLY if the full MoQ and encoder state was retained, otherwise you'll wreck downstream subscribers.

```
# old process
ANNOUNCE webex.com/kixelated
track=video group=68 payload=h264
track=video group=69 payload=h264

# segfault

# new process
ANNOUNCE webex.com/kixelated
track=video group=0 payload=av1
track=video group=1 payload=av1
```

Any existing subscribers for `webex.com/kixelated/video` will either crash or ignore the new data. It is a fatal error if the relay considers both connections as bitwise identical in this instance.


What Victor is proposing is that by default, ANNOUNCE is mutually exclusive unless you can provide an identical token. Something like:

```
# old process
ANNOUNCE webex.com/kixelated token=abc123
OBJECT track=video group=68 payload=h264
OBJECT track=video group=69 payload=h264

# segfault

# new process
ANNOUNCE webex.com/kixelated token=abc123
track=video group=70 payload=h264
track=video group=71 payload=h264
```

[...]

PR #1087, opened 2025-07-16
Explain how to resume publishing after losing state

afrind:
Fixes: #335

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -574,12 +574,12 @@ which tracks are delivered via Parameters, but the actual content of the tracks
 does not depend on those parameters; this is in contrast to protocols like HTTP,
 where request headers can alter the server response.
 
-A publisher that loses state and intends to resume publishing on the same Track
-risks colliding with previously published Objects and violating the above
-requirements.  A publisher can handle this in application specific ways, for
-example:
+A publisher that loses state (e.g. crashes) and intends to resume publishing on
+the same Track risks colliding with previously published Objects and violating
+the above requirements.  A publisher can handle this in application specific
+ways, for example:
 
-1. Selecting a unique Track Name or Track Namespace whenever it resumes
+1. Select a unique Track Name or Track Namespace whenever it resumes
    publishing. For example, it can base one of the Namespace tuple fields on the
    current time, or select a sufficiently large random value.
 2. Resume publishing under a previous Track Name and Namespace and set the
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -575,7 +575,7 @@ does not depend on those parameters; this is in contrast to protocols like HTTP,
 where request headers can alter the server response.
 
 A publisher that loses state and intends to resume publishing on the same Track
-risks of colliding with previously published Objects and violating the above
+risks colliding with previously published Objects and violating the above
 requirements.  A publisher can handle this in application specific ways, for
 example:
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -573,6 +573,18 @@ which tracks are delivered via Subscribe Parameters, but the actual content of
 the tracks does not depend on those parameters; this is in contrast to
 protocols like HTTP, where request headers can alter the server response.
 
+A publisher that loses state and intends to continue publishing has multiple
+options to confirm to the above requirements.
+
+1. Select a unique Track Name or Track Namespace each time it resumes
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -573,6 +573,18 @@ which tracks are delivered via Subscribe Parameters, but the actual content of
 the tracks does not depend on those parameters; this is in contrast to
 protocols like HTTP, where request headers can alter the server response.
 
+A publisher that loses state and intends to continue publishing has multiple
+options to confirm to the above requirements.
```
```suggestion
A publisher that loses state and intends to resume publishing can handle this in application-specific ways, which are beyond the scope of this specification. However, the following options MAY be considered:
```

draft-ietf-moq-transport.md
```diff
@@ -573,6 +573,18 @@ which tracks are delivered via Subscribe Parameters, but the actual content of
...
+
+1. Select a unique Track Name or Track Namespace each time it resumes
+   publishing. For example, it can base one of the Namespace tuple fields on the
+   current time, or select a sufficiently large random value.
+2. Continue under a previous Track Name and Namespace and set the initial Group
+   ID to a unique value guaranteed to be larger than all previously used groups.
+   This can be done by choosing a Group ID based on the current time.
+3. Use TRACK_STATUS or another mechanism to determine the previous state,
+   querying for the largest published Group ID.
```
```suggestion
1. Choose a unique Track Name or Track Namespace whenever resuming publishing. 
    This can be achieved by basing one of the Namespace tuple fields on the current time 
    or selecting a sufficiently large random value.
2. Resume publishing under the previous Track Name and Namespace, setting the 
    initial Group ID to a unique value that is guaranteed to be larger than all previously 
    used Group IDs. For example, the Group ID can be derived from the current time.
3.  Utilize TRACK_STATUS or an equivalent mechanism to query the previous state and 
      determine the largest published Group ID.
```

suhasHere (commented):
Few editorial clarifications , otherwise looks good

draft-ietf-moq-transport.md
vasilvv: I'd elaborate more on the specific problem that we are solving here, which is that:

1. A publisher tries to publish things on track X.
2. It is possible for it to crash and restart publications.
3. If it naively just starts publishing on track X, it risks colliding with its previous instance, violating the uniqueness constraint of MoQ.

draft-ietf-moq-transport.md
vasilvv: Is TRACK_STATUS actually guaranteed to work here?

draft-ietf-moq-transport.md
afrind: I tried to incorporate this feedback, please take another look.

draft-ietf-moq-transport.md
afrind: Incorporated most of this feedback

draft-ietf-moq-transport.md
```diff
@@ -574,6 +574,21 @@ which tracks are delivered via Parameters, but the actual content of the tracks
 does not depend on those parameters; this is in contrast to protocols like HTTP,
 where request headers can alter the server response.
 
+A publisher that loses state and intends to resume publishing on the same Track
+risks of colliding with previously published Objects and violating the above
```
```suggestion
risks colliding with previously published Objects and violating the above
```

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -574,6 +574,21 @@ which tracks are delivered via Parameters, but the actual content of the tracks
 does not depend on those parameters; this is in contrast to protocols like HTTP,
 where request headers can alter the server response.
 
+A publisher that loses state and intends to resume publishing on the same Track
+risks colliding with previously published Objects and violating the above
+requirements.  A publisher can handle this in application specific ways, for
+example:
```
```suggestion
A publisher that loses state (e.g. crashes) and intends to resume publishing on
the same Track risks colliding with previously published Objects and violating
the above requirements.  A publisher can handle this in application specific
ways, for example:
```

draft-ietf-moq-transport.md
```diff
@@ -574,6 +574,21 @@ which tracks are delivered via Parameters, but the actual content of the tracks
...
+
+1. Selecting a unique Track Name or Track Namespace whenever it resumes
+   publishing. For example, it can base one of the Namespace tuple fields on the
+   current time, or select a sufficiently large random value.
```
```suggestion
1. Select a unique Track Name or Track Namespace whenever it resumes
   publishing. For example, it can base one of the Namespace tuple fields on the
   current time, or select a sufficiently large random value.
```

PR #1096, opened 2025-07-18
Remove early FIN from the definition of malformed tracks

vasilvv:
Other criteria for malformed tracks are semantic errors, this one is a syntax error; if this happens, the sender almost certainly has a serious bug and it is reasonable to tear down entire connection (as with other syntax errors).


afrind:
> Agreed that this should be a fatal session error, do we already have text for that?

We have a SHOULD:

> If a stream ends gracefully in the middle of a serialized Object, the session SHOULD be terminated with a PROTOCOL_VIOLATION.

While looking for this, I noticed a number of statements in the Objects and Datagrams section that needs to be updated for PUBLISH

Issue #908, opened 2025-04-02
Feature modularity


vasilvv:
Currently, in order to ship an MoQ client (or a server), the implementer has to implement all features.  This makes sense as a requirement for relays, or maybe as something that we generally expect of an MoQ library.

I am worried, however, that "features actually used" to "features defined and mandatory to implement" ratio, as it currently stands, will hinder the adoption of the protocol.  If I am asking a smart TV vendor to implement a MoQ video playback, I don't want to require them to implement SUBSCRIBE_ANNOUNCES, probably ANNOUNCE itself as well.

ROLE was an underbaked version of that, and MAX_SUBSCRIBE_ID is definitely a component, but I feel like we need to do more (both in terms of ability to make things optional, and editorially)

afrind:
Thinking of HTTP - not every *client* needs to implement POST, HEAD, CONNECT or OPTIONS if it only wants to send a GET and receive a response.  Servers need to be ready to handle any of these, even if it only wants to serve GET.  The difference here is - in HTTP the server can't randomly send simple clients messages, but in MoQT it can.

All the MoQT verbs (except TRACK_STATUS?) now define a "Not Supported" error code.  If the concern that simple clients want to tell the peer in advance of what verbs they support, so the implementations can be even simpler?  Is it too much to say, you need to at least be able to receive the known frame types and send a well formed _ERROR?

vasilvv:
> Servers need to be ready to handle any of these, even if it only wants to serve GET.

Yes, but HTTP serializes all methods (that are not CONNECT) the same way, which makes rejecting unknown methods easy.

> Is it too much to say, you need to at least be able to receive the known frame types and send a well formed _ERROR?

Well, I do not expect most people to implement those in reality.  So if you need a read-only MoQ video server, and you're writing a library for that case, sending ANNOUNCE to it will most likely result either in the server closing your connection with an "unsupported message" error, or (worse) it just ignoring it entirely.  Thinking more about it, closing the connection is probably fine for most of those cases, but the text currently implies that all endpoints implementing MoQ will implement all of those frames, which will not be the case in practice whether we want it or not.

ianswett:
I agree that ANNOUNCE and SUBSCRIBE_ANNOUNCES, at least as currently specified, aren't useful for many deployments.

I'd like to re-evaluate whether some parts of MoQT are optional once we've landed PUBLISH and such.

Sy0307:
[...]

Issue #847, opened 2025-03-24
Intermediaries that are not relays


vasilvv:
We should have some discussions about intermediaries that are not relays and _do_ modify tracks (transmuxers, etc): we both need to acknowledge those exist, and give practical considerations as to how those should be done (e.g. what to do with the track names).

fluffy:
I don't think there is anything to say about them other than they must meet all the rules of a publisher and subscriber. They will tend to subscribe to some tracks and publish some new tracks. What else would we say about them ?

afrind:
Maybe acknowledging their existence and clarifying that if they are changing track data they MUST change the track name, group ID or object ID also, because MoQT objects are immutable.

afrind:
I added a definition of "intermediary" in #1107 but didn't add extra text explaining that something like a transmuxer needs to publish on new tracks. I didn't mark it as closing this issue in case folks still want some explanatory text.

afrind:
The definition I added originally in #1107 was apparently controversial and got removed from the final PR.  It was:

```
Intermediary:

: An entity that is both a Publisher and a Subscriber, but not the Original
Publisher or End Subscriber.
```

I'm open to adding some explanatory text or considerations for these entities (e.g. transmuxers) because the immutable object model makes them a little non-obvious.

afrind:
I will say that there are also intermediaries that are "relay-esque" in that the do not e.g. mux tracks, but they also don't follow the Relay rules exactly, so are not captial R Relays.  I'm thinking particularly of reverse proxies (often within CDNs), that act as transport servers and transport _clients_ and use configuration rather than MOQT constructs to establish routing rules.  I've heard confusion from people who come from the HTTP proxy world about the difference between a proxy and relay.  Saying more here might help.

PR #1107, opened 2025-07-26
Explain Modularity of MOQT

afrind:
Fixes: #908

Added a new term "intermediary" which was our old Relay definition, and redefined Relay to mean "an intermediary that follows all the rules". An example of a non-relay intermediary might be an ingest proxy, that only supports PUBLISH related messages upstream and downstream.

Added a top-level "Modularity" section between "Sessions" and "Publishing and Receiving Tracks" - open to suggestions if there's a better place.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -843,8 +843,8 @@ using only SUBSCRIBE related messages.  Limited endpoints SHOULD respond to any
 unsupported messages with the appropriate `NOT_SUPPORTED` error code or close
 the session, rather than ignoring them.
 
-Relays MUST all MOQT messages defined in this document, as well as processing
-rules described in {{relays-moq}}.
+Relays MUST implement all MOQT messages defined in this document, as well as
+processing rules described in {{relays-moq}}.
 
 # Publishing and Retrieving Tracks
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -836,8 +836,7 @@ subscribing endpoints as well as highly functional Relays.  Non-Relay endpoints
 MAY implement only the subset of functionality required to perform necessary
 tasks.  For example, a limited media player could operate using only SUBSCRIBE
 related messages.  Limited endpoints SHOULD respond to any unsupported messages
-with the appropriate `NOT_SUPPORTED` error code or close the session, rather
-than ignoring them.
+with the appropriate `NOT_SUPPORTED` error code, rather than ignoring them.
 
 Relays MUST implement all MOQT messages defined in this document, as well as
 processing rules described in {{relays-moq}}.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -194,10 +194,14 @@ End Subscriber:
 
 : A subscriber that initiates a subscription and does not send the data on to other subscribers.
 
-Relay:
+Intermediary:
 
 : An entity that is both a Publisher and a Subscriber, but not the Original
 Publisher or End Subscriber.
+
+Relay:
+
+: An intermediary that conforms to all requirements in {{relays-moq}}.
 
 Upstream:
 
@@ -828,6 +832,19 @@ For example, BBR's PROBE_RTT state halves the sending rate for more than a round
 in order to obtain an accurate minimum RTT. Similarly, Reno halves it's congestion
 window upon detecting loss.  In both cases, the large reduction in sending rate might
 cause issues with latency sensitive applications.
+
+# Modularity
+
+MOQT defines all messages necessary to implement both simple publishing or
+subscribing endpoints as well as highly functional Relays.  Non-Relay endpoints
+and intermediaries MAY implement only the subset of functionality required to
+perform necessary tasks.  For example, a limited media player could operate
+using only SUBSCRIBE related messages.  Limited endpoints SHOULD respond to any
+unsupported messages with the appropriate `NOT_SUPPORTED` error code or close
+the session, rather than ignoring them.
+
+Relays MUST all MOQT messages defined in this document, as well as processing
+rules described in {{relays-moq}}.
 
 # Publishing and Retrieving Tracks
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -194,14 +194,10 @@ End Subscriber:
 
 : A subscriber that initiates a subscription and does not send the data on to other subscribers.
 
-Intermediary:
-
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -829,6 +833,19 @@ in order to obtain an accurate minimum RTT. Similarly, Reno halves it's congesti
...
+MOQT defines all messages necessary to implement both simple publishing or
+subscribing endpoints as well as highly functional Relays.  Non-Relay endpoints
+and intermediaries MAY implement only the subset of functionality required to
+perform necessary tasks.  For example, a limited media player could operate
+using only SUBSCRIBE related messages.  Limited endpoints SHOULD respond to any
+unsupported messages with the appropriate `NOT_SUPPORTED` error code or close
+the session, rather than ignoring them.
+
+Relays MUST all MOQT messages defined in this document, as well as processing
+rules described in {{relays-moq}}.
```
```suggestion
Relays MUST implement all MOQT messages defined in this document, as well as
processing rules described in {{relays-moq}}.
```

draft-ietf-moq-transport.md
```diff
@@ -194,11 +194,15 @@ End Subscriber:
 
 : A subscriber that initiates a subscription and does not send the data on to other subscribers.
 
-Relay:
+Intermediary:
 
 : An entity that is both a Publisher and a Subscriber, but not the Original
 Publisher or End Subscriber.
```
suhasHere: I feel this will encourage non moq-compliant relays to be standardized, which may not be the direction we want here ?

afrind: MOQT is a toolbox that can be used to create many kinds of endpoints.  We are currently building a live ingest proxy which is both a publisher and a subscriber, not an original publisher or end consumer, and not a "Relay".  It will never support matching client publishers and and client subscribers per the rules in the relay section - that doesn't make sense in our architecture.  Such proxies will exist, we might as well give them a name -- see also #847.

I don't agree that this definition will encourage non moq-compliant _relays_ to be standardized. Rather this PR explicitly states that to be a capital "R" Relay, you have to follow the rules.

gwendalsimon: Another example: implementation of A/B watermarking. An intermediary is a subscriber to two tracks (A and B) from the original-publisher and a Publisher of one track (mixing groups from the A and B tracks) to end-subscriber.

suhasHere: We have active speaker servers that doesn't implement all of relay functionality for a while, but it's hard for me to call them an intermediary since
 - that's an overloaded term 
 - hard to define security properties in term of moqt 
 - it is very application specific even though it has subscribe and publish joint functionality 

The active speaker server nodes subscribes to all audio and publish subset of them based on some Audio activity detection logic. 

Just from the comments here , we have 3 different applications doing subset of moqt protocols and terminate original publishes in some application specific way before distributing downstream 


I am not against defining new entities but this needs more thinking, I feel.

afrind: I think we're reading a awful lot into this word appearing in the "Terms and Definitions" section.  Since it's only used in one place here:

> Non-Relay endpoints and intermediaries MAY implement only the subset of functionality required to perform necessary tasks.  

I can delete the `intermediary` definition, and just say:

> Non-Relay endpoints MAY implement only the subset of functionality required to perform necessary tasks.  

I would update the Relay definition as:

```
Relay:

: An entity that is both a Publisher and a Subscriber, but not the Original
Publisher or End Subscriber, and conforms to all requirements in {{relays-moq}}.
```

suhasHere: @afrind  the suggested changes works for me.

draft-ietf-moq-transport.md
```diff
@@ -194,11 +194,15 @@ End Subscriber:
...
 
+Relay:
+
+: An intermediary that conforms to all requirements in {{relays-moq}}.
```
```suggestion
Relay:

: An entity that is both a Publisher and a Subscriber, is not the Original
Publisher or End Subscriber, and conforms to all requirements in {{relays-moq}}.
```

draft-ietf-moq-transport.md
```diff
@@ -829,6 +833,19 @@ in order to obtain an accurate minimum RTT. Similarly, Reno halves it's congesti
...
 
+# Modularity
+
+MOQT defines all messages necessary to implement both simple publishing or
+subscribing endpoints as well as highly functional Relays.  Non-Relay endpoints
+and intermediaries MAY implement only the subset of functionality required to
+perform necessary tasks.  For example, a limited media player could operate
+using only SUBSCRIBE related messages.  Limited endpoints SHOULD respond to any
+unsupported messages with the appropriate `NOT_SUPPORTED` error code or close
+the session, rather than ignoring them.
```
```suggestion
subscribing endpoints as well as highly functional Relays.  Non-Relay endpoints
MAY implement only the subset of functionality required to perform necessary
tasks.  For example, a limited media player could operate using only SUBSCRIBE
related messages.  Limited endpoints SHOULD respond to any unsupported messages
with the appropriate `NOT_SUPPORTED` error code or close the session, rather
than ignoring them.
```

fluffy (approved):
LGTM

vasilvv (approved):

draft-ietf-moq-transport.md
```diff
@@ -829,6 +829,19 @@ in order to obtain an accurate minimum RTT. Similarly, Reno halves it's congesti
...
 
+# Modularity
+
+MOQT defines all messages necessary to implement both simple publishing or
+subscribing endpoints as well as highly functional Relays.  Non-Relay endpoints
+MAY implement only the subset of functionality required to perform necessary
+tasks.  For example, a limited media player could operate using only SUBSCRIBE
+related messages.  Limited endpoints SHOULD respond to any unsupported messages
+with the appropriate `NOT_SUPPORTED` error code or close the session, rather
+than ignoring them.
```
```suggestion
with the appropriate `NOT_SUPPORTED` error code, rather than ignoring them.
```

draft-ietf-moq-transport.md
```diff
@@ -829,6 +829,19 @@ in order to obtain an accurate minimum RTT. Similarly, Reno halves it's congesti
...
 cause issues with latency sensitive applications.
 
+# Modularity
+
+MOQT defines all messages necessary to implement both simple publishing or
+subscribing endpoints as well as highly functional Relays.  Non-Relay endpoints
+MAY implement only the subset of functionality required to perform necessary
+tasks.  For example, a limited media player could operate using only SUBSCRIBE
+related messages.  Limited endpoints SHOULD respond to any unsupported messages
+with the appropriate `NOT_SUPPORTED` error code or close the session, rather
```
ianswett: I'd prefer to not mention closing the session here, because it's unclear when that would be desirable.

PR #1108, opened 2025-07-26
RFC: Rename SUBSCRIBE_DONE to PUBLISH_DONE

afrind:
The SUBSCRIBE name is awkward since it can now end a subscription initiated either via SUBSCRIBE or PUBLISH.  I think this makes a better symmetry: the publisher ends with PUBLISH_DONE and the subscriber ends with UNSUBSCRIBE.

My other thought was SUBSCRIPTION_DONE but I prefer PUBLISH_DONE.


ianswett (approved):
I don't have a strong opinion either way on this one.  I'm inclined to let this sit for a bit longer in case someone does really care.

fluffy (approved):

sharmafb (approved):

Issue #746, opened 2025-03-01
Way for FETCH error to say range was too large


fluffy:
This needs discussion but I think to limit DOS attacks, we are going to want a RELAY to be able to tell a SUBSCRIBE the range requested was too large and *also* give information about what sort of range would work.

ianswett:
I'm unsure when this would be an issue?  This seems not so different from an HTTP proxy serving a multi-GB file.

As we were discussing at the interim, I think text about how a FETCH shouldn't generate more than one upstream FETCH is likely important, though.

afrind:
Do you mean SUBSCRIBE, FETCH or both?  Title says fetch and description says SUBSCRIBE.

afrind:
> As we were discussing at the interim, I think text about how a FETCH shouldn't generate more than one upstream FETCH is likely important, though.

A single FETCH will likely generate one upstream FETCH per gap in the cached range.  Otherwise the cache will have to pull objects it already has in cache.

ianswett:
I haven't been able to come up with a resource exhaustion or attack issue with a large number of groups.  I'd like to better understand the problem we're trying to fix before adding a mechnaism.

afrind:
Discussed in Madrid:

The sentiment in the room was it's fine to add a specific error code.  Because Groups and Objects are variable in size, there wasn't a clear proposal what to advertise as a maximum in setup, so we're not planning to add that for now.

fluffy:
Let me make a clear proposal that we specify the limit in bytes.

afrind:
Unfortunately, a limit in bytes is not that useful.  As a Relay, I would need to scan my cache for the requested range to compute it's size before sending any objects.  If the cache has gaps, the Relay also needs to fill them before determining this.

afrind:
Further, a limit in bytes doesn't achieve the aim of telling the subscriber how they can change their request to get it to succeed, since the subscriber doesn't know the size of the groups/objects they are requesting.  For that, the only thing that really makes sense to me would be a limit expressed in Groups and/or Objects.

Following up on Ian's earlier comment here:

>  I'd like to better understand the problem we're trying to fix before adding a mechanism.

[...]

PR #1109, opened 2025-07-26
Add FETCH_TOO_LARGE error code

afrind:
Fixes: #746

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2654,7 +2654,7 @@ EXPIRED_AUTH_TOKEN (0x12):
 
 FETCH_TOO_LARGE (0x13):
 : The FETCH request could not be satisfied because the requested range
-  exceeds an implementation defined limit.
+  exceeds the publisher's current limit.
 
 
 ## FETCH_CANCEL {#message-fetch-cancel}


vasilvv (approved):

draft-ietf-moq-transport.md
```diff
@@ -2652,6 +2652,10 @@ MALFORMED_AUTH_TOKEN (0x10):
 EXPIRED_AUTH_TOKEN (0x12):
 : Authorization token has expired ({{authorization-token}}).
 
+FETCH_TOO_LARGE (0x13):
+: The FETCH request could not be satisfied because the requested range
+  exceeds an implementation defined limit.
```
ianswett: nit: It might be based on load/etc as well.

```suggestion
  exceeds the peer's current limit.
```

```suggestion
  exceeds the publisher's current limit.
```

afrind:
@fluffy 

> I think this error should include a way to provide the publishers limit.

See this comment in the linked issue:

https://github.com/moq-wg/moq-transport/issues/746#issuecomment-3102047904

> Because Groups and Objects are variable in size, there wasn't a clear proposal what to advertise as a maximum in setup, so we're not planning to add that for now.

fluffy:
I disagree - I don't think we should keep adding half backed crap into the draft. If we don't know how to make this error be of any use, we should not add it until we have figured that out. I think the limit should be specified in bytes. It does not matter that objects and groups are of unknown size, the applications knows the characteristics of its data and can figure that out a rough estimate.  This same problems exists for the publisher to even know when to send this error.

fluffy:
I understand it is not trivial to decide how to know specify in what way the request was too large but without that, this error is not useful. The application receiving it has no way to know how to react to it. Similarly the applications sending it has no way to know if it should even send this error. 

My concrete proposal is we make this clearly about the sum of the size of the objects in bytes and we add in the error the the max limit. This allows the software that received the error to know what to do to correct from the error. More importantly it allows the publisher that received the request to know on how to implement computing and sending the error.

afrind:
I left this comment in #746 also -- 

Unfortunately, a limit in bytes is not that useful. As a Relay, I would need to scan my cache for the requested range to compute it's size before sending any objects. If the cache has gaps, the Relay also needs to fill them before determining this.

Let's keep the discussion of how to communicate a limit over in the issue.

fluffy:
Added text in issue

fluffy:
Just want to add here people convinced that bytes is not the right thing.

In meeting decided that will close issue and remove PR.

Issue #896, opened 2025-03-29
Should receiving Unrequested objects be a session error?


vasilvv:
Draft says:

> If an endpoint receives an Object it never requested, it SHOULD terminate the session with a protocol violation.

Is the protocol we currently define actually sufficiently resistant to various race conditions that this recommendation will not lead to spuriously closed connections?  Should we switch it to "ignore"?

ianswett:
I think this SHOULD is likely fine given we say "never requested".

afrind:
I think with PUBLISH we might need to relax this.  A publisher might get overly exuberant and start sending a track before publish OK.

PR #1112, opened 2025-07-26
Receiving unrequested Objects

afrind:
Fixes: #896

Because of PUBLISH, Objects with unknown Track Alias can arrive.  This shouldn't be a mandatory session error.

Explain that only publishers with some knowledge of subscriber interest should begin sending objects before PUBLISH_OK.

Explain that Objects can be in flight after UNSUBSCRIBE or PUBLISH_ERROR, and these ought not be treated as a new Track Alias.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -864,6 +864,11 @@ FETCH. A subscriber MUST send exactly one PUBLISH_OK or PUBLISH_ERROR in
 response to a PUBLISH. The peer SHOULD close the session with a protocol error
 if it receives more than one.
 
+Publishers SHOULD only send Objects belonging to Tracks of interest to the
+subscriber, as indicated by a SUBSCRIBE, FETCH or PUBLISH_OK message.
+Publishers with out-of-band knowledge of subscriber interest MAY begin sending
+Objects on PUBLISH initiated subscriptions before receiving PUBLISH_OK.
+
 A subscriber keeps subscription state until it sends UNSUBSCRIBE, or after
 receipt of a SUBSCRIBE_DONE or SUBSCRIBE_ERROR. Note that SUBSCRIBE_DONE does
 not usually indicate that state can immediately be destroyed, see
@@ -2283,6 +2288,9 @@ PUBLISH Message {
 
 * Parameters: The parameters are defined in {{version-specific-params}}.
 
+A subscriber receiving a PUBLISH for a Track it does not wish to receive SHOULD
+send PUBLISH_ERROR with error code `UNINTERESTED`, and STOP_SENDING any streams
+associated with that subscription.
 
 ## PUBLISH_OK {#message-publish-ok}
 
@@ -3029,12 +3037,6 @@ the datagram.
 An endpoint that receives an unknown stream or datagram type MUST close the
 session.
 
-The publisher only sends Objects after receiving a SUBSCRIBE or FETCH.  The
-publisher MUST NOT send Objects that are not requested.  If an endpoint receives
-an Object it never requested, it SHOULD terminate the session with a protocol
-violation. Objects can arrive after a subscription or fetch has been cancelled,
-so the session MUST NOT be teriminated in that case.
-
 Every Track has a single 'Object Forwarding Preference' and the Original
 Publisher MUST NOT mix different forwarding preferences within a single track
 (see {{malformed-tracks}}).
@@ -3045,6 +3047,11 @@ To optimize wire efficiency, Subgroups and Datagrams refer to a track by a
 numeric identifier, rather than the Full Track Name.  Track Alias is chosen by
 the publisher and included in SUBSCRIBE_OK ({{message-subscribe-ok}} or PUBLISH
 ({{message-publish}}).
+
+Objects can arrive after a subscription has been cancelled.  Subscribers SHOULD
+retain sufficient state to quickly discard these unwanted Objects, rather than
+treating them as belonging to an unknown Track Alias.
+
 
 ## Objects {#message-object}
 


draft-ietf-moq-transport.md
```diff
@@ -864,6 +864,11 @@ FETCH. A subscriber MUST send exactly one PUBLISH_OK or PUBLISH_ERROR in
 response to a PUBLISH. The peer SHOULD close the session with a protocol error
 if it receives more than one.
 
+Publishers SHOULD only send Objects belonging to Tracks of interest to the
+subscriber, as indicated by a SUBSCRIBE, FETCH or PUBLISH_OK message.
+Publishers with out-of-band knowledge of subscriber interest MAY begin sending
+Objects on PUBLISH initiated subscriptions before receiving PUBLISH_OK.
```
suhasHere: Publish with forward=1 does allow a publisher to start sending messages without waiting for publish_ok message.  This seems to block that flow.

afrind: This does not block any flows.  

Presumably anyone using Publish with forward=1 has "out of band knowledge of subscriber interest".  Also, line 867 is a SHOULD, not a MUST, so you can flood the network with unrequested objects if you want to.

Maybe it should read "_likely_ subscriber interest"?

```suggestion
Publishers SHOULD only send Objects for Tracks explicitly requested by the
subscriber through a SUBSCRIBE or FETCH message, or acknowledged in a PUBLISH_OK
message. However, Publishers with out-of-band knowledge of likely subscriber
interest MAY start sending Objects on PUBLISH-initiated subscriptions before
receiving a PUBLISH_OK response.
```

suhasHere (approved):
Minor nits and seems to look good to me

draft-ietf-moq-transport.md
```diff
@@ -864,6 +864,12 @@ FETCH. A subscriber MUST send exactly one PUBLISH_OK or PUBLISH_ERROR in
 response to a PUBLISH. The peer SHOULD close the session with a protocol error
 if it receives more than one.
 
+Publishers SHOULD only send Objects for Tracks explicitly requested by the
+subscriber through a SUBSCRIBE or FETCH message, or acknowledged in a PUBLISH_OK
+message. However, Publishers with out-of-band knowledge of likely subscriber
```
fluffy: I think this is phrased the wrong way - this sounds like it say you should not do X unless you have a unicorn in which case you should do X. 

I think it would be clearer to describe all this as the applications using the relays knows if it makes sense for the publisher to start sending data before the publish OK or not. And it can do either as it makes sense for that applications use case. Relays MUST be prepared to receive data before a publish OK

draft-ietf-moq-transport.md
```diff
@@ -2283,6 +2289,9 @@ PUBLISH Message {
 
 * Parameters: The parameters are defined in {{version-specific-params}}.
 
+A subscriber receiving a PUBLISH for a Track it does not wish to receive SHOULD
+send PUBLISH_ERROR with error code `UNINTERESTED`, and STOP_SENDING any streams
```
fluffy: I'm confused about the stop sending half of this sentence - it seems to me a subscriber would not be sending any streams. I suspect there is just some sort of typo here or clarification needed

fluffy (changes requested):

draft-ietf-moq-transport.md
afrind: STOP_SENDING is sent by the receiver to tell the publisher to stop.

The subscriber sends STOP_SENDING on streams the publisher has opened associated with an unwanted Track/Subscription.

draft-ietf-moq-transport.md
```diff
@@ -864,6 +864,12 @@ FETCH. A subscriber MUST send exactly one PUBLISH_OK or PUBLISH_ERROR in
 response to a PUBLISH. The peer SHOULD close the session with a protocol error
 if it receives more than one.
 
+Publishers SHOULD only send Objects for Tracks explicitly requested by the
+subscriber through a SUBSCRIBE or FETCH message, or acknowledged in a PUBLISH_OK
+message. However, Publishers with out-of-band knowledge of likely subscriber
+interest MAY start sending Objects on PUBLISH-initiated subscriptions before
+receiving a PUBLISH_OK response.
+
```
```suggestion
Publishers MAY start sending Objects on PUBLISH-initiated subscriptions before
receiving a PUBLISH_OK response to reduce latency.  Doing so can consume
unnecessary resources in cases where the Subscriber rejects the subscription
with PUBLISH_ERROR or sets Forward State=0 in PUBLISH_OK.

```

draft-ietf-moq-transport.md
```diff
@@ -2283,6 +2289,9 @@ PUBLISH Message {
 
 * Parameters: The parameters are defined in {{version-specific-params}}.
 
+A subscriber receiving a PUBLISH for a Track it does not wish to receive SHOULD
+send PUBLISH_ERROR with error code `UNINTERESTED`, and STOP_SENDING any streams
+associated with that subscription.
```
```suggestion
send PUBLISH_ERROR with error code `UNINTERESTED`, and abandon reading any
publisher initiated streams associated with that subscription using a
STOP_SENDING frame.
```

suhasHere (approved):

suhasHere (approved):

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -864,6 +864,11 @@ FETCH. A subscriber MUST send exactly one PUBLISH_OK or PUBLISH_ERROR in
 response to a PUBLISH. The peer SHOULD close the session with a protocol error
 if it receives more than one.
 
+Publishers MAY start sending Objects on PUBLISH-initiated subscriptions before
+receiving a PUBLISH_OK response to reduce latency.  Doing so can consume
+unnecessary resources in cases where the Subscriber rejects the subscription
+with PUBLISH_ERROR or sets Forward State=0 in PUBLISH_OK.
```
vasilvv: I think we should call out explicitly the fact that if you do that, some of the objects might get lost due to server needing to limit the amount of buffered object that are not associated with any subscription.

```suggestion
Publishers MAY start sending Objects on PUBLISH-initiated subscriptions before
receiving a PUBLISH_OK response to reduce latency.  Doing so can consume
unnecessary resources in cases where the Subscriber rejects the subscription
with PUBLISH_ERROR or sets Forward State=0 in PUBLISH_OK. It can also result in
the Subscriber dropping Objects if its buffering limits are exceeded (see
{{datagrams}} and {{subgroup-header}}).
```

afrind: @vasilvv updated, PTAL

vasilvv (approved):

Issue #641, opened 2024-12-11
When can a relay send SUBSCRIBE_DONE with status SUBSCRIPTION_ENDED?


afrind:
1. Relay has a single client subscribed to a single track with no end group specified
2. A second client joins with a fixed end group (N)

Here the relay is responsible for generating the SUBSCRIBE_DONE towards client 2, but when?

A triggering event could be a) receiving an END_OF_GROUP status or b) seeing an object with group > N.

The relay could send it immediately when it sees a triggering event, but if it does so, it has to pick the Stream Count at that time, and cannot open any new subgroup streams for late arriving objects in groups <= N.

It could start a timer when it sees the triggering event, and send SUBSCRIBE_DONE with a final stream count when the timer expires.

It would be nice to have a short circuit in this case, but it requires the original publisher include a Stream Count with END_OF_GROUP status.  This is also a per-hop value due to DELIVERY_TIMEOUTs that may have expired, which is a little weird, since otherwise object status and content is end-to-end.

afrind:
> and it's closed all streams it's going to open.
 
The relay cannot know this without knowing how many streams the upstream is sending in the last group in SUBSCRIBE. 

eg, say subscribe ends at group 2

S->R: G2, SG0, O1
S->R: G2, SG10, O3
S->R, G2, SG20, EOG
S->R: G3, SG0, O1

Are there more subgroups in G2?

martinduke:
it's not super clean, but at some point it's going to decide to stop, and then it can send DONE.

ianswett:
It seems like there are a few exacerbating factors here to determine if you've received (and therefore sent) all the Objects in a group.

An End Of Group can help, but only if you've received all the Objects.  If you haven't, you don't know if there's a gap in Object IDs or if there are one or more subgroups you haven't received yet.

afrind:
Individual Comment:

> An End Of Group can help, but only if you've received all the Objects. If you haven't, you don't know if there's a gap in Object IDs or if there are one or more subgroups you haven't received yet.

You can solve this with stream counting like with do for SUBSCRIBE_DONE.

afrind:
I'd like to add an option to include a Stream Count with End Of Group.  I see two mechanisms:

1) Define a special extension header - the problem with this is that the current extension design doesn't work with 'hop-by-hop' headers, because the default is to forward unknown headers and this would be wrong.  We'd have to make this header mandatory to understand (at least to not forward), or tweak the extension design.

[...]

Issue #1026, opened 2025-05-15
Proposed new requirements for expires


afrind:
Comments on #1012 that are actually new proposed requirements to draft-10:
 
* Client has to renew before the expiry time
* After the time is exceeded, relay can just remove the subscribe state with no signaling to the subscriber. 
* Make expires mandatory not advisory
* We need to talk about how much before and randomness so they don't all happen at the same time.

ianswett:
I very much oppose "After the time is exceeded, relay can just remove the subscribe state with no signaling to the subscriber."  That's just bound to cause problems.

What does "renew" mean in this context?  Is it auth, or something else?

MoQ is already going over a reliable QUIC connection, so I don't think there's any need for a signal that one wants to continue receiving the subscription, as WebRTC requires (if I remember correctly?).

fluffy:
This really comes into the we need to figure out the security design and then this probably be easier to figure out what bits we need where.

fluffy:
WebRTC needed consent at stun level so if new device got IP addess of old machine, the packets for old machine would stop. I don't think we have any problems like that here given it is all in a QUIC connection that takes care of that issue for us.

fluffy:
I think there are two design direction we could go

1) server tells client to reauth and gives it some time to do it.

2) client does reauth before old auth expires. 

I lean towards 2 as it pushes the timers onto client not server.

afrind:
It occurred to me that `expires` has a use outside of auth.  If a subscriber uses a filter that produces no objects for some period of time, I think the relay will need to reap that subscription.  eg:

For example, if the track is currently at group=1000 and I subscribe with an Absolute Start=1000000.  I'm forcing the relay to maintain state indefinitely.  The expires field is still advisory, but perhaps useful?  The relay can send SUBSCRIBE_DONE + EXPIRED.  If that's the case, any SUBSCRIBE_UPDATE, even one that changes nothing, should probably extend the life of the subscription?

I think this relay "idle" termination is required in a case where the the subscription has an End Group.  Unless the track is trivial, the relay can't possibly know that no more objects are coming that will match the filter, so the relay needs to be able to terminate the subscription.  See #641.

ianswett:
From Toronto:

Publisher sent as today, means subscription will end after that much time, clients should refresh with a SUBSCRIBE_UPDATE if they want to keep the subscription active.  When that is sent, if there's new auth, they should include that.

[...]

PR #1113, opened 2025-07-26
Filters don't (usually) determine the end of subscription

afrind:
Fixes: #641

Maybe this text belongs somewhere else -- does it apply to non-Relay publishers?

Does this affect our plans for client side ABR track switching?  Clients will likely need logic to Unsubscribe after receiving the end of a group on a switched-from track.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1167,6 +1167,12 @@ aggregate upstream subscriptions can subscribe using the Largest Object
 filter to avoid churn as downstream subscribers with disparate filters
 subscribe and unsubscribe from a track.
 
+A subscriber remains subscribed to a Track at a Relay until it unsubscribes or
+the upstream publisher terminates the subscription, and all subgroups are
+accounted for. Subscriber filters do not affect when the subscription ends
+unless the Relay knows definitively that no additional Objects can arrive that
+match the filter.
+
 ### Graceful Subscriber Relay Switchover
 
 This section describes behavior a subscriber MAY implement


draft-ietf-moq-transport.md
```diff
@@ -1167,6 +1167,12 @@ aggregate upstream subscriptions can subscribe using the Largest Object
 filter to avoid churn as downstream subscribers with disparate filters
 subscribe and unsubscribe from a track.
 
+A subscriber remains subscribed to a Track at a Relay until it unsubscribes or
+the upstream publisher terminates the subscription, and all subgroups are
+accounted for. Subscriber filters do not affect when the subscription ends
+unless the Relay knows definitively that no additional Objects can arrive that
+match the filter.
```
suhasHere: I think i get the intent but I wonder if it helps reword it as below

```suggestion
Subscriber filters do not directly determine when a subscription ends. A subscription ends when the subscriber unsubscribes, the publisher terminates it, or, in the case of Relay publishers, when the Relay confirms that no further objects matching the filter can be received.
```

afrind: I want to leave the door open to a relay ending the subscription when everything is guaranteed delivered.  For example, subscribe range is all of group 10, and relay receives Group 10, Object 0, Status End of Group.

suhasHere: How about the updated text ?

suhasHere (approved):
Made some editorial suggestions, otherwise looks good to me

draft-ietf-moq-transport.md
afrind: I feel like we're saying the same thing.  I'll let Ian pick :)

```suggestion
the upstream publisher terminates the subscription. Subscription filters only
allow a Relay to end the subscription when it knows that no additional Objects
can match the filter.
```

ianswett: A third suggestion :)

ianswett (commented):
I appreciate the clarification and suggested a slight variation.

draft-ietf-moq-transport.md
afrind: I think you mean "only allow _the Relay_ to end the subscription when _it_" knows, otherwise looks great to me.

ianswett: Updated.

draft-ietf-moq-transport.md
```diff
@@ -1167,6 +1167,11 @@ aggregate upstream subscriptions can subscribe using the Largest Object
 filter to avoid churn as downstream subscribers with disparate filters
 subscribe and unsubscribe from a track.
 
+A subscriber remains subscribed to a Track at a Relay until it unsubscribes or
+the upstream publisher terminates the subscription. Subscription filters only
+allow a Relay to end the subscription when it knows that no additional Objects
+can match the filter.
+
```
```suggestion
A subscriber remains subscribed to a Track at a Relay until it unsubscribes or
its authorization expires or the upstream publisher terminates the subscription. 
Subscription filters only allow a Relay to end the subscription when it knows that 
no additional Objects can match the filter.
```

draft-ietf-moq-transport.md
```diff
@@ -1167,6 +1167,11 @@ aggregate upstream subscriptions can subscribe using the Largest Object
 filter to avoid churn as downstream subscribers with disparate filters
 subscribe and unsubscribe from a track.
 
+A subscriber remains subscribed to a Track at a Relay until it unsubscribes or
+the upstream publisher terminates the subscription. Subscription filters only
+allow a Relay to end the subscription when it knows that no additional Objects
```
fluffy: I don't like text like this - it implies that relays needs to do this and that would be a challenging to implement in a way that did not have edge case problems across a multi relay networks.  I'd rather not have this unless we are going to exactly specify it.

fluffy (changes requested):
I like to say that filters don't end subscriptions.

draft-ietf-moq-transport.md
afrind: Good point regarding expires - though I would rather not mention _auth_ specifically, since expires presently doesn't.  I will update this text. We can also revisit once we have discussed #1026.

draft-ietf-moq-transport.md
afrind: @fluffy I don't want to imply that relays are required to keep score and always end subscriptions this way, that's computationally infeasible.

Do you think we should prohibit it entirely?

draft-ietf-moq-transport.md
```diff
@@ -1167,6 +1167,11 @@ aggregate upstream subscriptions can subscribe using the Largest Object
...
+can match the filter.
```
```suggestion
A subscriber remains subscribed to a Track at a Relay until it unsubscribes, the
upstream publisher terminates the subscription, or the subscription expires (see
{{subscribe_ok}}).  A subscription with a filter can reach a state where all
possible Objects matching the filter have been delivered to the subscriber.
Since tracking this can be prohibitively expensive, Relays are not required or
expected to do so.
```

draft-ietf-moq-transport.md
```diff
@@ -1167,6 +1167,13 @@ aggregate upstream subscriptions can subscribe using the Largest Object
 filter to avoid churn as downstream subscribers with disparate filters
 subscribe and unsubscribe from a track.
 
+A subscriber remains subscribed to a Track at a Relay until it unsubscribes, the
+upstream publisher terminates the subscription, or the subscription expires (see
+{{subscribe_ok}}).  A subscription with a filter can reach a state where all
+possible Objects matching the filter have been delivered to the subscriber.
+Since tracking this can be prohibitively expensive, Relays are not required or
+expected to do so.
+
```
```suggestion
A subscriber remains subscribed to a Track at a Relay until it unsubscribes, the
upstream publisher terminates the subscription, or the subscription expires (see
{{message-subscribe-ok}}).  A subscription with a filter can reach a state where
all possible Objects matching the filter have been delivered to the subscriber.
Since tracking this can be prohibitively expensive, Relays are not required or
expected to do so.

```

afrind:
@suhasHere @fluffy I updated, please take a look.  I'm unsure if you are asking for normative language prohibiting a relay from ending a subscription if it can detect that the filter has been completely satisfied, so I left it somewhat open-ended here.

I can see two reasons why we'd want it to be normative:

1) We don't want subscribers to accidentally assume that the Relay will take care of this for them, when some relays will never do it, even for simple cases.

2) The subscriber could theoretically use the delivery of the last object matching a filter to send a subscribe update widening the filter, once we allow widening.  CC: @ianswett

afrind:
Mentioned briefly in the interim today but didn't receive strong signal other than "relays gonna do".

suhasHere (approved):

vasilvv (approved):

Issue #782, opened 2025-03-17
Add a User-Agent-like field to a list of well-known setup parameters


vasilvv:
This should be useful for debugging issues that can be potentially implementation-specific.

(for WebTransport, you can already use User-Agent, but this would let you know what JS-based library you're talking to)

PR #1114, opened 2025-07-26
Add a free-form SETUP parameter identifying the implementation

afrind:
Fixes: #782

I called it MOQT_AGENT since ths server can also send this.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3646,6 +3646,8 @@ Implementations are advised to use timeouts to prevent resource
 exhaustion attacks by a peer that does not send expected data within
 an expected time.  Each implementation is expected to set its own limits.
 
+TODO: Security Considerations of MOQT_IMPLEMENTATION parameter
+
 # IANA Considerations {#iana}
 
 TODO: fill out currently missing registries:
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1719,8 +1719,11 @@ MAX_AUTH_TOKEN_CACHE_SIZE parameter in SERVER_SETUP (or the default value of 0).
 
 #### MOQT IMPLEMENTATION
 
-The MOQT_IMPLEMENTATION parameter (Parameter Type 0x05) is a UTF-8 encoded
-string describing the sender's MOQT implementation.
+The MOQT_IMPLEMENTATION parameter (Parameter Type 0x05) identifies the name and
+version of the sender's MOQT implementation.  This SHOULD be a UTF-8 encoded
+string [RFC3629], though the message does not carry information, such as
+language tags, that would aid comprehension by any entity other than the one
+that created the text.
 
 
 ## GOAWAY {#message-goaway}
@@ -3646,7 +3649,7 @@ Implementations are advised to use timeouts to prevent resource
 exhaustion attacks by a peer that does not send expected data within
 an expected time.  Each implementation is expected to set its own limits.
 
-TODO: Security Considerations of MOQT_IMPLEMENTATION parameter
+TODO: Security/Privacy Considerations of MOQT_IMPLEMENTATION parameter
 
 # IANA Considerations {#iana}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1720,7 +1720,7 @@ MAX_AUTH_TOKEN_CACHE_SIZE parameter in SERVER_SETUP (or the default value of 0).
 #### MOQT IMPLEMENTATION
 
 The MOQT_IMPLEMENTATION parameter (Parameter Type 0x05) is a UTF-8 encoded
-string describing the software agent of the sender.
+string describing the sender's MOQT implementation.
 
 
 ## GOAWAY {#message-goaway}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1717,6 +1717,12 @@ parameter as Alias Type USE_VALUE.  A client MUST handle registration failures
 of this kind by purging any Token Aliases that failed to register based on the
 MAX_AUTH_TOKEN_CACHE_SIZE parameter in SERVER_SETUP (or the default value of 0).
 
+#### MOQT AGENT
+
+The MOQT_AGENT parameter (Parameter Type 0x05) is a UTF-8 encoded
+string describing the software agent of the sender.
+
+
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1717,6 +1717,12 @@ parameter as Alias Type USE_VALUE.  A client MUST handle registration failures
 of this kind by purging any Token Aliases that failed to register based on the
 MAX_AUTH_TOKEN_CACHE_SIZE parameter in SERVER_SETUP (or the default value of 0).
 
+#### MOQT AGENT
```
```suggestion
#### MOQT USER-AGENT
```

suhasHere: the word Agent is bit overloaded these days.  I am fine with what you have, but make it explcit as User-Agent felt better. either works

suhasHere (approved):

draft-ietf-moq-transport.md
LPardue: Calling both ends user-agent is going to be super confusing.

Agent is overloaded I agree.

How about something like "IMPLEMENTATION_HINT", since this is really an optional hint containing unstructured data about implementation details

afrind: Dang Agent is overloaded.  I agree "USER" doesn't belong in the name here - it may be in use when no users are involved.  I like IMPLEMENTATION, but don't love HINT.  True this is completely unauthenticated and people can and will lie.  Let me try MOQT_IMPLEMENTATION.

#bikeshed

draft-ietf-moq-transport.md
```diff
@@ -1717,6 +1717,12 @@ parameter as Alias Type USE_VALUE.  A client MUST handle registration failures
...
+
+The MOQT_AGENT parameter (Parameter Type 0x05) is a UTF-8 encoded
+string describing the software agent of the sender.
```
```suggestion
#### MOQT IMPLEMENTATION

The MOQT_IMPLEMENTATION parameter (Parameter Type 0x05) is a UTF-8 encoded
string describing the software agent of the sender.
```

draft-ietf-moq-transport.md
```diff
@@ -1717,6 +1717,12 @@ parameter as Alias Type USE_VALUE.  A client MUST handle registration failures
 of this kind by purging any Token Aliases that failed to register based on the
 MAX_AUTH_TOKEN_CACHE_SIZE parameter in SERVER_SETUP (or the default value of 0).
 
+#### MOQT IMPLEMENTATION
+
+The MOQT_IMPLEMENTATION parameter (Parameter Type 0x05) is a UTF-8 encoded
+string describing the software agent of the sender.
```
```suggestion
#### MOQT IMPLEMENTATION

The MOQT_IMPLEMENTATION parameter (Parameter Type 0x05) is a UTF-8 encoded
string describing the sender's MOQT implementation.
```

LPardue (approved):

suhasHere:
Still LGTM.

vasilvv (approved):

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1717,6 +1717,12 @@ parameter as Alias Type USE_VALUE.  A client MUST handle registration failures
 of this kind by purging any Token Aliases that failed to register based on the
 MAX_AUTH_TOKEN_CACHE_SIZE parameter in SERVER_SETUP (or the default value of 0).
 
+#### MOQT IMPLEMENTATION
+
+The MOQT_IMPLEMENTATION parameter (Parameter Type 0x05) is a UTF-8 encoded
```
fluffy: Unless you are going to provide all the string prep rules and lang stuff for this string, I would not use UTF-8. It just bring useless review for something that is not meant to be human readable.

draft-ietf-moq-transport.md
```diff
@@ -1717,6 +1717,12 @@ parameter as Alias Type USE_VALUE.  A client MUST handle registration failures
...
+string describing the sender's MOQT implementation.
```
fluffy: suggest change "sender's MOQT implementation." to "name and version of sender's MOQT implementation."

draft-ietf-moq-transport.md
afrind: I think it is meant to be human readable.  This is likely to make its way into logs/log databases that will be analyzed by people.  I can see restricting it further (eg to printable ASCII) but would prefer not to make it a binary blob.

draft-ietf-moq-transport.md
afrind: Should we make it a structured field with a string name and a varint version number?  Honestly one of the uses of this field may be to workaround implementation/version specific specific bugs.  Making it entirely free-form means bespoke parsing rules -- this is what we have to do with User-Agent.

wilaw: +1 to structure. As soon as you provide a free-form text field, implementors will use it transfer huge JSON datasets.

draft-ietf-moq-transport.md
vasilvv: I would just copy whatever RFC 9000 says about CONNECTION_CLOSE reason phrase encoding.

afrind: For reference that is:

> Additional diagnostic information for the closure. This can be zero length if the sender chooses not to give details beyond the Error Code value. This SHOULD be a UTF-8 encoded string [[RFC3629](https://datatracker.ietf.org/doc/html/rfc3629)], though the frame does not carry information, such as language tags, that would aid comprehension by any entity other than the one that created the text.

afrind:
From Authors/Editors call, for now, we're just going to make it a blob.  Will remove the UTF-8 bit.

draft-ietf-moq-transport.md
```suggestion
The MOQT_IMPLEMENTATION parameter (Parameter Type 0x05) identifies the name and
version of the sender's MOQT implementation.  This SHOULD be a UTF-8 encoded
string [RFC3629], though the message does not carry information, such as
language tags, that would aid comprehension by any entity other than the one
that created the text.
```

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -3640,6 +3646,8 @@ Implementations are advised to use timeouts to prevent resource
 exhaustion attacks by a peer that does not send expected data within
 an expected time.  Each implementation is expected to set its own limits.
 
+TODO: Security Considerations of MOQT_IMPLEMENTATION parameter
```
```suggestion
TODO: Security/Privacy Considerations of MOQT_IMPLEMENTATION parameter
```

Issue #1037, opened 2025-06-02
SHOULD vs MUST for multi-publisher relay handling


afrind:
@suhasHere Raised some comments on #995 regarding normative statements for relays handling multiple publishers of the same namespace.  

The current text is:

> When a relay receives an incoming SUBSCRIBE request that triggers an upstream subscription, it SHOULD send a SUBSCRIBE request to each publisher that has announced the subscription's namespace, unless it already has an active subscription for the Objects requested by the incoming SUBSCRIBE request from all available publishers.

> When a relay receives an incoming ANNOUNCE for a given namespace, for each active upstream subscription that matches that namespace, it SHOULD send a SUBSCRIBE to the publisher that sent the ANNOUNCE.

Is SHOULD sufficient in these cases (and the PUBLISH equivalents), or is MUST required?

fluffy:
What are the cases that it would not send them ? The stuff we agreed on before needs this to work.

afrind:
> What are the cases that it would not send them ? The stuff we agreed on before needs this to work.

@fluffy you wrote this text in #525.  What did you have in mind?

suhasHere:
thanks @afrind  . My main reason to notice this ( but seems like I did miss old text somehow) was with SHOULD, we can't have make before break like flows work consistently across relay implementations.

I felt it would be logical to get it addressed when we are adding few new rules to the relay section in PUBLISH PR

@fluffy i know you had #525 , but would you be ok with the suggestion to move to MUST ?

fluffy:
I think in that 525 pr it was should unless there is already an active subscribe to one of them. Either way, I'm of the Justin Uberti view of if you are checking in a PR in a room that is mess, we need to clean up the room. I think we do need to have clearly defined behavior here and my view is many implementers ignore should so I try to avoid them and give clear guidance. It's clear that if some relays do this and some do not, things are not going to work.

suhasHere:
@martinduke  @gloinul  wondering if we can spends few mins on this issue ai IETF 123. This will help interoperability

ianswett:
Today at the MoQ session I suggested adding an indication on PUBLISH/SUBSCRIBE_OK and/or ANNOUNCE as to whether the person is authoritative for the entire Track or not.  If it's authoritative for all the data in the Track, then it's only necessary to subscribe to one to get all the Objects.

afrind:
Discussed in Madrid:

We've talked about a few multi-publisher use cases:

1. Make-before-break
[...]

Issue #845, opened 2025-03-24
Can relays guarantee Object deduplication?


vasilvv:
Draft says:

> The relay will drop duplicate objects received on both subscriptions.

We do say that relays are allowed to drop objects if they already have one in cache, but it is certainly not guaranteed that a relay will deduplicate objects, since the memory it uses to remember past objects is limited.

We should have text explicitly stating that if a publisher sends duplicate objects to the relay, there is non-determinism about how many exactly will be forwarded past the first one.

fluffy:
Would something like this text work for you?

"Given a relays will not keep state about all the objects they have received in the past, there is guarantee that a relay will not forward and object it has previously forwarded."

PR #1115, opened 2025-07-27
MUST for multi-publisher relay handling

afrind:
Fixes: #1037
Fixes: #845

Created a section on Multiple Publishers in the Relay section, moving some text there and adding some motivating use cases.  I included the spicy "distributed publisher" case at least for now (informative, not normative).

Updated the relevant SHOULDs to MUSTs.

Clarified that object deduplication is subject to implementation limits.

Some drive by capitalization of Track, Group, Object and Relay.  I've been slowly chipping away to avoid a giant PR.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1239,12 +1239,13 @@ When a relay receives an incoming PUBLISH message, it MUST send a PUBLISH
 request to each subscriber that has subscribed (via SUBSCRIBE_NAMESPACE)
 to the Track's namespace or prefix thereof.
 
-When a relay receives an incoming PUBLISH_NAMESPACE for a namespace that matches
-one or more existing subscriptions to other upstream sessions, it MUST send a
-SUBSCRIBE to the publisher that sent the PUBLISH_NAMESPACE for each matching
-subscription.  When it receives an incoming PUBLISH message for a Track that has
-active subscribers, it MUST respond with PUBLISH_OK, and SHOULD use Forward
-State=1.
+When a relay receives an authorized PUBLISH_NAMESPACE for a namespace that
+matches one or more existing subscriptions to other upstream sessions, it MUST
+send a SUBSCRIBE to the publisher that sent the PUBLISH_NAMESPACE for each
+matching subscription.  When it receives an authorized PUBLISH message for a
+Track that has active subscribers, it MUST respond with PUBLISH_OK.  If at least
+one downstream subscriber for the Track has Forward State=1, the Relay MUST use
+Forward State=1 in the reply.
 
 Relays use the Track Alias ({{track-alias}}) of an incoming Object to identify
 its Track and find the active subscribers. Relays MUST forward Objects to
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1136,20 +1136,19 @@ A Relay can receive PUBLISH_NAMESPACE for the same Track Namespace or PUBLISH
 messages for the same Track from multiple publishers and is expected to treat
 them uniformly.
 
-The is required in order to support the following use cases:
+This behavior is necessary to support the following scenarios:
 
 1. A publisher loses state and reconnects to the same Relay, but the Relay
    has not yet detected the old session has terminated.
-2. A publisher is attempting to gracefully migrate to a new session.
+2. A publisher is performing a graceful migration to a new session.
 3. A system that facilitates failover with multiple redundant publishers that
    produce identical Track content.
 4. A distributed set of endpoints, each publishing different content in the same
    Track.
 
-Relays MUST be able to process Objects for the same Track from multiple
-publishers and forward Objects to active matching subscriptions.  The Relay
-SHOULD attempt to deduplicate Objects when forwarding downstream, subject to
-implementation limits.
+Relays MUST handle Objects for the same Track from multiple publishers and
+forward them to active matching subscriptions. The Relay SHOULD attempt to
+deduplicate Objects before forwarding, subject to implementation constraints.
 
 ## Subscriber Interactions
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1133,7 +1133,7 @@ A cache MUST store all properties of an Object defined in
 ## Multiple Publishers
 
 A Relay can receive PUBLISH_NAMESPACE for the same Track Namespace or PUBLISH
-messages for the same Track from multiple publishers.  This following sections
+messages for the same Track from multiple publishers.  The following sections
 explain how Relays maintain subscriptions to all available publishers for a
 given Track.
 
@@ -1236,7 +1236,7 @@ State=0, the Relay MUST send SUBSCRIBE_UPDATE with Forward=1 to all publishers.
 If there are no active upstream subscriptions for the requested Track, the Relay
 MUST send a SUBSCRIBE request to each publisher that has published the
 subscription's namespace or prefix thereof.  If the SUBSCRIBE has Forward
-State=1, then the Relay MUST use Foward=1 when subscribing upstream.
+=1, then the Relay MUST use Forward=1 when subscribing upstream.
 
 When a relay receives an incoming PUBLISH message, it MUST send a PUBLISH
 request to each subscriber that has subscribed (via SUBSCRIBE_NAMESPACE)
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1137,16 +1137,12 @@ messages for the same Track from multiple publishers.  The following sections
 explain how Relays maintain subscriptions to all available publishers for a
 given Track.
 
-There are multiple scenarios where more than one publisher of the same Track
-connects to the same Relay, including, but not limited to:
-
-1. A publisher loses state and reconnects to the same Relay, but the Relay
-   has not yet detected the old session has terminated.
-2. A publisher is performing a graceful migration to a new session.
-3. A system that facilitates failover with multiple redundant publishers that
-   produce identical Track content.
-4. A distributed set of endpoints, each publishing different Objects in the same
-   Track.
+There is no specified limit to the number of publishers of a Track Namespace or
+Track.  An implementation can use mechanisms such as PUBLISH_ERROR,
+PUBLISH_NAMESPACE_ERROR, UNSUBSCRIBE, PUBLISH_NAMESPACE_CANCEL or GOAWAY if it
+cannot accept an additional publisher due to implementation constraints.
+Implementations can consider the establishment or idle time of the session or
+subscription to determine which publisher to reject or disconnect.
 
 Relays MUST handle Objects for the same Track from multiple publishers and
 forward them to active matching subscriptions. The Relay SHOULD attempt to
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1146,6 +1146,11 @@ The is required in order to support the following use cases:
 4. A distributed set of endpoints, each publishing different content in the same
    Track.
 
+Relays MUST be able to process Objects for the same Track from multiple
+publishers and forward Objects to active matching subscriptions.  The Relay
+SHOULD attempt to deduplicate Objects when forwarding downstream, subject to
+implementation limits.
+
 ## Subscriber Interactions
 
 Subscribers subscribe to tracks by sending a SUBSCRIBE
@@ -1165,11 +1170,6 @@ For successful subscriptions, the publisher maintains a list of
 subscribers for each Track. Each new Object belonging to the
 Track within the subscription range is forwarded to each active
 subscriber, dependent on the congestion response.
-
-Relays MUST be able to process Objects for the same Track from multiple
-publishers and forward Objects to active matching subscriptions.  The Relay
-SHOULD attempt to deduplicate Objects when forwarding downstream, subject to
-implementation limits.
 
 A relay MUST NOT reorder or drop objects received on a multi-object stream when
 forwarding to subscribers, unless it has application specific information.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1172,8 +1172,8 @@ given Track.
 
 There is no specified limit to the number of publishers of a Track Namespace or
 Track.  An implementation can use mechanisms such as PUBLISH_ERROR,
-PUBLISH_NAMESPACE_ERROR, UNSUBSCRIBE, PUBLISH_NAMESPACE_CANCEL or GOAWAY if it
-cannot accept an additional publisher due to implementation constraints.
+PUBLISH_NAMESPACE_ERROR, UNSUBSCRIBE or PUBLISH_NAMESPACE_CANCEL if it cannot
+accept an additional publisher due to implementation constraints.
 Implementations can consider the establishment or idle time of the session or
 subscription to determine which publisher to reject or disconnect.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1227,13 +1227,14 @@ PUBLISH_NAMESPACE namespace=(foobar).  Relays MUST forward SUBSCRIBE messages to
 all publishers and PUBLISH_NAMESPACE and PUBLISH messages to all subscribers
 that have a namespace prefix match.
 
-When a relay receives an incoming SUBSCRIBE that triggers an upstream
-subscription, it MUST send a SUBSCRIBE request to each publisher that has
-published the subscription's namespace or prefix thereof, unless it already has
-an active subscription for the Track requested by the incoming SUBSCRIBE request
-from all available publishers.  If it already has a matching upstream
-subscription in Forward State=0, it SHOULD send a SUBSCRIBE_UDPATE with
-Forward=1 to all publishers.
+When a Relay receives an authorized SUBSCRIBE for a Track with one or more
+active upstream subscriptions, it MUST reply with SUBSCRIBE_OK.  If the
+SUBSCRIBE has Forward State=1 and the upstream subscriptions are in Forward
+State=0, the Relay MUST send SUBSCRIBE_UPDATE with Forward=1 to all publishers.
+If there are no active upstream subscriptions for the requested Track, the Relay
+MUST send a SUBSCRIBE request to each publisher that has published the
+subscription's namespace or prefix thereof.  If the SUBSCRIBE has Forward
+State=1, then the Relay MUST use Foward=1 when subscribing upstream.
 
 When a relay receives an incoming PUBLISH message, it MUST send a PUBLISH
 request to each subscriber that has subscribed (via SUBSCRIBE_NAMESPACE)
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1133,8 +1133,8 @@ A cache MUST store all properties of an Object defined in
 ## Multiple Publishers
 
 A Relay can receive PUBLISH_NAMESPACE for the same Track Namespace or PUBLISH
-messages for the same Track from multiple publishers and is expected to treat
-them uniformly.
+messages for the same Track from multiple publishers and MUST handle them
+uniformly.
 
 This behavior is necessary to support the following scenarios:
 
@@ -1143,7 +1143,7 @@ This behavior is necessary to support the following scenarios:
 2. A publisher is performing a graceful migration to a new session.
 3. A system that facilitates failover with multiple redundant publishers that
    produce identical Track content.
-4. A distributed set of endpoints, each publishing different content in the same
+4. A distributed set of endpoints, each publishing different Objects in the same
    Track.
 
 Relays MUST handle Objects for the same Track from multiple publishers and
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1130,6 +1130,22 @@ A cache MUST store all properties of an Object defined in
 {{object-properties}}, with the exception of any extensions
 ({{object-extensions}}) that specify otherwise.
 
+## Multiple Publishers
+
+A Relay can receive PUBLISH_NAMESPACE for the same Track Namespace or PUBLISH
+messages for the same Track from multiple publishers and is expected to treat
+them uniformly.
+
+The is required in order to support the following use cases:
+
+1. A publisher loses state and reconnects to the same Relay, but the Relay
+   has not yet detected the old session has terminated.
+2. A publisher is attempting to gracefully migrate to a new session.
+3. A system that facilitates failover with multiple redundant publishers that
+   produce identical Track content.
+4. A distributed set of endpoints, each publishing different content in the same
+   Track.
+
 ## Subscriber Interactions
 
 Subscribers subscribe to tracks by sending a SUBSCRIBE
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,27 @@ A cache MUST store all properties of an Object defined in
 {{object-properties}}, with the exception of any extensions
 ({{object-extensions}}) that specify otherwise.
 
+## Multiple Publishers
+
+A Relay can receive PUBLISH_NAMESPACE for the same Track Namespace or PUBLISH
+messages for the same Track from multiple publishers and is expected to treat
+them uniformly.
```
```suggestion
A Relay can receive PUBLISH_NAMESPACE or PUBLISH messages for the same Track Namespace or Track from multiple publishers and must handle them uniformly.
```

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,27 @@ A cache MUST store all properties of an Object defined in
...
+
+The is required in order to support the following use cases:
```
```suggestion
This behavior is necessary to support the following scenarios:
```

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,27 @@ A cache MUST store all properties of an Object defined in
...
+
+1. A publisher loses state and reconnects to the same Relay, but the Relay
+   has not yet detected the old session has terminated.
+2. A publisher is attempting to gracefully migrate to a new session.
```
```suggestion
2. A publisher is performing a graceful migration to a new session.
```

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,27 @@ A cache MUST store all properties of an Object defined in
...
+3. A system that facilitates failover with multiple redundant publishers that
+   produce identical Track content.
+4. A distributed set of endpoints, each publishing different content in the same
+   Track.
+
+Relays MUST be able to process Objects for the same Track from multiple
+publishers and forward Objects to active matching subscriptions.  The Relay
+SHOULD attempt to deduplicate Objects when forwarding downstream, subject to
+implementation limits.
```
```suggestion
Relays MUST handle Objects for the same Track from multiple publishers and
forward them to active matching subscriptions. The Relay SHOULD attempt to
deduplicate Objects before forwarding, subject to implementation constraints.
```

suhasHere (approved):
LGTM. Few nits

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,27 @@ A cache MUST store all properties of an Object defined in
...
+them uniformly.
+
+The is required in order to support the following use cases:
+
+1. A publisher loses state and reconnects to the same Relay, but the Relay
+   has not yet detected the old session has terminated.
+2. A publisher is attempting to gracefully migrate to a new session.
+3. A system that facilitates failover with multiple redundant publishers that
+   produce identical Track content.
+4. A distributed set of endpoints, each publishing different content in the same
```
fluffy: I think "different objects" would be much better than "different content" here. One of the core invariants of the design that gets us eventually consistency in the relay network is that you can never publish the same object with different content. That has caused so much confusion at times I don't want to leak in any ambiguity here.

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,27 @@ A cache MUST store all properties of an Object defined in
 {{object-properties}}, with the exception of any extensions
 ({{object-extensions}}) that specify otherwise.
 
+## Multiple Publishers
+
+A Relay can receive PUBLISH_NAMESPACE for the same Track Namespace or PUBLISH
```
fluffy, line 1168: It can also receive multiple SUBSCRIBE OK from different publisher. Seems like we need to cover that too

draft-ietf-moq-transport.md
```diff
@@ -1146,26 +1167,22 @@ sufficient information to send a FETCH_OK immediately in response to a FETCH, it
 MUST withhold sending FETCH_OK until it does.
 
 For successful subscriptions, the publisher maintains a list of
-subscribers for each track. Each new Object belonging to the
-track within the subscription range is forwarded to each active
+subscribers for each Track. Each new Object belonging to the
+Track within the subscription range is forwarded to each active
```
fluffy: Just as a general stylistic thing, I really wish we could keep formatting changes and small typo stuff out of the design PRs but I know how it happens

draft-ietf-moq-transport.md
```diff
@@ -1216,25 +1229,26 @@ all publishers and PUBLISH_NAMESPACE and PUBLISH messages to all subscribers
 that have a namespace prefix match.
 
 When a relay receives an incoming SUBSCRIBE that triggers an upstream
-subscription, it SHOULD send a SUBSCRIBE request to each publisher that has
+subscription, it MUST send a SUBSCRIBE request to each publisher that has
 published the subscription's namespace or prefix thereof, unless it already has
-an active subscription for the Objects requested by the incoming SUBSCRIBE
-request from all available publishers.  If it already has a matching upstream
+an active subscription for the Track requested by the incoming SUBSCRIBE request
```
fluffy: not asking for a change here but just food for thought that might be a good idea ...

I wonder if it would be better to explain this like "when you get a subscribe, first you try and match agains  active subscribes, second if you don't get a match then you form a new one by sending to blah blah blah" - basically two steps where first you match, then if no match form new

draft-ietf-moq-transport.md
```diff
@@ -1216,25 +1229,26 @@ all publishers and PUBLISH_NAMESPACE and PUBLISH messages to all subscribers
...
-When a relay receives an incoming PUBLISH_NAMESPACE for a given namespace, for
-each active upstream subscription that matches that namespace, it SHOULD send a
-SUBSCRIBE to the publisher that sent the PUBLISH_NAMESPACE.  When it receives an
-incoming PUBLISH message for a track that has active subscribers, it SHOULD
-respond with PUBLISH_OK with Forward State=1.
+When a relay receives an incoming PUBLISH_NAMESPACE for a namespace that matches
+one or more existing subscriptions to other upstream sessions, it MUST send a
+SUBSCRIBE to the publisher that sent the PUBLISH_NAMESPACE for each matching
+subscription.  When it receives an incoming PUBLISH message for a Track that has
+active subscribers, it MUST respond with PUBLISH_OK, and SHOULD use Forward
```
fluffy: why would this not be froward state MUST be 1. Sure there may be a later PR that says when not to do this but the relays don't work if they don't get the data

fluffy (changes requested):
Few small things to change but I think this looks pretty good overall and going the right direction

draft-ietf-moq-transport.md
afrind: It can only get a SUBSCRIBE_OK once it issues the SUBSCRIBE, which is after the PUBLISH_NAMESPACE part happens.  There's separate text below that says when it needs to issue multiple SUBSCRIBES.

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,26 @@ A cache MUST store all properties of an Object defined in
 {{object-properties}}, with the exception of any extensions
 ({{object-extensions}}) that specify otherwise.
 
+## Multiple Publishers
+
+A Relay can receive PUBLISH_NAMESPACE for the same Track Namespace or PUBLISH
+messages for the same Track from multiple publishers and is expected to treat
+them uniformly.
```
```suggestion
messages for the same Track from multiple publishers and MUST handle them
uniformly.
```

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,26 @@ A cache MUST store all properties of an Object defined in
...
+
+This behavior is necessary to support the following scenarios:
+
+1. A publisher loses state and reconnects to the same Relay, but the Relay
+   has not yet detected the old session has terminated.
+2. A publisher is performing a graceful migration to a new session.
+3. A system that facilitates failover with multiple redundant publishers that
+   produce identical Track content.
+4. A distributed set of endpoints, each publishing different content in the same
```
```suggestion
4. A distributed set of endpoints, each publishing different Objects in the same
```

draft-ietf-moq-transport.md
afrind: I hear your feedback.  I got a little over-zealous here. I'll try to constrain the mini-corrections to paragraphs that are actually changing.

draft-ietf-moq-transport.md
afrind: Ah yes.  If the downstream forward state=1 for any subscriber then it MUST be 1 upstream.  If all downstream subscribers have forward state=0, I think elsewhere we say SHOULD send 1 upstream but offer the relay some flexibility (for example, DoS prevention).

afrind (commented):
Thanks for the quick reviews!

draft-ietf-moq-transport.md
```diff
@@ -1216,25 +1228,26 @@ all publishers and PUBLISH_NAMESPACE and PUBLISH messages to all subscribers
...
-each active upstream subscription that matches that namespace, it SHOULD send a
-SUBSCRIBE to the publisher that sent the PUBLISH_NAMESPACE.  When it receives an
-incoming PUBLISH message for a track that has active subscribers, it SHOULD
-respond with PUBLISH_OK with Forward State=1.
+When a relay receives an incoming PUBLISH_NAMESPACE for a namespace that matches
+one or more existing subscriptions to other upstream sessions, it MUST send a
+SUBSCRIBE to the publisher that sent the PUBLISH_NAMESPACE for each matching
+subscription.  When it receives an incoming PUBLISH message for a Track that has
+active subscribers, it MUST respond with PUBLISH_OK, and SHOULD use Forward
+State=1.
```
```suggestion
When a relay receives an authorized PUBLISH_NAMESPACE for a namespace that
matches one or more existing subscriptions to other upstream sessions, it MUST
send a SUBSCRIBE to the publisher that sent the PUBLISH_NAMESPACE for each
matching subscription.  When it receives an authorized PUBLISH message for a
Track that has active subscribers, it MUST respond with PUBLISH_OK.  If at least
one downstream subscriber for the Track has Forward State=1, the Relay MUST use
Forward State=1 in the reply.
```

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,26 @@ A cache MUST store all properties of an Object defined in
 {{object-properties}}, with the exception of any extensions
 ({{object-extensions}}) that specify otherwise.
 
+## Multiple Publishers
+
+A Relay can receive PUBLISH_NAMESPACE for the same Track Namespace or PUBLISH
+messages for the same Track from multiple publishers and MUST handle them
+uniformly.
```
vasilvv: I've no idea what "MUST handle them uniformly" means here, especially as a normative requirement.

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,26 @@ A cache MUST store all properties of an Object defined in
...
+
+This behavior is necessary to support the following scenarios:
+
+1. A publisher loses state and reconnects to the same Relay, but the Relay
+   has not yet detected the old session has terminated.
```
vasilvv: I am not sure this is a "multiple publisher" case, as at any individual instance in time, there exists only one publisher that is active?

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,26 @@ A cache MUST store all properties of an Object defined in
...
+2. A publisher is performing a graceful migration to a new session.
+3. A system that facilitates failover with multiple redundant publishers that
+   produce identical Track content.
```
vasilvv: 3 is actually two different use cases: one (3a) in which there are two redundant publishers sending the same media, and one (3b) where one of them is on standby.

In all of those cases, there is only one original publisher, but it publishes its track multiple times.  2 and 3b are similar in that most of the time, there is only one publisher speaking, so we might want to add signaling to distinguish this from 3a (which is a use case which we've not discussed that much in the working group, but probably should).

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,26 @@ A cache MUST store all properties of an Object defined in
...
+4. A distributed set of endpoints, each publishing different Objects in the same
+   Track.
```
vasilvv: This is the only use case that appears to imply two original publishers with disjoint state active at the same time.  I don't believe we've ever had consensus to support this.

vasilvv (changes requested):
This PR seems to be implying (as did the old text, though this has somehow escaped my notice before) that MoQT relays would take an arbitrarily large numbers of original publishers and aggregate the objects they publish into a single track that is then fanned out to an arbitrarily large number of subscribers.  This would impose a significant limitation on the ability of the relay networks to scale, and as such I don't think it is justifiable.

vasilvv:
Also, the text added seems to go far beyond the two issues cited, so you may want to open a separate issue to track that?

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,26 @@ A cache MUST store all properties of an Object defined in
 {{object-properties}}, with the exception of any extensions
 ({{object-extensions}}) that specify otherwise.
 
+## Multiple Publishers
+
+A Relay can receive PUBLISH_NAMESPACE for the same Track Namespace or PUBLISH
+messages for the same Track from multiple publishers and MUST handle them
+uniformly.
+
+This behavior is necessary to support the following scenarios:
```
suhasHere: I feel adding these use-cases might be causing confusion, as it can't be exhaustive or can't be specific and perhaps keep this PR scoped just to relay behavior alone?

draft-ietf-moq-transport.md
afrind: The previous text was `is expected to handle them uniformly`, I agree that the normative form doesn't make much sense either.  I think what I'm trying to capture is "if there's more than one publisher for a thing, then the relay does the same stuff to all of them"

draft-ietf-moq-transport.md
afrind: I think getting agreement on the problems we are trying to solve is important, let's try work through them a bit more.  If we stall out here and the normative parts of this PR have consensus, we can remove this.

draft-ietf-moq-transport.md
afrind: The relay might have two sessions for the same Track, and it doesn't know that only one is active.  It's one of the motivating use cases for why relays need to support more than one session per track.

draft-ietf-moq-transport.md
afrind: > add signaling

Do you want to file a separate issue or proposal?

draft-ietf-moq-transport.md
afrind: Agree it's still marked as Needs Discussion following the consensus call in the draft where we relaxed the temporal requirement for groups.  I will ask the chairs to help clarify this.  I think it matters less for the normative language in this PR, because the same rules work for _any_ multi-publisher SUBSCRIBE case, than in #1117.

suhasHere:
> hat MoQT relays would take an arbitrarily large numbers of original publishers and aggregate the objects they publish into a single track

I don't think aggregation of published tracks across the publishers is mandated in this PR. @vasilvv which statements seems to conclude the above observation ?

draft-ietf-moq-transport.md
suhasHere: how about "A Relay can receive PUBLISH_NAMESPACE or PUBLISH messages for the same Track Namespace or Track from multiple publishers and MUST process them consistently."

draft-ietf-moq-transport.md
suhasHere: We have discussed this use-case several times and we have submitted drafts that aligns with such a design. We had consensus on this and  as mentioned by Alan above we did address this in draft-11.  I am not seeing a reason to block such flows in the relay.

This PR doesn't define extra handling for same publisher coming across 2 sessions vs multiple publishers on the same track, at a relay layer it is serving tracks not publishers. As far as relay is concerned, it sees objects coming in for a given track across multiple sessions and relay does it usual filtering rules  (either dropping duplicates or applying subscription fitlers)  when serving such objects to the downstream subscribers.

afrind:
> I don't think aggregation of published tracks across the publishers is mandated in this PR. @vasilvv which statements seems to conclude the above observation ?

I agree with Victor - this is the implication.  By allowing multiple publishers of the same track and requiring the Relay to subscribe to all of them, the Relay fans-in and performs best-effort deduplication.  This is a form of aggregation.  That said, I think we need to do this in order to support some very real use-cases.

I wonder @vasilvv if adding a limit on the number of concurrent publishers helps address any of your concerns?  eg: a Relay MUST support at least N concurrent publishers of the same track, for some N > 1.  It's clear that no relay will support infinity.

draft-ietf-moq-transport.md
afrind: I think normative language in this section is not as helpful as below.  I will come up with alternate text.

draft-ietf-moq-transport.md
afrind: > We had consensus on this and as mentioned by Alan above 

I did not say we had consensus on this point.  I believe Victor filed an Objection following the draft in which that was introduced, and we have not resolved that.  I asked the chairs to clarify @martinduke @gloinul.

suhasHere:
> I agree with Victor - this is the implication. By allowing multiple publishers of the same track and requiring the Relay to subscribe to all of them, the Relay fans-in and performs best-effort deduplication. This is a form of aggregation.

Deduplication needs to be done even in a single publisher use-case and make before break use-cases and I want to keep them separeate.  But I am not seeing why would a relay needs to perform aggregation in these cases ? May be I don't understand what we mean by aggregation here ?

draft-ietf-moq-transport.md
gwendalsimon: I would simply discard this "handle them uniformly" part of the sentence. I don't see reasons a Relay would not act "uniformly", so I don't understand why we would need to mention anything here.

draft-ietf-moq-transport.md
gwendalsimon: As said in another comment, I am not a fan of the "MUST handle it uniformly". Similarly, I would not mention "the behavior is necessary" (because in my mind it is just the normal behavior). I would rephrase it such as "Multi-publisher can happen in many scenarios, including (but not restricted to)"

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,26 @@ A cache MUST store all properties of an Object defined in
...
+   has not yet detected the old session has terminated.
+2. A publisher is performing a graceful migration to a new session.
+3. A system that facilitates failover with multiple redundant publishers that
+   produce identical Track content.
+4. A distributed set of endpoints, each publishing different Objects in the same
+   Track.
+
+Relays MUST handle Objects for the same Track from multiple publishers and
+forward them to active matching subscriptions. The Relay SHOULD attempt to
+deduplicate Objects before forwarding, subject to implementation constraints.
```
gwendalsimon, line 1182: In the case the Track is such as group ID increases (and possibly sequentially), it could be expected the Relay to (try to) re-order groups that arrive out of order. 
Of course, I know it can be a dangerous process and I don't want to enter into a very complicated debate here... but...
Should we say something like "_The Relay MAY forward Objects from different publishers in an order that is not the same as the reception order._"?

draft-ietf-moq-transport.md
gwendalsimon: ... by the way, what about Publishers who have different publisher_priority? To handle the multiple track _uniformly_, should the Relay force all Publisher to have the same priority by setting the same subscriber_priority to all subscription?

draft-ietf-moq-transport.md
suhasHere: > Multi-publisher can happen in many scenarios, including (but not restricted to)"

I think this is a good suggestion

draft-ietf-moq-transport.md
afrind: @gwendalsimon We already constrain delivery order within the same track to be in the specified Group Order (which can differ from reception order) in the Priority section.  So I think we're good?

draft-ietf-moq-transport.md
```suggestion
A Relay can receive PUBLISH_NAMESPACE for the same Track Namespace or PUBLISH
messages for the same Track from multiple publishers.  This following sections
explain how Relays maintain subscriptions to all available publishers for a
given Track.
```

draft-ietf-moq-transport.md
```suggestion
There are multiple scenarios where more than one publisher of the same Track
connects to the same Relay, including, but not limited to:
```

draft-ietf-moq-transport.md
```diff
@@ -1215,26 +1229,29 @@ PUBLISH_NAMESPACE namespace=(foobar).  Relays MUST forward SUBSCRIBE messages to
...
-subscription in Forward State=0, it SHOULD send a SUBSCRIBE_UDPATE with
-Forward=1 to all publishers.
+When a Relay receives an authorized SUBSCRIBE for a Track with one or more
+active upstream subscriptions, it MUST reply with SUBSCRIBE_OK.  If the
+SUBSCRIBE has Forward State=1 and the upstream subscriptions are in Forward
+State=0, the Relay MUST send SUBSCRIBE_UPDATE with Forward=1 to all publishers.
+If there are no active upstream subscriptions for the requested Track, the Relay
+MUST send a SUBSCRIBE request to each publisher that has published the
+subscription's namespace or prefix thereof.  If the SUBSCRIBE has Forward
+State=1, then the Relay MUST use Foward=1 when subscribing upstream.
```
```suggestion
=1, then the Relay MUST use Forward=1 when subscribing upstream.
```

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,28 @@ A cache MUST store all properties of an Object defined in
 {{object-properties}}, with the exception of any extensions
 ({{object-extensions}}) that specify otherwise.
 
+## Multiple Publishers
+
+A Relay can receive PUBLISH_NAMESPACE for the same Track Namespace or PUBLISH
+messages for the same Track from multiple publishers.  This following sections
```
```suggestion
messages for the same Track from multiple publishers.  The following sections
```

draft-ietf-moq-transport.md
```diff
@@ -1215,26 +1229,29 @@ PUBLISH_NAMESPACE namespace=(foobar).  Relays MUST forward SUBSCRIBE messages to
...
-SUBSCRIBE to the publisher that sent the PUBLISH_NAMESPACE.  When it receives an
-incoming PUBLISH message for a track that has active subscribers, it SHOULD
-respond with PUBLISH_OK with Forward State=1.
+When a relay receives an authorized PUBLISH_NAMESPACE for a namespace that
+matches one or more existing subscriptions to other upstream sessions, it MUST
+send a SUBSCRIBE to the publisher that sent the PUBLISH_NAMESPACE for each
+matching subscription.  When it receives an authorized PUBLISH message for a
+Track that has active subscribers, it MUST respond with PUBLISH_OK.  If at least
+one downstream subscriber for the Track has Forward State=1, the Relay MUST use
+Forward State=1 in the reply.
```
suhasHere, line 1298: This is enforcing a behavior, where when a relay gets PUBLISH with forward=1 and there are no active subscriptions,  a Relay MUST respond with forward=0. This defeats the purpose of publish with forward=1.

suhasHere: Other than this , i think rest of the PR looks pretty good.

afrind: There is no MUST for Forward=0 in that case. It’s Relay choice and I believe there’s a SHOULD elsewhere. I suppose we could add another SHOULD here

draft-ietf-moq-transport.md
fluffy: We have discussed this at several face to face meetings. We implemented this and I don't see any reason it should have any negative impact on scalability.

draft-ietf-moq-transport.md
suhasHere: That seems fine to me per our discussions of making it mandatory only when there are active subscriptions.

afrind:
From Authors/Editors call:

Plan is to remove the use case descriptions for now and add non-normative considerations for what to do when the Relay exceeds its limits for publishers of a given Track (eg: you might want to unsubscribe from oldest idle publisher, or use goaway).

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,28 @@ A cache MUST store all properties of an Object defined in
...
+connects to the same Relay, including, but not limited to:
+
+1. A publisher loses state and reconnects to the same Relay, but the Relay
+   has not yet detected the old session has terminated.
+2. A publisher is performing a graceful migration to a new session.
+3. A system that facilitates failover with multiple redundant publishers that
+   produce identical Track content.
+4. A distributed set of endpoints, each publishing different Objects in the same
+   Track.
+
```
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,28 @@ A cache MUST store all properties of an Object defined in
 {{object-properties}}, with the exception of any extensions
 ({{object-extensions}}) that specify otherwise.
 
+## Multiple Publishers
+
+A Relay can receive PUBLISH_NAMESPACE for the same Track Namespace or PUBLISH
+messages for the same Track from multiple publishers.  The following sections
+explain how Relays maintain subscriptions to all available publishers for a
+given Track.
```
afrind, line 1171:
```suggestion
given Track.

There is no specified limit to the number of publishers of a Track Namespace or
Track.  An implementation can use mechanisms such as PUBLISH_ERROR,
PUBLISH_NAMESPACE_ERROR, UNSUBSCRIBE, PUBLISH_NAMESPACE_CANCEL or GOAWAY if it
cannot accept an additional publisher due to implementation constraints.
Implementations can consider the establishment or idle time of the session or
subscription to determine which publisher to reject or disconnect.
```

suhasHere (approved):
looks good to me

vasilvv (approved):

draft-ietf-moq-transport.md
```diff
@@ -1130,6 +1130,24 @@ A cache MUST store all properties of an Object defined in
...
+
+There is no specified limit to the number of publishers of a Track Namespace or
+Track.  An implementation can use mechanisms such as PUBLISH_ERROR,
+PUBLISH_NAMESPACE_ERROR, UNSUBSCRIBE, PUBLISH_NAMESPACE_CANCEL or GOAWAY if it
```
ianswett: GOAWAY?  That seems like a misuse of GOAWAY.

```suggestion
PUBLISH_NAMESPACE_ERROR, UNSUBSCRIBE, or PUBLISH_NAMESPACE_CANCEL if it
```

draft-ietf-moq-transport.md
```diff
@@ -1163,6 +1163,24 @@ A cache MUST store all properties of an Object defined in
...
+cannot accept an additional publisher due to implementation constraints.
+Implementations can consider the establishment or idle time of the session or
+subscription to determine which publisher to reject or disconnect.
```
```suggestion
There is no specified limit to the number of publishers of a Track Namespace or
Track.  An implementation can use mechanisms such as PUBLISH_ERROR,
PUBLISH_NAMESPACE_ERROR, UNSUBSCRIBE or PUBLISH_NAMESPACE_CANCEL if it cannot
accept an additional publisher due to implementation constraints.
Implementations can consider the establishment or idle time of the session or
subscription to determine which publisher to reject or disconnect.
```

ianswett (approved):

draft-ietf-moq-transport.md
afrind: This comment is about an older version of the text, I already removed GOAWAY.

Issue #1100, opened 2025-07-22
Prefix matching under specified


wilaw:
The current draft states:

> Prefix matching is used to determine which publishers receive a SUBSCRIBE or which subscribers receive a PUBLISH. For example, a SUBSCRIBE namespace=(foo,bar), track=x message will be forwarded to the sessions that sent ANNOUNCE namespace=(foo) and ANNOUNCE namespace=(foo, bar) respectively, but not one that sent ANNOUNCE namespace=(foobar).

This does not indicate if a match can be achieved where the prefix is not a contiguous tuple field . This seems to be an assumption of the draft but it is not explicitly stated.  Traditional prefix matching allows a left-side match but has no requirement that what is being matched is a complete tuple field. 

For example, if there is an ANNOUNCE namespace=(firstfield,secondfield), then do the following subscriptions match?

SUBSCRIBE namespace=(firstfield)
SUBSCRIBE namespace=(first)
SUBSCRIBE namespace=(firstfield,second)
SUBSCRIBE namespace=(first,second)
SUBSCRIBE namespace=(f,s)

When we say "prefix matching" do we instead mean that there must be an "exact match" for each tuple field, starting from the left?

sharmafb:
I would assume that it means that there must be an "exact match" for each tuple field, starting from the 0th element.

PR #1116, opened 2025-07-27
Explain Namespace Prefix Matching in more detail

afrind:
Fixes: #1100

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1205,22 +1205,24 @@ sends a PUBLISH_NAMESPACE_DONE. A subscriber indicates it will no longer
 subcribe to tracks in a namespace it previously responded PUBLISH_NAMESPACE_OK
 to by sending a PUBLISH_NAMESPACE_CANCEL.
 
-A Relay manages sessions from multiple publishers and subscribers, connecting
-them based on the Track Namespace or Full Track Name.  The Full Track Name in
-SUBSCRIBE is matched exactly against existing upstream subscriptions.
+A Relay connects publishers and subscribers by managing sessions based on the
+Track Namespace or Full Track Name. When a SUBSCRIBE message is sent, its Full
+Track Name is matched exactly against existing upstream subscriptions.
 
-Namespace Prefix Matching is further used to determine which publishers receive
-a SUBSCRIBE and which subscribers receive a PUBLISH.  In Namespace Prefix
-Matching, the Track Namespace tuple fields are matched from 1 to 32, with each
-tuple field requiring an exact match.  If a session's published or subscribed
-Track Namespace has equal or fewer tuple fields than the Track Namesapce in the
-message, it is considered a match. For example, a SUBSCRIBE namespace=(foo,bar),
-name=x message will be forwarded to the sessions that sent PUBLISH_NAMESPACE
-message with namespace=(foo) and namespace=(foo, bar) respectively, but not one
-that sent PUBLISH_NAMESPACE namespace=(foobar).
+Namespace Prefix Matching is further used to decide which publishers receive a
+SUBSCRIBE and which subscribers receive a PUBLISH. In this process, the tuples
+in the Track Namespace are matched sequentially, requiring an exact match for
+each field. If the published or subscribed Track Namespace has the same or fewer
+fields than the Track Namespace in the message, it qualifies as a match.
 
-Relays MUST forward SUBSCRIBE messages to all publishers and PUBLISH_NAMESPACE
-and PUBLISH messages to all subscribers that have a Namespace Prefix Match.
+For example:
+A SUBSCRIBE message with namespace=(foo, bar) and name=x will match sessions
+that sent PUBLISH_NAMESPACE messages with namespace=(foo) or namespace=(foo,
+bar).  It will not match a session with namespace=(foobar).
+
+Relays MUST forward SUBSCRIBE messages to all matching publishers and
+PUBLISH_NAMESPACE or PUBLISH messages to all matching subscribers.
+
 
 When a relay receives an incoming SUBSCRIBE that triggers an upstream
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1205,15 +1205,22 @@ sends a PUBLISH_NAMESPACE_DONE. A subscriber indicates it will no longer
...
-to determine which publishers receive a SUBSCRIBE or which subscribers receive a
-PUBLISH. For example, a SUBSCRIBE namespace=(foo,bar), track=x message will be
-forwarded to the sessions that sent PUBLISH_NAMESPACE namespace=(foo) and
-PUBLISH_NAMESPACE namespace=(foo, bar) respectively, but not one that sent
-PUBLISH_NAMESPACE namespace=(foobar).  Relays MUST forward SUBSCRIBE messages to
-all publishers and PUBLISH_NAMESPACE and PUBLISH messages to all subscribers
-that have a namespace prefix match.
+A Relay manages sessions from multiple publishers and subscribers, connecting
+them based on the Track Namespace or Full Track Name.  The Full Track Name in
+SUBSCRIBE is matched exactly against existing subscriptions.
```
vasilvv: I'm confused.  What is the context under which we match name in SUBSCRIBE against existing subscriptions?  Did you mean PUBLISH instead of SUBSCRIBE?

```suggestion
SUBSCRIBE is matched exactly against existing upstream subscriptions.
```

afrind: It should say 'upstream' subscriptions.  eg: a previous SUBSCRIBE resulted in an upstream SUBSCRIBE/OK exchange, or there was previously a PUBLISH/OK exchange for this track.

TimEvens: "The Full Track Name in SUBSCRIBE is matched exactly against existing subscriptions." is misleading and incorrect in that exactly matching is only for PUBLISH.  ANNOUNCE is prefix matched only against the namespace, sans name.

IMO, there needs to be text that describes the algorithm for selecting which publisher or set of publishers will receive the subscribe or Resume.  In our case, a PUBLISH will be in paused state (forward=0) until there is at lest one subscriber.  When there is a subscriber, we update the PUBLISH subscription (aka send a SUBSCRIBE_UPDATE) with `forward=1` to instruct the publisher to start forwarding. This maintains the same behavior with ANNOUNCE sending SUBSCRIBE when subscribers are present.

afrind: > "The Full Track Name in SUBSCRIBE is matched exactly against existing subscriptions." is misleading and incorrect in that exactly matching is only for PUBLISH.

It's PUBLISH and previous upstream SUBSCRIBE/OK flows.

> ANNOUNCE is prefix matched only against the namespace, sans name.

True, that is covered in the next paragraph, since it does not fall under "existing subscriptions".

> IMO, there needs to be text that describes the algorithm for selecting which publisher or set of publishers will receive the subscribe or Resume

There is such text, which is basically "all of them that have announced the namespace of prefix thereof" -- you might need to review the entire section to see if it matches your understand.

suhasHere: > In our case, a PUBLISH will be in paused state (forward=0) until there is at lest one subscriber.

I tend to disagree with this.  I tend to disagree with this.  A relay should be able to get data from publisher , who set forward=1 in PUBLISH and cache it or drop it on the floor, until a subscribe comes.

draft-ietf-moq-transport.md
```diff
@@ -1205,15 +1205,22 @@ sends a PUBLISH_NAMESPACE_DONE. A subscriber indicates it will no longer
...
+Matching, the Track Namespace tuple fields are matched from 1 to 32, with each
+tuple field requiring an exact match.  If a session's published or subscribed
+Track Namespace has equal or fewer tuple fields than the Track Namesapce in the
+message, it is considered a match. For example, a SUBSCRIBE namespace=(foo,bar),
+name=x message will be forwarded to the sessions that sent PUBLISH_NAMESPACE
+message with namespace=(foo) and namespace=(foo, bar) respectively, but not one
+that sent PUBLISH_NAMESPACE namespace=(foobar).
+
+Relays MUST forward SUBSCRIBE messages to all publishers and PUBLISH_NAMESPACE
+and PUBLISH messages to all subscribers that have a Namespace Prefix Match.
```
```suggestion
A Relay connects publishers and subscribers by managing sessions based on the
Track Namespace or Full Track Name. When a SUBSCRIBE message is sent, its Full
Track Name is matched exactly against existing upstream subscriptions.

Namespace Prefix Matching is further used to decide which publishers receive a
SUBSCRIBE and which subscribers receive a PUBLISH. In this process, the tuples
in the Track Namespace are matched sequentially, requiring an exact match for
each field. If the published or subscribed Track Namespace has the same or fewer
fields than the Track Namespace in the message, it qualifies as a match.

For example:
A SUBSCRIBE message with namespace=(foo, bar) and name=x will match sessions
that sent PUBLISH_NAMESPACE messages with namespace=(foo) or namespace=(foo,
bar).  It will not match a session with namespace=(foobar).

Relays MUST forward SUBSCRIBE messages to all matching publishers and
PUBLISH_NAMESPACE or PUBLISH messages to all matching subscribers.

```

draft-ietf-moq-transport.md
afrind: > a PUBLISH will be in paused state (forward=0) until there is at lest one subscriber.

Ours does the same, but I think the spec leaving this open is fine - I don't think there's call to mandate receiving objects that potentially nobody will ever want, but for some applications it could be appropriate.

> I tend to disagree with this

Are you asking for different language in the spec, or are you having an internal design discussion on this PR lol?

fluffy: I'll note that anyone that operates relays as a commercial service is highly likely to allow clients that ask for data to be cached and live will likely provide that service as asked. The key point here is the applications know what it needs - the relay has no clue what the application is much less what is best for it.

PR #1117, opened 2025-07-27
Explain FETCH routing for relays

afrind:
Fixes: #728

* Use the same rules as SUBSCRIBE to find available publishers
---
EDIT: 

Originally this PR had
* MUST send to at least one, and try against others for UNKNOWN_STATUS_IN_RANGE

But now it reads:

* MAY send to any publisher if there is more than one.
---

This leaves the door open for making the FETCHes in parallel, but doesn't require it.

I didn't see any other FETCH_ERROR codes that seemed retryable in this way, but if there are we can make the relay language more generic.

REMINDER for Editors: This is targeted against #1116, retarget main before merging.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1222,6 +1222,13 @@ that sent PUBLISH_NAMESPACE namespace=(foobar).
 Relays MUST forward SUBSCRIBE messages to all publishers and PUBLISH_NAMESPACE
 and PUBLISH messages to all subscribers that have a Namespace Prefix Match.
 
+When a Relay needs to make an upstream FETCH request, it determines the
+available publishers using the same matching rules as SUBSCRIBE. When more than
+one publisher is available, the Relay MUST send the FETCH to at least one of
+them.  If FETCH fails with `UNKNOWN_STATUS_IN_RANGE` (see {{message-fetch}}),
+the Relay MUST attempt the FETCH against other matching publishers until the
+FETCH succeeds or fails for a different reason.
+
 When a relay receives an incoming SUBSCRIBE that triggers an upstream
 subscription, it SHOULD send a SUBSCRIBE request to each publisher that has
 published the subscription's namespace or prefix thereof, unless it already has
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1224,10 +1224,7 @@ and PUBLISH messages to all subscribers that have a Namespace Prefix Match.
 
 When a Relay needs to make an upstream FETCH request, it determines the
 available publishers using the same matching rules as SUBSCRIBE. When more than
-one publisher is available, the Relay MUST send the FETCH to at least one of
-them.  If FETCH fails with `UNKNOWN_STATUS_IN_RANGE` (see {{message-fetch}}),
-the Relay MUST attempt the FETCH against other matching publishers until the
-FETCH succeeds or fails for a different reason.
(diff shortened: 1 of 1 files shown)


vasilvv (changes requested):
This sounds like a terrible idea given that we currently do not constrain the number of original publishers.

vasilvv:
Also, it doesn't work when the fetch is against a range that spans two different original publishers.

draft-ietf-moq-transport.md
```diff
@@ -1222,6 +1222,13 @@ that sent PUBLISH_NAMESPACE namespace=(foobar).
 Relays MUST forward SUBSCRIBE messages to all publishers and PUBLISH_NAMESPACE
 and PUBLISH messages to all subscribers that have a Namespace Prefix Match.
 
+When a Relay needs to make an upstream FETCH request, it determines the
+available publishers using the same matching rules as SUBSCRIBE. When more than
```
TimEvens, line 1226:
```suggestion
available publishers using the same matching rules as SUBSCRIBE. When more than
one publisher is available, the Relay MUST send the FETCH to the first matched
PUBLISH or longest prefix matched ANNOUNCE. 
```

draft-ietf-moq-transport.md
```diff
@@ -1222,6 +1222,13 @@ that sent PUBLISH_NAMESPACE namespace=(foobar).
...
+one publisher is available, the Relay MUST send the FETCH to at least one of
+them.  If FETCH fails with `UNKNOWN_STATUS_IN_RANGE` (see {{message-fetch}}),
```
TimEvens: The original set of publishers matched may be different (changed) by the time ERROR response is received. This should read that the relay **SHOULD** (**not** MUST) upon `UNKNOWN_STATUS_IN_RANGE` reattempt to find another publisher by first matching PUBLISH (more specific), then fallback to ANNOUNCE prefix matching, using the longest match first. The publisher that responded with error would be excluded in the reattempt. It would recursively do this till the list is empty when matching publishers based on PUBLISH and ANNOUNCE (prefix matching). 

While the above would be best to provide a succesful fetch, it's more work for the relay to perform this and could cause the relay to get stuck in a loop of trying to find an active publisher when publishers are in flux.

afrind:
Discussed with Authors/Editors: 

Plan is to say relay can pick any one authorized publisher for a fetch.  If an application supports multiple publishers with disjoint state and needs to support FETCH, it should use an intermediary to present itself to the relay as a single publisher.

draft-ietf-moq-transport.md
```diff
@@ -1222,6 +1222,13 @@ that sent PUBLISH_NAMESPACE namespace=(foobar).
...
+the Relay MUST attempt the FETCH against other matching publishers until the
+FETCH succeeds or fails for a different reason.
+
```
```suggestion
When a Relay needs to make an upstream FETCH request, it determines the
available publishers using the same matching rules as SUBSCRIBE. When more than
one publisher is available, the Relay MAY send the FETCH to any of them.

```

afrind:
> LGTM

@fluffy Please re-review since the thesis changed based on yesterday's call with @suhasHere and @vasilvv.

vasilvv (approved):

draft-ietf-moq-transport.md
afrind: I think we should discuss whether prefix length is a determinant separately from this PR - can you file a separate issue?

ianswett (approved):

Issue #1093, opened 2025-07-17
End of Group flag in Subgroup/Datagram Header creates new ways to malform a track


afrind:
Header indicates Contains End of Group and

1) subgroup contains no objects and ends with a FIN -- cannot determine EOG Object ID
2) subgroup/datagram ends with an object with status = Object Does Not Exist

afrind:
How would people feel about changing the end-of-group flag in the header to mean there's an implicit END_OF_GROUP object with ID = <last object id> + 1?  That at least fixes the second malformed case above, and allows for more symmetric APIs with cases where the EOG is explicitly vs implicitly sent.

martinduke:
The current design creates unfetchable objects and adds unacceptable latency to implementations. IMO changing the definition to mean there is an additional object with ID + 1 and EndOfGroup is imperative.

1) There is no way to express an object with non-normal status and a payload in a FETCH stream, so these objects cannot be delivered by FETCH.

2) The current design adds latency.  On a subgroup stream that contains EndOfGroup, a receiver cannot deliver an object to the upper layer until it either receives a byte from the next object header, or a FIN, because until then it doesn't know the status of the received object.

PR #1118, opened 2025-07-28
More explicit handling for end-of-group bit

afrind:
The previous "implicit" end of group handling both created additional malformed track cases, and was challenging to integrate into existing MoQ APIs.  Change the interpretation to mean it's the same as receiving an End of Group Object with ID + 1.

Fixes: #1093

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3234,8 +3234,18 @@ There are 8 defined Type values for OBJECT_DATAGRAM:
 | 0x07 | Yes           | Yes        | No        |
 |------|---------------|------------|-----------|
 
-For Type values where End of Group is Yes, the Object is the last Object in the
-Group.
+For Type values where End of Group is Yes, the effect is the same as receiving
+the current Datagram followed by an OBJECT_DATAGRAM_STATUS with the following
+values:
+
+~~~
+  Track Alias = datagram.Track Alias
+  Group ID = datagram.Group ID
+  Object ID = datagram.Object ID + 1
+  Publisher Priority = datagram.Publisher Priority
+  Extension Headers Length = 0
+  Object Status = End of Group
+~~~
 
 For Type values where Extensions Present is No, Extensions Headers Length is not
 present and the Object has no extensions.  When Extensions Present is Yes,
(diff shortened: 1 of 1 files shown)


martinduke (commented):
This implies that if I want extensions for my EndOfGroup object, I need to send that object explicitly. That's fine, I suppose, but it is a little different from when a single object had payload, extensions, and EoG status.

I don't understand many of the use cases for extensions, so I don't know if this is a problem.

This gets thorny if the relay decides to add an extension to an EoG object. If it's concerned it might do this, it can never initiate a stream with the EndOfGroup flag.

I am not sure compressing away the EndOfGroup object is well-conceived. It's already opened up a series of logical problems for what is a minor savings.

afrind:
> This implies that if I want extensions for my EndOfGroup object, I need to send that object explicitly. That's fine, I suppose, but it is a little different from when a single object had payload, extensions, and EoG status.

Yes, this is correct.

> I don't understand many of the use cases for extensions, so I don't know if this is a problem.
>
> This gets thorny if the relay decides to add an extension to an EoG object. If it's concerned it might do this, it can never initiate a stream with the EndOfGroup flag.

I can conceive of use cases for extensions in EOG -- metrics covering the entire group or something.

For a relay that might add extensions to EOG objects, it should be clear it can't use this flag.

> I am not sure compressing away the EndOfGroup object is well-conceived. It's already opened up a series of logical problems for what is a minor savings.

It was not well conceived in the first instance as revealed by implementation experience.  I was hoping this fixed it, but as you point out, it's a little fuzzy around extension handling.  Do you think we should remove the feature or try to fix-forward?

martinduke:
What if Object ID is VARINT62_MAX? I can't express the implied EOG object in a FETCH stream.

afrind:
> What if Object ID is VARINT62_MAX? I can't express the implied EOG object in a FETCH stream.

Wait for #1016 ?

martinduke:
> > What if Object ID is VARINT62_MAX? I can't express the implied EOG object in a FETCH stream.
> 
> Wait for #1016 ?

That just moves the problem to a different integer.

martinduke:
Having now implemented the existing spec (not this PR), and realized that this is saving 3 bytes per group under a narrow set of applicability conditions, I believe that all of this superstructure is not worth the lines of spec, lines of code, and ~dozen test cases it spawns. I propose we revert the whole thing.

fluffy:
For me, low bandwidth audio is the case that drives this being worth doing.

fluffy (approved):

afrind:
@vasilvv will address this after making a separate PR for #1223 and #1224

afrind:
#1342 will address this in a more complete way

Issue #870, opened 2025-03-26
Section 7 Minor Issues


martinduke:
Sec 7.1:
- The first paragraph doesn't mention FETCH, and should.

- "it SHOULD withhold its own SUBSCRIBE_OK until receiving one from upstream."  This really ought to be a MUST. First of all, if there is no upstream SUBSCRIBE, sending SUBSCRIBE_OK immediately is irresponsible. If there is a partial upstream SUBSCRIBE, behavior when the upstream sends SUBSCRIBE_ERROR is undefined. We could say SHOULD in this very narrow edge case, but would have to define that behavior. Better to just say MUST.

- "dependent on the congestion response". What does this mean? I think it means "send it as fast as congestion & flow control allow" but that is a condition on literally all things sent in the protocol.

- Re: Cache poisoning. Add an informative reference to RFC 9111?

Sec 7.2
- This section doesn't mention FETCH at all, and should

- The namespace=foobar example is not very helpful, because it doesn't illustrate a non-exact match.

- Mention of track_alias needs to be updated?

- "Relays MUST forward objects to matching subscribers in accordance to each subscription's priority, group order, and delivery timeout". s/in accordance to/in accordance with. Also, shouldn't this be SHOULD? All that stuff is best effort, no? In Sec 7.3 it says relays SHOULD respect priority, so one of these has to change.

- "If an upstream session is closed due to an unknown or invalid control message or Object, the relay MUST NOT continue to propagate that message or Object downstream, because it would enable a single session to close unrelated sessions." This only applies if there are multiple upstreams.

Sec 7.3
- "To mitigate this, a relay MAY read and write stream data out of order subject to flow control limits. See section 2.2 in [[QUIC](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#QUIC)]." What does this mean? It seems to conflict with a lot the assumptions that underlie the FIN/RESET strategy

afrind:
>"it SHOULD withhold its own SUBSCRIBE_OK until receiving one from upstream." This really ought to be a MUST. First of all, if there is no upstream SUBSCRIBE, sending SUBSCRIBE_OK immediately is irresponsible. If there is a partial upstream SUBSCRIBE, behavior when the upstream sends SUBSCRIBE_ERROR is undefined. We could say SHOULD in this very narrow edge case, but would have to define that behavior. Better to just say MUST

This is tracked in #882 

[...]

PR #1119, opened 2025-07-28
Address miscellaneous Relay section feedback

afrind:
Fixes: #870

* Mention Fetch
* Add cache poisoning to security TODO
* Coalesce the two sections describing object forwarding into one
* Remove ambiguous "congestion response"
* Remove inconsistent MUST regarding priority (Priority section says SHOULD)
* Remove old section about out-of-order stream forwarding
* Clean up paragraph on accidental error propagation

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1147,23 +1147,24 @@ A cache MUST store all properties of an Object defined in
 
 ## Subscriber Interactions
 
-Subscribers subscribe to tracks by sending a SUBSCRIBE
-({{message-subscribe-req}}) control message for each track of
-interest. Relays MUST ensure subscribers are authorized to access the
-content associated with the track. The authorization
-information can be part of subscription request itself or part of the
-encompassing session. The specifics of how a relay authorizes a user are outside
-the scope of this specification.
+Subscribers request Tracks by sending a SUBSCRIBE (see
+{{message-subscribe-req}}) or FETCH (see {{message-fetch}}) control message for
+each Track of interest. Relays MUST ensure subscribers are authorized to access
+the content associated with the Track. The authorization information can be part
+of request itself or part of the encompassing session. The specifics of how a
+relay authorizes a user are outside the scope of this specification.
 
 The relay MUST have an established upstream subscription before sending
 SUBSCRIBE_OK in response to a downstream SUBSCRIBE.  If a relay does not have
 sufficient information to send a FETCH_OK immediately in response to a FETCH, it
 MUST withhold sending FETCH_OK until it does.
 
-For successful subscriptions, the publisher maintains a list of
-subscribers for each track. Each new Object belonging to the
-track within the subscription range is forwarded to each active
-subscriber, dependent on the congestion response.
+For successful subscriptions, the publisher maintains a list of subscribers for
+each Track. Relays use the Track Alias ({{track-alias}}) of an incoming Object
+to identify its Track and find the active subscribers.  Each new Object
+belonging to the Track within the subscription range is forwarded to each active
+subscriber, according to the priority (see {{priorities}}) and delivery timeout
+(see {{delivery-timeout}}).
 
 Relays MUST be able to process objects for the same Full Track Name from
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1147,23 +1147,24 @@ A cache MUST store all properties of an Object defined in
...
-For successful subscriptions, the publisher maintains a list of
-subscribers for each track. Each new Object belonging to the
-track within the subscription range is forwarded to each active
-subscriber, dependent on the congestion response.
+For successful subscriptions, the publisher maintains a list of subscribers for
+each Track. Relays use the Track Alias ({{track-alias}}) of an incoming Object
+to identify its Track and find the active subscribers.  Each new Object
+belonging to the Track within the subscription range is forwarded to each active
+subscriber, according to the priority (see {{priorities}}) and delivery timeout
+(see {{delivery-timeout}}).
```
```suggestion
to identify its Track and find the active subscribers.  Each new Object
belonging to the Track is forwarded to each active subscriber, as allowed by the
subscription's filter (see {{message-subscribe-req}}), and delivered according
to the priority (see {{priorities}}) and delivery timeout (see
{{delivery-timeout}}).
```

draft-ietf-moq-transport.md
```diff
@@ -1302,13 +1298,6 @@ prioritize sending Objects based on {{priorities}}.
 A publisher SHOULD begin sending incomplete objects when available to
 avoid incurring additional latency.
 
-A relay that reads from one stream and writes to another in order can
-introduce head-of-line blocking.  Packet loss will cause stream data to
-be buffered in the library, awaiting in-order delivery, which could
-increase latency over additional hops.  To mitigate this, a relay MAY
-read and write stream data out of order subject to flow control
-limits.  See section 2.2 in {{QUIC}}.
```
suhasHere, line 1310: i think this is still needed and pipelining is useful for forwading subscriptions without accruing latency of full object assembly

suhasHere (changes requested):
Looking good overall.
Have a minor suggestion and like to re-add a deleted paragraph

draft-ietf-moq-transport.md
afrind: A couple comments:

1) This explicit MAY isn't really necessary for an implementation that wants to implement this scheme -- it can do whatever it wants, so long as it follows all the other rules laid down by QUIC and MOQT.
2) This behavior is complicated, likely requires customizing existing QUIC stacks and somewhat risky as raised by the linked issue, we shouldn't be encouraging it.
3) We already have language about pipelining objects on the previous line

> A publisher SHOULD begin sending incomplete objects when available to avoid incurring additional latency.

draft-ietf-moq-transport.md
```diff
@@ -1248,15 +1250,10 @@ SUBSCRIBE to the publisher that sent the PUBLISH_NAMESPACE.  When it receives an
...
-and delivery timeout.
-
-If an upstream session is closed due to an unknown or invalid control message
-or Object, the relay MUST NOT continue to propagate that message or Object
-downstream, because it would enable a single session to close unrelated
-sessions.
+If a Session is closed due to an unknown or invalid control message or Object,
+the Relay MUST NOT propagate that message or Object to another Session, because
+it would enable a single Session error to force an unrelated Session, which
+might be handling other subscriptions, to be closed.
```
sharmafb, line 1256: Is there a different mechanism that relays should use to convey to the client that the subscriptions/fetches corresponding to that upstream have been closed for this reason (e.g. a SUBSCRIBE_DONE with a specific error code)?

afrind: Partially, see "Malformed Tracks" for invalid Objects.  An invalid control message would likely result in something like SUBSCRIPTION_ENDED with a reason phrase, at least in our implementation.

draft-ietf-moq-transport.md
suhasHere: I think i am convinced, as the previous line is good enough here.

fluffy (approved):

Issue #876, opened 2025-03-27
Section 8 Issues


martinduke:
8.1.1.2 Delivery Timeout

Maybe this all gets wiped out by #667, but if not:

"If an earlier Object arrives later than subsequent Objects, relays can consider the receipt time as that of the next later Object, with the assumption that the Object's data was reordered."

This phrasing is tortured, and at first reading I thought it meant the opposite. This would be better if we used relative Locations rather than talking about "later" or "subsequent" packets.

8.2.2.1 PATH 

Please add a reference to 3.1.2, maybe eliminating the duplicate text in one of the places? When I read two separate specifications for something, I'm never sure which one is right.

8.3 GOAWAY

"draining state" is used without definition. Maybe just say "after sending a GOAWAY"?

8.4 MAX_SUBSCRIBE_ID (and 8.5)

No mention of FETCH, even though it's equally affected. Once again, a collective term for subscribe and fetch would be useful (as noted  in #860.

8.6 SUBSCRIBE

Unknown filters "MUST be treated as error." Protocol or Subscribe error?

8.8 SUBSCRIBE_ERROR

s/connection/session

8.11 SUBSCRIBE_DONE

UNSUBSCRIBE does not trigger SUBSCRIBE_DONE

8.12 FETCH

Start and End are locations; let's use the new notation.

8.12.1 

Fetch EndObject is always Subscribe Largest Object + 1, given the encoding. All the text after is wrong.

To be pedantic, Fetch StartGroup is Max(0, Subscribe Largest Group - Joining Start)

afrind:
> This phrasing is tortured, and at first reading I thought it meant the opposite. This would be better if we used relative Locations rather than talking about "later" or "subsequent" packets.

This now reads:

>  If an Object with a smaller ID arrives later than subsequent
Objects, relays can consider its receipt time as that of the Object with the
next larger Location, with the assumption that the Objects were reordered.

But the "subsequent Objects" is still there and it might still be confusing.  

Maybe:

If a relay receives an Object with Location L smaller than the Largest Object (see section Y), it can consider its receipt time as that of the Object with smallest Location larger than L, with the assumption that the Objects were reordered.


> 8.4 MAX_SUBSCRIBE_ID (and 8.5)
> No mention of FETCH

This has been resolved.

> 8.8 SUBSCRIBE_ERROR
> s/connection/session

This section on longer contains 'connection'

> UNSUBSCRIBE does not trigger SUBSCRIBE_DONE

[...]

PR #1120, opened 2025-07-28
Clarify DELIVERY_TIMEOUT reordering computation

afrind:
Addresses part of #876

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1573,10 +1573,7 @@ that an Object was not sent because the delivery timeout was exceeded.
 If both the subscriber and publisher specify the parameter, they use the min of
 the two values for the subscription.  The publisher SHOULD always specify the
 value received from an upstream subscription when there is one, and nothing
-otherwise.  If a Relay receives an Object with Location L smaller than the
-Largest Object (see {{message-subscribe-req}}), it can consider its receipt time
-as that of the Object with smallest Location larger than L, with the assumption
-that the Objects were reordered.
+otherwise.
 
 Publishers can, at their discretion, discontinue forwarding Objects earlier than
 the negotiated DELIVERY TIMEOUT, subject to stream closure and ordering


afrind:
> Note that this does not fix all of #876, though the tooling might close that issue.

Oops I thought everything else was already handled but rereading I missed some.  I will update.

draft-ietf-moq-transport.md
```diff
@@ -1573,9 +1573,10 @@ that an Object was not sent because the delivery timeout was exceeded.
 If both the subscriber and publisher specify the parameter, they use the min of
 the two values for the subscription.  The publisher SHOULD always specify the
 value received from an upstream subscription when there is one, and nothing
-otherwise.  If an Object with a smaller ID arrives later than subsequent
-Objects, relays can consider its receipt time as that of the Object with the
-next larger Location, with the assumption that the Objects were reordered.
+otherwise.  If a Relay receives an Object with Location L smaller than the
+Largest Object (see {{message-subscribe-req}}), it can consider its receipt time
+as that of the Object with smallest Location larger than L, with the assumption
+that the Objects were reordered.
```
suhasHere: May be this is just me,  I feel the new text is still confusing

afrind: @suhasHere : is it more or less confusing than the existing version?  Do you (or your favorite AI) have suggested text?  I'm trying to remove "subsequent" and "next larger", and change "ID" to "Location".

fluffy: actually, I think both version of this text are just wrong. I don't see how it works with object that were not reordered but deliberately sent at very very different times

afrind: @fluffy if we scope this sentence to within the same Group does that help?

fluffy: Uh, scoping to same subgroup might help but I think I am missing what this is trying acomplish

fluffy:
This seems like a design issue not minor design - if I am reading it correctly it seems to be non trivial change to how expiry would work - I'm also not sure I am understand all the context how what got us to this and might be a bit confused so help me get up to speed on it.

draft-ietf-moq-transport.md
afrind: This can't happen in a _subgroup_ because they are ordered by the transport and Object IDs must increase there.

afrind:
> I'm also not sure I am understand all the context how what got us to this and might be a bit confused so help me get up to speed on it.

From the git history, this sentence appeared in the first DELIVERY_TIMEOUT PR/commit in (#486) authored by @ianswett.  I don't know if we really need it or if we could just delete it and let Relays figure it out?

> This seems like a design issue not minor design - if I am reading it correctly it seems to be non trivial change to how expiry would work - 

This PR is not attempting to change any logic in the draft that has been consensus-called, only attempting to clarify it, so I don't think it's Design as written.  But if we're asking to remove this bit then it would be.

draft-ietf-moq-transport.md
yekuiwang: > Uh, scoping to same subgroup might help but I think I am missing what this is trying acomplish

I was also wondering why to have this sentence at all. Is there a problem to just remove this sentence?

afrind:
I could see this just being deleted.  Naïve relays will just do the normal delivery timeout thing with reordered objects and the worst case is that some objects are sent when they logically should have timed out. I can imagine an OP producing objects out of object ID order because of subgroups, eg: 

A: 1, 2, 3, 4
B: 101, 102, 103, 104

which is more fetch-cache friendly.  I don't think that impacts the delivery-timeout-reordering, because subgroups are already ordered, but it means that Object IDs within a group are not required to be temporal.  Or if folks disagree we should have a discussion about that.

draft-ietf-moq-transport.md
```suggestion
otherwise.
```

ianswett (approved):

vasilvv (approved):

PR #1123, opened 2025-07-30
Relay security considerations: SUBSCRIBE_NAMESPACE with short prefix

### LINKED ISSUE #1097: Relay security consideration: SUBSCRIBE_NAMESPACE with a short prefix
If an entity sends a relay a SUBSCRIBE_NAMESPACE message with a short prefix, it can cause the relay to send a lot of ANNOUNCE messages. As churn continues in the tree of prefixes being announced/unannounced, the relay would have to continue to send ANNOUNCE/UNANNOUNCE messages to the entity that had sent the SUBSCRIBE_NAMESPACE.

I was thinking of adding some text in the "Security Considerations" section in order to encourage Relay implementations to have protections in place for this. For instance, by only allowing SUBSCRIBE_NAMESPACE messages to more specific prefixes. Thoughts?

sharmafb:
Fixes #1097

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3690,7 +3690,9 @@ exhaustion attacks by a peer that does not send expected data within
 an expected time.  Each implementation is expected to set its own limits.
 
 
-## Relay memory consumption
+## Relay security considerations
+
+### State maintenance
 
 A Relay SHOULD have mechanisms to prevent malicious endpoints from flooding it
 with PUBLISH_NAMESPACE or SUBSCRIBE_NAMESPACE requests that could bloat data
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -3702,6 +3704,15 @@ such requests, or could have application-specific policies that can reject
 incoming PUBLISH_NAMESPACE or SUBSCRIBE_NAMESPACE requests that cause the state
 maintenance for the session to be excessive.
 
+### SUBSCRIBE_NAMESPACE with short prefixes
+
+A Relay SHOULD use authorization rules in order to prevent subscriptions closer
```
fluffy: my suggestion would be to change "SHOULd" to "can"

Subscribing to the whole namespace seems fairly common for things like metrics and logging of what is going on. I agree with the concern you raise here and agree with solution to it - I'm just not sure that means an relay can't support this.

afrind:
Please update ANNOUNCE to PUBLISH_NAMESPACE and UNANNOUNCE to PUBLISH_NAMESPACE_DONE to match current draft terminology.

sharmafb:
Addressed comments.

draft-ietf-moq-transport.md
```diff
@@ -3702,6 +3704,15 @@ such requests, or could have application-specific policies that can reject
 incoming PUBLISH_NAMESPACE or SUBSCRIBE_NAMESPACE requests that cause the state
 maintenance for the session to be excessive.
 
+### SUBSCRIBE_NAMESPACE with short prefixes
+
+A Relay can use authorization rules in order to prevent subscriptions closer
+to the root of a large prefix tree. Otherwise, if an entity sends a relay a
+SUBSCRIBE_NAMESPACE message with a short prefix, it can cause the relay to send
+a lot of PUBLISH_NAMESPACE messages. As churn continues in the tree of prefixes,
```
afrind: s/a lot/a large volume/

sharmafb: Addressed.

Issue #1126, opened 2025-08-04
Meaning of "or a more generic set including that namespace" in Section 5.2


yekuiwang:
Section 5.2 includes the following sentence:

> If a publisher is authoritative for a given namespace, or is a relay that has received an authorized ANNOUNCE for that namespace from an upstream publisher, it MUST send an ANNOUNCE to any subscriber that has subscribed to ANNOUNCE for that namespace, or a more generic set including that namespace.

I could not figure out what the final part, i.e., "or a more generic set including that namespace" means in the context. Any clarification would be appreciated. And whatever the clarification is, it might be good to have the clarification done in the text in the next version of the draft.

afrind:
If a relay receives Namespace = (a, b, c), it MUST send that to any session that has sent SUBSCRIBE_NAMESPACE (a, b, c), (a, b) or (a).

How about:

> it MUST send an ANNOUNCE to any subscriber that has subscribed to ANNOUNCE for that namespace, or a prefix of that namespace

PR #1138, opened 2025-08-05
more generic set -> prefix of

afrind:
Fixes: #1126

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -958,9 +958,8 @@ in a namespace without having received a PUBLISH_NAMESPACE for it.
 If a publisher is authoritative for a given namespace, or is a relay that has
 received an authorized PUBLISH_NAMESPACE for that namespace from an upstream
 publisher, it MUST send a PUBLISH_NAMESPACE to any subscriber that has
-subscribed via SUBSCRIBE_NAMESPACE for that namespace, or a more generic set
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -958,9 +958,8 @@ in a namespace without having received a PUBLISH_NAMESPACE for it.
 If a publisher is authoritative for a given namespace, or is a relay that has
 received an authorized PUBLISH_NAMESPACE for that namespace from an upstream
 publisher, it MUST send a PUBLISH_NAMESPACE to any subscriber that has
-subscribed via SUBSCRIBE_NAMESPACE for that namespace, or a more generic set
-including that namespace. A publisher MAY send the PUBLISH_NAMESPACE to any
-other subscriber.
+subscribed via SUBSCRIBE_NAMESPACE for that namespace, or a prefix of that
+namespace. A publisher MAY send the PUBLISH_NAMESPACE to any other subscriber.
```
suhasHere, line 962: Looks good, but one comment. I don't see why we have "A publisher MAY send the PUBLISH_NAMESPACE to any other subscriber."

afrind: You'll have to file another issue for that.  In general, the spec is loose about direct publisher <-> subscriber interactions, as opposed to ones that go through a relay.

Issue #1141, opened 2025-08-05
Semantics of SUBSCRIBE_UPDATE


yekuiwang:
> The Start Location MUST NOT decrease, and if it increases, there is no guarantee that the publisher has not already sent Objects with Locations smaller than the new Start Location. Similarly, the End Group MUST NOT increase, and if it decreases, there is no guarantee that the publisher has not already sent Objects with Locations larger than the new End Location.

I can understand the second sentence above, as it talks about that there is no guarantee that the publisher's operation is within the new range. However, I cannot figure out what the first sentence is talking about.

Was the phrase "no guarantee that the publisher has not already sent Objects with Locations smaller than the new Start Location" in the first sentence intended to be "no guarantee that the publisher has already sent all Objects with Locations smaller than the new Start Location", such that the first sentence would also talk about that there is no guarantee that the publisher's operation is within the new range?

afrind:
It's supposed to communicate:

Old Start Location = x
New Start Location = y

Publisher might have sent an object with `x < location  < y` between when the subscriber sends UPDATE and when the publisher receives it.

Maybe we can just say narrowing the subscription is subject to race conditions that might cause the subscriber to receive objects outside its currently specified filter.

Also, I think there was consensus to allow widening the filter and @ianswett was going to make that change?

Issue #1140, opened 2025-08-05
Semantics of Next Group Start


yekuiwang:
> Next Group Start (0x1): The filter start Location is {Largest Object.Group + 1, 0} and Largest Object is communicated in SUBSCRIBE_OK. If no content has been delivered yet, the filter Start Location is {0, 0}. There is no End Group - the subscription is open ended. For scenarios where the subscriber intends to start more than one group in the future, it can use an AbsoluteStart filter instead.

1. "start Location" -> "Start Location"
2. "to start more than one group in the future" -> "to start from more than one group in the future"

afrind:
Sounds fine.

afrind:
@yekuiwang : the editors are currently quite swamped with resolving issues.  We _greatly_ appreciate the thorough reviews, and I've tried to either acknowledge your proposed fix in the issue or made minor suggestions.  Can you take the lead in authoring one or more PRs to resolve these issues - anything labeled Editorial and not assigned to someone else is fair game?  We will review and merge.

yekuiwang:
> [@yekuiwang](https://github.com/yekuiwang) : the editors are currently quite swamped with resolving issues. We _greatly_ appreciate the thorough reviews, and I've tried to either acknowledge your proposed fix in the issue or made minor suggestions. Can you take the lead in authoring one or more PRs to resolve these issues - anything labeled Editorial and not assigned to someone else is fair game? We will review and merge.

Sure. Will do.

yekuiwang:
@afrind I just generated [PR#1143](https://github.com/moq-wg/moq-transport/pull/1143) that covers this issue and 9 other issues.

Issue #1139, opened 2025-08-05
Possibly incorrect cross referencing for Request ID


yekuiwang:
> Maximum Request ID: The Maximum Request ID for the session on which the endpoint is blocked. More on Request ID in [Section 8.7](https://www.ietf.org/archive/id/draft-ietf-moq-transport-13.html#message-subscribe-req).

The cross referencing for Request ID to Section 8.7 should probably be [Section 8.1](https://www.ietf.org/archive/id/draft-ietf-moq-transport-13.html#request-id).

afrind:
Yes, we originally had "Subscribe IDs" but changed to "Request IDs".

Issue #1134, opened 2025-08-05
An Object being delivered before sending any data of that Object


yekuiwang:
Section 8.2.1.2 (DELIVERY TIMEOUT Parameter) includes the following sentence:

> Publishers SHOULD consider whether the entire Object is likely to be delivered before sending any data for that Object, taking into account priorities, congestion control, and any other relevant information.

The first half of the sentence reads weird to me, as before sending any data for an Object, that Object won't be delivered in any case, right? Was "data" intended to be something like "metadata"?

sharmafb:
In this context “before sending any data for that Object” simply means “before you put the first byte of the Object on the wire.”
The guidance is: think ahead. If, given the current priorities, congestion-control state, delivery-timeout value, etc., you judge that you probably will not be able to transmit the whole Object before the timeout fires, you ought to refrain from starting to transmit it at all.

yekuiwang:
@sharmafb Got you. Thanks! With that understanding, it seems that it would be much clearer to change the sentence to be something like the following:

   Publishers SHOULD consider whether the entire Object can likely be successfully delivered within the timeout period before sending any data for that Object, taking into account priorities, congestion control, and any other relevant information.

afrind:
Your suggestion looks good.

PR #1143, opened 2025-08-06
Misc. editorial changes

yekuiwang:
The changes are for fixing the following issues:
1.	[#951](https://github.com/moq-wg/moq-transport/issues/951): Editorial issues in draft-ietf-moq-transport-10, section 2
2.	[#955](https://github.com/moq-wg/moq-transport/issues/955): On assigning Objects to Subgroups
3.	[#1127](https://github.com/moq-wg/moq-transport/issues/1127): Clarifications likely needed for "SHOULD report this to the application" in Section 5.2
4.	[#1128](https://github.com/moq-wg/moq-transport/issues/1128): Possible issues in handling of objects when subscriber priority is changed
5.	[#1129](https://github.com/moq-wg/moq-transport/issues/1129): Move last paragraph in Section 6.1 to Section 2.3.1
6.	[#1133](https://github.com/moq-wg/moq-transport/issues/1133): Wording of "neither the subscriber or publisher specify"
7.	[#1134](https://github.com/moq-wg/moq-transport/issues/1134): An Object been delivered before sending any data of that Object
8.	[#1139](https://github.com/moq-wg/moq-transport/issues/1139): Possibly incorrect cross referencing for Request ID
9.	[#1140](https://github.com/moq-wg/moq-transport/issues/1140): Semantics of Next Group Start
10.	[#1141](https://github.com/moq-wg/moq-transport/issues/1141): Semantics of SUBSCRIBE_UPDATE

Fixes: #951
Fixes: #955
Fixes: #1127
Fixes: #1128
Fixes: #1129
Fixes: #1133
Fixes: #1134
Fixes: #1139
Fixes: #1140
Fixes: #1141

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -431,9 +431,9 @@ ID without implying any relationship between them. In general, publishers assign
 objects to subgroups in order to leverage the features of streams as described
 above.
 
-When assigning Objects to different Subgroups, a reasonable tradeoff should be considered
-between having a good number of Subgroups (hence a good number of different priorities) in
-a group and minimizing the number of streams used.
+When assigning Objects to different Subgroups, a reasonable tradeoff should be
+considered between having a good number of Subgroups (hence a good number of
+different priorities) in a group and minimizing the number of streams used.
 
 ## Groups {#model-group}
 
@@ -957,9 +957,10 @@ namespace that exactly matches a namespace for which the peer sent an earlier
 PUBLISH_NAMESPACE (i.e. a PUBLISH_NAMESPACE ought not to be echoed back to its
 sender).
 
-An endpoint SHOULD report the reception of an ANNOUNCE_OK or ANNOUNCE_ERROR to
-the application to inform the search for additional subscribers for a namespace,
-or to abandon the attempt to publish under this namespace. This might
+An endpoint SHOULD report the reception of an PUBLISH_NAMESPACE_OK or
+PUBLISH_NAMESPACE_ERROR to the application to inform the search for additional
+subscribers for a namespace, or to abandon the attempt to publish under this
+namespace. This might
 be especially useful in upload or chat applications. A subscriber MUST send
 exactly one PUBLISH_NAMESPACE_OK or PUBLISH_NAMESPACE_ERROR in response to an
 PUBLISH_NAMESPACE. The publisher SHOULD close the session with a protocol error
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3617,7 +3617,7 @@ Objects in Group 10, as it sees fit.  A Track is considered malformed (see
 This extension is optional, as publishers might not know the prior gap gize, or
 there may not be a gap. If Prior Group ID Gap is not present, the receiver
 cannot infer any information about the existence of prior groups (see
-{{group-ordering}}).
+{{group-id-allocation}}).
 
 This extension can be added by the Original Publisher, but MUST NOT be added by
 relays. This extension MUST NOT be modified or removed.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -372,11 +372,11 @@ the display environment and available bandwidth. Each independently
 coded sequence of pictures in a resolution is sent as a group as the
 first picture in the sequence can be used as a random access point.
 This allows the client to join at the logical points where decoding
-of the stream can start without needing information before the join
+of the media can start without needing information before the join
 points. The temporal layers are sent as separate subgroups to allow
 the priority mechanism to favor lower temporal layers when there is
-not enough bandwidth to send all temporal layers. Each picture is sent
-as a single object.
+not enough bandwidth to send all temporal layers. Each frame of video
+is sent as a single object.
 
 ## Objects {#model-object}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -431,9 +431,15 @@ ID without implying any relationship between them. In general, publishers assign
 objects to subgroups in order to leverage the features of streams as described
 above.
 
-When assigning Objects to different Subgroups, a reasonable tradeoff should be
-considered between having a good number of Subgroups (hence a good number of
-different priorities) in a group and minimizing the number of streams used.
+In general, if Object B is dependent on Object A, then delivery of B can follow
+A, i.e. A and B can be usefully delivered over a single stream.  If an Object is
+dependent on all previous Objects in a Subgroup, it likely fits best in that
+Subgroup.  If an Object is not dependent on any of the Objects in a Subgroup, it
+likely belongs in a different Subgroup.
+
+When assigning Objects to different Subgroups, the Original Publisher makes a
+reasonable tradeoff between having an optimal mapping of Object relationships in
+a Group and minimizing the number of streams used.
 
 ## Groups {#model-group}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -450,7 +450,7 @@ point for subscriptions, so a subscriber that does not want to receive the
 entire Track can opt to receive only Groups starting from a given Group ID.
 Groups can contain any number of Objects.
 
-### Group ID Allocation
+### Group IDs
 
 Within a track, the original publisher SHOULD publish Group IDs which increase
 with time (where "time" is defined according to the internal clock of the media
@@ -3625,7 +3625,7 @@ Objects in Group 10, as it sees fit.  A Track is considered malformed (see
 This extension is optional, as publishers might not know the prior gap gize, or
 there may not be a gap. If Prior Group ID Gap is not present, the receiver
 cannot infer any information about the existence of prior groups (see
-{{group-id-allocation}}).
+{{group-ids}}).
 
 This extension can be added by the Original Publisher, but MUST NOT be added by
 relays. This extension MUST NOT be modified or removed.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -962,7 +962,7 @@ PUBLISH_NAMESPACE_ERROR to the application to inform the search for additional
 subscribers for a namespace, or to abandon the attempt to publish under this
 namespace. This might be especially useful in upload or chat applications. A
 subscriber MUST send exactly one PUBLISH_NAMESPACE_OK or PUBLISH_NAMESPACE_ERROR
-in response to an PUBLISH_NAMESPACE. The publisher SHOULD close the session with
+in response to a PUBLISH_NAMESPACE. The publisher SHOULD close the session with
 a protocol error if it receives more than one.
 
 A PUBLISH_NAMESPACE_DONE message withdraws a previous PUBLISH_NAMESPACE,
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -368,14 +368,15 @@ To give an example of how an application might use this data model,
 consider an application sending high and low resolution video using a
 codec with temporal scalability. Each resolution is sent as a separate
 track to allow the subscriber to pick the appropriate resolution given
-the display environment and available bandwidth. Each "group of pictures"
-in a video is sent as a group because the first frame is needed to
-decode later frames. This allows the client to join at the logical points
-where they can get the information to start decoding the media.
-The temporal layers are sent as separate sub groups to allow the
-priority mechanism to favour the base layer when there is not enough
-bandwidth to send both the base and enhancement layers. Each frame of
-video on a given layer is sent as a single object.
+the display environment and available bandwidth. Each independently
+coded sequence of pictures in a resolution is sent as a group as the
+first picture in the sequence can be used as a random access point.
+This allows the client to join at the logical points where decoding
+of the stream can start without needing information before the join
+points. The temporal layers are sent as separate subgroups to allow
+the priority mechanism to favor lower temporal layers when there is
+not enough bandwidth to send all temporal layers. Each picture is sent 
+as a single object.
 
 ## Objects {#model-object}
 
@@ -430,20 +431,9 @@ ID without implying any relationship between them. In general, publishers assign
 objects to subgroups in order to leverage the features of streams as described
 above.
 
-An example strategy for using stream properties follows. If object B is
-dependent on object A, then delivery of B can follow A, i.e. A and B can be
-usefully delivered over a single stream. Furthermore, in this example:
-
-- If an object is dependent on all previous objects in a Subgroup, it is added to
-that Subgroup.
-
-- If an object is not dependent on all of the objects in a Subgroup, it goes in
-a different Subgroup.
-
-- There are often many ways to compose Subgroups that meet these criteria. Where
-possible, choose the composition that results in the fewest Subgroups in a group
-to minimize the number of streams used.
-
+When assigning Objects to different Subgroups, a reasonable tradeoff should be considered
+between having a good number of Subgroups (hence a good number of different priorities) in
+a group and minimizing the number of streams used.
 
 ## Groups {#model-group}
 
@@ -454,7 +444,7 @@ point for subscriptions, so a subscriber that does not want to receive the
 entire Track can opt to receive only Groups starting from a given Group ID.
 Groups can contain any number of Objects.
 
-### Group Ordering
+### Group ID Allocation
 
 Within a track, the original publisher SHOULD publish Group IDs which increase
 with time (where "time" is defined according to the internal clock of the media
@@ -967,9 +957,9 @@ namespace that exactly matches a namespace for which the peer sent an earlier
 PUBLISH_NAMESPACE (i.e. a PUBLISH_NAMESPACE ought not to be echoed back to its
 sender).
 
-The receiver of a PUBLISH_NAMESPACE_OK or PUBLISH_NAMESPACE_ERROR SHOULD report
-this to the application to inform the search for additional subscribers for a
-namespace, or abandoning the attempt to publish under this namespace. This might
+An endpoint SHOULD report the reception of an ANNOUNCE_OK or ANNOUNCE_ERROR to
+the application to inform the search for additional subscribers for a namespace,
+or to abandon the attempt to publish under this namespace. This might
 be especially useful in upload or chat applications. A subscriber MUST send
 exactly one PUBLISH_NAMESPACE_OK or PUBLISH_NAMESPACE_ERROR in response to an
 PUBLISH_NAMESPACE. The publisher SHOULD close the session with a protocol error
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -430,20 +431,9 @@ ID without implying any relationship between them. In general, publishers assign
...
-- If an object is not dependent on all of the objects in a Subgroup, it goes in
-a different Subgroup.
-
-- There are often many ways to compose Subgroups that meet these criteria. Where
-possible, choose the composition that results in the fewest Subgroups in a group
-to minimize the number of streams used.
-
+When assigning Objects to different Subgroups, a reasonable tradeoff should be considered
+between having a good number of Subgroups (hence a good number of different priorities) in
+a group and minimizing the number of streams used.
```
```suggestion
When assigning Objects to different Subgroups, a reasonable tradeoff should be
considered between having a good number of Subgroups (hence a good number of
different priorities) in a group and minimizing the number of streams used.
```

draft-ietf-moq-transport.md
```diff
@@ -967,9 +957,9 @@ namespace that exactly matches a namespace for which the peer sent an earlier
 PUBLISH_NAMESPACE (i.e. a PUBLISH_NAMESPACE ought not to be echoed back to its
 sender).
 
-The receiver of a PUBLISH_NAMESPACE_OK or PUBLISH_NAMESPACE_ERROR SHOULD report
-this to the application to inform the search for additional subscribers for a
-namespace, or abandoning the attempt to publish under this namespace. This might
+An endpoint SHOULD report the reception of an ANNOUNCE_OK or ANNOUNCE_ERROR to
+the application to inform the search for additional subscribers for a namespace,
+or to abandon the attempt to publish under this namespace. This might
```
```suggestion
An endpoint SHOULD report the reception of an PUBLISH_NAMESPACE_OK or
PUBLISH_NAMESPACE_ERROR to the application to inform the search for additional
subscribers for a namespace, or to abandon the attempt to publish under this
namespace. This might
```

draft-ietf-moq-transport.md
afrind: I'm not sure we want to delete all of the text regarding dependencies.

draft-ietf-moq-transport.md
afrind: Some additional text wrapping is needed for this paragraph now that I fixed ANNOUNCE/PUBLISH_NAMESPACE.

draft-ietf-moq-transport.md
yekuiwang: Hmm, then it'd be great if you could read the problems I pointed out in https://github.com/moq-wg/moq-transport/issues/955 and suggest how to resolve them otherwise.

draft-ietf-moq-transport.md
yekuiwang: I am not in the context of the fix you mentioned and don't know what to do, thus I am expecting you to suggest the additional text changes.

afrind: I meant my suggested text which replaced ANNOUNCE with the new PUBLISH_NAMESPACE.  I just merged it into your branch; now the whole paragraph could use a reflow.

draft-ietf-moq-transport.md
```diff
@@ -368,14 +368,15 @@ To give an example of how an application might use this data model,
...
-video on a given layer is sent as a single object.
+the display environment and available bandwidth. Each independently
+coded sequence of pictures in a resolution is sent as a group as the
+first picture in the sequence can be used as a random access point.
+This allows the client to join at the logical points where decoding
+of the stream can start without needing information before the join
+points. The temporal layers are sent as separate subgroups to allow
+the priority mechanism to favor lower temporal layers when there is
+not enough bandwidth to send all temporal layers. Each picture is sent 
+as a single object.
```
```suggestion
the display environment and available bandwidth. Each independently
coded sequence of pictures in a resolution is sent as a group as the
first picture in the sequence can be used as a random access point.
This allows the client to join at the logical points where decoding
of the stream can start without needing information before the join
points. The temporal layers are sent as separate subgroups to allow
the priority mechanism to favor lower temporal layers when there is
not enough bandwidth to send all temporal layers. Each picture is sent
as a single object.
```

yekuiwang (commented):
"an PUBLISH_NAMESPACE_OK" -> "a PUBLISH_NAMESPACE_OK"

draft-ietf-moq-transport.md
```diff
@@ -2064,14 +2057,12 @@ A subscriber sends a SUBSCRIBE_UPDATE to a publisher to modify an existing
...
-Objects with Locations larger than the new End Location.  A publisher MUST
-terminate the session with a `PROTOCOL_VIOLATION` if the SUBSCRIBE_UPDATE
-violates these rules or if the subscriber specifies a request ID that has not
-existed within the Session.
+them. The Start Location MUST NOT decrease and the End Group MUST NOT increase.
+However, when a subscriber sends a SUBSCRIBE_UPDATE to narrow a subscription,
+there is no guarantee that the subscriber won't receive Objects outside the 
+narrowed range. A publisher MUST terminate the session with a
+`PROTOCOL_VIOLATION` if the SUBSCRIBE_UPDATE violates these rules or if the
+subscriber specifies a request ID that has not existed within the Session.
```
```suggestion
them. The Start Location MUST NOT decrease and the End Group MUST NOT increase.
However, when a subscriber sends a SUBSCRIBE_UPDATE to narrow a subscription,
there is no guarantee that the subscriber won't receive Objects outside the
narrowed range. A publisher MUST terminate the session with a
`PROTOCOL_VIOLATION` if the SUBSCRIBE_UPDATE violates these rules or if the
subscriber specifies a request ID that has not existed within the Session.
```

draft-ietf-moq-transport.md
```diff
@@ -957,9 +957,10 @@ namespace that exactly matches a namespace for which the peer sent an earlier
 PUBLISH_NAMESPACE (i.e. a PUBLISH_NAMESPACE ought not to be echoed back to its
 sender).
 
-An endpoint SHOULD report the reception of an ANNOUNCE_OK or ANNOUNCE_ERROR to
-the application to inform the search for additional subscribers for a namespace,
-or to abandon the attempt to publish under this namespace. This might
+An endpoint SHOULD report the reception of an PUBLISH_NAMESPACE_OK or
```
yekuiwang: "an PUBLISH_NAMESPACE_OK" ->  "a PUBLISH_NAMESPACE_OK"

draft-ietf-moq-transport.md
```diff
@@ -967,13 +957,13 @@ namespace that exactly matches a namespace for which the peer sent an earlier
...
-exactly one PUBLISH_NAMESPACE_OK or PUBLISH_NAMESPACE_ERROR in response to an
-PUBLISH_NAMESPACE. The publisher SHOULD close the session with a protocol error
-if it receives more than one.
+An endpoint SHOULD report the reception of a PUBLISH_NAMESPACE_OK or
+PUBLISH_NAMESPACE_ERROR to the application to inform the search for additional
+subscribers for a namespace, or to abandon the attempt to publish under this
+namespace. This might be especially useful in upload or chat applications. A
+subscriber MUST send exactly one PUBLISH_NAMESPACE_OK or PUBLISH_NAMESPACE_ERROR
+in response to an PUBLISH_NAMESPACE. The publisher SHOULD close the session with
+a protocol error if it receives more than one.
```
```suggestion
in response to a PUBLISH_NAMESPACE. The publisher SHOULD close the session with
a protocol error if it receives more than one.
```

draft-ietf-moq-transport.md
```diff
@@ -430,20 +431,9 @@ ID without implying any relationship between them. In general, publishers assign
...
-- If an object is not dependent on all of the objects in a Subgroup, it goes in
-a different Subgroup.
-
-- There are often many ways to compose Subgroups that meet these criteria. Where
-possible, choose the composition that results in the fewest Subgroups in a group
-to minimize the number of streams used.
-
+When assigning Objects to different Subgroups, a reasonable tradeoff should be
+considered between having a good number of Subgroups (hence a good number of
+different priorities) in a group and minimizing the number of streams used.
```
```suggestion
In general, if Object B is dependent on Object A, then delivery of B can follow
A, i.e. A and B can be usefully delivered over a single stream.  If an Object is
dependent on all previous Objects in a Subgroup, it likely fits best in that
Subgroup.  If an Object is not dependent on any of the Objects in a Subgroup, it
likely belongs in a different Subgroup.

When assigning Objects to different Subgroups, the Original Publisher makes a
reasonable tradeoff between having an optimal mapping of Object relationships in
a Group and minimizing the number of streams used.
```

afrind: @yekuiwang How does this suggested text look to you?  Also CC: @martinduke who authored the original text.

Note a somewhat subtle change from the original:

was:

> If an Object is not dependent on _all_ of the Objects in a Subgroup

to 

> If an Object is not dependent on _any_ of the Objects in a Subgroup

I think cases where there are partial dependencies are best resolved by the Original Publisher that understands them, but if there's no dependencies at all then another subgroup is probably better.

martinduke: I much prefer @afrind's wording. An example is useful, and the original PR doesn't provide any hints at all.

yekuiwang: As I commented in the github issue, it is OK to me anyway as long as the potential issues have been considered such that a fully informed decision is made. The current new wording is at least an improvement to the old one, including the subtle change mentioned and the softening.

afrind (approved):

draft-ietf-moq-transport.md
```diff
@@ -368,14 +368,15 @@ To give an example of how an application might use this data model,
...
-where they can get the information to start decoding the media.
-The temporal layers are sent as separate sub groups to allow the
-priority mechanism to favour the base layer when there is not enough
-bandwidth to send both the base and enhancement layers. Each frame of
-video on a given layer is sent as a single object.
+the display environment and available bandwidth. Each independently
+coded sequence of pictures in a resolution is sent as a group as the
+first picture in the sequence can be used as a random access point.
+This allows the client to join at the logical points where decoding
+of the stream can start without needing information before the join
```
suhasHere: "decoding of stream" seems bit ambiguous. I prefer the original text where it is referring to decoding of the media,

draft-ietf-moq-transport.md
```diff
@@ -368,14 +368,15 @@ To give an example of how an application might use this data model,
...
+points. The temporal layers are sent as separate subgroups to allow
+the priority mechanism to favor lower temporal layers when there is
+not enough bandwidth to send all temporal layers. Each picture is sent
```
suhasHere: s/picture/frame of video

draft-ietf-moq-transport.md
```diff
@@ -454,7 +450,7 @@ point for subscriptions, so a subscriber that does not want to receive the
 entire Track can opt to receive only Groups starting from a given Group ID.
 Groups can contain any number of Objects.
 
-### Group Ordering
+### Group ID Allocation
```
suhasHere: Group Ordering seems more appropriate

draft-ietf-moq-transport.md
yekuiwang: Agreed that "decoding of the media" is better herein, as "stream" has a special meaning in this document, which I did not realize when drafting the changes of this entire paragraph.

draft-ietf-moq-transport.md
```diff
@@ -2064,14 +2062,12 @@ A subscriber sends a SUBSCRIBE_UPDATE to a publisher to modify an existing
...
-Objects with Locations larger than the new End Location.  A publisher MUST
-terminate the session with a `PROTOCOL_VIOLATION` if the SUBSCRIBE_UPDATE
-violates these rules or if the subscriber specifies a request ID that has not
-existed within the Session.
+them. The Start Location MUST NOT decrease and the End Group MUST NOT increase.
+However, when a subscriber sends a SUBSCRIBE_UPDATE to narrow a subscription,
+there is no guarantee that the subscriber won't receive Objects outside the
+narrowed range. A publisher MUST terminate the session with a
+`PROTOCOL_VIOLATION` if the SUBSCRIBE_UPDATE violates these rules or if the
+subscriber specifies a request ID that has not existed within the Session.
```
suhasHere: I am big supporter of brevity, but in this case the extra details matters i feel

draft-ietf-moq-transport.md
yekuiwang: @afrind suggested to change the title of "Group Ordering" in context of https://github.com/moq-wg/moq-transport/issues/1129. So given the context of https://github.com/moq-wg/moq-transport/issues/1129, you still think having the title "Group Ordering" herein is more appropriate?

draft-ietf-moq-transport.md
yekuiwang: @suhasHere OK, so maybe you can take a look at https://github.com/moq-wg/moq-transport/issues/1141 and suggest changes that would resolve the issue pointed out there.

draft-ietf-moq-transport.md
yekuiwang: @suhasHere Sorry I don't understand the comment. Can you clarify the point a bit more?

yekuiwang:
Test, to see how to make the PR link sign in the Type column appear in the list of github issues.

The changes are for fixing the following issues:

https://github.com/moq-wg/moq-transport/issues/951: Editorial issues in draft-ietf-moq-transport-10, section 2
https://github.com/moq-wg/moq-transport/issues/955: On assigning Objects to Subgroups
https://github.com/moq-wg/moq-transport/issues/1127: Clarifications likely needed for "SHOULD report this to the application" in Section 5.2
https://github.com/moq-wg/moq-transport/issues/1128: Possible issues in handling of objects when subscriber priority is changed
https://github.com/moq-wg/moq-transport/issues/1129: Move last paragraph in Section 6.1 to Section 2.3.1
https://github.com/moq-wg/moq-transport/issues/1133: Wording of "neither the subscriber or publisher specify"
https://github.com/moq-wg/moq-transport/issues/1134: An Object been delivered before sending any data of that Object
https://github.com/moq-wg/moq-transport/issues/1139: Possibly incorrect cross referencing for Request ID
https://github.com/moq-wg/moq-transport/issues/1140: Semantics of Next Group Start
https://github.com/moq-wg/moq-transport/issues/1141: Semantics of SUBSCRIBE_UPDATE

yekuiwang:
Fixes: https://github.com/moq-wg/moq-transport/issues/951, https://github.com/moq-wg/moq-transport/issues/955, https://github.com/moq-wg/moq-transport/issues/1127, https://github.com/moq-wg/moq-transport/issues/1128, https://github.com/moq-wg/moq-transport/issues/1129, https://github.com/moq-wg/moq-transport/issues/1133, https://github.com/moq-wg/moq-transport/issues/1134, https://github.com/moq-wg/moq-transport/issues/1139, https://github.com/moq-wg/moq-transport/issues/1140, https://github.com/moq-wg/moq-transport/issues/1141

draft-ietf-moq-transport.md
```diff
@@ -368,14 +368,15 @@ To give an example of how an application might use this data model,
...
-video on a given layer is sent as a single object.
+the display environment and available bandwidth. Each independently
+coded sequence of pictures in a resolution is sent as a group as the
+first picture in the sequence can be used as a random access point.
+This allows the client to join at the logical points where decoding
+of the stream can start without needing information before the join
+points. The temporal layers are sent as separate subgroups to allow
+the priority mechanism to favor lower temporal layers when there is
+not enough bandwidth to send all temporal layers. Each picture is sent
+as a single object.
```
```suggestion
the display environment and available bandwidth. Each independently
coded sequence of pictures in a resolution is sent as a group as the
first picture in the sequence can be used as a random access point.
This allows the client to join at the logical points where decoding
of the media can start without needing information before the join
points. The temporal layers are sent as separate subgroups to allow
the priority mechanism to favor lower temporal layers when there is
not enough bandwidth to send all temporal layers. Each frame of video
is sent as a single object.
```

draft-ietf-moq-transport.md
afrind: This section is about the object model, so I like focusing on "how the OP selects IDs" as that related to the model.  Maybe just Group IDs?  Or Group ID Ordering?

draft-ietf-moq-transport.md
afrind: I think "outside the narrowed range" is clear enough, and the brevity helps in understanding compared to "Objects with Locations smaller than the new Start Location/Objects with Locations larger than the new End Location"

gwendalsimon: Since there is an effort to be more rigorous in the writing, I would suggest:
- Remove the "However".  The sentence about race condition of a narrowed range is not related to the previous sentence by a However.
- Replace "won't" ==> "will not". Contraction must be avoided in written language
- The full sentence "However, ... narrowed range" is out of scope of this paragraph. The paragraph is about not widening the range. This remark about race condition in a narrow range is a side note on subscribe_update, it would require an extra paragraph if one thinks it is important to mention (I don't)
- The sentence "A publisher MUST terminate the session with a `PROTOCOL_VIOLATION` if the SUBSCRIBE_UPDATE violates **these** rules..." is hard to understand with respect to the previous sentences. One previous sentence says "Subscriptions can only be narrowed...", so is it a "can only" or a MUST? Is it one of the rules? One previous sentence is about race condition. Is it a rule? Where is the violation? Is there any other rule?

```suggestion
them. The Start Location MUST NOT decrease and the End Group MUST NOT increase.
A publisher MUST terminate the session with a `PROTOCOL_VIOLATION` if the
SUBSCRIBE_UPDATE violates these rules or if the subscriber specifies a request
ID that has not existed within the Session.

When a subscriber narrows their subscription, it might still receive objects
outside the new range if the publisher sent them before the update was
processed.
```

draft-ietf-moq-transport.md
suhasHere: ah makes sense. Then can we just called it Group IDs ?

draft-ietf-moq-transport.md
```diff
@@ -1031,9 +1027,9 @@ request.  It is specified in the SUBSCRIBE or FETCH message, and can be
 updated via SUBSCRIBE_UPDATE message.  The subscriber priority of an individual
 schedulable object is the subscriber priority of the request that caused that
 object to be sent. When subscriber priority is changed, a best effort SHOULD be
-made to apply the change to all objects that have not been sent, but it is
+made to apply the change to all objects that have not been scheduled, but it is
```
suhasHere, line 1030: sent is fine or "scheduled to be sent" is fine too.

suhasHere (commented):
minor nits, looks fine

suhasHere (approved):

draft-ietf-moq-transport.md
```suggestion
### Group IDs
```

draft-ietf-moq-transport.md
```diff
@@ -3627,7 +3625,7 @@ Objects in Group 10, as it sees fit.  A Track is considered malformed (see
 This extension is optional, as publishers might not know the prior gap gize, or
 there may not be a gap. If Prior Group ID Gap is not present, the receiver
 cannot infer any information about the existence of prior groups (see
-{{group-ordering}}).
+{{group-id-allocation}}).
```
```suggestion
{{group-ids}}).
```

draft-ietf-moq-transport.md
afrind, line 1030: @suhasHere I'm ok with the proposed change here.  An object is either unscheduled, scheduled or sent (leaving out acknowledged, since subgroup objects can rxmit).  When priorities update, all unscheduled objects have the new priority - but SHOULD is fine since all of priorities is SHOULD.  Objects that have been scheduled but not sent (eg perhaps buffered in the quic stack) are handled in an implementation dependent way.

Issue #1146, opened 2025-08-06
All Error message formats are the same


martinduke:
TRACK_STATUS was changed to be exactly like SUBSCRIBE, so the section for each message just refers to the original format.

However, all the REQUEST_ERROR messages are identical. They could all use the same language. Better yet, let's collapse it all into a single REQUEST_ERROR message and use the request ID to disambiguate what it means.

martinduke:
If we did this, I could eliminate a lot of code by having a function for general request correctness checking (is the request ID correct, did I send/receive GOAWAY, does the track exist etc) and just send REQUEST_ERROR if any of that fails.

afrind:
I hadn't thought about making a single REQUEST_ERROR message, but I had thought about sharing a parser/serializer for all error messages.

If we shared a single message, would you collapse the error code spaces (eg: < 100=common errors 100=subscribe errors, 200=publish errors, etc)?

martinduke:
Error codes don't really matter that much, but I'd have as many codes as possible be common. Certain numbers would mean different things for different types of request. This is pretty much how it is with the current numbering, and in fact I use a single set of constants for request error codes.

Issue #1036, opened 2025-05-29
Should we get rid of FETCH_ERROR + No Objects?


afrind:
The current text says both things -- a publisher can send FETCH_ERROR + No Objects, but FETCH_OK + an empty data stream is not forbidden, and as defined means the same thing.

I'm torn here - as a FETCH cache implementer, handling No Objects as a FETCH_ERROR is annoying.  However, forcing publishers to open and close an empty stream when they know there are no objects seems wasteful.

afrind:
Discussed in Madrid:

Cullen/Suhas in favor of removing FETCH_ERROR + No Objects, and Mo said to keep both paths.  No other opinions.

There doesn't seem to be clear consensus to make a change without additional input.  Would anyone else like to comment?

yekuiwang:
Having multiple ways of doing the same thing in a standard is always not a good thing, unless there is significant benefits of doing that.

Having multiple ways of doing the same thing in a standard, and then specifying one as an error and the other not as an error is kind of sick.

Therefore, I am for removing one of the two. As for which one to remove, no strong opinion, but I tend to think that it is more straightforward to keep "FETCH_ERROR + No Objects" and forbid FETCH_OK + an empty data stream, and this is also more consistent with other cases of "no content for your request" like `TRACK_DOES_NOT_EXIST`, `TRACK_ENDED` (or `DOES_NOT_EXIST` in PR #1159 ).

afrind:
> Having multiple ways of doing the same thing in a standard is always not a good thing, unless there is significant benefits of doing that.
> Having multiple ways of doing the same thing in a standard, and then specifying one as an error and the other not as an error is kind of sick.

This resonates with me :)

>  "FETCH_ERROR + No Objects" ... is more consistent with other cases of "no content for your request" ...

Looking at RFC 7233 (HTTP Range Requests), that also treats this condition as an error (416 Range Not Satisfiable).  This seems convincing to me...unless there's a case in which we want to send the FETCH_OK before we even know if there are objects or not?

> Cullen/Suhas in favor of removing FETCH_ERROR + No Objects, and Mo said to keep both paths. 

I'll note this suggestion (removing FETCH_OK+empty) is option C here, so not sure how close we are to achieving consensus.

martinduke:
I agree with @yekuiwang. Pick one. I'd poll the implementers and see what the majority of people do, because it really doesn't matter.

fluffy:
+1 yekuiwang

[...]

PR #1159, opened 2025-08-14
Consolidate all the Error Message types

### LINKED ISSUE #1156: PUBLISH_ERROR should use the Auth Error codes
PUBLISH accepts Auth tokens, so the errors apply.

### LINKED ISSUE #1157: Consolidate all the _ERROR messages into REQUEST_ERROR
They have the same format and same role in the state machine. Error codes are slightly different but mostly advisory.

Each request as a unique Request ID, so there is no ambiguity.

This eliminates quite a bit of tedious code and tedious spec text.

### LINKED ISSUE #1158: MALFORMED_TRACK should apply to SUBSCRIBE_ERROR and TRACK_STATUS_ERROR
Not just FETCH_ERROR.

martinduke:
Fixes #1156
Fixes #1157 
Fixes #1158

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -859,9 +859,9 @@ processing rules described in {{relays-moq}}.
 A subscription can be initiated by either a publisher or a subscriber.  A
 publisher initiates a subscription to a track by sending the PUBLISH message.
 The subscriber either accepts or rejects the subscription using PUBLISH_OK or
-PUBLISH_ERROR.  A subscriber initiates a subscription to a track by sending the
+REQUEST_ERROR.  A subscriber initiates a subscription to a track by sending the
 SUBSCRIBE message.  The publisher either accepts or rejects the subscription
-using SUBSCRIBE_OK or SUBSCRIBE_ERROR.  Once either of these sequences is
+using SUBSCRIBE_OK or REQUEST_ERROR.  Once either of these sequences is
 successful, the subscription can be updated by the subscriber using
 SUBSCRIBE_UPDATE, terminated by the subscriber using UNSUBSCRIBE, or terminated
 by the publisher using PUBLISH_DONE.
@@ -881,19 +881,19 @@ content. The communication of intent and authorization can be accomplished by
 the client sending SUBSCRIBE_NAMESPACE, or conveyed in other mechanisms out of
 band.
 
-A publisher MUST send exactly one SUBSCRIBE_OK or SUBSCRIBE_ERROR in response to
-a SUBSCRIBE. It MUST send exactly one FETCH_OK or FETCH_ERROR in response to a
-FETCH. A subscriber MUST send exactly one PUBLISH_OK or PUBLISH_ERROR in
+A publisher MUST send exactly one SUBSCRIBE_OK or REQUEST_ERROR in response to
+a SUBSCRIBE. It MUST send exactly one FETCH_OK or REQUEST_ERROR in response to a
+FETCH. A subscriber MUST send exactly one PUBLISH_OK or REQUEST_ERROR in
 response to a PUBLISH. The peer SHOULD close the session with a protocol error
 if it receives more than one.
 
 A subscriber keeps subscription state until it sends UNSUBSCRIBE, or after
-receipt of a PUBLISH_DONE or SUBSCRIBE_ERROR. Note that PUBLISH_DONE does not
+receipt of a PUBLISH_DONE or REQUEST_ERROR. Note that PUBLISH_DONE does not
 usually indicate that state can immediately be destroyed, see
 {{message-publish-done}}.
 
 A subscriber keeps FETCH state until it sends FETCH_CANCEL, receives
-FETCH_ERROR, or receives a FIN or RESET_STREAM for the FETCH data stream. If the
+REQUEST_ERROR, or receives a FIN or RESET_STREAM for the FETCH data stream. If the
 data stream is already open, it MAY send STOP_SENDING for the data stream along
 with FETCH_CANCEL, but MUST send FETCH_CANCEL.
 
@@ -906,11 +906,11 @@ The publisher can immediately delete subscription state after sending
 PUBLISH_DONE, but MUST NOT send it until it has closed all related streams. It
 can destroy all FETCH state after closing the data stream.
 
-A SUBSCRIBE_ERROR indicates no objects will be delivered, and both endpoints can
+A REQUEST_ERROR indicates no objects will be delivered, and both endpoints can
 immediately destroy relevant state. Objects MUST NOT be sent for requests that
 end with an error.
 
-A FETCH_ERROR indicates that both endpoints can immediately destroy state.
+A REQUEST_ERROR indicates that both endpoints can immediately destroy state.
 Since a relay can start delivering FETCH Objects from cache before determining
 the result of the request, some Objects could be received even if the FETCH
 results in error.
@@ -942,11 +942,11 @@ PUBLISH_NAMESPACE_DONE or PUBLISH messages for that namespace, or more specific
 part of that namespace.
 
 A publisher MUST send exactly one SUBSCRIBE_NAMESPACE_OK or
-SUBSCRIBE_NAMESPACE_ERROR in response to a SUBSCRIBE_NAMESPACE. The subscriber
+REQUEST_ERROR in response to a SUBSCRIBE_NAMESPACE. The subscriber
 SHOULD close the session with a protocol error if it detects receiving more than
 one.
 
-The receiver of a SUBSCRIBE_NAMESPACE_OK or SUBSCRIBE_NAMESPACE_ERROR ought to
+The receiver of a SUBSCRIBE_NAMESPACE_OK or REQUEST_ERROR ought to
 forward the result to the application, so the application can decide which other
 publishers to contact, if any.
 
@@ -975,10 +975,10 @@ PUBLISH_NAMESPACE (i.e. a PUBLISH_NAMESPACE ought not to be echoed back to its
 sender).
 
 An endpoint SHOULD report the reception of a PUBLISH_NAMESPACE_OK or
-PUBLISH_NAMESPACE_ERROR to the application to inform the search for additional
+REQUEST_ERROR to the application to inform the search for additional
 subscribers for a namespace, or to abandon the attempt to publish under this
 namespace. This might be especially useful in upload or chat applications. A
-subscriber MUST send exactly one PUBLISH_NAMESPACE_OK or PUBLISH_NAMESPACE_ERROR
+subscriber MUST send exactly one PUBLISH_NAMESPACE_OK or REQUEST_ERROR
 in response to a PUBLISH_NAMESPACE. The publisher SHOULD close the session with
 a protocol error if it receives more than one.
 
@@ -1352,11 +1352,11 @@ The following Message Types are defined:
 |-------|-----------------------------------------------------|
 | 0x1A  | REQUESTS_BLOCKED ({{message-requests-blocked}})     |
 |-------|-----------------------------------------------------|
+| 0x2   | REQUEST_ERROR  ({{message-request-ok}})             |
+|-------|-----------------------------------------------------|
 | 0x3   | SUBSCRIBE ({{message-subscribe-req}})               |
 |-------|-----------------------------------------------------|
 | 0x4   | SUBSCRIBE_OK ({{message-subscribe-ok}})             |
-|-------|-----------------------------------------------------|
-| 0x5   | SUBSCRIBE_ERROR ({{message-subscribe-error}})       |
 |-------|-----------------------------------------------------|
 | 0x2   | SUBSCRIBE_UPDATE ({{message-subscribe-update}})     |
 |-------|-----------------------------------------------------|
@@ -1368,13 +1368,9 @@ The following Message Types are defined:
 |-------|-----------------------------------------------------|
 | 0x1E  | PUBLISH_OK ({{message-publish-ok}})                 |
 |-------|-----------------------------------------------------|
-| 0x1F  | PUBLISH_ERROR ({{message-publish-error}})           |
-|-------|-----------------------------------------------------|
 | 0x16  | FETCH ({{message-fetch}})                           |
 |-------|-----------------------------------------------------|
 | 0x18  | FETCH_OK ({{message-fetch-ok}})                     |
-|-------|-----------------------------------------------------|
-| 0x19  | FETCH_ERROR ({{message-fetch-error}})               |
 |-------|-----------------------------------------------------|
 | 0x17  | FETCH_CANCEL ({{message-fetch-cancel}})             |
 |-------|-----------------------------------------------------|
@@ -1382,13 +1378,9 @@ The following Message Types are defined:
 |-------|-----------------------------------------------------|
 | 0xE   | TRACK_STATUS_OK ({{message-track-status-ok}}        |
 |-------|-----------------------------------------------------|
-| 0xF   | TRACK_STATUS_ERROR ({{message-track-status-error}}) |
-|-------|-----------------------------------------------------|
 | 0x6   | PUBLISH_NAMESPACE  ({{message-pub-ns}})             |
 |-------|-----------------------------------------------------|
 | 0x7   | PUBLISH_NAMESPACE_OK ({{message-pub-ns-ok}})        |
-|-------|-----------------------------------------------------|
-| 0x8   | PUBLISH_NAMESPACE_ERROR ({{message-pub-ns-error}})  |
 |-------|-----------------------------------------------------|
 | 0x9   | PUBLISH_NAMESPACE_DONE  ({{message-pub-ns-done}})   |
 |-------|-----------------------------------------------------|
@@ -1397,8 +1389,6 @@ The following Message Types are defined:
 | 0x11  | SUBSCRIBE_NAMESPACE ({{message-subscribe-ns}})      |
 |-------|-----------------------------------------------------|
 | 0x12  | SUBSCRIBE_NAMESPACE_OK ({{message-sub-ns-ok}})      |
-|-------|-----------------------------------------------------|
-| 0x13  | SUBSCRIBE_NAMESPACE_ERROR ({{message-sub-ns-error}} |
 |-------|-----------------------------------------------------|
 | 0x14  | UNSUBSCRIBE_NAMESPACE ({{message-unsub-ns}})        |
 |-------|-----------------------------------------------------|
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1369,12 +1369,12 @@ The following Message Types are defined:
 |-------|-----------------------------------------------------|
 | 0x1A  | REQUESTS_BLOCKED ({{message-requests-blocked}})     |
 |-------|-----------------------------------------------------|
+| 0x2   | REQUEST_ERROR  ({{message-request-ok}})             |
```
afrind: 0x2 is SUBSCRIBE_UPDATE

Please pick a new unused codepoint.

draft-ietf-moq-transport.md
```diff
@@ -3800,18 +3579,24 @@ TODO: register the URI scheme and the ALPN and grease the Extension types
...
+| DOES_NOT_EXIST             | 0x4  | SUBSCRIBE, FETCH, TRACK_STATUS, SUBSCRIBE_NAMESPACE | {{message-request-error}} |
+| UINTERESTED                | 0x4  | PUBLISH, PUBLISH_NAMESPACE | {{message-request-error}} |
+| INVALID_RANGE              | 0x5  | SUBSCRIBE, FETCH | {{message-request-error}} |
+| PREFIX_OVERLAP             | 0x5  | SUBSCRIBE_NAMESPACE | {{message-request-error}} |
+| NO_OBJECTS                 | 0x6  | FETCH          | {{message-request-error}} |
+| INVALID_JOINING_REQUEST_ID | 0x7  | FETCH          | {{message-request-error}} |
+| UNKNOWN_STATUS_IN_RANGE    | 0x8  | FETCH          | {{message-request-error}} |
+| MALFORMED_TRACK            | 0x9  | SUBSCRIBE, FETCH, TRACK_STATUS | {{message-request-error}} |
+| MALFORMED_AUTH_TOKEN       | 0x10 | All            | {{message-request-error}} |
+| EXPIRED_AUTH_TOKEN         | 0x12 | All            | {{message-request-error}} |
```
afrind: I think we could do something interesting with the ranges here, all the common errors in the 0-X range, generic subscriber, generic publisher, then message specific error ranges, etc.

martinduke:
Fixes #1146

draft-ietf-moq-transport.md
martinduke: Agreed. I don't wanna do it in this PR.

afrind:
Do we want to include the control message type in REQUEST_ERROR as an additional check that the responder is using the correct request id?

martinduke:
I don't think so. If the receiver can't keep request_id straight this is not the only problem there will be.

draft-ietf-moq-transport.md
```diff
@@ -1856,6 +1846,82 @@ REQUESTS_BLOCKED Message {
...
+
+* Error Reason: Provides a text description of the request error. See
+ {{reason-phrase}}.
+
+The application SHOULD use a relevant error code in REQUEST_ERROR,
+as defined below. Most codepoints have identical meanings for various request
+types, but some have request-specific meanings.
+
+INTERNAL_ERROR (0x0):
+: An implementation specific or generic error occurred.
```
fluffy, line 1990: I'd like to remove this error. It it completely unspecified what the device receiving this error does when it gets it.

draft-ietf-moq-transport.md
```diff
@@ -1856,6 +1846,82 @@ REQUESTS_BLOCKED Message {
...
+UNAUTHORIZED (0x1):
+: The subscriber is not authorized to subscribe to the given track.
+
+TIMEOUT (0x2):
+: The subscription could not be completed before an implementation specific
+  timeout. For example, a relay could not establish an upstream subscription
+  within the timeout.
+
+NOT_SUPPORTED (0x3):
+: The endpoint does not support the type of request.
```
fluffy, line 2002: I think we should remove this and have better extensibility approach for deciding what is supported at setup time.

draft-ietf-moq-transport.md
```diff
@@ -1856,6 +1846,82 @@ REQUESTS_BLOCKED Message {
...
+DOES_NOT_EXIST or UNINTERESTED (0x4):
+: In response to SUBSCRIBE, FETCH, TRACK_STATUS, or SUBSCRIBE_NAMESPACE, the
+track or namespace is not available at the publisher. For PUBLISH or
+PUBLISH_NAMESPACE, the subscriber is not interested in the track or namespace.
+
+INVALID_RANGE or PREFIX_OVERLAP (0x5):
+: In response to SUBSCRIBE or FETCH, specified range of Locations cannot be
+accessed with the request type. In response to SUBSCRIBE_NAMESPACE, the
+namespace prefix overlaps with another SUBSCRIBE_NAMESPACE in the same
+session.
```
fluffy: This should be two different errors

draft-ietf-moq-transport.md
```diff
@@ -1856,6 +1846,82 @@ REQUESTS_BLOCKED Message {
...
+
+NO_OBJECTS (0x6):
+: In response to FETCH, no objects exist between the Start and End Locations.
```
fluffy: This is not really an error and I thought we decided to just sent a fetch response with no data in it.

draft-ietf-moq-transport.md
```diff
@@ -1856,6 +1846,82 @@ REQUESTS_BLOCKED Message {
...
+: In response to a FETCH, the requested range contains an object with unknown
+status.
+
+MALFORMED_TRACK (0x9):
+: In response to a SUBSCRIBE, FETCH, or TRACK_STATUS, a relay publisher detected
+the track was malformed (see {{malformed-tracks}}).
+
+MALFORMED_AUTH_TOKEN (0x10):
+: Invalid Auth Token serialization during registration (see
+  {{authorization-token}}).
```
fluffy, line 2006: Seems like we need both a Malformed and Invalid auth token error. Name does not match description.

draft-ietf-moq-transport.md
```diff
@@ -1856,6 +1846,82 @@ REQUESTS_BLOCKED Message {
...
+
+EXPIRED_AUTH_TOKEN (0x12):
+: Authorization token has expired ({{authorization-token}}).
```
fluffy, line 2009: We probably need a bunch more text around when this can be sent.

draft-ietf-moq-transport.md
```diff
@@ -3800,18 +3579,24 @@ TODO: register the URI scheme and the ALPN and grease the Extension types
...
-| UNAUTHORIZED          | 0x1  | {{message-subscribe-error}} |
-| TIMEOUT               | 0x2  | {{message-subscribe-error}} |
-| NOT_SUPPORTED         | 0x3  | {{message-subscribe-error}} |
-| TRACK_DOES_NOT_EXIST  | 0x4  | {{message-subscribe-error}} |
-| INVALID_RANGE         | 0x5  | {{message-subscribe-error}} |
-| MALFORMED_AUTH_TOKEN  | 0x10 | {{message-subscribe-error}} |
-| EXPIRED_AUTH_TOKEN    | 0x12 | {{message-subscribe-error}} |
+### REQUEST_ERROR Codes {#iana-request-error}
+
+| Name                       | Code | Request Types  |Specification              |
```
fluffy: I would arrange this table differently. The problem is this is going to be hard to add new messages new because a new command would make changes all over this table. Here is my proposal. We have this table but we remove from in the "Request Type" column. Then we have a set of extra tables. We have a table for each request type that lists the Names of errors that are allowed in responses to that message. 
 
 New requests will define a new table of the second type, and if they add a new error type, append it to the end of the first.

draft-ietf-moq-transport.md
fluffy: I view HTTP errors as a strong anti pattern in how to do this. I don't think class of errors ever helped even in HTTP. Error are only really usefull if the receiver can do something when they receive them.

fluffy (changes requested):
Thanks for starting on cleaning up the error stuff.

draft-ietf-moq-transport.md
afrind: This PR is only about consolidating the messages, so we won't be changing any error codes or semantics here.  I will file an issue for you with your comments about existing error codes.

draft-ietf-moq-transport.md
afrind: This has already been discussed and resolved in the editors' copy, see #1107 (which you approved) and #908.

draft-ietf-moq-transport.md
afrind: I agree, I don't think we want any codes having two meanings.  It sounds like @martinduke wanted to make these changes in two PRs.

draft-ietf-moq-transport.md
afrind: See https://github.com/moq-wg/moq-transport/issues/1036#issuecomment-3113298608.  We discussed in Madrid but got very little and conflicting input.

draft-ietf-moq-transport.md
afrind: We have UNAUTHORIZED (well formed) and MALFORMED.  Can you describe how INVALID is different from either of these two?

draft-ietf-moq-transport.md
afrind: We have some discussion in #1026

draft-ietf-moq-transport.md
afrind: > Error are only really usefull if the receiver can do something when they receive them.

Strong disagree.  In practice, errors also feed monitoring systems that have alarms and operators behind them.  Even if every error had the same client action "give up" or "retry", being able to group these by the actual problem has a lot of value.

> I don't think class of errors ever helped even in HTTP.

As someone who operates an HTTP CDN at scale, I will say that error classes are very useful, for the above reason, and you can look at any HTTP implementation and I virtually guarantee you will find code that divides status by 100 and looks at the result.

That said,my suggestion to band the error codes had less to do with the way that HTTP does it (by function), but instead just grouping them by verb (subscribe specific, publish specific, etc).

draft-ietf-moq-transport.md
afrind: +1 this will get very messy.

draft-ietf-moq-transport.md
afrind: #1162

martinduke:
If I can synthesize this thread:

1) Do not have a single Codepoint mean different things for different messages. I'll renumber these to give ranges to particular request types.

2) There is interest in changing some of these errors, but that will be tracked in other errors/PRs.

I can live with that, and will update accordingly.

draft-ietf-moq-transport.md
martinduke: done, though I didn't bother to create separate tables

afrind (approved):

ianswett (approved):
Nice!

draft-ietf-moq-transport.md
```diff
@@ -1862,6 +1852,93 @@ REQUESTS_BLOCKED Message {
...
+
+DOES_NOT_EXIST (0x10):
+: The track or namespace is not available at the publisher.
+
+INVALID_RANGE (0x11):
+: In response to SUBSCRIBE or FETCH, specified range of Locations cannot be
+accessed with the request type.
+
+MALFORMED_TRACK (0x12):
+: In response to a SUBSCRIBE, FETCH, or TRACK_STATUS, a relay publisher detected
```
afrind: I repurposed #1158 to discuss the cacheability of malformed track which impacts whether subscribe or track status can result in a malformed track.  I recommend keeping this constrained to FETCH until that resolves.
```suggestion
: In response to a FETCH, a relay publisher detected
```

draft-ietf-moq-transport.md
fluffy: This is not a great way to write spec. Justin Uberti was the person that really convinced me a better way to write specs was not just rearrange the furniture in a dirty room, but instead, if you are making changes in the room, clean up the room.

afrind: @fluffy I understand the sentiment, but I disagree that _this_ PR should be blocked on resolving every individual error code semantic. That is completely orthogonal to what is happening here.

gloinul: @fluffy I will note that the Internal Error code exists for the non-consoldidated Error responses in -13 which has WG consensus. Thus, it is a separate proposal to remove the Internal Error code for which there exists an issue #1162. I don't see you (Cullen) having a comment here that is against consolidating the Error responses for all requests?

gloinul:
As WG Chair.

> I have no idea how to implement receiving an error such as INTERNAL_ERROR. If a relay sends an upstream subscribe and gets that, what does it send to the downstream subscribe.

@fluffy there exists an issue #1162 for that aspect. I also reviewed your other comments that appeared to also be issues with specific error codes. What I can see these error codes do exists in draft -13 which we have run the WG consensus on. Thus changing the errors that are defined is a change to what we have WG consensus on. Which we clearly can given the WG process for raising issues. 

I don't see any comment from you that appears to be against the consolidation of multiple request specific error responses into a single one. Thus, can you remove your request for change here and keep this PR focused on only the consolidation?

afrind:

There were no objections to moving forward with this PR as is and addressing error code semantics separately.

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1862,6 +1852,93 @@ REQUESTS_BLOCKED Message {
...
+
+The application SHOULD use a relevant error code in REQUEST_ERROR,
+as defined below. Most codepoints have identical meanings for various request
+types, but some have request-specific meanings.
+
+INTERNAL_ERROR (0x0):
+: An implementation specific or generic error occurred.
+
+UNAUTHORIZED (0x1):
+: The subscriber is not authorized to subscribe to the given track.
```
```suggestion
: The subscriber is not authorized to perform the requested action on the given
track.
```

draft-ietf-moq-transport.md
```diff
@@ -1862,6 +1852,93 @@ REQUESTS_BLOCKED Message {
...
+
+TIMEOUT (0x2):
+: The subscription could not be completed before an implementation specific
+  timeout. For example, a relay could not establish an upstream subscription
+  within the timeout.
+
+NOT_SUPPORTED (0x3):
+: The endpoint does not support the type of request.
```
suhasHere, line 2002:
```suggestion
: The endpoint processing the request does not support the type of request.
```

draft-ietf-moq-transport.md
```diff
@@ -1862,6 +1852,93 @@ REQUESTS_BLOCKED Message {
...
+
+MALFORMED_AUTH_TOKEN (0x4):
+: Invalid Auth Token serialization during registration (see
+  {{authorization-token}}).
+
+EXPIRED_AUTH_TOKEN (0x5):
+: Authorization token has expired ({{authorization-token}}).
+
+Below are errors for use by the publisher. They can appear in response to
+SUBSCRIBE, FETCH, TRACK_STATUS, and SUBSCRIBE_NAMESPACE, unless otherwise noted.
```
suhasHere, line 2012: what is the expected behavior, if these exist elsewhere ?

draft-ietf-moq-transport.md
```diff
@@ -1862,6 +1852,93 @@ REQUESTS_BLOCKED Message {
...
+
+DOES_NOT_EXIST (0x10):
+: The track or namespace is not available at the publisher.
+
+INVALID_RANGE (0x11):
+: In response to SUBSCRIBE or FETCH, specified range of Locations cannot be
+accessed with the request type.
```
```suggestion
: In response to SUBSCRIBE or FETCH, specified Filter or range of Locations
cannot be satisfied.
```

draft-ietf-moq-transport.md
```diff
@@ -1862,6 +1852,93 @@ REQUESTS_BLOCKED Message {
...
+
+MALFORMED_TRACK (0x12):
+: In response to a SUBSCRIBE, FETCH, or TRACK_STATUS, a relay publisher detected
+the track was malformed (see {{malformed-tracks}}).
+
+The following are errors for use by the subscriber. They can appear in response
+to PUBLISH or PUBLISH_NAMESPACE, unless otherwise noted.
+
+UNINTERESTED (0x20):
+: The subscriber is not interested in the track or namespace.
```
suhasHere, line 2029: Under what conditions, a publisher chooses to use this error. This error seems to be not much helpful, unless I am missing something

draft-ietf-moq-transport.md
```diff
@@ -2044,61 +2121,6 @@ session with a `PROTOCOL_VIOLATION` ({{session-termination}}).
...
-: The subscription could not be completed before an implementation specific
-  timeout. For example, a relay could not establish an upstream subscription
-  within the timeout.
-
-NOT_SUPPORTED (0x3):
-: The endpoint does not support the SUBSCRIBE method.
-
-TRACK_DOES_NOT_EXIST (0x4):
-: The requested track is not available at the publisher.
-
```
suhasHere, line 2136: We are missing this error. This seems to be useful one ?

draft-ietf-moq-transport.md
```diff
@@ -2646,75 +2626,6 @@ Values of 0x0 and those larger than 0x2 are a protocol error.
...
-UNAUTHORIZED (0x1):
-: The subscriber is not authorized to fetch from the given track.
-
-TIMEOUT (0x2):
-: The fetch could not be completed before an implementation specific timeout.
-  For example, a relay could not FETCH missing objects within the timeout.
-
-NOT_SUPPORTED (0x3):
-: The endpoint does not support the FETCH method.
```
suhasHere, line 2750: We are missing this error too

draft-ietf-moq-transport.md
```diff
@@ -2646,75 +2626,6 @@ Values of 0x0 and those larger than 0x2 are a protocol error.
...
-
-TRACK_DOES_NOT_EXIST (0x4):
-: The requested track is not available at the publisher.
```
suhasHere, line 2753: This one is missing too

draft-ietf-moq-transport.md
```diff
@@ -2985,63 +2832,6 @@ Message"}
...
-
-TIMEOUT (0x2):
-: The operation could not be completed before an implementation specific
-  timeout.
-
-NOT_SUPPORTED (0x3):
-: The endpoint does not support the SUBSCRIBE_NAMESPACE method.
-
-NAMESPACE_PREFIX_UNKNOWN (0x4):
-: The namespace prefix is not available for subscription.
```
suhasHere, line 3096: we are missing these error codes .

draft-ietf-moq-transport.md
```diff
@@ -2985,63 +2832,6 @@ Message"}
...
-
-NAMESPACE_PREFIX_OVERLAP (0x5):
-: The namespace prefix overlaps with another SUBSCRIBE_NAMESPACE in the same
-  session.
```
suhasHere, line 3100: same as above .

draft-ietf-moq-transport.md
afrind: For that, we need to resolve #1175 - I expect they would be treated the same way?

draft-ietf-moq-transport.md
afrind: This was already discussed in #704.

draft-ietf-moq-transport.md
afrind: See line 1895

draft-ietf-moq-transport.md
afrind: Now called `DOES_NOT_EXIST`, see line 1908

draft-ietf-moq-transport.md
afrind: Now called `DOES_NOT_EXIST`; see line 1908.

draft-ietf-moq-transport.md
afrind: Line 1927

draft-ietf-moq-transport.md
afrind: I believe this PR is consolidating this error with `DOES_NOT_EXIST`, is that right @martinduke?

draft-ietf-moq-transport.md
afrind: Isn't "processing the request" implied since it's in the error message?  Which other endpoint would it be?

afrind:
> Thanks for the PR. The consolidation seems to be good. I noticed some of the error conditions are missing and they seems to serve a distinct purpose. Love to see them added before we merge.

I think all of the errors are still there.  DOES_NOT_EXIST consolidates a couple different cases now, but is as unambiguous as the previous set based on the type of request.

Issue #441, opened 2024-04-24
Ability to filter which Objects are sent in a Subscription


wilaw:
Imagine a track, sent as frame-per stream, with an I-frame every 2s and each GOP in a Group. 
In order to create a trick-play preview of this track, it would be convenient to just receive the I-frames, i.e object 0 from each group.

To do this , we should create the ability in specify an OBJECT-FILTER in our SUBSCRIBE. The object filter can have a simply syntax to specify a range of object numbers that should be delivered. For example: 

```
[0] - only send objects with an ID of 0
[0,5] - only send objects with an ID of 0 or 5.
[0..5] - only send objects in the range 0 to 5 inclusive.
[0..2, 16..20] - only send objects in the range 0 to 5 inclusive, or 16..20 inclusive.
```

We could optionally extend this concept to group numbers too. I don't have an immediate use case for that, but I could imagine one involving future layered encodings in the same track, where I wanted to pull multiplexed ranges of groups.

VMatrix1900:
The concept of filter is useful. By putting it in the subscribe message, each individual client is able to create a shadow sub-track which is tailored to its own needs. I wonder if the similar result can be accomplished by server creating multiple tracks for all clients to subscribe.

ianswett:
There was interest in this in Boston, but it makes sense to wait until Subscribe, Fetch, etc settle out.

afrind:
I wonder if this is a good candidate for testing the extensibility of filter types per #790 ?

afrind:
See also #1007 for BPF suggestion.

martinduke:
In Montreal: Ted Hardie points out that there are privacy considerations for relays acting on feedback tracks. Should at least write about them.

afrind:
Closed by #1765

Issue #1068, opened 2025-07-08
Outside-in compression: scaling MOQT for high capacity ingest


wilaw:
Forgive the Pied Piper Reference. The purpose of this issue is to initiate a design discussion around making MOQT efficient for many-to-one distribution. 

Much of the preceding design of MOQT assumes either few-to-few (e.g. video conferencing) or one-to-many (live sports) distribution. We can leverage caching to enable scalable egress, combined with a fan-out architecture for non-cacheable real-time content. 

But what about the use-case of "fan-in" - millions of sensors providing real-time telemetry to a single monitoring point? In this case, caching does no good from a scale perspective, as the content is only ever going to one subscriber. The fan-out architecture works in reverse to collapse all upstream data onto downstream pipes. 

What features do we need in MOQT to make high capacity ingest efficient?:
1. We want a way for a publisher to signal to all relays that a track should not be cached. Currently the MAX CACHE DURATION Parameter can only be present in a SUBSCRIBE_OK, FETCH_OK or TRACK_STATUS message. It cannot be sent by a publisher. Setting the DELIVERY TIMEOUT Parameter to zero is not equivalent to do-not-cache. 
2. Keep-alives for long running sessions. Imagine a security camera that sends sparse data once every 5min. We can rely on application level pings to keep the connection alive, or should web build in a mechanism into the protocol to keep connections alive (with auth protection so untrusted clients cannot keep connections open forever). 
3. Can the central collectors use the relays as configurable filters? What if we do want to leverage the cache to "batch" the collection of telemetry? For example, sensors output temperature data every second, but the collector only wants to retrieve the last 5mi of data every 5min. We could create a FORWARD_FILTER Parameter which the subscriber can send during subscription, which specifies a key parameter and a threshold value. For example, I subscribe only to objects for which the "foo" parameter is > 10. 
4. Much telemetry data is timeline-based. Can we have a standard mode in which a track is identified as a timeline and make sure that works efficiently for our major actions (SUBSCRIBE, FETCH etc). Is setting each group number to be a millisecond epoch timestamp sufficiently robust? 
5. Publisher authentication and authorization: is our current token scheme sufficient for high-capacity ingest?
6. End-to-end encryption: we allow the payload to be encrypted, but is this sufficient for sensitive IoT data at city-wide scale?

What else might be important for high-capacity ingest that we can address now before we fail at scale? 

Cheers
Will

LPardue:
[...]

PR #1164, opened 2025-08-20
RFC: Generic Filters for Subscriptions and FETCH

afrind:
Based on proposal in #1068

Fixes: #441 

One thing that is strange is allowing Group ID filters here, since these overlap with the built-in filters in Subscribe and Fetch.  This may be what Victor was suggesting in Stockholm.

The pathological case of this encoding is for every other object, which would double the length of the Values array.  We could fix it by shifting the start value left by one and using the LSB to indicate single vs range, but it seemed like overkill.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1665,8 +1665,8 @@ Extension ID is present only when Operand is Extension ID (value 0x4) and
 indicates the Extension to filter. This MUST indicate an Extension with an
 integer type.  Objects without the Extension do not pass the filter.
 
-Values is an array of integers which encode the values of interest.  The
-array is a sequence of pairs indicating the start and length of
+Values is an array of variable-length integers (i) which encode the values of
+interest.  The array is a sequence of pairs indicating the start and length of
 the matching range.  The Start is encoded as a delta from the previous End, or
 from 0 for the first element.  The length indicates the number of elements
 including Start to match.  An odd number of elements indicates the final range


draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,58 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+Negated = Operand and Flag & 0x1
+Operand = Operand and Flag >> 1
+~~~
+
+Operand | Meaning
+0x0 | No Filter
+0x1 | Group ID
+0x2 | Subgroup ID
+0x3 | Object ID
+0x4 | Extension ID
```
afrind, line 1658:
```suggestion
0x4 | Extension ID
0x5 | Publisher Priority
```

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,59 @@ multi-object stream will expire earlier than Objects later in the stream. Once
 Objects have expired from cache, their state becomes unknown, and a relay that
 handles a downstream request that includes those Objects re-requests them.
 
+#### OBJECT FILTER Parameter
+
+The OBJECT_FILTER parameter(Parameter Type 0x05) MAY appear in SUBSCRIBE,
+SUBSCRIBE_UPDATE, PUBLISH_OK, TRACK_STATUS or FETCH message.  It is a structure
```
suhasHere, line 1639: how about adding this to SubscribeNamespace as well ?

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,59 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+indicating Objects or ranges of Objects that the Publisher will deliver.
+
+~~~
+Object Filter {
+  Type (0x5),
+  Length (i),
+  Operand and Flag (i),
```
```suggestion
  Operand and Operator (i),
```

suhasHere: Where Operator can be one !-, =, >, < . I think this will cover a good set of use-cases.

draft-ietf-moq-transport.md
afrind: It's unclear what adding a filter to SUBSCRIBE_NAMESPACE would do without #1047?  Or you have different use cases in mind?  Let's keep this focused on subscriptions.

draft-ietf-moq-transport.md
afrind: All of those can be express via range filters + negation:

```
=  -> Equal
!= -> Negate + Equal
>  -> Range with no end
<  -> Negate + Range with no end
```

It's also unclear what < > mean with multiple ranges.

fluffy: nit but instead of messing around with the explaining to people how to extract bits, just use bitfield definitions so make this 

Operand(4) 
NegateFlag (1)

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,59 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+match.  A length of 0 indicates no endpoint.  For example [ 10, 3, 1, 1, 20, 0 ]
+would match values 10, 11, 12, 14 and 34-max.  Only the first (excluding
+Extension ID) or last value can be 0.  If an endpoint receives a 0 value
+anywhere else, it MUST treat the track as Malformed (see {{malformed-tracks}}).
+
+All filters can be removed from a Subscription by sending SUBSCRIBE_UPDATE with
+Operand No Filter.
+
+This parameter MAY appear more than once, and filters are cumulative.  When
+filters are applied, delivery rules regarding Subgroups and FETCH responses are
```
fluffy: I don't want to deal with an unlimited number of filters operations. How about a max of 4 unless someone can come up with a very good use case for more. Keep in mind I would like to be able to implement this on something like a FPGA or silicon for wire speed filtering.

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,59 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+modified.  Within a Subgroup, the Publisher is allowed to send an Object on a
+Subgroup stream even when it is not the next Object if the expected Objects
+did not pass the filter.  In a FETCH response, the Subscriber can only infer
```
fluffy: I might be reading what you are saying here wrong but ...

If you mean you can send stuff that does not pass the filter, then no, I don't think this should be allowed. I think the stream should be ended if this happens on reliable stream and the things should just be filtered if on datagram. 

If you mean that if the filter remove object N but object N+1 is the subgroup passes the filter, you can send N+1 even though N is missing ... then I need to think about what this does to caches downstream.

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,59 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+Objects do not exist from their absence if they pass all filters.
```
fluffy: Agree with this but Depending on the filter, that may be close to impossible so I guess fetch will have to use reasonable filters to be able to infer this.

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,59 @@ multi-object stream will expire earlier than Objects later in the stream. Once
 Objects have expired from cache, their state becomes unknown, and a relay that
 handles a downstream request that includes those Objects re-requests them.
 
+#### OBJECT FILTER Parameter
+
+The OBJECT_FILTER parameter(Parameter Type 0x05) MAY appear in SUBSCRIBE,
+SUBSCRIBE_UPDATE, PUBLISH_OK, TRACK_STATUS or FETCH message.  It is a structure
+indicating Objects or ranges of Objects that the Publisher will deliver.
```
fluffy, line 1640: Can it occur more than once ? If I want to filter on only object where the group_id is greater than 100 and publish priority is greater than 10, do I use two ?

afrind: Yes, you composite them.

draft-ietf-moq-transport.md
afrind: I realized < X is also trivially implemented as a range from [0,X]

draft-ietf-moq-transport.md
afrind: A max seems very sensible.

draft-ietf-moq-transport.md
afrind: I am not saying you should send stuff that doesn't pass the filter. I was thinking the latter.  For example f you have a subgroup with objects 0-9, and a filter that says only even objects, then you can put [ 0, 2, 4, 5, 8 ] in a single stream, even though you know you are omitting objects.

That said, I think maybe that is a really bad idea, and we should just use the normal logic and the publisher has to reset the stream and open a new one wherever it skips.

draft-ietf-moq-transport.md
afrind: Yes, I realized this too -- eg if I filter on Object ID or Group ID, this makes sense, but for other filter types, the subscriber would never know that Object X was not sent because it had the wrong extension value or it just didn't exist.

afrind (commented):
> I can see how this might make sense for fetch but it seems to really complicate subscribe and I don't see the use cases it solves. To be specific, I can not imagine a subscriber where I want groups 100 to 200, then skip group 201 to 299, then give me groups 300 to 400.

See the linked issue (and also #441) where @wilaw lists out some use cases.  You may be right about having a _list_ of allowed ranges though.

> This is a fairly significant breaking change and I think we should talk more about what we are trying to fix before getting into this. I do see some value in writing down a more generic and normalize filter that provides more capabilities as long as it does not complicate things so we may end up at exactly this but I'd like to consider the goals more fist.

Completely agree we need to spend time discussing this before moving forward (note "RFC" in PR title).  I wrote this to help kickstart the discussion, and because it's among the biggest "API level" changes remaining, which the chairs have asked us to prioritize.

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,59 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+~~~
+
+Operand and Flag contains both an enumeration indicating the Operand, with the
+least significant bit indicating if the filter is negated.
+
+~~~
+Negated = Operand and Flag & 0x1
+Operand = Operand and Flag >> 1
+~~~
+
```
```suggestion
~~~
Object Filter {
  Type (0x5),
  Length (i),
  Operand (7),
  Negation Flag (1),
  [Extension ID (i),
  Values (i) ...
}
~~~

```

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,59 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+Operand | Meaning
+0x0 | No Filter
+0x1 | Group ID
+0x2 | Subgroup ID
+0x3 | Object ID
+0x4 | Extension ID
+0x5 | Publisher Priority
+
```
afrind, line 1660:
```suggestion

When Negation Flag is 1, the result of the filter is negated to determine which Objects pass the filter.
```

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,59 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+Operand is Extension ID, the first Value indicates the Extension ID to filter.
+This value MUST indicate an Extension with an integer type.  The remainder of
+the array is a sequence of pairs indicating the start and length of the matching
+range.  The Start is encoded as a delta from the previous End, or from 0 for the
+first element.  The length indicates the number of elements including Start to
+match.  A length of 0 indicates no endpoint.  For example [ 10, 3, 1, 1, 20, 0 ]
+would match values 10, 11, 12, 14 and 34-max.  Only the first (excluding
+Extension ID) or last value can be 0.  If an endpoint receives a 0 value
+anywhere else, it MUST treat the track as Malformed (see {{malformed-tracks}}).
+
```
```suggestion
Extension ID is present only when Operated is Extension ID (value 0x4) and
indicates the Extension to filter. This MUST indicate an Extension with an
integer type.  Objects without the Extension do not pass the filter.

Values is an array of integers which encode the values of interest.  The
remainder of the array is a sequence of pairs indicating the start and length of
the matching range.  The Start is encoded as a delta from the previous End, or
from 0 for the first element.  The length indicates the number of elements
including Start to match.  An odd number of elements indicates the final range
has no endpoint.  For example [ 10, 3, 1, 1, 20 ] would match values 10, 11, 12,
14 and 34-max.  Only the first value can be 0.  If an endpoint receives a 0
value anywhere else or length of the Values array exceeds TBD, it MUST reject
the request with `INVALID_FILTER`.
```

draft-ietf-moq-transport.md
```suggestion
This parameter MAY appear up to TBD2 times in a request, and filters are
cumulative.  If the parameter appears more than TBD2 times, the endpoint MUST
reject the request with `INVALID_FILTER`. When filters are applied, delivery
rules regarding FETCH responses are modified (see {{message-fetch}}).  In a
filtered FETCH response, the Subscriber can only infer Objects do not exist from
their absence if they pass all filters.  This is only possible when filtering on
Group ID and/or Object ID, as the Subscriber lacks other fields needed to
evaluate the filters.

Publishers MUST respect Subgroup delivery rules even for filtered responses.  If
an Object in the middle of a Subgroup is filtered out, the Publisher MUST NOT
place another Object in that Subgroup stream (see {{closing-subgroup-streams}}).
```

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,62 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+Operand | Meaning
+0x0 | No Filter
+0x1 | Group ID
+0x2 | Subgroup ID
+0x3 | Object ID
+0x4 | Extension ID
+0x5 | Publisher Priority
+
+When Negation Flag is 1, the result of the filter is negated to determine which Objects pass the filter.
+Extension ID is present only when Operated is Extension ID (value 0x4) and
```
```suggestion
When Negation Flag is 1, the result of the filter is negated to determine which
Objects pass the filter.

Extension ID is present only when Operated is Extension ID (value 0x4) and
```

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,64 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+0x1 | Group ID
+0x2 | Subgroup ID
+0x3 | Object ID
+0x4 | Extension ID
+0x5 | Publisher Priority
+
+When Negation Flag is 1, the result of the filter is negated to determine which
+Objects pass the filter.
+
+Extension ID is present only when Operated is Extension ID (value 0x4) and
```
```suggestion
Extension ID is present only when Operand is Extension ID (value 0x4) and
```

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,64 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+remainder of the array is a sequence of pairs indicating the start and length of
+the matching range.  The Start is encoded as a delta from the previous End, or
+from 0 for the first element.  The length indicates the number of elements
+including Start to match.  An odd number of elements indicates the final range
+has no endpoint.  For example [ 10, 3, 1, 1, 20 ] would match values 10, 11, 12,
+14 and 34-max.  Only the first value can be 0.  If an endpoint receives a 0
+value anywhere else or length of the Values array exceeds TBD, it MUST reject
+the request with `INVALID_FILTER`.
+All filters can be removed from a Subscription by sending SUBSCRIBE_UPDATE with
+Operand No Filter.
```
afrind, line 1679:
```suggestion

All filters can be removed from a Subscription by sending SUBSCRIBE_UPDATE with
Operand No Filter.
```

draft-ietf-moq-transport.md
suhasHere: adding it to namespace will enable object filtering across the tracks, where the base case is a single subscription . I was suggesting it would be nice to think the filter more generically

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,65 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+Operand | Meaning
+0x0 | No Filter
+0x1 | Group ID
+0x2 | Subgroup ID
+0x3 | Object ID
+0x4 | Extension ID
+0x5 | Publisher Priority
+
+When Negation Flag is 1, the result of the filter is negated to determine which
+Objects pass the filter.
```
wilaw, line 1662: Should you define that Negation flag has only two allowed values - 0 and 1? Or is it acceptable to pass any value !1?

wilaw:
> I can see how this might make sense for fetch but it seems to really complicate subscribe and I don't see the use cases it solves. To be specific, I can not imagine a subscriber where I want groups 100 to 200, then skip group 201 to 299, then give me groups 300 to 400.

Sparse non-contiguous ranges are really a filter use-case for FETCH. Since we want a common filter parameter between FETCH and SUBSCRIBE, these just won't be used much in SUBSCRIBE, but will be used with FETCH. 

For SUBSCRIBE, the use-cases for filters outside of Sparse non-contiguous ranges are quite real and some examples are described below. Additionally, when combined with extensions, we get some very nice behaviors. 

1. Sensors send temperature summary data in the first Object of each Group, then updates every minute as a new object. I want to monitor only the summary data by receiving only the first object in every group
FILTER ( =, object ID, 0 )

1. A live video feed codes its groups using unix epoch time. I want to subscribe to the feed starting  tonight at 7pm 
FILTER ( >, group ID, 1752138462)

1. Only forward me live security camera groups where the activity level is greater than or equal to 25%. The activity level is added to each object as the custom extension with a type of 400.
FILTER (>=, extension 400, 25)

1. Only forward me live temperature data from zip code 94925. The zip code is stored as the custom extension header 1444.
FILTER (=, extension 1444, 94925)

1. Only forward me log data from the media player with subscriber ID 847362511. The subscriber ID is sent in extension header 768 by the media player.
FILTER (=, extension 768, 847362511)

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,65 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+
+Extension ID is present only when Operand is Extension ID (value 0x4) and
+indicates the Extension to filter. This MUST indicate an Extension with an
+integer type.  Objects without the Extension do not pass the filter.
+
+Values is an array of integers which encode the values of interest.  The
+remainder of the array is a sequence of pairs indicating the start and length of
+the matching range.  The Start is encoded as a delta from the previous End, or
```
wilaw: > The remainder of the array is a sequence of pairs indicating the start and length of
> the matching range

Saying that the second value indicates the length of the _matching range_ is contradictory to the examples, as it suggests the length within which samples will match. For example a start of 10 and a length of 3 indicates that the matching range is from 10..13 , whereas it is intended to be 10..12. 

I think explicitly defining the matching range is more intuitive for users than specifying a non-inclusive final value. For example, if I am filtering on zip code data, and I want 94552, i have to set the filter length to point at the imaginary 94553, which is not a valid zip code. 

So suggestion is modify the definition so that the length is an inclusive match. This changes the example given later and also relaxes the requirement around which values can be zero. To specify matches for 10, 11, 12, 14 and 34-max we would code [10, 2, 2, 0, 20]

wilaw (changes requested):
Thanks for implementing this Alan!

My main feedback is about making the length of the match inclusive rather than exclusive. When dealing with GroupIDs etc it doesn't matter much, but when we get into use-cases where the values represent real-world discrete things (like ZIP codes, device IDs, temperatures, serial numbers etc), then specifying the inclusive end of the range seems better than one more than that.

draft-ietf-moq-transport.md
afrind: @suhasHere I am amenable to using the same parameter to filter Tracks in response to SUBSCRIBE_NAMESPACE rather than Objects in SUBSCRIBE, but it requires more semantics (for example, most operands defined here don't make any sense).  Can you file an issue explaining the requirements for Track filtering first?

draft-ietf-moq-transport.md
afrind: It's only 1 bit now, so 0-1 are implicit

draft-ietf-moq-transport.md
afrind: API and wire format don't need to be the same. I figured most APIs would take in a list of absolute ranges and run through an encoder which produce the values on the wire -- especially since subsequent range offsets are delta encoded.  It's also a little strange to say it's a sequence of [Delta, Length-1] pairs, though it does buy one more value for one byte encodings by giving a meaning to 0 length.

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,65 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+SUBSCRIBE_UPDATE, PUBLISH_OK, TRACK_STATUS or FETCH message.  It is a structure
+indicating Objects or ranges of Objects that the Publisher will deliver.
+
+~~~
+Object Filter {
+  Type (0x5),
+  Length (i),
+  Operand (7),
+  Negation Flag (1),
+  [Extension ID (i),
```
```suggestion
  [Extension ID (i),]

File: draft-ietf-moq-transport.md [hunk 17]
```diff
@@ -1633,6 +1633,65 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+
+When Negation Flag is 1, the result of the filter is negated to determine which
+Objects pass the filter.
+
+Extension ID is present only when Operand is Extension ID (value 0x4) and
+indicates the Extension to filter. This MUST indicate an Extension with an
+integer type.  Objects without the Extension do not pass the filter.
+
+Values is an array of integers which encode the values of interest.  The
+remainder of the array is a sequence of pairs indicating the start and length of
```
yekuiwang: What does "the remainder" here mean?

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,65 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+the matching range.  The Start is encoded as a delta from the previous End, or
+from 0 for the first element.  The length indicates the number of elements
+including Start to match.  An odd number of elements indicates the final range
```
yekuiwang, line 1672: If we specify the length field to indicate the number of elements minus 1, then the same number of bits can represent more elements, i.e., the capability or bit efficiency is slightly improved.

draft-ietf-moq-transport.md
afrind: I think a previous version had the first element optionally be the extension Id, but it was pulled out

draft-ietf-moq-transport.md
afrind: It only saves one value though at the expense of some readability?

draft-ietf-moq-transport.md
```suggestion
array is a sequence of pairs indicating the start and length of
```

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,65 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+indicating Objects or ranges of Objects that the Publisher will deliver.
+
+~~~
+Object Filter {
+  Type (0x5),
+  Length (i),
+  Operand (7),
+  Negation Flag (1),
+  [Extension ID (i),]
+  Values (i) ...
```
```suggestion
  Values (..) ...
```

draft-ietf-moq-transport.md
```diff
@@ -1633,6 +1633,65 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+
+When Negation Flag is 1, the result of the filter is negated to determine which
+Objects pass the filter.
+
+Extension ID is present only when Operand is Extension ID (value 0x4) and
+indicates the Extension to filter. This MUST indicate an Extension with an
+integer type.  Objects without the Extension do not pass the filter.
+
+Values is an array of integers which encode the values of interest.  The
+array is a sequence of pairs indicating the start and length of
```
```suggestion
Values is an array of variable-length integers (i) which encode the values of
interest.  The array is a sequence of pairs indicating the start and length of
```

draft-ietf-moq-transport.md
yekuiwang: Yes. Just a suggestion for consideration. In some other standards this is always done.

afrind:
Parking pending Design Team input

mzanaty:
In the Toronto interim, it was decided to setup a design team for object/track filters (PR#1164, I#441, 1068, 1206). Mo, Martin, Suhas, Victor, Will, YK volunteered. Please let me know if you want to join these meetings. Next meeting is Friday October 10 at 9:00am PDT.
Immediately after the interim, we had an impromptu initial meeting with the folks still lingering in Toronto: Mo, Victor, Cullen, Mike, Magnus. We converged on a simplified design.
- The 'Operands' were removed in favor of separate parameter types for each filter type.
- The 'Values' array was removed in favor of a single range, Start/End inclusive with no delta coding. End=0 means no end unless Start=0. For multiple ranges, use multiple parameters.
- The 'Negation' flag was removed. We discussed allowing Start>End to mean negation. Maybe unnecessary, since any desired ranges can be expressed without negation using an extra range. For example, not 3-7 can be expressed as 0-2 and 8-0.
- The track filter only specifies NumberOfTopTracks. The selection metric is a new Selector extension. Some folks preferred this opaque Selector to avoid leaking private info, so the relay network is not aware these are voice activity, temperature, etc. metrics. However, this means only a single metric can be the Selector for all subscribers. An alternative is to specify the extension ID as well, which can be the Selector or another extension. List feedback preferred this alternative.
- A setup parameter will negotiate the number of filter ranges supported, including 0 to disable support.

mzanaty:
Oct 10, 2025 design meeting with Mo, Suhas, Victor, Will, YK.

Parameters in Subscribe, Subscribe_Update, Fetch, Publish_Ok:
No Filter { Type=TBD+0, Length=0 }
Group ID Filter { Type=TBD+2, Length, Start, [End] }
Subgroup ID Filter { Type=TBD+4, Length, Start, [End] }
Object ID Filter { Type=TBD+6, Length, Start, [End] }
Pub Priority Filter { Type=TBD+8, Length, Start, [End] }
Extension ID Filter { Type=TBD+a, Length, Extension ID, Start, [End] }

Parameter in Subscribe_Namespace:
Track Selection Filter { Type=TBD+c, Length, Extension ID, MaxTracksSelected }

Setup Parameters:
MAX_OBJECT_FILTER_RANGES (=0 to disable)
MAX_TRACKS_SELECTED (=0 to disable)
Applies to each type of filter per subscription

Track Selection Details
- Multiple Filters
   - Object filters evaluate first, then track selection.
- Tie Breaker
   - Same selection metric value, keep old track.
   - If app wants new track it can embed time in metric.
- Suggest to change metric infrequently? Start of group? Avoid erratic selection churn.
- Evict silent tracks? Avoid rogue publishers that send a max metric to evict others then send nothing to remain on top.
- Track Alias and Name needed upon new track selection
   - Send PUBLISH to deliver this
   - Parameters in PUBLISH may differ upon each new track selection
      - Delivery Timeout
      - Max Cache Duration
      - Publisher Priority

Consider in later PR: Remove SUBSCRIPTION_FILTER parameter? Need another way to say Next Group.

afrind:
@mzanaty Thanks for the update.

> Send PUBLISH to deliver this

Does that mean evicted tracks also get PUBLISH_DONE to keep the total number of subscriptions under this SUBSCRIBE_NAMESPACE=MaxTracksSelected?  I'd prefer that because it keeps the subscription state machine the same.  If you don't do this, you will leak subscriptions causing excessive state, and, with current mechanisms, likely run out of request IDs.

> Consider in later PR: Remove SUBSCRIPTION_FILTER parameter? Need another way to say Next Group.

I'd prefer that the draft have only one group filter mechanism at a time, so if you want to replace SUBSCRIPTION_FILTER, wait to add `Group ID Filter` until that follow up PR.

suhasHere:
> Does that mean evicted tracks also get PUBLISH_DONE to keep the total number of subscriptions under this SUBSCRIBE_NAMESPACE=MaxTracksSelected? I'd prefer that because it keeps the subscription state machine the same. If you don't do this, you will leak subscriptions causing excessive state, and, with current mechanisms, likely run out of request IDs.

This will be challenging. since this a dynamic selection and track can be selected at different points in time. Those tracks are all active , sending publish_done is incorrect here.

afrind:
> Those tracks are all active

What if MaxSelected is 5 and there are 1000 tracks, but every track has a moment where it's selector is in the top 5.  Then I end up with 1000 subscriptions.  Is that the intent?

mzanaty:
Oct 17, 2025 design meeting with Martin, Mo, Suhas, Victor, YK.

Decided to defer details for Fetch and removing Subscription Filter until core design is done.

Decided to put all Start/End ranges of a single filter type in a single parameter. Discussed but rejected putting all filter types in a single super parameter. Subscribe_Update always replaces all ranges of a given filter type, or removes the filter if no ranges (Length=0), which eliminated the "No Filter" type. 

Parameters in Subscribe, Subscribe_Update, Publish_Ok: (defer Fetch details)
Group Filter { Type=TBD1, Length, [Start, [End]]… }
Subgroup Filter { Type=TBD2, Length, [Start, [End]]… }
Object Filter { Type=TBD3, Length, [Start, [End]]… }
Priority Filter { Type=TBD4, Length, [Start, [End]]… }
Extension Filter { Type=TBD5, Length, [Extension ID, Start, [End]]… }

End is optional in the last range, meaning no End if omitted. For the Extension Filter, also allow End=0 to mean no End when Start>0, to allow different Extension IDs to have no End.

Need to clarify how filters (specifically Object or Extension Filters) can cause object gaps in subgroup streams, and how FIN is handled.

No changes to the Track Selection Filter. It remains as below.

Parameter in Subscribe_Namespace:
Track Selection Filter { Type=TBD6, Length, Extension ID, MaxTracksSelected }

Discussed whether deselected tracks get signaled in control messages like Publish Done. That results in extra control messages and subscription state churn if deselcted tracks later get reselected. However, many deselected tracks can pollute subscription state, so it may be good to allow publishers to purge state for many old deselected tracks. Need more discussion if "many old" can be left to implementations or we need to specify guidance.

afrind:
> Group Filter { Type=TBD1, Length, [Start, [End]]… }

For your next meeting can you discuss the interaction with SUBSCRIPTION_FILTER, or if you are replacing it entirely?

> Need to clarify how filters (specifically Object or Extension Filters) can cause object gaps in subgroup streams, and how FIN is handled.

See #1295.  I think the answer is along those lines -- if you omit an object from a subgroup, you MUST reset the stream.

> Need more discussion if "many old" can be left to implementations or we need to specify guidance.

We definitely need some guidance, even if the algorithm to close deselected subscriptions is implementation specific.  Eventually you can run out of request IDs.

mzanaty:
The final design intends to replace SUBSCRIPTION_FILTER entirely. But there are details (such as Next Group) that the team wants to defer until the core design aspects are finalized first.

mzanaty:
Nov 5, 2025 design meeting with Martin, Mo, Suhas, Y.K.

On Wed, Nov 5, 2025 at 8:36 AM Martin Duke (martin.h.duke@gmail.com) wrote:

1. SUBSCRIBE and SUBSCRIBE_NAMESPACE (and PUBLISH_OK and SUBSCRIBE_UPDATE) in one PR, FETCH in a separate PR.
3. Add three parameters; start location, end group, and next group. If there is no start location or next group parameter, the default will be Largest Object. Mo will add text that makes Largest Object clearer to him.
4. We believe this retains all the capabilities of the existing framework in the editors' draft.
5. All parameters and filters have an AND relationship. To be delivered, an object has to comply with all parameters and filters.
6. For FETCH, we will retain start and end location fields. the new parameters do not apply. The filters do.
7. SUBSCRIBE_NAMESPACE can have all these filters and parameters. They are default applied to any PUBLISHes that arrive in that namespace. PUBLISH_OK is always sent, and can update the filters/parameters for a specific subscription in that NAMESPACE.
However, any SUBSCRIBE in that namespace does not inherit params/filters from SUBSCRIBE_NAMESPACE.
8. Add a field to track selection filter to SUBSCRIBE_NAMESPACE called 'latch-time' on how long a publisher should remain in the relay's table without publishing an update on its metric.
9. If two tracks have the same metric, the relay will continue to forward the one that it's already forwarding to break the tie.
10. There is a track selection filter field 'max deselected tracks'. When the number of open PUBLISHes exceeds the limit, the publisher MUST send PUBLISH_DONE to bound receiver state. The fields 'max tracks' is meant to bound incoming message traffic, which is a different constraint.

mzanaty:
Nov 6, 2025 design meeting with Martin, Mo, Suhas.

From: Martin Duke [martin.h.duke@gmail.com]
Sent: Monday, November 10, 2025 1:07 PM
To: Mo Zanaty [mzanaty@cisco.com]; Ye-Kui Wang [yekui.wang@bytedance.com]; Suhas Nandakumar  [snandaku@cisco.com]
Cc: Law, Will [wilaw@akamai.com]; Victor Vasiliev [vasilvv@google.com]
Subject: Re: Summary of design team meeting

Update: Mo and Suhas want to get rid of LargestObject -- but if there's a joining fetch, that implicitly turns the subscribe into largest object.

My response:
	1.	I don't have any use cases, so do what you want. You have to convince Ye Kui?
	2.	If this happens, the default filter should be NextGroup, which eliminates a couple of parameters.
	3.	Organizationally, it's harder to land this if you're removing capabilities. Be prepared to back off to the original proposal if there's pushback.

afrind:
> Add three parameters; start location, end group, and next group. If there is no start location or next group parameter, the default will be Largest Object. Mo will add text that makes Largest Object clearer to him.

Shouldn't omitting start / next group mean the subscription is unfiltered?

> SUBSCRIBE_NAMESPACE can have all these filters and parameters. They are default applied to any PUBLISHes that arrive in that namespace.

But not to previously established subscriptions in that namespace, I assume.

> Add a field to track selection filter to SUBSCRIBE_NAMESPACE called 'latch-time' on how long a publisher should remain in the relay's table without publishing an update on its metric.

bikeshed: latch-time doesn't seem like a great name.  Something like min-publisher-freshness or something?

> If two tracks have the same metric, the relay will continue to forward the one that it's already forwarding to break the tie.

Ties go to earliest publisher.  Seems reasonable.

> max tracks / max deselected tracks

I think this is good pattern.  If we're using "selected" to define a track that passes the top-N filter at the current moment, then maybe call it "max selected tracks" and "max tracks".  So I SUBSCRIBE_NAMESPACE max-selected=5, max=10 (keep up to 5 displaced tracks open), or max-selected=3, max-tracks=3 (close displaced tracks immediately).  I think the idea of max-tracks is useful independent of filtering or top-N capabilities, to limit what you get when you SUB_NS.

> Mo and Suhas want to get rid of LargestObject

This seems orthogonal.  If we want to get this landed before the draft-16 cut, let's keep it as focused as possible.

mzanaty:
See latest design in #1401

PR #1207, opened 2025-09-05
Remove multiline [] notation for Extensions

afrind:
Addresses #708

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3237,7 +3237,15 @@ definition of the extension, Extension Headers MAY be modified, added, removed,
 and/or cached by relays.
 
 Object Extension Headers are serialized as Key-Value-Pairs (see
-{{moq-key-value-pair}}).
+{{moq-key-value-pair}}), prefixed by the length of the serialized
+Key-Value-Pairs, in bytes.
+
+~~~
+struct Extensions {
+  Extension Headers Length (i),
+  Extension headers (..),
+}
+~~~
 
 Header types are registered in the IANA table 'MOQ Extension Headers'.
 See {{iana}}.
@@ -3276,8 +3284,7 @@ OBJECT_DATAGRAM {
   Group ID (i),
   [Object ID (i),]
   Publisher Priority (8),
-  [Extension Headers Length (i),
-  Extension headers (...)],
+  [Extensions (..),]
   [Object Status (i),]
   [Object Payload (..),]
 }
@@ -3304,10 +3311,10 @@ There are 10 defined Type values for OBJECT_DATAGRAM.
 * End of Group: For Type values where End of Group is "Yes" the Object is the
   last Object in the Group.
 
-* Extensions Present: If Extensions Present is "Yes" the Extension Headers
-  Length and Extension headers fields are included. If an endpoint receives a
-  datagram with Extensions Present as "Yes" and a Extension Headers Length of 0,
-  it MUST close the session with PROTOCOL_VIOLATION.
+* Extensions Present: If Extensions Present is "Yes" the Extensions field is
+  included. If an endpoint receives a datagram with Extensions Present as "Yes"
+  and a Extension Headers Length of 0, it MUST close the session with
+  PROTOCOL_VIOLATION.
 
 * Object ID Present: If Object ID Present is No, the Object ID field is omitted
   and the Object ID is 0.  When Object ID Present is Yes, the Object ID field is
@@ -3414,10 +3421,10 @@ Subgroup ID field in the header and the Subgroup ID is either 0 (for Types
 0x10-11 and 0x18-19) or the Object ID of the first object transmitted in this
 subgroup (for Types 0x12-13 and 0x1A-1B).
 
-For Type values where Extensions Present is No, Extensions Headers Length is
-not present and all Objects have no extensions.  When Extensions Present is
-Yes, Extension Headers Length is present in all Objects in this subgroup.
-Objects with no extensions set Extension Headers Length to 0.
+For Type values where Extensions Present is No, the Extensions field is never
+present and all Objects have no extensions.  When Extensions Present is Yes, the
+Extensions field is present in all Objects in this subgroup.  Objects with no
+extensions set Extension Headers Length to 0.
 
 To send an Object with `Object Forwarding Preference` = `Subgroup`, find the open
 stream that is associated with the subscription, `Group ID` and `Subgroup ID`,
@@ -3438,8 +3445,7 @@ unless there is an Prior Object ID Gap extesnion header (see
 ~~~
 {
   Object ID Delta (i),
-  [Extension Headers Length (i),
-  Extension headers (...)],
+  [Extensions (..)],
   Object Payload Length (i),
   [Object Status (i)],
   Object Payload (..),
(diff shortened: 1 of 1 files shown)


yekuiwang (approved):

RichLogan:
If the extensions type is now an explicitly defined thing, could we not use that in the message specs? I.e `[Extensions (Extensions)]` rather than `(..)`? This is how `Location` is used today (e.g `[Start Location (Location)]`). Doubly so perhaps, given it is length prefixed and the notation uses a different notation for varint length prefixed byte arrays than `(..)` ?

Also I think the `struct` keyword is not used in the other definitions, but is present here.

draft-ietf-moq-transport.md
```diff
@@ -3237,7 +3237,15 @@ definition of the extension, Extension Headers MAY be modified, added, removed,
 and/or cached by relays.
 
 Object Extension Headers are serialized as Key-Value-Pairs (see
-{{moq-key-value-pair}}).
+{{moq-key-value-pair}}), prefixed by the length of the serialized
+Key-Value-Pairs, in bytes.
+
+~~~
+struct Extensions {
```
```suggestion
Extensions {
```

draft-ietf-moq-transport.md
```diff
@@ -3276,8 +3284,7 @@ OBJECT_DATAGRAM {
   Group ID (i),
   [Object ID (i),]
   Publisher Priority (8),
-  [Extension Headers Length (i),
-  Extension headers (...)],
+  [Extensions (..),]
```
afrind, line 3299:
```suggestion
  [Extensions (Extensions),]
```

draft-ietf-moq-transport.md
```diff
@@ -3438,8 +3445,7 @@ unless there is an Prior Object ID Gap extesnion header (see
 ~~~
 {
   Object ID Delta (i),
-  [Extension Headers Length (i),
-  Extension headers (...)],
+  [Extensions (..)],
```
afrind, line 3460:
```suggestion
  [Extensions (Extensions)],
```

draft-ietf-moq-transport.md
```diff
@@ -3574,8 +3580,7 @@ Each object sent on a fetch stream after the FETCH_HEADER has the following form
   Subgroup ID (i),
   Object ID (i),
   Publisher Priority (8),
-  Extension Headers Length (i),
-  [Extension headers (...)],
+  Extensions (..),
```
afrind, line 3595:
```suggestion
  Extensions (Extensions),
```

RichLogan (approved):
Cheers! Things like this make machine parsing the draft easier too.

draft-ietf-moq-transport.md
```diff
@@ -3414,10 +3421,10 @@ Subgroup ID field in the header and the Subgroup ID is either 0 (for Types
...
 subgroup (for Types 0x12-13 and 0x1A-1B).
 
-For Type values where Extensions Present is No, Extensions Headers Length is
-not present and all Objects have no extensions.  When Extensions Present is
-Yes, Extension Headers Length is present in all Objects in this subgroup.
-Objects with no extensions set Extension Headers Length to 0.
+For Type values where Extensions Present is No, the Extensions field is never
+present and all Objects have no extensions.  When Extensions Present is Yes, the
+Extensions field is present in all Objects in this subgroup.  Objects with no
+extensions set Extension Headers Length to 0.
```
suhasHere: I feel the behavior here and in line 3317 conflict on when extensions with header length is 0. I am incline to align the datagram section with this one where having extension length of 0 is not a protocol violation ?

@afrind  thoughts ?

suhasHere (commented):
This looks good and I did had one one observation which need not be addressed in this PR though.

draft-ietf-moq-transport.md
afrind: @suhasHere in a DATAGRAM, it _never_ makes sense to encode 0 length extensions, because you can use the type bit to encode that instead.  Subgroups can contain multiple objects and hence need a 0 length option.

This is an editorial PR, please file a new issue if you want a design change.

draft-ietf-moq-transport.md
```diff
@@ -3276,8 +3284,7 @@ OBJECT_DATAGRAM {
   Group ID (i),
   [Object ID (i),]
   Publisher Priority (8),
-  [Extension Headers Length (i),
-  Extension headers (...)],
+  [Extensions (Extensions),]
```
fluffy: This syntax is very confusing but  ... How do you know how many extensions exist ?

fluffy (changes requested):
I like getting rid of multi line array - I don't even think it is valid sysntax

I'm not sure exactly what your are proposing here and not 100% sure it work but I am probably just not understanding. Might need a bit more clarification

draft-ietf-moq-transport.md
afrind: I realized this is the wrong syntax.  It should be:

`Extensions (..)`

and below:

`If Extensions Present is yes, the Extensions structure defined in {{object-extensions}} is included.`


(And that structure includes the length)

```suggestion
  [Extensions (Extensions),]
```
```suggestion
  [Extensions (..),]
```

draft-ietf-moq-transport.md
```diff
@@ -3438,8 +3445,7 @@ unless there is an Prior Object ID Gap extesnion header (see
 ~~~
 {
   Object ID Delta (i),
-  [Extension Headers Length (i),
-  Extension headers (...)],
+  [Extensions (Extensions)],
```
```suggestion
  [Extensions (..)],
```

draft-ietf-moq-transport.md
```diff
@@ -3574,8 +3580,7 @@ Each object sent on a fetch stream after the FETCH_HEADER has the following form
   Subgroup ID (i),
   Object ID (i),
   Publisher Priority (8),
-  Extension Headers Length (i),
-  [Extension headers (...)],
+  Extensions (Extensions),
```
```suggestion
  Extensions (..),
```

draft-ietf-moq-transport.md
```diff
@@ -3304,10 +3311,10 @@ There are 10 defined Type values for OBJECT_DATAGRAM.
...
   last Object in the Group.
 
-* Extensions Present: If Extensions Present is "Yes" the Extension Headers
-  Length and Extension headers fields are included. If an endpoint receives a
-  datagram with Extensions Present as "Yes" and a Extension Headers Length of 0,
-  it MUST close the session with PROTOCOL_VIOLATION.
+* Extensions Present: If Extensions Present is "Yes" the Extensions field is
+  included. If an endpoint receives a datagram with Extensions Present as "Yes"
+  and a Extension Headers Length of 0, it MUST close the session with
+  PROTOCOL_VIOLATION.
```
```suggestion
* Extensions Present: If Extensions Present is "Yes" the Extensions structure
  defined in {{object-extensions}} is included. If an endpoint receives a
  datagram with Extensions Present as "Yes" and a Extension Headers Length of 0,
  it MUST close the session with PROTOCOL_VIOLATION.
```

draft-ietf-moq-transport.md
```suggestion
For Type values where Extensions Present is No, the Extensions field is never
present and all Objects have no extensions.  When Extensions Present is Yes, the
Extensions structure defined in {{object-extensions}} is present in all Objects
in this subgroup.  Objects with no extensions set Extension Headers Length to 0.
```

ianswett (approved):

PR #1209, opened 2025-09-08
Moving pieces of Subscribe, Refactor State Management

afrind:
1. Move the text describing Forward behavior and subscription filters out of SUBSCRIBE section, since these apply to subscriptions initiated via PUBLISH also.
  a. Relay specific forward behavior moved to relay section
3. Add a subsection heading for subscription state management
4. Move the text describing fetch state management into its own section
 
This PR only moves text around and does not change any descriptions or wire format behaviors.  A second PR will make changes.  If you want text moved to a different place, leave that feedback here.  If you want different descriptions or behaviors, please leave that specific feedback on the next PR.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -176,7 +176,7 @@ Endpoint:
 
 Peer:
 
-: The other endpoint than the one being described
+: The other endpoint than the one being described.
 
 Publisher:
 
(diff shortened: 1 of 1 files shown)


ianswett (approved):
Thanks for moving this, we have too much text in the frame definition sections, so this is a good move in the right direction.

PR #1210, opened 2025-09-08
Parameters for Group Order, Subscribe Priority and Subscription Filter

afrind:
Note: This is targeted against PR #1209.

This is a proposal based on the most recent discussion of parameters and extensibility.  

It moves Group Order and Subscriber Priority to parameters in SUBSCRIBE, SUBSCRIBE_UPDATE, PUBLISH, PUBLISH_OK.  The PR specifies a value for the receiver to use if the parameter is omitted.

It defines a structure for a Subscription Filter as it exists today.  
  * The wire format for SUBSCRIBE and PUBLISH_OK use this directly (not as a param, no wire format change)
  * SUBSCRIBE_UPDATE is changed to accept a parameter containing this filter when the subscriber wants to change it

Items of note:

1. This adds a new filter type "No Filter", addressing #1035.
2. Relative filter types are only allowed in SUBSCRIBE, because they require the SUBSCRIBE_OK containing Largest to communicate what actually happened, but maybe this could be relaxed?
3. The values used by Group Order are different depending on the message.  This preserves existing behavior while eliminating the "Default" wire value previously allowed in SUBSCRIBE.
4. Forward is retained as a Fixed field, rather than in all messages where it appears.  This is based on the fact that there's no sensible "common" value.
5. Other than Forward, everything that can be changed in SUBSCRIBE_UPDATE is now a parameter, and omitted means "unchanged".

Fixes: #872 
Fixes: #878 
Fixes: #1035 
Closes: #1043
Closes: #1044 
Closes: #1059
Closes: #1046

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -993,8 +993,8 @@ specified in `Start Location`, the remainder of that Group passes the
 filter. `End Group` MUST specify the same or a larger Group than specified in
 `Start Location`.
 
-An endpoint that receives a Filter Type other than the above MUST be close the
-session with a `PROTOCOL_VIOLATION`.
+An endpoint that receives a filter type other than the above MUST close the
+session with `PROTOCOL_VIOLATION`.
 
 ## Fetch State Management
 


afrind:
Note this PR does not parameterize Group Order, Subscriber Priority or any Location information in FETCH/OK, but it did remove Group Order from FETCH_OK, because the publisher _cannot_ change the order requested by the subscriber and "default" was not an option there.  Also removed discussion about priorities from FETCH group order, since it is not priority related in a FETCH context.

draft-ietf-moq-transport.md
```diff
@@ -1721,6 +1741,42 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+SUBSCRIBE_UPDATE, SUBSCRIBE_OK, PUBLISH or PUBLISH_OK.  It is an enum indicating
+how to prioritize Objects from different groups within the same subscription
+(see {{priorities}}). The allowed values are Ascending (0x1) or Descending (0x2)
+be used. If an endpoint receives a value outside this range, it MUST close the
+session with `PROTOCOL_VIOLATION`.
+
+If omitted from SUBSCRIBE, the publisher's preference is used and communicated
+in SUBSCRIBE_OK.  If omitted from SUBSCRIBE_OK, PUBLISH or PUBLISH_OK, the
+receiver uses Ascending (0x1).  If omitted from SUBSCRIBE_UPDATE, the value is
+unchanged.
```
fluffy: That this has a different semantic meaning depending on what messages it is in makes me think we should have two different parameters types

draft-ietf-moq-transport.md
```diff
@@ -1721,6 +1741,42 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+
+#### SUBSCRIPTION FILTER Parameter {#subscription-filter}
+
+The SUBSCRIPTION_FILTER paramter (Parameter Type 0x21) MAY appear in a
+SUBSCRIBE_UDPATE message. It is a length-prefixed Subscription Filter (see
+{{subscription-filters}}).  If the length of the Subscription Filter does not
+match the parameter length, the publisher MUST close the session with
```
fluffy: So this is a bit off what we discussed on the extensibility The filter type would be encoded here as tag and value because the receiver will know the length as it is mandatory to understand. There is no length for it so no chance to mismatch.

draft-ietf-moq-transport.md
afrind: What's funny is the semantics are only different _when omitted_.  I don't know how to fix that with multiple parameters, but I'm open to suggestions.

draft-ietf-moq-transport.md
afrind: 100% that is the plan but we're going to handle it holistically in a separate PR.  See #1184, which I think should be resolved at the same type as #794.

afrind:

* No one expressed a desire to use relative filters in PUBLISH_OK or SUBSCRIBE_UPDATE
* No requests to update end without start
* Group Order - we might want to split this into two parameters

Subscriber: GROUP_ORDER, default = Don't Care
Publisher: GROUP_ORDER_PREFERENCE, default = Ascending

If the subscriber omits GROUP_ORDER from PUBLISH_OK it's implicitly accepting the Publisher preference.
If the subscribe specifies an order other than the publisher's preference, it finds out in SUBSCRIBE_OK

* No one objected to keeping Forward as an inline field in all messages

draft-ietf-moq-transport.md
afrind: Now I see what you mean and we will try with two different params and see how it looks

draft-ietf-moq-transport.md
fluffy: OK - that makes int pretty confusing to review

afrind:
Fixes: #872

draft-ietf-moq-transport.md
```diff
@@ -946,28 +960,34 @@ to network reordering or prioritization, relays can receive Objects with
 Locations smaller than  `Largest Object` after the SUBSCRIBE is processed, but
 these Objects do not pass the Largest Object filter.
 
+This Filter Type is only allowed in SUBSCRIBE. If a publisher receives this
+Filter Type in PUBLISH_OK or SUBSCRIBE_UDPATE, it MUST close the session with
+`PROTOCOL_VIOLATION`.
+
```
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -946,28 +960,34 @@ to network reordering or prioritization, relays can receive Objects with
...
-message MAY be less than the `Largest Object` observed at the publisher. If the
-specified `End Group` is the same group specified in `Start`, the remainder of
-that Group passes the filter. `End Group` MUST specify the same or a larger Group
-than specified in `Start`.
-
-An endpoint that receives a filter type other than the above MUST be close the
+This Filter Type is only allowed in SUBSCRIBE. If a publisher receives this
+Filter Type in PUBLISH_OK or SUBSCRIBE_UDPATE, it MUST close the session with
+`PROTOCOL_VIOLATION`.
+
```
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -1721,6 +1741,42 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+how to prioritize Objects from different groups within the same subscription
+(see {{priorities}}). The allowed values are Ascending (0x1) or Descending (0x2)
+be used. If an endpoint receives a value outside this range, it MUST close the
+session with `PROTOCOL_VIOLATION`.
+
+If omitted from SUBSCRIBE, the publisher's preference is used and communicated
+in SUBSCRIBE_OK.  If omitted from SUBSCRIBE_OK, PUBLISH or PUBLISH_OK, the
+receiver uses Ascending (0x1).  If omitted from SUBSCRIBE_UPDATE, the value is
+unchanged.
+
```
afrind, line 1772:
```suggestion
#### GROUP ORDER Parameter {#group-order}

The GROUP_ORDER parameter (Parameter Type 0x22) MAY appear in a SUBSCRIBE,
SUBSCRIBE_UPDATE or PUBLISH_OK.  It is an enum indicating how to prioritize
Objects from different groups within the same subscription (see
{{priorities}}). The allowed values are Ascending (0x1) or Descending (0x2) be
used. If an endpoint receives a value outside this range, it MUST close the
session with `PROTOCOL_VIOLATION`.

If omitted from SUBSCRIBE, the publisher's preference is used and communicated
in SUBSCRIBE_OK via PUBLISHER_GROUP_ORDER_PREFERENCE.  If omitted from
SUBSCRIBE_OK or PUBLISH_OK, the receiver uses Ascending (0x1).  If omitted from
SUBSCRIBE_UPDATE, the value is unchanged.

#### PUBLISHER GROUP ORDER PREFERENCE Parameter {#publisher-group-order}

The PUBLISHER_GROUP_ORDER_PREFERENCE parameter (Parameter Type 0x24) MAY appear
in a SUBSCRIBE_OK or PUBLISH message.  It is an enum indicating the Publisher's
preference for Group Order (see {{priorities}}). The allowed values are
Ascending (0x1) or Descending (0x2) be used. If an endpoint receives a value
outside this range, it MUST close the session with `PROTOCOL_VIOLATION`.

If omitted, the receiver uses Ascending (0x1).

```

draft-ietf-moq-transport.md
```diff
@@ -1721,6 +1741,42 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+#### SUBSCRIPTION FILTER Parameter {#subscription-filter}
+
+The SUBSCRIPTION_FILTER paramter (Parameter Type 0x21) MAY appear in a
+SUBSCRIBE_UDPATE message. It is a length-prefixed Subscription Filter (see
+{{subscription-filters}}).  If the length of the Subscription Filter does not
```
yekuiwang: It may only appear in a SUBSCRIBE_UPDATE message?

draft-ietf-moq-transport.md
```diff
@@ -922,21 +922,35 @@ end with an error.
 
 ### Subscription Filters
 
-The subscriber specifies a filter on the subscription to allow
-the publisher to identify which objects need to be delivered.
+Subscribers can specify a filter on a subscription indicating to the publisher
```
yekuiwang, line 940: Should a default be specified? If so, meaning the type of No Filter?

If a default is specified, then "can specify a filter" should be changed to "can include a SUBSCRIPTION_FILTER parameter".

draft-ietf-moq-transport.md
```diff
@@ -922,21 +922,35 @@ end with an error.
...
+subscription.
 
-There are 4 types of filters:
+A Subscription Filter has the following structure:
+
+~~~
+Subscription Filter {
+  Filter Type (i),
+  [Start Location (Location)],
+  [End Group (i)],
```
```suggestion
  [Start Location (Location),]
  [End Group (i),]

File: draft-ietf-moq-transport.md [hunk 9]
```diff
@@ -922,21 +922,35 @@ end with an error.
 
 ### Subscription Filters
 
-The subscriber specifies a filter on the subscription to allow
-the publisher to identify which objects need to be delivered.
+Subscribers can specify a filter on a subscription indicating to the publisher
+which Objects to send.
 
 All filters have a Start Location and an optional End Group.  Only objects
 published or received via a subscription having Locations greater than or
```
yekuiwang, line 944:
```suggestion
If the filter type is No Filter (0x0), the subscription is not filtered and all
Objects published or received via the subscription are sent to the
Subscriber. Otherwise, all filters have a Start Location and an optional
End Group, and only objects published or received via a subscription
having Locations greater than or

File: draft-ietf-moq-transport.md [hunk 10]
```diff
@@ -922,21 +922,35 @@ end with an error.
...
+  Filter Type (i),
+  [Start Location (Location)],
+  [End Group (i)],
+}
+~~~
+
+There are 5 types of filters:
+
+No Filter (0x0): The subscription is not filtered and all Objects published or
+received via subscription are sent to the Subscriber.
```
```suggestion
received via a subscription are sent to the Subscriber.

File: draft-ietf-moq-transport.md (hunk 7 above)
Comment (afrind) [reply draft-ietf-moq-transport.md:940]: In this PR, the filter is not an optional parameter in SUBSCRIBE or PUBLISH_OK.  The default when omitted from SUBSCRIBE_UPDATE is "no change".

File: draft-ietf-moq-transport.md (hunk 6 above)
Comment (afrind) [reply draft-ietf-moq-transport.md]: That's the proposal here.

File: draft-ietf-moq-transport.md (hunk 9 above)
Comment (afrind) [reply draft-ietf-moq-transport.md:944]: It's semantic whether Filter Type = "No Filter" is a filter.  I'm inclined to say it _isn't_ and the original text "All filters have a Start..." is ok.

File: draft-ietf-moq-transport.md [hunk 11]
```diff
@@ -922,21 +922,35 @@ end with an error.
...
+subscription.
 
-There are 4 types of filters:
+A Subscription Filter has the following structure:
+
+~~~
+Subscription Filter {
+  Filter Type (i),
+ [Start Location (Location),]
+ [End Group (i),]
```
```suggestion
  Filter Type (i),
  [Start Location (Location),]
  [End Group (i),]
```

draft-ietf-moq-transport.md
afrind: @fluffy see the proposed text splitting to two parameters below.  It unfortunately didn't change the weirdness in the "what does it mean if it's missing" cases, but maybe it's cleaner?

draft-ietf-moq-transport.md
```diff
@@ -1721,6 +1733,42 @@ multi-object stream will expire earlier than Objects later in the stream. Once
 Objects have expired from cache, their state becomes unknown, and a relay that
 handles a downstream request that includes those Objects re-requests them.
 
+
+#### SUBSCRIBER PRIORITY Parameter {#subscriber-priority)
+
+The SUBSCRIBER_PRIORITY parameter (Parameter Type 0x20) MAY appear in a
+SUBSCRIBE, SUBSCRIBE_UPDATE or PUBLISH_OK message. It is an integer expressing
```
afrind: Should we replace the `Subscriber Priority` field in FETCH in this PR as well?

draft-ietf-moq-transport.md
yekuiwang: I see. In SUBSCRIBE etc. the structure is there, not the parameter.

draft-ietf-moq-transport.md
yekuiwang: OK, in that case, then we should not say below "There are 5 types of filters:"

draft-ietf-moq-transport.md
yekuiwang: Definitely yes. Otherwise in the future there will be lot of questions asking why should Subscriber Priority for FETCH be treated differently.

draft-ietf-moq-transport.md
```diff
@@ -922,21 +922,35 @@ end with an error.
...
 
-The `Largest Object` is defined to be the object with the largest Location
-({{location-structure}}) in the track from the perspective of the endpoint
-processing the SUBSCRIBE message. Largest Object updates when the first byte of
-an Object with a larger Location than the previous value is published or
-received through a subscription.
+Some filters are defined to be relative to the `Largest Object`. The `Largest
+Object` is the Object with the largest Location ({{location-structure}}) in the
+Track from the perspective of the publisher processing a SUBSCRIBE
+message. Largest Object updates when the first field of an Object with a
```
ianswett: First field is also a bit odd.  Could it be when you begin receiving the payload so you've gotten all the headers and you know they're valid?  Alternatively, it could just be when the Object's Group and Object IDs are known?

draft-ietf-moq-transport.md
```diff
@@ -922,21 +922,35 @@ end with an error.
...
+}
+~~~
+
+There are 5 types of filters:
+
+No Filter (0x0): The subscription is not filtered and all Objects published or
+received via a subscription are sent to the Subscriber.
 
 Largest Object (0x2): The filter Start Location is `{Largest Object.Group,
 Largest Object.Object + 1}` and `Largest Object` is communicated in
```
ianswett, line 970: Nit: Largest Object is also communicated in PUBLISH

draft-ietf-moq-transport.md
ianswett: Changing GROUP_ORDER in SUBSCRIBE_UPDATE is odd, do we need to allow that?  It also feels like it could be somewhat costly CPU-wise?

draft-ietf-moq-transport.md
```diff
@@ -1721,6 +1733,42 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+
+If omitted from SUBSCRIBE, the publisher's preference is used and communicated
+in SUBSCRIBE_OK.  If omitted from SUBSCRIBE_OK, PUBLISH or PUBLISH_OK, the
+receiver uses Ascending (0x1).  If omitted from SUBSCRIBE_UPDATE, the value is
+unchanged.
+
+#### SUBSCRIPTION FILTER Parameter {#subscription-filter}
+
+The SUBSCRIPTION_FILTER paramter (Parameter Type 0x21) MAY appear in a
+SUBSCRIBE_UDPATE message. It is a length-prefixed Subscription Filter (see
```
```suggestion
SUBSCRIBE_UPDATE message. It is a length-prefixed Subscription Filter (see
```

draft-ietf-moq-transport.md
```diff
@@ -922,21 +922,35 @@ end with an error.
...
+
+~~~
+Subscription Filter {
+  Filter Type (i),
+  [Start Location (Location),]
+  [End Group (i),]
+}
+~~~
+
+There are 5 types of filters:
```
```suggestion
Filter Type can have one of the following values:
```

draft-ietf-moq-transport.md
afrind: I will change that text

draft-ietf-moq-transport.md
```diff
@@ -922,21 +922,35 @@ end with an error.
...
-({{location-structure}}) in the track from the perspective of the endpoint
-processing the SUBSCRIBE message. Largest Object updates when the first byte of
-an Object with a larger Location than the previous value is published or
-received through a subscription.
+Some filters are defined to be relative to the `Largest Object`. The `Largest
+Object` is the Object with the largest Location ({{location-structure}}) in the
+Track from the perspective of the publisher processing a SUBSCRIBE
+message. Largest Object updates when the first field of an Object with a
+Location larger than the previous value is published or received through a
+subscription.
```
```suggestion
message. Largest Object updates when the first byte of an Object with a Location
larger than the previous value is published or received through a subscription.
```

draft-ietf-moq-transport.md
```diff
@@ -1721,6 +1733,42 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
 
+
+#### SUBSCRIBER PRIORITY Parameter {#subscriber-priority)
+
+The SUBSCRIBER_PRIORITY parameter (Parameter Type 0x20) MAY appear in a
+SUBSCRIBE, SUBSCRIBE_UPDATE or PUBLISH_OK message. It is an integer expressing
+the priority of a subscription relative to other subscriptions and fetch
+responses in the same session. Lower numbers get higher priority.  See
+{{priorities}}.  The range is restricted to 0-255.  If a publisher receives a
+value outside this range, it MUST close the session with `PROTOCOL_VIOLATION`.
```
afrind, line 1755:
```suggestion
The SUBSCRIBER_PRIORITY parameter (Parameter Type 0x20) MAY appear in a
SUBSCRIBE, SUBSCRIBE_UPDATE, PUBLISH_OK or FETCH message. It is an integer
expressing the priority of a subscription relative to other subscriptions and
fetch responses in the same session. Lower numbers get higher priority.  See
{{priorities}}.  The range is restricted to 0-255.  If a publisher receives a
value outside this range, it MUST close the session with `PROTOCOL_VIOLATION`.
```

draft-ietf-moq-transport.md
```diff
@@ -1721,6 +1733,42 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+SUBSCRIBE_UPDATE, SUBSCRIBE_OK, PUBLISH or PUBLISH_OK.  It is an enum indicating
+how to prioritize Objects from different groups within the same subscription
+(see {{priorities}}). The allowed values are Ascending (0x1) or Descending (0x2)
+be used. If an endpoint receives a value outside this range, it MUST close the
+session with `PROTOCOL_VIOLATION`.
+
+If omitted from SUBSCRIBE, the publisher's preference is used and communicated
+in SUBSCRIBE_OK.  If omitted from SUBSCRIBE_OK, PUBLISH or PUBLISH_OK, the
+receiver uses Ascending (0x1).  If omitted from SUBSCRIBE_UPDATE, the value is
+unchanged.
```
```suggestion
If omitted from SUBSCRIBE or PUBLISH_OK, the publisher's preference is used and
communicated in SUBSCRIBE_OK.  If omitted from SUBSCRIBE_OK or PUBLISH, the
receiver uses Ascending (0x1).
```

draft-ietf-moq-transport.md
```diff
@@ -1721,6 +1733,42 @@ multi-object stream will expire earlier than Objects later in the stream. Once
...
+If omitted from SUBSCRIBE or PUBLISH_OK, the publisher uses the value 128.
+
+#### GROUP ORDER Parameter {#group-order}
+
+The GROUP_ORDER parameter (Parameter Type 0x22) MAY appear in a SUBSCRIBE,
+SUBSCRIBE_UPDATE, SUBSCRIBE_OK, PUBLISH or PUBLISH_OK.  It is an enum indicating
+how to prioritize Objects from different groups within the same subscription
+(see {{priorities}}). The allowed values are Ascending (0x1) or Descending (0x2)
+be used. If an endpoint receives a value outside this range, it MUST close the
+session with `PROTOCOL_VIOLATION`.
```
afrind, line 1767:
```suggestion
The GROUP_ORDER parameter (Parameter Type 0x22) MAY appear in a SUBSCRIBE,
SUBSCRIBE_OK, PUBLISH or PUBLISH_OK.  It is an enum indicating how to prioritize
Objects from different groups within the same subscription (see
{{priorities}}). The allowed values are Ascending (0x1) or Descending (0x2) be
used. If an endpoint receives a value outside this range, it MUST close the
session with `PROTOCOL_VIOLATION`.
```

draft-ietf-moq-transport.md
afrind: I removed changing Group Order from this PR for now, pending working group signal on #1045 (and #607)

ianswett:
From Toronto:

Filters should be optional on SUBSCRIBE and PUBLISH_OK, and Group Order becomes a param in Fetch.  Otherwise this PR is heading in the right direction.

Whether Forward should become a param is in a separate PR that already exists.

draft-ietf-moq-transport.md
```diff
@@ -2173,24 +2193,18 @@ SUBSCRIBE_UPDATE Message {
...
-* Start Location : The starting location.
-
-* End Group: The end Group ID, plus 1. A value of 0 means the subscription is
-open-ended.
-
-* Subscriber Priority: Specifies the priority of a subscription relative to
-other subscriptions in the same session. Lower numbers get higher priority.
-See {{priorities}}.
+  existing Request ID.  The publisher MUST close the session with `
+  `PROTOCOL_VIOLATION if the subscriber specifies an invalid Subscription
```
```suggestion
  `PROTOCOL_VIOLATION` if the subscriber specifies an invalid Subscription
```

afrind:
Unfortunately, I have mismerged this PR horribly, redone in #1273

PR #1219, opened 2025-09-14
Dynamically generating new Groups

### LINKED ISSUE #1028: Protocol needs a better way to support applications that use dynamic sized groups

ianswett:
This is an alternative to #1060 using existing mechanisms in the draft, Relative Joining FETCH and MAX_CACHE_DURATION.

Fixes #1028

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1248,6 +1248,37 @@ all active subscriptions to the new relay. The new relay will send a
 response to the subscribes and if they are successful, the subscriptions
 to the old relay can be stopped with an UNSUBSCRIBE.
 
+### Joining an Ongoing Track
+
+The MoQ Object model is designed with the concept that the beginning of
+a Group is a joint point, so in order for a subscriber to join a Track,
+it needs to request an existing or future Group.  Different applications will
+have different approaches for when to begin a new Group.
+
+To join a Track at a past Group, a SUBSCRIBE for Largest Object followed
+by a Joining FETCH for the intended start Group can be sent.  To join a Track
+at the next Group, a SUBSCRIBE for Next Group Start can be sent.
+
+#### Dynamically Starting New Groups
+
+Some applications might want to only begin a new Group when needed, because
+the information needed to join a Track is large relative to the incremental
+information in other Objects.  Additionally, it is ideal to only contact the
+original publisher for a new Group when it's been long enough since a prior
+Group was published to avoid overwhelming the Original Publisher.
+
+This can be achieved by having the subscriber join a Track at the largest Group
+(A Joining Relative Fetch with Joining Start = 0) and Publisher specifying
+a MAX_CACHE_DURATION that is the minimum interval at which it's willing to
+start a new Group.
+
+If there are no subscribers, the subscription will go to the Original Publisher
+and it can begin a Group. If a largest Group has been produced and is still
+cached in a Relay, it is returned and the new subscriber joins.  If a largest
+Group has ben produced, but it is not cached, the subscription will go to the
+Original Publisher and the Original Publisher can begin a new Group and return
+that to the newly joining subcriber.
+
 
 ## Publisher Interactions
 


draft-ietf-moq-transport.md
```diff
@@ -1248,6 +1248,37 @@ all active subscriptions to the new relay. The new relay will send a
...
+(A Joining Relative Fetch with Joining Start = 0) and Publisher specifying
+a MAX_CACHE_DURATION that is the minimum interval at which it's willing to
+start a new Group.
+
+If there are no subscribers, the subscription will go to the Original Publisher
+and it can begin a Group. If a largest Group has been produced and is still
+cached in a Relay, it is returned and the new subscriber joins.  If a largest
+Group has ben produced, but it is not cached, the subscription will go to the
+Original Publisher and the Original Publisher can begin a new Group and return
+that to the newly joining subcriber.
```
suhasHere, line 1280: We might need some mechanism to limit such requests reaching Original Publisher in the cases where cached data is either not useful or not available, esp when there are huge number of subscribers triggering joining fetches.

One more point may be worth considering is, for large groups in cache, the objects get delivered at network rate for all such subscribers and cause momentary congestion over the downlinks.

draft-ietf-moq-transport.md
```diff
@@ -1248,6 +1248,37 @@ all active subscriptions to the new relay. The new relay will send a
...
+Some applications might want to only begin a new Group when needed, because
+the information needed to join a Track is large relative to the incremental
+information in other Objects.  Additionally, it is ideal to only contact the
+original publisher for a new Group when it's been long enough since a prior
+Group was published to avoid overwhelming the Original Publisher.
+
+This can be achieved by having the subscriber join a Track at the largest Group
+(A Joining Relative Fetch with Joining Start = 0) and Publisher specifying
+a MAX_CACHE_DURATION that is the minimum interval at which it's willing to
+start a new Group.
```
suhasHere, line 1273: I wonder how would MAX_CACHE_DURATION might interact with publishers who don't follow fixed sized groups.

draft-ietf-moq-transport.md
```diff
@@ -1248,6 +1248,37 @@ all active subscriptions to the new relay. The new relay will send a
...
+#### Dynamically Starting New Groups
+
+Some applications might want to only begin a new Group when needed, because
+the information needed to join a Track is large relative to the incremental
+information in other Objects.  Additionally, it is ideal to only contact the
+original publisher for a new Group when it's been long enough since a prior
+Group was published to avoid overwhelming the Original Publisher.
+
+This can be achieved by having the subscriber join a Track at the largest Group
+(A Joining Relative Fetch with Joining Start = 0) and Publisher specifying
```
suhasHere, line 1271: This is a neat idea. Our current implementation of media client does follow this design.

libquicr media clients does joining fetch today when it knows the cache is not too deep  or data is available in the cache, to sync to the latest group really quick. For cases where this would not work, solutions that request OP to generate new sync point seems useful.

draft-ietf-moq-transport.md
```diff
@@ -1248,6 +1248,37 @@ all active subscriptions to the new relay. The new relay will send a
...
+a MAX_CACHE_DURATION that is the minimum interval at which it's willing to
+start a new Group.
+
+If there are no subscribers, the subscription will go to the Original Publisher
+and it can begin a Group. If a largest Group has been produced and is still
+cached in a Relay, it is returned and the new subscriber joins.  If a largest
+Group has ben produced, but it is not cached, the subscription will go to the
```
fluffy, line 1278: tiny typo s/ben/been/

draft-ietf-moq-transport.md
```diff
@@ -1248,6 +1248,37 @@ all active subscriptions to the new relay. The new relay will send a
...
+by a Joining FETCH for the intended start Group can be sent.  To join a Track
+at the next Group, a SUBSCRIBE for Next Group Start can be sent.
+
+#### Dynamically Starting New Groups
+
+Some applications might want to only begin a new Group when needed, because
+the information needed to join a Track is large relative to the incremental
+information in other Objects.  Additionally, it is ideal to only contact the
+original publisher for a new Group when it's been long enough since a prior
+Group was published to avoid overwhelming the Original Publisher.
```
fluffy, line 1268: When to start a new group can be a bit more than just an sufficient time has passed. Publishers might want to sync with a camera angle change for example.

draft-ietf-moq-transport.md
```diff
@@ -1248,6 +1248,37 @@ all active subscriptions to the new relay. The new relay will send a
...
+
+This can be achieved by having the subscriber join a Track at the largest Group
+(A Joining Relative Fetch with Joining Start = 0) and Publisher specifying
+a MAX_CACHE_DURATION that is the minimum interval at which it's willing to
```
fluffy, line 1272: I think we would want a new parameter for this because many things will have cache durations that are much longer than the max group length. For example, a streaming video might want to make a new group every 30 seconds during normal playout but be willing to make them every 5 seconds when new people are joining and want to cache data for 30 minutes so viewers can scrub back and see older stuff.

draft-ietf-moq-transport.md
```diff
@@ -1248,6 +1248,37 @@ all active subscriptions to the new relay. The new relay will send a
...
+start a new Group.
+
+If there are no subscribers, the subscription will go to the Original Publisher
+and it can begin a Group. If a largest Group has been produced and is still
+cached in a Relay, it is returned and the new subscriber joins.  If a largest
```
fluffy, line 1277: But this will return all the frames that have been produced up to this point in the groups and that is a huge chunk of bandwidth which likely will not work. The app is trying to request a group to get a new join point exactly so it does not need to do a fetch of all this old frames and the bandwidth that requires.

draft-ietf-moq-transport.md
```diff
@@ -1248,6 +1248,37 @@ all active subscriptions to the new relay. The new relay will send a
...
+Group has ben produced, but it is not cached, the subscription will go to the
+Original Publisher and the Original Publisher can begin a new Group and return
```
fluffy, line 1279: In the case of many downstream subscribers all joining and roughly the same time and a non caching relay, this would result in each one of them creating a separate upstream request which we explicitly want to avoid because it causes "start of service" attack on the original publisher

fluffy (changes requested):
I think I like #1060 better. The weaknesses of this approach is that in the non caching relay case, it does not aggregate the upstream requests for new groups and in the other case of a caching relay, it does get a new group but instead floods the subscriber will all the objects in the group of to that point.

afrind:
> the weaknesses of this approach is that in the non caching relay case, it does not aggregate the upstream requests for new groups and in the other case of a caching relay, it does get a new group but instead floods the subscriber will all the objects in the group of to that point.

Both caching and non-caching relays can (and probably should) aggregate upstream fetches for the same data.

That said, I think the limitation that "min group duration == max cache duration" is significant.  I don't see how to solve that by adding a separate parameter without also changing the semantics of FETCH, since the joining FETCH to the relay will get a cache hit up to max cache duration, and not trigger a new group.

afrind:
The high level text on joining tracks here (first section) has been moved into #1060.  Close?

PR #1220, opened 2025-09-14
Add Forward to SUBSCRIBE_NAMESPACE

ianswett:
Because you might not want all the Objects for all Tracks until you find out what the Tracks are.

This has been discussed a number of times, including during the Params discussions, but I am not sure if it fixes any filed issues.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2969,6 +2969,7 @@ SUBSCRIBE_NAMESPACE Message {
   Length (16),
   Request ID (i),
   Track Namespace Prefix (..),
+  Forward (8),
   Number of Parameters (i),
   Parameters (..) ...
 }
@@ -2986,6 +2987,12 @@ SUBSCRIBE_NAMESPACE Message {
   ("example.com", "meeting=123") would match both.  If an endpoint receives a
   Track Namespace Prefix consisting of 0 or greater than than 32 Track Namespace
   Fields, it MUST close the session with a `PROTOCOL_VIOLATION`.
+
+ 
+* Forward: The Forward value that new subscriptions resulting from this
+  SUBSCRIBE_NAMESPACE will have. If 0, the subscriber will be notified of all
+  available tracks in the namespace, but not receive Objects unless it changes
+  Forward to 1 in PUBLISH_OK.
 
 * Parameters: The parameters are defined in {{version-specific-params}}.
 


draft-ietf-moq-transport.md
```diff
@@ -2987,6 +2988,12 @@ SUBSCRIBE_NAMESPACE Message {
   Track Namespace Prefix consisting of 0 or greater than than 32 Track Namespace
   Fields, it MUST close the session with a `PROTOCOL_VIOLATION`.
 
+ 
+* Forward: The Forward value that new subscriptions resulting from this
+  SUBSCRIBE_NAMESPACE will have. If 0, the subscriber will be notified of all
+  available tracks in the namespace, but not receive Objects unless it changes
+  Forward to 1 in PUBLISH_OK.
```
```suggestion
  Forward to 1 in PUBLISH_OK or SUBSCRIBE_UPDATE.
```

draft-ietf-moq-transport.md
```diff
@@ -2987,6 +2988,12 @@ SUBSCRIBE_NAMESPACE Message {
   Track Namespace Prefix consisting of 0 or greater than than 32 Track Namespace
   Fields, it MUST close the session with a `PROTOCOL_VIOLATION`.
 
+
+* Forward: The Forward value that new subscriptions resulting from this
+  SUBSCRIBE_NAMESPACE will have. If 0, the subscriber will be notified of all
+  available tracks in the namespace, but not receive Objects unless it changes
+  Forward to 1 in PUBLISH_OK or SUBSCRIBE_UPDATE.
```
```suggestion
  available tracks in the namespace via PUBLISH messages with Forward set to
  0. The subscriber can elect to receive Objects by setting Forward to 1 in
  PUBLISH_OK or SUBSCRIBE_UPDATE.
```

afrind:
> Should Priority be added as well?

This was sort of addressed in #1047 ?

draft-ietf-moq-transport.md
suhasHere: I like the original intent here of sending the objects along with publish_ok if the forward is true on the subscribe_namespace. This will avoid extra round trip when the application knows it is going to do an OK for the publishes that will come from this namespace subscription

```suggestion
* Forward: The Forward value determines the behavior of new subscriptions created under this SUBSCRIBE_NAMESPACE. If set to 0, the subscriber will be notified of all available tracks in the namespace through PUBLISH messages with Forward set to 0. The subscriber can then opt to receive objects by setting Forward to 1 in a PUBLISH_OK or SUBSCRIBE_UPDATE message. If set to 1, PUBLISH messages (with Forward set to 1) and the corresponding objects are forwarded as they are received and the subscriber may choose to respond with Forward set to 0 in a PUBLISH_OK or SUBSCRIBE_UPDATE message, if necessary.
```

ianswett: This suggestion looks fine as well I think?  I accepted Alan's before seeing this, but I don't think I have a strong opinion.

suhasHere: Yeah . I wanted to clarify that if a subscriber chooses to forward=1 in subscribe namespace, it can expect to  receive publish_ok and objects associated with tracks without awaiting a round trip of publish_ok. I am totally fine if you feel comfortable that the accpeted text captures the intent.

draft-ietf-moq-transport.md
```diff
@@ -2987,6 +2988,13 @@ SUBSCRIBE_NAMESPACE Message {
   Track Namespace Prefix consisting of 0 or greater than than 32 Track Namespace
   Fields, it MUST close the session with a `PROTOCOL_VIOLATION`.
 
+
+* Forward: The Forward value that new subscriptions resulting from this
+  SUBSCRIBE_NAMESPACE will have. If 0, the subscriber will be notified of all
+  available tracks in the namespace via PUBLISH messages with Forward set to
+  0. The subscriber can elect to receive Objects by setting Forward to 1 in
+  PUBLISH_OK or SUBSCRIBE_UPDATE.
```
```suggestion
* Forward: The Forward value that new subscriptions resulting from this
  SUBSCRIBE_NAMESPACE will have (see {{subscriptions}}).
```

Issue #1216, opened 2025-09-09
SUBSCRIBE_UPDATE can not cause an Object to be delivered multiple times?


yekuiwang:
In Sect. 9.10 [SUBSCRIBE_UPDATE](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#name-subscribe_update):

> Unlike a new subscription, SUBSCRIBE_UPDATE can not cause an Object to be delivered multiple times.

I failed to find a constraint that prohibits SUBSCRIBE_UPDATE from causing an Object to be delivered multiple times. If no such constraint exists, then is this statement correct?

afrind:
I think maybe we should remove this sentence.  There was a point in the document when we allowed multiple subscriptions to the same track, so perhaps "a new subscription" would cause multiple objects to be delivered.  But now such a subscription is not allowed, and there's other language saying that sometimes subscribers will receive duplicate objects -- Relays support multiple publishers and can't track everything that's been delivered to every subscriber.

Issue #1213, opened 2025-09-08
On "close the Session with Key-Value Formatting error"


yekuiwang:
In Sect. 9.2.1.1 ([AUTHORIZATION TOKEN](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#name-authorization-token))

> If the Token structure cannot be decoded, the receiver MUST close the Session with Key-Value Formatting error.

Not sure whether this is an error at all, but the style of "Key-Value Formatting error" seems not aligned with other places of closing the Session with an error code, and it is not resolved in PR #1159.

Since I am not sure whether this is a minor editorial error, I did not include it in issue #1211.

afrind:
> Not sure whether this is an error at all:

I think it is an error.  By Token Structure, we mean:

```
Token {
  Alias Type (i),
  [Token Alias (i),]
  [Token Type (i),]
  [Token Value (..)]
}
```

as opposed to the application specific "Token Value" contents might be.

So an example might be 

`Type=AUTH_TOKEN, Length = 1, Value={Alias Type=2}`

Here the value ends after the Alias Type, but it should also include a Token Alias.  Key-Value Formatting error is chosen because the length or value is incorrect.

I'd say it's editorial to change it to KEY_VALUE_FORMATTING_ERROR.

Issue #1211, opened 2025-09-08
Some minor editorial issues


yekuiwang:
(Note: This issue is to just to document some newly identified minor editorial problems, for which the fixes, probably together with fixes for some earlier editorial issues, are to be provided in one PR.)

In Sect. 8.4.1 ([Graceful Publisher Network Switchover](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#name-graceful-publisher-network-)): 

> Ideally, the subscriptions downstream from the relay do no observe this change, and keep receiving the Objects on the same subscription.

"do no" -> "do not"

In Sect. 8.5 ([Relay Object Handling](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#name-relay-object-handling)):

> A publisher SHOULD begin sending incomplete objects when available to avoid incurring additional latency.

This sentence is about publisher handling of objects, while the entire Section 8 is about relays. Should this sentence be moved to somewhere else?

afrind:
> This sentence is about publisher handling of objects, while the entire Section 8 is about relays. Should this sentence be moved to somewhere else?

Yes.  Maybe in the "Subscriptions" section before this paragraph:

"Publishers MAY start sending Objects on PUBLISH-initiated subscriptions...

yekuiwang:
In Sect. 9.1 ([Request ID](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#name-request-id)):

> There are independent Request IDs for each endpoint.

This sentence does not disallow dependent Request IDs for each endpoint. Was the intent actually to say that "Request IDs for one endpoint are independent from those for another endpoint"?

yekuiwang:
In Sect. 9.2:

> Setup message parameters use a namespace that is constant across all MoQ Transport versions.

MoQ Transport -> MOQT

yekuiwang:
In Sect. 9:

[Figure 3](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#figure-3): [MOQT Message](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#name-moqt-message)

MOQT Message -> MOQT Control Message

afrind:
> This sentence does not disallow dependent Request IDs for each endpoint. Was the intent actually to say that "Request IDs for one endpoint are independent from those for another endpoint"?

Maybe 

> Request IDs for one endpoint increment independently from those sent by the peer endpoint.

?

[...]

Issue #1203, opened 2025-09-04
Step 4 of the Scheduling Algorithm should be updated


yekuiwang:
At the beginning of Sect. 7.2 ([Scheduling Algorithm](https://www.ietf.org/archive/id/draft-ietf-moq-transport-14.html#name-scheduling-algorithm)), we have: 

> When an MOQT publisher has multiple schedulable objects it can choose between, the objects SHOULD be selected as follows:
> 1. If two objects have different subscriber priorities associated with them, the one with the highest subscriber priority is scheduled to be sent first.
> 2. If two objects have the same subscriber priority, but different publisher priorities, the one with the highest publisher priority is scheduled to be sent first.
> 3. If two objects in response to the same request have the same subscriber and publisher priority, but belong to two different groups of the same track, the group order of the associated subscription is used to decide the one that is scheduled to be sent first.
> 4. If two objects in response to the same request belong to the same group of the same track, the one with the lowest Subgroup ID (for tracks with delivery preference Subgroup), or the lowest Object ID (for tracks with delivery preference Datagram) is scheduled to be sent first.

Step 4 should be changed to be as follows:

> 4. If two objects in response to the same request _have the same subscriber and publisher priority and_ belong to the same group of the same track, the one with the lowest Subgroup ID (for tracks with delivery preference Subgroup), or the lowest Object ID (for tracks with delivery preference Datagram) is scheduled to be sent first.

afrind:
It's a little redundant, but ok.

If two schedulable objects had different subscriber or publisher priority, the algorithm would have terminated at step 1 or 2.  Further, since there is only a single subscriber priority per track, then these necessarily have the same subscriber priority.

I guess step 3 mentions it, so we should be consistent one way or the other though.

yekuiwang:
> Further, since there is only a single subscriber priority per track, 

Another new fact to me! I just went through all places containing the 50 instances of "Priority" in -v14, but could not find where is this specified or implied. Could you please let me know?

afrind:
> Subscriber Priority is a priority number associated with an individual request. It is specified in the SUBSCRIBE or FETCH message, and can be updated via SUBSCRIBE_UPDATE message. The subscriber priority of an individual schedulable object is the subscriber priority of the request that caused that object to be sent.

yekuiwang:
[...]

PR #1222, opened 2025-09-14
Misc. editorial changes set2 to -v14

yekuiwang:
Fixes: #1193
Fixes: #1196
Fixes: #1198
Fixes: #1203
Fixes: #1211
Fixes: #1213
Fixes: #1216


draft-ietf-moq-transport.md
```diff
@@ -907,6 +908,9 @@ FETCH. A subscriber MUST send exactly one PUBLISH_OK or PUBLISH_ERROR in
 response to a PUBLISH. The peer SHOULD close the session with a protocol error
 if it receives more than one.
 
+A publisher SHOULD begin sending incomplete objects when available to avoid
```
ianswett, line 911: This might be difficult for some implementations, so I'm not sure we want a SHOULD?  Maybe RECOMMENDED?

```suggestion
It is RECOMMENDED a publisher begin sending incomplete Objects when available to avoid
```

draft-ietf-moq-transport.md
```diff
@@ -1462,14 +1464,15 @@ the length of the Message Payload, the receiver MUST close the session with a
...
-not expected, it MUST close the session with `INVALID_REQUEST_ID`.
+and frequency of requests.  Request IDs for one endpoint increment independently
+from those sent by the peer endpoint.  The client's Request ID starts at 0 and
+are even and the server's Request ID starts at 1 and are odd.  The Request ID
+increments by 2 with each FETCH, SUBSCRIBE, SUBSCRIBE_UPDATE,
+SUBSCRIBE_NAMESPACE, PUBLISH, PUBLISH_NAMESPACE or TRACK_STATUS request.
+Other messages with a Request ID field reference the Request ID of another
+message for correlation. If an endpoint receives a Request ID that is not valid
+for the peer, or a new request with a Request ID that is not expected, it MUST
+close the session with `INVALID_REQUEST_ID`.
```
```suggestion
message for correlation. If an endpoint receives a Request ID that is not valid
for the peer, or a new request with a Request ID that is not the next in
sequence or exceeds the received MAX_REQUEST_ID, it MUST close the session with
`INVALID_REQUEST_ID`.
```

draft-ietf-moq-transport.md
```diff
@@ -3170,9 +3172,11 @@ A canonical MoQ Object has the following information:
 the Object {{priorities}}.
 
 * Object Forwarding Preference: An enumeration indicating how a publisher sends
-an object. The preferences are Subgroup and Datagram.  When in response to a
-SUBSCRIBE, an Object MUST be sent according to its `Object Forwarding
-Preference`, described below.
+an object. The preferences are Subgroup and Datagram.  Note that the Original
+Publisher determines the Forwarding Preference for the entire Track, and is a
+Track property that is implicitly signaled by the delivery of any Object using
```
ianswett, line 3177: The forwarding preference is only implicitly signaled?  That feels odd, but nothing to fix in this PR.

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1554,10 +1557,10 @@ Token {
 |------|------------|------------------------------------------------------|
 
 
-* Token Alias - a Session-specific integer identifier that references a Token
-  Value. There are separate Alias spaces for the client and server (e.g.: they
-  can each register Alias=1). Once a Token Alias has been registered, it cannot
-  be re-registered by the same sender in the Session without first being
+* Token Alias - a Session-and-endpoint-specific integer identifier that references
```
suhasHere: Just trying to understand the context behind this change a bit more. How are we defining end-point in this case ?
I am not sure if this is purely editorial change , should we create a different PR ?

draft-ietf-moq-transport.md
```diff
@@ -1763,7 +1766,7 @@ value is 0, so if not specified, the peer MUST NOT send requests.
 
 The MAX_AUTH_TOKEN_CACHE_SIZE parameter (Parameter Type 0x04) communicates the
 maximum size in bytes of all actively registered Authorization tokens that the
-server is willing to store per Session. This parameter is optional. The default
+endpoint is willing to store per Session. This parameter is optional. The default
```
suhasHere, line 1770: isn't this server only state ?

draft-ietf-moq-transport.md
```diff
@@ -2403,7 +2305,8 @@ PUBLISH_OK Message {
 * Group Order: Indicates the subscription will be delivered in
   Ascending (0x1) or Descending (0x2) order by group. See {{priorities}}.
   Values of 0x0 and those larger than 0x2 are a protocol error. This
-  overwrites the GroupOrder specified PUBLISH.
+  overwrites the GroupOrder specified in the PUBLISH this message is
+  replying to.
```
suhasHere: I feel the original sentence was fine here

draft-ietf-moq-transport.md
yekuiwang: Thanks Ian for reviewing the changes. FYI: This sentence was moved from another section without any change.

BTW, I thought SHOULD wording and RECOMMENDED wording are equivalent and have the exact same meaning. Am I missing anything herein?

draft-ietf-moq-transport.md
yekuiwang: I like these changes, which make "not expected" much clearer.

draft-ietf-moq-transport.md
yekuiwang: That surprised me as well.

draft-ietf-moq-transport.md
yekuiwang: Thanks Suhas for reviewing the changes. It was intended to be a purely editorial change. Discussions for this change are herein:
- https://github.com/moq-wg/moq-transport/issues/1211#issuecomment-3271533610
- https://github.com/moq-wg/moq-transport/issues/1211#issuecomment-3271637151
- https://github.com/moq-wg/moq-transport/issues/1211#issuecomment-3271679654
- https://github.com/moq-wg/moq-transport/issues/1211#issuecomment-3271726301
- https://github.com/moq-wg/moq-transport/issues/1211#issuecomment-3271750976

draft-ietf-moq-transport.md
yekuiwang: The discussions for this change are:
- https://github.com/moq-wg/moq-transport/issues/1211#issuecomment-3271725266
- https://github.com/moq-wg/moq-transport/issues/1211#issuecomment-3271729495

draft-ietf-moq-transport.md
yekuiwang: The original sentence is "This overwrites the GroupOrder specified PUBLISH." The changed sentence is "This overwrites the GroupOrder specified in the PUBLISH this message is replying to."

To me, the original sentence is incomplete. At least "specified PUBLISH" should be changed to "specified in the  PUBLISH".

But in any case, you editors decide.

draft-ietf-moq-transport.md
```diff
@@ -410,7 +410,8 @@ scheduling of sending data on active streams.
 
 Every object within a Group belongs to exactly one Subgroup.
 
-Objects from two subgroups cannot be sent on the same stream. Objects from the
+When Objects are transported in a stream that is not a response to a FETCH, Objects
+from two subgroups cannot be sent on the same stream, and Objects from the
```
```suggestion
When Objects are sent in a subscription (see {{subscriptions}}),  Objects
from two subgroups cannot be sent on the same stream, and Objects from the
```

draft-ietf-moq-transport.md
afrind: Let's keep this PR editorial and just move the existing text.

draft-ietf-moq-transport.md
afrind: Maybe "session-and-endpoint" is not necessary after all given the context from next sentence.

eg: a session-specific integer...  There are separate Alias spaces for client and server.

draft-ietf-moq-transport.md
afrind: Since MoQT is symmetric and either endpoint can send requests which contain auth tokens, I think the client needs to follow the same rules -- at least that's how I interpreted it.  The default for this param is 0 anyways, so not sending means you don't need any state.  Do we need a design issue for discussion?

draft-ietf-moq-transport.md
```suggestion
  overwrites the GroupOrder specified in the corresponding PUBLISH message.
```

draft-ietf-moq-transport.md
```diff
@@ -2451,6 +2354,105 @@ NOT_SUPPORTED (0x3):
 UNINTERESTED (0x4):
 : The namespace or track is not of interest to the endpoint.
 
+## PUBLISH_DONE {#message-publish-done}
```
afrind, line 2357: I'd prefer this section move was in its own PR.  In any case I assume none of the text changed?

draft-ietf-moq-transport.md
```diff
@@ -3170,9 +3172,11 @@ A canonical MoQ Object has the following information:
...
 
 * Object Forwarding Preference: An enumeration indicating how a publisher sends
-an object. The preferences are Subgroup and Datagram.  When in response to a
-SUBSCRIBE, an Object MUST be sent according to its `Object Forwarding
-Preference`, described below.
+an object. The preferences are Subgroup and Datagram.  Note that the Original
+Publisher determines the Forwarding Preference for the entire Track, and is a
+Track property that is implicitly signaled by the delivery of any Object using
+either Subgroups or Datagrams. Once the property is established for one Object
+of a Track, the same value MUST be used for all Objects of the Track.
```
afrind, line 3179: You removed the old MUST

>  an Object MUST be sent according to its `Object Forwarding Preference`, 

This was for relays and we should try to preserve that, since this PR Is editorial.

draft-ietf-moq-transport.md
yekuiwang: @afrind Is that sufficiently clear, since joining fetches are associated with a subscription? It seems to me that wording it this way readers can still be confused that this statement applies to the response to a joining fetch?

draft-ietf-moq-transport.md
yekuiwang: OK

draft-ietf-moq-transport.md
yekuiwang: Looks good.

draft-ietf-moq-transport.md
yekuiwang: Right, no change as part of the move; that's why I put the move in a separate commit.

draft-ietf-moq-transport.md
yekuiwang: The following sentence is equivalent, isn't it?

> Once the property is established for one Object of a Track, the same value MUST be used for all Objects of the Track.

And the original sentence has an issue with the very first Object to be sent, as for that Object, the OP determines the value of the Forwarding Preference, not the other way around that the sending needs to follow the Forwarding Preference value.

draft-ietf-moq-transport.md
afrind: joining fetches are not subscriptions, they are fetches that dynamically determine the range.

draft-ietf-moq-transport.md
afrind: Ah, there's another missing unexpected case - there could be a gap (eg: 0, 4) is invalid.

draft-ietf-moq-transport.md
afrind: ok

draft-ietf-moq-transport.md
afrind: I think we lost a little explicitness about the fact that relays preserve the property that it would be good to retain somehow.  Let's keep the first half of original sentence in there after your new bit:

> When in response to a SUBSCRIBE, an Object MUST be sent according to its `Object Forwarding
Preference`.

draft-ietf-moq-transport.md
yekuiwang: OK, in that case, it is fine. But in the draft there is no definition of "subscription", and in many places "subscription" is used that could easily be interpreted by readers to cover fetches as well. For example, in the very first instance of "subscription" in the abstract: "MOQT allows a producer of media to publish data and have it consumed via subscription by a multiplicity of endpoints." This sentence gives me an impression that it covers all means of consumption of published data, thus it includes fetches. There are many other places like this in the draft. Therefore, I strongly suggest to add a clarification or maybe a definition of "subscription" to mean what you said, such that readers won't have confusions on this aspect.

draft-ietf-moq-transport.md
yekuiwang: OK. Since your emphasized the explicitness is about "relays", maybe it is good to explicitly mention "relays" in that sentence (then the sentence is self-contained even for the very first Object to be sent, as for relays the property has been determined by the OP for all Objects).

draft-ietf-moq-transport.md
afrind: This referenced section in my suggestion defines a subscription: https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#name-subscriptions.  The intro and other sections in the draft might need updating, but I intend for "subscription" to mean "a flow of objects initiated via the SUBSCRIBE or PUBLISH messages".

draft-ietf-moq-transport.md
```diff
@@ -1554,10 +1557,10 @@ Token {
...
 
 
-* Token Alias - a Session-specific integer identifier that references a Token
-  Value. There are separate Alias spaces for the client and server (e.g.: they
-  can each register Alias=1). Once a Token Alias has been registered, it cannot
-  be re-registered by the same sender in the Session without first being
+* Token Alias - a Session-and-endpoint-specific integer identifier that references
+  a Token Value. There are separate Alias spaces for the client and server (e.g.:
+  they can each register Alias=1). Once a Token Alias has been registered, it
+  cannot be re-registered by the same endpoint in the Session without first being
```
```suggestion
* Token Alias - a Session-specific integer identifier that references a Token
  Value. There are separate Alias spaces for the client and server (e.g.: they
  can each register Alias=1). Once a Token Alias has been registered, it cannot
  be re-registered by the same endpoint in the Session without first being
```

draft-ietf-moq-transport.md
afrind, line 3179:
```suggestion
of a Track, the same value MUST be used for all Objects of the Track.
In a subscription, an Object MUST be sent according to its `Object Forwarding
Preference`.
```

draft-ietf-moq-transport.md
yekuiwang: OK, thanks. Let me flag the following in an issue: The intro and other sections in the draft might need updating regarding the use of the word "subscription", which should mean "a flow of objects initiated via the SUBSCRIBE or PUBLISH messages".

draft-ietf-moq-transport.md
afrind: I retained the original sentence without adding Relays for now. Maybe a subsequent PR can move this to the relay section, but I wanted to ensure this was editorial.

yekuiwang: OK, thanks! Let me flag this in another issue to avoid it being easily forgotten.

PR #1239, opened 2025-09-23
Make normative text on subgroups in different streams

fluffy:


fluffy:
I think this is editorial but some people might disagree

PR #1240, opened 2025-09-23
Clarify it is QUIC implementations

fluffy:


draft-ietf-moq-transport.md
```diff
@@ -404,7 +404,7 @@ When a Track's forwarding preference (see {{object-properties}}) is
 description in the remainder of this section does not apply.
 
 Streams offer in-order reliable delivery and the ability to cancel sending
-and retransmission of data. Furthermore, many implementations offer the ability
+and retransmission of data. Furthermore, many QUIC implementations offer the ability
```
afrind: Prioritization is also a WebTransport feature.  Should we say "QUIC and WebTransport implementations"?

PR #1242, opened 2025-09-23
fix language to what will need to be in RFC

fluffy:


draft-ietf-moq-transport.md
```diff
@@ -690,7 +690,7 @@ The first stream opened is a client-initiated bidirectional control stream where
 the endpoints exchange Setup messages ({{message-setup}}), followed by other
 messages defined in {{message}}.
 
-This draft only specifies a single use of bidirectional streams. Objects are
+This specification only specifies a single use of bidirectional streams. Objects are
 sent on unidirectional streams.  Because there are no other uses of
 bidirectional streams, a peer MAY close the session as a `PROTOCOL_VIOLATION` if
 it receives a second bidirectional stream.
```
```suggestion
This specification only specifies a single use of bidirectional streams. Objects
are sent on unidirectional streams.  Because there are no other uses of
bidirectional streams, a peer MAY close the session as a `PROTOCOL_VIOLATION` if
it receives a second bidirectional stream.
```

PR #1260, opened 2025-09-23
Initial proposal warning to not be over zealous in putting in error c…

### LINKED ISSUE #1254: Caching of unauthorized token response for unauthorized fails
Section 9.2.1.1 

A receiver MAY store an error code (eg: UNAUTHORIZED or MALFORMED_AUTH_TOKEN) in place of the Token Type and Token Alias if any future message referencing the Token Alias will result in that error.

This text will store in the cache an UNAUTHORIZED that will in the future might give an AUTHORIZED result. So there is important to be careful when storing UNAUTHORIZED in the token cache. This appear to need a bit more careful wording of the difference for example expired, and validation of other properties that may give valid results in a later re-evaluation. Time and location are two token properties that can give this result.

gloinul:
…odes. Intended to resolve #1254

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1657,7 +1657,7 @@ previously registered tokens without potentially terminating the entire Session.
 A receiver MAY store an error code (eg: `UNAUTHORIZED` or
 `MALFORMED_AUTH_TOKEN`) in place of the Token Type and Token Alias if any future
 message referencing the Token Alias will result in that error. However, it is
-important to no store an error code for a token that might be valid in the
+important to not store an error code for a token that might be valid in the
 future or due to some other property becoming fulfilled which currently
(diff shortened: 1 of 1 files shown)


afrind (approved):

draft-ietf-moq-transport.md
```diff
@@ -1656,9 +1656,11 @@ in the token cache, even if the message fails for other reasons, including
 previously registered tokens without potentially terminating the entire Session.
 A receiver MAY store an error code (eg: `UNAUTHORIZED` or
 `MALFORMED_AUTH_TOKEN`) in place of the Token Type and Token Alias if any future
-message referencing the Token Alias will result in that error. The size of a
-registered cache entry includes the length of the Token Value, regardless of
-whether it is stored.
+message referencing the Token Alias will result in that error. However, it is
+important to no store an error code for a token that might be valid in the
```
```suggestion
important to not store an error code for a token that might be valid in the
```

Issue #1182, opened 2025-08-30
Proposal for REQUEST_OK message


afrind:
I'd like to add a generic REQUEST_OK with the following structure:

```
REQUEST_OK {
  Request ID (i),
  Length (16),
  Number of Parameters (i),
  Parameters (..)
}
```

This would immediately replace PUBLISH_NAMESPACE_OK and SUBSCRIBE_NAMESPACE_OK, while adding an extensibility point (parameters) to those messages.  I would also use REQUEST_OK (along with #1157) to address #1003.

Assuming we reach consensus on moving Content Exists/Largest and Expires to parameters (#1052 #1012), we would also replace TRACK_STATUS_OK with REQUEST_OK, which would solve #1064 and #1144.

SUBSCRIBE_OK, PUBLISH_OK and FETCH_OK seem to have enough specific fields to warrant being specialized messages.

fluffy:
I don't feel strongly about this but I don't see any reason to make this change. It just makes it hard to talk about the protocol and explain it. It will not make the parser smaller or faster in any real way.

afrind:
It reduces the bloat in the document specifying repetitious things, makes the protocol simpler to implement, and helps resolve the three issues listed above.  
Consolidating the error messages (see #1157) has made measurable size reductions in our implementation.
Agree there is no speed win.

martinduke:
I support this, but don't think the win is as big as REQUEST_ERROR.

afrind:
> I support this, but don't think the win is as big as REQUEST_ERROR.

I agree, since REQUEST_ERROR covers everything but REQUEST_OK wouldn't.

yekuiwang:
I also support this proposal for the merits @afrind mentioned above.

sharmafb:
I support this as well.

fluffy:
I feel pretty strongly against this and it goes against the extensibility stuff we agreed to before.If an SUBCRIBE_OK has a required field, the the publish_OK does not, I don't want them encoded as parameters. I think it results in more error in implementations and makes the text of the document much harder to understand. 

I'd like time to discuss this before we make a decision.

afrind:
> I feel pretty strongly against this and it goes against the extensibility stuff we agreed to before.If an SUBCRIBE_OK has a required field, the the publish_OK does not, I don't want them encoded as parameters. 

[...]

PR #1274, opened 2025-09-26
REQUEST_OK message

afrind:
This can be used for simple OK messages without message specific parameters.

Fixes: #1182

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1031,12 +1031,12 @@ part of that namespace.  This includes echoing back PUBLISH or PUBLISH_NAMESPACE
 messages to the endpoint that sent them.  If an endpoint accepts its own
 PUBLISH, this behaves as self-subscription described in {{subscriptions}}.
 
-A publisher MUST send exactly one SUBSCRIBE_NAMESPACE_OK or
+A publisher MUST send exactly one REQUEST_OK or
 REQUEST_ERROR in response to a SUBSCRIBE_NAMESPACE. The subscriber
 SHOULD close the session with a protocol error if it detects receiving more than
 one.
 
-The receiver of a SUBSCRIBE_NAMESPACE_OK or REQUEST_ERROR ought to
+The receiver of a REQUEST_OK or REQUEST_ERROR ought to
 forward the result to the application, so the application can decide which other
 publishers to contact, if any.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1957,6 +1955,28 @@ REQUESTS_BLOCKED Message {
...
+~~~
+REQUEST_OK Message {
+  Type (i) = 0x7,
+  Length (16),
+  Request ID (i),
+  Number of Parameters (i),
+  Parameters (..) ...
+}
+~~~
+{: #moq-transport-pub-ns-ok format title="MOQT REQUEST_OK Message"}
```
```suggestion
{: #moq-transport-request-ok format title="MOQT REQUEST_OK Message"}
```

draft-ietf-moq-transport.md
```diff
@@ -1957,6 +1955,28 @@ REQUESTS_BLOCKED Message {
...
+
+* Request ID: The Request ID of the PUBLISH_NAMESPACE ({{message-pub-ns}}) this
+  message is replying to.
```
```suggestion
* Request ID: The Request ID to which this message is replying.
```

afrind:
> LG, but it feels like we could move more things over to use this generic REQUEST_OK?

After parameterizing the things agreed to last week we can also move PUBLISH_OK.  However:

SUBSCRIBE_OK -- has Track Alias field
FETCH_OK -- has "End of Track" and "End Location"

Per relative agreement on the principle of what's a field vs param, I don't think it makes sense to param those fields though.

ianswett:
> > LG, but it feels like we could move more things over to use this generic REQUEST_OK?
> 
> After parameterizing the things agreed to last week we can also move PUBLISH_OK. However:
> 
> SUBSCRIBE_OK -- has Track Alias field FETCH_OK -- has "End of Track" and "End Location"
> 
> Per relative agreement on the principle of what's a field vs param, I don't think it makes sense to param those fields though.

Could we replace SUBSCRIBE_OK with PUBLISH?

On FETCH_OK, I agree we shouldn't change it for now, but I wonder if we need "End of Track"?  Also, what if the Track ends while you're fetching?

afrind:
> Could we replace SUBSCRIBE_OK with PUBLISH?

It would have the Full Track Name in it, which is weird?

> On FETCH_OK, I agree we shouldn't change it for now, but I wonder if we need "End of Track"? Also, what if the Track ends while you're fetching?

I think you need End of Track to distinguish it from the case where End < Requested End because End > Largest vs End > End of Track.

ianswett:
> > Could we replace SUBSCRIBE_OK with PUBLISH?
> 
> It would have the Full Track Name in it, which is weird?
> 
> > On FETCH_OK, I agree we shouldn't change it for now, but I wonder if we need "End of Track"? Also, what if the Track ends while you're fetching?
> 
> I think you need End of Track to distinguish it from the case where End < Requested End because End > Largest vs End > End of Track.

If the track ends while the Fetch is completing, you wouldn't know the end of the track, but I guess the fetch ends at that Object anyway?  Something feels weird about that, but I understand that's how it works.

afrind:
> If the track ends while the Fetch is completing, you wouldn't know the end of the track, but I guess the fetch ends at that Object anyway? Something feels weird about that, but I understand that's how it works.

So you mean you FETCH to (10,10) and that is the largest object when FETCH_OK is sent but then the track ends?  It seems like to know that (currently) there has to be an END_OF_TRACK object at (10,11) or (11,0).

suhasHere (approved):

yekuiwang (approved):

PR #1275, opened 2025-09-26
Subscribe Update Acknowledgements

afrind:
Fixes: #1003

Depends on #1274

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2184,10 +2184,10 @@ When a subscriber narrows their subscription, it might still receive objects
 outside the new range if the publisher sent them before the update was
 processed.
 
-There is no control message in response to a SUBSCRIBE_UPDATE, because it is
-expected that it will always succeed and the worst outcome is that it is not
-processed promptly and some extra objects from the existing subscription are
-delivered.
+The receiver of a SUBSCRIBE_UPDATE MUST respond with exactly one REQUEST_OK
+or REQUEST_ERROR message indicating if the update was successful.  When an
+update is unsuccessful, the publisher MUST also terminate the subscription with
+PUBLISH_DONE.
 
 Like SUBSCRIBE, End Group MUST be greater than or equal to the Group specified
 in `Start`.


draft-ietf-moq-transport.md
```diff
@@ -2184,10 +2184,10 @@ When a subscriber narrows their subscription, it might still receive objects
...
 processed.
 
-There is no control message in response to a SUBSCRIBE_UPDATE, because it is
-expected that it will always succeed and the worst outcome is that it is not
-processed promptly and some extra objects from the existing subscription are
-delivered.
+The receiver of a SUBSCRIBE_UPDATE MUST respond with exactly one REQUEST_OK
+or REQUEST_ERROR message indicating if the update was successful.  When an
+update is unsuccessful, the publisher MUST also terminate the subscription with
+PUBLISH_DONE.
```
afrind: Do we need a new PUBLISH_DONE error code for SUBSCRIBE_UPDATE failed?

afrind: When we allow the publisher to send SUBSCRIBE_UPDATE, the flow is like this:

```
P->S: SUB_UPDATE
S->P: REQUEST_ERROR
// Does S also send UNSUBSCRIBE here?
P->S: PUBLISH_DONE  // if so, this is unnecessary
```

ianswett: Probably?

ianswett (approved):
This looks good and like what we talked about, but it also feels awkward to both fail the UPDATE and then also have to separately fail the subscription.

afrind:
> This looks good and like what we talked about, but it also feels awkward to both fail the UPDATE and then also have to separately fail the subscription.

It's less about "separately failing the subscription" and more that sending the PUBLISH_DONE is required for any publisher terminated subscription, since it includes the Stream Count.

draft-ietf-moq-transport.md
suhasHere: I think this is fine, but would like to also add "Sender of the REQUEST_ERROR" MUST also clean up associated subscription state.

suhasHere (approved):

ianswett (commented):
Still LGTM

vasilvv (approved):

Issue #1278, opened 2025-09-26
Clarify whether Message Parameters are hop-by-hop or end-to-end


ianswett:
I thought all control messages and the parameters associated with them were hop-by-hop.  If that was true, it'd be possible to use TV for all of them if we wanted and not do an even/odd parameter space algorithm.  I don't have a strong opinion that this would be better, but it would be possible.

@vasilvv  said that Parameters on PUBLISH (and SUBSCRIBE_OK) like MAX_CACHE_DURATION are end-to-end, which was initially surprising, but I agree that MAX_CACHE_DURATION should be preserved down the caching hierarchy.  Though that doesn't inherently mean it's end-to-end.

It seems pretty clear that parameters for most messages should not be blindly propagated upstream (ie: SUBSCRIBE params), but is PUBLISH different?  Or is it the fact that MAX_CACHE_DURATION is part of the base transport spec?

Possibly we're mixing two different things: Parameters on subscriptions and properties of a Track? (ie: Track Extensions)

afrind:
SUBSCRIBE / PUBLISH_OK params are definitely hop-by-hop, because of possible fan-out at the relay.

I agree there's ambiguity in publisher specified parameters like MAX_CACHE_DURATION and DELIVERY_TIMEOUT.  I also think there's value in Track properties/extensions, but #781 / #1176 has been parked.

ianswett:
I think it'd be really nice to separate things that are hop-by-hop from E2E and make all the hop-by-hop things TV and the end-to-end things more like HTTP Headers.  MAX_CACHE_DURATION seems like a track property, since it's only sent by the publisher, but DELIVERY_TIMEOUT does not.

suhasHere:
Isn't DELOBERY_TIMEOUT also original publisher chosen property since tis the only one that understand the timing requirements of the objects ?

afrind:
When sent P->S DELIVERY_TIMEOUT seems like it should be end-to-end but not S->P. That’s even weirder.

afrind:
DYNAMIC_GROUPS in #1060 is another publisher sent parameter that needs to propagate towards the subscriber.

martinduke:
I think they're all hop-by-hop and are not guaranteed to be stable along the entire path, but in practice there is no reason for relays to alter some of these values.

In the case of delivery timeout, a relay *could* actually make it shorter due to memory limitations.

afrind:
Discussed at 10/8 Interim -

[...]

PR #1285, opened 2025-10-06
RFC: Define End-To-End parameters

afrind:
A strawman that could address #1278 

1. Make unknown message params a PROTOCOL_VIOLATION
2. Make all subscriber message params hop-by-hop
3. Define publisher message parameter scope case-by-case:

e2e: 
PUBLISHER_PRIORITY
MAX_CACHE_DURATION
DELIVERY_TIMEOUT
[GROUP_ORDER]
[DYNAMIC_GROUPS]

hbh:
AUTH_TOKEN
EXPIRES
[LARGEST_OBJECT]
[FORWARD]

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -245,32 +245,6 @@ This document uses stream management terms described in {{?RFC9000, Section
 This document uses the conventions detailed in ({{?RFC9000, Section 1.3}})
 when describing the binary encoding.
 
-As a quick reference, the following list provides a non normative summary
-of the parts of RFC9000 field syntax that are used in this specification.
-
-x (L):
-
-: Indicates that x is L bits long.
-
-x (i):
-
-: Indicates that x holds an integer value using the variable-length
-  encoding as described in ({{?RFC9000, Section 16}}).
-
-x (..):
-
-: Indicates that x can be any length including zero bits long.  Values
- in this format always end on a byte boundary.
-
-[x (L)]:
-
-: Indicates that x is optional and has a length of L.
-
-x (L) ...:
-
-: Indicates that x is repeated zero or more times and that each instance
-  has a length of L.
-
 To reduce unnecessary use of bandwidth, variable length integers SHOULD
 be encoded using the least number of bytes possible to represent the
 required value.
@@ -287,8 +261,10 @@ Location {
 ~~~
 {: #moq-location format title="Location structure"}
 
-In this document, the constituent parts of any Location A can be referred to
-using A.Group or A.Object.
+In this document, a Location can be expressed in the form of {GroupID,
+ObjectID}, where GroupID and ObjectID indicate the Group ID and Object ID of the
+Location, respectively.  The constituent parts of any Location A can be referred
+to using A.Group or A.Object.
 
 Location A < Location B if:
 
@@ -301,29 +277,37 @@ pairs in which the key is a variable length integer and the value
 is either a variable length integer or a byte field of arbitrary
 length.
 
+Key-Value-Pairs encode a Type value as a delta from the previous Type value,
+or from 0 if there is no previous Type value. This is efficient on the wire
+and makes it easy to ensure there is only one instance of a type when needed.
+The previous Type value plus the Delta Type MUST NOT be greater than 2^64 - 1.
+If a Delta Type is received that would be too large, the Session MUST be closed
+with a `PROTOCOL_VIOLATION`.
+
 Key-Value-Pair is used in both the data plane and control plane, but
 is optimized for use in the data plane.
 
 ~~~
 Key-Value-Pair {
-  Type (i),
+  Delta Type (i),
   [Length (i),]
   Value (..)
 }
 ~~~
 {: #moq-key-value-pair format title="MOQT Key-Value-Pair"}
 
-* Type: an unsigned integer, encoded as a varint, identifying the
-  type of the value and also the subsequent serialization.
+* Delta Type: an unsigned integer, encoded as a varint, identifying the Type
+  as a delta encoded value from the previous Type, if any. The Type identifies
+  the type of value and also the subsequent serialization.
 * Length: Only present when Type is odd. Specifies the length of the Value field
   in bytes. The maximum length of a value is 2^16-1 bytes.  If an endpoint
   receives a length larger than the maximum, it MUST close the session with a
-  Protocol Violation.
+  `PROTOCOL_VIOLATION`.
 * Value: A single varint encoded value when Type is even, otherwise a
   sequence of Length bytes.
 
 If a receiver understands a Type, and the following Value or Length/Value does
-not match the serialization defined by that Type, the receiver MUST terminate
+not match the serialization defined by that Type, the receiver MUST close
 the session with error code `KEY_VALUE_FORMATTING_ERROR`.
 
 ### Reason Phrase Structure {#reason-phrase}
@@ -390,6 +374,23 @@ payloads.
 
 Objects within a Group are in ascending order by Object ID.
 
+From the perspective of a subscriber or a cache, an Object can be in three
+possible states:
+
+1. The Object is known to not exist. This state is permanent. MOQT has multiple
+   ways to communicate that a certain range of objects does not exist,
+   including the Object Status field, and the use of gaps in FETCH responses.
+2. The Object is known to exist. From this state, it can transition to not
+   existing, but not vice versa.
+3. The state of the Object is unknown, either because it has not been yet
+   received, or it has not been produced yet.
+
+Whenever the publisher communicates that certain objects do not exist, this
+fact is expressed as a contiguous range of non-existent objects and
+by include extension headers indicating the group/object gaps; MOQT
+implementers should take that into account when selecting appropriate data
+structures.
+
 ## Subgroups {#model-subgroup}
 
 A subgroup is a sequence of one or more objects from the same group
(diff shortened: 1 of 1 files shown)


fluffy:
I think we should allow unknown messages that are passed between end subscriber and publisher but that seems a different issue.  Perhaps we need a separate issue for this part of it.

draft-ietf-moq-transport.md
```diff
@@ -1561,6 +1560,16 @@ messages and for all other message types. SETUP message parameter types
...
 
+### Parameter Scope
+
+Message Parameters sent by the subscriber are always intended for the peer
+endpoint and are not forwarded by Relays.  Some message parameters sent by the
+publisher are intended for the peer endpoint and are not forwarded by Relays,
+while others relate to the Track or subscription and are considered End-to-End,
+as noted in the parameter definition.  End-to-End Parameters sent in PUBLISH,
+SUBSCRIBE_OK, TRACK_STATUS_OK or FETCH_OK MUST be cached by Relays and included
+in future downstream subscriptions or fetches for that Track.
```
michalhosna, line 1711: Are the parameters immutable for the lifetime of a given track? 

If parameters are mutable, how should the caching work? Only the last one is valid? What should existing connections see? 

What about multi-publishers? Must all of them send the same values for e2e params?

fluffy (changes requested):
I think we need to get our extensibility story  agreed on before we can do this part of it.

draft-ietf-moq-transport.md
```diff
@@ -1575,24 +1575,23 @@ for the peer, or a new request with a Request ID that is not the next in
...
 `INVALID_REQUEST_ID`.
 
-## Parameters {#params}
+## Message Parameters {#params}
 
-Some messages include a Parameters field that encodes optional message
-elements.
+Some messages include a Parameters field that encodes optional message elements.
+All parameters included in a message MUST be defined in the negotiated version
+of MOQT or negotiated via Setup Parameters.  An endpoint that receives an
```
ianswett, line 1678: There's one wrinkle I realize, which is that we are using Parameters in Setup to negotiate which Parameters are supported, but we need to allow for unknown Parameters in Setup.

draft-ietf-moq-transport.md
```diff
@@ -1575,24 +1575,23 @@ for the peer, or a new request with a Request ID that is not the next in
...
+unknown message parameters MUST close the session with `PROTOCOL_VIOLATION`.
```
ianswett, line 1679:
```suggestion
unknown Parameter MUST close the session with `PROTOCOL_VIOLATION`.
```

draft-ietf-moq-transport.md
```diff
@@ -1605,6 +1604,16 @@ messages and for all other message types. SETUP message parameter types
 are defined in {{setup-params}}. Version-specific parameter types are defined
 in {{version-specific-params}}.
 
+### Parameter Scope
+
+Message Parameters sent by the subscriber are always intended for the peer
```
ianswett, line 1705: nit: New Group Request is sort of forwarded?

draft-ietf-moq-transport.md
```diff
@@ -1893,6 +1923,9 @@ to the Largest Group, it does not send a NEW_GROUP_REQUEST upstream.
 After sending a NEW_GROUP_REQUEST upstream, the request is considered
 outstanding until the Largest Group increases.
 
+When sent by a publisher, this parameter is not End-To-End (see
```
ianswett: Isn't this sent by a Subscriber?

draft-ietf-moq-transport.md
```diff
@@ -1773,6 +1788,9 @@ multi-object stream will expire earlier than Objects later in the stream. Once
 Objects have expired from cache, their state becomes unknown, and a relay that
 handles a downstream request that includes those Objects re-requests them.
 
+When sent by a publisher, this parameter is End-To-End (see
```
ianswett: Nit: This is correct, but given the Publisher is the only one who sends it, how about:
```suggestion
MAX CACHE DURATION is only sent by a publisher, and is End-To-End (see
```

draft-ietf-moq-transport.md
```diff
@@ -1605,6 +1604,16 @@ messages and for all other message types. SETUP message parameter types
 are defined in {{setup-params}}. Version-specific parameter types are defined
 in {{version-specific-params}}.
 
+### Parameter Scope
+
+Message Parameters sent by the subscriber are always intended for the peer
+endpoint and are not forwarded by Relays.  Some message parameters sent by the
+publisher are intended for the peer endpoint and are not forwarded by Relays,
```
ianswett, line 1707:
```suggestion
publisher are scoped to the Session and are not forwarded by Relays,
```

draft-ietf-moq-transport.md
afrind: Hmm I think I was trying to distinguish "message" params from "setup" params, but might need to be more crisp.

draft-ietf-moq-transport.md
afrind: I disagree.  NGR is an instruction to the terminating endpoint that could result in another NGR going upstream, but it is not "forwarded" in the same way that objects are.  It's subtle.

draft-ietf-moq-transport.md
afrind: With #1374 I assume _all_ publisher **parameters** are scoped to a message, while "Extension Headers" (eg: properties) are end-to-end.

draft-ietf-moq-transport.md
afrind: Oh this is a good question.  Can a publisher send REQUEST_UPDATE and change either parameters or extensions/track properties, or the properties are immutable?

draft-ietf-moq-transport.md
```diff
@@ -1893,6 +1923,9 @@ to the Largest Group, it does not send a NEW_GROUP_REQUEST upstream.
 After sending a NEW_GROUP_REQUEST upstream, the request is considered
 outstanding until the Largest Group increases.
 
+When sent by a publisher, this parameter is not End-To-End (see
+{{parameter-scope}}).
+
```
```suggestion
```

draft-ietf-moq-transport.md
ianswett: Alternatively, we could decide that Setup Messages use Extension Headers of some sort, but that feels odd somehow.

draft-ietf-moq-transport.md
ianswett: It seems like an edge case, which is one of the reasons I resisted it.  But I think it makes your above statement incorrect?

afrind: I see NEW_GROUP_REQUEST as intended for the first hop.  If that hop is a relay, it takes that as input to it's own algorithm, which sometimes emits another NGR upstream.

fluffy: Alan's way of thinking about this makes sense to me. We likely will have other things with specific aggregation handling.

draft-ietf-moq-transport.md
```diff
@@ -1605,6 +1604,16 @@ messages and for all other message types. SETUP message parameter types
 are defined in {{setup-params}}. Version-specific parameter types are defined
 in {{version-specific-params}}.
 
+### Parameter Scope
+
+Message Parameters sent by the subscriber are always intended for the peer
+endpoint and are not forwarded by Relays.  Some message parameters sent by the
+publisher are intended for the peer endpoint and are not forwarded by Relays,
+while others relate to the Track or subscription and are considered End-to-End,
```
fluffy, line 1708: I don't love End-to-end as the name here. The point of these is you do not have to negotiation support for them in the setup and unknown ones are allowed. That is going to make the text really hard for the non end to end parameters.  The issue is they only flow one way and end-to-end might suggest they can flow both ways.  I did like separating them totally from parameters (other than same code point space) and calling them properties. I think that makes it easier to understand how they are different. Another option might be "track properties" but then I don't want to have to refer to the other ones as "non track propties" everywhere we use them.

draft-ietf-moq-transport.md
fluffy: Ian and I were proposing that if you get a new one if a message, it overrides whatever was in cache before for that property. And when downstreams requests, it gets latest in cache with no upstream check to see if the cache i fresh.

draft-ietf-moq-transport.md
```diff
@@ -1812,6 +1833,9 @@ SUBSCRIBE_OK or TRACK_STATUS_OK is used. If omitted in PUBLISH_OK, the
 publisher's preference from PUBLISH is used. If omitted from SUBSCRIBE_OK,
 TRACK_STATUS_OK, PUBLISH or FETCH, the receiver uses Ascending (0x1).
 
+When sent by a publisher, this parameter is End-To-End (see
+{{parameter-scope}}).
+
```
fluffy, line 1951: This just high lights that we should have two separate things of subscriber group order and publisher group order.

draft-ietf-moq-transport.md
```diff
@@ -1854,6 +1881,9 @@ Relays MUST preserve the value of this parameter received from an upstream
 publisher in SUBSCRIBE_OK or PUBLISH when sending these messages to downstream
 subscribers.
 
+When sent by a publisher, this parameter is End-To-End (see
```
fluffy: Is there any time this is not sent by a publisher ?

draft-ietf-moq-transport.md
```diff
@@ -1784,6 +1802,9 @@ subscription inherit this priority, unless they specifically override it.
 
 The subscription has Publisher Priorty 128 if this parameter is omitted.
 
+When sent by a publisher, this parameter is End-To-End (see
```
fluffy: I think this is always send by publisher. If that is correct we should change this text.

draft-ietf-moq-transport.md
afrind: I think mutability makes sense for some of these but not others, so it might be on a case-by-case what you are allowed to change (with unknown=allow).  I do worry a little bit about how that intersects our immutable Track model and what the edge cases might be.

draft-ietf-moq-transport.md
afrind: We tried spitting it that way and the PR was incredibly awkward, so we let it roll with the current spelling.  Perhaps as this group of extensibility PRs comes into focus we can take another swing at splitting this.

draft-ietf-moq-transport.md
afrind: I agree it will be easier to explain and conceptualize if we don't call everything "parameters" and have this distinction between end-to-end and not.

draft-ietf-moq-transport.md
ianswett: SGTM, happy to keep this

draft-ietf-moq-transport.md
ianswett: I'd still love to get rid of Group Order, but that's TBD.

afrind:
I think we should land #1374 first, then rework this one to use Track Extensions instead of "end-to-end" parameters.

afrind:
See replacement #1390

PR #1286, opened 2025-10-07
clarify joining fetch end location

### LINKED ISSUE #1252: Fetch end location and start location of subscribe
The "largest object" is undefined in section 9.17 of -14. This needs to get clarified to say that the subscribe saves what the largest object was when the subscribe started, then later joining fetch use that number that was saved as the end location of the fetch.  The current text is pretty unclear on details of this.

suhasHere:
Fixes #1252

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -854,7 +854,7 @@ processing rules described in {{relays-moq}}.
 
 # Publishing and Retrieving Tracks
 
-## Subscriptions
+## Subscriptions {#subscriptions}
 
 A subscription can be initiated by either a publisher or a subscriber.  A
 publisher initiates a subscription to a track by sending the PUBLISH message.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2555,11 +2555,12 @@ Joining Fetch {
 #### Joining Fetch Range Calculation
 
 The Largest Location value from the corresponding
-subscription is used to calculate the end of a Joining Fetch so the
+subscription, is used to calculate the end of a Joining Fetch so the
```
```suggestion
subscription is used to calculate the end of a Joining Fetch, so the
```

draft-ietf-moq-transport.md
```diff
@@ -2555,11 +2555,12 @@ Joining Fetch {
...
 Objects retrieved by the FETCH and SUBSCRIBE are contiguous and non-overlapping.
 
 The publisher receiving a Joining Fetch sets the End Location to {Subscribe
-Largest Location.Object + 1}.
+Largest Location.Object + 1}. Here Subscribe Largest Location is
+saved value when the subscription started.
```
```suggestion
Largest Location.Object + 1}. Here Subscribe Largest Location is the
saved value from when the subscription started.
```

ianswett (commented):
Some small suggestions

afrind:
Since this places a requirement on SUBSCRIBE, I think we should mention that requirement either in the Subscriptions or SUBSCRIBE section too, maybe something like:

> A publisher MUST save the Largest Location communicated in PUBLISH or SUBSCRIBE_OK when establishing a subscription.  This value can be used in a Joining FETCH (see {{joining-fetches}}) at any time while the subscription is active.

PR #1287, opened 2025-10-07
Clarify what happens when max_cache_duration parameter is omitted

### LINKED ISSUE #1263: How long can I cache objects in a track with no Max Cache Duration?
This is currently optional and there is no specified default.

The HTTP interpretation is "forever until it is evicted"

suhasHere:
Fixes #1263

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1729,9 +1729,8 @@ multi-object stream will expire earlier than Objects later in the stream. Once
 Objects have expired from cache, their state becomes unknown, and a relay that
 handles a downstream request that includes those Objects re-requests them.
 
-If the MAX_CACHE_DURATION parameter is not present in a message, the Objects
-never expires until implementation constraints cause them to be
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1721,14 +1721,18 @@ any other relevant information.
...
 relay MUST NOT start forwarding any individual Object received through this
 subscription or fetch after the specified number of milliseconds has elapsed
 since the beginning of the Object was received.  This means Objects earlier in a
 multi-object stream will expire earlier than Objects later in the stream. Once
 Objects have expired from cache, their state becomes unknown, and a relay that
 handles a downstream request that includes those Objects re-requests them.
 
+If the MAX_CACHE_DURATION parameter is not present in a message, the Objects
+never expires until implementation constraints cause them to be
+evicted from the cache.
```
```suggestion
If the MAX_CACHE_DURATION parameter is not sent by the publisher, the Objects
can be cached until implementation constraints cause them to be evicted.
```

PR #1273, opened 2025-09-25
Parameters for Group Order, Subscribe Priority and Subscription Filter (redo)

afrind:
Note: this is a redo of #1210, targeted against main

This is a proposal based on the most recent discussion of parameters and extensibility.  

It moves Group Order and Subscriber Priority to parameters in SUBSCRIBE, SUBSCRIBE_UPDATE, PUBLISH, PUBLISH_OK.  The PR specifies a value for the receiver to use if the parameter is omitted.

It defines a structure for a Subscription Filter as it exists today.  
  * The wire format for SUBSCRIBE and PUBLISH_OK use this directly (not as a param, no wire format change)
  * SUBSCRIBE_UPDATE is changed to accept a parameter containing this filter when the subscriber wants to change it

Items of note:

1. This adds a new filter type "No Filter", addressing #1035.
2. Relative filter types are only allowed in SUBSCRIBE, because they require the SUBSCRIBE_OK containing Largest to communicate what actually happened, but maybe this could be relaxed?
3. The values used by Group Order are different depending on the message.  This preserves existing behavior while eliminating the "Default" wire value previously allowed in SUBSCRIBE.
4. Forward is retained as a Fixed field, rather than in all messages where it appears.  This is based on the fact that there's no sensible "common" value.
5. Other than Forward, everything that can be changed in SUBSCRIBE_UPDATE is now a parameter, and omitted means "unchanged".

Fixes: #872 
Fixes: #878 
Fixes: #1035 
Closes: #1043
Closes: #1044 
Closes: #1059

PR #1288, opened 2025-10-07
Allow subscribe_update to increase the end location

### LINKED ISSUE #770: Not being able to have multiple subscriptions is a challenge for ABR

suhasHere:
Fixes #770

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2192,15 +2192,15 @@ session with a `PROTOCOL_VIOLATION` ({{session-termination}}).
 ## SUBSCRIBE_UPDATE {#message-subscribe-update}
 
 A subscriber sends a SUBSCRIBE_UPDATE to a publisher to modify an existing
-subscription. Subscriptions can only be narrowed, not widened, as an attempt to
-widen could fail. If Objects with Locations smaller than the current
+subscription. The Start Location MUST not decrease, as an attempt to
+to do so could fail. If Objects with Locations smaller than the current
 subscription's Start Location are required, FETCH can be used to retrieve
-them. The Start Location MUST NOT decrease and the End Group MUST NOT increase.
-A publisher MUST terminate the session with a `PROTOCOL_VIOLATION` if the
+them. A publisher MUST terminate the session with a `PROTOCOL_VIOLATION` if the
 SUBSCRIBE_UPDATE violates these rules or if the subscriber specifies a request
 ID that has not existed within the Session.
 
-When a subscriber narrows their subscription, it might still receive objects
+When a subscriber narrows their subscription (increase the Start Location and/or
+decrease the End Group), it might still receive objects
 outside the new range if the publisher sent them before the update was
 processed.
 


draft-ietf-moq-transport.md
```diff
@@ -2192,15 +2192,15 @@ session with a `PROTOCOL_VIOLATION` ({{session-termination}}).
...
 A subscriber sends a SUBSCRIBE_UPDATE to a publisher to modify an existing
-subscription. Subscriptions can only be narrowed, not widened, as an attempt to
-widen could fail. If Objects with Locations smaller than the current
+subscription. The Start Location MUST not decrease, as an attempt to
+to do so could fail. If Objects with Locations smaller than the current
 subscription's Start Location are required, FETCH can be used to retrieve
-them. The Start Location MUST NOT decrease and the End Group MUST NOT increase.
-A publisher MUST terminate the session with a `PROTOCOL_VIOLATION` if the
+them. A publisher MUST terminate the session with a `PROTOCOL_VIOLATION` if the
 SUBSCRIBE_UPDATE violates these rules or if the subscriber specifies a request
```
yekuiwang: Now there is only one rule left from the above description.

```suggestion
SUBSCRIBE_UPDATE violates this rule or if the subscriber specifies a request

File: draft-ietf-moq-transport.md [hunk 2]
```diff
@@ -2192,15 +2192,15 @@ session with a `PROTOCOL_VIOLATION` ({{session-termination}}).
...
-SUBSCRIBE_UPDATE violates these rules or if the subscriber specifies a request
+them. A publisher MUST terminate the session with a `PROTOCOL_VIOLATION` if the
+SUBSCRIBE_UPDATE violates this rule or if the subscriber specifies a request
 ID that has not existed within the Session.
 
-When a subscriber narrows their subscription, it might still receive objects
+When a subscriber narrows their subscription (increase the Start Location and/or
+decrease the End Group), it might still receive objects
 outside the new range if the publisher sent them before the update was
 processed.
```
```suggestion
When a subscriber narrows their subscription (increase the Start Location and/or
decrease the End Group), it might still receive objects
outside the new range if the publisher sent them before the update was
processed. When a subscriber increases the End Location, the Largest Object
might be larger than the old End Location, in which case some Objects would
need to be fetched.
```

suhasHere: I like the suggestion.

afrind:
This is going to conflict with #1273.  Can we land that and rebase this?

afrind:
@suhasHere can you resolve conflicts with trunk?

draft-ietf-moq-transport.md
```diff
@@ -2281,16 +2281,20 @@ session with a `PROTOCOL_VIOLATION` ({{session-termination}}).
...
+SUBSCRIBE_UPDATE violates this rule or if the subscriber specifies a request
+ID that has not existed within the Session.
+
+When a subscriber narrows their subscription (increase the Start Location and/or
+decrease the End Group), it might still receive objects
 outside the new range if the publisher sent them before the update was
-processed.
+processed. When a subscriber increases the End Location, the Largest Object
+might be larger than the old End Location, in which case some Objects would
+need to be fetched.
```
afrind: It's not clear to me what you are trying to say.  Is it that extending the End could create a gap that is not easily detectable? e.g.:

```
S->P: SUBSCRIBE start=10,0, end=10,5
P->S: SUB_OK
Publish 10,0 - 10,10; 10,6 -> 10,10 are not sent to S
S->P: SUBSCRIBE_UPDATE End=10,20
```

If so, maybe something like this:

When a subscriber increases the End Location, the Largest Object at the publisher might already be larger than the previous End Location.  This will create a gap in the subscription.  The REQUEST_OK in response to the SUBSCRIBE_UPDATE will include the LARGEST_OBJECT parameter, and the subscriber can issue a FETCH to retrieve the omitted Objects, if any.

draft-ietf-moq-transport.md
```diff
@@ -2281,16 +2281,20 @@ session with a `PROTOCOL_VIOLATION` ({{session-termination}}).
...
-widened, as an attempt to widen could fail. If Objects with Locations smaller
-than the current subscription filter's Start Location are required, FETCH can be
-used to retrieve them. The Start Location MUST NOT decrease and the End Group
-MUST NOT increase.  A publisher MUST terminate the session with a
-`PROTOCOL_VIOLATION` if the SUBSCRIBE_UPDATE violates these rules.
-
-When a subscriber narrows their subscription, it might still receive objects
+When updating the Subscription Filter (see {{subscription-filters}}),
+the Start Location MUST not decrease, as an attempt to
+to do so could fail. If Objects with Locations smaller than the current
```
afrind, line 2286: I don't know that "it could fail" is a sufficient justification.  Should we explain in more detail why this is disallowed?

draft-ietf-moq-transport.md
```diff
@@ -2281,16 +2281,20 @@ session with a `PROTOCOL_VIOLATION` ({{session-termination}}).
...
+subscription's Start Location are required, FETCH can be used to retrieve
+them. A publisher MUST terminate the session with a `PROTOCOL_VIOLATION` if the
+SUBSCRIBE_UPDATE violates this rule or if the subscriber specifies a request
+ID that has not existed within the Session.
```
afrind: This requirement (valid ID) is already on line 2332, so I don't think we need it here?  Let's keep this section just on the semantics of the range update.

```suggestion
SUBSCRIBE_UPDATE violates this rule.
```

draft-ietf-moq-transport.md
suhasHere: I like the suggestion. @ianswett  thoughts ?

vasilvv (approved):

draft-ietf-moq-transport.md
```diff
@@ -2281,16 +2281,19 @@ session with a `PROTOCOL_VIOLATION` ({{session-termination}}).
...
+them. A publisher MUST terminate the session with a `PROTOCOL_VIOLATION` if the
+SUBSCRIBE_UPDATE violates this rule.
+
+When a subscriber narrows their subscription (increase the Start Location and/or
+decrease the End Group), it might still receive objects
 outside the new range if the publisher sent them before the update was
-processed.
+processed. When a subscriber increases the End Location, the Largest Object
+might be larger than the old End Location, in which case some Objects would
+need to be fetched.
```
```suggestion
processed. When a subscriber increases the End Location, the Largest Object at
the publisher might already be larger than the previous End Location. This will
create a gap in the subscription. The REQUEST_OK in response to the
SUBSCRIBE_UPDATE will include the LARGEST_OBJECT parameter, and the subscriber
can issue a FETCH to retrieve the omitted Objects, if any.
```

draft-ietf-moq-transport.md
afrind: Let's handle this in an editorial PR

afrind (approved):

draft-ietf-moq-transport.md
```diff
@@ -2281,16 +2281,21 @@ session with a `PROTOCOL_VIOLATION` ({{session-termination}}).
...
-MUST NOT increase.  A publisher MUST terminate the session with a
-`PROTOCOL_VIOLATION` if the SUBSCRIBE_UPDATE violates these rules.
-
-When a subscriber narrows their subscription, it might still receive objects
+When updating the Subscription Filter (see {{subscription-filters}}),
+the Start Location MUST not decrease, as an attempt to
+to do so could fail. If Objects with Locations smaller than the current
+subscription's Start Location are required, FETCH can be used to retrieve
+them. A publisher MUST terminate the session with a `PROTOCOL_VIOLATION` if the
+SUBSCRIBE_UPDATE violates this rule.
```
afrind, line 2289:
```suggestion
to do so could fail. If Objects with Locations smaller than the current
subscription's Largest Location are required, FETCH can be used to retrieve
them. A publisher MUST terminate the session with a `PROTOCOL_VIOLATION` if the
SUBSCRIBE_UPDATE violates this rule.
```

Issue #1144, opened 2025-08-06
Is TRACK_STATUS_OK with track_alias > 0 an error?


martinduke:
There is no normative language here, so it's unclear if anything can be enforced.

afrind:
If we keep this "set it to 0" logic then it should probably be normative.  I think there might be some value in communicating a real Track Alias - particularly for publisher/subscriber systems that use fixed Track Alias.  I'm not sure how critical it is.

martinduke:
I have no opinion about this except that the spec be explicit. We can say "the publisher puts whatever it wants here" and have no enforcement, or require it is zero and kill the session, or say the publisher SHOULD put in the actual track alias if it has one.

yekuiwang:
Saying "the publisher SHOULD put in the actual track alias if it has one" seems making a lot of sense; even MUST seems good.

afrind:
Might as well make it MUST if we care at all.

PR #1291, opened 2025-10-07
Mandate track alias value in TRACK_STATUS_OK

suhasHere:
Addresses #1144

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2757,8 +2757,9 @@ The TRACK_STATUS_OK message format is identical to the SUBSCRIBE_OK message
 ({{message-subscribe-ok}}).
 
 The publisher populates the fields of TRACK_STATUS_OK exactly as it would have
-populated a SUBSCRIBE_OK, setting Track Alias to 0. It is not considered an error
-if Track Alias 0 is already in use by an active subscription.
+populated a SUBSCRIBE_OK. The Track Alias MUST be set to 0. It is not
+considered an error if Track Alias 0 is already in use by an active
+subscription.
 
 ## PUBLISH_NAMESPACE {#message-pub-ns}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2756,10 +2756,11 @@ to a successful TRACK_STATUS message.
 The TRACK_STATUS_OK message format is identical to the SUBSCRIBE_OK message
 ({{message-subscribe-ok}}).
 
-The publisher populates the fields of TRACK_STATUS_OK exactly as it would have
-populated a SUBSCRIBE_OK. The Track Alias MUST be set to 0. It is not
+The publisher populates the fields of TRACK_STATUS_OK in the same way as
+SUBSCRIBE_OK. The Track Alias MUST be set to 0. It is not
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2756,9 +2756,11 @@ to a successful TRACK_STATUS message.
 The TRACK_STATUS_OK message format is identical to the SUBSCRIBE_OK message
 ({{message-subscribe-ok}}).
 
-The publisher populates the fields of TRACK_STATUS_OK exactly as it would have
-populated a SUBSCRIBE_OK, setting Track Alias to 0. It is not considered an error
-if Track Alias 0 is already in use by an active subscription.
+The publisher populates the fields of TRACK_STATUS_OK in the same way as
+SUBSCRIBE_OK. The Track Alias MUST be set to 0. It is not
```
yekuiwang, line 2760:
```suggestion
for SUBSCRIBE_OK. The Track Alias MUST be set to 0. It is not

Comment (suhasHere) [reply draft-ietf-moq-transport.md:2760]: @yekuiwang  I think the original commit text seems more appropriate since we want to highlight that the logic behind populating fields mimcs SUBSCRIBE_OK

Comment (yekuiwang) [reply draft-ietf-moq-transport.md:2760]: Yes, I understand. I just feel that English grammar wise, 

> The publisher populates the fields of TRACK_STATUS_OK in the same way as for SUBSCRIBE_OK.

is better than

> The publisher populates the fields of TRACK_STATUS_OK in the same way as SUBSCRIBE_OK.

As the former seems to be a better shortcut for "The publisher populates the fields of TRACK_STATUS_OK in the same way as the publisher would populate the fields of SUBSCRIBE_OK".

But, maybe I am wrong, and in any case, feel free to decide. Thanks!

**Comment by afrind** [comment]
> Hopefully we'll finally remove TRACK_STATUS, but this text LGTM

There's no plan to remove TRACK_STATUS.  

I think we can replace TRACK_STATUS_OK (SUBSCRIBE_OK) with REQUEST_OK (#1274), which will also fix this.  But for now, this is good.

File: draft-ietf-moq-transport.md (hunk 1 above)
Comment (afrind) [reply draft-ietf-moq-transport.md:2760]: I would keep the original text here about how to formulate the message and combine with your new text that the receiver MUST close if it gets non-zero.

File: draft-ietf-moq-transport.md [hunk 2]
```diff
@@ -2756,9 +2756,11 @@ to a successful TRACK_STATUS message.
...
+considered an error if Track Alias 0 is already in use by an active
+subscription. However, the subscriber MUST terminate the session with
+a `PROTOCOL_VIOLATION` if the Track Alias is not set to 0.
```
afrind, line 2763:
```suggestion
The publisher populates the fields of TRACK_STATUS_OK exactly as it would have
populated a SUBSCRIBE_OK, setting Track Alias to 0. It is not considered an
error if Track Alias 0 is already in use by an active subscription. The
subscriber MUST terminate the session with a `PROTOCOL_VIOLATION` if it receives
a non-zero Track Alias.
```

afrind:
#1294 solves this a different way, by killing TRACK_STATUS_OK completely.

afrind:
Planning to do this with #1294

Issue #996, opened 2025-05-05
On repetition of the AUTHORIZATION TOKEN parameter


yekuiwang:
Sec. 8.2.1.1 includes the following sentence:

> The AUTHORIZATION TOKEN parameter MAY be repeated within a message.

So is it allowed to be repeated within a message with exactly the same parameter value? If so, what is the purpose of such repetition? Or is the intent really to say that multiple instances of the AUTHORIZATION TOKEN parameter with different parameter values are allowed within a message?

afrind:
I think the same token type + token value in the same request, or the same alias in the same request, should probably be an error.

For sure you can have multiple tokens with different types, and possibly multiple tokens with the same type.

I might need @wilaw or @fluffy to weigh in how how they think the tokens will be populated in multi-CDN requests and the use-cases for all of those.

yekuiwang:
Trying to progress this one. The exactly same sentence is still in draft-13.

How about changing the sentence from 

> The AUTHORIZATION TOKEN parameter MAY be repeated within a message.

to

> The AUTHORIZATION TOKEN parameter MAY be repeated within a message as long as the multiple instances of the AUTHORIZATION TOKEN parameter have different values.

ianswett:
In Toronto:

Allowing two of the same token type is fine.  People thought repeating it was a bad idea, so MAY close the connection if you notice, MUST NOT send duplicates.

afrind:
>  People thought repeating it was a bad idea

You mean repeating the same "Token Type, Token Value pair in a single parameter set"?

PR #1293, opened 2025-10-07
Define duplication rules for header extensions

### LINKED ISSUE #1121: Parameters are only one per Message by default, Objects are unstated
Parameters clearly state that only one can occur in a Message unless otherwise stated:

> Senders MUST NOT repeat the same parameter type in a message unless the parameter definition explicitly allows multiple instances of that type to be sent in a single message.

Object Extension Headers don't have any similar text, which I forgot about.  I find it confusing that we have a clear default in one case and not the other.  Also, I expect most Object extension headers to be limited to 1 per Object, so making that the default is sensible.

suhasHere:
Fixes  #1121

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3707,6 +3707,8 @@ y = e2e Encrypted Data
 EXT 1 and EXT N can be modified or removed by Relays
 ~~~
 
+An Object MUST NOT contain more than one instance of this extension header.
+
 ## Prior Object ID Gap
 
 Prior Object ID Gap (Extension Header Type 0x3E) is a variable length integer


draft-ietf-moq-transport.md
```diff
@@ -3080,9 +3080,10 @@ of the Object payload and not as an extension header.
 
 Extension Headers are defined in external specifications and registered in an
 IANA table {{iana}}. These specifications define the type and value of the
-header, along with any rules concerning processing, modification, caching and
-forwarding. A relay which is coded to implement these rules is said to
-"support" the extension.
+header, as well as the rules for duplication, processing,
+modification, caching and forwarding. Unless explicitly restricted by the
+extension's specification, extension headers are duplicable by default.
```
afrind: "duplication" as a verb has at least two meanings.  I think it's ok to leave that out of the first sentence, and consider it part of "processing".

The new sentence then might flow better as a standalone after this paragraph.

'duplicable' is not a very common word - so maybe:

Unless explicitly restricted by the extension's specification, a extension header can be repeated multiple times in the same Object.

draft-ietf-moq-transport.md
```diff
@@ -3585,8 +3586,8 @@ there may not be a gap. If Prior Group ID Gap is not present, the receiver
 cannot infer any information about the existence of prior groups (see
 {{group-ids}}).
 
-This extension can be added by the Original Publisher, but MUST NOT be added by
-relays. This extension MUST NOT be modified or removed.
+Only a single extension of this type can be added by the Original Publisher,
```
afrind: line 3575 already has this restriction.  I think it might be best to have this as a separate statement if we put it at the end:

> An Object MUST NOT contain more than one instance of this extension header.

Or if it's allowed for the extension:

> An Object MAY contain more than one instance of this extension header.

draft-ietf-moq-transport.md
```diff
@@ -3655,8 +3656,8 @@ there may not be a gap. If Prior Object ID Gap is not present, the receiver
 cannot infer any information about the existence of prior objects (see
 {{model-object}}).
 
-This extension can be added by the Original Publisher, but MUST NOT be added by
-relays. This extension MUST NOT be modified or removed.
+Only a single extension of this type can be added by the Original Publisher,
```
afrind: line 3647 already has this restriction.

Also I think you missed it for IMMUTABLE_EXTENSIONS.

draft-ietf-moq-transport.md
```diff
@@ -3080,9 +3080,10 @@ of the Object payload and not as an extension header.
...
 IANA table {{iana}}. These specifications define the type and value of the
-header, along with any rules concerning processing, modification, caching and
-forwarding. A relay which is coded to implement these rules is said to
-"support" the extension.
+header, as well as the rules for duplication, processing,
+modification, caching and forwarding. Unless explicitly restricted by the
+extension's specification, extension headers are duplicable by default.
+A relay designed to enforce these rules is considered to "support" the extension.
 
 If unsupported by the relay, Extension Headers MUST NOT be modified, MUST be
```
afrind, line 3129: Would this now read better in the singular, like:

> If an Extension Header is not supported by the Relay, it MUST NOT be modified, etc

draft-ietf-moq-transport.md
ianswett: For params, we took the approach that only one could occur unless the param said otherwise.  I think that's a better default, partially because it takes cognitive load to realize the default for one is only one and the default for the other allows for multiple.

afrind: So you are agreeing that this PR is correct but you want to flip the default for params too?

yekuiwang: What is the purpose of duplicating or repeating a Parameter multiple times with the same value in an Object? As discussed in #996 (fixed by #1276), at least for the AUTHORIZATION TOKEN parameter, simple duplication or repetition with exactly the same value should not be allowed. Similar rule should be applied to other Parameters as well?

suhasHere: May be it's not clear. This issue was supporting adding multiple extensions of same type. Duplication is misleading as Alan pointed

yekuiwang: Sorry, I put my comment in a wrong place. What Alan suggested above is:

> Unless explicitly restricted by the extension's specification, a extension header can be repeated multiple times in the same Object.

With that wording, my comment above still applies.

afrind: > Unless explicitly restricted by the extension's specification, a extension header can be repeated multiple times in the same Object with different values.

?

suhasHere: This works for me

suhasHere:
@afrind @ianswett  how should we proceed here ?

draft-ietf-moq-transport.md
yekuiwang: > This works for me

+1

afrind:
> @afrind @ianswett how should we proceed here ?

@suhasHere - can you resolve the conflicts with trunk and integrate the suggestions here?

draft-ietf-moq-transport.md
```diff
@@ -3108,9 +3108,10 @@ be terminated with a `PROTOCOL_VIOLATION` ({{session-termination}}).
 Any object with a status code other than zero MUST have an empty payload.
 
 #### Object Extension Header {#object-extensions}
-Any Object with status Normal can have extension headers.  If an endpoint
-receives extension headers on Objects with status that is not Normal, it MUST close the
-session with a `PROTOCOL_VIOLATION`.
+
+Any Object MAY have extension headers except those with Object Status 'Object
+Does Not Exist'.  If an endpoint receives a non-existent Object containing
+extension headers it MUST close the session with a `PROTOCOL_VIOLATION`.
```
afrind: The merge blew away the previous text that forbid extensions on non-normal (not just DNE).

```suggestion
Any Object with status Normal can have extension headers.  If an endpoint
receives extension headers on Objects with status that is not Normal, it MUST close the
session with a `PROTOCOL_VIOLATION`.
```

draft-ietf-moq-transport.md
ianswett: Params today are limited to 1 by default.  I'd like that to be true for header extensions as well.

draft-ietf-moq-transport.md
```diff
@@ -3119,9 +3120,10 @@ of the Object payload and not as an extension header.
 
 Extension Headers are defined in external specifications and registered in an
 IANA table {{iana}}. These specifications define the type and value of the
-header, along with any rules concerning processing, modification, caching and
-forwarding. A relay which is coded to implement these rules is said to
-"support" the extension.
+header, as well as the rules for duplication, processing,
+modification, caching and forwarding. Unless explicitly restricted by the
+extension's specification, extension headers are duplicable by default.
+A relay designed to enforce these rules is considered to "support" the extension.
```
```suggestion
header, as well as the rules for processing, modification, caching and
forwarding. All such specifications MUST specify whether multiple values of the
same extension are allowed on a single Object.  A relay that enforces these
rules is considered to "support" the extension.  If a Relay does not support an
extension header, it MUST assume multiple values are allowed.
```

draft-ietf-moq-transport.md
```diff
@@ -3656,8 +3659,8 @@ there may not be a gap. If Prior Group ID Gap is not present, the receiver
 cannot infer any information about the existence of prior groups (see
 {{group-ids}}).
 
-This extension can be added by the Original Publisher, but MUST NOT be added by
-relays. This extension MUST NOT be modified or removed.
+Only a single extension of this type can be added by the Original Publisher,
+but MUST NOT be added by relays. This extension MUST NOT be modified or removed.
```
```suggestion
This extension can be added by the Original Publisher, but MUST NOT be added by
relays. This extension MUST NOT be modified or removed.

An Object MUST NOT contain more than one instance of this extension header.
```

draft-ietf-moq-transport.md
```diff
@@ -3726,8 +3729,8 @@ there may not be a gap. If Prior Object ID Gap is not present, the receiver
 cannot infer any information about the existence of prior objects (see
 {{model-object}}).
 
-This extension can be added by the Original Publisher, but MUST NOT be added by
-relays. This extension MUST NOT be modified or removed.
+Only a single extension of this type can be added by the Original Publisher,
+but MUST NOT be added by relays. This extension MUST NOT be modified or removed.
```
```suggestion
This extension can be added by the Original Publisher, but MUST NOT be added by
relays. This extension MUST NOT be modified or removed.

An Object MUST NOT contain more than one instance of this extension header.
```

afrind (approved):

Issue #1233, opened 2025-09-22
Semantic FIN and Filters / Forward


vasilvv:
Currently, we use FIN to indicate that a subgroup stream has been completed; this FIN has a semantic meaning (relays have to forward it).

How does this interact with filters?  If I have a filter that stops at `(15, 10)`, but the subgroup goes on, the publisher is expected to send a FIN after `(15, 10)`, which is not where the original publisher would put it.

How does this interact with forward?  Assume I have a really long group.  If I send Forward=0, does the publisher has to cleanly fin the subgroup stream, reset it or what?  What happens if I set Forward=1, and that subgroup is still open?  Or if it's been closed so long ago that the relay evicted all the information about it from the cache?

afrind:
Presently you cannot set an end object in a filter, only an end group. I think this issue might have been one of the motivations?

If we land a proposal like #1164 and the last object in a subgroup is omitted, I think the answer is - you wait for an ack and reset, or send reset_stream_at.

afrind:
I missed the second part of your issue

> How does this interact with forward? Assume I have a really long group. If I send Forward=0, does the publisher has to cleanly fin the subgroup stream, reset it or what? What happens if I set Forward=1, and that subgroup is still open?

Relays/publishers don't have to immediately terminate open streams when Forward goes from 0->1.  The only normative rules are you can't send a FIN if you didn't send the final object in the subgroup, and a relay MUST NOT forward an Object on an existing Subgroup stream unless it is the *next* Object in that Subgroup.  But if Forward=0 causes an object to be missed - the logical thing to do is reset it.  If Forward=0 and the end of subgroup is reached _and_ the final object has been sent, I think it's fine to FIN the stream.

> Or if it's been closed so long ago that the relay evicted all the information about it from the cache?

I think the same logic applies.

Are you asking for editorial clarification or a design change?

vasilvv:
I am not sure if it's editorial or design, but we definitely should describe what happens to existing streams when Forward changes.

afrind:
I think we can address this with an Editorial PR.

PR #1295, opened 2025-10-09
Omitting a subgroup object necessitates reset

afrind:
Fixes: #1233

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3411,7 +3411,7 @@ stream, it MUST use a RESET_STREAM or RESET_STREAM_AT
 {{!I-D.draft-ietf-quic-reliable-stream-reset}} frame. This includes, but is
 not limited to:
 
-* An open Subgroup exceeding its Delivery Timeout
+* An Object in an open Subgroup exceeding its Delivery Timeout
 * Early termination of subscription due to an UNSUBSCRIBE message
 * A publisher's decision to end the subscription early
 * A SUBSCRIBE_UPDATE moving the subscription's End Group to a smaller Group or


draft-ietf-moq-transport.md
```diff
@@ -3408,11 +3408,17 @@ MUST close the stream with a FIN.
...
 stream, it MUST use a RESET_STREAM or RESET_STREAM_AT
-{{!I-D.draft-ietf-quic-reliable-stream-reset}} frame. This includes an open
-Subgroup exceeding its Delivery Timeout, early termination of subscription due
-to an UNSUBSCRIBE message, a publisher's decision to end the subscription early,
-or a SUBSCRIBE_UPDATE moving the subscription's End Group to a smaller Group or
-the Start Location to a larger Location.  When RESET_STREAM_AT is used, the
+{{!I-D.draft-ietf-quic-reliable-stream-reset}} frame. This includes, but is
+not limited to:
+
+* An open Subgroup exceeding its Delivery Timeout
```
```suggestion
* An Object in an open Subgroup exceeding its Delivery Timeout
```

draft-ietf-moq-transport.md
```diff
@@ -3408,11 +3408,17 @@ MUST close the stream with a FIN.
...
+* Early termination of subscription due to an UNSUBSCRIBE message
+* A publisher's decision to end the subscription early
+* A SUBSCRIBE_UPDATE moving the subscription's End Group to a smaller Group or
+  the Start Location to a larger Location
+* Omitting a Subgroup Object due to the subcriber's Forward State
```
suhasHere, line 3419: I think we also need to add Subscribe update changing forward state too ?

afrind: Changing the forward state by itself does not necessitate closing streams _unless_ it causes an object to be missed

ianswett: If you the publisher opened a stream, presumably the forward state at that time was true, so does it make sense to say "A change in the Forward state from true to false"?

```suggestion
* A change in the Forward state from true to false
```

afrind: You made basically the same comment as Suhas so I have the same reply.  Simply changing the forward state is not what necessitates closing the stream. If an object is published every 1s, I can flip forward off and on again between objects and retain the stream.  The problem only happens when I go to publish an object on that subgroup and forward=0 --> then you have to reset.

Of course, your implementation can choose to reset upon receiving Forward=0 too, but nothing breaks if you leave it open longer.

suhasHere: May be what I was intending to capture is 
If Forward state moves from 0 to 1 and there is a gap in the next object Id, then the stream needs to be reset

afrind: > If Forward state moves from 0 to 1 and there is a gap in the next object Id, then the stream needs to be reset

I tried to capture it with this:

> Omitting a Subgroup Object due to the subcriber's Forward State

Issue #1251, opened 2025-09-23
Location math not defined


fluffy:
Section 9.16.2.1 of -14 uses math on locations but we don't define that. I think it would be better to rewrite this not to need math on locations

PR #1298, opened 2025-10-09
Remove Location math

afrind:
Use prose to explain how to set Locations

Fixes: #1251

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2651,10 +2651,10 @@ The Largest Location value from the corresponding
 subscription is used to calculate the end of a Joining Fetch so the
 Objects retrieved by the FETCH and SUBSCRIBE are contiguous and non-overlapping.
 
-The publisher receiving a Joining Fetch sets the End Location Group to 
+The publisher receiving a Joining Fetch sets the End Location Group to
 `Largest Location.Group` and End Location Object to `Largest Location.Object` plus
 one. `Largest Location` refers to the value sent in the `PUBLISH` or `SUBSCRIBE_OK`
-message used to establish the subscription being joined..
+message used to establish the subscription being joined.
 
 Note: the last Object included in the Joining FETCH response is Subscribe
 Largest Location.  The "plus one" above indicates the equivalent Standalone Fetch


draft-ietf-moq-transport.md
```diff
@@ -2649,18 +2651,20 @@ The Largest Location value from the corresponding
 subscription is used to calculate the end of a Joining Fetch so the
 Objects retrieved by the FETCH and SUBSCRIBE are contiguous and non-overlapping.
 
-The publisher receiving a Joining Fetch sets the End Location to {Subscribe
-Largest Location.Object + 1}.
+The publisher receiving a Joining Fetch sets the End Location Group to
```
```suggestion
The publisher receiving a Joining Fetch sets the End Location Group to
```

draft-ietf-moq-transport.md
```diff
@@ -2649,18 +2651,20 @@ The Largest Location value from the corresponding
 subscription is used to calculate the end of a Joining Fetch so the
 Objects retrieved by the FETCH and SUBSCRIBE are contiguous and non-overlapping.
 
-The publisher receiving a Joining Fetch sets the End Location to {Subscribe
-Largest Location.Object + 1}.
+The publisher receiving a Joining Fetch sets the End Location Group to 
+`Largest Location.Group` and End Location Object to `Largest Location.Object` plus
+one. `Largest Location` refers to the value sent in the `PUBLISH` or `SUBSCRIBE_OK`
+message used to establish the subscription being joined..
```
```suggestion
message used to establish the subscription being joined.
```

ianswett:
I think this looks ok, but I believe I'm also happy with the current text, so I'm not that motivated to change it.

yekuiwang:
> I think this looks ok, but I believe I'm also happy with the current text, so I'm not that motivated to change it.

Same feeling

afrind:
@ianswett @yekuiwang @fluffy : other thoughts on how to resolve 1251 besides this?

draft-ietf-moq-transport.md
```diff
@@ -972,16 +972,18 @@ Subscription Filter {
 
 Filter Type can have one of the following values:
 
-Largest Object (0x2): The filter Start Location is `{Largest Object.Group,
-Largest Object.Object + 1}` and `Largest Object` is communicated in
+Largest Object (0x2): The filter Start Location Group is `Largest Object.Group`
+and the Start Location Object is the `Largest Object.Object` plus one;
+`Largest Object` is communicated in
```
suhasHere, line 977: this will conflict with LARGEST_OBJECT parameter PR .. we may need to update it so

yekuiwang:
> @ianswett @yekuiwang @fluffy : other thoughts on how to resolve 1251 besides this?

To me, the current text in this aspect is more compact and cleaner. If that compact expression needs to be defined, we can define it.  To me, it would be sufficient to simply add something like the following:

> A Location can be expressed in the form of {grpId, objId}, where grpId and objId indicate the Group and Object, respectively, of the Location.

afrind:
Closing in favor of #1319

PR #1315, opened 2025-10-16
Delta Encode Key-Value-Pairs for Params and Headers

### LINKED ISSUE #748: Extension (and parameter) ordering
I wouldn't expect the order of either extension headers or parameters to matter, but there is some simplification by having them in a fixed order.  We could delta encode the type varint, which would force them to be in ascending order by extension/param type and save a bit of space.  This could also eliminate the possibility of duplicates if we chose.

ianswett:
Fixes #748

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1765,7 +1765,7 @@ aliases.
 #### DELIVERY TIMEOUT Parameter {#delivery-timeout}
 
 The DELIVERY TIMEOUT parameter (Parameter Type 0x02) MAY appear in a
-TRACK_STATUS, REQUEST_OK (in response to TRACK_STATUS), PUBLISH, PUBLISH_OK,
+REQUEST_OK (in response to TRACK_STATUS), PUBLISH, PUBLISH_OK,
 SUBSCRIBE, SUBSCRIBE_OK, or SUBSCRIBE_UDPATE message.
 
 It is the duration in milliseconds the relay SHOULD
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -273,22 +273,25 @@ Location A < Location B if:
 Key-Value-Pair is a flexible structure designed to carry key/value
 pairs in which the key is a variable length integer and the value
 is either a variable length integer or a byte field of arbitrary
-length.
+length. Key-Value-Pairs encode a Type value as a delta from the
+previous Type value. If there is no previous Type value, it's an
+absolute Type.
```
```suggestion
previous Type value, or from 0 if there is no previous Type value.
```

draft-ietf-moq-transport.md
```diff
@@ -1584,7 +1587,10 @@ Receivers ignore unrecognized parameters.
 The number of parameters in a message is not specifically limited, but the
 total length of a control message is limited to 2^16-1 bytes.
 
-Parameters are serialized as Key-Value-Pairs {{moq-key-value-pair}}.
+Parameters are serialized as Key-Value-Pairs {{moq-key-value-pair}} in
```
afrind, line 1631: Do you need to say the same about extensions?

ianswett: Good point, I added a short clause there as well.

vasilvv:
Mostly LGTM, but you should mention that the resulting ID should not exceed a certain max value.

draft-ietf-moq-transport.md
```diff
@@ -1623,7 +1625,10 @@ Receivers ignore unrecognized parameters.
...
+increasing Parameter ID order with a delta encoding. This is efficient on the
+wire and makes it easy to ensure there is only one instance of parameters
+that cannot be repeated.
```
```suggestion
that cannot be repeated. The previous Type value plus the Delta Type MUST NOT be
greater than 2^62 - 1.  If a Delta Type is received that would be too large, the Session
SHOULD be closed with a Protocol Violation.
```

ianswett:
> Mostly LGTM, but you should mention that the resulting ID should not exceed a certain max value.

Good point, I added text on that: https://github.com/moq-wg/moq-transport/pull/1315/commits/addc07d6d11bf649793f34dbacda386c5699e59e

draft-ietf-moq-transport.md
```diff
@@ -1626,7 +1628,12 @@ Receivers ignore unrecognized parameters.
 The number of parameters in a message is not specifically limited, but the
 total length of a control message is limited to 2^16-1 bytes.
 
-Parameters are serialized as Key-Value-Pairs {{moq-key-value-pair}}.
+Parameters are serialized as Key-Value-Pairs {{moq-key-value-pair}} in
+increasing Parameter ID order with a delta encoding. This is efficient on the
+wire and makes it easy to ensure there is only one instance of parameters
+that cannot be repeated. The previous Type value plus the Delta Type MUST NOT be
+greater than 2^62 - 1.  If a Delta Type is received that would be too large, the Session
+SHOULD be closed with a Protocol Violation.
```
michalhosna:
```suggestion
MUST be closed with a `PROTOCOL_VIOLATION`.
```
I do think this should be enforced. I think the rest of the draft always have 'MUST' for any encoding related protocol violations. Why be more permissive here? 

There seems to be inconsistency between "Protocol Violation" and `PROTOCOL_VIOLATION`, the latter seems to be used in vast majority.

michalhosna: Is there a reason to have this limit only for parameters and not for extensions?
I.e. why not move this to the Key-Value-Pair section?

michalhosna (changes requested):
I am not sure why this change is split between the Parameters and Extensions Header sections. I think all the change text can and should be only in the Key-Value-Pair section.
 
Sorry for the late review. Feel free to ignore me.

draft-ietf-moq-transport.md
```diff
@@ -1626,7 +1628,12 @@ Receivers ignore unrecognized parameters.
 The number of parameters in a message is not specifically limited, but the
 total length of a control message is limited to 2^16-1 bytes.
 
-Parameters are serialized as Key-Value-Pairs {{moq-key-value-pair}}.
+Parameters are serialized as Key-Value-Pairs {{moq-key-value-pair}} in
+increasing Parameter ID order with a delta encoding. This is efficient on the
+wire and makes it easy to ensure there is only one instance of parameters
+that cannot be repeated. The previous Type value plus the Delta Type MUST NOT be
+greater than 2^62 - 1.  If a Delta Type is received that would be too large, the Session
```
```suggestion
greater than 2^64 - 1.  If a Delta Type is received that would be too large, the Session
```

draft-ietf-moq-transport.md
ianswett: MUST SGTM

Issue #1309, opened 2025-10-13
A single Subgroup stream (not a single Subgroup) has a single publisher priority


yekuiwang:
In Section 7 of the editor's copy, we have:

> A single subgroup or datagram has a single publisher priority. 

However, we know that Objects in a fetch response can have publisher priority present on Object level thus Objects of the same Subgroup could have different publisher priorities.

It is therefore suggested to change the above sentence to be as follows:

> A single Subgroup stream or datagram has a single publisher priority.

afrind:
A fetch response can encode a different publisher priority per each object in a single subgroup, but such a subgroup might be malformed?  In response to Subscribe, you can only have one priority per subgroup.

We thought about preventing this encoding in FETCH compression PR, but decided it was too much complexity because of receiver state.

So I think the original text is in the right spirit, and we can make it an error to specify different priorities in a fetch response subgroup.

yekuiwang:
OK, I will make a PR for this. 

BTW, I just checked which error code should be used, and figured out it should be "MALFORMED_TRACK", and when I searched for "MALFORMED_TRACK" in the editor's copy, I noticed that are two different error codes, code values 0x12 and 0x07, with the same name of "MALFORMED_TRACK". Is it really needed to define two different error codes for malformed track? If yes, won't it be better to give them different names to be clearer and to avoid possible confusion?

yekuiwang:
BTW again, if all Objects of each Subgroup in a fetch stream must have the same publisher priority, it would be so much more efficient to signal the publisher priority values in the fetch header, one value for each Subgroup. Would that be a very bad idea?

afrind:
> I noticed that are two different error codes, code values 0x12 and 0x07, with the same name of "MALFORMED_TRACK".

Ah yes, one goes in FETCH_ERROR and another goes in PUBLISH_DONE.  Maybe there should even be a third for resetting the fetch data stream when it's malformed?

Yes we should either give them the same value or different names.

> BTW again, if all Objects of each Subgroup in a fetch stream must have the same publisher priority, it would be so much more efficient to signal the publisher priority values in the fetch header, one value for each Subgroup. Would that be a very bad idea?

[...]

PR #1317, opened 2025-10-17
Malformed track due to multiple priorities for one subgroup and use o…

yekuiwang:
Fixes: #1309

This PR includes changes for the two following purposes:
1. Add the definition of malformed track due to detection of multiple priorities for one subgroup in a FETCH response.
2. Use one error code value for the 'MALFORMED_TRACK' error code (currently two values are used).

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -513,9 +513,9 @@ include:
 1. An Object is received in a FETCH response with the same Group ID as the
    previous Object, but whose Object ID is not strictly larger than the previous
    object.
-2. An Object is received in a FETCH response with the same Subgroup ID as a
-   previous Object in the same FETCH response, but whose Publisher Priority
-   is different from the Publisher Priority of that previous Object.
+2.  In a FETCH response, an Object with a particular Subgroup ID is received, but its
+     Publisher Priority is different from that of the previous Object with the same
+     Subgroup ID.
 3. An Object is received in an Ascending FETCH response whose Group ID is smaller
    than the previous Object in the response.
 4. An Object is received in a Descending FETCH response whose Group ID is larger


draft-ietf-moq-transport.md
```diff
@@ -2165,8 +2168,10 @@ INVALID_RANGE (0x11):
 cannot be satisfied.
 
 MALFORMED_TRACK (0x12):
-: In response to a FETCH, a relay publisher detected
-the track was malformed (see {{malformed-tracks}}).
+: In response to a FETCH, a relay publisher detected the track was
+malformed (see {{malformed-tracks}}). This error code may also
+be used when sending a PUBLISH_DONE message (see {{message-publish-done}})
+or  resetting a fetch stream upon the detection of a malformed track.
```
```suggestion
or resetting a fetch stream upon the detection of a malformed track.

Comment (afrind) [inline draft-ietf-moq-transport.md]: data stream reset codes have their own section 13.2.4

File: draft-ietf-moq-transport.md [hunk 2]
```diff
@@ -3862,6 +3865,7 @@ TODO: register the URI scheme and the ALPN and grease the Extension types
 | CANCELLED        | 0x1  | {{closing-subgroup-streams}} |
 | DELIVERY_TIMEOUT | 0x2  | {{closing-subgroup-streams}} |
 | SESSION_CLOSED   | 0x3  | {{closing-subgroup-streams}} |
+| MALFORMED_TRACK  | 0x12 | {{malformed-tracks}}         |
```
afrind: Can you add text about this to section 10.4.3 "closing subgroup streams" and update the reference here?

yekuiwang: Done. Please check whether it is good now. 

BTW, section 10.4.3 is titled "closing subgroup streams", but it seems that it also covers closing fetch streams. Should the title be updated at least?

yekuiwang: > BTW, section 10.4.3 is titled "closing subgroup streams", but it seems that it also covers closing fetch streams. Should the title be updated at least?

@afrind Have you noticed the above question?

afrind: Yes we should reorganize or retitle or something.  You can file an editorial issue or PR.

draft-ietf-moq-transport.md
```diff
@@ -510,29 +510,32 @@ to MOQT constraints. Such a Track is considered malformed.  Some example
 conditions that constitute a malformed track when detected by a receiver
 include:
 
-1. An Object is received in a FETCH response with the same Group as the
-   previous Object, but whose Object ID is not strictly larger than the previous
-   object.
-2. An Object is received in an Ascending FETCH response whose Group ID is smaller
+1. An Object is received in a FETCH response with the same Group ID as the
+   previous Object in the same FETCH response, but whose Object ID is not
+   larger than that of the previous Object.
```
suhasHere: i feel the original text was clear on this one. May be just s/Group/Group ID

draft-ietf-moq-transport.md
```diff
@@ -510,29 +510,32 @@ to MOQT constraints. Such a Track is considered malformed.  Some example
...
+2. An Object is received in a FETCH response with the same Subgroup ID as a
+   previous Object in the same FETCH response, but whose Publisher Priority
+   is different from the Publisher Priority of that previous Object.
```
```suggestion
2. An Object is received in a FETCH response with the same Subgroup ID as the
   previous Object , but whose Publisher Priority
   is different from that of previous Object.
```

draft-ietf-moq-transport.md
```diff
@@ -2236,8 +2239,8 @@ INVALID_RANGE (0x11):
 cannot be satisfied.
 
 MALFORMED_TRACK (0x12):
-: In response to a FETCH, a relay publisher detected
-the track was malformed (see {{malformed-tracks}}).
+: In response to a FETCH, a relay publisher detected that the track was
+malformed (see {{malformed-tracks}}).
```
suhasHere, line 2243: is there a change here ?

draft-ietf-moq-transport.md
yekuiwang: OK

draft-ietf-moq-transport.md
yekuiwang: Changing from "a previous Object" to "the previous Object" herein is not OK, as that Object being talked about herein can be earlier than "the previous Object", which might have a different Subgroup ID. What about the following:

> In a FETCH response, an Object with a particular Subgroup ID is received, but its Publisher Priority is different from that of the previous Object with the same Subgroup ID.

draft-ietf-moq-transport.md
yekuiwang: Yes, but only added "that".

yekuiwang:
> Few nits to simplify the text

Thanks for the review. I responded to your comments/questions. Will make corresponding changes if you are OK with my suggested rewording on the one with "a previous" or "the previous" or if you have yet a different suggestion.

draft-ietf-moq-transport.md
```suggestion
1. An Object is received in a FETCH response with the same Group ID as the
   previous Object, but whose Object ID is not strictly larger than the previous
   object.

File: draft-ietf-moq-transport.md [hunk 6]
```diff
@@ -510,29 +510,32 @@ to MOQT constraints. Such a Track is considered malformed.  Some example
...
 include:
 
-1. An Object is received in a FETCH response with the same Group as the
+1. An Object is received in a FETCH response with the same Group ID as the
    previous Object, but whose Object ID is not strictly larger than the previous
    object.
-2. An Object is received in an Ascending FETCH response whose Group ID is smaller
+2. An Object is received in a FETCH response with the same Subgroup ID as a
+   previous Object in the same FETCH response, but whose Publisher Priority
+   is different from the Publisher Priority of that previous Object.
```
```suggestion
2.  In a FETCH response, an Object with a particular Subgroup ID is received, but its
     Publisher Priority is different from that of the previous Object with the same
     Subgroup ID.

File: draft-ietf-moq-transport.md [hunk 7]
```diff
@@ -510,29 +510,32 @@ to MOQT constraints. Such a Track is considered malformed.  Some example
...
 
-1. An Object is received in a FETCH response with the same Group as the
+1. An Object is received in a FETCH response with the same Group ID as the
    previous Object, but whose Object ID is not strictly larger than the previous
    object.
-2. An Object is received in an Ascending FETCH response whose Group ID is smaller
+2.  In a FETCH response, an Object with a particular Subgroup ID is received, but its
+     Publisher Priority is different from that of the previous Object with the same
+     Subgroup ID.
+3. An Object is received in an Ascending FETCH response whose Group ID is smaller
```
suhasHere, line 519: This can happen due to re-ordering between group boundaries/streams. Not sure, if it is easy to enforce?

draft-ietf-moq-transport.md
```diff
@@ -510,29 +510,32 @@ to MOQT constraints. Such a Track is considered malformed.  Some example
 conditions that constitute a malformed track when detected by a receiver
 include:
```
suhasHere, line 511: Looks like some of these are true only for Streams. Should we make it explicit?

draft-ietf-moq-transport.md
afrind: This is for FETCH

draft-ietf-moq-transport.md
```diff
@@ -510,29 +510,32 @@ to MOQT constraints. Such a Track is considered malformed.  Some example
...
 
-1. An Object is received in a FETCH response with the same Group as the
+1. An Object is received in a FETCH response with the same Group ID as the
```
ianswett, line 513: **Editorial Suggestion:**
As discussed, I think it'd be nice to separate this into two lists in the future: One that applies to both SUBSCRIBE and FETCH, and one that only applies to FETCH.

PR #1192, opened 2025-09-03
Some editorial changes to v14

yekuiwang:
This PR suggests some editorial changes, including a fix for the remaining item in #536 (i.e., removal of the new field types (b) and (tuple)) and fixes for #1186, #1187, and #1190, and some changes that are very minor and obvious, to v14 of the MOQT draft.

Fixes: #1186
Fixes: #1187
Fixes: #1190 
Fixes: #536

PR #1320, opened 2025-10-20
draft-15 release notes

afrind:


yekuiwang:
@afrind PR #1192, which includes removal of the new field types (b) and (tuple), might worth mentioning.

yekuiwang:
> @afrind PR #1192, which includes removal of the new field types (b) and (tuple), might worth mentioning.

Never mind, I just noticed that -v15 has already been submitted.

PR #1322, opened 2025-10-20
Allow Start Location to decrease in SUBSCRIBE_UPDATE

### LINKED ISSUE #1314: Inability to move a Subscription's Start Location forward and only having one Subscription to a Track causes problems
If one has a Subscription starting in the future (ie: I asked to start at Group 7 and the current Group is 5), and then I decide I want to start at Largest Object, that can't be done with SUBSCRIBE_UPDATE today.  Additionally, one can't just issue a second subscription for the gap between Largest Object and the current start group (ie: Group 7) because you can only have one active Subscription for a Track at once.

I noticed this when reading #1288 

Related to #770

ianswett:
Allow it to decrease as small as Largest Location, similar to a new subscription.

Fixes #1314


ianswett:
Closing this PR because I accidentally deleted SUBSCRIBE_UPDATE in a bunch of spots

PR #1323, opened 2025-10-20
Allow Start Location to decrease in SUBSCRIBE_UPDATE

### LINKED ISSUE #1314: Inability to move a Subscription's Start Location forward and only having one Subscription to a Track causes problems

ianswett:
Allow it to decrease as small as Largest Location, similar to a new subscription.

Fixes #1314

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -275,8 +275,14 @@ Location A < Location B if:
 Key-Value-Pair is a flexible structure designed to carry key/value
 pairs in which the key is a variable length integer and the value
 is either a variable length integer or a byte field of arbitrary
-length. Key-Value-Pairs encode a Type value as a delta from the
-previous Type value, or from 0 if there is no previous Type value.
+length.
+
+Key-Value-Pairs encode a Type value as a delta from the previous Type value,
+or from 0 if there is no previous Type value. This is efficient on the wire
+and makes it easy to ensure there is only one instance of a type when needed.
+The previous Type value plus the Delta Type MUST NOT be greater than 2^64 - 1.
+If a Delta Type is received that would be too large, the Session MUST be closed
+with a `PROTOCOL_VIOLATION`.
 
 Key-Value-Pair is used in both the data plane and control plane, but
 is optimized for use in the data plane.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2339,22 +2339,23 @@ SUBSCRIBE_OK Message {
...
-When updating the Subscription Filter (see {{subscription-filters}}),
-the Start Location MUST not decrease, as an attempt to
-to do so could fail. If Objects with Locations smaller than the current
-subscription's Largest Location are required, FETCH can be used to retrieve
-them. A publisher MUST terminate the session with a `PROTOCOL_VIOLATION` if the
-SUBSCRIBE_UPDATE violates this rule.
+When a subscriber decreases the Start Location of the Subscription Filter
+(see {{subscription-filters}}), if the Start Location is less than the Track's
+Largest Location, the Start Location is increased to Largest Location,
+similar to SUBSCRIBE. If Objects with Locations smaller than the current
```
afrind: This is no longer true.  1) we moved this to Subscription Filters. 2) we removed the "increase to largest object" behavior from Absolute* filters.

I think you just want to allow this with the relative filter types?  Do you have a use case for absolute?

ianswett: The use case for Absolute is that I'm trying to decrease it to a specific Group and due to timing, an Object from that Group was published in the meantime.  I'd rather have the 'increase to largest object' behavior than failure.

afrind: My recollection of why we removed clamping start to largest was to allow Absolute filters starting at 0 to mean "anything".  Since there's now a "No Filter" option _maybe_ you can add this back?  But you need to update the text here in any case.

ianswett: But re-reading the existing text, it looks like it doesn't get increased, to largest object, it just gets set to the specified absolute value and if new Objects are produced that match the filter, they flow through.  I think that's fine, but I need to update this PR if so.

afrind (changes requested):
You need to update this

ianswett:
> You need to update this

Updated to address your comment.

afrind:
It looks like you now support arbitrary decrease in start location, instead or clamping it to Largest Location.  Maybe that's ok?  We were very worried about this before but now it doesn't seem as scary.

ianswett:
> It looks like you now support arbitrary decrease in start location, instead or clamping it to Largest Location. Maybe that's ok? We were very worried about this before but now it doesn't seem as scary.

Yes, if a Subscription filter really is a filter, then I think it seems reasonable to allow decreasing it to any value?  But I can go either way on this one.  I'm not sure I know enough to have a strong opinion.

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -2354,22 +2354,22 @@ SUBSCRIBE_OK Message {
...
+When a subscriber increases the End Location, the Largest Object at
 the publisher might already be larger than the previous End Location. This will
 create a gap in the subscription. The REQUEST_OK in response to the
 SUBSCRIBE_UPDATE will include the LARGEST_OBJECT parameter, and the subscriber
 can issue a FETCH to retrieve the omitted Objects, if any.
 
+When a subscriber narrows their subscription (increase the Start Location and/or
+decrease the End Group), it might still receive objects
+outside the new range if the publisher sent them before the update was
+processed.
```
```suggestion
When a subscriber narrows their subscription (increase the Start Location and/or
decrease the End Group), it might still receive Objects outside the
new range if the publisher sent them before the update was processed.
```

draft-ietf-moq-transport.md
```diff
@@ -2354,22 +2354,21 @@ SUBSCRIBE_OK Message {
...
 subscription.
 
-When updating the Subscription Filter (see {{subscription-filters}}),
-the Start Location MUST not decrease, as an attempt to
-to do so could fail. If Objects with Locations smaller than the current
-subscription's Largest Location are required, FETCH can be used to retrieve
-them. A publisher MUST close the session with a `PROTOCOL_VIOLATION` if the
-SUBSCRIBE_UPDATE violates this rule.
+When a subscriber decreases the Start Location of the Subscription Filter
+(see {{subscription-filters}}), the Start Location can be less the Track's
```
```suggestion
(see {{subscription-filters}}), the Start Location can be smaller than the Track's

afrind:
Discussed in 11/17 Interim:

This change is ok to proceed.

Issue #1064, opened 2025-07-07
TRACK_STATUS[_OK] has useless fields


martinduke:
I guess my review was 2 minutes too late for -13, so I filed a new issue.

I do not think there is anything to be gained by making TRACK_STATUS and TRACK_STATUS_OK use identical formats to SUBSCRIBE and SUBSCRIBE_OK, so these are just wasted bits.

If there is a master plan to eventually fold everything back into REQUEST and REQUEST_OK, or parameterize the fields that are wasted here, then I will happily close the issue.

martinduke:
Moreover, while in the PR discussion Alan articulated rules for how to fill and interpret these fields, none of those rules made it into the draft.

afrind:
>  master plan to eventually fold everything back into REQUEST and REQUEST_OK

I don't think there's a clearly articulated master plan for that.  @ianswett posted to the mailing list outlining proposed principles for determining if a value should be a field or a parameter.  We haven't yet had a discussion or confirmation that the working group agrees with those principles.  I think we've reached a point now where all the _ERROR messages have the same format (Type, Length, Request ID, Error Code, Reason Phrase).

> parameterize the fields that are wasted here

We have to see how the parameterization discussion shakes out.  I don't see all the fields as wasted, since eg: you can set Group Order to "Default" in order to learn the publisher's Group Order preference. Forward and Filter Type don't currently give you any information.  Track Alias in the response is also not helpful.

afrind:
Comment from @fluffy in #1255 

> Having an used and invalid info will just cause confusion

suhasHere:
#1281 has same or similar concerns :-)

afrind:
Comments from 1281:

> PR https://github.com/moq-wg/moq-transport/pull/1273 opened up few discussions on if Priority, GroupOrder and Filter is applicable to Track Status message. This stems from symmetry of message encoding between Subscribe Message and Track Status message. Moving to parameters makes it more obvious ( may be not ?)

> We can probably resolve in 2 ways

> Define TrackStatus to be not symmetrical to Subscribe and make it very focussed on its purpose
> Explicitly define what parameters are applicable for TrackStatus in the spec, if we decide to move with symmetry design approach

afrind:
I think with #1273 and #1274, this can be easily fixed.

1. SUBSCRIBE's only fixed field will be Full Track Name, which is also what is needed for TRACK_STATUS.
[...]

PR #1325, opened 2025-10-21
Remove delivery related params from TRACK_STATUS for Subscribers

afrind:
Fixes: #1064


afrind:
> Q: Do we want to disallow these? I can imagine someone wanting to know what the values are for when they do SUBSCRIBE?

This PR only removes the subscriber side params, things like DELIVERY_TIMEOUT and PUBLISHER_PRIORITY can still come in the REQUEST_OK response to TRACK_STATUS.

ianswett (approved):
I missed that this only changes it for Subscribers, so LGTM

Issue #1265, opened 2025-09-24
How to serve a partial-hit FETCH when the original publisher is unavailable


afrind:
If a cache has some objects in a FETCH range cached, but a range of objects with unknown status, and the original publisher is offline/unavailable, the cache must stop serving the FETCH at the first unknown object.  The subscriber has no way to know how many objects are unknown, so doesn't have an efficient way to skip over the gap.

FETCH_OK may have already been sent, so can't necessarily send a FETCH_ERROR when this happens
RESET_STREAM only carries an error code and can't encode any information about the unknown range
The absence of the OP may be either temporary or permanent

We could have an Object Status like "unknown" or "unknown range", but we just tentatively agreed to remove serving object status in FETCH responses.

afrind:
Related to #945 - at least the proposed solution is the same.

afrind:
Talking to Victor - one idea is to use a flag in the serialization byte to signal a "Trailer" that gives you additional information about the FETCH in the data plane.

Issue #1057, opened 2025-06-25
Is knowing the status of all Objects in the Fetch practical?


ianswett:
Currently we say:
> If an Original Publisher receives a FETCH with a range that includes an object with
> unknown status, it MUST return FETCH_ERROR with code Unknown Status in Range.

If I issue a Fetch for 1 million Objects, I have to be sure the status isn't unknown before returning FETCH_OK, because I can't return FETCH_OK and then FETCH_ERROR.

Is this really practical?

afrind:
The use case for UNKNOWN_STATUS_IN_RANGE is for distributed publishers.  It's expected that a given publisher will instantly know that a requested range covers groups that it is not responsible for.  See also #1117.

Consider a relatively straightforward example - a publishing client (C), a long term storage system (S) and a single relay (R) and a fetch client (F).  

```
S->R: PUBLISH_NAMESPACE (foo)
S->R: SUBSCRIBE_NAMESPACE (foo)
C->R: PUBLISH (foo, name=x)
R->S: PUBLISH (foo, name=x)
C->R: Objects starting from Group ID = [start time]
Objects go to both R and S (and downstream clients)

Client crashes and restarts
C->R: PUBLISH (foo, name=x)
R->S: PUBLISH (foo, name=x)
C->R: Objects starting from Group ID = [new start time]
...
F->R: FETCH start=A, end=B // where R < new start time
R->C: FETCH start=A, end=B
C->R: FETCH_ERROR, UNKNOWN_STATUS_IN_RANGE
R->S: FETCH start=A, end=B
S->R: FETCH_OK
```

I'm not saying this is a great architecture for ingest/storage, but it's at least plausible.

vasilvv:
I thought the other use case is when I recover after the crash, and I know that I can only serve group starting with a certain group ID?  In that case, this does seem practical.

I am not sure this is a good idea, though?  Mostly because if I start at time 100, and the subscriber requests [50, 150], I think they'd much rather get the [100, 150] range than an error.

afrind:
#728 notes mention that there's interest in trimming the range to what is known by the publisher, rather than failing outright.  I could be ok with that, however, FETCH is already quite complicated so adding more mechanisms, particularly when not many people have implemented the ones we have, gives me pause.

Also, @vasilvv what is your proposed normative relay handling when I have the above scenario -- two publishers with different answers for FETCH [50, 150]?

vasilvv:
[...]

PR #1331, opened 2025-10-22
Allow unknown ranges in a FETCH response

### LINKED ISSUE #945: FETCH stream stalls if Objects exist and will be published, but are lower priority
It is possible that an original publisher has multiple tracks (ie: 1080p, 720p, 360p) and the highest quality is lower priority than the others.

In such a case, the original publisher may want to eventually publish all the content from all tracks, but might not consistently have the bandwidth.

In cases such as live sports, the jitter buffer may be 10-30 seconds and so a SUBSCRIBE+Joining FETCH would likely be used.

If the FETCH happened to hit Objects that had been created, but the original publisher doesn't have bandwidth to send yet, I believe it would have to hang until those Objects were published, since otherwise FETCH indicates a permanent gap is present.

Having a FETCH pause for a moment because it has to go upstream will be normal, but having a FETCH potentially pause for minutes to hours with no clear reason why seems problematic.

A fairly simple solution to this problem is to add an Object Status that is "Not Yet Available" or "Currently Unavailable" or something similar to indicate that it might be available in the future, but it can't be retrieved in a timely manner.

afrind:
This is a proposal to address several open FETCH issues.

Replace UNKNOWN_STATUS_IN_RANGE with a data-stream encoding of an UNKNOWN range.  This requires leaving DOES_NOT_EXIST serializations in fetch, so a omitted range can be split between DNE and UNKNOWN.  But that seems OK to me.

I'm unsure if it addresses #945 completely, but it gives a bandwidth constrained publisher another option that doesn't completely fail the fetch. (Ian believes it does resolve #945, unless others object)

Fixes: #1265 
Fixes: #1057 
Fixes: #945

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3661,14 +3661,14 @@ the Object ID.
 
 #### End of Range
 
-When Serialization Flags indicates an End of Range (e.g. values 0x8C or 0x10C), 
-the Group ID and Object ID
-fields are present.  Subgroup ID, Priority and Extensions are not present. All
-Objects with Locations between the last serialized Object, if any, and this
-Location, inclusive, either do not exist (when Serialization Flags is 0x8C) or
-are unknown (0x10C).  A publisher SHOULD NOT use `End of Non-Existent Range` in
-a FETCH response except to split a range of Objects that will not be serialized
-into those that are known not to exist and those with unknown status.
+When Serialization Flags indicates an End of Range (e.g. values 0x8C or 0x10C),
+the Group ID and Object ID fields are present.  Subgroup ID, Priority and
+Extensions are not present. All Objects with Locations between the last
+serialized Object, if any, and this Location, inclusive, either do not exist
+(when Serialization Flags is 0x8C) or are unknown (0x10C).  A publisher SHOULD
+NOT use `End of Non-Existent Range` in a FETCH response except to split a range
+of Objects that will not be serialized into those that are known not to exist
+and those with unknown status.
 
 ## Examples
 


draft-ietf-moq-transport.md
```diff
@@ -2761,9 +2757,11 @@ A publisher MUST send fetched groups in the requested group order, either
 ascending or descending. Within each group, objects are sent in Object ID order;
 subgroup ID is not used for ordering.
 
-If an Original Publisher receives a FETCH with a range that includes an object with
-unknown status, it MUST return REQUEST_ERROR with code UNKNOWN_STATUS_IN_RANGE.
-
+If a Publisher receives a FETCH with a range that includes an object with
```
```suggestion
If a Publisher receives a FETCH with a range that includes one or more Objects with
```

ianswett (commented):
Thanks, LGTM

afrind:
@vasilvv would you like to propose alternate spelling?

afrind:
Discussed in 11/17 Interim:

The wg liked this general functionality but would like a richer set of "UNKNOWN" reasons like "publisher unavailable", etc.  and agree #1342 should go first.

vasilvv:
I've complained about the proposed syntax in this PR, so I am proposing the alternative:

If the highest bit in serialization flags is set, the following message is not an object, but some fetch-related metadata.  The remaining 7 bits denote the type of the metadata in question.  The first byte immediately after that indicates metadata length.

We define Flags = 0x80 to mean "range of objects with unknown status".  It has the following syntax:
```
Serialization Flags (8) = 0x80,
Metadata Length (i),
Start Group (i),
Start Object (i),
End Group (i),
End Object (i)
```

Not sure if the range end should be inclusive or not, probably not.

draft-ietf-moq-transport.md
```diff
@@ -3645,6 +3648,19 @@ the prior Object, the Subscriber MUST close the session with a
...
+Value | Meaning
+0x00 | Unknown
+0x01 | Does not Exist
+
+Any other value is a `PROTOCOL_VIOLATION`.  When Object Fetch Status is present,
+all Objects with Locations between the last serialized Object, if any, and this
+one, inclusive, have the given status.  A publisher SHOULD NOT use `Does not
+Exist` in a FETCH response except to split a range of Objects that will not be
+serialized into those that are known not to exist and those which are Unknown.
+
```
afrind: It is a protocol violation to include a subgroup, priority or extensions on a Fetch Status.

afrind:
Consider making serialization a varint rather than a byte.  When it fits in 1 byte, bottom values are flags.  Then define special types above 1-byte max.
Victor's proposal has appeal that you cannot encode invalid things like extensions on unknown range.

draft-ietf-moq-transport.md
```diff
@@ -3614,7 +3616,17 @@ format:
 ~~~
 {: #object-fetch-format title="MOQT Fetch Object Fields"}
 
-The Serialization Flags field defines the serialization of the Object.
+The Serialization Flags field defines the serialization of the Object.  It is
+a variable-length integer.  When less than 64, the bits repesent flags described
```
```suggestion
a variable-length integer.  When less than 64, the bits represent flags described
```

draft-ietf-moq-transport.md
```diff
@@ -3649,6 +3659,16 @@ When encoding an Object with a Forwarding Preference of "Datagram" (see
 {{object-properties}}), the Publisher treats it as having a Subgroup ID equal to
 the Object ID.
 
+#### End of Range
+
+When Serialization Flags indicates an End of Range, the Group ID and Object ID
```
```suggestion
When Serialization Flags indicates an End of Range (e.g. values 0x8C or 0x10C), 
the Group ID and Object ID
```

suhasHere (approved):
minor editorial nits. LGTM

PR #1334, opened 2025-10-22
Consistent use of `Established`

afrind:
I did't update the one referring to SUBSCRIBE_NAMESPACE, since we don't have a format state machine there yet.


yekuiwang:
> I did't update the one referring to SUBSCRIBE_NAMESPACE, since we don't have a format state machine there yet.

I noticed that. I was not sure whether that change should be made, but made it anyway to draw your attention (and then I would know :-) ).

ianswett (approved):

Issue #1313, opened 2025-10-15
Joining FETCH as a separate control message creates edge cases and feature gaps


ianswett:
The point of joining FETCH is to save the round trip, therefore the only time its valuable to send a joining FETCH is immediately after sending the SUBSCRIBE.

There are the obvious edge cases, such as referencing a subscription that no longer exists (#1261) or failed for some reason. There are probably more that I'm not thinking of.

~~Additionally, there's no FETCH_UPDATE today, so I can change the priority of the SUBSCRIBE, but not the corresponding Joining FETCH.~~

I also don't think the transition is handled well in some cases, as I commented in a now closed issue: https://github.com/moq-wg/moq-transport/issues/598#issuecomment-2603109830. They key problem being that the latest Object could have greatly increased between the time the SUBSCRIBE+JOINING_FETCH are made and when the JOINING_FETCH catches up realtime.  If they were a single message, it would be relatively simple to say that if Objects greater than the Largest Object have been published since the subscription started, they MAY be sent on the FETCH stream.

Merging the two messages simplifies this, and makes it clear how to correctly use this feature.  This would lose a slight bit of functionality in terms of being able to separately prioritize or cancel the FETCH and SUBSCRIBE, but in practice you could cancel either by narrowing the SUBFETCH (or whatever the combined verb is called) with SUBSCRIBE_UPDATE

Somewhat related to #1039 and #861

afrind:
> the only time its valuable to send a joining FETCH is immediately after sending the SUBSCRIBE.

You can send it after receiving a PUBLISH also?

> Additionally, there's no FETCH_UPDATE today, so I can change the priority of the SUBSCRIBE, but not the corresponding Joining FETCH.

This is not joining fetch specific, and the wg had little appetite to update it when discussed in Toronto (#1204).

As for the "subscribe+joining fetch -15 seconds" use case, can you elaborate more on what is actually broken?  It seems to me that in the worst case, the relay buffers live subscription data until the fetch completes, then delivers it.  Also, I think the scenario (backfill the buffer) works better as:

At time T:
SUBSCRIBE Largest Object, Pri=1
FETCH -15s, Descending, Pri=2 // lower priority
... 
FETCH_CANCEL // once total buffer len = 15s

If it takes 1s to download 5s of track data (fetch or subscribe), I'll have a 15s jitter buffer after 3s, starting at **T-12**

ianswett:
[...]

PR #1335, opened 2025-10-23
Allow joining FETCH for PUBLISH and SUBSCRIBE_UPDATE forward=1

afrind:
This also restricts joining fetch to subscriptions in Forward State=1, since joining a subscription in Forward State 0 does not have the intended "no gaps" effect.

This means that subscriptions initiated in Forward=0 need to update the join point when Forward transitions to 1.  Possible flow:

```
P->S: PUBLISH, Forward=0, Largest Object={10, 5}
P: Object 11,0
S->P: PUBLISH_OK, Forward=1
S->P: Joining FETCH
P->S: FETCH_OK, End=11,0 (inclusive)
P->S (fetch stream): 11,0
P->S (subscribe stream): Object 11,1
```

Note in the above flow, a standalone FETCH to Largest Object would miss 11,0.

This PR also allows re-joining a track that was temporarily in Forward=0:

```
S->P: SUBSCRIBE, Forward=1
S->P: Joining FETCH
...
S->P: SUBSCRIBE_UDPATE, Forward=0
...
S->P: SUBSCRIBE_UDPATE, Forward=1
S->P: Joining FETCH
P->S: REQUEST_OK, Largest={21,0}
P->S (fetch stream): Object 21,0
```

Addresses an issue raised in #1313

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1198,10 +1198,11 @@ is a join point, so in order for a subscriber to join a Track, it needs to
 request an existing Group or wait for a future Group.  Different applications
 will have different approaches for when to begin a new Group.
 
-To join a Track at a past Group, the subscriber sends a SUBSCRIBE, PUBLISH_OK or SUBSCRIBE_UPDATE with Filter
-Type `Largest Object` followed by a Joining FETCH (see {{joining-fetches}}) for
-the intended start Group, which can be relative.  To join a Track at the next
-Group, the subscriber sends a SUBSCRIBE with Filter Type `Next Group Start`.
+To join a Track at a past Group, the subscriber sends a SUBSCRIBE, PUBLISH_OK or
+REQUEST_UPDATE with Forward State 1 followed by a Joining FETCH (see
+{{joining-fetches}}) for the intended start Group, which can be relative.  To
+join a Track at the next Group, the subscriber sends a SUBSCRIBE with Filter
+Type `Next Group Start`.
 
 #### Dynamically Starting New Groups
 
@@ -2830,9 +2831,9 @@ a Location relative to the Largest group.
 A Subscriber can use a Joining Fetch to, for example, fill a playback buffer with a
 certain number of groups prior to the live edge of a track.
 
-A Joining Fetch is only permitted when the associated subscription Filter
-Type is Largest Object and the Forward State is 1; any other value results in closing
-the session with a `PROTOCOL_VIOLATION`.
+A Joining Fetch is only permitted when the associated subscription has
+Forward State 1; otherwise the publisher MUST close the session with a
+`PROTOCOL_VIOLATION`.
 
 If no Objects have been published for the track the publisher MUST
 respond with a REQUEST_ERROR with error code `INVALID_RANGE`.


draft-ietf-moq-transport.md
```diff
@@ -2658,30 +2658,29 @@ Joining Fetch {
...
-Largest Location.Object + 1}. Here Subscribe Largest Location is the
-saved value from when the subscription started (see {{subscriptions}}).
+The publisher receiving a Joining Fetch sets the End Location to
+{Joining Location.Group, Joining Location.Object + 1} (see {{subscriptions}}.
 
-Note: the last Object included in the Joining FETCH response is Subscribe
-Largest Location.  The `+ 1` above indicates the equivalent Standalone Fetch
+Note: the last Object included in the Joining FETCH response is Joining
+Location.  The `+ 1` above indicates the equivalent Standalone Fetch
 encoding.
```
```suggestion
Note: the last Object included in the Joining FETCH response is the Object
at the Joining Location.  The `+ 1` above indicates the equivalent Standalone
Fetch encoding.

File: draft-ietf-moq-transport.md [hunk 2]
```diff
@@ -926,6 +921,12 @@ the subscription is established, the subscriber can update the Forward State by
 sending SUBSCRIBE_UPDATE.  Control messages, such as PUBLISH_DONE
 ({{message-publish-done}}) are still sent on subscriptions in Forward State 0.
 
+A publisher MUST save the Largest Location communicated in SUBSCRIBE_OK, PUBLISH
+or REQUEST_OK in response to a SUBSCRIBE_UPDATE that changes the Forward State
```
```suggestion
or REQUEST_OK (in response to a SUBSCRIBE_UPDATE) that changes the Forward State
```

ianswett:
Based on my initial review, I think this PR looks sensible and it's definitely a use case worth solving, but this feels bad.

draft-ietf-moq-transport.md
```diff
@@ -2658,30 +2658,29 @@ Joining Fetch {
 }
 ~~~
 
-* Joining Request ID: The Request ID of the existing subscription to be
+* Joining Request ID: The Request ID of the `Established` subscription to be
   joined. If a publisher receives a Joining Fetch with a Request ID that does
-  not correspond to an existing Subscribe in the same session, it MUST return
+  not correspond to an `Established` subscription in the same session, it MUST return
```
afrind: Some other states are valid. `Pending (subscriber)`

afrind: Fixed in #1363

ianswett:
Per discussion, there's already an issue where if you use NGR, there are two possibilities:
 1) NGR will not produce a new group, so you'd like to Joining FETCH for the first part of the Group.
 2) NGR will produce a new group, so there's no need for a Joining FETCH

This is true for both SUBSCRIBE and SUBSCRIBE_UPDATE, so this PR doesn't change it.

afrind:
Parked pending Joining Fetch resolution post Boulder

afrind:
Per Boulder Discussion: 

I removed the restriction on the filter type since it is no longer relevant.  A Joining FETCH can be issued against any type of subscription in Forward State=1 -- the Fetch goes from the specified start (relative to the Joining Location or Absolute) to the Joining Location.

Issue #1256, opened 2025-09-23
Evaluation of authorization tokens given current unauthroized may need "retry after" error response


gloinul:
If one have authorization tokens that has properties that may result in unauthorized result currently, but may be produce valid at a later time, such as time limited access (in the future) or at another location appear to have the need for an error code or error + additional information to tell the requester to retry this request at a later time.

afrind:
I think a Retry-After field is generally useful in REQUEST_ERROR

afrind:
We will cover the generic Retry-After tomorrow when discussion #1175 and #1231

afrind:
@martinduke owns making a proposal to revamp errors with more specific information, including Retry-After semantics, which I believe will solve this.  Parking this issue pending that.

afrind:
See #1339

Issue #1175, opened 2025-08-25
Handling of non-defined error codes


michalhosna:
Hi,

The draft says an application “SHOULD use a relevant error/status code as defined below” in several places.

- Are implementations allowed to send an error/status code that is not specified in the draft?
- What is the expected behavior if an implementation receives such a non-defined code (e.g., treat as generic failure, ignore)? 
- Do we want to allow implementation-specific error codes with some kind of reserved range?

Sorry if this was already discussed — I wasn’t able to find a prior thread in issues, PRs, or the mailing list.

LPardue:
Responding as an individual and error code enthusiast

> Are implementations allowed to send an error/status code that is not specified in the draft?

Yes they should be. Error codes are an extension point of the protocol

> What is the expected behavior if an implementation receives such a non-defined code (e.g., treat as generic failure, ignore)?

We can turn the question around and ask what does a receiver do with an error code it knows? The answer is basically the same - deal with the fact something errored independent of the code, then possibly log the code and present to user 

> Do we want to allow implementation-specific error codes with some kind of reserved range?

I would advise us not to reserve any special range apriori. The space is large any way. At the point some custom error codes might be useful, you'd probably be better off with application-level mecahnsim to exchange information (i.e. something on top of moqt). You'd probably implementations to be aware if they are speaking to each other, a shared reserved range doesn't address that itself and would seem to be likely to increase the chance of collisions. Better to negotiate the use of custom things via an extension if really needed. The process for requesting moqt extension codepoints should be clear and simple enough to encourage people to do it this way.

afrind:
+1 to everything @LPardue said

Since there's now effectively a "User-Agent" setting, you can use the combo of that and the non-standard code to figure out what happened.

michalhosna:
I agree that sending non-specified errors should be allowed and properly handled. If we agree on this, I think the spec should state it explicitly.

(In the wild: [streaming-university/moqtail errors out on unknown error codes](https://github.com/streaming-university/moqtail/blob/73860deb323e64fe32154210e0975569991da6ea/libs/moqtail-rs/src/model/control/constant.rs#L247-L250))

> We can turn the question around and ask what does a receiver do with an error code it knows? The answer is basically the same - deal with the fact something errored independent of the code, then possibly log the code and present to user

[...]

PR #1339, opened 2025-10-24
Add Retry Interval to REQUEST_ERROR

martinduke:
Fixes the request error part of #1175.

Fixes: #1256

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -339,7 +339,7 @@ purposes such as logging, representing track filenames, or use in
 certain authorization verification schemes. The namespace and track name
 are binary, so they need to be converted to a safe form.
 
-The following format is RECOMMONDED:
+The following format is RECOMMENDED:
 
 * Each of the namespace tuples are rendered in order with a hyphen (-)
   between them followed by the track name with a double hyphen (--)
@@ -430,15 +430,15 @@ using at least as many streams as there are Subgroups,
 typically with a one-to-one mapping between Subgroups and streams.
 
 When an Object's forwarding preference (see {{object-properties}}) is
-"Datagram", it is not sent in Subgroups and the
-description in the remainder of this section does not apply.
+"Datagram", it is not sent in Subgroups, does not belong to a Subgroup in any
+way, and the description in the remainder of this section does not apply.
 
 Streams offer in-order reliable delivery and the ability to cancel sending and
 retransmission of data. Furthermore, many QUIC and WebTransport implementations
 offer the ability to control the relative scheduling priority of pending stream
 data.
 
-Every object within a Group belongs to exactly one Subgroup.
+Every Object within a Group belongs to exactly one Subgroup or Datagram.
 
 When Objects are sent in a subscription (see {{subscriptions}}),  Objects
 from two subgroups MUST NOT be sent on the same stream, and Objects from the
@@ -536,18 +536,24 @@ Track Namespace Field {
 * Track Namespace Field Value: A sequence of bytes that forms a Track Namespace
   Field.
 
+Each Track Namespace Field Value MUST contain at least one byte. If an endpoint
+receives a Track Namespace Field with a Track Namespace Field Length of 0, it
+MUST close the session with a `PROTOCOL_VIOLATION`.
+
 The structured nature of Track Namespace allows relays and applications to
 manipulate prefixes of a namespace. If an endpoint receives a Track Namespace
 consisting of 0 or greater than 32 Track Namespace Fields, it MUST close the
 session with a `PROTOCOL_VIOLATION`.
 
-Track Name is a sequence of bytes that identifies an individual track within the
-namespace.
+Track Name is a sequence of bytes, possibly empty, that identifies an individual
+track within the namespace.
 
-The maximum total length of a Full Track Name is 4,096 bytes. The length of a Full
-Track Name is computed as the sum of the Track Namespace Field Length fields
-and the Track Name Length field.  If an endpoint receives a Full Track Name
-exceeding this length, it MUST close the session with a `PROTOCOL_VIOLATION`.
+The maximum total length of a Full Track Name is 4,096 bytes. The length of a
+Full Track Name is computed as the sum of the Track Namespace Field Length
+fields and the Track Name Length field. The length of a Track Namespace is the
+sum of the Track Namespace Field Length fields. If an endpoint receives a Track
+Namespace or a Full Track Name exceeding 4,096 bytes, it MUST close the session
+with a `PROTOCOL_VIOLATION`.
 
 In this specification, both the Track Namespace Fields and the Track Name
 are not constrained to a specific encoding. They carry a sequence of bytes and
@@ -646,7 +652,11 @@ specifications define the type and value of the header, along with any rules
 concerning processing, modification, caching and forwarding.
 
 If unsupported by the relay, Extension Headers MUST NOT be modified, MUST be
-cached as part of the Track or Object and MUST be forwarded by relays.
+cached as part of the Track or Object and MUST be forwarded by relays.  If a
+Track or Object arrives with a different set of unknown extensions than
+previously cached, the most recent set SHOULD replace any cached values,
+removing any unknown values not present in the new set.  Relays MUST NOT attempt
+to merge sets of unknown extensions received in different messages.
 
 If supported by the relay and subject to the processing rules specified in the
 definition of the extension, Extension Headers MAY be modified, added, removed,
@@ -1340,6 +1350,11 @@ to different subscriptions or FETCH responses, but have the same subscriber and
 publisher priority.  The ordering in those cases is implementation-defined,
 though the expectation is that all subscriptions will be able to send some data.
 
+A publisher might not utilize the entire available congestion window,
+session flow control, or all available streams for lower
+priority Objects if it expects higher priority Objects will be available to send
+in the near future or it wants to reserve some bandwidth for control messages.
+
 Given the critical nature of control messages and their relatively
 small size, the control stream SHOULD be prioritized higher than all
 subscribed Objects.
@@ -1480,7 +1495,7 @@ all possible Objects matching the filter have been delivered to the subscriber.
 Since tracking this can be prohibitively expensive, Relays are not required or
 expected to do so.
 
-### Graceful Subscriber Relay Switchover
+### Graceful Subscriber Relay Switchover {#graceful-subscriber-switchover}
 
 This section describes behavior a subscriber MAY implement
 to allow for a better user experience when a relay sends a GOAWAY.
@@ -1568,7 +1583,7 @@ the Relay MUST NOT propagate that message or Object to another Session, because
 it would enable a single Session error to force an unrelated Session, which
 might be handling other subscriptions, to be closed.
 
-### Graceful Publisher Network Switchover
+### Graceful Publisher Network Switchover {#graceful-publisher-switchover}
 
 This section describes a behavior that a publisher MAY
 choose to implement to allow for a better user experience when
@@ -1741,11 +1756,11 @@ versions. All other messages use a version-specific namespace.
 For example, the integer '1' can refer to different parameters for Setup
 messages and for all other message types. SETUP message parameter types
 are defined in {{setup-params}}. Version-specific parameter types are defined
-in {{version-specific-params}}.
+in {{message-params}}.
 
-### Version Specific Parameters {#version-specific-params}
+### Message Parameters {#message-params}
 
-Each version-specific parameter definition indicates the message types in which
+Each message parameter definition indicates the message types in which
 it can appear. If it appears in some other type of message, it MUST be ignored.
 Note that since Setup parameters use a separate namespace, it is impossible for
 these parameters to appear in Setup messages.
@@ -1753,10 +1768,9 @@ these parameters to appear in Setup messages.
 #### AUTHORIZATION TOKEN Parameter {#authorization-token}
 
 The AUTHORIZATION TOKEN parameter (Parameter Type 0x03) MAY appear in a
-CLIENT_SETUP, SERVER_SETUP, PUBLISH, SUBSCRIBE, REQUEST_UPDATE,
-SUBSCRIBE_NAMESPACE, PUBLISH_NAMESPACE, TRACK_STATUS or FETCH message. This
-parameter conveys information to authorize the sender to perform the operation
-carrying the parameter.
+PUBLISH, SUBSCRIBE, REQUEST_UPDATE, SUBSCRIBE_NAMESPACE, PUBLISH_NAMESPACE,
+TRACK_STATUS or FETCH message. This parameter conveys information to authorize
+the sender to perform the operation carrying the parameter.
 
 The parameter value is a Token structure containing an optional Session-specific
 Alias. The Alias allows the sender to reference a previously transmitted Token
@@ -1872,8 +1886,10 @@ SUBSCRIBE, SUBSCRIBE_OK, or SUBSCRIBE_UDPATE message.
 It is the duration in milliseconds the relay SHOULD
 continue to attempt forwarding Objects after they have been received.  The start
 time for the timeout is based on when the Object Headers are received, and does
-not depend upon the forwarding preference. There is no explicit signal that an
-Object was not sent because the delivery timeout was exceeded.
+not depend upon the forwarding preference. Objects with forwarding preference
+'Datagram' are not retransmitted when lost, so the Delivery Timeout only limits
+the amount of time they can be queued before being sent. There is no explicit
+signal that an Object was not sent because the delivery timeout was exceeded.
 
 DELIVERY_TIMEOUT, if present, MUST contain a value greater than 0.  If an
 endpoint receives a DELIVERY_TIMEOUT equal to 0 it MUST close the session
@@ -1894,7 +1910,9 @@ causing the publisher to exceed its resource limits, the publisher MAY terminate
 the subscription with error `TOO_FAR_BEHIND`.
 
 If an object in a subgroup exceeds the delivery timeout, the publisher MUST
-reset the underlying transport stream (see {{closing-subgroup-streams}}).
+reset the underlying transport stream (see {{closing-subgroup-streams}}) and
+SHOULD NOT attempt to open a new stream to deliver additional Objects in that
+Subgroup.
 
 When sent by a subscriber, this parameter is intended to be specific to a
 subscription, so it SHOULD NOT be forwarded upstream by a relay that intends
@@ -2159,9 +2177,10 @@ initiation.
 
 #### AUTHORIZATION TOKEN {#setup-auth-token}
 
-See {{authorization-token}}.  The endpoint can specify one or more tokens in
-CLIENT_SETUP or SERVER_SETUP that the peer can use to authorize MOQT session
-establishment.
+The AUTHORIZATION TOKEN setup parameter (Parameter Type 0x03)) is funcionally
+equivalient to the AUTHORIZATION TOKEN message parameter, see {{authorization-token}}.
+The endpoint can specify one or more tokens in CLIENT_SETUP or SERVER_SETUP
+that the peer can use to authorize MOQT session establishment.
 
 If a server receives an AUTHORIZATION TOKEN parameter in CLIENT_SETUP with Alias
 Type REGISTER that exceeds its MAX_AUTH_TOKEN_CACHE_SIZE, it MUST NOT fail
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2196,66 +2197,83 @@ REQUEST_ERROR Message {
...
+later time. If so, the sender of REQUEST_ERROR includes a Retry Interval in the
+message. If it is sending more than one such message within a second or so, it
+SHOULD jitter the Retry Interval to avoid a flood of retries. The Retry Interval
+MAY be zero if the request can be retried immediately.
+
+If the sender has no information as to when a request is likely to be
+successful, it MAY jitter around a default interval of 30 seconds.
+
+If a Retry Interval exceeds the lifetime of a necessary authentication token
+used in the request, so that a retry at that time would faile, the sender SHOULD
```
```suggestion
used in the request, so that a retry at that time would fail, the sender SHOULD
```

draft-ietf-moq-transport.md
```diff
@@ -2196,66 +2197,83 @@ REQUEST_ERROR Message {
...
+use a non-retryable error code indicating a new authentication token is needed.
+
+INTERNAL_ERROR:
```
afrind, line 2359: Do we need a way to specify a non-retryable internal error?

draft-ietf-moq-transport.md
```diff
@@ -2196,66 +2197,83 @@ REQUEST_ERROR Message {
...
 
 Below are errors for use by the publisher. They can appear in response to
 SUBSCRIBE, FETCH, TRACK_STATUS, and SUBSCRIBE_NAMESPACE, unless otherwise noted.
 
-DOES_NOT_EXIST (0x10):
+DOES_NOT_EXIST:
 : The track or namespace is not available at the publisher.
 
-INVALID_RANGE (0x11):
+INVALID_RANGE:
```
afrind, line 2391: I think INVALID_RANGE might need to split.  It's for End < Start [not retryable] and FETCH Start > Largest [retryable?]

draft-ietf-moq-transport.md
```diff
@@ -2196,66 +2197,83 @@ REQUEST_ERROR Message {
...
-MALFORMED_TRACK (0x12):
+MALFORMED_TRACK:
 : In response to a FETCH, a relay publisher detected
 the track was malformed (see {{malformed-tracks}}).
 
 The following are errors for use by the subscriber. They can appear in response
 to PUBLISH or PUBLISH_NAMESPACE, unless otherwise noted.
 
-UNINTERESTED (0x20):
+UNINTERESTED:
```
afrind, line 2402: Are there both flavors of uninterested?

draft-ietf-moq-transport.md
```diff
@@ -2196,66 +2197,83 @@ REQUEST_ERROR Message {
...
 
-EXPIRED_AUTH_TOKEN (0x5):
+EXPIRED_AUTH_TOKEN:
 : Authorization token has expired ({{authorization-token}}).
 
 Below are errors for use by the publisher. They can appear in response to
 SUBSCRIBE, FETCH, TRACK_STATUS, and SUBSCRIBE_NAMESPACE, unless otherwise noted.
 
-DOES_NOT_EXIST (0x10):
+DOES_NOT_EXIST:
```
afrind, line 2388: Do we need both flavors of DOES_NOT_EXIST?  eg a simple app that gets a SUBSCRIBE for a track it will never publish never wants to see that again.  A relay might simply not have that track "right now".

draft-ietf-moq-transport.md
```diff
@@ -2196,66 +2197,83 @@ REQUEST_ERROR Message {
...
 to PUBLISH or PUBLISH_NAMESPACE, unless otherwise noted.
 
-UNINTERESTED (0x20):
+UNINTERESTED:
 : The subscriber is not interested in the track or namespace.
 
 Errors below can only be used in response to one message type.
 
-PREFIX_OVERLAP (0x30):
+PREFIX_OVERLAP:
```
afrind, line 2407: This is sort of retryable if I unsubscribe to a namespace first?  Maybe this is fine.

draft-ietf-moq-transport.md
```diff
@@ -2196,66 +2197,83 @@ REQUEST_ERROR Message {
...
-as defined below. Most codepoints have identical meanings for various request
-types, but some have request-specific meanings.
+as defined below and assigned in {{iana-request-error}}. Most codepoints have
+identical meanings for various request types, but some have request-specific
+meanings.
 
-INTERNAL_ERROR (0x0):
+Odd error codes indicate the request is retryable with the same parameters at a
+later time. If so, the sender of REQUEST_ERROR includes a Retry Interval in the
+message. If it is sending more than one such message within a second or so, it
```
afrind: Do you mean the same error to multiple endpoints, or multiple errors to the same endpoint, or both?

draft-ietf-moq-transport.md
```diff
@@ -2196,66 +2197,83 @@ REQUEST_ERROR Message {
...
+SHOULD jitter the Retry Interval to avoid a flood of retries. The Retry Interval
```
afrind: Do we need more text explaining what "jitter the retry interval" means?

afrind (commented):
Thanks for this proposal.

This also fixes #1256.

draft-ietf-moq-transport.md
martinduke: both. why would it matter which?

draft-ietf-moq-transport.md
martinduke: It seems clear to me. propose something if you want it.

draft-ietf-moq-transport.md
martinduke: No action because you said it was fine.

draft-ietf-moq-transport.md
afrind: I'd recommend mentioning "across one or more sessions" to make it clear that the second case is included?

draft-ietf-moq-transport.md
martinduke: Sure, made the change. Not sure why you wouldn't just send SUBSCRIBE when interested, but I don't really care.

draft-ietf-moq-transport.md
```diff
@@ -2196,66 +2197,83 @@ REQUEST_ERROR Message {
...
+MAY be zero if the request can be retried immediately.
```
```suggestion
SHOULD apply randomization to each retry interval so that retries are spread out
over time, minimizing the risk of synchronized retry storms.  The Retry Interval
MAY be zero if the request can be retried immediately.
```

draft-ietf-moq-transport.md
```diff
@@ -2196,66 +2197,83 @@ REQUEST_ERROR Message {
...
+
+If the sender has no information as to when a request is likely to be
+successful, it MAY jitter around a default interval of 30 seconds.
```
```suggestion
successful, it MAY apply randomization around a default interval of 30 seconds.
```

draft-ietf-moq-transport.md
```diff
@@ -2187,6 +2187,7 @@ REQUEST_ERROR Message {
   Length (16),
   Request ID (i),
   Error Code (i),
+  [Retry Interval (i)],
```
ianswett: In terms of framing, I think I'd prefer to not split error codes and instead make the Retry Interval 'optional' by reserving 0 as a special value that's equivalent to it not being present.

This also avoids us having a discussion about which of the error codes should be retryable.

ianswett (commented):
This looks sensible, but I'd prefer to not split error codes and make the retry-after always optional

draft-ietf-moq-transport.md
```diff
@@ -2212,66 +2213,88 @@ REQUEST_ERROR Message {
...
+are spread out over time, minimizing the risk of synchronized retry storms.  The
+Retry Interval value MAY be one if the request can be retried immediately.
 
-UNAUTHORIZED (0x1):
+If the sender has no information as to when a request is likely to be
+successful, it MAY apply randomization around a default interval of 30 seconds.
+
+If a Retry Interval exceeds the lifetime of a necessary authentication token
+used in the request, so that a retry at that time would fail, the sender SHOULD
+use an error code indicating a new authentication token is needed.
```
afrind: Do we really mean this?  Like if I get a DOES_NOT_EXIST with a valid auth token, and Retry-After 30 (since it will exist then), but the token expires in 15, I change the error to UNAUTHORIZED?

That seems super confusing.  I think I'd rather get the real error code, and burn the round trip when my token actually expires.

draft-ietf-moq-transport.md
```diff
@@ -2212,66 +2213,88 @@ REQUEST_ERROR Message {
...
 The application SHOULD use a relevant error code in REQUEST_ERROR,
-as defined below. Most codepoints have identical meanings for various request
-types, but some have request-specific meanings.
+as defined below and assigned in {{iana-request-error}}. Most codepoints have
+identical meanings for various request types, but some have request-specific
+meanings.
 
-INTERNAL_ERROR (0x0):
-: An implementation specific or generic error occurred.
+If request is retryable with the same parameters at a later time, the sender of
```
```suggestion
If a request is retryable with the same parameters at a later time, the sender of
```

draft-ietf-moq-transport.md
```diff
@@ -2212,66 +2213,88 @@ REQUEST_ERROR Message {
...
+REQUEST_ERROR includes a non-zero Retry Interval in the message. If it is
+sending more than one such message within a second or so across one or more
+sessions, it SHOULD apply randomization to each retry interval so that retries
+are spread out over time, minimizing the risk of synchronized retry storms.  The
+Retry Interval value MAY be one if the request can be retried immediately.
```
```suggestion
Retry Interval value is 1 if the request can be retried immediately.
```

draft-ietf-moq-transport.md
```diff
@@ -2212,66 +2213,88 @@ REQUEST_ERROR Message {
...
 
-UNAUTHORIZED (0x1):
+If the sender has no information as to when a request is likely to be
+successful, it MAY apply randomization around a default interval of 30 seconds.
```
ianswett: This is ok, but do we need this?

draft-ietf-moq-transport.md
```diff
@@ -2212,66 +2213,88 @@ REQUEST_ERROR Message {
...
+
+If a Retry Interval exceeds the lifetime of a necessary authentication token
+used in the request, so that a retry at that time would fail, the sender SHOULD
+use an error code indicating a new authentication token is needed.
+
+INTERNAL_ERROR:
+: An implementation specific or generic error occurred. This might be retryable
+or not, depending on the imlementation conditions that caused the error.
```
```suggestion
or not, depending on the implementation conditions that caused the error.
```

ianswett (commented):
Some small comments, but this looks close.

draft-ietf-moq-transport.md
```suggestion
are spread out over time, minimizing the risk of synchronized retry storms.  A
Retry Interval value of 1 indicates the request can be retried immediately.
```

draft-ietf-moq-transport.md
afrind: Or make it a SHOULD?

ianswett: I'm inclined to delete this sentence because I don't think it's needed.

draft-ietf-moq-transport.md
afrind: I am inclined to remove this entirely.

draft-ietf-moq-transport.md
```diff
@@ -2212,66 +2213,88 @@ REQUEST_ERROR Message {
...
+REQUEST_ERROR includes a non-zero Retry Interval in the message. If it is
+sending more than one such message within a second or so across one or more
+sessions, it SHOULD apply randomization to each retry interval so that retries
+are spread out over time, minimizing the risk of synchronized retry storms.  A
+Retry Interval value of 1 indicates the request can be retried immediately.
 
-UNAUTHORIZED (0x1):
+If the sender has no information as to when a request is likely to be
+successful, it MAY apply randomization around a default interval of 30 seconds.
+
```
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -2212,66 +2213,88 @@ REQUEST_ERROR Message {
...
+If a Retry Interval exceeds the lifetime of a necessary authentication token
+used in the request, so that a retry at that time would fail, the sender SHOULD
+use an error code indicating a new authentication token is needed.
+
```
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -3924,6 +3940,13 @@ TODO: register the URI scheme and the ALPN and grease the Extension types
 | INVALID_JOINING_REQUEST_ID | 0x32 | {{message-request-error}} |
 | UNKNOWN_STATUS_IN_RANGE    | 0x33 | {{message-request-error}} |
 
+The range of error codes including 0x100 to 0xffff is reserved for
```
fluffy: I'm not keen on implementation specific error code. I think that contributes to non interoperable implementations. I'd like to have some discussion on this.

draft-ietf-moq-transport.md
```diff
@@ -3924,6 +3940,13 @@ TODO: register the URI scheme and the ALPN and grease the Extension types
...
+implementation-speceific codes.
+
+The range of error codes starting with 0x10000 is reserved for
```
fluffy: Can we right this up in the normal way we would IANA. This should be there is a way to get IANA to pre allocate a code.

fluffy: I don't see any reason we need to invent new IANA processes beyond what is in rfc8126

draft-ietf-moq-transport.md
```diff
@@ -2212,66 +2213,81 @@ REQUEST_ERROR Message {
...
+as defined below and assigned in {{iana-request-error}}. Most codepoints have
+identical meanings for various request types, but some have request-specific
+meanings.
+
+If a request is retryable with the same parameters at a later time, the sender of
+REQUEST_ERROR includes a non-zero Retry Interval in the message. If it is
+sending more than one such message within a second or so across one or more
+sessions, it SHOULD apply randomization to each retry interval so that retries
+are spread out over time, minimizing the risk of synchronized retry storms.  A
+Retry Interval value of 1 indicates the request can be retried immediately.
```
fluffy: I think it needs to do randomization even if it is not doing multiple requests. A common problem is Start of Service attacks. With IP phones we regularly see cities of over 10 million people loose power, then all restart at the same time often all with a set top box from of some sort from the same vendor. This forms a Start Of Service attack where the normal mitigations is for the server to be able to respond with retry after and it is more robust to spreading out the load if that can be spread. 

My suggestion would be just make this very specific that the min time to respond was a number between the values returned and 125% of value returned with uniform random distribution in that time. Sure some clients won't bother but the ones that cause SoS attacks will.

draft-ietf-moq-transport.md
```diff
@@ -2212,66 +2213,81 @@ REQUEST_ERROR Message {
 
 * Error Code: Identifies an integer error code for request failure.
 
+* Retry Interval: The minimum time (in seconds) before the request SHOULD be
+  sent again, plus one. If the value is 0, the request SHOULD NOT be retried.
```
fluffy, line 2343: Just as something to think about ... ff you use seconds instead of ms, it makes it hard for the server to pace multiple clients but I can see how either would work.

afrind:
I added this to the agenda for tomorrow's meeting (11/17) and in I would like to get a show of hands on whether we will or will not support implementation defined codes.

afrind:
Talked to @fluffy :

1. Keep the current sender jitter rather
2. Switch from sec to ms

Regarding error code extensibility, we do want application defined codes, but don't necessarily want to mix them in the same field as the existing MOQT error codes (eg: how QUIC CONNECTION_CLOSE has either QUIC error or an application error).  One idea is to have a separate Application Error field.

draft-ietf-moq-transport.md
```diff
@@ -2212,66 +2213,81 @@ REQUEST_ERROR Message {
 
 * Error Code: Identifies an integer error code for request failure.
 
+* Retry Interval: The minimum time (in seconds) before the request SHOULD be
```
```suggestion
* Retry Interval: The minimum time (in milliseconds) before the request SHOULD be
```

draft-ietf-moq-transport.md
```diff
@@ -2212,66 +2213,81 @@ REQUEST_ERROR Message {
...
-
-UNAUTHORIZED (0x1):
+as defined below and assigned in {{iana-request-error}}. Most codepoints have
+identical meanings for various request types, but some have request-specific
+meanings.
+
+If a request is retryable with the same parameters at a later time, the sender of
+REQUEST_ERROR includes a non-zero Retry Interval in the message. If it is
+sending more than one such message within a second or so across one or more
+sessions, it SHOULD apply randomization to each retry interval so that retries
```
```suggestion
sending more than one such message within a second or so across multiple
sessions, it SHOULD apply randomization to each retry interval so that retries
```

draft-ietf-moq-transport.md
```diff
@@ -2212,66 +2213,81 @@ REQUEST_ERROR Message {
...
+are spread out over time, minimizing the risk of synchronized retry storms.  A
+Retry Interval value of 1 indicates the request can be retried immediately.
+
+INTERNAL_ERROR:
+: An implementation specific or generic error occurred. This might be retryable
+or not, depending on the implementation conditions that caused the error.
```
```suggestion
: An implementation specific or generic error occurred.
```

draft-ietf-moq-transport.md
```diff
@@ -2212,66 +2213,81 @@ REQUEST_ERROR Message {
...
 : Authorization token has expired ({{authorization-token}}).
 
 Below are errors for use by the publisher. They can appear in response to
 SUBSCRIBE, FETCH, TRACK_STATUS, and SUBSCRIBE_NAMESPACE, unless otherwise noted.
 
-DOES_NOT_EXIST (0x10):
-: The track or namespace is not available at the publisher.
+DOES_NOT_EXIST:
+: The track or namespace is not available at the publisher. This might be
+retryable or not, if the target might exist later.
```
```suggestion
: The track or namespace is not available at the publisher.
```

draft-ietf-moq-transport.md
```diff
@@ -2212,66 +2213,81 @@ REQUEST_ERROR Message {
...
 
-INVALID_RANGE (0x11):
+INVALID_RANGE:
 : In response to SUBSCRIBE or FETCH, specified Filter or range of Locations
-cannot be satisfied.
+cannot be satisfied. This might be retryable if the range is expected to have
+objects in the future.
```
```suggestion
cannot be satisfied.
```

draft-ietf-moq-transport.md
```diff
@@ -2212,66 +2213,81 @@ REQUEST_ERROR Message {
...
 malformed (see {{malformed-tracks}}).
 
 The following are errors for use by the subscriber. They can appear in response
 to PUBLISH or PUBLISH_NAMESPACE, unless otherwise noted.
 
-UNINTERESTED (0x20):
-: The subscriber is not interested in the track or namespace.
+UNINTERESTED:
+: The subscriber is not interested in the track or namespace. This might be
+retryable if it expects to be interested later.
```
```suggestion
: The subscriber is not interested in the track or namespace.
```

draft-ietf-moq-transport.md
```diff
@@ -3924,6 +3940,13 @@ TODO: register the URI scheme and the ALPN and grease the Extension types
 | INVALID_JOINING_REQUEST_ID | 0x32 | {{message-request-error}} |
 | UNKNOWN_STATUS_IN_RANGE    | 0x33 | {{message-request-error}} |
 
+The range of error codes including 0x100 to 0xffff is reserved for
+implementation-speceific codes.
+
+The range of error codes starting with 0x10000 is reserved for
+provisional error codes that are under consideration for a permanent
+code point by the IETF.
+
```
```suggestion
```

draft-ietf-moq-transport.md
afrind: Discussed with Ian, we don't think we need to be too prescriptive here in _how_ servers jitter to handle specific cases like this - it should be sufficient to say that they SHOULD jitter when sending a lot of errors.

afrind (approved):

draft-ietf-moq-transport.md
```diff
@@ -2313,33 +2314,44 @@ REQUEST_ERROR Message {
...
+as defined below and assigned in {{iana-request-error}}. Most codepoints have
+identical meanings for various request types, but some have request-specific
+meanings.
+
+If a request is retryable with the same parameters at a later time, the sender of
+REQUEST_ERROR includes a non-zero Retry Interval in the message. If it is
+sending more than one such message within a second or so across multiple
+sessions, it SHOULD apply randomization to each retry interval so that retries
+are spread out over time, minimizing the risk of synchronized retry storms.  A
+Retry Interval value of 1 indicates the request can be retried immediately.
```
```suggestion
If a request is retryable with the same parameters at a later time, the sender
of REQUEST_ERROR includes a non-zero Retry Interval in the message. To minimize
the risk of synchronized retry storms, the sender can apply randomization to
each retry interval so that retries are spread out over time.  A Retry Interval
value of 1 indicates the request can be retried immediately.
```

vasilvv (approved):

Issue #1340, opened 2025-10-24
What happens if both endpoints try to establish a subscription to the same track?


afrind:
Current text (in SUBSCRIBE definition) 

> A subscriber MUST NOT establish multiple concurrent subscriptions for a track within a single session and publishers SHOULD treat this as a protocol violation.

1. We should move this restriction to the subscriptions section, since it applies to PUBLISH also
2. It's possible for the publisher and subscriber to race establishing a subscription to the same track.  This probably shouldn't be a PROTOCOL_VIOLATION. Instead we can clearly say which one wins, and which one gets a REQUEST_ERROR.  Publisher wins?

> If a publisher receives a SUBSCRIBE request for a subscription in the Pending state, it MUST reject that request with a REQUEST_ERROR with code DUPLICATE_SUBSCRIPTION.  If a subscriber receives a PUBLISH for a subscription in the Pending state, it MUST first terminate its pending subscription UNSUBSCRIBE before sending PUBLISH_OK.

yekuiwang:
Looks good to let the publisher win. 

> If a subscriber receives a PUBLISH for a subscription in the Pending state, it MUST first terminate its pending subscription UNSUBSCRIBE before sending PUBLISH_OK.

If the subscriber receives the REQUEST_ERROR with code DUPLICATE_SUBSCRIPTION before trying to terminate its pending subscription using UNSUBSCRIBE, then it no longer needs to terminate its pending subscription using UNSUBSCRIBE before sending PUBLISH_OK, right?

afrind:
S->P: SUBSCRIBE id=1
P->S: PUBLISH id=0
P->S: REQUEST_ERROR id=1

If the subscriber processes the REQUEST_ERROR, that subscription moves from `Pending (subscriber)` to `Terminated`, so there's no need to UNSUBSCRIBE.

afrind:
> If a publisher receives a SUBSCRIBE request for a subscription in the `Pending (publisher)` state, it MUST reject that request with a REQUEST_ERROR with code DUPLICATE_SUBSCRIPTION. If a subscriber receives a PUBLISH for a subscription in the `Pending (subscriber)` state, it MUST transition the subscription to `Terminated` before sending PUBLISH_OK.

And the state machine diagram is missing sender transitions out of Pending states

Pending (subscriber) -> send UNSUBSCRIBE -> Terminated
Pending (publisher) -> send PUBLISH_DONE -> Terminated

PR #1341, opened 2025-10-25
Duplicate subscription processing

afrind:
1) Makes duplicate subscriptions a request error rather than a session error, since it can happen accidentally 2) explain who wins publish/subscribe races
3) update the state machine to indicate that the initiator of a subscription can cancel it from the pending state

We actually allow > 1 subscription to the same track in the same session, with self-subscribe.  The restriction is actually per track + role.

Fixes: #1340

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -947,7 +947,7 @@ with the same role, it MUST fail that request with a `DUPLICATE_SUBSCRIPTION`
 error.
 
 If a publisher receives a SUBSCRIBE request for a Track with an existing
-subscription not in `Pending (publisher)` state, it MUST fail that request with
+subscription in `Pending (publisher)` state, it MUST fail that request with
 a `DUPLICATE_SUBSCRIPTION` error. If a subscriber receives a PUBLISH for a Track
 with a subscription in the `Pending (Subscriber)` state, it MUST ensure the
 subscription it initiated transitions to the `Terminated` state before sending


draft-ietf-moq-transport.md
```diff
@@ -931,14 +932,26 @@ prior messages other than SETUP.  Relays MUST NOT send any PUBLISH messages
 without knowing the client is interested in and authorized to receive the
 content. The communication of intent and authorization can be accomplished by
 the client sending SUBSCRIBE_NAMESPACE, or conveyed in other mechanisms out of
-band.
+band./
```
```suggestion
band.
```

draft-ietf-moq-transport.md
```diff
@@ -939,6 +940,18 @@ subscriptions initiated by other endpoints, and all published Objects will be
 forwarded back to the endpoint, subject to priority and congestion response
 rules.
 
+For a given Track, an endpoint can have at most one subscription to the Track
```
```suggestion
For a given Track, an endpoint can have at most one subscription to a Track
```

draft-ietf-moq-transport.md
```diff
@@ -939,6 +940,18 @@ subscriptions initiated by other endpoints, and all published Objects will be
...
+acting as the publisher and at most one acting as a subscriber.  If an endpoint
+receives a message attempting to establish a second subscription to a Track
+with the same role, it MUST fail that request with a `DUPLICATE_SUBSCRIPTION`
+error.
+
+If a publisher receives a SUBSCRIBE request for a Track with an existing
+subscription not in `Pending (publisher)` state, it MUST fail that request with
+a `DUPLICATE_SUBSCRIPTION` error. If a subscriber receives a PUBLISH for a
+subscription in the `Pending (Subscriber)` state, it MUST ensure the
+subscription transitions to the `Terminated` state before sending PUBLISH_OK.
```
ianswett: Which subscription?  The one it initiated?

draft-ietf-moq-transport.md
```diff
@@ -939,6 +940,18 @@ subscriptions initiated by other endpoints, and all published Objects will be
 forwarded back to the endpoint, subject to priority and congestion response
 rules.
 
+For a given Track, an endpoint can have at most one subscription to the Track
+acting as the publisher and at most one acting as a subscriber.  If an endpoint
+receives a message attempting to establish a second subscription to a Track
+with the same role, it MUST fail that request with a `DUPLICATE_SUBSCRIPTION`
+error.
+
+If a publisher receives a SUBSCRIBE request for a Track with an existing
```
ianswett, line 949: The idea here is that when PUBLISH and SUBSCRIBE are both in flight, PUBLISH wins?

ianswett (commented):
Looking good, thanks

draft-ietf-moq-transport.md
```diff
@@ -939,6 +940,18 @@ subscriptions initiated by other endpoints, and all published Objects will be
 forwarded back to the endpoint, subject to priority and congestion response
 rules.
 
+For a given Track, an endpoint can have at most one subscription to the Track
+acting as the publisher and at most one acting as a subscriber.  If an endpoint
```
suhasHere, line 944: There can me more than one publisher to  a track and the same is true for Subscriber

afrind: @suhasHere in the same session?  See the existing text which I moved/updated at line 2265 

> A subscriber MUST NOT establish multiple concurrent subscriptions for a track within a single session and publishers SHOULD treat this as a protocol violation.

I guess we didn't have rules for the publisher establishing concurrent subscriptions, but what is the use case?

draft-ietf-moq-transport.md
afrind: Yes, I'm amenable to subscriber wins too, we just need to pick one.

draft-ietf-moq-transport.md
```diff
@@ -939,6 +940,18 @@ subscriptions initiated by other endpoints, and all published Objects will be
...
+receives a message attempting to establish a second subscription to a Track
+with the same role, it MUST fail that request with a `DUPLICATE_SUBSCRIPTION`
+error.
+
+If a publisher receives a SUBSCRIBE request for a Track with an existing
+subscription not in `Pending (publisher)` state, it MUST fail that request with
+a `DUPLICATE_SUBSCRIPTION` error. If a subscriber receives a PUBLISH for a
+subscription in the `Pending (Subscriber)` state, it MUST ensure the
+subscription transitions to the `Terminated` state before sending PUBLISH_OK.
```
```suggestion
a `DUPLICATE_SUBSCRIPTION` error. If a subscriber receives a PUBLISH for a Track
with a subscription in the `Pending (Subscriber)` state, it MUST ensure the
subscription it initiated transitions to the `Terminated` state before sending
PUBLISH_OK.
```

draft-ietf-moq-transport.md
```diff
@@ -939,6 +940,19 @@ subscriptions initiated by other endpoints, and all published Objects will be
...
 rules.
 
+For a given Track, an endpoint can have at most one subscription to a Track
+acting as the publisher and at most one acting as a subscriber.  If an endpoint
+receives a message attempting to establish a second subscription to a Track
+with the same role, it MUST fail that request with a `DUPLICATE_SUBSCRIPTION`
+error.
+
+If a publisher receives a SUBSCRIBE request for a Track with an existing
+subscription not in `Pending (publisher)` state, it MUST fail that request with
```
```suggestion
subscription in `Pending (publisher)` state, it MUST fail that request with
```

ianswett (approved):

draft-ietf-moq-transport.md
ianswett: Can you add an intro sentence that indicates the intent?

afrind: We couldn't think of anything off the top of our head.

PR #1342, opened 2025-10-27
Redo the way we deal with missing Objects and Object Status

### LINKED ISSUE #1093: End of Group flag in Subgroup/Datagram Header creates new ways to malform a track

### LINKED ISSUE #1223: Do we ever need to explicitly send "Object Does Not Exist" on the wire?
Between other existing object status codes, and the Prior Object ID extension we added recently, I am not sure if there is a scenario under which we would need to explicitly send one of those on the wire.  Should we remove it?

### LINKED ISSUE #1224: Do we need Object Status and Prior Group/Object ID extensions in FETCH replies?
For both of the current Prior ID Gap extensions, we say:

> This extension can be added by the Original Publisher, but MUST NOT be added by relays. This extension MUST NOT be modified or removed.

This makes sense for SUBSCRIBE, but when you FETCH those, you can already figure out there is a gap from the fact that there are no objects in that range.

End of Group (and other status codes) has a similar issue.  If a FETCH reply has an object (6, 0), followed by an object (7, 0), one does not need a special End of Group at (6, 1); the only extra information there could come from the associated subgroup ID, or from the extensions.  I've heard people express at one of the recent calls skepticism as to whether using End of Group extensions is a good idea (we do currently say that "Any Object may have extension headers except those with Object Status 'Object Does Not Exist'").  If don't allow those, we can remove End of Group from FETCH replies; I think this actually applies to all objects with non-Normal status.

vasilvv:
Among other things, removes object status from FETCH, simplifies semantics of it in SUBSCRIBE, and allows relays to rephrase those.

Fixes #1093
Fixes #1223
Fixes #1224

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -539,6 +539,14 @@ include:
    Subgroup stream before a FIN.
 5. A Subgroup is received over multiple transport streams terminated by FIN with
    different final Objects.
+6. An Object is received in a Group whose Object ID is larger than the final
+   Object in the Group.  The final Object in a Group is the Object with Status
+   END_OF_GROUP or the last Object sent in a FETCH that requested the entire
+   Group.
+7. An Object is received on a Track whose Group and Object ID are larger than the
+   final Object in the Track.  The final Object in a Track is the Object with
+   Status END_OF_TRACK or the last Object sent in a FETCH whose response indicated
+   End of Track.
 8. The same Object is received more than once with different Payload or
     other immutable properties.
 9. An Object is received with a different Forwarding Preference than previously


draft-ietf-moq-transport.md
```diff
@@ -1296,20 +1302,23 @@ Objects or MAY use them to update certain cached fields. Implementations that
...
-duplicate object in this case.
+For ranges of objects that do not exist, relays MAY change the representation
+of a missing range to a semantically equivalent one.  For instance, a relay may
+change an End-of-Group="Y" Subgroup Header to an equivalent object with an End
+of Group status, or a Prior Group ID Gap extension could be removed in FETCH,
+where it's redundant.
+
+Note that due to reordering, an implementation can receive an Object after
+receiving an indication that the Object in question does not exist.  The
+endpoint SHOULD NOT cache or forward the duplicate object in this case.
```
yekuiwang: Just trying to understand:
1) Why reordering can make an Object to be received after receiving an indication that the Object in question does not exist?
2) Why is such an Object called a duplicate Object?

draft-ietf-moq-transport.md
```diff
@@ -3077,8 +3085,8 @@ Preference`.
   within the Group. This field is omitted if the `Object Forwarding Preference`
   is Datagram.
 
-* Object Status: An enumeration used to indicate missing
-objects or mark the end of a group or track. See {{object-status}} below.
+* Object Status: An enumeration used to indicate missing objects or mark the
+  end of a group or track. See {{object-status}} below.
```
```suggestion
* Object Status: An enumeration used to indicate whether the Object is a normal Object
  or mark the end of a group or track. See {{object-status}} below.

File: draft-ietf-moq-transport.md [hunk 3]
```diff
@@ -3091,34 +3099,24 @@ NOT be processed by a relay. Only present when 'Object Status' is Normal (0x0).
...
-         Group produced in this Track with Object ID zero. This status also
-         indicates the specified Group has ended. Publishers MUST NOT publish an
-         Object with a Location larger than this Location (see
-         {{malformed-tracks}}). This SHOULD be cached.
+* 0x3 := Indicates End of Group. Indicates that no objects with the specified
+         Group ID and the Object ID that is greater than or equal to the one
+         specified exist.
+
+* 0x4 := Indicates End of Track. Indicates that no objects with the location
+         that is equal to or greater than the one specified exist.
```
yekuiwang, line 3128: Hmm, now the semantics seem to be the same for EoG and EoT. Shouldn't the EoG and EoT property be clearly described instead?

draft-ietf-moq-transport.md
```diff
@@ -3550,7 +3549,6 @@ format:
   [Publisher Priority (8),]
   [Extensions (..),]
   Object Payload Length (i),
-  [Object Status (i),]
```
yekuiwang, line 3553: It is said above that "The Object Status is a field that is only present in objects that are delivered
via a SUBSCRIPTION, and is absent in Objects delivered via a FETCH." But now it seems completely removed?

draft-ietf-moq-transport.md
```diff
@@ -3091,34 +3099,24 @@ NOT be processed by a relay. Only present when 'Object Status' is Normal (0x0).
 
 #### Object Status {#object-status}
 
-The Object Status informs subscribers what objects will not be received
-because they were never produced, are no longer available, or because they
-are beyond the end of a group or track.
+The Object Status is a field that is only present in objects that are delivered
+via a SUBSCRIPTION, and is absent in Objects delivered via a FETCH.  It allows
+the publisher to explicitly communicate that a specific range of objects does
+not exist.
```
yekuiwang, line 3116: With the removal of "Object Does Not Exist", the Object Status field only indicates EoG or EoT, then it might be easier for readers to understand to just say that it indicates EoG or EoT, instead of saying that it indicates that a specific range of objects does not exist?

draft-ietf-moq-transport.md
```diff
@@ -522,14 +536,6 @@ include:
    Subgroup stream before a FIN.
 5. A Subgroup is received over multiple transport streams terminated by FIN with
    different final Objects.
-6. An Object is received in a Group whose Object ID is larger than the final
```
afrind, line 525: It still seems like there can be malformed errors like this, right?

draft-ietf-moq-transport.md
```diff
@@ -1296,20 +1302,23 @@ Objects or MAY use them to update certain cached fields. Implementations that
 update the cache need to protect against cache poisoning.  The only Object
 fields that can be updated are the following:
 
-1. Object Status can transition from any status to Object Does Not Exist in
-   cases where the object is no longer available.  Transitions between Normal,
-   End of Group and End of Track are invalid.
+1. Object can transition from existing to not existing in cases where the
+   object is no longer available.
 3. Object Header Extensions can be added, removed or updated, subject
```
```suggestion
2. Object Header Extensions can be added, removed or updated, subject
```

draft-ietf-moq-transport.md
afrind: 1. The publisher has redundant ingest paths to a relay
2. The "duplicate" language may be too strong.  The indication that the object doesn't exist may have come some other way than receiving an Object with a particular status.

draft-ietf-moq-transport.md
afrind: EOG is saying no objects exist between [{g,o}, {g+1,0})
EOT is saying no objects exist at or after {g,o}

draft-ietf-moq-transport.md
```diff
@@ -3234,8 +3232,9 @@ There are 10 defined Type values for OBJECT_DATAGRAM.
 | 0x2C | No | No | Yes | No | Status |
 | 0x2D | No | Yes | Yes | No | Status |
 
-* End of Group: For Type values where End of Group is "Yes" the Object is the
-  last Object in the Group.
+* End of Group: For Type values where End of Group is "Yes", indicates that no
+  Object with the same Group ID and the Object ID greater than the ID of the
+  current Object exists.
```
```suggestion
  Object with the same Group ID and an Object ID greater than `Object ID` exists.
```

draft-ietf-moq-transport.md
```diff
@@ -3366,10 +3365,10 @@ There are 24 defined Type values for SUBGROUP_HEADER:
...
 |------|---------------|-----------------|------------|--------------|---------------|
 
-For Type values where Contains End of Group is Yes, the last Object in this
-Subgroup stream before a FIN is the last Object in the Group.  If the Subgroup
-stream is terminated with a RESET_STREAM or RESET_STREAM_AT, the receiver cannot
-determine the End of Group Object ID.
+For Type values where Contains End of Group is Yes, the FIN indicates that the
+Objects that have the same Group ID and an Object ID larger than the last
+Object received on the stream do not exist.  This does not apply to
+RESET_STREAM or RESET_STREAM_AT.
```
```suggestion
For Type values where Contains End of Group is Yes, the subscriber can infer the
final Object in the Group when the data stream is terminated by a FIN.  In this
case, Objects that have the same Group ID and an Object ID larger than the last
Object received on the stream do not exist.  This does not apply when the data
stream is terminated with a RESET_STREAM or RESET_STREAM_AT.
```

draft-ietf-moq-transport.md
afrind: This is the FETCH object format, so it seems consistent.  I do have need of this field in #1331 however, so we should think about how to reconcile these.

draft-ietf-moq-transport.md
```diff
@@ -3683,9 +3678,11 @@ cannot infer any information about the existence of prior groups (see
 {{group-ids}}).
 
 This extension can be added by the Original Publisher, but MUST NOT be added by
-relays. This extension MUST NOT be modified or removed.
+relays. This extension MAY be removed by relay when the object in question is
+served via FETCH, but it MUST NOT be modified or removed otherwise.
```
afrind: It does not seem harmful to me to leave this extension in.  First, it could be in the Immutable section.  Second, the FETCH may not cover the full range of non-existent objects conveyed by the header.  eg: Group 10 has PGIDG=3 and I get a FETCH for group 10.  Does the relay need to serialize the extension?

draft-ietf-moq-transport.md
```diff
@@ -3756,7 +3751,8 @@ cannot infer any information about the existence of prior objects (see
 {{model-object}}).
 
 This extension can be added by the Original Publisher, but MUST NOT be added by
-relays. This extension MUST NOT be modified or removed.
+relays. This extension MAY be removed by relay when the object in question is
+served via FETCH, but it MUST NOT be modified or removed otherwise.
```
afrind: Same comment as above.

draft-ietf-moq-transport.md
yekuiwang: Ahh, got it. Thanks!

draft-ietf-moq-transport.md
vasilvv: I think all malformed errors of this kind are now just a violation of the object uniqueness principle.

draft-ietf-moq-transport.md
```diff
@@ -3091,34 +3099,24 @@ NOT be processed by a relay. Only present when 'Object Status' is Normal (0x0).
...
-* 0x4 := Indicates End of Track. Group ID is either the largest Group produced
-         in this Track with Object ID one greater than the largest Object
-         produced in that Group, or Group ID is one greater than the largest
-         Group produced in this Track with Object ID zero. This status also
-         indicates the specified Group has ended. Publishers MUST NOT publish an
-         Object with a Location larger than this Location (see
-         {{malformed-tracks}}). This SHOULD be cached.
+* 0x3 := Indicates End of Group. Indicates that no objects with the specified
+         Group ID and the Object ID that is greater than or equal to the one
+         specified exist.
```
```suggestion
         specified exist in the group identified by the Group ID.
```

draft-ietf-moq-transport.md
suhasHere: Not sure why the suggestion to remove the extension

draft-ietf-moq-transport.md
```diff
@@ -366,6 +366,20 @@ payloads.
 
 Objects within a Group are in ascending order by Object ID.
 
+From the perspective of a subscriber or a cache, an Object can be in three
+possible states:
+
+1. The Object is known to not exist. This state is permanent.
```
suhasHere: I think adding how it is known to not exist will help clarify

afrind:
@vasilvv To clarify text on extension removal

draft-ietf-moq-transport.md
afrind: @vasilvv agreed to clarify this bit

vasilvv:
Hopefully addressed all of the review comments.

draft-ietf-moq-transport.md
afrind: Below we have language that if you receive an object that was previously known not to exist, it's not a malformed error and you consider this to be a result of re-ordering, and ignore the object.  

But what about this:

Subsrcribe
Object 10,1, Object Gap=1
...
Later
Fetch 10,0
Object 10,0

This still seems malformed

draft-ietf-moq-transport.md
```diff
@@ -1296,20 +1304,23 @@ Objects or MAY use them to update certain cached fields. Implementations that
...
+of Group status, or a Prior Group ID Gap extension could be removed in FETCH,
+where it's redundant.
 
-Note that due to reordering, an implementation can receive a duplicate Object
-with a status of Normal, End of Group or End of Track after receiving a previous
-status of Object Does Not Exist.  The endpoint SHOULD NOT cache or forward the
-duplicate object in this case.
+Note that due to reordering, an implementation can receive an Object after
+receiving an indication that the Object in question does not exist.  The
+endpoint SHOULD NOT cache or forward the duplicate object in this case.
```
```suggestion
endpoint SHOULD NOT cache or forward the object in this case.
```

draft-ietf-moq-transport.md
afrind: We have conflicting language in the group/object ID gap sections that say it's malformed to receive a normal object in a previously defined gap.

draft-ietf-moq-transport.md
```diff
@@ -366,6 +366,22 @@ payloads.
...
+1. The Object is known to not exist. This state is permanent. MOQT has multiple
+   ways to communicate that a certain range of objects does not exist,
+   including the Object Status field, and the use of gaps in FETCH responses.
+2. The Object is known to exist. From this state, it can transition to not
+   existing, but not vice versa.
+3. The state of the Object is unknown, either because it has not been yet
+   received, or it has not been produced yet.
+
+Whenever the publisher communicates that certain objects do not exist, this
+fact is expressed as a contiguous range of non-existent objects; MOQT
```
```suggestion
Whenever the publisher communicates that certain objects do not exist, this
fact is expressed as a contiguous range of non-existent objects and
by include extension headers indicating the group/object gaps; MOQT
```

ianswett (approved):

Issue #1262, opened 2025-09-23
No way to indicate that PUBLISHes from SUBSCRIBE_NAMESPACE were blocked


ianswett:
There is a generic REQUESTS_BLOCKED frame, but there's no way to specifically indicate that a PUBLISH couldn't be sent as a result.

Is that ok?  Also, if you become unblocked, does the publisher then send any queued PUBLISH messages?  If only one more PUBLISH is possible, does the priority of the SUBSCRIBE_NAMESPACE factor into the decision of which PUBLISH to send first?

afrind:
Assuming we split wildcard subscribe and discovery into two operations, a subscriber can detect this is the case by observing a) the peer is out of request IDs and b) there are more discovered tracks than subscriptions in the given namespace.

Parking this pending resolution of #843

afrind:
If PUBLISH is blocked on request IDs, and more become available, does a publisher need to queue up tracks pending for SUBSCRIBE_NAMESPACE?

afrind:
Can a publisher end a SUBSCRIBE_NAMESPACE because there are too many queued PUBLISH messages?

ianswett:
Now that #1344 has landed, I lean towards adding a PUBLISH_BLOCKED (or PUBLISH_FAILED, etc) message on the response stream to SUBSCRIBE_NAMESPACE to tell the Subscriber when this occurs.  This avoids needing to buffer these.  If a Subscriber receives this message, it knows it will not receive the PUBLISH and if it wants the Track, it needs to SUBSCRIBE.

I think the PUBLISH_BLOCKED could be the same as the PUBLISH message, since I believe we want Track Extension Headers included?

I can also see the Filtered (ie: Top N) SUBSCRIBE_NAMESPACE using this mechanism as a lighter weight way to inform the Subscriber about a Track's existence without creating any state.

ianswett:
Unparking this so we can discuss it when we have time.  We should at least be clear about what should happen today in the current draft.

Issue #1168, opened 2025-08-22
Ephemeral vs Long-lived Request IDs


afrind:
All requests have a request ID that is used as a correlator between the request its response.  SUBSCRIBE and FETCH also use this Request ID for subsequent purposes (SUBSCRIBE_UPDATE, PUBLISH_DONE, FETCH_CANCEL) and it has a longer lifetime.  This means it's possible to consume all Request IDs with subscriptions and not have any credit left to send other messages like FETCH, TRACK_STATUS or PUBLISH_NAMESPACE.  This is exacerbated by #1106 which means you might not even be able to update subscriptions you created.

We could

1) Leave this as is
2) Split the ID space
3) Something else?

sharmafb:
I'm not sure that I see why this is an issue. If the peer updates the MAX_REQUEST_ID appropriately (as it should), then this won't be a problem, right?

yekuiwang:
My feeling is that option 1 ( leave this as is) is likely the best. The value space of Request ID can be as large as 2^62, thus careful implementations should be able to handle Request IDs in a workable fashion, noting that there is the mechanism to increase max Request ID already specified through the MAX_REQUEST_ID message. Also note that to do option 2 (Split the ID space) would not only require implementations to have separable tables to track different types of Request IDs, but also the MAX_REQUEST_ID message need to be updated to add request types. Another consideration is that splitting the ID space does not seem save anything, on bits-on-the-wire efficiency or implementation complexities. Just my two cents.

afrind:
> I'm not sure that I see why this is an issue. If the peer updates the MAX_REQUEST_ID appropriately (as it should), then this won't be a problem, right?

The root of the issue is that Request ID is solving two problems:

1.  Limiting the number of requests the peer can send without more credit
2. Limiting the amount of state the receiver is committing to

For example, our implementation only releases request ID's taken by a SUBSCRIBE when the subscription completes.  If say, MAX_REQUEST_ID is 10 and the peer makes 10 open-ended subscriptions, we won't grant more requests unless they UNSUBSCRIBE.

fluffy:
I'm understand why we might want to limit the number of outstanding subscribe, publish, fetch etc but each of these should probably be it's own limit not merged together. Trying to limit the number of request_id seems the wrong way to do this. 

Overall, I very skeptical that these limits actually help with the ways a system can DDOS MoQ and we should have a serious design discussion about DOS attacks, what things can be billed for by relays, and how to deal with this.

[...]

PR #1332, opened 2025-10-22
Change SUBSCRIBE_UPDATE to REQUEST_UPDATE and expand ability to update

afrind:
This creates a mechanism to refresh auth credentials for PUBLISH, SUBSCRIBE_NAMESPACE and PUBLISH_NAMESPACE.

A side-effect is that it allows the subscriber to change the priority of a FETCH (we closed #1204 already but this would address it).

If we decide to address #1270, this will help.

Fixes: #1267 
Fixes: #1204

PR #1344, opened 2025-10-31
Put SUBSCRIBE_NAMESPACE on a stream, make Namespaces and PUBLISH independent

### LINKED ISSUE #843: ANNOUNCE via SUBSCRIBE_ANNOUNCES vs ANNOUNCE by itself
I feel like we should distinguish those two.  If I'm a publisher and I send an ANNOUNCE, I should get an OK or ERROR in response.  If I am a client who asks relay to SUBSCRIBE_ANNOUNCES, I should not have to OK/ERROR everything relay sends to me.

### LINKED ISSUE #1305: There is no way for the publisher to cancel SUBSCRIBE_NAMESPACE
Say that a subscriber sends SUBSCRIBE_NAMESPACE with a valid authorization and the publisher OKs it.

Some time later, the authorization expires. Alternatively, the publisher is a relay that has an upstream subscription to a namespace, but that session suddenly dies.

How can the publisher signal this? It can PUBLISH_NAMESPACE_DONE or PUBLISH_DONE anything that was delivered, but that's a different thing from canceling the subscribe.

### LINKED ISSUE #1310: PUBLISH_NAMESPACE is a control message, but doesn't change subscription state, and blocks other control messages
PUBLISH_NAMESPACE (formerly ANNOUNCE) is a control message, but doesn't directly alter a SUBSCRIBE or FETCH.

This seems odd to me, but more importantly, if I get a flood of PUBLISH_NAMESPACE messages on the control stream in response to a new SUBSCRIBE_NAMESPACE, that blocks all the other actual control stream traffic, some of which may be somewhat time-sensitive like UNSUBSCRIBE or SUBSCRIBE_UPDATE.  When one issues a SUBSCRIBE or FETCH, the Objects come on a different stream, preventing this problem.

As an associated point, I'm not sure if we need Parameters on PUBLISH_NAMESPACE, but that seems like a secondary concern.

### LINKED ISSUE #1348: No way to avoid PUBLISH when using SUBSCRIBE_NAMESPACE
> If the FORWARD parameter ([Section 9.2.1.10](https://www.ietf.org/archive/id/draft-ietf-moq-transport-15.html#forward-parameter)) is present in this message and equal to 0, PUBLISH messages resulting from this SUBSCRIBE_NAMESPACE will set the FORWARD parameter to 0. If the FORWARD parameter is equal to 1 or omitted from this message, PUBLISH messages resulting from this SUBSCRIBE_NAMESPACE will set the FORWARD parameter to 1, or indicate that value by omitting the parameter (see [Section 5.1](https://www.ietf.org/archive/id/draft-ietf-moq-transport-15.html#subscriptions)).

What this means is that `SUBSCRIBE_NAMESPACE` will trigger both `PUBLISH` and `PUBLISH_NAMESPACE`. There's no way to get only `PUBLISH_NAMESPACE`.

Currently, I have to send `SUBSCRIBE_NAMESPACE forward=0` and `PUBLISH_ERROR` all of the incoming `PUBLISH forward=0` messages. I don't want to learn about every (currently known) track within a namespace, and will `SUBSCRIBE` to tracks when I actually want them. I just want to learn about the existence of namespaces, not tracks.

ianswett:
I did not remove Request ID entirely in this PR, because it makes the most sense to do that all at once, but I did create new 'NAMESPACE' and 'NAMESPACE_DONE' messages that are separate from PUBLISH_NAMESPACE/PUBLISH_NAMESPACE_DONE and much simpler.

Saves bytes on the wire by not repeating the 'Track Namespace Prefix' for every Namespace.

Removes UNSUBSCRIBE_NAMESPACE because we can just close the stream.

Fixes #1348 
Fixes #1310
Fixes #1305
Fixes part of #1168 
Fixes #843

Further possible changes:
  1) Don't send REQUEST_ERROR on errors, QUIC RESET_STREAM with the application error code. Loses Reason Phrase
  2) Don't send REQUEST_OK on success.  Loses response parameters.
  3) Allow for subscribing to full Track names and not just Namespaces.
  4) Prohibit sending NAMESPACE_DONE before a NAMESPACE is received for a SUBSCRIBE_NAMESPACE
  5) Add a PUBLISH_BLOCKED sent on the response stream when a new PUBLISH couldn't be sent #1262 
  6) Drop the length from NAMESPACE and NAMESPACE_DONE

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -725,7 +725,10 @@ the endpoints exchange Setup messages ({{message-setup}}), followed by other
 messages defined in {{message}}.
 
 This specification only specifies two uses of bidirectional streams, the control
-stream and SUBSCRIBE_NAMESPACE. Objects are sent on unidirectional streams.
+stream, which begins with CLIENT_SETUP, and SUBSCRIBE_NAMESPACE. Bidirectional
+streams MUST NOT begin with any other message type unless negotiated. If they
+do, the peer MUST close the Session with a Protocol Violation. Objects are sent on
+unidirectional streams.
 
 A unidirectional stream containing Objects or bidirectional stream(s) containing a
 SUBSCRIBE_NAMESPACE could arrive prior to the control stream, in which case the


draft-ietf-moq-transport.md
```diff
@@ -1598,7 +1598,7 @@ and frequency of requests.  Request IDs for one endpoint increment independently
 from those sent by the peer endpoint.  The client's Request ID starts at 0 and
 are even and the server's Request ID starts at 1 and are odd.  The Request ID
 increments by 2 with each FETCH, SUBSCRIBE, SUBSCRIBE_UPDATE,
-SUBSCRIBE_NAMESPACE, PUBLISH, PUBLISH_NAMESPACE or TRACK_STATUS request.
```
afrind, line 1667: I think SUBSCRIBE_NAMESPACE still needs a Request ID, at least for now.

draft-ietf-moq-transport.md
```diff
@@ -2885,6 +2885,24 @@ PUBLISH_NAMESPACE Message {
...
+## NAMESPACE {#message-pub-ns}
+
+The NAMESPACE message is similar to the PUBLISH_NAMESPACE message, except
+it is in response to a SUBSCRIBE_NAMESPACE request. Because it is never
+sent on the control stream, it can use the same type.
+
+~~~
+NAMESPACE Message {
+  Type (i) = 0x6,
+  Length (16),
```
afrind: Aren't these 3 bytes completely redundant?

draft-ietf-moq-transport.md
```diff
@@ -2932,16 +2950,18 @@ PUBLISH_NAMESPACE_CANCEL Message {
 
 ## SUBSCRIBE_NAMESPACE {#message-subscribe-ns}
 
-The subscriber sends the SUBSCRIBE_NAMESPACE control message to a publisher to
-request the current set of matching published namespaces and `Established`
-subscriptions, as well as future updates to the set.
+The subscriber sends a SUBSCRIBE_NAMESPACE control message on a new
+bidirectional stream to a publisher to request the current set of matching
```
afrind, line 3071: How do you distinguish the control stream from this stream?

draft-ietf-moq-transport.md
```diff
@@ -2960,15 +2980,19 @@ SUBSCRIBE_NAMESPACE Message {
...
+ 
 * Parameters: The parameters are defined in {{version-specific-params}}.
 
-The publisher will respond with REQUEST_OK or
-REQUEST_ERROR.  If the SUBSCRIBE_NAMESPACE is successful, the publisher will
-immediately forward existing PUBLISH_NAMESPACE and PUBLISH messages that match
-the Track Namespace Prefix that have not already been sent to this subscriber.
-If the set of matching PUBLISH_NAMESPACE messages changes, the publisher sends
-the corresponding PUBLISH_NAMESPACE or PUBLISH_NAMESPACE_DONE message.
+The publisher will respond with REQUEST_OK or REQUEST_ERROR.  If the
```
afrind: I think you mean to say these will come on the response half of the bidi stream.

draft-ietf-moq-transport.md
```diff
@@ -2960,15 +2980,19 @@ SUBSCRIBE_NAMESPACE Message {
...
+SUBSCRIBE_NAMESPACE is successful, the publisher will send matching
+NAMESPACE messages on the response stream if they are requested.
```
afrind: Do we need or want a mechanism to get only the track _names_ in the same format as namespaces, or we think only PUBLISH is sufficient?

draft-ietf-moq-transport.md
```diff
@@ -2960,15 +2980,19 @@ SUBSCRIBE_NAMESPACE Message {
...
+Also, any matching PUBLISH messages without an active Subscription will be
```
afrind: s/active/\`Established\`/

draft-ietf-moq-transport.md
```diff
@@ -2960,15 +2980,19 @@ SUBSCRIBE_NAMESPACE Message {
   Track Namespace Prefix consisting of 0 or greater than than 32 Track Namespace
   Fields, it MUST close the session with a `PROTOCOL_VIOLATION`.
 
-
+* Subscribe Options: Allows subscribers to request PUBLISH (0x00),
+  NAMESPACE (0x01), or both (0x02) for a given SUBSCRIBE_NAMESPACE request.
```
afrind, line 3101: I assume I can't change this value later and prefix_overlap still applies.  Therefore, if I want to go from NAMESPACE to both, I need to UNSUBSCRIBE_NS and SUBSCRIBE_NS again?

draft-ietf-moq-transport.md
afrind: Did you want to remove the common prefix with the one in the request?

draft-ietf-moq-transport.md
afrind: Ah the response stream can have both `NAMESPACE` and `PUBLISH_NAMESPACE_DONE` so you need type (and also `REQUEST_OK`/`ERROR`, maybe?).  Length feels useless here, but I can live with it.

draft-ietf-moq-transport.md
```diff
@@ -2885,6 +2885,24 @@ PUBLISH_NAMESPACE Message {
 
 * Parameters: The parameters are defined in {{version-specific-params}}.
 
+## NAMESPACE {#message-pub-ns}
```
```suggestion
## NAMESPACE_NOTIFY {#message-pub-ns}
```

draft-ietf-moq-transport.md
ianswett: Agreed, I kept it below, at least for now.

draft-ietf-moq-transport.md
```diff
@@ -1598,7 +1598,7 @@ and frequency of requests.  Request IDs for one endpoint increment independently
 from those sent by the peer endpoint.  The client's Request ID starts at 0 and
 are even and the server's Request ID starts at 1 and are odd.  The Request ID
 increments by 2 with each FETCH, SUBSCRIBE, SUBSCRIBE_UPDATE,
-SUBSCRIBE_NAMESPACE, PUBLISH, PUBLISH_NAMESPACE or TRACK_STATUS request.
+PUBLISH, PUBLISH_NAMESPACE or TRACK_STATUS request.
```
```suggestion
SUBSCRIBE_NAMESPACE, PUBLISH, PUBLISH_NAMESPACE or TRACK_STATUS request.
```

draft-ietf-moq-transport.md
ianswett: I don't have a strong opinion, but NAMESPACE is shorter

draft-ietf-moq-transport.md
```diff
@@ -2885,6 +2885,24 @@ PUBLISH_NAMESPACE Message {
 
 * Parameters: The parameters are defined in {{version-specific-params}}.
 
+## NAMESPACE {#message-pub-ns}
+
+The NAMESPACE message is similar to the PUBLISH_NAMESPACE message, except
+it is in response to a SUBSCRIBE_NAMESPACE request. Because it is never
+sent on the control stream, it can use the same type.
```
```suggestion
sent on the control stream, it can use the same type value.  Because all messages are in
response to a single SUBSCRIBE_NAMESPACE, only the tuples of the namespace after the
namespace being subscribed to are included.
```

draft-ietf-moq-transport.md
ianswett: Yes, I think I need a type.  The Length is a bit more obviously redundant, but that goes back to the TLV vs TV conversation.  I think there's a good argument that this isn't on the control stream and one could use TV even if we're using TLV elsewhere.

draft-ietf-moq-transport.md
ianswett: Yes, updated.

draft-ietf-moq-transport.md
ianswett: I think it still applies.  I wasn't intending to solve that problem.

draft-ietf-moq-transport.md
ianswett: I think we do want that, but I was trying to keep this small.  Happy to add it into this PR if we're all sure we want it.

draft-ietf-moq-transport.md
ianswett: It's a new bidi stream and it starts with the SUBSCRIBE_NAMESPACE message.

draft-ietf-moq-transport.md
```suggestion
Also, any matching PUBLISH messages without an `Established` Subscription will be
```

draft-ietf-moq-transport.md
ianswett: True, I updated the text above.  This text is fairly duplicative I think, but I guess we can move it in another PR?

ianswett:
> What does the subscriber do with the write half of the bidi stream after the SUBSCRIBE_NAMESPACE? FIN, or leave open?

I'm inclined to leave it open, at least until we decide if want to send REQUEST_UPDATE on it.

> 
> How does this impact UNSUBSCRIBE_NAMESPACE, if at all?

I removed it in the more recent update, because it wasn't needed.
> 
> If we introduce REQUEST_UPDATE (#1332), where does that go?

On the open bidi stream.

draft-ietf-moq-transport.md
ianswett: Though maybe the prefix overlap problem isn't so much a problem now that it's clear which SUBSCRIBE_NAMESPACE it's in response to?

draft-ietf-moq-transport.md
```diff
@@ -1571,16 +1572,14 @@ The following Message Types are defined:
 |-------|-----------------------------------------------------|
 | 0xD   | TRACK_STATUS ({{message-track-status}})             |
 |-------|-----------------------------------------------------|
-| 0x6   | PUBLISH_NAMESPACE  ({{message-pub-ns}})             |
+| 0x6   | PUBLISH_NAMESPACE/NAMESPACE  ({{message-pub-ns}})   |
```
afrind: I don't think reusing this code point is helping anyone

draft-ietf-moq-transport.md
afrind: This stream is currently a mini control stream that supports:

REQUEST_OK
REQUEST_ERROR
NAMESPACE
PUBLISH_NAMESPACE_DONE

There's some nice code reusability there, enough that I can live with the length redundancy.  The alternative is quite a bit more spec text, but I can live with that too if folks are concerned.

draft-ietf-moq-transport.md
afrind: So if the first stream I receive starts with SUBSCRIBE_NAMESPACE, I hold it pending one that starts with CLIENT_SETUP?

draft-ietf-moq-transport.md
afrind: The reasoning behind prefix overlap was to prevent sending the same NAMESPACE to two subscribers, since it's redundant.

draft-ietf-moq-transport.md
```diff
@@ -2960,15 +2983,20 @@ SUBSCRIBE_NAMESPACE Message {
...
 * Parameters: The parameters are defined in {{version-specific-params}}.
 
-The publisher will respond with REQUEST_OK or
-REQUEST_ERROR.  If the SUBSCRIBE_NAMESPACE is successful, the publisher will
-immediately forward existing PUBLISH_NAMESPACE and PUBLISH messages that match
-the Track Namespace Prefix that have not already been sent to this subscriber.
-If the set of matching PUBLISH_NAMESPACE messages changes, the publisher sends
-the corresponding PUBLISH_NAMESPACE or PUBLISH_NAMESPACE_DONE message.
+The publisher will respond with REQUEST_OK or REQUEST_ERROR on the response half
+of the bidi stream.  If the SUBSCRIBE_NAMESPACE is successful, the publisher will
```
afrind: spell out bidirectional

draft-ietf-moq-transport.md
```diff
@@ -2960,15 +2983,20 @@ SUBSCRIBE_NAMESPACE Message {
...
+send matching NAMESPACE messages on the response stream if they are requested.
+If it is an error, the stream will be immediately closed via FIN.
+Also, any matching PUBLISH messages without an `Established` Subscription will be
+sent on the control stream. When there are changes to the namespaces or
```
afrind, line 3110: Only if options is PUBLISH or both

draft-ietf-moq-transport.md
```diff
@@ -1103,26 +1103,27 @@ The syntax of these messages is described in {{message}}.
...
 forward the result to the application, so the application can decide which other
 publishers to contact, if any.
 
-An UNSUBSCRIBE_NAMESPACE withdraws a previous SUBSCRIBE_NAMESPACE. It does not
-prohibit original publishers from sending further PUBLISH_NAMESPACE or PUBLISH
-messages, but relays MUST NOT send any further PUBLISH messages to a client
-without knowing the client is interested in and authorized to receive the
-content.
+A SUBSCRIBE_NAMESPACE can be cancelled by the subscriber closing the stream with
+either a FIN or RESET_STREAM. Cancelling does not prohibit original publishers
```
afrind: It can be cancelled by _either_ endpoint by sending a FIN or RESET.

draft-ietf-moq-transport.md
ianswett: Sure, but it would no longer be a correctness problem.  Before, I believe there was ambiguity.

draft-ietf-moq-transport.md
ianswett: Again, I'm trying to minimize spec churn.

draft-ietf-moq-transport.md
ianswett: Suggestions welcome

draft-ietf-moq-transport.md
ianswett: Does that need to be stated?

draft-ietf-moq-transport.md
ianswett: Without 0RTT clearly described, which it isn't today, I'm unsure how that could happen?

draft-ietf-moq-transport.md
```suggestion
of the stream.  If the SUBSCRIBE_NAMESPACE is successful, the publisher will
```

draft-ietf-moq-transport.md
```suggestion
A SUBSCRIBE_NAMESPACE can be cancelled by closing the stream with
either a FIN or RESET_STREAM. Cancelling does not prohibit original publishers
```

draft-ietf-moq-transport.md
afrind: Yes but the formats are different, and they mean different things.  Just use a new code point.

draft-ietf-moq-transport.md
kixelated: Yeah @afrind if right.

It depends on if the QUIC library returns streams in stream_id order, but any bidirectional stream should be buffered until the CLIENT_SETUP is processed. In fact, the same is true for any unidirectional streams if pipelining a PUBLISH is valid (debatable with max_request_id), but that's a lot easier to implement.

draft-ietf-moq-transport.md
```diff
@@ -2932,16 +2954,18 @@ PUBLISH_NAMESPACE_CANCEL Message {
...
+published namespaces and/or `Established` subscriptions, as well as future
+updates to the set.
 
 ~~~
 SUBSCRIBE_NAMESPACE Message {
   Type (i) = 0x11,
   Length (16),
   Request ID (i),
   Track Namespace Prefix (..),
+  Subscribe Options (i),
```
kixelated, line 3081: Should this be a parameter? Could it have a reasonable default?

draft-ietf-moq-transport.md
ianswett: I did in the updated PR.

draft-ietf-moq-transport.md
ianswett: Another question is whether it would be shared with any other message, and I don't think it would be?

draft-ietf-moq-transport.md
ianswett: I think it's worth considering, but we can make it separable.

draft-ietf-moq-transport.md
ianswett: Thanks, I understand the issue now.  Yes, I can call out this potential issue.

ianswett: PTAL at https://github.com/moq-wg/moq-transport/pull/1344/commits/2465b19681f2744448debf80fd0676088eae7147

draft-ietf-moq-transport.md
```diff
@@ -1103,26 +1106,27 @@ The syntax of these messages is described in {{message}}.
 
 If the subscriber is aware of a namespace of interest, it can send
 SUBSCRIBE_NAMESPACE to publishers/relays it has established a session with. The
-recipient of this message will send any relevant PUBLISH_NAMESPACE,
+recipient of this message will send any relevant NAMESPACE,
 PUBLISH_NAMESPACE_DONE or PUBLISH messages for that namespace, or more specific
```
```suggestion
recipient of this message will send any relevant NAMESPACE,
NAMESPACE_DONE or PUBLISH messages for that namespace, or more specific
```

afrind:
Discussed 11/13 with Authors/Editors -- plan to present to the wg at the interim

draft-ietf-moq-transport.md
```diff
@@ -2894,6 +2900,27 @@ PUBLISH_NAMESPACE Message {
...
+in the 'Track Namespace Suffix'.
+
+~~~
+NAMESPACE Message {
+  Type (i) = 0x8,
+  Length (16),
+  Track Namespace Suffix (..),
+}
+~~~
+{: #moq-transport-pub-ns-format title="MOQT NAMESPACE Message"}
```
```suggestion
{: #moq-transport-ns-format title="MOQT NAMESPACE Message"}
```

afrind:
Presented at 11/17 Interim:

Plan is to proceed with this design.

draft-ietf-moq-transport.md
```diff
@@ -684,10 +684,13 @@ The first stream opened is a client-initiated bidirectional control stream where
...
 
-This specification only specifies a single use of bidirectional streams. Objects
-are sent on unidirectional streams.  Because there are no other uses of
-bidirectional streams, a peer MAY close the session as a `PROTOCOL_VIOLATION` if
-it receives a second bidirectional stream.
+This specification only specifies two uses of bidirectional streams, the control
+stream and SUBSCRIBE_NAMESPACE. Objects are sent on unidirectional streams.
+
+A unidirectional stream or bidirectional stream containing a SUBSCRIBE_NAMESPACE
+could arrive prior to the control stram, in which case the data SHOULD be buffered.
```
```suggestion
could arrive prior to the control stream, in which case the data SHOULD be buffered
until the control stream arrives and setup is complete.
```

draft-ietf-moq-transport.md
```diff
@@ -1104,26 +1107,27 @@ The syntax of these messages is described in {{message}}.
 
 If the subscriber is aware of a namespace of interest, it can send
 SUBSCRIBE_NAMESPACE to publishers/relays it has established a session with. The
-recipient of this message will send any relevant PUBLISH_NAMESPACE,
-PUBLISH_NAMESPACE_DONE or PUBLISH messages for that namespace, or more specific
+recipient of this message will send any relevant NAMESPACE,
+NAMESPACE_DONE or PUBLISH messages for that namespace, or more specific
 part of that namespace.  This includes echoing back PUBLISH or PUBLISH_NAMESPACE
 messages to the endpoint that sent them.  If an endpoint accepts its own
```
afrind: PUBLISH_NAMESPACE is not longer echoed?

draft-ietf-moq-transport.md
```diff
@@ -1104,26 +1107,27 @@ The syntax of these messages is described in {{message}}.
...
 PUBLISH, this behaves as self-subscription described in {{subscriptions}}.
 
-A publisher MUST send exactly one REQUEST_OK or
-REQUEST_ERROR in response to a SUBSCRIBE_NAMESPACE. The subscriber
-SHOULD close the session with a protocol error if it detects receiving more than
-one.
+The subscriber sends SUBSCRIBE_NAMESPACE on a new bidirectional stream and the
+publisher MUST send REQUEST_OK or REQUEST_ERROR as the first message on the
+bidirectional stream in response to a SUBSCRIBE_NAMESPACE. The subscriber
```
```suggestion
publisher MUST send exactly one REQUEST_OK or REQUEST_ERROR as the first message on the
bidirectional stream in response to a SUBSCRIBE_NAMESPACE. The subscriber
```

draft-ietf-moq-transport.md
```diff
@@ -2912,6 +2939,27 @@ PUBLISH_NAMESPACE_DONE Message {
 * Request ID: The Request ID of the PUBLISH_NAMESPACE that is being terminated. See
   {{message-subscribe-req}}.
 
+## NAMESPACE_DONE {#message-namespace-done}
+
+The publisher sends the `NAMESPACE_DONE` control message to indicate its
+intent to stop serving new subscriptions for tracks within the provided Track
+Namespace. All NAMESPACE messages are in response to a SUBSCRIBE_NAMESPACE,
```
afrind: This bit about NAMESPACE messages applies to both NS and NS_DONE, and is in the NS_DONE section.

draft-ietf-moq-transport.md
```diff
@@ -2894,6 +2900,27 @@ PUBLISH_NAMESPACE Message {
...
+  Type (i) = 0x8,
+  Length (16),
+  Track Namespace Suffix (..),
+}
+~~~
+{: #moq-transport-ns-format title="MOQT NAMESPACE Message"}
+
+* Track Namespace Suffix: Specifies the final portion of a track's
+  namespace as defined in {{track-name}}. The namespace begins with the
+  'Track Namespace Prefix' specified in {message-subscribe-ns}.
```
afrind: "final portion" is a bit under defined. Maybe 

Specifies the final portion of a track's namespace as defined in {{track-name}} after removing namespace tuples included in `Track Namespace Prefix`...

afrind (commented):
This looks close

draft-ietf-moq-transport.md
ianswett: I think it probably should be, to minimize any behavior change, but the PUBLISH_NAMESPACE would turn into a NAMESPACE message on the wire.

draft-ietf-moq-transport.md
ianswett: Isn't this obvious?

draft-ietf-moq-transport.md
```diff
@@ -2894,6 +2901,27 @@ PUBLISH_NAMESPACE Message {
...
```
```suggestion
  namespace as defined in {{track-name}} after removing namespace tuples included in
  'Track Namespace Prefix' {message-subscribe-ns}.
```

draft-ietf-moq-transport.md
ianswett: Fair point, any suggestions?  Just remove it?

draft-ietf-moq-transport.md
```diff
@@ -2941,16 +2990,18 @@ PUBLISH_NAMESPACE_CANCEL Message {
...
+published namespaces and/or `Established` subscriptions, as well as future
+updates to the set.
 
 ~~~
 SUBSCRIBE_NAMESPACE Message {
   Type (i) = 0x11,
   Length (16),
   Request ID (i),
   Track Namespace Prefix (..),
+  Subscribe Options (i),
```
suhasHere, line 3081: Can one send 2 subscribe_namespace with 2 different options

draft-ietf-moq-transport.md
```diff
@@ -2912,6 +2941,27 @@ PUBLISH_NAMESPACE_DONE Message {
 * Request ID: The Request ID of the PUBLISH_NAMESPACE that is being terminated. See
   {{message-subscribe-req}}.
 
+## NAMESPACE_DONE {#message-namespace-done}
+
+The publisher sends the `NAMESPACE_DONE` control message to indicate its
+intent to stop serving new subscriptions for tracks within the provided Track
+Namespace. All NAMESPACE messages are in response to a SUBSCRIBE_NAMESPACE,
```
```suggestion
Namespace. All NAMESPACE_DONE messages are in response to a SUBSCRIBE_NAMESPACE,
```

draft-ietf-moq-transport.md
afrind: No, this falls under duplicate prefix text elsewhere.

draft-ietf-moq-transport.md
```diff
@@ -701,10 +701,14 @@ The first stream opened is a client-initiated bidirectional control stream where
...
-bidirectional streams, a peer MAY close the session as a `PROTOCOL_VIOLATION` if
-it receives a second bidirectional stream.
+This specification only specifies two uses of bidirectional streams, the control
+stream and SUBSCRIBE_NAMESPACE. Objects are sent on unidirectional streams.
+
+A unidirectional stream or bidirectional stream containing a SUBSCRIBE_NAMESPACE
+could arrive prior to the control stream, in which case the data SHOULD be buffered
+until the control stream arrives and setup is complete.
+If an implementation does not want to buffer, it MAY reset other bidirectional
+streams before the session and control stream are established.
```
vasilvv, line 737: This PR does not really provide a good way for implementations to figure out what stream is the control stream.

draft-ietf-moq-transport.md
```diff
@@ -1610,14 +1616,16 @@ The following Message Types are defined:
 |-------|-----------------------------------------------------|
 | 0x6   | PUBLISH_NAMESPACE  ({{message-pub-ns}})             |
 |-------|-----------------------------------------------------|
+| 0x8   | NAMESPACE  ({{message-namespace}})                  |
+|-------|-----------------------------------------------------|
 | 0x9   | PUBLISH_NAMESPACE_DONE  ({{message-pub-ns-done}})   |
 |-------|-----------------------------------------------------|
+| 0xE   | NAMESPACE_DONE  ({{message-namespace-done}})        |
+|-------|-----------------------------------------------------|
 | 0xC   | PUBLISH_NAMESPACE_CANCEL ({{message-pub-ns-cancel}})|
```
vasilvv, line 1657: It looks like we're repeating the mistake we had in HTTP/3 where we are adding streams without type identifiers and then have some semantic rules about what messages can appear where.  Can we have an explicit type for bidi streams instead?

afrind: @vasilvv the current PR is adequate, in that it either starts with CLIENT_SETUP or it starts with SUBSCRIBE_NAMESPACE.  This is extensible since have varint frame type so have unlimited ways to start streams.  It seems like it would be fairly redundant to have:

SETUP_STREAM
CLIENT_SETUP

and

SUB_NS_STREAM
SUB_NS

Right?

afrind: I guess the counter-argument is if in the future you want to start a stream with a non-message.  This is what webtransport ran into and it got a little ugly, but ultimately resolved it by reserving a frame type and giving it new semantics.

ianswett: I definitely don't want to repeat that mistake, but in this case we are adding type/message identifiers.

draft-ietf-moq-transport.md
ianswett: There is a type identifier at the beginning of the control stream, which is the SETUP messages.

draft-ietf-moq-transport.md
```diff
@@ -701,10 +701,14 @@ The first stream opened is a client-initiated bidirectional control stream where
...
 
-This specification only specifies a single use of bidirectional streams. Objects
-are sent on unidirectional streams.  Because there are no other uses of
-bidirectional streams, a peer MAY close the session as a `PROTOCOL_VIOLATION` if
-it receives a second bidirectional stream.
+This specification only specifies two uses of bidirectional streams, the control
+stream and SUBSCRIBE_NAMESPACE. Objects are sent on unidirectional streams.
+
+A unidirectional stream or bidirectional stream containing a SUBSCRIBE_NAMESPACE
+could arrive prior to the control stream, in which case the data SHOULD be buffered
```
```suggestion
A unidirectional stream containing Objects or bidirectional stream(s) containing a SUBSCRIBE_NAMESPACE
could arrive prior to the control stream, in which case the data SHOULD be buffered
```

draft-ietf-moq-transport.md
```diff
@@ -1133,26 +1137,28 @@ The syntax of these messages is described in {{message}}.
...
-one.
+recipient of this message will send any relevant NAMESPACE,
+NAMESPACE_DONE or PUBLISH messages for that namespace, or more specific
+part of that namespace.  This includes echoing back published Tracks and/or Track
+Namespaces under the SUBSCRIBE_NAMESPACE prefix to the endpoint that sent them.
+If an endpoint accepts its own PUBLISH, this behaves as self-subscription described
+in {{subscriptions}}.
+
+The subscriber sends SUBSCRIBE_NAMESPACE on a new bidirectional stream and the
+publisher MUST send REQUEST_OK or REQUEST_ERROR as the first message on the
```
```suggestion
publisher MUST send a single REQUEST_OK or REQUEST_ERROR as the first message on the
```

draft-ietf-moq-transport.md
```diff
@@ -1133,26 +1137,28 @@ The syntax of these messages is described in {{message}}.
...
 forward the result to the application, so the application can decide which other
 publishers to contact, if any.
 
-An UNSUBSCRIBE_NAMESPACE withdraws a previous SUBSCRIBE_NAMESPACE. It does not
-prohibit original publishers from sending further PUBLISH_NAMESPACE or PUBLISH
-messages, but relays MUST NOT send any further PUBLISH messages to a client
-without knowing the client is interested in and authorized to receive the
-content.
+A SUBSCRIBE_NAMESPACE can be cancelled by closing the stream with
+either a FIN or RESET_STREAM. Cancelling does not prohibit original publishers
```
suhasHere, line 1184:
```suggestion
A SUBSCRIBE_NAMESPACE can be cancelled by the creator the stream, by closing the stream with
either a FIN or RESET_STREAM. Cancelling does not prohibit original publishers
```

draft-ietf-moq-transport.md
```diff
@@ -3001,15 +3053,20 @@ SUBSCRIBE_NAMESPACE Message {
...
-If the set of matching PUBLISH_NAMESPACE messages changes, the publisher sends
-the corresponding PUBLISH_NAMESPACE or PUBLISH_NAMESPACE_DONE message.
+The publisher will respond with REQUEST_OK or REQUEST_ERROR on the response half
+of the stream.  If the SUBSCRIBE_NAMESPACE is successful, the publisher will
+send matching NAMESPACE messages on the response stream if they are requested.
+If it is an error, the stream will be immediately closed via FIN.
+Also, any matching PUBLISH messages without an `Established` Subscription will be
+sent on the control stream. When there are changes to the namespaces or
+subscriptions being published and the subscriber is subscribed to them,
+the publisher sends the corresponding NAMESPACE, PUBLISH_NAMESPACE_DONE,
```
```suggestion
the publisher sends the corresponding NAMESPACE, NAMESPACE_DONE,
```

draft-ietf-moq-transport.md
vasilvv: I can see that this works, but I don't see why we would want to do that.  If there are two types of streams, and they have basically disjoint message type list, why do we want them to have the same format?

draft-ietf-moq-transport.md
```diff
@@ -724,10 +724,14 @@ The first stream opened is a client-initiated bidirectional control stream where
...
-This specification only specifies a single use of bidirectional streams. Objects
-are sent on unidirectional streams.  Because there are no other uses of
-bidirectional streams, a peer MAY close the session as a `PROTOCOL_VIOLATION` if
-it receives a second bidirectional stream.
+This specification only specifies two uses of bidirectional streams, the control
+stream and SUBSCRIBE_NAMESPACE. Objects are sent on unidirectional streams.
+
+A unidirectional stream containing Objects or bidirectional stream(s) containing a SUBSCRIBE_NAMESPACE
+could arrive prior to the control stream, in which case the data SHOULD be buffered
+until the control stream arrives and setup is complete.
```
```suggestion
A unidirectional stream containing Objects or bidirectional stream(s) containing a
SUBSCRIBE_NAMESPACE could arrive prior to the control stream, in which case the
data SHOULD be buffered until the control stream arrives and setup is complete.
```

draft-ietf-moq-transport.md
ianswett: It can be cancelled by either the creator or the peer.

afrind:
Discussed with Authors/Editors:

1) make it clear that there are two (for now) kinds of bidi streams, and the first byte identifies what kind it is (CLIENT_SETUP or SUBSCRIBE_NAMESPACE).
2) clearly define which control messages are allowed on each kind of stream -- some messages can only go on one or the other and some can go on both.
3) add normative text what to do if you receive a bidi stream with type that doesn't match one of the valid ones.

ianswett:
I addressed 1 and 3 and filed an issue (#1387) for 2.

Issue #1103, opened 2025-07-25
Delivery timeout, Datagrams, and Caching


ianswett:
The draft is quite clear that even if the forwarding preference is Datagram, the Delivery Timeout applies:
> "The start time for the timeout is based on when the beginning of the Object is received, and does not depend upon the forwarding preference."

So if I have no delivery timeout, I expect to get all the Objects.  But then there's text about how a datagram might be dropped along the path if a hop doesn't have sufficient MTU.  Of course publishers will try to avoid that, but this feels like a "You'll get all the Objects, except you might not and you'll never know what."

I expect this to create challenges for caches, unless publishers are very good at indicating all Object ID gaps, Group ID gaps, and end of Group.

I'll note this also means I can have (almost) fully reliable Datagram subscriptions and fully unreliable Subgroup subscriptions.  Both are quite implementable, but if that's the case, why have a Datagram forwarding preference at all?

afrind:
I'm not sure exactly what change you are asking for in this issue, and how it relates to the title.

A "fully reliable datagram subscription" would require a QUIC stack that supports retransmission of datagrams, which is rare, and I don't think we want to make this a requirement (yes, I know Google's can do this).

I think it's known that with DELIVERY_TIMEOUT, subgroups can be dropped or reset, but when it's not present, the intention is that you will get everything or your subscription will be dropped with TOO_FAR_BEHIND.  Don't use datagrams or delivery timeout if you want to get everything.

> why have a Datagram forwarding preference at all?

I think folks have made it clear that the three properties of a datagram (< MTU, not flow controlled, not retransmitted) are ideal for audio, which is part of the core MOQT use case.  That said, I've started to wonder why we still disallow mixing and matching subgroups and datagrams within a single Track.  My implementation does not enforce this requirement on the sender or receiver.  In MOQT the currency is Objects, and as long as you respect the per-object Forwarding Preference at Relays, nothing will break.  Someone approached me in Madrid asking specifically for relaxing this restriction (Huawei?). 

Does anyone have a strong reason to disallow mixing and matching?

ianswett:
In discussion, there was also disagreement about whether Datagrams are ever retransmitted at any point in the path, particularly if there's a long or no delivery timeout.

yekuiwang:
[...]

PR #1350, opened 2025-11-04
Enable mixing datagrams with streams in one track

Weihang1996:
As discussed in Toronto, Datagrams to be mixed with Subgroups in one single track should be allowed. (see #1103) 
This PR fixes the related descriptions in the draft.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1180,7 +1180,7 @@ MOQT maintains priorities between different schedulable objects.
 A schedulable object in MOQT is either:
 
 1. The first or next Object in a Subgroup that is in response to a subscription.
-2. An Object with delivery preference Datagram.
+2. An Object with forwarding preference Datagram.
 3. An Object in response to a FETCH where that Object is the next
    Object in the response.
 


ianswett (approved):
Thanks for the PR!

draft-ietf-moq-transport.md
```diff
@@ -3072,13 +3067,10 @@ A canonical MoQ Object has the following information:
...
-an object. The preferences are Subgroup and Datagram.  Note that the Original
-Publisher determines the Forwarding Preference for the entire Track, and is a
-Track property that is implicitly signaled by the delivery of any Object using
-either Subgroups or Datagrams. Once the property is established for one Object
-of a Track, the same value MUST be used for all Objects of the Track.
-In a subscription, an Object MUST be sent according to its `Object Forwarding
-Preference`.
+an object. The preferences are Subgroup and Datagram.  `Object Forwarding
+Preference` is the property of an individual Object, and is not required to
+remain consistent in one Track.  In a subscription, an Object MUST be sent
```
```suggestion
Preference` is a property of an individual Object and can vary among
Objects in the same Track.  In a subscription, an Object MUST be sent
```

draft-ietf-moq-transport.md
```diff
@@ -1179,8 +1177,7 @@ MOQT maintains priorities between different schedulable objects.
 A schedulable object in MOQT is either:
 
 1. The first or next Object in a Subgroup that is in response to a subscription.
-2. An Object in response to a subscription that belongs to a Track with
-   delivery preference Datagram.
+2. An Object in response to a subscription with delivery preference Datagram.
```
vasilvv: I think this should still be "An Object with delivery preference Datagram"?

draft-ietf-moq-transport.md
```diff
@@ -535,8 +535,6 @@ include:
    End of Track.
 9. The same Object is received more than once with different Payload or
     other immutable properties.
-10. An Object is received with a different Forwarding Preference than previously
-    observed from the same Track.
```
vasilvv, line 547: I assume that it is still illegal to change delivery preference of the same object (i.e. to republish a datagram on a subgroup or vice versa)?

draft-ietf-moq-transport.md
```diff
@@ -1306,8 +1303,7 @@ fields that can be updated are the following:
    to the constraints of the specific header extension.
 
 An endpoint that receives a duplicate Object with an invalid Object Status
-change, or with a different Forwarding Preference, Subgroup ID, Priority or
-Payload MUST treat the track as Malformed.
+change, Subgroup ID, Priority or Payload MUST treat the track as Malformed.
```
vasilvv: See my comment above.  I don't see why we need to make the forwarding preference of an individual object mutable, given that almost everything about objects is otherwise immutable.

draft-ietf-moq-transport.md
suhasHere: +1 i would like this to be resolved before merging this PR

Weihang1996: I agree that we should keep the forwarding preference immutable, but Clause 10 makes me feel like it's saying "an object is received with a different forwarding preference than the forwarding preference observed from the previous objects in the same track". Maybe we should make it clearer. 
Is "A duplicated Object is received with a different Forwarding Preference" better?

draft-ietf-moq-transport.md
Weihang1996: Yes, I agree.

draft-ietf-moq-transport.md
Weihang1996: Yes, I think we should keep the original text here.

Weihang1996:
> It is not entirely clear to me what the subgroup IDs associated with datagrams mean in the world where we mix those.

This commit has been updated to address Victor's comments.

afrind:
> I think we need to be clearer about how prioritization is supposed to work for a Group with both Datagrams and Subgroups?

The next object to send on a subgroup and a datagram (in the same group) are both schedulable. Since they have the same pub priority, it should fallback to subgroupID vs object ID?  Now, that could be weird, but we're also on the 4th tiebreaker, and they could be the same (!).

draft-ietf-moq-transport.md
```diff
@@ -543,8 +543,7 @@ include:
    End of Track.
 9. The same Object is received more than once with different Payload or
     other immutable properties.
-10. An Object is received with a different Forwarding Preference than previously
-    observed from the same Track.
+10. A duplicated Object is received with a different Forwarding Preference.
```
```suggestion
10. An Object is received with a different Forwarding Preference than previously
    observed.
```

draft-ietf-moq-transport.md
```diff
@@ -1235,8 +1233,8 @@ the objects SHOULD be selected as follows:
    decide the one that is scheduled to be sent first.
 3. If two objects in response to the same request have the same subscriber
    and publisher priority and belong to the same group of the same track, the
-   one with **the lowest Subgroup ID** (for tracks with delivery preference
-   Subgroup), or **the lowest Object ID** (for tracks with delivery preference
+   one with **the lowest Subgroup ID** (for objects with delivery preference
+   Subgroup), or **the lowest Object ID** (for objects with delivery preference
    Datagram) is scheduled to be sent first.
```
```suggestion
   Datagram) is scheduled to be sent first.  If the two objects have
   different Forwarding Preferences the order is implementation dependent.
```

draft-ietf-moq-transport.md
```diff
@@ -1235,9 +1234,10 @@ the objects SHOULD be selected as follows:
    decide the one that is scheduled to be sent first.
 3. If two objects in response to the same request have the same subscriber
    and publisher priority and belong to the same group of the same track, the
-   one with **the lowest Subgroup ID** (for tracks with delivery preference
-   Subgroup), or **the lowest Object ID** (for tracks with delivery preference
-   Datagram) is scheduled to be sent first.
+   one with **the lowest Subgroup ID** (for objects with delivery preference
+   Subgroup), or **the lowest Object ID** (for objects with delivery preference
```
```suggestion
   one with **the lowest Subgroup ID** (for objects with forwarding preference
   Subgroup), or **the lowest Object ID** (for objects with forwarding preference
```

draft-ietf-moq-transport.md
```diff
@@ -1180,8 +1180,7 @@ MOQT maintains priorities between different schedulable objects.
 A schedulable object in MOQT is either:
 
 1. The first or next Object in a Subgroup that is in response to a subscription.
-2. An Object in response to a subscription that belongs to a Track with
-   delivery preference Datagram.
+2. An Object with delivery preference Datagram.
```
```suggestion
2. An Object with forwarding preference Datagram.
```

ianswett (approved):

suhasHere (approved):

Issue #1099, opened 2025-07-22
Common way to format namespace tuple and track names in logs and drafts


fluffy:
Writing things that look like HTP URI which are not URI just causes confusion

fluffy:
I will point out the file https://datatracker.ietf.org/doc/draft-jennings-moq-file/ also needs this and should alight with it.

afrind:
Subject to #764, I propose this:

`("a", "b", "c"), "name"`

Of course, all of those fields are binary data with no restricted characters, but for readability, please don't use example namespace fields of track names with unprintable characters or `()", `

afrind:
In MoQT we currently have these examples:

Full Track Name: 
`namespace=(foo,bar), track=x`

Namespaces:
`(foo)`
`(foo, bar)`
`"example.com", "meeting=123", "participant=100")`
`("example.com", "meeting=123") `

wilaw:
Since MOQT allows arbitrary binary blobs for namespace tuple fields and names, we need to specify how these should be encoded as strings. base64 would seem a candidate in this regard. 

Additionally, to save space in log lines, can we forego the wrapping quotes? For example

(a,b,c),name
(YWxwaGE=,YnJhdm8=),Y2hhcmxpZQ==

Obviously this requires tuple fields without ( or , characters , but that is satisfied by base64 for binary fields and a simple constraint on text fields.

fluffy:
I would prefer we stay in the set of characters you can cut and paste on standard shells and they do not interpret them as this happens constantly when working with CLI tools. 

I would prefer that normal ascii characters were  non transformed for a tuple of "example" 

I would prefer character set that was not weird to embed in a HTTP URL.

This leads me to something like - as separators between tuples and percent encoding for non allowed values. 

So something like chars other than a-z A-Z 0-9 get percent encoded and - in between. 

I can see adding in perhaps a few more character out out the URL safe alphabet

I had base64 before but that is just a PITA to go from the base64 to what the name was when debugging particularly given usage where big parts of the name are pretty readable in ascii .

Please do consider the files use case as well in the draft I pointed to in first post.

vasilvv:
As a data point, our implementation currently uses `("a", "b", "c")`, where individual strings are C-escaped, and the track name is the last entry in the tuple (we don't parse that, that's the output format).

fluffy:
On call today I agree to write a PR for this.

Issue #1234, opened 2025-09-23
Restrict strings that will be placed in log to be safe


fluffy:
We have seen many problems with security logs getting strange and non printable character sent to them that need to be removed. We should limit reason phrases and MOQT_IMPLEMENTATION parameter to things that are safe to log. 

Given the use of this is not for end users, I would suggest limiting to code points hexadecimal 20 through 7E.

afrind:
Won't Track Names and Namespaces also end up in logs (see #1099, though also #1174)?  If so, every implementation that logs strings will need a `sanitizeForLogging(string)` method, which we can also call on `MOQT_IMPLEMENTATION` and 🔥 `Reason Phrase` 🔥

fluffy:
A best practice for security is to have a common way things are sanitizeForLogging - specifically I think we should recommend a way to format track names and namspaces both in logs and in the way we write examples in specifications. The reason this is a brest practice is so that security tools that look at the logs have a common format to be looking at. 

I would suggest we use the same thing we use in the moq-files draft as it both deals with binary and keeps "safe" things easy to read

afrind:
@fluffy are you asking to restrict the character set (as in this issue) or are you asking for an informative mechanism for logging and specs (#1099) or both?

fluffy:
Both are needed and how we do namespace / tracknames is a subset of the problem

fluffy:
On call today, 

1) it was clear we should at least say something like "remember to sanitize reason code before logging"

2) It was less clear if we should i18s reason code or not. AD is going to poke around and see if he has advice for us. Magnus is going to contact i18s directorate to see if they have advice for us.

afrind:
Parked pending #1099 PR and advice from AD / and i18n.

afrind:
I see this comment from 8 months back:

> It was less clear if we should i18s reason code or not. AD is going to poke around and see if he has advice for us. Magnus is going to contact i18s directorate to see if they have advice for us.

@gloinul 

Honestly if we just take these two fields and restrict to safe US ASCII chars all my use cases are met.

gloinul:
I clearly forgot about contacting the I18s directorate. From a lot of perspective restricting reason phrases etc to US ASCII make things simpler and these are clearly not intended to be read by regular users, only implementors and those deploying the protocol. 

Do I need to contact i18s directorate?

afrind:
[...]

PR #1355, opened 2025-11-04
Start text on formatting names for logs

fluffy:
@afrind - have a look at this and rewrite if you want. I think some of it could just be deleted as it more motivational than needed in a spec but added it with the view we can delete it while reviewing the PR. 

Thanks

Fixes: #1099

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -277,29 +277,37 @@ pairs in which the key is a variable length integer and the value
 is either a variable length integer or a byte field of arbitrary
 length.
 
+Key-Value-Pairs encode a Type value as a delta from the previous Type value,
+or from 0 if there is no previous Type value. This is efficient on the wire
+and makes it easy to ensure there is only one instance of a type when needed.
+The previous Type value plus the Delta Type MUST NOT be greater than 2^64 - 1.
+If a Delta Type is received that would be too large, the Session MUST be closed
+with a `PROTOCOL_VIOLATION`.
+
 Key-Value-Pair is used in both the data plane and control plane, but
 is optimized for use in the data plane.
 
 ~~~
 Key-Value-Pair {
-  Type (i),
+  Delta Type (i),
   [Length (i),]
   Value (..)
 }
 ~~~
 {: #moq-key-value-pair format title="MOQT Key-Value-Pair"}
 
-* Type: an unsigned integer, encoded as a varint, identifying the
-  type of the value and also the subsequent serialization.
+* Delta Type: an unsigned integer, encoded as a varint, identifying the Type
+  as a delta encoded value from the previous Type, if any. The Type identifies
+  the type of value and also the subsequent serialization.
 * Length: Only present when Type is odd. Specifies the length of the Value field
   in bytes. The maximum length of a value is 2^16-1 bytes.  If an endpoint
   receives a length larger than the maximum, it MUST close the session with a
-  Protocol Violation.
+  `PROTOCOL_VIOLATION`.
 * Value: A single varint encoded value when Type is even, otherwise a
   sequence of Length bytes.
 
 If a receiver understands a Type, and the following Value or Length/Value does
-not match the serialization defined by that Type, the receiver MUST terminate
+not match the serialization defined by that Type, the receiver MUST close
 the session with error code `KEY_VALUE_FORMATTING_ERROR`.
 
 ### Reason Phrase Structure {#reason-phrase}
@@ -395,6 +403,23 @@ payloads.
 
 Objects within a Group are in ascending order by Object ID.
 
+From the perspective of a subscriber or a cache, an Object can be in three
+possible states:
+
+1. The Object is known to not exist. This state is permanent. MOQT has multiple
+   ways to communicate that a certain range of objects does not exist,
+   including the Object Status field, and the use of gaps in FETCH responses.
+2. The Object is known to exist. From this state, it can transition to not
+   existing, but not vice versa.
+3. The state of the Object is unknown, either because it has not been yet
+   received, or it has not been produced yet.
+
+Whenever the publisher communicates that certain objects do not exist, this
+fact is expressed as a contiguous range of non-existent objects and
+by include extension headers indicating the group/object gaps; MOQT
+implementers should take that into account when selecting appropriate data
+structures.
+
 ## Subgroups {#model-subgroup}
 
 A subgroup is a sequence of one or more objects from the same group
@@ -404,8 +429,8 @@ stream and are sent on a single stream whenever possible. A Group is delivered
 using at least as many streams as there are Subgroups,
 typically with a one-to-one mapping between Subgroups and streams.
 
-When a Track's forwarding preference (see {{object-properties}}) is
-"Datagram", Objects are not sent in Subgroups and the
+When an Object's forwarding preference (see {{object-properties}}) is
+"Datagram", it is not sent in Subgroups and the
 description in the remainder of this section does not apply.
 
 Streams offer in-order reliable delivery and the ability to cancel sending and
@@ -565,7 +590,7 @@ include:
 9. The same Object is received more than once with different Payload or
     other immutable properties.
 10. An Object is received with a different Forwarding Preference than previously
-    observed from the same Track.
+    observed.
 
 The above list of conditions is not considered exhaustive.
 
@@ -609,6 +634,29 @@ ways, for example:
 3. Use TRACK_STATUS or similar mechanism to query the previous state to
    determine the largest published Group ID.
 
+## Extension Headers {#extension-headers}
+
+Tracks and Objects can have additional relay-visible fields, known as Extension
+Headers, which do not require negotiation, and can be used to alter
+MoQT Object distribution.
+
+Extension Headers are defined in {{moqt-extension-headers}} as well as external
+specifications and are registered in an IANA table {{iana}}. These
+specifications define the type and value of the header, along with any rules
+concerning processing, modification, caching and forwarding.
+
+If unsupported by the relay, Extension Headers MUST NOT be modified, MUST be
+cached as part of the Track or Object and MUST be forwarded by relays.
+
+If supported by the relay and subject to the processing rules specified in the
+definition of the extension, Extension Headers MAY be modified, added, removed,
+and/or cached by relays.
+
+Extension Headers are serialized as Key-Value-Pairs (see {{moq-key-value-pair}}).
+
+Header types are registered in the IANA table 'MOQ Extension Headers'.
+See {{iana}}.
+
 # Sessions {#session}
 
 ## Session establishment {#session-establishment}
@@ -630,7 +678,7 @@ such as TCP, and applications using MoQ might need to fallback to
 another protocol when QUIC or WebTransport aren't available.
 
 MOQT uses ALPN in QUIC and "WT-Available-Protocols" in WebTransport
-({{WebTransport, Section 3.4}}) to perform version negotiation.
+({{WebTransport, Section 3.3}}) to perform version negotiation.
 
 \[\[RFC editor: please remove the remainder of this section before publication.]]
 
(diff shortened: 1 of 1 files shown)


fluffy:
Fixes #1234

draft-ietf-moq-transport.md
```diff
@@ -324,6 +324,32 @@ Reason Phrase {
   such as language tags, that would aid comprehension by any entity other than
   the one that created the text.
 
+## Representing Namespace and Track Names
+
+There is often the need to render namespace tuples and track names in
+logs or other places. The namespace and track name are binary so they
+need to be converted to a safe form for logging.  Having a common way to
+do this is particularly useful for tools that monitory security
```
```suggestion
do this is particularly useful for tools that monitor security
```

draft-ietf-moq-transport.md
```diff
@@ -324,6 +324,32 @@ Reason Phrase {
...
+* Each of the namespace tuples are rendered in order with a period
+  between them followed by the track name with a minus between the last
+  namespace and track name.
+
+* Bytes in the range a-z, A-Z, and 0-9 are are output as is while bytes
+  all other bytes are percent encoded as a percent symbol followed by
+  exactly two lower case hex digits.
+
+For example, if a namespace had two two tuples, the first with the bytes
+0x48,0x69 , the next with bytes 0xFF, and a track name with the byte
```
```suggestion
0x48,0x69, the next with bytes 0xFF, and a track name with the byte
```

draft-ietf-moq-transport.md
```diff
@@ -324,6 +324,32 @@ Reason Phrase {
...
+
+There is often the need to render namespace tuples and track names in
+logs or other places. The namespace and track name are binary so they
+need to be converted to a safe form for logging.  Having a common way to
+do this is particularly useful for tools that monitory security
+incidents across relays from different vendors.  The following format is
+RECOMMONDED:
+
+* Each of the namespace tuples are rendered in order with a period
+  between them followed by the track name with a minus between the last
```
```suggestion
  between them followed by the track name with a hyphen between the last
```

draft-ietf-moq-transport.md
```diff
@@ -324,6 +324,32 @@ Reason Phrase {
...
+  namespace and track name.
+
+* Bytes in the range a-z, A-Z, and 0-9 are are output as is while bytes
```
afrind: This seems overly restrictive, as there are plenty of printable ASCII chars.  I noticed @suhasHere uses full track names like this (`/` is inserted as a tuple separator in my log):

000001/app=01/conf=000003/media=C1[h264,width=1920,height=1080,fps=30
,br=2000]/endpoint=0016/

Is there a reason to forbid symbols like `,`, `=`, `[`, and `]`.  The less escaping the easier reading logs will be on the eyes.

I do think it's imperative we escape the tuple and namespace/name separators that appear in the names.

afrind (commented):
This seems directionally right.  We can probably bikeshed on `.` and `-` but I could live with them I think.

draft-ietf-moq-transport.md
suhasHere: we can change the way the draft says us to :-) Might be a good reason for this PR

ianswett: AI says:
"The ASCII characters considered safe to print to logs are the printable graphic characters, which are decimal codes 32 through 126. These include: 
Space (decimal 32)
Alphanumeric characters (a-z, A-Z, 0-9)
Punctuation and symbols (!, ", #, $, %, &, ', (, ), *, +, ,, -, ., /, :, ;, <, =, >, ?, @, [, \, ], ^, _, `, {, |, }, ~) "

afrind: Let's escape whitespace.

afrind:
I was thinking maybe we should try to align with a scheme that already exists, so we don't need to write another encoder.  The proposed scheme is close to URL Safe encoding, which allows only 0-9, A-Z, a-z and these four characters:

- (hyphen) (ASCII 45)
. (dot) (ASCII 46)
_ (underscore) (ASCII 95)
~ (tilde) (ASCII 126)

But to use that, we can't select `.` and `-` as tuple and name separators.  So if we chose, say `/` and `+`, you could do:

`result = '/'.join(map(urllib.parse.quote, namespace)) + '+' + urllib.parse.quote(name)`

Examples:

```
namespace = ['simple', 'path', 'no-specials]
name = 'track'
```

`simple/path/no-specials+track`

```
namespace = ['path', 'with spaces', 'and special chars!']
name = 'hello world'
```

`path/with%20spaces/and%20special%20chars%21+hello%20world`

draft-ietf-moq-transport.md
fluffy: Keep in mind that *if* we want to align with files draft which I think is good idea, then we need to restrict to the file safe set. This use . which is not always file safe but it should not have a name that ends up starting with .

fluffy:
I think the starting point is do we want URL safe or or file name safe. 

If we are going to align this with file names, I think most people do now want / in them. If we want them to look like URL then / might be a good choice. 

The more I thought about this, the more, the more I think it is good to be able to tell if the last one is a track name or just part of the name space with no track name. So I think we should have different separators between tuples in the namespace from what separates the track name.

draft-ietf-moq-transport.md
afrind: I probably need to go read the files draft, and think about how it uses full-track names encoded as strings.  Note that if your first namespace tuple is empty (0 length, which we don't prohibit, I think?), you will get one starting with a dot.

afrind (commented):
The requirements of what this encoding is for was not clear from the issue or this PR, both of which are talking about what to put in _logs_, not filenames or URLs.  I  suggested URL encoding not because I want to put this in a URL, but because it a) it has a very restricted unencoded charset that is close what the PR proposed b) uses %xx escaping, as proposed here, and c) the encoders already exist.

If the requirement is really to come up with a way for putting track names in file names or URLs, that will generate different tradeoffs (see this MSF/WARP issue https://github.com/moq-wg/warp-streaming-format/issues/60).  For logs, I just need an encoding that will give me back the original full track name bytes unambiguously.

> The more I thought about this, the more, the more I think it is good to be able to tell if the last one is a track name or just part of the name space with no track name. So I think we should have different separators between tuples in the namespace from what separates the track name.

100% agree there should be two different separators.

fluffy:
I tried to explain it it the text of the PR that contains 

"The goal of this format is to have a format that is both filename and URL safe." 

This type of stuff is much harder to do with issues than PR having a discussion on a call or meeting where we can walk through why the design is like this. I'm not married to filename safe or anything but I definitely do not find the argument writing an encode for this is too much work very motivating.

afrind:
> "The goal of this format is to have a format that is both filename and URL safe."

Sorry, I missed that.

> I'm not married to filename safe or anything but I definitely do not find the argument writing an encode for this is too much work very motivating.

Me neither.  I can live with what's in this PR.  I did want to ask the question about reusing something that's already out there, since I think we're already going to get some sideways looks about defining our own varints (which I think are good).

afrind:
Discussed in 11/17 interim:

The wg was ok with expanding the scope of this PR to use this format anywhere a track name needs to be escaped.  We should probably make that recommendation more prominent.  There wasn't any specific feedback on character sets or re-use though.

draft-ietf-moq-transport.md
```diff
@@ -324,6 +324,32 @@ Reason Phrase {
...
+
+* Each of the namespace tuples are rendered in order with a period
+  between them followed by the track name with a hyphen between the last
+  namespace and track name.
+
+* Bytes in the range a-z, A-Z, and 0-9 are are output as is while bytes
+  all other bytes are percent encoded as a percent symbol followed by
+  exactly two lower case hex digits.
+
+For example, if a namespace had two two tuples, the first with the bytes
```
```suggestion
For example, if a namespace had two tuples, the first with the bytes
```

draft-ietf-moq-transport.md
```diff
@@ -324,6 +324,32 @@ Reason Phrase {
...
+0x48,0x69, the next with bytes 0xFF, and a track name with the byte
+0x30 would render as the string "Hi.%ff-0". The goal of this format is
+to have a format that is both filename and URL safe. It allows many
+common names to be rendered it an easily human readable form while still
+supporting all binary values.
```
```suggestion
to have a format that is both filename and URL safe. It allows many
common names to be rendered in an easily human readable form while still
supporting all binary values.
```

draft-ietf-moq-transport.md
```diff
@@ -324,6 +324,32 @@ Reason Phrase {
   such as language tags, that would aid comprehension by any entity other than
   the one that created the text.
 
+## Representing Namespace and Track Names
+
+There is often the need to render namespace tuples and track names in
+logs or other places. The namespace and track name are binary so they
+need to be converted to a safe form for logging.  Having a common way to
+do this is particularly useful for tools that monitor security
+incidents across relays from different vendors.  The following format is
```
```suggestion
There is often the need to render namespace tuples and track names 
to be consumed for logging, representing track filename or in
certain authorization verification schemes, for example. The namespace 
and track name are binary so they need to be converted to a safe
form. The following format is
```

suhasHere (approved):

vasilvv:
My feedback:
- Filename-safe and URL-safe is a good goal.
- ASCII Alphanumerics definitely should be unescaped.
- Underscore probably should be unescaped.
- If you're aiming for URL-safe, you probably don't want to use percent?
- Characters that are not URL-escaped are `! * ( ) _ - .`
- Of those, `*` is not allowed in filenames, and `!` is a bit questionable.
- This leaves us with `! ( ) _ - .`

This is a bit annoying.  Some potential approaches:
- `ns1-ns2-ns_three(track.20name)` -- I think there is already precedent for using dots for escaping?
- `ns1!ns2!ns_three-track.20name` -- similar to what's currently in the text, for everyone nostalgic for the good old UUCP days.

afrind:
We talked about this:

ns1!ns!ns_three..track.20name

ns1!ns!.!ns_three!!track.20name

wilaw:
For the record, I find

```
example.2enet-teamA-projectX--report

mydomain.2eio-users-12--profilePicture

company.2eorg-departments-marketing--budget2024
```


much easier to read than

```
example.2enet!teamA!projectX!!report

mydomain.2eio!users!123!!profilePicture

company.2eorg!departments!marketing!!budget2024
```
The bangs are not naturally occurring in standard text and don't fall comfortably on the eye.

afrind:
@wilaw whichever char we pick can't be printed unescaped though.  Are you ok losing `-`?

fluffy:
@wilaw In discussion about this, @vasilvv wanted to keep both - and _ as usable in the text of the tuples without escaping. I don't feel strongly.

wilaw:
> @wilaw whichever char we pick can't be printed unescaped though. Are you ok losing -?

I am. Its use as a divider can be substituted with _

> @wilaw In discussion about this, @vasilvv wanted to keep both - and _ as usable in the text of the tuples without escaping. I don't feel strongly.

The reason I care about readability is that I'd like to align the MSF [URL syntax](https://github.com/moq-wg/msf/issues/60) with this log file format. Consistency between the two makes sense. @vasilvv  - if it makes the MSF URLs easier to read for all the people preparing, using and debugging them, can we use - for the tuple boundary and then still be able to use _ as a separator in the naming?

fluffy:
I have slight preference for the - over ! as then we can use the same thing in https://datatracker.ietf.org/doc/draft-jennings-moq-file - I get ! can be used in some file systems but it interactions with shells makes it not great.

afrind:
@wilaw @fluffy I updated to use - and -- instead of ! and !!.  Let me know what you think.

draft-ietf-moq-transport.md
```diff
@@ -324,6 +324,43 @@ Reason Phrase {
...
+There is often the need to render namespace tuples and track names to be
+consumed for logging, representing track filename or in certain authorization
+verification schemes, for example. The namespace and track name are binary so
+they need to be converted to a safe form. The following format is RECOMMONDED:
+
+* Each of the namespace tuples are rendered in order with a hyphen (-) between
+  them followed by the track name with a double hyphen (--) between the last
+  namespace and track name.
+
+* Bytes in the range a-z, A-Z, 0-9 as well as ! (0x21) and _ (0x5f) are are
```
```suggestion
* Bytes in the range a-z, A-Z, 0-9 as well as ! (0x21) and _ (0x5f) are
```

vasilvv:
I don't feel strongly about losing either `_` or `-`, I just think we need to keep at least one of those.

For what it's worth, `!` does not look unnatural to me, but I don't care strongly either way.

fluffy:
OK, I cleaned this up a bit and reduced to just one example as the other examples did not add much more insight. 

I think this is at the point that everyone who comments on it "can live with it" in this form.

PR #1357, opened 2025-11-05
Editorial: Unify sesion termination

michalhosna:
Unification makes it much easier to search through the document. Especially when backreferencing or validating an implementation. 

As advised publicly by Alan, I am not filing an editorial issue; instead, I am directly filling a PR.

Starting this as a single PR. Happy to split into multiple PRs or drop some changes if deemed better. I didn't want to unnecessarily pollute PRs. 

### Close session vs. terminate session:

I know the document has a `## Termination  {#session-termination}` section. But even in it, "terminate" and "close" are both used. I do think "close" is better, because there are other uses of the word "terminate". 

https://github.com/moq-wg/moq-transport/blob/586da24d41327cafc22dd4d886b5e2602d166116/draft-ietf-moq-transport.md?plain=1#L1288-L1289

Looking back at this, this may also need clarification.

I tried to do a minimum number of changes for unification, so I chose to use "close" for session and keep "termination" for streams and subscriptions.

Sorry if this is a wrong call; English is (obviously) not my 1st language.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -299,7 +299,7 @@ Key-Value-Pair {
   sequence of Length bytes.
 
 If a receiver understands a Type, and the following Value or Length/Value does
-not match the serialization defined by that Type, the receiver MUST terminate
+not match the serialization defined by that Type, the receiver MUST close
 the session with error code `KEY_VALUE_FORMATTING_ERROR`.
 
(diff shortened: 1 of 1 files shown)


yekuiwang:
Being consistent is good. The changes look good to me.

ianswett (approved):

PR #1362, opened 2025-11-06
Add Prior Group Subscription Filter type

### LINKED ISSUE #1358: Subscribing to the start of the current Group could be optimized
Assuming Joining FETCH stays as is, there is an optimization to joining at the current Group that I believe is worth considering.

Just because there is a Largest Object, doesn't mean all Objects less than that are available.  This means that a Joining FETCH could be blocked waiting for Objects the Relay has never seen.

For example, if there was a base layer and a temporal enhancement layer, and the base layer was higher priority, a Relay might have all of the base layer and none of the enhancement layer.  If a Joining FETCH came in, the Subscriber would receive the first Object in the base layer and then be stuck there until the enhancement layer started arriving.

In some cases, this is just underutilizing the available bandwidth, but in other cases if the enhancement layer doesn't arrive within the delivery timeout, the playback can't begin.

In contrast, if both the Joining FETCH and SUBSCRIBE start at the beginning of the current Group, the FETCH could skip over any Objects (PR #1331) that are not yet available and then they'd be delivered via SUBSCRIBE.

The farther back in time the FETCH is, the less likely this is an issue, but I think it could be a substantial issue for the "Join at the current Group" case that we expect to be extremely common.

ianswett:
Allows starting at the current group when the 'Prior Group' value is 0.

Allows faster Track joining in the presence of reordering.

Fixes #1358

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1000,6 +1000,13 @@ Subscription Filter {
 
 Filter Type can have one of the following values:
 
+Largest Group (0x0): The filter Start Location is `{Largest Object.Group, 0}`
+and `Largest Object` is communicated in SUBSCRIBE_OK. If no content has been
+delivered yet, the filter Start Location is {0, 0}. There is no End Group -
+the subscription is open ended. Note that Objects before or equal to
+'Largest Object' will only be delivered if they arrive at the Publisher after
+the Subscription is created.
+
 Largest Object (0x2): The filter Start Location is `{Largest Object.Group,
 Largest Object.Object + 1}` and `Largest Object` is communicated in
 SUBSCRIBE_OK. If no content has been delivered yet, the filter Start Location is
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1028,8 +1028,10 @@ Subscription Filter {
 
 Filter Type can have one of the following values:
 
-Largest Group (0x0): The filter Start Location is `{Largest Object.Group, 0}`
-and `Largest Object` is communicated in SUBSCRIBE_OK. If no content has been
+Prior Group (0x0): The filter specifies a 'GroupsPrior' value as a varint
+and starts that many groups prior to the current Group.
+Start Location is `{Largest Object.Group - GroupsPrior, 0}` and
+`Largest Object` is communicated in SUBSCRIBE_OK. If no content has been
 delivered yet, the filter Start Location is {0, 0}. There is no End Group -
 the subscription is open ended. Note that Objects before or equal to
 'Largest Object' will only be delivered if they arrive at the Publisher after


afrind (approved):

suhasHere (approved):

afrind:
Briefly covered this in the interim today but did not have time to discuss.

afrind:
Discussed in 11/17 Interim:

wg felt this was ok.

gwendalsimon:
We discussed Largest Group quite a long time ago, didn't we?

One point is kind of philosophical: FETCH is for the past, SUBSCRIBE is for the present and the future. Largest Group for SUBSCRIBE breaks this rule.

More practically, we have the case where the Relay does not have the first Objects of the current group (because it does not have a cache, or because these Objects are too old to be delivered from cache, or because it did not have any active Subscription at the time these Objects were delivered). It may send the same request upstream, but the upstream Relay or the Original Publisher may have started a new Group. To be safe, the Relay should send a FETCH instead. 

Why is the combination of SUBSCRIBE Largest Object + Joining FETCH Offset 0 not good?

ianswett:
The issue describes the problem, but if we do something along the lines of merging Subscribe and Fetch (#1377) those would also be resolved.

martinduke:
I am not convinced that "FETCH everything, but return objects out of order" is a real use case. If it is a real use case, then let's just make a FETCH parameter that does this rather than have a  SUBSCRIBE filter that affects how FETCH works.

One problem with this PR is that the notion of SUBSCRIBE that returns "new objects" < largest_location was expressly separated from FETCH, and this mingles them in a way that might cause problems.

afrind:
Discussed 12/1:

Parking until we have a larger discussion on joining fetch concerns.

ianswett:
> I am not convinced that "FETCH everything, but return objects out of order" is a real use case. If it is a real use case, then let's just make a FETCH parameter that does this rather than have a SUBSCRIBE filter that affects how FETCH works.

+1

> 
> One problem with this PR is that the notion of SUBSCRIBE that returns "new objects" < largest_location was expressly separated from FETCH, and this mingles them in a way that might cause problems.

SUBSCRIBE returns Objects less than Largest Location today, since you can specify an absolute range.  This just avoids the latency of doing a TRACK_STATUS then issuing a SUBSCRIBE starting at the Group you want.

In my mind, a key point is that there may be some/many Objects less than Largest Object that are not currently available.  They may have been 'published' by the original publisher, but you can't get them yet.  So the concept of FETCH is for the past and SUBSCRIBE is for the future is only true if Publishers are strictly publishing in Object ID order and there's no reordering, basically TCP.

ianswett:
> 
> Why is the combination of SUBSCRIBE Largest Object + Joining FETCH Offset 0 not good?

One example is if every other Object is in a different Subgroup, the Joining FETCH will hang waiting for Object 1, even if it has all of the base Subgroup.  If the other Subgroup has lower priority, it's natural it might be delivered later.

gwendalsimon:
> > Why is the combination of SUBSCRIBE Largest Object + Joining FETCH Offset 0 not good?
> 
> One example is if every other Object is in a different Subgroup, the Joining FETCH will hang waiting for Object 1, even if it has all of the base Subgroup. If the other Subgroup has lower priority, it's natural it might be delivered later.

I don't understand this use case. Seriously. Anyway FETCH can have a higher priority. It seems to me like a lot of "if"s to ask for such a major change, with all the consequences. 

As I said earlier, the Publisher / Relay may not be even able to get the first object of the group (nothing in cache / too old objects to be delivered / request sent upstream but Publisher already serving the next group...). The PR does not mention the behavior of the Relay for any of the different cases that can arrive.

draft-ietf-moq-transport.md
```diff
@@ -1028,6 +1028,15 @@ Subscription Filter {
 
 Filter Type can have one of the following values:
 
+Prior Group (0x0): The filter specifies a 'GroupsPrior' value as a varint
+and starts that many groups prior to the current Group.
+Start Location is `{Largest Object.Group - GroupsPrior, 0}` and
+`Largest Object` is communicated in SUBSCRIBE_OK. If no content has been
+delivered yet, the filter Start Location is {0, 0}. There is no End Group -
```
afrind, line 1167: No content or GroupsPrior > LargestGroup

draft-ietf-moq-transport.md
```diff
@@ -1028,6 +1028,15 @@ Subscription Filter {
 
 Filter Type can have one of the following values:
 
+Prior Group (0x0): The filter specifies a 'GroupsPrior' value as a varint
```
afrind, line 1163: It feels just a little strange the PriorGroup=0 is CurrentGroup.  I think we should support the use case but the name is strange.

afrind:
In general this seems fine and easy to implement (it just sets the subscription filter range), and I'm not sure it even qualifies as Joining FETCH Dissent.  If you combine it with Joining FETCH, Fill Timeout=0 (#1490), you will get everything, no duplicates, no HoLB, assuming all the gaps in the cache are late objects and not missing for a different reason (upstream delivery timeout).

vasilvv (approved):

martinduke:
I would prefer that we not merge this PR in favor of the REWIND proposal, which I will try this weekend.

afrind:
Agreed, we can hold off here

ianswett:
Closing in favor of #1642

PR #1366, opened 2025-11-06
Key-Value-Pairs Delta encoding editorial tweaks. 

michalhosna:
#1315 Follow up. 

Resubmitting my late-review proposal. I suspect my comment was ignored because getting a wire-format change into main without late unnecessary editorial discussion is good. That's why I am resubmitting. Editors, if I got that wrong, and you like the current wording, feel free to close this.

~I am not sure where the last-minute change of 2^62-1 to 2^64-1 came from. I think allowing bigger value than is encodable in a single delta is wrong. It doesn't really add any value and makes the handling complex, how is one supposed to encode it if needed without prior pairs? If I understand that wrongly and there is need for 2^64-1 I think there should be an explanation somewhere.~

I do think the size limit should be enforced for all Key-Value-Pairs.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -275,8 +275,14 @@ Location A < Location B if:
 Key-Value-Pair is a flexible structure designed to carry key/value
 pairs in which the key is a variable length integer and the value
 is either a variable length integer or a byte field of arbitrary
-length. Key-Value-Pairs encode a Type value as a delta from the
-previous Type value, or from 0 if there is no previous Type value.
+length.
+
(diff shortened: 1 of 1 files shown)


afrind:
We bumped it to 2^64-1 now so we wouldn't forget when we made VARINT_MAX bigger in #1016

michalhosna:
> We bumped it to 2^64-1 now so we wouldn't forget when we made VARINT_MAX bigger in https://github.com/moq-wg/moq-transport/pull/1016

Ok, that makes sense. Sorry. I didn't know that merging that is close.

afrind (approved):

PR #1371, opened 2025-11-09
Clarify MoQT Extensibility

### LINKED ISSUE #790: Clarify what is the extension model for filter types
Section 8.6 Subscribe in -10 says the following: 

A filter type other than the above MUST be treated as error.

I think it would be good to clarify what the extension model for new filter types are. Is filter types something that should have an IANA registry? Is the method to disable the above MUST for extension to use a negotiated parameter or does it need to use a different version of the protocol?

### LINKED ISSUE #795: Extension model for the Fetch Types
So currently Fetch (Section 8.12 of -10) states:

A Fetch Type other than 0x1 or 0x2 MUST be treated as an error.

So what is the extension model for the Fetch Type? Does it require a new MoQ version, or should one use parameters to negotiate extensions. I think similar to the Filter types in Subscribe (see #790).

I think this may point to at least a declaration of intention when it comes to fields that do not allow additional values in this document.

ianswett:
Fixes #790
Fixes #795

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -665,7 +665,7 @@ specific to the underlying transport protocol usage (see {{session}}).
 
 Endpoints use the exchange of Setup messages to negotiate MOQT extensions.
 Extensions can define new Message types, new Parameters, or new framing for
-Fetch or Subscription response streams.
+Data Streams and Datagrams.
 
 The client and server MUST include all Setup Parameters {{setup-params}}
 required for the negotiated MOQT version in CLIENT_SETUP and SERVER_SETUP.


draft-ietf-moq-transport.md
```diff
@@ -663,28 +663,21 @@ specific to the underlying transport protocol usage (see {{session}}).
 
 ## Extension Negotiation {#extension-negotiation}
 
-Endpoints use the exchange of Setup messages to negotiate any MOQT extensions
-to use.
+Endpoints use the exchange of Setup messages to negotiate MOQT extensions.
+Extensions can define new Message types, new Parameters, or new framing for
+Fetch or Subscription response streams.
```
```suggestion
Data Streams and Datagrams.
```

PR #1372, opened 2025-11-09
Forward State Clarification

ianswett:
I was working on a different change, but got sidetracked.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -663,27 +663,20 @@ specific to the underlying transport protocol usage (see {{session}}).
 
 ## Extension Negotiation {#extension-negotiation}
 
-Endpoints use the exchange of Setup messages to negotiate any MOQT extensions
-to use.
+Endpoints use the exchange of Setup messages to negotiate MOQT extensions.
+Extensions can define new Message types, new Parameters, or new framing for
+Data Streams and Datagrams.
 
-The client includes all Setup Parameters {{setup-params}} required for the
-negotiated MOQT version in CLIENT_SETUP.
+The client and server MUST include all Setup Parameters {{setup-params}}
+required for the negotiated MOQT version in CLIENT_SETUP and SERVER_SETUP.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -928,14 +928,12 @@ SUBSCRIBE_OK when establishing a subscription. This value can be used in a
...
-subscription sets the initial Forward State in either PUBLISH or SUBSCRIBE.  The
-sender of PUBLISH_OK can update the Forward State based on its preference.  Once
-the subscription is established, the subscriber can update the Forward State by
-sending SUBSCRIBE_UPDATE.  Control messages, such as PUBLISH_DONE
-({{message-publish-done}}) are still sent on subscriptions in Forward State 0.
+All `Established` subscriptions have a Forward State which is either 0 or 1.
+The publisher does not send Objects if the Forward State is 0, and does send them
+if the Forward State is 1.  The initiator of the subscription sets the initial
+Forward State in either PUBLISH or SUBSCRIBE.  The subscriber can send PUBLISH_OK
+or SUBSCRIBE_UPDATE to update the Forward State. Control messages, such as
```
afrind, line 928: I sort of liked the original text better here -- in a publish flow the pub_ok has to come before subscribe_update, though I am realizing that in a subscribe flow, maybe the subscriber can pipeline SUBSCRIBE_UDPATE?  But we should tackle that separately since that impacts the state machine diagram, and possibly language about the publisher being tolerant of receiving SUB_UPDATE on a subscription terminated with subscribe_error.

ianswett: I thought the point of some changes we made recently with filters is to allow a subscriber to pipeline SUBSCRIBE_UPDATE? #872

afrind: yes, SUB + SUB_UPDATE should be allowed, but PUB_OK needs to come before SUB_UPDATE the other way round.  Let's tackle separately.

afrind (approved):

PR #1373, opened 2025-11-09
Add Extensions for Tracks to PUBLISH and SUBSCRIBE_OK

### LINKED ISSUE #781: MoQT lacks support for track level object extension

ianswett:
I don't love the name 'Extensions', so I'm open to changing it in this PR or in a separate PR.

Fixes #781

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2334,7 +2334,8 @@ SUBSCRIBE_OK Message {
   Request ID (i),
   Track Alias (i),
   Number of Parameters (i),
-  Parameters (..) ...
+  Parameters (..) ...,
+  Extensions (..),
 }
 ~~~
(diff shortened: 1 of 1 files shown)


wilaw:
> I don't love the name 'Extensions', 

Can we just called them "Headers"? That seems reasonably intuitive.

Issue #1316, opened 2025-10-17
Possible improvements on MoQ support of VOD applications


yekuiwang:
There was a great goal/wish of MoQ being able to unify the worlds of conversational, live streaming, and VOD applications (https://www.ietf.org/blog/moq-overview/). As side topics, @afrind and I touched a couple of times, e.g., in #1205 and #1309, on support of VOD applications in MOQT. From the discussions, I learned that:

1. For support of VOD applications based on MOQT, only FETCH requests can be used, as for VOD all Objects are "already published" or "from the past", while SUBCRIBE is only for "newly published or received" Objects.
2. Because of bullet 1, multiple streams cannot be used for VOD, unless Subgroup or Priority based Object filter, e.g., as discussed in #1164, is added to the MOQT draft and allowed to be used with FETCH.
3. Because Subgroup based streams cannot be used, signalling of priority cannot be in the stream header, while all Objects of one Subgroup are required to have the same priority. (@afrind does not think this is a major issue, which is probably true, but it would be very nice if this aspect can be improved.)
4. @afrind mentioned, in both #1205 and #1309, that due to lack of information about the consumption/playback rate, a relay cannot effectively serve past content, but rather VOD support with MOQT has to rely on use of many FETCH requests, just like use of many HTTP requests in HTTP adaptive streaming (e.g., DASH and HLS). 

Bullet 2 depends on #1164, and bullet 3 somehow depends on bullet 2. 

On bullet 4 above, I can think of the following possible ways of conveying information about the consumption/playback rate (i.e., media bitrate) such that relays can use it:
a) Require that catalog tracks are not encrypted, which ensures that relays can access all information in the catalog tracks, including the Bitrate field.
b) Add an Object Extension Header to convey the bitrate information.
c) Add a Bitrate Parameter to conveying the bitrate information, and allow this parameter to appear in a TRACK_STATUS_OK and whatever other control messages that deemed useful to have this parameter.

Opinions?

afrind:
CC: VOD enthusiast @wilaw

wilaw:
+1 to ensuring that MOQT is robust over the VOD playback use-case. 

>  VOD support with MOQT has to rely on use of many FETCH requests

[...]

Issue #1205, opened 2025-09-05
Potential wording issues of "newly published objects" and "already published objects"


yekuiwang:
In Sect. 9.7 (SUBSCRIBE): 
> A subscription causes the publisher to send _newly published objects_ for a track.

In Sect.  9.16 (FETCH): 
> A subscriber issues a FETCH to a publisher to request a range of _already published objects_ within a track.

Clarification questions:

1. Aren't "newly published objects" a kind of "already published objects"? They were already published even though they were newly published, right?

2. An original publisher has Objects of a track of a VOD content stored for a very very long time, thus it does not have any "newly published objects" for the track. For such an original publisher, can a subscriber send a SUBSCRIBE to it to request for Objects of the track?

afrind:
I agree this could use some clarification.

An object is "newly published" in this context when a Publisher sends the Object in a subscription, or is received via an established subscription.  Only such objects can be received via SUBSCRIBE.

So can you SUBSCRIBE to a VOD?

There's no way for a _generic_ caching relay to play the VOD back to you at anything other than line rate.  That said, you could write a VOD server that understands the media rate and acts like an original publisher.  I believe Zafer Gurel has implemented something like this in the demos he's shown at recent interims.

yekuiwang:
Regarding whether SUBSCRIBE can be used for a VOD content, let me try an example, for a better understanding.

Below is my understanding how streaming of VOD content could work:

C: client, R: relay, OP: Original Publisher (i.e., server)

1. C->R: SUBSCRIBE trackA
2. R->OP: SUBCRIBE trackA
3. OP->R: Objects of trackA
4. R->C: Objects of trackA

However, if SUBSCRIBE can only be used for "newly published objects", then is step 2 above OK? Because in step 2 the relay uses SUBSCRIBE to request Objects that do not seem to be "newly published objects".

afrind:
I think VOD in your example can be somewhat fraught, because MOQ relays are designed to work with live content.  Think about how the relay would process a second SUBSCRIBE AbsoluteStart=0,0 -- it will coalesce it with the existing subscription, causing the second subscriber to join the first wherever it is watching (which is a different kind of application -- watch party!).  You would need some kind of specialized VOD server rather than a _Relay_ to make this work.

yekuiwang:
Hmm, I am learning a lot! Thanks! 

[...]

PR #1377, opened 2025-11-23
RFC: Fetch is dead

### LINKED ISSUE #861: Are Groups really join points anymore / SUBSCRIBE start on group boundaries
In 2.3 and elsewhere, it mentions that a Group is a join point.

It does make intuitive sense for a Group to be a join point, as starting at Object 0 gives you the key frame.

However, the protocol doesn't really elevate the group as an important join point anymore. LatestObject (and its potential descendants) don't care about Groups. AbsoluteStart, AbsoluteRange, and regular FETCH don't privilege Object 0 at all.

We're just left with Joining FETCH that uses groups as a join point.

What I'd propose is that AbsoluteStart and AbsoluteRange simply have a start group rather than a start group and object. Is there a use case for a SUBSCRIBE to a future group starting mid-group? It sounds nonsensical.

Alternately, we could change the text to make less sweeping claims about join points.

### LINKED ISSUE #1039: Simplifying joining at the latest available join point.

FETCH relative joining 0

### LINKED ISSUE #1168: Ephemeral vs Long-lived Request IDs

### LINKED ISSUE #1313: Joining FETCH as a separate control message creates edge cases and feature gaps

### LINKED ISSUE #1343: Cancelling a subscription can be slow
In HTTP/3, we made sure that one could cancel a HTTP request at the QUIC layer, in case the upper HTTP layers were busy.  This was informed by experience with the importance of TCP reset working.

In MoQ today, UNSUBSCRIBE is on the control stream, so it's HoL blocked behind other messages.  Today this is made worse by the fact that there might be a number of PUBLISH_NAMESPACE messages, but even without that, it'd be nice to be able to immediately cancel a Subscription without waiting for any pending SUBSCRIBE_UPDATEs/etc to be processed.

### LINKED ISSUE #1391: Unclear how to use Joining FETCH with New Group Request
Per discussion on #1335 , there's an issue where if you use NGR, there are two possibilities:
 1) NGR will not produce a new group, so you'd like to Joining FETCH for the first part of the Group.
 2) NGR will produce a new group, so there's no need for a Joining FETCH

In the second case, if you issue a Joining FETCH, you get a gap between the current Largest Object and the start of the next Group, which is seemingly not the intent of Joining FETCH, but you also aren't going to use the Joining FETCH anyway so maybe it doesn't matter?

@suhasHere indicated he didn't use Joining FETCH with NGR, so maybe not a practical problem.

ianswett:
Removes Fetch and Joining Fetch in favor of moving to a Subscribe model that uses a single bidirectional control stream per Subscription.  Does not change how Objects are sent.

Removes Request ID from SUBSCRIBE_UPDATE, but not entirely.

Adds the 'Relative Past Group (0x5)' filter type to replicate the relative joining fetch functionality.

Some open PRs need to land for this to fully come together, so this is more of a proof-of-concept.

This is currently parked in favor of the more incremental change to move Requests to bidi streams, #1389

Fixes #1391
Fixes #1343
Fixes #1313 
Fixes #1168
Fixes #1039

May Fix #861

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -457,8 +457,8 @@ by doing a Fetch upstream, if necessary.
 
 Applications that cannot produce Group IDs that increase with time are limited
 to the subset of MOQT that does not compare group IDs. Subscribers to these
-Tracks SHOULD NOT use range filters which span multiple Groups in FETCH or
-SUBSCRIBE.  SUBSCRIBE and FETCH delivery use Group Order, so they could have
+Tracks SHOULD NOT use range filters which span multiple Groups in
+SUBSCRIBE.  SUBSCRIBE uses Group Order, which could have
 an unexpected delivery order if Group IDs do not increase with time.
 
 The amount of time elapsed between publishing an Object in Group ID N and in a
@@ -565,8 +565,8 @@ include:
 
 The above list of conditions is not considered exhaustive.
 
-When a subscriber detects a Malformed Track, it MUST UNSUBSCRIBE any
-subscription and FETCH_CANCEL any fetch for that Track from that publisher, and
+When a subscriber detects a Malformed Track, it MUST unsubscribe from any
+subscription for that Track from that publisher, and
 SHOULD deliver an error to the application.  If a relay detects a Malformed
 Track, it MUST immediately terminate downstream subscriptions with PUBLISH_DONE
 and reset any fetch streams with Status Code `MALFORMED_TRACK`. Object(s)
@@ -818,8 +818,7 @@ The server sends a GOAWAY message, signaling the client to establish a new
 session and migrate any `Established` subscriptions. The GOAWAY message optionally
 contains a new URI for the new session, otherwise the current URI is
 reused. The server SHOULD close the session with `GOAWAY_TIMEOUT` after a
-sufficient timeout if there are still open subscriptions or fetches on a
-connection.
+sufficient timeout if there are still open subscriptions on a connection.
 
 When the server is a subscriber, it SHOULD send a GOAWAY message to downstream
 subscribers prior to any UNSUBSCRIBE messages to upstream publishers.
@@ -935,9 +934,7 @@ response to a PUBLISH. The peer SHOULD close the session with a protocol error
 if it receives more than one.
 
 A publisher MUST save the Largest Location communicated in PUBLISH or
-SUBSCRIBE_OK when establishing a subscription. This value can be used in a
-Joining FETCH (see {{joining-fetches}}) at any time while the subscription is
-active.
+SUBSCRIBE_OK when establishing a subscription.
 
 All `Established` subscriptions have a Forward State which is either 0 or 1.
 The publisher does not send Objects if the Forward State is 0, and does send them
@@ -1036,6 +1033,11 @@ to network reordering or prioritization, relays can receive Objects with
 Locations smaller than  `Largest Object` after the SUBSCRIBE is processed, but
 these Objects do not pass the Largest Object filter.
 
+Relative Past Group (0x5): The number of Groups in the past is specified explicitly.
+The filter Start Location is `{Largest Object.Group - NumGroups, 0}`.  There is no
+End Group - the subscription is open ended. A NumGroups value of 0 starts the
+subscription at at the current Group.
+
 Next Group Start (0x1): The filter Start Location is `{Largest Object.Group + 1,
 0}` and `Largest Object` is communicated in SUBSCRIBE_OK. If no content has been
 delivered yet, the filter Start Location is {0, 0}.  There is no End Group -
@@ -1067,8 +1069,7 @@ request an existing Group or wait for a future Group.  Different applications
 will have different approaches for when to begin a new Group.
 
 To join a Track at a past Group, the subscriber sends a SUBSCRIBE with Filter
-Type `Largest Object` followed by a Joining FETCH (see {{joining-fetches}}) for
-the intended start Group, which can be relative.  To join a Track at the next
+Type `Relative Past Group`. To join a Track at the next
 Group, the subscriber sends a SUBSCRIBE with Filter Type `Next Group Start`.
 
 #### Dynamically Starting New Groups
@@ -1092,27 +1093,6 @@ Another possible subscriber pattern is to send a SUBSCRIBE with Filter Type
 `Next Group Start` and NEW_GROUP_REQUEST equal to 0.  The value of
 DYNAMIC_GROUPS in SUBSCRIBE_OK will indicate if the publisher supports dynamic
 groups. A publisher that does will begin the next group as soon as practical.
-
-## Fetch State Management
-
-The publisher MUST send exactly one FETCH_OK or REQUEST_ERROR in response to a
-FETCH.
-
-A subscriber keeps FETCH state until it sends FETCH_CANCEL, receives
-REQUEST_ERROR, or receives a FIN or RESET_STREAM for the FETCH data stream. If the
-data stream is already open, it MAY send STOP_SENDING for the data stream along
-with FETCH_CANCEL, but MUST send FETCH_CANCEL.
-
-The Publisher can destroy fetch state as soon as it has received a
-FETCH_CANCEL. It MUST reset any open streams associated with the FETCH. It can
-also destroy state after closing the FETCH data stream.
-
-It can destroy all FETCH state after closing the data stream with a FIN.
-
-A REQUEST_ERROR indicates that both endpoints can immediately destroy state.
-Since a relay can start delivering FETCH Objects from cache before determining
-the result of the request, some Objects could be received even if the FETCH
-results in error.
 
 
 # Namespace Discovery {#track-discovery}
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1379,17 +1359,17 @@ deduplicate Objects before forwarding, subject to implementation constraints.
...
+Relays MUST ensure subscribers are authorized to access the content associated
+with the Track. The authorization information can be part of request itself or
+part of the encompassing session. The specifics of how a relay authorizes a user
+are outside the scope of this specification.
 
 The relay MUST have an `Established` upstream subscription before sending
 SUBSCRIBE_OK in response to a downstream SUBSCRIBE.  If a relay does not have
-sufficient information to send a FETCH_OK immediately in response to a FETCH, it
-MUST withhold sending FETCH_OK until it does.
+sufficient information to send a SUBSCRIBE_OK, itMUST withhold sending
```
zafergurel, line 1371:
```suggestion
sufficient information to send a SUBSCRIBE_OK, it MUST withhold sending
```

acbegen:
Having "fetch" functionality within Subscribe or as a separate message has been discussed in the past. Following the [slides](https://datatracker.ietf.org/meeting/119/materials/slides-119-moq-subscribe-and-fetch) (slides 16-17) and then the subsequent [PR](https://github.com/moq-wg/moq-transport/pull/421).

I am curious whether we should reopen this discussion. Having something in the spec for 1.5 years and having people implement it, then changing it completely, I think, deserves some strong justification. Keeping Subscribe and Fetch separate, I thought, was also something those who did not want to implement Fetch wanted in the first place.

draft-ietf-moq-transport.md
```diff
@@ -2300,20 +2272,26 @@ PREFIX_OVERLAP (0x30):
...
 
 ## SUBSCRIBE {#message-subscribe-req}
 
-A subscription causes the publisher to send newly published objects for a track.
+A subscription causes the publisher to send Objects for a Track. Newly
+published Objects are published according to their forwarding preference
+and previously published Objects are delivered on a single unidirectional
+stream, allowing the receiver to apply flow control.
+
+There are 4 types of subscriptions:
```
gwendalsimon, line 2286: Actually three?

draft-ietf-moq-transport.md
```diff
@@ -2300,20 +2272,26 @@ PREFIX_OVERLAP (0x30):
...
 
-Subscribe only requests newly published or received Objects.  Objects from the
-past are retrieved using FETCH ({{message-fetch}}).
+Code | Subscription Type
+0x0 | Only newly published Objects
+0x1 | Only previously published Objects
+0x2 | All previously and newly published Objects
```
gwendalsimon, line 2291: The wording is a bit confusing. "All previously ..." may be interpreted as we need to send the whole history of the track. Maybe "Both previously and newly published objects"? Or "Some previously published objects and the newly published objects"?

draft-ietf-moq-transport.md
```diff
@@ -2421,50 +2433,23 @@ The format of SUBSCRIBE_UPDATE is as follows:
...
-* Subscription Request ID: The Request ID of the SUBSCRIBE
-  ({{message-subscribe-req}}) this message is updating.  This MUST match an
-  existing Request ID.  The publisher MUST close the session with `
-  `PROTOCOL_VIOLATION` if the subscriber specifies an invalid Subscription
-  Request ID.
-
 * Parameters: The parameters are defined in {{version-specific-params}}.
 
 
-## UNSUBSCRIBE {#message-unsubscribe}
```
gwendalsimon, line 2443: why is UNSUBSCRIBE dead as well? Collateral victim?

acbegen: Indeed, it still shows up in one place (likely to be a leftover) but why are we removing Unsubscribe for terminating the sessions?

gloinul:
I have to say this was not what I expected as a solution based on the earlier discussion. I thought the main solution to #1343 was to send control messages on different streams so there would be no HOL and you have a request context for updates. On #1313 I thought the obvious solution was to create a new verb that represents the combination. I think bunching all under subscribe have several impacts which I don't think necessary are in the best interest for the WG:

1. The name of the verb does not represent what the protocol does
2. It makes the separation into different implementation capabilities more unclear as it would not follow along the line of verb. A relay might only support subscribe, or subscribe+fetch, or be capable of joining fetch. Because I think it has been clear that there are usage of the protocol that has no need of caching or what was happening in the past, only what is being published after the end subscriber joined.
3. It will make it hard to discuss as we can't just subscribe, you have to make clear which mode of operations you are talking about.  
4. Although renaming all as part of Subscribe may in fact not be larger than keeping different verbs it still gives the impression that the WG is not making progress on the protocol and that we reshuffeling things completely again. 

So I have to ask is there actually a gain in deleting the verb names in favor calling all Subscribe? Are we actually getting anything additionally from that part of the change?

gwendalsimon:
Summary of the first comments: Stop trying to make FETCH _not_ happen!

yekuiwang:
If FETCH is gone, then the key differences between requesting "past" and "future" objects/groups would be gone, right? For example, currently all objects in a FETCH response need to be in one stream; with this PR those same set of objects would be part of a subscription and then each subgroup in each group would be in its own stream, right?

If so, then removing FETCH appears to be a unification of requesting "past" and "future" objects/groups, or, in other words, a unification of live streaming and VOD streaming, which is very good thing, isn't it? And then this PR would fix the following issues: #1205, #1316.

gloinul:
> If FETCH is gone, then the key differences between requesting "past" and "future" objects/groups would be gone, right? For example, currently all objects in a FETCH response need to be in one stream; with this PR those same set of objects would be part of a subscription and then each subgroup in each group would be in its own stream, right?
> 

My reading of this PR is that it does not actually remove this distinction, please see the PR. There is a whole section on sending Sending Previously Published Objects (See line 2338). 

> If so, then removing FETCH appears to be a unification of requesting "past" and "future" objects/groups, or, in other words, a unification of live streaming and VOD streaming, which is very good thing, isn't it? And then this PR would fix the following issues: #1205, #1316.

That is not how I interpret this PR. So far it is primarily a name squashing affair to me. It removes the name fetch, to instead use the verb SUBSCRIBE to implement all three of:
SUBSCRIBE
FETCH
SUBSCRIBE+FETCH (the one verb version of Joining FETCH).

Then it delivers what would have been "fetched" over its own stream. Rather than in one stream per group. 

I think the name of this PR is misleading, it is easy to interpret it as that the functionality of FETCH is dead, instead of what I believe is meant, namely that only the name that is to be killed.

kixelated:
My biggest problem with FETCH is how it introduces head-of-line blocking. A relay has to merge N input streams (object ID ascending) into 1 output stream (group+object ID ascending). I think this PR would remove that, right?

afrind:
> My biggest problem with FETCH is how it introduces head-of-line blocking. A relay has to merge N input streams (object ID ascending) into 1 output stream (group+object ID ascending).

This is only a thing if you are trying to FETCH live object, which is not something MOQT supports.  A FETCH response always ends at Largest Object, even if the requested end is larger.  So you construct the fetch response from your cache, or by sending a FETCH upstream, or some mixture of both.

kixelated:
> > My biggest problem with FETCH is how it introduces head-of-line blocking. A relay has to merge N input streams (object ID ascending) into 1 output stream (group+object ID ascending).
> 
> This is only a thing if you are trying to FETCH live object, which is not something MOQT supports. A FETCH response always ends at Largest Object, even if the requested end is larger. So you construct the fetch response from your cache, or by sending a FETCH upstream, or some mixture of both.

What I mean by interleaving N streams, is interleaving N sources of objects: 
- one stream might be a group from cold storage (ex. fetched from S3 via HTTP)
- another might be a cached group from a SUBSCRIBE (in RAM, not really a stream)
- another might be a FETCH for a potential gap within that cached group (ex. object 68 is missing, does it exist?)
- another might be an live group from a SUBSCRIBE **that is smaller than Largest Object**

The last one is because groups and objects arrive out of order on purpose. You can totally have a "live object" that it is smaller than Largest Object and should be served via FETCH, but it hasn't arrived yet due to congestion. Yes a relay *could* issue a FETCH instead of waiting for the stream to finish, but it's kind of silly to have a FETCH overlap with an active SUBSCRIBE (just wait instead of downloading it twice).

In general it's a pain in the butt being forced to serve objects in order. I think this all gets fixed if FETCH could serve sub-groups independently just like SUBSCRIBE. The end user experience would also be better because objects can be immediately served from cache (utilizing the network), instead of blocking on the fetch from cold storage. It's the same concept behind being able to serve HTTP/2/3 responses in parallel instead of HTTP/1 pipelining.

wilaw:
>  I think this all gets fixed if FETCH could serve sub-groups independently just like SUBSCRIBE.

+1 . The original design of the FETCH response as a single stream was an imagined convenience stemming from the assumption that all the objects were in cache and therefore easily sequence-able. Implementation experience shows this to be otherwise. In many cases all the objects _will_ be in cache and easily sequence-able and the relay _may_ indeed use a single stream to return them. But in the case that they are not, flexibility should be allowed. A client that is capable of receiving SUBSCRIBE response can also correctly receive the FETCH across across multiple streams.

afrind:
One of the design requirements of FETCH was "I can tell I got everything" and "I know what does not exist".  Delivering subgroups out of order in different streams breaks this property.

kixelated:
> One of the design requirements of FETCH was "I can tell I got everything" and "I know what does not exist". Delivering subgroups out of order in different streams breaks this property.

That is a design requirement for SUBSCRIBE too. Otherwise, it's not possible to implement a HLS/DASH-like experience where media is buffered instead of skipped.

ex. A player issues a SUBSCRIBE and receives group 6 and group 8. The player buffers waiting for 7, but it may never arrive if the group was dropped or never existed.

I understand the design that head-of-line blocking acts an implicit signal that 7 doesn't exist because 8 was served next. However, you could have an explicit message instead: `OBJECT_DROPPED` or something. Same result but without the performance ramifications of HoLB and it would also work for SUBSCRIBE.

This would also take the guess-work out of PUBLISH_DONE's `Stream Count`. Every object between start-end would be accounted for, either received via a data stream or skipped via a `OBJECT_DROPPED` range.

PR #1382, opened 2025-11-28
Clarify datagrams and subgroups

### LINKED ISSUE #1381: Datagrams and streams intermixed in the same subgroup.
PR #1350 allowed Datagrams and streams to coexist on the same track, as agreed in Toronto.

Here are the minutes of that discussion: https://datatracker.ietf.org/doc/minutes-interim-2025-moq-22-202509251330/ (Ctrl+F for "Delivery Timeout Mix and Match")

This PR is missing two parts of that discussion:
1) We agreed that a single subgroup cannot contain both datagrams and streams. 
2) I noted that we have to encode the subgroup ID in the DATAGRAM frame, or reserve a specific subgroup ID for datagrams, or something. The current mapping where subgroup_id == object_id does not work if subgroups have arbitrary IDs. I would propose that subgroup ID zero be reserved for datagrams. Datagrams exist outside of stream priorities, so that sort of works. If pub and sub priorities,  and group IDs , are equal, it makes sense that datagrams would be delivered to QUIC first. Zero is a special subgroup that has no implied dependencies between  its objects.

Upon further reflection, I'm not sure that the current state completely breaks any functions of the protocol, but it does confuse the idea that a subgroup has a particular priority. I would not like to wave through this change to the object model without at least discussing it.

martinduke:
Fixes #1381

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -536,18 +536,24 @@ Track Namespace Field {
 * Track Namespace Field Value: A sequence of bytes that forms a Track Namespace
   Field.
 
+Each Track Namespace Field Value MUST contain at least one byte. If an endpoint
+receives a Track Namespace Field with a Track Namespace Field Length of 0, it
+MUST close the session with a `PROTOCOL_VIOLATION`.
+
 The structured nature of Track Namespace allows relays and applications to
 manipulate prefixes of a namespace. If an endpoint receives a Track Namespace
 consisting of 0 or greater than 32 Track Namespace Fields, it MUST close the
 session with a `PROTOCOL_VIOLATION`.
 
-Track Name is a sequence of bytes that identifies an individual track within the
-namespace.
+Track Name is a sequence of bytes, possibly empty, that identifies an individual
+track within the namespace.
 
-The maximum total length of a Full Track Name is 4,096 bytes. The length of a Full
-Track Name is computed as the sum of the Track Namespace Field Length fields
-and the Track Name Length field.  If an endpoint receives a Full Track Name
-exceeding this length, it MUST close the session with a `PROTOCOL_VIOLATION`.
+The maximum total length of a Full Track Name is 4,096 bytes. The length of a
+Full Track Name is computed as the sum of the Track Namespace Field Length
+fields and the Track Name Length field. The length of a Track Namespace is the
+sum of the Track Namespace Field Length fields. If an endpoint receives a Track
+Namespace or a Full Track Name exceeding 4,096 bytes, it MUST close the session
+with a `PROTOCOL_VIOLATION`.
 
 In this specification, both the Track Namespace Fields and the Track Name
 are not constrained to a specific encoding. They carry a sequence of bytes and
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -3626,8 +3629,8 @@ the prior Object, the Subscriber MUST close the session with a
 The Extensions structure is defined in {{object-extensions}}.
 
 When encoding an Object with a Forwarding Preference of "Datagram" (see
-{{object-properties}}), the Publisher treats it as having a Subgroup ID equal to
-the Object ID.
+{{object-properties}}), the object has no Subgroup ID. The publisher MUST SET bit 0x40 to '1'. When 0x40 is
+set, it SHOULD set the two least significant bits to zero and the subscriber MUST ignore the bits.
```
```suggestion
{{object-properties}}), the object has no Subgroup ID. The publisher MUST SET bit 0x40 to '1'.
When 0x40 is set, it SHOULD set the two least significant bits to zero and the subscriber
MUST ignore the bits.
```

draft-ietf-moq-transport.md
```diff
@@ -401,15 +401,16 @@ using at least as many streams as there are Subgroups,
...
+way, and the description in the remainder of this section does not apply.
 
 Streams offer in-order reliable delivery and the ability to cancel sending and
 retransmission of data. Furthermore, many QUIC and WebTransport implementations
 offer the ability to control the relative scheduling priority of pending stream
 data.
 
-Every object within a Group belongs to exactly one Subgroup.
+Every object within a Group belongs to exactly one Subgroup (but zero if its
+forwarding preference is Datagram).
```
```suggestion
Every Object within a Group belongs to exactly one Subgroup or Datagram.
```

draft-ietf-moq-transport.md
```diff
@@ -3077,7 +3078,9 @@ An endpoint that receives an unknown stream or datagram type MUST close the
 session.
 
 Every Object has a 'Object Forwarding Preference' and the Original Publisher
-MAY mix different forwarding preference within a single track.
+MAY mix different forwarding preference within a single track. As a Subgroup
+is by definition a set of objects delivered with Subgroup forwarding preference,
+objects with a Subgroup ID have the same forwarding preference.
```
ianswett: I don't think the second sentence is necessary, but I tried to make it clearer in case you want to keep it.

```suggestion
MAY use both Subgroups and Datagrams within a Group or Track. A Subgroup
is a set of Objects with the Subgroup forwarding preference, so Objects
within a Subgroup have the same forwarding preference.
```

yekuiwang: "within a Subgroup ID" should be "within the same Subgroup ID" or  "within a Subgroup"?

afrind:
Note this will now conflict with #1331 and require spilling over into a second byte for datagrams until we land updated varints.

afrind:
Resolved conflicts.  We can land this and datagrams in FETCH will have a 2 byte serialization field until #1016 lands, but that could be fine.

draft-ietf-moq-transport.md
afrind: >  A Subgroup is a set of Objects with the Subgroup forwarding preference, so Objects within a Subgroup have the same forwarding preference.

Is this necessary at all?

martinduke: Like all clarifications, it's not "necessary". I don't care that much if you keep it.

suhasHere (approved):
LGTM and also agree on simplifying the text as other have commented

draft-ietf-moq-transport.md
```diff
@@ -3188,7 +3188,9 @@ An endpoint that receives an unknown stream or datagram type MUST close the
...
```
```suggestion
MAY use both Subgroups and Datagrams within a Group or Track.
```

PR #1389, opened 2025-12-08
Move SUBSCRIBE, PUBLISH, FETCH, TRACK_STATUS, and PUBLISH_NAMESPACE to bidi streams

### LINKED ISSUE #632: What happens when FETCH_HEADER is lost?
I have the following test case where the publisher makes an API error and publishes an object on a fetch stream with the wrong length:

C->S: FETCH
S->C: FETCH_OK
S->C: open fetch stream (2), 
S->C: write FETCH_HEADER (2, delayed/lost)
S: API error, RESET_STREAM (2)

The client is expecting a fetch stream that will never come.  

The client needs to have a timer on the arrival of the FETCH_HEADER in any case, because a malicious or buggy server might never send one on purpose.  We should add text to this effect.

We can also short circuit the timer in cases like the above by using RESET_STREAM_AT (always deliver the FETCH_HEADER).

### LINKED ISSUE #1070: Is it a protocol error to get ANNOUNCE_CANCEL before ANNOUNCE_OK/ANNOUNCE_ERROR?
A well-behaved subscriber should not do this.

allowing ANNOUNCE_CANCEL first makes the state machine more complicated.

### LINKED ISSUE #1168: Ephemeral vs Long-lived Request IDs

### LINKED ISSUE #1247: Do we need REQUEST_BLOCKED
Could there be a better way to get new ID space ? Does limiting the request ID space help with any problems were are worried about ?

### LINKED ISSUE #1336: MAX_REQUEST_ID prevents pipelining
Even with ALPN version negotiation, both the client and server have to wait for the receipt of the `SERVER_SETUP`/`CLIENT_SETUP` before any subscriptions can be issued. Otherwise, they run a risk of a protocol violation by sending any requests while the default `MAX_REQUEST_ID` is still 0. 

tl;dr the RTT that could be saved by ALPN is prevented by MAX_REQUEST_ID.

### LINKED ISSUE #1343: Cancelling a subscription can be slow

### LINKED ISSUE #1351: Do we need stricter text for Request IDs?
The draft currently says:

>  The client's Request ID starts at 0 and are even and the server's Request ID starts at 1 and are odd

and

> The Request ID increments by 2 with each FETCH, ...

It also says

> If an endpoint receives a Request ID that is not valid for the peer, or a new request with a Request ID that is not the next in sequence or exceeds the received MAX_REQUEST_ID, it MUST close the session with INVALID_REQUEST_ID.

I know many, me included, have translated to our stacks mandating 0 as a starting point for clients, 1 for servers, and mandatory increment steps of +2, meaning that if I get a Request ID different from 0 as the first ID from a client, I'll close the connection, and I'll do the same if I get 0 first and 4 next.

The issue I see is that there's no MUST for the starting points, and for the increments either. This means that a loose interpretation could have people see them as a generic starting point and a need for always going up, but not as strictly as they should: meaning that 4, 10, 68 could be seen as a valid always growing sequence of Request IDs from a client. I did meet such a use case in my interop, where one of the clients that hit my relay was using Request ID 8 as the first ID, rather than 0.

ianswett:
Removes UNSUBSCRIBE, FETCH_CANCEL, and PUBLISH_NAMESPACE_DONE, replacing them with STOP_SENDING.

Removes MAX_REQUESTS and REQUESTS_BLOCKED.  QUIC Stream limits are used to limit the number of simultaneous requests.

Adds a 'Dependent RequestID' Message Parameter that allows one to make a Request dependent upon a prior Request to achieve the HoLB functionality that is lost by moving to multiple streams.

Fixes #1168
Fixes #1247
Fixes #1336
Fixes #1343
Fixes #1351
Fixes #1070

Could fix #632 if we moved the Fetch Objects to the Response stream
Improves Startup Latency #83 

This PR currently does not move FETCH response Objects to the response of the bidi stream, but it could.  The only negative is we'd be mixing response Objects and any REQUST_OK control messages that result from a REQUEST_UPDATE.

Potential future benefits:
* Provides an obvious place to put extensions like @vasilvv Object ACK proposal
* Could provide data on the Subscription or Track, such as the Largest Object in a Group, or Objects that weren't sent due to Delivery Timeout.

Closes #1505

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2557,7 +2557,7 @@ SUBSCRIBE Message {
   Type (vi64) = 0x3,
   Length (16),
   Request ID (vi64),
-  Required Request ID (vi64),
+  Required Request ID Delta (vi64),
   Track Namespace (..),
   Track Name Length (vi64),
   Track Name (..),
@@ -2569,7 +2569,7 @@ SUBSCRIBE Message {
 
 * Request ID: See {{request-id}}.
 
-* Required Request ID: See {{required-request-id}}.
+* Required Request ID Delta: See {{required-request-id}}.
 
 * Track Namespace: Identifies the namespace of the track as defined in
   ({{track-name}}).
@@ -2640,7 +2640,7 @@ REQUEST_UPDATE Message {
   Type (vi64) = 0x2,
   Length (16),
   Request ID (vi64),
-  Required Request ID (vi64),
+  Required Request ID Delta (vi64),
   Number of Parameters (vi64),
   Parameters (..) ...
 }
@@ -2649,7 +2649,7 @@ REQUEST_UPDATE Message {
 
 * Request ID: See {{request-id}}.
 
-* Required Request ID: See {{required-request-id}}.
+* Required Request ID Delta: See {{required-request-id}}.
 
 * Parameters: The parameters are defined in {{message-params}}.
 
@@ -2685,7 +2685,7 @@ PUBLISH Message {
   Type (vi64) = 0x1D,
   Length (16),
   Request ID (vi64),
-  Required Request ID (vi64),
+  Required Request ID Delta (vi64),
   Track Namespace (..),
   Track Name Length (vi64),
   Track Name (..),
@@ -2699,7 +2699,7 @@ PUBLISH Message {
 
 * Request ID: See {{request-id}}.
 
-* Required Request ID: See {{required-request-id}}.
+* Required Request ID Delta: See {{required-request-id}}.
 
 * Track Namespace: Identifies a track's namespace as defined in ({{track-name}})
 
@@ -2965,7 +2965,7 @@ FETCH Message {
   Type (vi64) = 0x16,
   Length (16),
   Request ID (vi64),
-  Required Request ID (vi64),
+  Required Request ID Delta (vi64),
   Fetch Type (vi64),
   [Standalone (Standalone Fetch),]
   [Joining (Joining Fetch),]
@@ -2977,7 +2977,7 @@ FETCH Message {
 
 * Request ID: See {{request-id}}.
 
-* Required Request ID: See {{required-request-id}}.
+* Required Request ID Delta: See {{required-request-id}}.
 
 * Fetch Type: Identifies the type of Fetch, whether Standalone, Relative
   Joining or Absolute Joining.
@@ -3119,7 +3119,7 @@ PUBLISH_NAMESPACE Message {
   Type (vi64) = 0x6,
   Length (16),
   Request ID (vi64),
-  Required Request ID (vi64),
+  Required Request ID Delta (vi64),
   Track Namespace (..),
   Number of Parameters (vi64),
   Parameters (..) ...
@@ -3129,7 +3129,7 @@ PUBLISH_NAMESPACE Message {
 
 * Request ID: See {{request-id}}.
 
-* Required Request ID: See {{required-request-id}}.
+* Required Request ID Delta: See {{required-request-id}}.
 
 * Track Namespace: Identifies a track's namespace as defined in
   {{track-name}}.
@@ -3190,7 +3190,7 @@ SUBSCRIBE_NAMESPACE Message {
   Type (vi64) = 0x11,
   Length (16),
   Request ID (vi64),
-  Required Request ID (vi64),
+  Required Request ID Delta (vi64),
   Track Namespace Prefix (..),
   Subscribe Options (vi64),
   Number of Parameters (vi64),
@@ -3201,7 +3201,7 @@ SUBSCRIBE_NAMESPACE Message {
 
 * Request ID: See {{request-id}}.
 
-* Required Request ID: See {{required-request-id}}.
+* Required Request ID Delta: See {{required-request-id}}.
 
 * Track Namespace Prefix: A Track Namespace structure as described in
   {{track-name}} with between 0 and 32 Track Namespace Fields.  This prefix is


draft-ietf-moq-transport.md
```diff
@@ -2945,19 +2901,21 @@ received a SUBSCRIBE message, except it does not create downstream subscription
...
 does not send PUBLISH_DONE for this request, and the subscriber cannot send
-REQUEST_UPDATE or UNSUBSCRIBE.
+REQUEST_UPDATE.
 
 ## PUBLISH_NAMESPACE {#message-pub-ns}
-
-The publisher sends the PUBLISH_NAMESPACE control message to advertise that it
-has tracks available within a Track Namespace. The receiver verifies the
-publisher is authorized to publish tracks under this namespace.
+
```
```suggestion

```

draft-ietf-moq-transport.md
```diff
@@ -722,19 +723,21 @@ can be used.
...
-A unidirectional stream containing Objects or bidirectional stream(s) containing a
-SUBSCRIBE_NAMESPACE could arrive prior to the control stream, in which case the
-data SHOULD be buffered until the control stream arrives and setup is complete.
-If an implementation does not want to buffer, it MAY reset other bidirectional
-streams before the session and control stream are established.
+messages defined in {{message}}. The control stream begins with a single
+CLIENT_SETUP message.
+
+In addition to the control stream, this specification uses bidirectional streams
+that begin with five types of message: TRACK_STATUS, SUBSCRIBE, PUBLISH, FETCH,
```
```suggestion
that begin with six types of message: TRACK_STATUS, SUBSCRIBE, PUBLISH, FETCH,
```

draft-ietf-moq-transport.md
```diff
@@ -924,9 +927,9 @@ successful, the subscription moves to the `Established` state and can
 be updated by the subscriber using REQUEST_UPDATE.  Either endpoint
 can terminate an `Established` subscription, moving it to the
 `Terminated` state.  The subscriber terminates a subscription in the
-`Pending (Subscriber)` or `Established` states using
-UNSUBSCRIBE, the publisher terminates a subscription in the `Pending
-(Publisher)` or `Established` states using PUBLISH_DONE.
+`Pending (Subscriber)` or `Established` states by cancelling the
+bidirectional stream. The publisher terminates a subscription in the
```
```suggestion
`Pending (Subscriber)` or `Established` states by sending STOP_SENDING.
The publisher terminates a subscription in the
```

draft-ietf-moq-transport.md
```diff
@@ -951,7 +954,7 @@ REQUEST_ERROR |    SUBSCRIBE_OK |    | PUBLISH_OK       | REQUEST_ERROR
               |            |             |       | REQUEST_UPDATE
               |            +-------------+ <-----+
               |                 |    |                  |
-              +---- UNSUBSCRIBE |    | PUBLISH_DONE ----+
+              +-- Cancel Stream |    | PUBLISH_DONE ----+
```
```suggestion
              +--- STOP_SENDING |    | PUBLISH_DONE ----+
```

draft-ietf-moq-transport.md
```diff
@@ -1014,13 +1017,13 @@ the Subscriber dropping Objects if its buffering limits are exceeded (see
 
 ### Subscription State Management
 
-A subscriber keeps subscription state until it sends UNSUBSCRIBE, or after
+A subscriber keeps subscription state until it cancels the stream, or after
```
```suggestion
A subscriber keeps subscription state until it sends STOP_SENDING, or after
```

draft-ietf-moq-transport.md
```diff
@@ -1014,13 +1017,13 @@ the Subscriber dropping Objects if its buffering limits are exceeded (see
...
 receipt of a PUBLISH_DONE or REQUEST_ERROR. Note that PUBLISH_DONE does not
 usually indicate that state can immediately be destroyed, see
 {{message-publish-done}}.
 
 The Publisher can destroy subscription state as soon as it has received
-UNSUBSCRIBE. It MUST reset any open streams associated with the SUBSCRIBE.
+Stop Sending. It MUST reset any open streams associated with the SUBSCRIBE.
```
```suggestion
STOP_SENDING. It MUST reset any open streams associated with the SUBSCRIBE.
```

draft-ietf-moq-transport.md
```diff
@@ -1312,8 +1314,8 @@ publisher priority.  The ordering in those cases is implementation-defined,
 though the expectation is that all subscriptions will be able to send some data.
 
 Given the critical nature of control messages and their relatively
-small size, the control stream SHOULD be prioritized higher than all
-subscribed Objects.
+small size, the control stream and bid request streams SHOULD be prioritized
```
```suggestion
small size, the control stream and bidi request streams SHOULD be prioritized
```

draft-ietf-moq-transport.md
```diff
@@ -1485,9 +1487,9 @@ The authorization and identification of the publisher depends on the way the
 relay is managed and is application specific.
 
 When a publisher wants to stop new subscriptions for a published namespace it
-sends a PUBLISH_NAMESPACE_DONE. A subscriber indicates it will no longer
-subcribe to Tracks in a namespace it previously responded REQUEST_OK
-to by sending a PUBLISH_NAMESPACE_CANCEL.
+sends a STOP_SENDING on the bidi stream to withdraw the PUBLISH_NAMESPACE.
+A subscriber indicates it will no longer subcribe to Tracks in a namespace
```
```suggestion
A subscriber indicates it will no longer subscribe to Tracks in a namespace
```

draft-ietf-moq-transport.md
```diff
@@ -2396,8 +2392,8 @@ time to receive objects in the future.
 
 ## SUBSCRIBE_OK {#message-subscribe-ok}
 
-A publisher sends a SUBSCRIBE_OK control message for successful
-subscriptions.
+A publisher sends a SUBSCRIBE_OK as the first message on the incoming
+bidi stream for successful subscriptions.
```
```suggestion
A publisher sends a SUBSCRIBE_OK as the first response message on the
bidi stream for successful subscriptions.
```

draft-ietf-moq-transport.md
```diff
@@ -3374,8 +3313,11 @@ Header field values.
 
 Streams aside from the control stream MAY be canceled due to congestion
 or other reasons by either the publisher or subscriber. Early termination of a
-stream does not affect the MoQ application state, and therefore has no
-effect on outstanding subscriptions.
+unidirectional stream does not affect the MoQ application state, and therefore has
+no effect on outstanding subscriptions. Termination of a bidi request stream
+terminates the Subscription, Fetch or Track Status request. Publishers SHOULD NOT
+RESET the bidi response stream, and instead SHOULD send the appropriate final
+message indicating the response is complete and FIN the bidi stream.
```
```suggestion
terminates the Subscription, Fetch, Track Status, Publish Namespace, or Subscribe Namespace
request. Publishers SHOULD NOT RESET the bidi response stream, and instead SHOULD send
the appropriate final message indicating the response is complete and FIN the bidi stream.
```

draft-ietf-moq-transport.md
```diff
@@ -565,12 +565,13 @@ include:
...
 The above list of conditions is not considered exhaustive.
 
-When a subscriber detects a Malformed Track, it MUST UNSUBSCRIBE any
-subscription and FETCH_CANCEL any fetch for that Track from that publisher, and
-SHOULD deliver an error to the application.  If a relay detects a Malformed
-Track, it MUST immediately terminate downstream subscriptions with PUBLISH_DONE
-and reset any fetch streams with Status Code `MALFORMED_TRACK`. Object(s)
-triggering Malformed Track status MUST NOT be cached.
+When a subscriber detects a Malformed Track, it MUST send unsubsribe any
+subscription and cancel any fetch for that Track from that publisher by sending
```
afrind: EPARSE.  it must cancel the corresponding subscription or fetch

draft-ietf-moq-transport.md
```diff
@@ -1203,20 +1207,18 @@ An endpoint SHOULD report the reception of a REQUEST_OK or
...
 namespace. This might be especially useful in upload or chat applications. A
-subscriber MUST send exactly one REQUEST_OK or REQUEST_ERROR
-in response to a PUBLISH_NAMESPACE. The publisher SHOULD close the session with
-a protocol error if it receives more than one.
+subscriber MUST send exactly one REQUEST_OK or REQUEST_ERROR as the first
+message on the bidi stream in response to a PUBLISH_NAMESPACE. The publisher
+SHOULD close the session with a protocol error if it receives more than one.
 
-A PUBLISH_NAMESPACE_DONE message withdraws a previous PUBLISH_NAMESPACE,
+A PUBLISH_NAMESPACE is withdrawn by sending STOP_SENDING on the bidi stream,
```
afrind: Why is this not a RESET_STREAM / FIN on the write half?

draft-ietf-moq-transport.md
```diff
@@ -1203,20 +1207,18 @@ An endpoint SHOULD report the reception of a REQUEST_OK or
...
 although it is not a protocol error for the subscriber to send a SUBSCRIBE or
-FETCH message for a track in a namespace after receiving an
-PUBLISH_NAMESPACE_DONE.
+FETCH message for a track in a namespace after the namespace is withdrawn.
 
 A subscriber can send PUBLISH_NAMESPACE_CANCEL to revoke acceptance of an
```
afrind: This should be STOP_SENDING on the bidi + FIN or RESET.

draft-ietf-moq-transport.md
```diff
@@ -1312,8 +1314,8 @@ publisher priority.  The ordering in those cases is implementation-defined,
 though the expectation is that all subscriptions will be able to send some data.
 
 Given the critical nature of control messages and their relatively
-small size, the control stream SHOULD be prioritized higher than all
-subscribed Objects.
+small size, the control stream and bidi request streams SHOULD be prioritized
```
afrind: relative prioritization of these?

draft-ietf-moq-transport.md
```diff
@@ -1400,7 +1402,7 @@ given Track.
 
 There is no specified limit to the number of publishers of a Track Namespace or
 Track.  An implementation can use mechanisms such as REQUEST_ERROR,
-UNSUBSCRIBE or PUBLISH_NAMESPACE_CANCEL if it cannot
+STOP_SENDING or PUBLISH_NAMESPACE_CANCEL if it cannot
```
afrind: maybe "unsubscribing via STOP_SENDING"

draft-ietf-moq-transport.md
```diff
@@ -1485,9 +1487,9 @@ The authorization and identification of the publisher depends on the way the
 relay is managed and is application specific.
 
 When a publisher wants to stop new subscriptions for a published namespace it
-sends a PUBLISH_NAMESPACE_DONE. A subscriber indicates it will no longer
-subcribe to Tracks in a namespace it previously responded REQUEST_OK
-to by sending a PUBLISH_NAMESPACE_CANCEL.
+sends a STOP_SENDING on the bidi stream to withdraw the PUBLISH_NAMESPACE.
```
afrind: See above

draft-ietf-moq-transport.md
```diff
@@ -1642,16 +1642,12 @@ The following Message Types are defined:
...
 |-------|-----------------------------------------------------|
 | 0x6   | PUBLISH_NAMESPACE  ({{message-pub-ns}})             |
 |-------|-----------------------------------------------------|
 | 0x8   | NAMESPACE  ({{message-namespace}})                  |
 |-------|-----------------------------------------------------|
-| 0x9   | PUBLISH_NAMESPACE_DONE  ({{message-pub-ns-done}})   |
-|-------|-----------------------------------------------------|
 | 0xE   | NAMESPACE_DONE  ({{message-namespace-done}})        |
 |-------|-----------------------------------------------------|
 | 0xC   | PUBLISH_NAMESPACE_CANCEL ({{message-pub-ns-cancel}})|
```
afrind: let's kill this too

afrind (commented):
I think we should be crisper about what to do with the write half of the stream when you stop_sending the read half.

Possibly defining unsubscribe as an action and using that instead of STOP_SENDING will keep the existing clarity.

draft-ietf-moq-transport.md
ianswett: I guess the control stream should be higher, because the only things left on it are SETUP and GOAWAY?

draft-ietf-moq-transport.md
ianswett: I agree, but currently it has an Error Code and Reason Phrase, though I'm not sure why?  It'd be easy enough to specify an Error Code if we wanted to, so maybe that's enough?

draft-ietf-moq-transport.md
```suggestion
When a subscriber detects a Malformed Track, it MUST cancel any corresponding
subscription or fetches for that Track from that publisher by sending
```

draft-ietf-moq-transport.md
```suggestion
unsubscribing via STOP_SENDING or PUBLISH_NAMESPACE_CANCEL if it cannot
```

draft-ietf-moq-transport.md
```diff
@@ -751,24 +752,41 @@ can be used.
...
-SUBSCRIBE_NAMESPACE could arrive prior to the control stream, in which case the
-data SHOULD be buffered until the control stream arrives and setup is complete.
-If an implementation does not want to buffer, it MAY reset other bidirectional
-streams before the session and control stream are established.
+messages defined in {{message}}. The control stream begins with a single
+CLIENT_SETUP message.
+
+In addition to the control stream, this specification uses bidirectional request
+streams that begin with six types of message: TRACK_STATUS, SUBSCRIBE, PUBLISH,
+FETCH, PUBLISH_NAMESPACE, and SUBSCRIBE_NAMESPACE. Bidirectional streams MUST NOT
```
```suggestion
In addition to the control stream, this specification uses bidirectional streams
to carry requests.  A request stream begins with one of these six message types:
TRACK_STATUS, SUBSCRIBE, PUBLISH, FETCH, PUBLISH_NAMESPACE, and
SUBSCRIBE_NAMESPACE. Bidirectional streams MUST NOT
```

draft-ietf-moq-transport.md
```diff
@@ -751,24 +752,41 @@ can be used.
...
+begin with any other message type unless negotiated. If they do, the peer MUST
+close the Session with a Protocol Violation. Objects are sent on unidirectional
```
```suggestion
close the Session with a `PROTOCOL_VIOLATION`. Objects are sent on unidirectional
```

draft-ietf-moq-transport.md
```diff
@@ -751,24 +752,41 @@ can be used.
...
+streams.
+
+A unidirectional stream containing Objects or bidirectional stream(s) beginning
+with a message other than CLIENT_SETUP could arrive prior to the control stream,
+in which case the stream data SHOULD be buffered until the control stream arrives
+and setup is complete. If an implementation does not want to buffer or if it is
+certain the message type is not supported, it MAY reset other bidirectional streams
+before the session and control stream are established.
```
```suggestion
and setup is complete. If an implementation does not want to buffer or if
the message type is not supported, it MAY reset such bidirectional streams
before the session and control stream are established.
```

draft-ietf-moq-transport.md
```diff
@@ -751,24 +752,41 @@ can be used.
...
 
 The control stream MUST NOT be closed at the underlying transport layer during the
 session's lifetime.  Doing so results in the session being closed as a
 `PROTOCOL_VIOLATION`.
 
+### Request Cancellation and Rejection {#request-cancellation}
+
+Once a request stream has been opened, the request MAY be cancelled by either
+endpoint. Subscribers cancel requests if the response is no longer of interest;
+Publishers cancel requests if they are unable to or choose not to respond.
```
```suggestion
endpoint. Senders cancel requests if the response is no longer of interest;
Receivers cancel requests if they are unable to or choose not to respond.
```

draft-ietf-moq-transport.md
```diff
@@ -751,24 +752,41 @@ can be used.
...
+
+Implementations SHOULD cancel requests by abruptly terminating any directions of a
+stream that are still open. To do so, an implementation resets the sending parts of
+streams and aborts reading on the receiving parts of streams; see ({{QUIC, Section 2.4}}).
```
afrind: I know this is the RFC9114 language but I think it could be updated to match usage in the document - 
```suggestion
Implementations SHOULD cancel requests by abruptly terminating any directions of
a stream that are still open using RESET_STREAM / RESET_STREAM_AT or
STOP_SENDING.
```

draft-ietf-moq-transport.md
```diff
@@ -751,24 +752,41 @@ can be used.
...
+
+When the Publisher rejects a request without performing any application processing,
+The PUBLISHER SHOULD send a REQUEST_ERROR and FINs the stream.
```
```suggestion
The PUBLISHER SHOULD send a REQUEST_ERROR and FIN the stream.
```

draft-ietf-moq-transport.md
```diff
@@ -1157,13 +1166,14 @@ groups. A publisher that does will begin the next group as soon as practical.
...
-with FETCH_CANCEL, but MUST send FETCH_CANCEL.
+A subscriber keeps FETCH state until it cancels the request
+((see {{request-cancellation}})), receives REQUEST_ERROR, or receives a FIN or
+RESET_STREAM for the FETCH data stream. If the data stream is already open,
+the subscriber MAY send STOP_SENDING for the data stream in addition to
+STOP_SENDING on the bidi stream, but MUST send STOP_SENDING for the bidi stream.
 
 The Publisher can destroy fetch state as soon as it has received a
-FETCH_CANCEL. It MUST reset any open streams associated with the FETCH. It can
+STOP_SENDING. It MUST reset any open streams associated with the FETCH. It can
```
afrind: Or "as soon as it has been cancelled by the subscriber"

draft-ietf-moq-transport.md
```diff
@@ -1232,20 +1242,19 @@ An endpoint SHOULD report the reception of a REQUEST_OK or
...
-a protocol error if it receives more than one.
+subscriber MUST send exactly one REQUEST_OK or REQUEST_ERROR as the first
+message on the bidi stream in response to a PUBLISH_NAMESPACE. The publisher
+SHOULD close the session with a protocol error if it receives more than one.
 
-A PUBLISH_NAMESPACE_DONE message withdraws a previous PUBLISH_NAMESPACE,
-although it is not a protocol error for the subscriber to send a SUBSCRIBE or
-FETCH message for a track in a namespace after receiving an
-PUBLISH_NAMESPACE_DONE.
+A PUBLISH_NAMESPACE is withdrawn by cancelling the Request
```
afrind: request or Request -- let's be consistent.

draft-ietf-moq-transport.md
afrind: Agree it should go 

1. control
2. requests
3. data

Should there be any guidance about prioritizing the request/response streams relative to _each other_?  Particularly SUBSCRIBE_NAMESPACE responses can be larger.

draft-ietf-moq-transport.md
```diff
@@ -1513,10 +1522,10 @@ an explicit PUBLISH_NAMESPACE.
 The authorization and identification of the publisher depends on the way the
 relay is managed and is application specific.
 
-When a publisher wants to stop new subscriptions for a published namespace it
-sends a PUBLISH_NAMESPACE_DONE. A subscriber indicates it will no longer
-subcribe to Tracks in a namespace it previously responded REQUEST_OK
-to by sending a PUBLISH_NAMESPACE_CANCEL.
+When a publisher wants to stop new subscriptions for a published namespace, it
+cancels the request (see {{request-cancellation}}) to withdraw the PUBLISH_NAMESPACE.
+A subscriber also indicates it will no longer subscribe to Tracks in a namespace
```
```suggestion
A subscriber indicates it will no longer subscribe to Tracks in a namespace
```

draft-ietf-moq-transport.md
```diff
@@ -1513,10 +1522,10 @@ an explicit PUBLISH_NAMESPACE.
...
+by cancelling the request.
```
```suggestion
by cancelling the PUBLISH_NAMESPACE request.
```

draft-ietf-moq-transport.md
```diff
@@ -1617,8 +1626,8 @@ prioritize sending Objects based on {{priorities}}.
 
 # Control Messages {#message}
 
-MOQT uses a single bidirectional stream to exchange control messages, as
-defined in {{session-init}}.  Every single message on the control stream is
+MOQT uses bidirectional streams to exchange control messages, as
+defined in {{session-init}}.  Every single message on a control stream is
```
```suggestion
defined in {{session-init}}.  Every message on a control or request stream is
```

draft-ietf-moq-transport.md
```diff
@@ -2423,14 +2337,13 @@ time to receive objects in the future.
 
 ## SUBSCRIBE_OK {#message-subscribe-ok}
 
-A publisher sends a SUBSCRIBE_OK control message for successful
-subscriptions.
+A publisher sends a SUBSCRIBE_OK as the first response message on the
+bidi stream for successful subscriptions.
```
afrind, line 2597: Do we need a section that explains that for a given request type, the first message on the response stream must be either REQUEST_ERROR or the request-specific ok message, and anything else is PROTOCOL_VIOLATION?

SUBSCRIBE: SUBSCRIBE_OK
FETCH: FETCH_OK
PUBLISH: PUBLISH_OK
TRACK_STATUS: REQUEST_OK
PUBLISH_NS: REQUEST_OK
SUBSCRIBE_NS: REQUEST_OK

draft-ietf-moq-transport.md
```diff
@@ -2583,32 +2461,30 @@ PUBLISH_OK.
...
 * Parameters: The parameters are defined in {{version-specific-params}}.
 
 TODO: A similar section to SUBSCRIBE about how the publisher handles a
 filter that is entirely behind Largest Object or is otherwise invalid.
 
 ## PUBLISH_DONE {#message-publish-done}
 
-A publisher sends a `PUBLISH_DONE` message to indicate it is done publishing
-Objects for that subscription.  The Status Code indicates why the subscription
+A publisher sends a `PUBLISH_DONE` message as the final message on the
```
afrind: Do we need to say something about when FIN comes, or if another message comes before FIN?

draft-ietf-moq-transport.md
```diff
@@ -3403,8 +3206,11 @@ Header field values.
 
 Streams aside from the control stream MAY be canceled due to congestion
 or other reasons by either the publisher or subscriber. Early termination of a
-stream does not affect the MoQ application state, and therefore has no
-effect on outstanding subscriptions.
+unidirectional stream does not affect the MoQ application state, and therefore has
+no effect on outstanding subscriptions. Termination of a bidi request stream
+terminates the Subscription, Fetch, Track Status, Publish Namespace, or Subscribe Namespace
+request. Publishers SHOULD NOT RESET the bidi response stream, and instead SHOULD send
+the appropriate final message indicating the response is complete and FIN the bidi stream.
```
afrind: 1) subscribers and publisher both respond to requests
2) what i've already sent request_ok before cancellation, are you saying I should send request_error?  How is that better than RESET_STREAM + Cancelled?

draft-ietf-moq-transport.md
```diff
@@ -3557,7 +3363,7 @@ stream, it MUST use a RESET_STREAM or RESET_STREAM_AT
 not limited to:
 
 * An Object in an open Subgroup exceeding its Delivery Timeout
-* Early termination of subscription due to an UNSUBSCRIBE message
+* Early termination of subscription due to stream cancellation
```
```suggestion
* Early termination of subscription due to request cancellation
```

draft-ietf-moq-transport.md
```diff
@@ -3638,10 +3444,9 @@ INTERNAL_ERROR (0x0):
 : An implementation specific error.
 
 CANCELLED (0x1):
-: The subscriber requested cancellation via UNSUBSCRIBE, FETCH_CANCEL or
-  STOP_SENDING, or the publisher ended the subscription, in which case
-  PUBLISH_DONE ({{message-publish-done}}) will have a more detailed status
-  code.
+: The subscriber requested cancellation via STOP_SENDING, or the publisher
```
afrind: Would it be more clear to have separate CANCEL_REQUEST and CANCELLATION_REQUESTED codes?

draft-ietf-moq-transport.md
```diff
@@ -3638,10 +3444,9 @@ INTERNAL_ERROR (0x0):
...
+  ended the subscription, in which case PUBLISH_DONE ({{message-publish-done}})
+  will have a more detailed status code.
```
afrind: This existing text doesn't cover fetch data streams.

draft-ietf-moq-transport.md
```diff
@@ -2828,7 +2704,8 @@ FETCH Message {
 ~~~
 {: #moq-transport-fetch-format title="MOQT FETCH Message"}
 
-* Request ID: See {{request-id}}.
+* Fetch ID: An ID used to identify the unidirectional stream containing
+  Objects requested by the Fetch. The ID MUST be unique within a Session.
```
afrind: I get that subscribe ID and fetch ID only need to be unique.  Should we offer a recommendation that implementations do anything in particular?

draft-ietf-moq-transport.md
ianswett: I think Request is better, in which case we should probably define the term as well?

draft-ietf-moq-transport.md
ianswett: Maybe, though I might be inclined to do that later.  There might be other text about Requests to reorg into its own section.

draft-ietf-moq-transport.md
```suggestion
  Objects requested by the Fetch. The ID MUST be unique within a Session.
  To ensure Fetch IDs are not reused, implementations SHOULD issue them
  sequentially.
```

draft-ietf-moq-transport.md
ianswett: Done. I'm inclined to leave out guidance on prioritizing them relative to one anther.  Hopefully they won't typically be that large?

draft-ietf-moq-transport.md
```diff
@@ -2583,32 +2462,30 @@ PUBLISH_OK.
...
 TODO: A similar section to SUBSCRIBE about how the publisher handles a
 filter that is entirely behind Largest Object or is otherwise invalid.
 
 ## PUBLISH_DONE {#message-publish-done}
 
-A publisher sends a `PUBLISH_DONE` message to indicate it is done publishing
-Objects for that subscription.  The Status Code indicates why the subscription
+A publisher sends a `PUBLISH_DONE` message as the final message on the
+subscription's bidi stream to indicate it is done publishing Objects for that
+subscription.  The Status Code indicates why the subscription
```
```suggestion
A publisher sends a `PUBLISH_DONE` message as the final message before
closing the subscription's bidi stream to indicate it is done publishing Objects
for that subscription.  The Status Code indicates why the subscription
```

draft-ietf-moq-transport.md
ianswett: I tried to clarify, PTAL.

draft-ietf-moq-transport.md
ianswett: I think it's on the SUBSCRIBE_DONE case when we'd like a FIN if possible?

draft-ietf-moq-transport.md
```diff
@@ -3403,8 +3209,11 @@ Header field values.
 
 Streams aside from the control stream MAY be canceled due to congestion
 or other reasons by either the publisher or subscriber. Early termination of a
-stream does not affect the MoQ application state, and therefore has no
-effect on outstanding subscriptions.
+unidirectional stream does not affect the MoQ application state, and therefore has
+no effect on outstanding subscriptions. Termination of a bidi request stream
+terminates the Subscription, Fetch, Track Status, Publish Namespace, or Subscribe Namespace
+request. Publishers SHOULD NOT RESET the bidi response stream, and instead SHOULD send
+the appropriate final message indicating the response is complete and FIN the bidi stream.
```
```suggestion
request. When possible, Publishers SHOULD send a SUBSCRIBE_DONE when terminating a
subscription instead of abruptly terminating the stream.
```

draft-ietf-moq-transport.md
```diff
@@ -3403,8 +3209,11 @@ Header field values.
 
 Streams aside from the control stream MAY be canceled due to congestion
 or other reasons by either the publisher or subscriber. Early termination of a
-stream does not affect the MoQ application state, and therefore has no
-effect on outstanding subscriptions.
+unidirectional stream does not affect the MoQ application state, and therefore has
+no effect on outstanding subscriptions. Termination of a bidi request stream
+terminates the Subscription, Fetch, Track Status, Publish Namespace, or Subscribe Namespace
+request. When possible, Publishers SHOULD send a SUBSCRIBE_DONE when terminating a
```
```suggestion
request. When possible, Publishers SHOULD send a PUBLISH_DONE when terminating a
```

draft-ietf-moq-transport.md
```diff
@@ -3638,10 +3447,9 @@ INTERNAL_ERROR (0x0):
 : An implementation specific error.
 
 CANCELLED (0x1):
-: The subscriber requested cancellation via UNSUBSCRIBE, FETCH_CANCEL or
-  STOP_SENDING, or the publisher ended the subscription, in which case
-  PUBLISH_DONE ({{message-publish-done}}) will have a more detailed status
-  code.
+: The subscriber requested cancellation via STOP_SENDING, or the publisher
+  ended the subscription, in which case PUBLISH_DONE ({{message-publish-done}})
+  will have a more detailed status code.
```
```suggestion
: The subscriber or publisher cancelled the Request. For Subscriptions,
  PUBLISH_DONE ({{message-publish-done}}) will have a more detailed status code.
```

draft-ietf-moq-transport.md
ianswett: Good point, I updated it.  PTAL.

draft-ietf-moq-transport.md
```diff
@@ -600,12 +600,13 @@ include:
...
-Track, it MUST immediately terminate downstream subscriptions with PUBLISH_DONE
-and reset any fetch streams with Status Code `MALFORMED_TRACK`. Object(s)
-triggering Malformed Track status MUST NOT be cached.
+When a subscriber detects a Malformed Track, it MUST cancel any corresponding
+subscription or fetches for that Track from that publisher
+(see {{request-cancellation}}), and SHOULD deliver an error to the application.
+If a relay detects a Malformed Track, it MUST immediately terminate downstream
+subscriptions with PUBLISH_DONE and reset any fetch streams with
+Status Code `MALFORMED_TRACK`. Object(s) triggering Malformed Track status
+MUST NOT be cached.
```
martinduke, line 695: Is the implication here that the publisher cannot serve any future requests for the track? Or is it only requests that contain the malformed object?

If the former, is there a time limit or is the track corrupted forever and ever?

If the latter, why do I have to terminate all FETCH regardless of Location?

draft-ietf-moq-transport.md
```diff
@@ -963,9 +973,9 @@ successful, the subscription moves to the `Established` state and can
 be updated by the subscriber using REQUEST_UPDATE.  Either endpoint
 can terminate an `Established` subscription, moving it to the
 `Terminated` state.  The subscriber terminates a subscription in the
-`Pending (Subscriber)` or `Established` states using
-UNSUBSCRIBE, the publisher terminates a subscription in the `Pending
-(Publisher)` or `Established` states using PUBLISH_DONE.
+`Pending (Subscriber)` or `Established` states by sending STOP_SENDING.
+The publisher terminates a subscription in the
+`Pending (Publisher)` or `Established` states using PUBLISH_DONE.
```
martinduke: PUBLISH_DONE + FIN

draft-ietf-moq-transport.md
martinduke: maybe rank them by subscriber priority?

draft-ietf-moq-transport.md
```diff
@@ -1444,10 +1454,10 @@ given Track.
 
 There is no specified limit to the number of publishers of a Track Namespace or
 Track.  An implementation can use mechanisms such as REQUEST_ERROR,
```
martinduke:
```suggestion
Track.  An implementation can use mechanisms such as REQUEST_ERROR or
```

draft-ietf-moq-transport.md
ianswett: This is existing text, I just tried to update it to remove the UNSUBSCRIBE and FETCH_CANCEL messages.

I think it's only requests that contain the malformed Object or Group?  Good point on FETCH.  I think the idea is that the publisher is corrupt or behaving poorly, so the Track shouldn't be fetched from them?

draft-ietf-moq-transport.md
```suggestion
`Pending (Publisher)` or `Established` states by sending PUBLISH_DONE
and closing the stream.
```

draft-ietf-moq-transport.md
ianswett: I don't have a strong preference on that.  Functionally, I don't think it matters, but it might be helpful for debugging?

draft-ietf-moq-transport.md
ianswett: I've changed it to only talk about subscriptions and SUBSCRIBE_DONE.

draft-ietf-moq-transport.md
ianswett: I clarified that control is higher priority than requests and allowed publishers to prioritize control stream by the subscriber priority, as Martin suggested.

draft-ietf-moq-transport.md
```diff
@@ -1356,8 +1367,9 @@ priority Objects if it expects higher priority Objects will be available to send
 in the near future or it wants to reserve some bandwidth for control messages.
 
 Given the critical nature of control messages and their relatively
-small size, the control stream SHOULD be prioritized higher than all
-subscribed Objects.
+small size, the control stream SHOULD be prioritized highest, followed by the
+bidi request streams and then all subscribed Objects. Control streams MAY be
```
afrind: Bidi request streams MAY?

draft-ietf-moq-transport.md
```suggestion
A PUBLISH_NAMESPACE is withdrawn by cancelling the request
```

draft-ietf-moq-transport.md
```suggestion
bidi request streams and then all subscribed Objects. Bidi request streams MAY be
```

draft-ietf-moq-transport.md
```diff
@@ -661,12 +661,13 @@ include:
...
-Track, it MUST immediately terminate downstream subscriptions with PUBLISH_DONE
-and reset any fetch streams with Status Code `MALFORMED_TRACK`. Object(s)
-triggering Malformed Track status MUST NOT be cached.
+When a subscriber detects a Malformed Track, it MUST cancel any corresponding
+subscription or fetches for that Track from that publisher
+(see {{request-cancellation}}), and SHOULD deliver an error to the application.
+If a relay detects a Malformed Track, it MUST immediately terminate downstream
+subscriptions with PUBLISH_DONE and reset any fetch streams with
+Status Code `MALFORMED_TRACK`. Object(s) triggering Malformed Track status
+MUST NOT be cached.
```
suhasHere, line 695: is this part of this PR ? Can we move it to a different PR

draft-ietf-moq-transport.md
```diff
@@ -865,20 +884,20 @@ PROTOCOL_VIOLATION (0x3):
...
-: The session was closed because the endpoint used a Request ID that was
-  smaller than or equal to a previously received request ID, or the least-
-  significant bit of the request ID was incorrect for the endpoint.
+: The endpoint received a Request ID with an incorrect least significant
+  bit for the sender, or a duplicate Request ID. See {{request-id}}.
+
+INVALID_REQUIRED_REQUEST_ID (0x7):
+: The endpoint received a Required Request ID that does not have the same
+  parity as the Request ID, or is greater than the Request ID. See
+  {{request-id}}.
```
```suggestion
  parity as the Request ID, or is less than 0. See {{required-request-id}}.
```

draft-ietf-moq-transport.md
```diff
@@ -1763,21 +1771,35 @@ Payload, which is defined by each message type.  If the length does not match
...
+session with `INVALID_REQUEST_ID`.
+
+## Required Request ID {#required-request-id}
+
+Every request message includes a Required Request ID field, which is a delta
+from the Request ID.  The receiver MUST deliver the referenced request to the
+application before delivering this request.  Required Request ID MUST have the
+same parity (least significant bit) as Request ID. If an endpoint receives a
+Required Request ID with different parity, it MUST close the session with
+`INVALID_REQUIRED_REQUEST_ID`. A Required Request ID value of 0 indicates
```
```suggestion
Every request message includes a Required Request ID Delta field, which is a delta
subtracted from Request ID to determine the Required Request ID. The receiver
MUST deliver the referenced request to the application before delivering this request.  Required Request ID MUST have the same parity (least significant bit) as Request ID
and Required Request ID Delta MUST NOT be larger than Request ID. If an endpoint
receives an invalid Required Request ID, it MUST close the session with
`INVALID_REQUIRED_REQUEST_ID`. A Required Request ID Delta value of 0 indicates
```

draft-ietf-moq-transport.md
```diff
@@ -1763,21 +1771,35 @@ Payload, which is defined by each message type.  If the length does not match
...
+
+If an endpoint receives a Request ID where the least significant bit is
+incorrect for the sender, or a duplicate Request ID, it MUST close the
+session with `INVALID_REQUEST_ID`.
+
+## Required Request ID {#required-request-id}
+
+Every request message includes a Required Request ID Delta field, which is a delta
+subtracted from Request ID to determine the Required Request ID. The receiver
+MUST deliver the referenced request to the application before delivering this request.  Required Request ID MUST have the same parity (least significant bit) as Request ID
```
afrind: This should be half-delta or delta-count. Odd numbers for delta never make sense.

might add clarity to explain how to compute required request id from the delta.

```suggestion
MUST deliver the referenced request to the application before delivering this request.
Required Request ID MUST have the same parity (least significant bit) as Request ID
```

vasilvv:
(I'm OOO, and haven't had a chance to read the full PR yet, but following up on the discussion we had at an editors/authors call)

I think we should handle blocks at a level higher than the requests level.  My simplified proposal:
- A message `WAIT_FOR x` causes the receiver to stop reading control stream until `x` is triggered, or timeout occurs.
- A message `UNBLOCK x` causes all `WAIT_FOR x` messages to unblock.
- Messages can occur on any control stream.
- We need some text about state tracking.

draft-ietf-moq-transport.md
```diff
@@ -822,24 +823,42 @@ can be used.
...
+begin with any other message type unless negotiated. If they do, the peer MUST
+close the Session with a `PROTOCOL_VIOLATION`. Objects are sent on unidirectional
+streams.
+
+A unidirectional stream containing Objects or bidirectional stream(s) beginning
+with a message other than CLIENT_SETUP could arrive prior to the control stream,
+in which case the stream data SHOULD be buffered until the control stream arrives
+and setup is complete. If an implementation does not want to buffer or if
+the message type is not supported, it MAY reset such bidirectional streams
+before the session and control stream are established.
```
```suggestion
before the session and control streams are established.
```

draft-ietf-moq-transport.md
```diff
@@ -822,24 +823,42 @@ can be used.
...
+Once a request stream has been opened, the request MAY be cancelled by either
+endpoint. Senders cancel requests if the response is no longer of interest;
+Receivers cancel requests if they are unable to or choose not to respond.
+
+Implementations SHOULD cancel requests by abruptly terminating any directions of
+a stream that are still open using RESET_STREAM / RESET_STREAM_AT or
+STOP_SENDING.
+
+When the Publisher rejects a request without performing any application processing,
+The PUBLISHER SHOULD send a REQUEST_ERROR and FIN the stream.
```
suhasHere: would this apply to both publisher and the subscriber ?

draft-ietf-moq-transport.md
```diff
@@ -865,20 +884,19 @@ PROTOCOL_VIOLATION (0x3):
...
 INVALID_REQUEST_ID (0x4):
-: The session was closed because the endpoint used a Request ID that was
-  smaller than or equal to a previously received request ID, or the least-
-  significant bit of the request ID was incorrect for the endpoint.
+: The endpoint received a Request ID with an incorrect least significant
+  bit for the sender, or a duplicate Request ID. See {{request-id}}.
+
+INVALID_REQUIRED_REQUEST_ID (0x7):
+: The endpoint received a Required Request ID that does not have the same
+  parity as the Request ID, or is less than 0. See {{required-request-id}}.
```
suhasHere: this is the first time I am seeing a reference to something less than 0. wonder why would we need to specify it ?

draft-ietf-moq-transport.md
```diff
@@ -1241,13 +1260,14 @@ groups. A publisher that does will begin the next group as soon as practical.
...
 
-A subscriber keeps FETCH state until it sends FETCH_CANCEL, receives
-REQUEST_ERROR, or receives a FIN or RESET_STREAM for the FETCH data stream. If the
-data stream is already open, it MAY send STOP_SENDING for the data stream along
-with FETCH_CANCEL, but MUST send FETCH_CANCEL.
+A subscriber keeps FETCH state until it cancels the request
+((see {{request-cancellation}})), receives REQUEST_ERROR, or receives a FIN or
+RESET_STREAM for the FETCH data stream. If the data stream is already open,
+the subscriber MAY send STOP_SENDING for the data stream in addition to
+STOP_SENDING on the bidi stream, but MUST send STOP_SENDING for the bidi stream.
```
```suggestion
RESET_STREAM for the FETCH data stream. If the data stream is already open,
the subscriber wishing to cancel the FETCH, MAY send STOP_SENDING  for both  
data stream and  the request bidi stream. It MUST always send STOP_SENDING for 
request the bidi stream.
```

draft-ietf-moq-transport.md
```diff
@@ -1241,13 +1260,14 @@ groups. A publisher that does will begin the next group as soon as practical.
...
 
 The Publisher can destroy fetch state as soon as it has received a
-FETCH_CANCEL. It MUST reset any open streams associated with the FETCH. It can
+STOP_SENDING. It MUST reset any open streams associated with the FETCH. It can
```
```suggestion
STOP_SENDING . It MUST reset any open streams (data or request) associated with the FETCH. It can
```

draft-ietf-moq-transport.md
```diff
@@ -1324,20 +1344,19 @@ An endpoint SHOULD report the reception of a REQUEST_OK or
...
 
-A subscriber can send PUBLISH_NAMESPACE_CANCEL to revoke acceptance of an
-PUBLISH_NAMESPACE, for example due to expiration of authorization
-credentials. The message enables the publisher to PUBLISH_NAMESPACE again with
-refreshed authorization, or discard associated state. After receiving an
-PUBLISH_NAMESPACE_CANCEL, the publisher does not send PUBLISH_NAMESPACE_DONE.
+A subscriber can cancel the request to revoke acceptance of a
+PUBLISH_NAMESPACE. If the reason for cancellation is expiration of
+authorization credentials, the publisher can PUBLISH_NAMESPACE again with
+refreshed authorization, or close the stream and discard associated state.
```
```suggestion
authorization credentials, the publisher can send PUBLISH_NAMESPACE again
on a new bidi stream with refreshed authorization, or close the stream and 
discard associated state.
```

draft-ietf-moq-transport.md
```diff
@@ -1324,20 +1344,19 @@ An endpoint SHOULD report the reception of a REQUEST_OK or
...
+(see {{request-cancellation}}), although it is not a protocol error for
+the subscriber to send a SUBSCRIBE or FETCH message for a track in a
+namespace after the namespace is withdrawn.
 
-A subscriber can send PUBLISH_NAMESPACE_CANCEL to revoke acceptance of an
-PUBLISH_NAMESPACE, for example due to expiration of authorization
-credentials. The message enables the publisher to PUBLISH_NAMESPACE again with
-refreshed authorization, or discard associated state. After receiving an
-PUBLISH_NAMESPACE_CANCEL, the publisher does not send PUBLISH_NAMESPACE_DONE.
+A subscriber can cancel the request to revoke acceptance of a
```
```suggestion
A subscriber can cancel the request (see {{request-cancellation}}) to revoke acceptance of a
```

draft-ietf-moq-transport.md
```diff
@@ -1783,21 +1791,36 @@ Payload, which is defined by each message type.  If the length does not match
...
+the Request ID of a SUBSCRIBE.
+
+The client generates even numbered Request IDs, starting at 0, and the
+server generates odd numbered Request IDs, starting at 1.  Each
+endpoint increments its Request ID by 2 for each new request.
+
+Each SUBSCRIBE, PUBLISH, FETCH, SUBSCRIBE_NAMESPACE, PUBLISH_NAMESPACE,
+REQUEST_UPDATE, and TRACK_STATUS message consumes a Request ID. Only
+request messages include a Request ID; response messages do not, since
+they are sent on the same bidirectional stream.
```
```suggestion
they are sent on the same bidirectional stream as the request.
```

draft-ietf-moq-transport.md
```diff
@@ -1783,21 +1791,36 @@ Payload, which is defined by each message type.  If the length does not match
...
+
+If an endpoint receives a Request ID where the least significant bit is
+incorrect for the sender, or a duplicate Request ID, it MUST close the
+session with `INVALID_REQUEST_ID`.
+
+## Required Request ID {#required-request-id}
+
+Every request message includes a Required Request ID Delta field, which is
+multiplied by two and subtracted from Request ID to determine the
+Required Request ID. The receiver MUST deliver the referenced request to the
```
suhasHere: I am totally fine for this to be not delta encoded. It adds no value, i feel. Am i missing something ?

draft-ietf-moq-transport.md
```diff
@@ -2537,14 +2495,13 @@ time to receive objects in the future.
 
 ## SUBSCRIBE_OK {#message-subscribe-ok}
 
-A publisher sends a SUBSCRIBE_OK control message for successful
-subscriptions.
+A publisher sends a SUBSCRIBE_OK as the first response message on the
```
suhasHere, line 2596: not sure the refeernce to "First response" .. do we have any other response ?
Also in earlier sections we do say that Subscribe request can be cancelled in Pending state ?

draft-ietf-moq-transport.md
ianswett: This is existing text, I minorly updated it to remove references to UNSUBSCRIBE and FETCH_CANCEL.

draft-ietf-moq-transport.md
ianswett: Yes, I think so.  Will update.

draft-ietf-moq-transport.md
```diff
@@ -862,24 +863,42 @@ can be used.
...
+Once a request stream has been opened, the request MAY be cancelled by either
+endpoint. Senders cancel requests if the response is no longer of interest;
+Receivers cancel requests if they are unable to or choose not to respond.
+
+Implementations SHOULD cancel requests by abruptly terminating any directions of
+a stream that are still open using RESET_STREAM / RESET_STREAM_AT or
+STOP_SENDING.
+
+When the Publisher rejects a request without performing any application processing,
+The PUBLISHER SHOULD send a REQUEST_ERROR and FIN the stream.
```
```suggestion
When an endpoint rejects a request without performing any application processing,
it SHOULD send a REQUEST_ERROR and FIN the stream.
```

draft-ietf-moq-transport.md
```suggestion
RESET_STREAM for the FETCH data stream. If the data stream is already open,
the subscriber wishing to cancel the FETCH MAY send STOP_SENDING for the
data stream as well as the the bidi request stream. It MUST send STOP_SENDING
for the bidi request stream.
```

draft-ietf-moq-transport.md
```diff
@@ -1281,13 +1300,15 @@ groups. A publisher that does will begin the next group as soon as practical.
...
+((see {{request-cancellation}})), receives REQUEST_ERROR, or receives a FIN or
+RESET_STREAM for the FETCH data stream. If the data stream is already open,
+the subscriber wishing to cancel the FETCH MAY send STOP_SENDING for the
+data stream as well as the the bidi request stream. It MUST send STOP_SENDING
+for the bidi request stream.
 
 The Publisher can destroy fetch state as soon as it has received a
-FETCH_CANCEL. It MUST reset any open streams associated with the FETCH. It can
+STOP_SENDING. It MUST reset any open streams associated with the FETCH. It can
 also destroy state after closing the FETCH data stream.
```
```suggestion
STOP_SENDING. It MUST reset the bidi request stream and unidirectional
data stream associated with the FETCH. It can also destroy state after closing
the FETCH data stream.
```

draft-ietf-moq-transport.md
ianswett: Would 'first message on the response side of the bidi stream..' be better?

draft-ietf-moq-transport.md
ianswett: I'm not sure we need a special error for this anymore, but I updated the text.

draft-ietf-moq-transport.md
```diff
@@ -905,20 +924,19 @@ PROTOCOL_VIOLATION (0x3):
...
 INVALID_REQUEST_ID (0x4):
-: The session was closed because the endpoint used a Request ID that was
-  smaller than or equal to a previously received request ID, or the least-
-  significant bit of the request ID was incorrect for the endpoint.
+: The endpoint received a Request ID with an incorrect least significant
+  bit for the sender, or a duplicate Request ID. See {{request-id}}.
+
+INVALID_REQUIRED_REQUEST_ID (0x7):
+: The endpoint received a Required Request ID that does not have the same
+  parity as the Request ID, or is less than 0. See {{required-request-id}}.
```
```suggestion
: The endpoint received a Required Request ID Delta that results
in an invalid Request ID. See {{required-request-id}}.
```

vasilvv:
Read the entire PR.  It overall looks good, except for the request ID mechanism, which seems really undercooked (there definitely needs to be at least more thought put into the timeouts, etc).  Also see my counterproposal above.

draft-ietf-moq-transport.md
```diff
@@ -3630,8 +3505,11 @@ Header field values.
 
 Streams aside from the control stream MAY be canceled due to congestion
 or other reasons by either the publisher or subscriber. Early termination of a
-stream does not affect the MoQ application state, and therefore has no
-effect on outstanding subscriptions.
+unidirectional stream does not affect the MoQ application state, and therefore has
+no effect on outstanding subscriptions. Termination of a bidi request stream
+terminates the Subscription, Fetch, Track Status, Publish Namespace, or Subscribe Namespace
+request. When possible, Publishers SHOULD send a PUBLISH_DONE when terminating a
+subscription instead of abruptly terminating the stream.
```
```suggestion
subscription instead of abruptly terminating the associated control stream.
```

draft-ietf-moq-transport.md
```diff
@@ -1825,21 +1836,36 @@ Payload, which is defined by each message type.  If the length does not match
...
+
+Every request message includes a Required Request ID Delta field, which is
+multiplied by two and subtracted from Request ID to determine the
+Required Request ID. The receiver MUST deliver the referenced request to the
+application before delivering this request. Two times the Required Request ID
+Delta MUST NOT be larger than Request ID. If an endpoint receives a delta that
+is too large for the Request ID, it MUST close the session with
+`INVALID_REQUIRED_REQUEST_ID`. A Required Request ID Delta value of 0 indicates
+there is no dependency. If the required request does not arrive, the receiver
+will time out the dependent request.
```
suhasHere, line 1872:
```suggestion

Every request message includes a Required Request ID Delta
field that specifies a dependency on a prior request. The 
Required Request ID is computed as:
~~~
    Required Request ID = Request ID - (2 × Required Request ID
  Delta)
~~~
  
  A Required Request ID Delta of 0 indicates no dependency. When
   a dependency exists, the receiver MUST deliver the referenced
   request to the application before delivering the dependent
  request. If the required request does not arrive, the receiver
   will time out the dependent request.

  The delta is scaled by two because request IDs from each
  endpoint use alternating parity (odd or even), so valid
  dependencies always differ by a multiple of two. An endpoint
  MUST close the session with INVALID_REQUIRED_REQUEST_ID if it
  receives a delta where 2 × Required Request ID Delta exceeds
  the Request ID.
```

Issue #606, opened 2024-11-06
 timeout properties got broken 


fluffy:
TTL got moved into delivery timeout in subscribe OK but that lost the property to set a different timeout for different sub groups.  

It is not clear where a max cache parameter can be in the protocol. 

It is not clear interaction of max cache and delivery timeout in subscribe OK are both needed.

afrind:
I renamed the issue to reflect the primary issue with the current draft.

> It is not clear where a max cache parameter can be in the protocol.

This is true, let's file a separate issue. My understanding is that it's useful in both SUBSCRIBE_OK and FETCH_OK - does that seem right?

> It is not clear interaction of max cache and delivery timeout in subscribe OK are both needed.

My understanding is:

DELIVERY_TIMEOUT only applies to SUBSCRIBE (not FETCH) -- how long should a relay attempt to deliver an object before dropping/resetting.
MAX_CACHE_DURATION tells a cache the max time to keep an object in cache (to serve via FETCH) before expiring it.  A subsequent fetch after expiry would necessitate re-fetching from origin.

If that's right then having both in SUBSCRIBE_OK seems valid.  If that's not right, let's file a separate issue for this too.

martinduke:
I don't understand the use case for different delivery timeouts. Surely the delivery timeout relates to when the media will be consumed, which is independent of priority?

ianswett:
This was filed a while ago, but I'm very unclear what the use case is for different delivery timeouts per subgroup, particularly since delivery timeout is typically expected to be subscriber chosen.

Getting the higher temporal scalability layer after playing the content isn't useful and if the base layer is higher priority, you'd always deliver that first.  There might be optimizations around implementing the intent of the client to never transmit the higher temporal scalability layer if it was clear it wouldn't get through, but I think that's a different issue.

afrind:
Discussed in Stockholm - @vasilvv is planning to make an alternate proposal for delivery timeout in #667.   We will park this until that is resolved, but if we keep the current design we will allow an optional per-subgroup delivery timeout for the publisher (the subscriber is still restricted to one per Track).

afrind:
There's a relatively simple fix now - allow DT extension on the first object of a SG to override track level extension.

PR #1390, opened 2025-12-09
Move track properties to extensions, scope parameters

afrind:
1. Move the following to Track Extensions/Headers/Properties, so it's clear they are carried end-to-end:

DYNAMIC_GROUPS
MAX_CACHE_DURATION
PUBLISHER_PRIORITY
PUBLISHER_GROUP_ORDER_PREFERENCE (split from GROUP_ORDER param) DELIVERY_TIMEOUT (split from DELIVERY_TIMEOUT param)

2. Reject unknown *Message* parameters -- all must be negotiated.  Unknown Setup Params are of course OK.

3. All *parameters* are explicitly scoped to a single hop.

I didn't say anything about updates. It's easy to say "new values overwrite old ones", but there's no message to update a Track extension.

Replaces #1285
Fixes: #1278

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -332,6 +332,35 @@ Reason Phrase {
   such as language tags, that would aid comprehension by any entity other than
   the one that created the text.
 
+## Representing Namespace and Track Names
+
+There is often a need to render namespace tuples and track names for
+purposes such as logging, representing track filenames, or use in
+certain authorization verification schemes. The namespace and track name
+are binary, so they need to be converted to a safe form.
+
+The following format is RECOMMENDED:
+
+* Each of the namespace tuples are rendered in order with a hyphen (-)
+  between them followed by the track name with a double hyphen (--)
+  between the last namespace and track name.
+
+* Bytes in the range a-z, A-Z, 0-9 as well as _ (0x5f) are output as is,
+  while all other bytes are encoded as a period (.) symbol followed by
+  exactly two lower case hex digits.
+
+The goal of this format is to have a format that is both filename and
+URL safe. It allows many common names to be rendered in an easily human
+readable form while still supporting binary values.
+
+Example:
+
+~~~
+example.2enet-team2-project_x--report
+  Namespace tuples: (example.net, team2, project_x)
+  Track name: report
+~~~
+
 # Object Data Model {#model}
 
 MOQT has a hierarchical data model, comprised of tracks which contain
@@ -401,15 +430,15 @@ using at least as many streams as there are Subgroups,
 typically with a one-to-one mapping between Subgroups and streams.
 
 When an Object's forwarding preference (see {{object-properties}}) is
-"Datagram", it is not sent in Subgroups and the
-description in the remainder of this section does not apply.
+"Datagram", it is not sent in Subgroups, does not belong to a Subgroup in any
+way, and the description in the remainder of this section does not apply.
 
 Streams offer in-order reliable delivery and the ability to cancel sending and
 retransmission of data. Furthermore, many QUIC and WebTransport implementations
 offer the ability to control the relative scheduling priority of pending stream
 data.
 
-Every object within a Group belongs to exactly one Subgroup.
+Every Object within a Group belongs to exactly one Subgroup or Datagram.
 
 When Objects are sent in a subscription (see {{subscriptions}}),  Objects
 from two subgroups MUST NOT be sent on the same stream, and Objects from the
@@ -507,18 +536,24 @@ Track Namespace Field {
 * Track Namespace Field Value: A sequence of bytes that forms a Track Namespace
   Field.
 
+Each Track Namespace Field Value MUST contain at least one byte. If an endpoint
+receives a Track Namespace Field with a Track Namespace Field Length of 0, it
+MUST close the session with a `PROTOCOL_VIOLATION`.
+
 The structured nature of Track Namespace allows relays and applications to
 manipulate prefixes of a namespace. If an endpoint receives a Track Namespace
 consisting of 0 or greater than 32 Track Namespace Fields, it MUST close the
 session with a `PROTOCOL_VIOLATION`.
 
-Track Name is a sequence of bytes that identifies an individual track within the
-namespace.
+Track Name is a sequence of bytes, possibly empty, that identifies an individual
+track within the namespace.
 
-The maximum total length of a Full Track Name is 4,096 bytes. The length of a Full
-Track Name is computed as the sum of the Track Namespace Field Length fields
-and the Track Name Length field.  If an endpoint receives a Full Track Name
-exceeding this length, it MUST close the session with a `PROTOCOL_VIOLATION`.
+The maximum total length of a Full Track Name is 4,096 bytes. The length of a
+Full Track Name is computed as the sum of the Track Namespace Field Length
+fields and the Track Name Length field. The length of a Track Namespace is the
+sum of the Track Namespace Field Length fields. If an endpoint receives a Track
+Namespace or a Full Track Name exceeding 4,096 bytes, it MUST close the session
+with a `PROTOCOL_VIOLATION`.
 
 In this specification, both the Track Namespace Fields and the Track Name
 are not constrained to a specific encoding. They carry a sequence of bytes and
@@ -649,7 +684,7 @@ such as TCP, and applications using MoQ might need to fallback to
 another protocol when QUIC or WebTransport aren't available.
 
 MOQT uses ALPN in QUIC and "WT-Available-Protocols" in WebTransport
-({{WebTransport, Section 3.4}}) to perform version negotiation.
+({{WebTransport, Section 3.3}}) to perform version negotiation.
 
 \[\[RFC editor: please remove the remainder of this section before publication.]]
 
@@ -1311,6 +1346,11 @@ to different subscriptions or FETCH responses, but have the same subscriber and
 publisher priority.  The ordering in those cases is implementation-defined,
 though the expectation is that all subscriptions will be able to send some data.
 
+A publisher might not utilize the entire available congestion window,
+session flow control, or all available streams for lower
+priority Objects if it expects higher priority Objects will be available to send
+in the near future or it wants to reserve some bandwidth for control messages.
+
 Given the critical nature of control messages and their relatively
 small size, the control stream SHOULD be prioritized higher than all
 subscribed Objects.
@@ -1451,7 +1491,7 @@ all possible Objects matching the filter have been delivered to the subscriber.
 Since tracking this can be prohibitively expensive, Relays are not required or
 expected to do so.
 
-### Graceful Subscriber Relay Switchover
+### Graceful Subscriber Relay Switchover {#graceful-subscriber-switchover}
 
 This section describes behavior a subscriber MAY implement
 to allow for a better user experience when a relay sends a GOAWAY.
@@ -1504,7 +1544,9 @@ A SUBSCRIBE message with namespace=(foo, bar) and name=x will match sessions
 that sent PUBLISH_NAMESPACE messages with namespace=(foo) or namespace=(foo,
 bar).  It will not match a session with namespace=(foobar).
 
-Relays MUST forward SUBSCRIBE messages to all matching publishers and
+Relays MUST send SUBSCRIBE messages to all matching publishers. This includes
+matching both Established subscriptions on the Full Track Name and Namespace
+Prefix Matching against published Namespaces.  Relays MUST forward
 PUBLISH_NAMESPACE or PUBLISH messages to all matching subscribers.
 
 When a Relay needs to make an upstream FETCH request, it determines the
@@ -1537,7 +1579,7 @@ the Relay MUST NOT propagate that message or Object to another Session, because
 it would enable a single Session error to force an unrelated Session, which
 might be handling other subscriptions, to be closed.
 
-### Graceful Publisher Network Switchover
+### Graceful Publisher Network Switchover {#graceful-publisher-switchover}
 
 This section describes a behavior that a publisher MAY
 choose to implement to allow for a better user experience when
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1684,34 +1684,44 @@ sequence or exceeds the received MAX_REQUEST_ID, it MUST close the session with
...
 
-Setup message parameters use a namespace that is constant across all MOQT
+Setup Parameters use a namespace that is constant across all MOQT
 versions. All other messages use a version-specific namespace.
 For example, the integer '1' can refer to different parameters for Setup
 messages and for all other message types. SETUP message parameter types
 are defined in {{setup-params}}. Version-specific parameter types are defined
 in {{version-specific-params}}.
 
+The Parameters in SUBSCRIBE, PUBLISH_OK and FETCH MUST NOT cause the publisher
```
ianswett: I think it's generally true that Parameters shouldn't cause the Publisher to modify Objects, though I agree that these are the most likely culprits.  I'd be inclined to drop the specific messages?

```suggestion
Message Parameters in SUBSCRIBE, PUBLISH_OK and FETCH MUST NOT cause the publisher
```

draft-ietf-moq-transport.md
```diff
@@ -1835,8 +1845,7 @@ aliases.
 #### DELIVERY TIMEOUT Parameter {#delivery-timeout}
 
 The DELIVERY TIMEOUT parameter (Parameter Type 0x02) MAY appear in a
-REQUEST_OK (in response to TRACK_STATUS), PUBLISH, PUBLISH_OK,
-SUBSCRIBE, SUBSCRIBE_OK, or SUBSCRIBE_UDPATE message.
+PUBLISH_OK, SUBSCRIBE, or SUBSCRIBE_UDPATE message.
```
ianswett: SUBSCRIBE_UPDATE was replaced by REQUEST_UPDATE, correct?  Also, UPDATE is spelled wrong

draft-ietf-moq-transport.md
```diff
@@ -1919,8 +1899,7 @@ the value 128.
 #### GROUP ORDER Parameter {#group-order}
 
 The GROUP_ORDER parameter (Parameter Type 0x22) MAY appear in a SUBSCRIBE,
-SUBSCRIBE_OK, REQUEST_OK (in response to TRACK_STATUS), PUBLISH,
-PUBLISH_OK or FETCH.
+or FETCH.
```
ianswett: Not PUBLISH_OK?

draft-ietf-moq-transport.md
```diff
@@ -3755,6 +3719,71 @@ SUBGROUP_HEADER {
...
+A subscription has Publisher Priorty 128 if this extension is omitted.
+
+#### PUBLISHER GROUP ORDER PREFERENCE {#group-order-pref}
+
+The PUBLISHER_GROUP_ORDER_PREFERENCE extension (Extension Header Type 0x22) is a
+Track Extension.
+
+It
+is an enum indicating the publisher's preference for prioritizing Objects from
+different groups within the
```
ianswett: Reflow

```suggestion
It is an enum indicating the publisher's preference for prioritizing Objects
from different groups within the
```

draft-ietf-moq-transport.md
afrind: This text was just moved.  Can we follow up

draft-ietf-moq-transport.md
```suggestion
PUBLISH_OK, SUBSCRIBE, or REQUEST_UPDATE message.
```

draft-ietf-moq-transport.md
```suggestion
PUBLISH_OK, or FETCH.
```

draft-ietf-moq-transport.md
```diff
@@ -3755,6 +3719,70 @@ SUBGROUP_HEADER {
 The following Extension Headers are defined in MOQT. Each Extension Header
 specifies whether it can be used with Tracks, Objects, or both.
 
+
+#### DELIVERY TIMEOUT {#delivery-timeout-ext}
+
+The DELIVERY TIMEOUT extension (Extension Header Type 0x02) is a Track
```
ianswett, line 3806: Observation: Now that it's an extension header, we could allow it on Subgroups or Datagrams if we really wanted to?  I think that was requested in the past?

afrind: That's a good idea, see #606, though setting a different DT per object in a subgroup seems excessive but annoying to prevent.

draft-ietf-moq-transport.md
```diff
@@ -1684,34 +1684,44 @@ sequence or exceeds the received MAX_REQUEST_ID, it MUST close the session with
 
 ## Parameters {#params}
 
-Some messages include a Parameters field that encodes optional message
-elements.
+Some messages include a Parameters field that encodes optional message elements.
+Parameters in the CLIENT_SETUP and SERVER_SETUP messages are called Setup
+Parameters.  Parameters in other control messages are Message Parameters.
+Receivers ignore unrecognized Setup Parameters.  All Message Parameters MUST be
```
suhasHere, line 1732: I feel ignoring is not what we need here. May be what we need is, Setup parameters MUST be negotiated by peers else return protocol violation. ( please note, this is not tied to a MOQT version, so we get the flexibility)

ianswett: That limits one's ability to negotiate extensions.  Setup parameters is where you would negotiate those extensions, so it's the place where unknown parameters need to be allowed.

draft-ietf-moq-transport.md
```diff
@@ -3755,6 +3719,70 @@ SUBGROUP_HEADER {
...
+subscription or fetch after the specified number of milliseconds has elapsed
+since the beginning of the Object was received.  This means Objects earlier in a
+multi-object stream will expire earlier than Objects later in the stream. Once
+Objects have expired from cache, their state becomes unknown, and a relay that
+handles a downstream request that includes those Objects re-requests them.
+
+If the MAX_CACHE_DURATION extension is not sent by the publisher, the Objects
+can be cached until implementation constraints cause them to be evicted.
+
+#### PUBLISHER PRIORITY {#publisher-priority}
```
vasilvv: Should this be DEFAULT PUBLISHER PRIORITY?

draft-ietf-moq-transport.md
```diff
@@ -3755,6 +3719,70 @@ SUBGROUP_HEADER {
...
+The PUBLISHER PRIORITY extension (Extension Header Type 0x0E) is a Track Extension
+that specifies the priority of
+a subscription relative to other subscriptions in the same session.  The value
+is from 0 to 255 and lower numbers get higher priority.  See
+{{priorities}}. Priorities above 255 are invalid. Subgroups and Datagrams for this
+subscription inherit this priority, unless they specifically override it.
+
+A subscription has Publisher Priorty 128 if this extension is omitted.
+
+#### PUBLISHER GROUP ORDER PREFERENCE {#group-order-pref}
```
vasilvv: Similarly, DEFAULT PUBLISHER GROUP ORDER?

draft-ietf-moq-transport.md
afrind: I don't see how you will make extensibility work if you don't allow unrecognized setup params in v1 -- how else to you attempt to negotiate a v1 extension with a peer that might not support it?

draft-ietf-moq-transport.md
```diff
@@ -3755,6 +3719,70 @@ SUBGROUP_HEADER {
...
+Objects have expired from cache, their state becomes unknown, and a relay that
+handles a downstream request that includes those Objects re-requests them.
+
+If the MAX_CACHE_DURATION extension is not sent by the publisher, the Objects
+can be cached until implementation constraints cause them to be evicted.
+
+#### PUBLISHER PRIORITY {#publisher-priority}
+
+The PUBLISHER PRIORITY extension (Extension Header Type 0x0E) is a Track Extension
+that specifies the priority of
```
```suggestion
#### DEFAULT PUBLISHER PRIORITY {#publisher-priority}

The DEFAULT PUBLISHER PRIORITY extension (Extension Header Type 0x0E) is a Track
Extension that specifies the priority of
```

draft-ietf-moq-transport.md
```diff
@@ -3755,6 +3719,70 @@ SUBGROUP_HEADER {
...
+is from 0 to 255 and lower numbers get higher priority.  See
+{{priorities}}. Priorities above 255 are invalid. Subgroups and Datagrams for this
+subscription inherit this priority, unless they specifically override it.
+
+A subscription has Publisher Priorty 128 if this extension is omitted.
+
+#### PUBLISHER GROUP ORDER PREFERENCE {#group-order-pref}
+
+The PUBLISHER_GROUP_ORDER_PREFERENCE extension (Extension Header Type 0x22) is a
+Track Extension.
```
```suggestion
#### DEFAULT PUBLISHER GROUP ORDER {#group-order-pref}

The DEFAULT_PUBLISHER_GROUP_ORDER extension (Extension Header Type 0x22) is a
Track Extension.
```

draft-ietf-moq-transport.md
afrind: @suhasHere This is not a change from the current document, which says:

Receivers ignore unrecognized parameters.

What I changed here is left setup alone, and made message parameters more strict.  We can talk about different extension negotiation parameters, but I think this PR is ready on its own.

suhasHere: Yes, existing document lacks extension mechanism and this PR and related discussions was to address those concerns. 

By unrecognizable and ignorable, Setup parameter must be recognized by the Peer , but it need not have to be in a given protocol version.  I am feeling bit uncomfortable to allow to blindly ignore setup parameter by the peer ( in this case relay)


Just thinking loud, say, i have a setup param, called disable_rtx_use_rs_fec_only and the publisher ignored it altogether.  Should the receiver figure out the publisher doesn't support this behavior by observing data plane and seeing RTX packets or knows the peer understood by observing rs_fec packets or consider peer doesn't understand by seeing something like fec but which doesn't look like rs_fec but fails recovety?

May be I am thinking wrong here, but love to understand what am i missing ?

fluffy:
I have asked for agenda time to discuss this before we make changes. I still feel very strongly that trying to force this into the draft before we have agreement on what we are trying to do is making this WG far more antagonistic than it needs to be. 

Lets take an example, what we had discussed was a way in set up to negotiate  that the  client and server both support NewFeatureX but that data that is passed when you use NewFeatureX might be one or more parameters that could be big small or just flags. 

The terminology in this PR confuse the messages we use to pass the data for NewFeatureX with the way we pass the data that is needed to use NewFeatureX.

afrind:
@martinduke @gloinul Please take a look and provide direction to the editors.

afrind:
I filed #1407 to track discussion about changing the extension negotiation design.

draft-ietf-moq-transport.md
suhasHere: @ianswett  thoughts on the above ?

ianswett: The server/publisher can indicate in its SERVER_SETUP that it supports the disable_rtx_use_rs_fec_only extension if knowing whether its supported is important.  MP-QUIC does this by requiring both endpoints to send the initial_max_path_id transport param: https://www.ietf.org/archive/id/draft-ietf-quic-multipath-18.html#section-2

Alternatively, QUIC ACK_FREQUENCY is an optimization, and only uses one Transport Param to unilaterally advertise support: https://www.ietf.org/archive/id/draft-ietf-quic-ack-frequency-13.html#section-3

Each extension needs to decide how important it is to know if the peer supports it.  Commonly it ends up falling out of the document's larger requirements.  For example, MP-QUIC requires both endpoints to say that a given Path ID can be used/opened before opening one, so it makes sense to set that initial value in the handshake.

Even if you control all your infrastructure, it's difficult to ensure 100% of servers support something before a client requests it.  Equally importantly, having an untested feature in 100% of servers leaves you with a potential 0-day vulnerability if the new code has a crash bug or other issue (don't ask me how I know).

suhasHere: Agreed with the reasoning. May be what we need is 
Sender lists all the parameters that needs peer to respond and peer responds with ones it can support ( in server setup). This will give a clear indication of the peer's ability in supporting features required by those parameters. 

Completely ignoring the support is very ambiguous though.

would that work ?

martinduke:
> @martinduke @gloinul Please take a look and provide direction to the editors.

I don't have a record of an agenda request from Cullen. I've emailed him to ask for one.

ianswett:
> I have asked for agenda time to discuss this before we make changes. I still feel very strongly that trying to force this into the draft before we have agreement on what we are trying to do is making this WG far more antagonistic than it needs to be.
> 
> Lets take an example, what we had discussed was a way in set up to negotiate that the client and server both support NewFeatureX but that data that is passed when you use NewFeatureX might be one or more parameters that could be big small or just flags.
> 
> The terminology in this PR confuse the messages we use to pass the data for NewFeatureX with the way we pass the data that is needed to use NewFeatureX.

Are you saying that using the same term, 'Parameter' for both SETUP messages, where parameters are not required to be understood, and Message Parameters, where Parameters are required to be understood, is making the draft confusing?  if so, I can see splitting these into two registries and using different names if that would be clearer?

suhasHere (approved):
+1 given that we need to address setup parameter negotiation issue (#1407)

draft-ietf-moq-transport.md
```diff
@@ -1910,7 +1918,7 @@ reset the underlying transport stream (see {{closing-subgroup-streams}}) and
 SHOULD NOT attempt to open a new stream to deliver additional Objects in that
 Subgroup.
 
-When sent by a subscriber, this parameter is intended to be specific to a
+This parameter is intended to be specific to a
 subscription, so it SHOULD NOT be forwarded upstream by a relay that intends
 to serve multiple subscriptions for the same track.
```
sharmafb, line 1923: Isn't this redundant now that we've explicitly specified "Message Parameters are always intended for the peer endpoint only and are not forwarded by Relays"?

afrind: I'm ok living with the redundancy at least for now.

PR #1393, opened 2025-12-12
Allow the '*' or the empty namespace in SUBSCRIBE_NAMESPACE

### LINKED ISSUE #1346: SUBSCRIBE_NAMESPACE * is invalid
I want to receive all namespaces published by the peer, however the draft specifies the number of namespace parts MUST be larger than 0.

ianswett:
Fixes #1346 

This doesn't substantially change the properties of the system vs requiring one.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2345,10 +2345,6 @@ INVALID_JOINING_REQUEST_ID(0x32):
 : In response to a Joining FETCH, the referenced Request ID is not an
 `Established` Subscription.
 
-UNKNOWN_STATUS_IN_RANGE(0x33):
-: In response to a FETCH, the requested range contains an object with unknown
-status.
-
 ## SUBSCRIBE {#message-subscribe-req}
 
 A subscription causes the publisher to send newly published objects for a track.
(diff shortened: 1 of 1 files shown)


afrind (approved):

draft-ietf-moq-transport.md
```diff
@@ -3088,13 +3088,13 @@ SUBSCRIBE_NAMESPACE Message {
 * Request ID: See {{request-id}}.
 
 * Track Namespace Prefix: A Track Namespace structure as described in
-  {{track-name}} with between 1 and 32 Track Namespace Fields.  This prefix is
+  {{track-name}} with between 0 and 32 Track Namespace Fields.  This prefix is
```
suhasHere, line 3089: I am not sure we allow fully empty namespaces.

afrind: We don't right now, see linked issue.  I tend to agree with @ianswett though -- the abuse risk of subscribe namespace isn't materially different at 0 vs 1 tuple minimum.  The risk needs to be addressed by auth, rather than a protocol restriction

suhasHere: its more of a data model question than security risk and what should a relay do if there is a zero tuples in track namespace ( say if auth allows it ) ?

afrind: All namespaces/tracks at the relay match

fluffy (approved):
I think this is a good idea. There will be use case like debugging where it could make since for a colocated server to subscribe to *.

sharmafb (approved):

sandarsh (approved):

Issue #1380, opened 2025-11-26
How does a caching relay handle different unknown extensions?


afrind:
The text currently says:

> Object Header Extensions can be added, removed or updated, subject to the constraints of the specific header extension.

However, if a cache doesn't support the extension, then it has no idea what the constraints are and we need to specify a default behavior.

The question also applies to Track level extensions in #1374.

I think the only reasonable behavior might be append and resolve duplicates:

A=B + A=C = {A=B, A=C}
A=B + empty = {A=B}
A=B + A=B = {A=B}

afrind:
Discussed in 12/1 Interim:

The strawman proposal above is problematic.

WG would like:

1. Supported extensions are processed according to their own rules
2. Unsupported extensions are taken wholesale from a single publisher (no attempts to merge)
3. Caches SHOULD serve the unknown extensions that arrived most recently (this allows publishers to "fix" corruption).  SHOULD

PR #1395, opened 2025-12-13
Clarify handling of unknown extensions

afrind:
Per the issue:

* Supported extensions are processed according to their own rules
* Unsupported extensions are taken wholesale from a single publisher (no attempts to merge)
* Caches SHOULD serve the unknown extensions that arrived most recently (this allows publishers to "fix" corruption). SHOULD

Fixes: #1380

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -618,10 +618,10 @@ concerning processing, modification, caching and forwarding.
 
 If unsupported by the relay, Extension Headers MUST NOT be modified, MUST be
 cached as part of the Track or Object and MUST be forwarded by relays.  If a
-Track or Object arrives with a different set of unknown extensions, the most
-recent set SHOULD replace any cached values, removing any unknown values not
-present in the new set.  Relays MUST NOT attempt to merge sets of unknown
-extensions received in different messages.
+Track or Object arrives with a different set of unknown extensions than
+previously cached, the most recent set SHOULD replace any cached values,
+removing any unknown values not present in the new set.  Relays MUST NOT attempt
+to merge sets of unknown extensions received in different messages.
 
 If supported by the relay and subject to the processing rules specified in the
 definition of the extension, Extension Headers MAY be modified, added, removed,


draft-ietf-moq-transport.md
```diff
@@ -617,7 +617,10 @@ specifications define the type and value of the header, along with any rules
 concerning processing, modification, caching and forwarding.
 
 If unsupported by the relay, Extension Headers MUST NOT be modified, MUST be
-cached as part of the Track or Object and MUST be forwarded by relays.
+cached as part of the Track or Object and MUST be forwarded by relays.  If a
+Track or Object arrives with a different set of unknown extensions, the most
+recent set SHOULD update any cached values, removing any unknown values not
```
ianswett: I would think it would be 'replace', not 'update'?  I'm not sure it's safe to assume that merging even known headers from two different sets is wise?

```suggestion
recent set SHOULD replace any cached values, removing any unknown values not
```

draft-ietf-moq-transport.md
```diff
@@ -617,7 +617,10 @@ specifications define the type and value of the header, along with any rules
 concerning processing, modification, caching and forwarding.
 
 If unsupported by the relay, Extension Headers MUST NOT be modified, MUST be
-cached as part of the Track or Object and MUST be forwarded by relays.
+cached as part of the Track or Object and MUST be forwarded by relays.  If a
+Track or Object arrives with a different set of unknown extensions, the most
+recent set SHOULD replace any cached values, removing any unknown values not
+present in the new set.
```
michalhosna: I think the current wording is not restrictive enough to cover the non-merging requirement.

From #1380
> WG would like
> 2. Unsupported extensions are taken wholesale from a single publisher (no attempts to merge)

I tried to write a concrete suggestion, but now I am not sure if we want to only disallow merging of the extension values or also forbid merging/union of the extension set.

Forbidding that would make it possible to drop whole extensions from the cache and make sure that the set is always sensible, but the behavior of known extensions may start to get weird.

michalhosna (changes requested):

draft-ietf-moq-transport.md
afrind: Can you clarify what you mean by merging extension values or merging/union of the extension set?  I think the behavior we want is the following:

If the cached track/object has a set of supported extensions X and a set of unknown extensions Y, and we get another extension block for that track/object with supported extensions X' and unknown extensions Y'.  Then the cache SHOULD be updated to contain exactly:

merge_supported(X, X'), Y'

Anything in Y not in Y' is deleted.

michalhosna: "Merging values" (Appending): `{A=1} + {A=2} = {A=1, A=2}`. We don't want this. 

Merge/union of the extension set `{A=1, B=1} + {A=2, C=2}`
  1. Allow union/merge of the set = `{A=2, B=1, C=2}` 
  2. Disallow union/merge of the set = `{A=1, B=1}` or `{A=2, C=2}`. Either one is valid, with preference for the latter. 
  
The text recommends the 2nd option but is missing any hard requirements on the merge handling. I think we need a hard requirement (MUST) that forbids values merging/appending. 

I am not sure if we want to forbid "set merge/union" in the "MUST" part or just discourage it in the "SHOULD" part.

```suggestion
present in the new set.  Relays MUST NOT attempt to merge sets of unknown
extensions received in different messages.
```

ianswett (approved):

michalhosna (approved):

draft-ietf-moq-transport.md
```diff
@@ -617,7 +617,11 @@ specifications define the type and value of the header, along with any rules
 concerning processing, modification, caching and forwarding.
 
 If unsupported by the relay, Extension Headers MUST NOT be modified, MUST be
-cached as part of the Track or Object and MUST be forwarded by relays.
+cached as part of the Track or Object and MUST be forwarded by relays.  If a
+Track or Object arrives with a different set of unknown extensions, the most
+recent set SHOULD replace any cached values, removing any unknown values not
+present in the new set.  Relays MUST NOT attempt to merge sets of unknown
```
suhasHere: this is performance intensive to do per object. For objects, the set of extensions is stored per object isn't it ? I wonder, if we just store it as-is ?

afrind: I think that's what the text is saying - you replace all cached extensions E with new extensions E'.  This implicitly deletes unknown extensions in E not present in E' so shouldn't be performance intensive.

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -617,7 +617,11 @@ specifications define the type and value of the header, along with any rules
...
+extensions received in different messages.
```
```suggestion
Track or Object arrives with a different set of unknown extensions than
previously cached, the most recent set SHOULD replace any cached values,
removing any unknown values not present in the new set.  Relays MUST NOT attempt
to merge sets of unknown extensions received in different messages.
```

Issue #1349, opened 2025-11-04
Does a single subscribe match both a PUBLISH and PUBLISH_NAMESPACE from different publishers?


afrind:
Consider this sequence:

P1->R: PUBLISH a/b
P2->R: PUBLISH_NAMESPACE a
S->R: SUBSCRIBE a/b

Does the relay send a SUBSCRIBE to P2?

afrind:
Discussed in 12/1 Interim:

Suhas said "yes" and no one else expressed an opinion.  It seems relatively consistent with our multipublisher design.

PR #1397, opened 2025-12-13
Relays match SUBSCRIBE to both Tracks and Namespaces

afrind:
Fixes: #1349

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2347,10 +2347,6 @@ INVALID_JOINING_REQUEST_ID(0x32):
 : In response to a Joining FETCH, the referenced Request ID is not an
 `Established` Subscription.
 
-UNKNOWN_STATUS_IN_RANGE(0x33):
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1504,7 +1504,9 @@ A SUBSCRIBE message with namespace=(foo, bar) and name=x will match sessions
 that sent PUBLISH_NAMESPACE messages with namespace=(foo) or namespace=(foo,
 bar).  It will not match a session with namespace=(foobar).
 
-Relays MUST forward SUBSCRIBE messages to all matching publishers and
+Relays MUST forward SUBSCRIBE messages to all matching publishers. This includes
```
ianswett: I feel like forward is the wrong word, since it sort of implies that the SUBSCRIBE is just sent up the chain with no modification.

```suggestion
Relays MUST send SUBSCRIBE messages to all matching publishers. This includes
```

suhasHere (approved):

Issue #1385, opened 2025-12-07
Section 10.3.1, Table 4: The table has duplicate rows with different "Type" values.


kirki58:
In the "Data streams and Datagrams > Datagrams > Object Datagram" Section of the draft document, We can see 2 table entries that are duplicates of 2 others and they defy the bitmask logic behind Object Datagram Type values.

These are:

- `0x2C` --> is a duplicate of `0x28` yet, it has a different type value plus the binary representation of the table in the form of (Status? - 0 - [Priority?]' , [ObjectId?]' , EndOfGroup?, Extensions?) doesn't really come to add up to `0x2C` in hex but rather adds up to `0x28`.
- `0x2D` --> is a duplicate of `0x29` yet, it has a different type value plus the binary representation of the table in the form of (Status? - 0 - [Priority?]' , [ObjectId?]' , EndOfGroup?, Extensions?) doesn't really come to add up to `0x2D` in hex but rather adds up to `0x29`.

Thanks to @zafergurel for double-checking this, As he stated the issue is probably caused by a copy-paste of rows and than those rows being forgotten to have an update on their values

I can't really assume what is being purposed with those 2 rows but the possibe fix to the issue is updating the rows as:

| Type | End Of Group | Extensions | Object ID | Priority Present | Status / Payload |
|---------|---------------------|-----------------|----------------|------------------------|-------------------------|
| `0x2C` | No                  | No               | No              | No                       | Status                  |
| `0x2D` | No                  | Yes               | No              | No                       | Status                  |

afrind:
I think this was an editorial mistake -- unless anyone has actually implemented.

PR #1400, opened 2025-12-13
Describe bit fields instead of tables for Datagram and Subgroup Header

afrind:
Fixes: #1385

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3333,39 +3333,39 @@ OBJECT_DATAGRAM {
 {: #object-datagram-format title="MOQT OBJECT_DATAGRAM"}
 
 The Type field in the OBJECT_DATAGRAM takes the form 0b00X0XXXX (or the set of
-values from 0x00 to 0x0F, 0x20 to 0x2F). However, not all Type values in this range
-are valid. The four low-order bits and bit 5 of the Type field determine which fields
-are present in the datagram:
+values from 0x00 to 0x0F, 0x20 to 0x2F). However, not all Type values in this
+range are valid. The four low-order bits and bit 5 of the Type field determine
+which fields are present in the datagram:
 
-* The **EXTENSIONS** bit (0x01) in the Type is set to indicate that there is an
-  Extensions field present. When set to 1, the Extensions structure defined in
-  {{object-extensions}} is present. When set to 0, the Extensions field is absent.
-  If an endpoint receives a datagram with the EXTENSIONS bit set and an Extension
-  Headers Length of 0, it MUST close the session with a `PROTOCOL_VIOLATION`.
+* The **EXTENSIONS** bit (0x01) indicates when the Extensions field is
+  present. When set to 1, the Extensions structure defined in
+  {{object-extensions}} is present. When set to 0, the Extensions field is
+  absent.  If an endpoint receives a datagram with the EXTENSIONS bit set and an
+  Extension Headers Length of 0, it MUST close the session with a
+  `PROTOCOL_VIOLATION`.
 
 * The **END_OF_GROUP** bit (0x02) indicates End of Group. When set to 1, this
-  indicates that no Object with the same Group ID and an Object ID greater than the
-  Object ID in this datagram exists.
+  indicates that no Object with the same Group ID and an Object ID greater than
+  the Object ID in this datagram exists.
 
-* The **NO_OBJECT_ID** bit (0x04) in the Type is set to indicate that the Object ID
-  field is absent and the Object ID is 0. When set to 1, the Object ID field is
-  omitted. When set to 0, the Object ID field is present.
+* The **ZERO_OBJECT_ID** bit (0x04) indicates when the Object ID field is
+  present.  When set to 1, the Object ID field is omitted and the Object ID is
+  0. When set to 0, the Object ID field is present.
 
-* The **NO_PRIORITY** bit (0x08) in the Type is set to indicate that the Priority
-  field is absent. When set to 1, the Priority field is omitted and this Object
-  inherits the Publisher Priority specified in the control message that established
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -3300,63 +3300,35 @@ OBJECT_DATAGRAM {
...
-  * For Type values 0x00 through 0x07, the Object Payload is present
-    and the Object Status field is omitted.
-
-    There is no explicit length field for the Object Payload. The entirety of the
-    transport datagram following the Object header fields contains the payload.
-
-  * For Type values 0x20, 0x21, 0x24 and 0x25 the Object Status field is present
-    and there is no Object Payload.
+The Type field in the OBJECT_DATAGRAM takes the form 0b00X0XXXX (or the set of
+values from 0x00 to 0x0F, 0x20 to 0x2F). The three low-order bits and bit 5
```
afrind: We need to list the values that are not valid (eg: status + extensions)

draft-ietf-moq-transport.md
```diff
@@ -3406,81 +3378,39 @@ SUBGROUP_HEADER {
...
-present and all Objects have no extensions.  When Extensions Present is Yes, the
-Extensions structure defined in {{object-extensions}} is present in all Objects
-in this subgroup.  Objects with no extensions set Extension Headers Length to 0.
-
-For Type values where Priority Present is No, Priority is not present and this
-Subgroup inherits the Publisher Priority specified in the control message that
-established the subscription.  When Priority Present is Yes, the Priority field
-is present in the Subgroup header.
+The Type field in the SUBGROUP_HEADER takes the form 0b00X1XXXX (or the set of
+values from 0x10 to 0x1F, 0x30 to 0x3F), where bit 4 is always set to 1. The
```
afrind: same, there are invalid values here as well

draft-ietf-moq-transport.md
kirki58: Hey, aren't status datagrams allowed to have extensions? I think *status + end-of-group* should be considered invalid?

ianswett: No, only Objects have extensions.  The Objects can be empty, but there's no reason to attach properties to a thing that doesn't exist.

draft-ietf-moq-transport.md
```diff
@@ -3300,63 +3300,35 @@ OBJECT_DATAGRAM {
...
+  field is absent. When set to 1, the Priority field is omitted and this Object
+  inherits the Publisher Priority specified in the control message that established
+  the subscription. When set to 0, the Priority field is present.
+
+* The **STATUS** bit (0x20) determines whether the datagram contains an Object Status
+  or Object Payload. When set to 1, the Object Status field is present and there is
+  no Object Payload. When set to 0, the Object Payload is present and the Object
+  Status field is omitted. There is no explicit length field for the Object Payload;
+  the entirety of the transport datagram following the Object header fields contains
+  the payload.
```
sharmafb: Why are some of these phrased as negatives (e.g. NO_OBJECT_ID) and others phrased as positives (e.g. STATUS)? Can we be consistent with our conventions here?

afrind: The current bits evolved in a way intended to be somewhat backwards compatible with earlier version.  I think we can normalize all the bits when we get to the final product such that 1 means "included" wherever possible.  Also the "is subgroup" bit probably shouldn't be bit 4.

yekuiwang:
Hmm, a good editorial improvement.

yekuiwang (approved):

mzanaty:
Why not define Type using RFC 9000 notation for actual bit fields? Like:
Type (2) = 0,
Stream (1), // 1=stream, 0=datagram
End of Group (1),
Priority Present (1),
ID Present (1),
ID Zero (1),
Extensions Present (1)

draft-ietf-moq-transport.md
ianswett: Does this still need to be fixed?

draft-ietf-moq-transport.md
```diff
@@ -3331,63 +3331,35 @@ OBJECT_DATAGRAM {
...
-  * For Type values 0x00 through 0x07, the Object Payload is present
-    and the Object Status field is omitted.
-
-    There is no explicit length field for the Object Payload. The entirety of the
-    transport datagram following the Object header fields contains the payload.
-
-  * For Type values 0x20, 0x21, 0x24 and 0x25 the Object Status field is present
-    and there is no Object Payload.
+The Type field in the OBJECT_DATAGRAM takes the form 0b00X0XXXX (or the set of
+values from 0x00 to 0x0F, 0x20 to 0x2F). The three low-order bits and bit 5
```
ianswett: Isn't it the four low order bits, not three?

```suggestion
values from 0x00 to 0x0F, 0x20 to 0x2F). The four low-order bits and bit 5
```

draft-ietf-moq-transport.md
```diff
@@ -3331,63 +3332,47 @@ OBJECT_DATAGRAM {
...
+  Extensions field present. When set to 1, the Extensions structure defined in
+  {{object-extensions}} is present. When set to 0, the Extensions field is absent.
+  If an endpoint receives a datagram with the EXTENSIONS bit set and an Extension
+  Headers Length of 0, it MUST close the session with a `PROTOCOL_VIOLATION`.
+
+* The **END_OF_GROUP** bit (0x02) indicates End of Group. When set to 1, this
+  indicates that no Object with the same Group ID and an Object ID greater than the
+  Object ID in this datagram exists.
+
+* The **NO_OBJECT_ID** bit (0x04) in the Type is set to indicate that the Object ID
```
```suggestion
* The **ZERO_OBJECT_ID** bit (0x04) in the Type is set to 1 to indicate that the Object ID
```

PR #1401, opened 2025-12-15
Add new Filters for Objects/Tracks in Sub/NS/PubOk/Upd/Fetch, replace Subscription Filters

mzanaty:
Add new Filters for Objects/Tracks in Sub/NS/PubOk/Upd/Fetch, replace Subscription Filters
```
 LOCATION_FILTER { Type=0x21, Len, [StartG, [StartO, [EndG, [EndO]]]] }
    GROUP_FILTER { Type=0x23, Len, [StartG, [EndG]]... }
 SUBGROUP_FILTER { Type=0x25, Len, [StartS, [EndS]]... }
   OBJECT_FILTER { Type=0x27, Len, [StartO, [EndO]]... }
 PRIORITY_FILTER { Type=0x29, Len, [StartP, [EndP]]... }
EXTENSION_FILTER { Type=0x2B, Len, [ExtType, StartV, [EndV]]... }
    TRACK_FILTER { Type=0x2D, Len, [ExtType, MaxS, MaxD, MaxT] }

G=GroupID, S=SubgroupID, O=ObjectID, P=Priority, V=Value of Extension
MaxS=MaxTracksSelected, MaxD=MaxTracksDeselected, MaxT=MaxTimeSelected
```

Fixes #1206
Fixes #441
Fixes: #899

PR #1404, opened 2025-12-20
RFC: Add MAX_SEND_LOCATION to provide backpressure to a Subscription

### LINKED ISSUE #869: Difficult to limit resource consumption of a Subscription
For FETCH, flow control on the stream can limit the resource consumption.

For Subscribe, there's no limit to the number of Subgroups that a publisher can publish to simultaneously, which could allow one Track to consume a huge number of streams, potentially starving resources from other subscriptions.

A simple solution would be to add a Parameter to SUBSCRIBE that allows limiting the number of simultaneous unidirectional streams for the subscription.

ianswett:
MAX_SEND_LOCATION consists of a Group ID followed by a Subgroup ID, equal to or above which Objects should not be sent.

I considered making it a Location (Group ID, Object ID), but streams already have flow control and limits, so I thought that was slightly better, but I could be convinced that Location is better if people have a strong preference.

Relevant to #1316 and #1358

Fixes #869

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -332,6 +332,35 @@ Reason Phrase {
   such as language tags, that would aid comprehension by any entity other than
   the one that created the text.
 
+## Representing Namespace and Track Names
+
+There is often a need to render namespace tuples and track names for
+purposes such as logging, representing track filenames, or use in
+certain authorization verification schemes. The namespace and track name
+are binary, so they need to be converted to a safe form.
+
+The following format is RECOMMONDED:
+
+* Each of the namespace tuples are rendered in order with a hyphen (-)
+  between them followed by the track name with a double hyphen (--)
+  between the last namespace and track name.
+
+* Bytes in the range a-z, A-Z, 0-9 as well as _ (0x5f) are output as is,
+  while all other bytes are encoded as a period (.) symbol followed by
+  exactly two lower case hex digits.
+
+The goal of this format is to have a format that is both filename and
+URL safe. It allows many common names to be rendered in an easily human
+readable form while still supporting binary values.
+
+Example:
+
+~~~
+example.2enet-team2-project_x--report
+  Namespace tuples: (example.net, team2, project_x)
+  Track name: report
+~~~
+
 # Object Data Model {#model}
 
 MOQT has a hierarchical data model, comprised of tracks which contain
@@ -1843,8 +1872,10 @@ SUBSCRIBE, SUBSCRIBE_OK, or SUBSCRIBE_UDPATE message.
 It is the duration in milliseconds the relay SHOULD
 continue to attempt forwarding Objects after they have been received.  The start
 time for the timeout is based on when the Object Headers are received, and does
-not depend upon the forwarding preference. There is no explicit signal that an
-Object was not sent because the delivery timeout was exceeded.
+not depend upon the forwarding preference. Objects with forwarding preference
+'Datagram' are not retransmitted when lost, so the Delivery Timeout only limits
+the amount of time they can be queued before being sent. There is no explicit
+signal that an Object was not sent because the delivery timeout was exceeded.
 
 DELIVERY_TIMEOUT, if present, MUST contain a value greater than 0.  If an
 endpoint receives a DELIVERY_TIMEOUT equal to 0 it MUST close the session


draft-ietf-moq-transport.md
```diff
@@ -1947,6 +1947,26 @@ publisher MUST close the session with `PROTOCOL_VIOLATION`.
...
+prevents too many Objects from being sent at once. The Subscription will
+still deliver the same Objects that it would without PAUSE_AT, once the
+PAUSE_AT Parameter is removed or increased to a sufficiently large value.
+
+PAUSE_AT can limit the streams and bandwidth consumed by a single
+Subscription, reducing the likelihood of the Session running out of stream
+or data flow control.
+
+If a Subscription remains paused for too long, a Publisher MAY decide to
+terminate the Subscription, just like it can for those with Forward=0.
```
suhasHere: what should be the expected behavior if the parameter is omitted or set to 0 . May be both means there is no pausing ?

I am not able to understand how would application know when and what value to set at ? If we decide to limit, may be it is better as a session setup parameter on max number of subscriptions and max number of streams per subscription along with a message to modify it when the limit is reached ?

vasilvv:
So far, our approach to backpressure in MOQ has been "if you cannot handle the track at the rate it is generated, unsubscribe".  Is there a reason we should reconsider this?

draft-ietf-moq-transport.md
```diff
@@ -1947,6 +1947,26 @@ publisher MUST close the session with `PROTOCOL_VIOLATION`.
...
+The PAUSE_AT parameter (Parameter Type 0x23) MAY appear in a SUBSCRIBE,
+PUBLISH_OK, or REQUEST_UPDATE (for a subscription) message. It is a
+length-prefixed Group ID and Subgroup ID.  Objects with a larger Group ID
+or an equal Group ID and larger or equal Subgroup ID SHOULD NOT be sent
+for the Subscription until PAUSE_AT is increased.
+
+PAUSE_AT is different from the end of a Subscription filter because it only
+prevents too many Objects from being sent at once. The Subscription will
+still deliver the same Objects that it would without PAUSE_AT, once the
+PAUSE_AT Parameter is removed or increased to a sufficiently large value.
```
fluffy: This seems like a very major change because now my relay had to be able to separately queue all objects for all downstream subscribers. This is simply not feasible for a non caching relay.

afrind: At a minimum, your outbound QUIC stack already has to buffer stream data until it has CWND, stream and connection flow control credit and is then acknowledged, or the delivery timeout, none of which are in the control of the relay.  That's why a publisher is allowed to terminate a subscription with TOO_FAR_BEHIND if the queue becomes too large.

draft-ietf-moq-transport.md
afrind: I think this PR would read better if you called it MAX_SEND_LOCATION or something - to indicate it's similar to other kinds of flow control.  I would expect not set means MAX_SEND_LOCATION=<maxint, maxint>?

afrind: Also forward=0 isn't quite the right reference, you should refer to the TOO_FAR_BEHIND section.

ianswett: Thanks for the naming suggestion, I renamed this to MAX_SEND_LOCATION.

Is it ok to say that a length of 0 (ie: REQUEST_UPDATE) means the value is cleared, equivalent to <maxint, maxint>?

draft-ietf-moq-transport.md
ianswett: I agree it's not a small change, but as Alan said, a Subscription can already get quite behind today.

One use case for this is a MoQ relay that is only pull through with no caching (sort of like HTTPS LBs commonly are today), then this provides a way for the relay to put backpressure upstream when the Original Subscriber starts getting behind.

draft-ietf-moq-transport.md
```diff
@@ -1947,6 +1947,26 @@ publisher MUST close the session with `PROTOCOL_VIOLATION`.
 If omitted from SUBSCRIBE or PUBLISH_OK, the subscription is
 unfiltered.  If omitted from SUBSCRIBE_UDPATE, the value is unchanged.
 
+#### MAX_SEND_LOCATION Parameter {#pause-at}
+
+The MAX_SEND_LOCATION parameter (Parameter Type 0x23) MAY appear in a
+SUBSCRIBE, PUBLISH_OK, or REQUEST_UPDATE (for a subscription) message.
+It is a length-prefixed Group ID and Subgroup ID.  Objects with a larger
```
afrind, line 1985: It occurs to me that Group ID + Subgroup ID is not a Location, so the name needs more work.

Also, subgroup IDs aren't necessarily sent in any particular order, so setting a max may be challenging.

Would you need the equivalent of BLOCKED?

draft-ietf-moq-transport.md
```diff
@@ -1947,6 +1947,26 @@ publisher MUST close the session with `PROTOCOL_VIOLATION`.
...
+MAX_SEND_LOCATION is different from the end of a Subscription filter because it
+only prevents too many Objects from being sent at once. The Subscription will
+still deliver the same Objects that it would without MAX_SEND_LOCATION, once the
+MAX_SEND_LOCATION Parameter is removed or increased to a sufficiently large value.
+
+MAX_SEND_LOCATION can limit the streams and bandwidth consumed by a single
+Subscription, reducing the likelihood of the Session running out of stream
+or data flow control.
+
+If a Subscription remains paused for too long, a Publisher MAY decide to
```
afrind, line 1998: We currently say this in DELIVERY_TIMEOUT:

> If a subscriber fails to consume Objects at a sufficient rate, causing the publisher to exceed its resource limits, the publisher MAY terminate the subscription with error TOO_FAR_BEHIND.

draft-ietf-moq-transport.md
```diff
@@ -1947,6 +1947,26 @@ publisher MUST close the session with `PROTOCOL_VIOLATION`.
...
+terminate the Subscription with error `TOO_FAR_BEHIND`.
+
```
afrind, line 2000: You need to say what the default is when not present

afrind:
I have an alternate idea for subgroup limits, which is a total subgroup count for the subscription.  This way the subscriber doesn't need to know as much about the subgroup ID assignments in the track (are the sg ids 1, 2, 3 or 10000, 20000, 30000)

draft-ietf-moq-transport.md
fluffy: Buffering in the QUIC stack makes this *way* worse. Once it is in the quic stack, you have a copy per subscriber receiving it. While it is buffered in the MoQ stack, you can just keep a refernce to the copy in cache for each subscriber that still needs to receive the data. Plus when it is in the MoQ side, you can still time it out, change send orders due to priorities of new objects that arrived. I think the well designed relays are going to tend to mostly only hand objects to the QUIC stack at the point the stack has a good chance of sending it very soon.

afrind: >  Once it is in the quic stack, you have a copy per subscriber receiving it. 

This depends on your QUIC stack.  Some stacks support reference counted buffers.

afrind: Also, this PR has been overhauled since the text that spawned this thread.

draft-ietf-moq-transport.md
```diff
@@ -1978,6 +1978,26 @@ publisher MUST close the session with `PROTOCOL_VIOLATION`.
 If omitted from SUBSCRIBE or PUBLISH_OK, the subscription is
 unfiltered.  If omitted from SUBSCRIBE_UDPATE, the value is unchanged.
 
+#### MAX_SEND_LOCATION Parameter {#pause-at}
+
+The MAX_SEND_LOCATION parameter (Parameter Type 0x23) MAY appear in a
+SUBSCRIBE, PUBLISH_OK, or REQUEST_UPDATE (for a subscription) message.
```
TimEvens, line 1984: Considering `REQUEST_UPDATE` has a wider scope, it would be ideal to specifically call out which messages instead of a vague "for subscriptions"...

draft-ietf-moq-transport.md
```diff
@@ -1978,6 +1978,26 @@ publisher MUST close the session with `PROTOCOL_VIOLATION`.
 If omitted from SUBSCRIBE or PUBLISH_OK, the subscription is
 unfiltered.  If omitted from SUBSCRIBE_UDPATE, the value is unchanged.
 
+#### MAX_SEND_LOCATION Parameter {#pause-at}
+
+The MAX_SEND_LOCATION parameter (Parameter Type 0x23) MAY appear in a
```
TimEvens, line 1983: I see value in setting this in `SUBSCRIBE_NAMESPACE` so that matched tracks can inherit the max location at start and can request update to move it forward. It would be nice if matched tracks inherit defaults/`SUBSCRIBE_NAMESPACE` parameters that are applied to matched tracks.

TimEvens, line 1983: Code point 0x23 appears to conflict with #1401  `GROUP_FILTER`.

TimEvens (changes requested):
As written, `MAX_SEND_LOCATION` functions as a group/subgroup-level filter, but it is not clearly described as a filter on subscribed objects. It would be clearer to explicitly define it as an object filter, potentially renaming it to something like `OBJECT_FILTER_MAX_LOCATION` to align with that intent.

My concern with this parameter is that it will introduce churn due to repeated `REQUEST_UPDATE` calls to advance the limit. Additionally, increasing the limit could result in microbursting.

The PR also does not clearly describe the resume semantics when the limit is increased. As written, it appears to suggest that once the limit is raised, delivery resumes from where it previously stopped — effectively behaving like a joining fetch triggered by the limit increase.

For example, suppose the max is set to group 5, subgroup 2. The limit is not increased, but objects continue flowing and the track advances to group 10, subgroup 100. If the client later increases the limit to group 50, subgroup 1000, the sender would need to deliver all objects from group 5, subgroup 2 up to the current head. This would likely cause microbursting and would also introduce additional complexity around delivery timeouts, since these resumed objects would now be subject to timeout semantics based on the limit increase.

If the intent is not to perform a joining fetch on limit increase, then buffering would be required to support resumption. However, buffering is something many implementations try to avoid — especially given that cache already serves a similar role.

There are also use cases where `TOO_FAR_BEHIND` should not result in terminating the subscription. In those cases, it may make more sense to jump to head upon a significant limit increase rather than attempting to deliver the entire backlog. One possible approach would be to allow limit increases within a bounded delta; if that delta is exceeded, the subscription would skip ahead to head instead of replaying all missed objects.

ianswett:
Closing this for now in favor of the Rewind proposal: https://github.com/martinduke/draft-duke-moq-subscribe-rewind

We can revisit flow control ideas later if they're still of interest.

PR #1406, opened 2025-12-21
Clarify Delivery Timeout for datagrams

### LINKED ISSUE #1103: Delivery timeout, Datagrams, and Caching

ianswett:
Now that forwarding preferences can be mixed within a Track, I think this is the only other change needed to fix #1103.

Fixes #1103

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -332,6 +332,35 @@ Reason Phrase {
   such as language tags, that would aid comprehension by any entity other than
   the one that created the text.
 
+## Representing Namespace and Track Names
+
+There is often a need to render namespace tuples and track names for
+purposes such as logging, representing track filenames, or use in
+certain authorization verification schemes. The namespace and track name
+are binary, so they need to be converted to a safe form.
+
+The following format is RECOMMONDED:
+
+* Each of the namespace tuples are rendered in order with a hyphen (-)
+  between them followed by the track name with a double hyphen (--)
+  between the last namespace and track name.
+
+* Bytes in the range a-z, A-Z, 0-9 as well as _ (0x5f) are output as is,
+  while all other bytes are encoded as a period (.) symbol followed by
+  exactly two lower case hex digits.
+
+The goal of this format is to have a format that is both filename and
+URL safe. It allows many common names to be rendered in an easily human
+readable form while still supporting binary values.
+
+Example:
+
+~~~
+example.2enet-team2-project_x--report
+  Namespace tuples: (example.net, team2, project_x)
+  Track name: report
+~~~
+
 # Object Data Model {#model}
 
 MOQT has a hierarchical data model, comprised of tracks which contain


draft-ietf-moq-transport.md
```diff
@@ -1843,8 +1843,10 @@ SUBSCRIBE, SUBSCRIBE_OK, or SUBSCRIBE_UDPATE message.
 It is the duration in milliseconds the relay SHOULD
 continue to attempt forwarding Objects after they have been received.  The start
 time for the timeout is based on when the Object Headers are received, and does
-not depend upon the forwarding preference. There is no explicit signal that an
-Object was not sent because the delivery timeout was exceeded.
+not depend upon the forwarding preference. Objects with forwarding preference
+'Datagram' are not retransmitted when lost, so the Delivery Timeout only limits
```
afrind, line 1876: I think we should also define the properties of datagrams higher up in the doc, but that can wait for a larger editorial pass.

ianswett: I agree that this text is way too far into the doc.  I was quite surprised by that, TBH.  We should definitely do a bit more editorial text movement sometime soonish.  I'm not sure the sections are wrong, as much as the subsections need to be moved up.

draft-ietf-moq-transport.md
```diff
@@ -1843,8 +1843,10 @@ SUBSCRIBE, SUBSCRIBE_OK, or SUBSCRIBE_UDPATE message.
...
+the amount of time they can be queued before being sent. There is no explicit
```
martinduke, line 1877: s/can/ought to be

some APIs will probably not allow you to revoke a datagram.

afrind: or SHOULD?

ianswett: There's a SHOULD above, so I don't think we need another one?

```suggestion
the amount of time they are queued before being sent. There is no explicit
```

afrind: Actually, I think `can` is fine if the concern is about implementations not having a mechanism to cancel a queued datagram - `are` is probably worse.  The sentence is:

> the Delivery Timeout only limits the amount of time they can be queued before being sent.

As @ianswett points out, this is a SHOULD. Further, we don't define what "sent" means here, but in the priority section we explain it can mean "pass to the next layer".

And implementations may have to add more capabilities if they want to be maximally functional and performant.

PR #1408, opened 2025-12-21
A Publisher might not use the congestion window

### LINKED ISSUE #471: Should any priority values indicate "Less than best effort"? / How do you upswitch without causing user harm?
There was a discussion at the interim about how having one or more priority values indicate less than best effort might be useful, and I believe client side ABR is a good use case.

How does a client decide to switch to a higher quality?  With the proposed priority scheme, the higher quality could be made lower priority, but it doesn't prevent higher quality data from consuming a large portion of the bandwidth and causing congestion that negatively impacts delivery of the lower quality.

To give a typical example, the client is considering switching to the next highest quality, which is 2x the high bitrate of the current video quality.  The current quality is consuming slightly more than 1/2 of the subscriber's bandwidth, so the subscriber will not have the bandwidth to switch to the higher quality.  Without caution, too much data could be sent and cause bufferbloat and/or loss.

My strawman proposal is that LBE means is that if the publisher only had LBE content to send, it would alter its behavior in terms of how much/when to send.  For example, the core BBR model sets the CWND to 2*BDP the vast majority of the time, but maybe LBE would not be sent unless CWND is < 1.5*BDP.  This ensures there is always CWND available for higher priority data.  Ideally BBR should be pacing limited, and when it's CWND limited, something is typically going poorly.

ianswett:
Fixes #471

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1485,7 +1485,7 @@ all possible Objects matching the filter have been delivered to the subscriber.
 Since tracking this can be prohibitively expensive, Relays are not required or
 expected to do so.
 
-### Graceful Subscriber Relay Switchover
+### Graceful Subscriber Relay Switchover {#graceful-subscriber-switchover}
 
 This section describes behavior a subscriber MAY implement
 to allow for a better user experience when a relay sends a GOAWAY.
@@ -1573,7 +1573,7 @@ the Relay MUST NOT propagate that message or Object to another Session, because
 it would enable a single Session error to force an unrelated Session, which
 might be handling other subscriptions, to be closed.
 
-### Graceful Publisher Network Switchover
+### Graceful Publisher Network Switchover {#graceful-publisher-switchover}
 
 This section describes a behavior that a publisher MAY
 choose to implement to allow for a better user experience when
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1340,6 +1340,10 @@ to different subscriptions or FETCH responses, but have the same subscriber and
 publisher priority.  The ordering in those cases is implementation-defined,
 though the expectation is that all subscriptions will be able to send some data.
 
+A publisher might not utilize the entire available congestion window for lower
```
afrind: do you want to mention flow control window also?

ianswett: Hmm, interesting point, is this better?

```suggestion
A publisher might not utilize the entire available congestion window,
session flow control, or all available streams for lower
```

draft-ietf-moq-transport.md
```diff
@@ -1340,6 +1340,10 @@ to different subscriptions or FETCH responses, but have the same subscriber and
...
+priority Objects if it expects higher priority Objects to be available to send
```
```suggestion
priority Objects if it expects higher priority Objects will be available to send
```

draft-ietf-moq-transport.md
afrind: lgtm

PR #1409, opened 2026-01-03
Split malformed Tracks into two lists

### LINKED ISSUE #1345: Separate the list of reasons for malformed tracks into two lists
In discussions in #1317, @ianswett suggested that it'd be nice to separate the list of reasons for malformed tracks into two lists in the future: One that applies to both SUBSCRIBE and FETCH, and one that only applies to FETCH.

ianswett:
Fixes #1345

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -124,7 +124,7 @@ rapidly detect and respond to congestion.
 
 The parallel nature of QUIC streams can provide improvements in the face
 of loss. A goal of MOQT is to design a streaming protocol to leverage
-the transmission benefits afforded by parallel QUIC streams as well
+the transmission benefits afforded by parallel QUIC streams as well as
 exercising options for flexible loss recovery.
 
 ### Convergence
@@ -259,7 +259,7 @@ after the first 0 and subsequent bytes, if any, represent the integer value,
 encoded in network byte order.
 
 Integers are encoded in 1, 2, 3, 4, 5, 6, 8, or 9 bytes and can encode up to 64
-bit unsinged integers. The following table summarizes the encoding properties.
+bit unsigned integers. The following table summarizes the encoding properties.
 
 |--------------|----------------|-------------|------------------------|
 | Leading Bits | Length (bytes) | Usable Bits | Range                  |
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -564,34 +564,37 @@ to MOQT constraints. Such a Track is considered malformed.  Some example
...
-10. An Object is received with a different Forwarding Preference than previously
+6. An Object is received with a different Forwarding Preference than previously
     observed.
 
+Additionally, a FETCH responce is considered malformed if:
+
+1. An Object is received in a FETCH response with the same Group ID as the
+   previous Object, but whose Object ID is not strictly larger than the previous
+   object.
+2. In a FETCH response, an Object with a particular Subgroup ID is received, but its
```
```suggestion
2. An Object with a particular Subgroup ID is received, but its
```

draft-ietf-moq-transport.md
```diff
@@ -564,34 +564,37 @@ to MOQT constraints. Such a Track is considered malformed.  Some example
 conditions that constitute a malformed track when detected by a receiver
 include:
```
afrind: receiver of a subscription?

draft-ietf-moq-transport.md
```diff
@@ -561,37 +561,40 @@ Namespace Fields or Track Name such that exact comparison works.
 
 There are multiple ways a publisher can transmit a Track that does not conform
 to MOQT constraints. Such a Track is considered malformed.  Some example
-conditions that constitute a malformed track when detected by a receiver
+conditions that constitute a malformed track when detected by a subscriber
```
afrind, line 655: FETCH is issued by subscribers too - so this needs to be focused on subscriptions.

draft-ietf-moq-transport.md
```diff
@@ -561,37 +561,40 @@ Namespace Fields or Track Name such that exact comparison works.
...
    Status END_OF_TRACK or the last Object sent in a FETCH whose response indicated
    End of Track.
-9. The same Object is received more than once with different Payload or
+5. The same Object is received more than once with different Payload or
     other immutable properties.
-10. An Object is received with a different Forwarding Preference than previously
+6. An Object is received with a different Forwarding Preference than previously
     observed.
 
+Additionally, a FETCH responce is considered malformed if:
```
martinduke:
```suggestion
Additionally, a FETCH response is considered malformed if:
```

afrind:
>  IMO it should be a protocol violation if a sender does any of these.

Do you want to blow up the entire connection if the publisher sends bad FETCH data?

I wonder if we could eliminate a bunch of these by making FETCH locations (when included explicitly) delta encoded...

martinduke:
> > IMO it should be a protocol violation if a sender does any of these.
> 
> Do you want to blow up the entire connection if the publisher sends bad FETCH data?
> 
> I wonder if we could eliminate a bunch of these by making FETCH locations (when included explicitly) delta encoded...

Why wouldn't I? The publisher is not obeying the protocol in an easily detectable way.

draft-ietf-moq-transport.md
ianswett: The intent is the first list applies to 'Tracks', so either Subscriptions or Fetches, and the second list is only applicable to Fetch.

1 and 2 seem specific to Subscriptions, but the other 3 are not?

ianswett:
I'm closing this in favor of delta encoding Fetch to remove 3 of the cases.

PR #1410, opened 2026-01-04
Subscribers can migrate networks too

### LINKED ISSUE #848: Network Switchover
Draft says:

> This section describes behavior that a publisher MAY choose to implement to allow for a better users experience when switching between networks, such as WiFi to Cellular or vice versa.

This feels like a situation that is better suited for QUIC connection migration.  The latter paragraph discusses an endpoint that "does not have QUIC connection migration available", but it seems to me like this is the wrong direction -- the solution for the discussed problem is connection migration, and if the client has access to two network interfaces at the same time, it is more than capable of performing it.

ianswett:
Fixes #848 

This text might be better in the "Migration" section? https://github.com/moq-wg/moq-transport/blob/main/draft-ietf-moq-transport.md#migration-session-migration

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -339,7 +339,7 @@ purposes such as logging, representing track filenames, or use in
 certain authorization verification schemes. The namespace and track name
 are binary, so they need to be converted to a safe form.
 
-The following format is RECOMMONDED:
+The following format is RECOMMENDED:
 
 * Each of the namespace tuples are rendered in order with a hyphen (-)
   between them followed by the track name with a double hyphen (--)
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1559,45 +1560,32 @@ When a relay receives an authorized PUBLISH_NAMESPACE for a namespace that
...
 If a Session is closed due to an unknown or invalid control message or Object,
 the Relay MUST NOT propagate that message or Object to another Session, because
 it would enable a single Session error to force an unrelated Session, which
 might be handling other subscriptions, to be closed.
 
-### Graceful Publisher Network Switchover
+### Graceful Publisher Relay Switchover
+
+This section describes a behavior that a publisher MAY implement to improve
+user experience when a relay sends a GOAWAY or the Subscriber switches between
```
sharmafb: Did you mean to write Publisher here, and not Subscriber?

ianswett: Yes, thanks.

```suggestion
user experience when a relay sends a GOAWAY or the publisher switches between
```

draft-ietf-moq-transport.md
```diff
@@ -1482,8 +1482,9 @@ expected to do so.
 
 ### Graceful Subscriber Relay Switchover {#graceful-subscriber-switchover}
 
-This section describes behavior a subscriber MAY implement
-to allow for a better user experience when a relay sends a GOAWAY.
+This section describes a behavior that a publisher MAY implement to improve
```
```suggestion
This section describes a behavior that a Subscriber MAY implement to improve
```

PR #1418, opened 2026-01-13
Clarification for Track Alias uniqueness

sharmafb:
By adding "by a publisher," I'm clarifying that the namespace for Track Aliases is distinct for each direction of the connection.
* The Client (as publisher) manages its own set of Aliases (e.g., Alias 1 maps to ClientTrackA).
* The Server (as publisher) manages its own set of Aliases (e.g., Alias 1 maps to ServerTrackB).


afrind (approved):

afrind:
I might let a few more folks read this instead of sneaking it into -16

ianswett (approved):

PR #1425, opened 2026-01-19
Fix example namespace in draft-ietf-moq-transport.md

itzmanish:
Fixes small typo on example section for representing namespace and track names.


ianswett (approved):

afrind:
+1 Victor -- `.` is escaped in this encoding.

Issue #1424, opened 2026-01-16
We Need MOQPACK


martinduke:
Now that Auth Tokens can arrive on the SUB_NS stream, Token Registry operations are no longer HOL blocked. The draft needs language for how to handle this case.

I don't really care which of many solutions we choose.

afrind:
The band-aid solution to this problem is you can't use a TOKEN_ALIAS until you've seen the response (OK or ERROR) to the request that had REGISTER.  This would prevent pipelined requests from using aliases.

Another reasonable option is to remove token compression from the draft, and move it to an extension.

If we need the full complexity of the QPACK dynamic table, with sending requests that may block on the arrival of an out-of-order request, we would do better to find a way to import that (also likely via an extension) rather than recreate it.

See #969

ianswett:
SUBSCRIBE_NS can safely reference token aliases in CLIENT_SETUP, right?

martinduke:
yes -- I think anything registered in CLIENT_SETUP is globally safe.

afrind:
Yes, the issue in 16 is tokens registered in SUB NS can’t be reliably referenced elsewhere until after a response, and those registered on the control stream can’t be referenced in SUB NS until a response.

PR #1430, opened 2026-01-23
Define how to use auth cache safely with multiple streams

afrind:
Fixes: #1424

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1876,12 +1876,12 @@ the combination of Token Type and Token Value are unique after resolving any
 aliases.
 
 Messages carrying the AUTHORIZATION TOKEN parameter can appear on different
-streams. Because stream processing order can be different than send order, the
+control streams. Because stream processing order can be different than send order, the
 receiver and sender can have inconsistent views of the token cache state.
 
-Senders MUST NOT send USE_ALIAS on one stream for an alias registered on a
+Senders MUST NOT send USE_ALIAS on one control stream for an alias registered on a
 different stream until the sender has received a response to the message
-containing the REGISTER. Senders MAY use USE_ALIAS on the same stream as the
+containing the REGISTER. Senders MAY use USE_ALIAS on the same control stream as the
 REGISTER without waiting for a response.
 
 Senders MUST NOT send DELETE for an alias while any message using USE_ALIAS with


draft-ietf-moq-transport.md
```diff
@@ -1875,6 +1875,18 @@ The AUTHORIZATION TOKEN parameter MAY be repeated within a message as long as
 the combination of Token Type and Token Value are unique after resolving any
 aliases.
 
+Messages carrying the AUTHORIZATION TOKEN parameter can appear on different
+streams. Because stream processing order is not guaranteed, the receiver and
+sender can have inconsistent views of the token cache state.
```
```suggestion
streams. Because streams can be processed in a different order than messages are sent,
the receiver and sender can have inconsistent views of the token cache state.
```

draft-ietf-moq-transport.md
```diff
@@ -1875,6 +1875,18 @@ The AUTHORIZATION TOKEN parameter MAY be repeated within a message as long as
...
+
+Senders MUST NOT send USE_ALIAS on one stream for an alias registered on a
+different stream until the sender has received a response to the message
```
ianswett, line 1883: a successful response?

afrind: No actually.  An error response (eg track_not_found) still updates the auth cache.

draft-ietf-moq-transport.md
afrind: How about?

```suggestion
streams. Because stream processing order can be different than send order, the
receiver and sender can have inconsistent views of the token cache state.
```

afrind:
> I see the intent behind this PR, may be this is all right. But i feel bit uncomfortable to tie transport streams to auth state.

It's easier to think of this as "compression state" rather than "auth state".  Another viable alternative is to punt auth token compression entirely. We'll have a discussion on Monday.

draft-ietf-moq-transport.md
```diff
@@ -1875,6 +1875,18 @@ The AUTHORIZATION TOKEN parameter MAY be repeated within a message as long as
 the combination of Token Type and Token Value are unique after resolving any
 aliases.
 
+Messages carrying the AUTHORIZATION TOKEN parameter can appear on different
+streams. Because stream processing order can be different than send order, the
```
```suggestion
control streams. Because stream processing order can be different than send order, the
```

draft-ietf-moq-transport.md
```diff
@@ -1875,6 +1875,18 @@ The AUTHORIZATION TOKEN parameter MAY be repeated within a message as long as
...
+receiver and sender can have inconsistent views of the token cache state.
+
+Senders MUST NOT send USE_ALIAS on one stream for an alias registered on a
```
```suggestion
Senders MUST NOT send USE_ALIAS on one control stream for an alias registered on a
```

draft-ietf-moq-transport.md
```diff
@@ -1875,6 +1875,18 @@ The AUTHORIZATION TOKEN parameter MAY be repeated within a message as long as
...
+different stream until the sender has received a response to the message
+containing the REGISTER. Senders MAY use USE_ALIAS on the same stream as the
```
```suggestion
containing the REGISTER. Senders MAY use USE_ALIAS on the same control stream as the
```

Issue #1405, opened 2025-12-21
Single Object Subgroups don't need a Subgroup ID


ianswett:
A key purpose of a Subgroup ID is to allow the second half of a Subgroup to be delivered on a different stream and put together with the first half if the stream is unexpected terminated for any reason.

If a Stream only has one Object, there's no need for that.  

This matters because currently Subgroup ID is a Varint, which is a huge space and a challenge for prioritization.  The Object per stream use case is the one case I expect there could be more than 256 Subgroups today.  If we could eliminate it, we could go down to one byte, just like Priority.

My strawman is to do what we do for Objects in Datagrams: https://github.com/moq-wg/moq-transport/blob/main/draft-ietf-moq-transport.md#datagrams

afrind:
There's already a way to convey a subgroup on the wire without a subgroup ID (see e.g. Stream Header Type 0x12).  There's only two other places where the subgroup ID matters:

1) it's included in FETCH responses
2) it's a tiebreaker for priority within the same track and group

I think 1 doesn't matter that much.

For 2, perhaps a concrete change is to eliminate subgroup ID as a tiebreaker, and use "the first object ID on the subgroup".  For multi-object subgroups, I expect there won't be that many, and publishers will differentiate on priority.  For many one-object subgroups (or datagrams), there's only one object ID anyways.  This makes it easier to compare subgroups and datagrams that have the same publisher priority too.

ianswett:
I agree that 1 doesn't matter much.

First Object ID on the Subgroup seems sensible to me for prioritization.  I thought we already said that lower Object IDs are sent first if everything else is equal, but I can't find that text in the Priorities section (https://github.com/moq-wg/moq-transport/blob/main/draft-ietf-moq-transport.md#priorities-priorities).  Right now we explicitly say that Subgroup vs Datagram prioritization can be implementation dependent if everything else is equal.

I agree there likely won't be many multi-object subgroups, which is why I was trying to find a way to get Subgroup ID down to a byte.  I also think the Datagram Object framing makes a bit more sense for a single Object stream, as well as being slightly more compact.

afrind:
The only place the size of such a subgroup ID would matter is prioritization, so if we take it out of the prioritization algorithm, why do you want to make it a byte?  It's not sent on the wire anyways.

ianswett:
Discussion of the PR so far indicates we don't really want to bother with #1593 

[...]

PR #1446, opened 2026-01-31
Remove subgroup ID from priority tiebreakers

afrind:
Replace it with "first object ID in the subgroup".

This is only for cases where the publisher set equal pri for multiple subgroups (eg: publisher has no preference).  It makes it easier to compare mixed datagram / subgroup tracks.

Fixes: #1405

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1395,10 +1395,9 @@ the objects SHOULD be selected as follows:
    decide the one that is scheduled to be sent first.
 3. If two objects in response to the same request have the same subscriber
    and publisher priority and belong to the same group of the same track, the
-   one with **the lowest Subgroup ID** (for objects with forwarding preference
-   Subgroup), or **the lowest Object ID** (for objects with forwarding preference
-   Datagram) is scheduled to be sent first.  If the two objects have
-   different Forwarding Preferences the order is implementation dependent.
+   one with **the lowest Object ID** is scheduled to be sent first.  For objects
+   with forwarding preference Subgroup, the first Object ID in the Subgroup is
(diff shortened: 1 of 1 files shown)


ianswett (approved):

vasilvv:
I'm a bit confused about this.  I thought we wanted to use subgroup IDs for layer prioritization?

afrind:
> I'm a bit confused about this. I thought we wanted to use subgroup IDs for layer prioritization?

The primary way to prioritize subgroups as layers is with **Publisher Priority**.  The only reason to use SG-ID was for a group with 256+ subgroups that are logically the same priority, but sequenced.  You can still do that with this proposal since the SGs are likely single-object subgroups with the "sg-id == first object id" compression bit -- like datagrams but > than MTU and retransmittable.

You can play dice with the universe and organize set your pub pri equal for multi-object subgroups too, but it might get weird if someone joins mid-group -- the first object in the sg might not reflect the actual ordering intent.

vasilvv:
Ah, I see.  I guess this works, though SGID space is larger than publisher priorities.

afrind:
Discussed in Boulder:

For multi-subgroup objects that happen to start in the middle of the subgroup (eg largest object), first object ID is wrong and will result in strange prioritization.  Using subgroup ID seems preferable.

Perhaps the most reasonable way to remove the "implementation dependent" comparison between Datagram Object ID and Subgroup ID is -- just compare them _anyways_.

1) It seems likely that the subgroups have low numbers and there are few (0, 1, 2) and the datagrams will generally have higher object number since there are more of them.
2) This is the _third_ tiebreaker, these objects are in the same Track and Group and have the same publisher priority.  If the publisher cared, it would have set different priorities.

Issue #1402, opened 2025-12-15
Support the lingering SUBSCRIBE use case


ianswett:
At the recent interop, the Cisco implementation's implementation of SUBSCRIBE allows the SUBSCRIBE to linger even if nothing has been published yet.  Other implementations immediately returned when the Track hadn't been published.  There's a real use case here, which is to ensure the beginning of the Track doesn't get clipped off when a new Original Publisher joins.

This is very similar to what SUBSCRIBE_NAMESPACE allows, since it elicits PUBLISH from the Publisher for the specified namespace.  My best guess is that the gap here is that there are use cases you don't want to subscribe to the entire Namespace, but I'm not sure?

afrind:
Example flows:

Lingering Subscribe:

```
S->R: SUBSCRIBE, MaxWait=T
No publishers yet, start timer
P->R: PUBLISH
R->S: SUBSCRIBE_OK
```

```
S->R: SUBSCRIBE, MaxWait=T
No publishers yet, start timer
P->R: PUBLISH_NAMESPACE
R->P: PUB_NS_OK, SUBSCRIBE
P->R: SUBSCRIBE_OK
R->S: SUBSCRIBE_OK
```

SUBSCRIBE_NAMESPACE

```
S->R: SUBSCRIBE_NAMESPACE
R->S: SUB_NS_OK
P->R: PUBLISH
R->S: PUBLISH
```

```
S->R: SUBSCRIBE_NAMESPACE foo
R->S: SUB_NS_OK
P->R: PUBLISH_NAMESPACE foo

crickets?
```
Maybe this is just not a valid design?

```
S->R: SUBSCRIBE_NAMESPACE foo
R->S: SUB_NS_OK
P->R: PUBLISH_NAMESPACE foo/bar
R->S: NAMESPACE [foo/]bar
S->R: SUBSCRIBE foo/bar, track=t
R->P: SUBSCRIBE foo/bar, track=t
```

suhasHere:
We did discuss this as part of IETF 125 , slides https://datatracker.ietf.org/meeting/122/materials/slides-122-moq-early-media-00

I think when  multiple clients try to pub and sub at the same time, there is a chance that the ordering of the control messages may cause media clipping

afrind:
The response to the discussion at IETF was to address using SUBSCRIBE_NAMESPACE + PUBLISH.  Along with joining FETCH, it seems to functionally address the problem described.

What are the remaining gaps?

PR #1447, opened 2026-01-31
Add RENDEZVOUS_TIMEOUT parameter for lingering SUBSCRIBE

afrind:
Adds a new SUBSCRIBE parameter that allows subscribers to specify how long they are willing to wait for a publisher to become available. 

EDIT: first draft also used for reconnect after disconnection, but removed after feedback. See #1489 

Fixes: #1402

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -791,8 +791,8 @@ application.
 The client can establish a connection to an MOQT server identified by a given
 URI by setting up a QUIC connection to the host and port identified by the
 `authority` section of the URI. The `authority`, `path-abempty` and `query`
-portions of the URI are also transmitted in SETUP parameters (see
-{{setup-params}}). If the port is omitted in the URI, a default port of 443 is
+portions of the URI are also transmitted in Setup Options (see
+{{setup-options}}). If the port is omitted in the URI, a default port of 443 is
 used for setting up the QUIC connection.
 
 ### Connection URL
@@ -808,7 +808,7 @@ Endpoints use the exchange of Setup messages to negotiate MOQT extensions.
 Extensions can define new Message types, new Parameters, or new framing for
 Data Streams and Datagrams.
 
-The client and server MUST include all Setup Parameters {{setup-params}}
+The client and server MUST include all Setup Options {{setup-options}}
 required for the negotiated MOQT version in CLIENT_SETUP and SERVER_SETUP.
 
 Clients request the use of extensions by specifying Parameters in CLIENT_SETUP.
@@ -1065,17 +1065,18 @@ a SUBSCRIBE. A subscriber MUST send exactly one PUBLISH_OK or REQUEST_ERROR in
 response to a PUBLISH. The peer SHOULD close the session with a protocol error
 if it receives more than one.
 
-A publisher MUST save the Largest Location communicated in PUBLISH or
-SUBSCRIBE_OK when establishing a subscription. This value can be used in a
-Joining FETCH (see {{joining-fetches}}) at any time while the subscription is
-active.
-
 All `Established` subscriptions have a Forward State which is either 0 or 1.
 The publisher does not send Objects if the Forward State is 0, and does send them
 if the Forward State is 1.  The initiator of the subscription sets the initial
 Forward State in either PUBLISH or SUBSCRIBE.  The subscriber can send PUBLISH_OK
 or REQUEST_UPDATE to update the Forward State. Control messages, such as
 PUBLISH_DONE ({{message-publish-done}}) are sent regardless of the forward state.
+
+A publisher MUST save the Largest Location communicated in SUBSCRIBE_OK, PUBLISH
+or REQUEST_OK (in response to a REQUEST_UPDATE) that changes the Forward State
+from 0 to 1.  This value is called the Joining Location and can be used in a
+Joining FETCH (see {{joining-fetches}}) while the subscription is in the
+`Established` state.
 
 Either endpoint can initiate a subscription to a track without exchanging any
 prior messages other than SETUP.  Relays MUST NOT send any PUBLISH messages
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1980,6 +1980,39 @@ successfully delivered within the timeout period before sending any data
 for that Object, taking into account priorities, congestion control, and
 any other relevant information.
 
+#### RENDEZVOUS TIMEOUT Parameter {#rendezvous-timeout}
```
ianswett, line 1986: Nit: This isn't a very obvious name to me.  Something like "Lingering Timeout" might be better?  Or "Awaiting Timeout"?

draft-ietf-moq-transport.md
```diff
@@ -1980,6 +1980,39 @@ successfully delivered within the timeout period before sending any data
...
+
+The RENDEZVOUS_TIMEOUT parameter (Parameter Type 0x04) MAY appear in a
+SUBSCRIBE message.
```
ianswett: Why couldn't it occur in PUBLISH_OK as well to allow the Publisher to indicate that it's the max time it's willing to wait for something to be published?

draft-ietf-moq-transport.md
```diff
@@ -1980,6 +1980,39 @@ successfully delivered within the timeout period before sending any data
...
+
+It is the duration in milliseconds the subscriber is willing to wait for a
+publisher to become available. This applies when a relay receives a SUBSCRIBE
+for a track that has no current publisher, or when a publisher disconnects
+during an active subscription.
+
+If the RENDEZVOUS_TIMEOUT is present, the relay SHOULD hold the subscription
+and wait for a publisher to appear or reconnect, up to the specified duration.
+If a publisher becomes available within this time, the relay proceeds with
+the subscription normally. If the timeout expires without a publisher, the
```
ianswett: When does the Subscriber send a PUBLISH_OK?  I'm not sure if the relay would know the Largest Object until a publisher becomes available?

If so, then I don't think the following paragraph is necessary?

draft-ietf-moq-transport.md
```diff
@@ -1980,6 +1980,39 @@ successfully delivered within the timeout period before sending any data
...
+the Track, even when DELIVERY_TIMEOUT (see {{delivery-timeout}}) is not
+specified. Subscribers that require complete delivery of all Objects without
+gaps SHOULD NOT set RENDEZVOUS_TIMEOUT, or should be prepared to handle
+potential discontinuities.
+
+The relay or publisher MAY use a shorter timeout than requested by the
+subscriber. For example, a relay might limit the maximum rendezvous timeout
+to protect its resources.
+
+If RENDEZVOUS_TIMEOUT is absent, the relay SHOULD immediately return
```
ianswett: I'd like to make this a MUST, but maybe it's unenforceable?

ianswett (commented):
Thanks Alan, some small comments.  I'm particularly interested in when we expect the PUBLISH_OK to be sent.

draft-ietf-moq-transport.md
```diff
@@ -1980,6 +1980,39 @@ successfully delivered within the timeout period before sending any data
...
 any other relevant information.
 
+#### RENDEZVOUS TIMEOUT Parameter {#rendezvous-timeout}
+
+The RENDEZVOUS_TIMEOUT parameter (Parameter Type 0x04) MAY appear in a
+SUBSCRIBE message.
+
+It is the duration in milliseconds the subscriber is willing to wait for a
+publisher to become available. This applies when a relay receives a SUBSCRIBE
+for a track that has no current publisher, or when a publisher disconnects
```
ianswett: I think applying this param for the second case is adding edge cases I don't want to worry about, like the paragraph below.

draft-ietf-moq-transport.md
afrind: For whatever reason, I don't like "lingering" and rendezvous is a cool French word.

draft-ietf-moq-transport.md
afrind: I think the second case is a valid use case, so we will have to deal with it eventually.  The only question is if we want to control that timeout differently from the original subscription.

draft-ietf-moq-transport.md
afrind: This only applies to SUBSCRIBE, not PUBLISH.  PUBLISH is implicitly "lingering".  If your question is "when does the publisher send SUBSCRIBE_OK?" - then I think the answer is "when a publisher is known".  If your auth or other part of your request is invalid, you get REQUEST_ERROR immediately.  I would like to avoid the "100 continue" style response that says "your request was fine and I'm waiting for pubs".

draft-ietf-moq-transport.md
ianswett: I think it's valid, but I can see a few ways something like that could arise in the network, and I don't think this is necessarily the right param for this one.  I can imagine being ok waiting 5+ minutes for another Publisher to join, but only wanting an upstream transition/restart to allow for a much smaller transition time, like 5 or 10 seconds.

draft-ietf-moq-transport.md
ianswett: The current name is very non-obvious to me, so I'd prefer something else, but maybe others have better suggestions.

draft-ietf-moq-transport.md
```diff
@@ -1993,6 +1993,32 @@ successfully delivered within the timeout period before sending any data
...
+
+The relay MAY use a shorter timeout than requested by the subscriber. For
+example, a relay might limit the maximum rendezvous timeout to protect its
+resources.
+
+A value of 0 indicates the subscriber does not want to wait and expects
+an immediate response.
+
+If RENDEZVOUS_TIMEOUT is absent, the relay MUST immediately return
+REQUEST_ERROR with error code DOES_NOT_EXIST if no publisher is available.
```
vasilvv: I think you should merge those to paragraph, and just say that `0` is the default.

draft-ietf-moq-transport.md
```diff
@@ -1993,6 +1993,32 @@ successfully delivered within the timeout period before sending any data
...
+It is the duration in milliseconds the subscriber is willing to wait for a
+publisher to become available. This applies when a relay receives a SUBSCRIBE
+for a Track that has no current publisher.
+
+If the RENDEZVOUS_TIMEOUT is present, the relay SHOULD hold the subscription
+and wait for a publisher to appear, up to the specified duration. The relay
+does not send SUBSCRIBE_OK until a publisher becomes available. If a publisher
+becomes available within this time, the relay proceeds with the subscription
+normally. If the timeout expires without a publisher, the relay SHOULD respond
+with REQUEST_ERROR with error code TIMEOUT.
```
zafergurel, line 2000: The actual timeout value used by the relay may be in the error response because the relay may use a shorter timeout.

afrind: There's nowhere in the error response to convey the value, and it's more or less observable by timing the SUBSCRIBE->ERROR time at the subscriber.  I think we could massage the language a little bit though to indicate this is a subscriber preference -- e.g. this is the maximum amount of time a relay should wait.

draft-ietf-moq-transport.md
afrind: It's an homage to the French heritage of MOQ

draft-ietf-moq-transport.md
```suggestion
A value of 0 indicates the subscriber does not want to wait and expects an
immediate response.  The relay MUST immediately return REQUEST_ERROR with error
code DOES_NOT_EXIST if no publisher is available

If RENDEZVOUS_TIMEOUT is absent, the default is 0.
```

PR #1451, opened 2026-02-01
Allow multiple Subscriptions to a Track

ianswett:
This was motivated by the Filters PR #1401, though there are existing use cases, such as a relay wanting to propagate two disjoint SUBSCRIBEs from different Subscribers upstream on the same upstream Session.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1089,18 +1089,14 @@ subscriptions initiated by other endpoints, and all published Objects will be
 forwarded back to the endpoint, subject to priority and congestion response
 rules.
 
-For a given Track, an endpoint can have at most one subscription to a Track
-acting as the publisher and at most one acting as a subscriber.  If an endpoint
-receives a message attempting to establish a second subscription to a Track
-with the same role, it MUST fail that request with a `DUPLICATE_SUBSCRIPTION`
-error.
+For a given Track, an endpoint SHOULD avoid overlapping Subscriptions because
+it could be unclear which Subscription an Object is for. If a Publisher has
+multiple overlapping Subscriptions within a Session, it MAY send duplicate
+Objects.
 
-If a publisher receives a SUBSCRIBE request for a Track with an existing
-subscription in `Pending (publisher)` state, it MUST fail that request with
-a `DUPLICATE_SUBSCRIPTION` error. If a subscriber receives a PUBLISH for a Track
-with a subscription in the `Pending (Subscriber)` state, it MUST ensure the
-subscription it initiated transitions to the `Terminated` state before sending
-PUBLISH_OK.
+If a publisher receives an identical SUBSCRIBE request for a Track with an
+existing subscription in `Pending (publisher)` state, it MAY fail that request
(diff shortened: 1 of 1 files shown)


afrind:
I'm surprised there's not more to change than just this.  At the very least, we should talk about how track aliases are assigned -- I assume multiple subs to the same track receive different aliases - not sure how well that plays with systems that use hash(name) as the alias.

ianswett:
> I'm surprised there's not more to change than just this. At the very least, we should talk about how track aliases are assigned -- I assume multiple subs to the same track receive different aliases - not sure how well that plays with systems that use hash(name) as the alias.

I added some text (and fixed the link to the track-alias section) to indicate it was ok to use the same Track Alias or different ones.  If we can guarantee the Track Alias is always different, we don't need my text about overlapping subscriptions causing potential ambiguity.

draft-ietf-moq-transport.md
```diff
@@ -3250,6 +3246,8 @@ Objects can arrive after a subscription has been cancelled.  Subscribers SHOULD
 retain sufficient state to quickly discard these unwanted Objects, rather than
 treating them as belonging to an unknown Track Alias.
 
+If multiple Subscriptions are established for the same Track, the Publisher
+MAY use the same Track Alias for all of them, but it is not required to.
```
afrind, line 3481: I'm confused how to implement a subscriber in this system.

subscribe(trackName, filter-set-1) -> produces a stream of objects matching filter-set-1
subscribe(trackName, filter-set-2) -> produces a stream of objects matching filter-set-2

Since the publisher is evaluating the filters, if we allow the alias to be reused, the data plane needs a component that indicating which **subscription(s)** a given object matched.

eg: filter set 1 is "objects 0, 2, 4, 6, 8" and filter set 2, is "0, 3, 6, 9".

If using the same track alias, the subscriber will see (0,2,3,4,6,8,9) and has to re-run the filters to re-sort the objects.  This means no duplicates but 2x the filter processing.

Alternately, the publisher opens 3 streams for this subgroup: only-1, only-2, both, and the subgroup header lists the request IDs?

both: 0, 6
1: 2, 4, 8
2: 3, 9

This retains no duplicates and only 1x processing, but makes subgroup reconstruction and in-order delivery a subscriber-side nightmare.

It's significantly simpler to say that multiple subscriptions to the same track need _unique aliases_, and eat the duplicates.

alias1: 0,2,4,6,8
alias2: 0,3,6,9

0 and 6 are duplicated with only publisher-side filter processing.  Subscriber-side subgroup reconstruction is automatic.  But team "alias=hash" needs a different approach.

afrind:
Does this solve an open issue?

ianswett:
> Does this solve an open issue?

I don't want to do Object filters as described without this change, because I think it's a DoS vector.  The @suhasHere and @mzanaty thought it would be better to keep this as a separate PR for readability, and it has some potentially valuable use cases by itself.

ianswett:
Closing this and we can create a new PR after we discuss this in London.

fluffy (commented):
I think it should be very deterministic what the relay does with overlapping subscriptions. I don't want to get error from some relays and not get errors from others.

Issue #1407, opened 2025-12-21
Unknown Setup Parameters and Extension Negotiation


afrind:
The current draft states:

> Receivers ignore unrecognized parameters.

This applies to both setup and message parameters.

The current design is that a client includes parameters that represent extensions it supports and wishes to use and possibly parameters related to that extension.  ALPN is not suitable for this because of possible combinatorial explosion when there are N different extensions.

So as an example:

~~~
CLIENT_SETUP:
Parameters:
HOT_EXTENSION: 1
HOT_EXTENSION_DEFAULT_TEMP: 100
HOT_EXTENSION_MAX_TEMP: 1000

SERVER_SETUP that supports HOT_EXTENSION:
Parameters:
HOT_EXTENSION: 1
HOT_EXTENSION_MAX_TEMP: 500

SERVER_SETUP that has never heard of HOT_EXTENSION:
Parameters:
~~~

This design allows clients to speculatively add all the information necessary to setup to enable an extension without any risk of blowing up the session.

This design has been in place through all consensus calls issued in 2025, including the ALPN change in draft-15.

In #1390 discussions, some folks would like to revisit this design.

There's intersection with #1370, because allowing the control streams to be independent of one another has implications for how one negotiates extensions.

ianswett:
I can imagine wanting to require auth, but that's defined in the core draft, so by definition it can't be unknown.

If the peer doesn't support something and you need it, and it's an extension (not a new alpn), then I think the only approach is to fail the handshake after finding out the peer doesn't support whatever extension you need.

This approach has worked quite well for QUIC, so I'm not sure why MoQ would be that different?

vasilvv:
I find this complaint odd, since this is very similar to how TLS, QUIC, HTTP/2, HTTP/3, and probably a lot of other protocols work.  I don't see really good alternatives (except maybe for ALPS, which would be an improvement in some sense, but sadly ALPS draft has been inactive for years).

suhasHere:
Here is my comment from the PR #1390  ( copying here for context) , which seems to not work in terms of extensibility.

By unrecognizable and ignorable, Setup parameter must be recognized by the Peer , but it need not have to be in a given protocol version. I am feeling bit uncomfortable to allow to blindly ignore setup parameter by the peer ( in this case relay)

[...]

PR #1461, opened 2026-02-05
Rename Setup Parameters to Setup Options

afrind:
We can bikeshed on this name, but at least it isn't used anywhere else.

- Rename "Parameters" section to "Message Parameters" with clean separation
- Rename "Setup Parameters" subsection to "Setup Options"
- Use "Option Type" instead of "Parameter Type" for Setup Options
- Update wire format diagrams to use "Number of Setup Options"
- Add Setup Options serialization paragraph explaining namespace separation
- Capitalize "Setup Options" consistently throughout
- Fix typos: "funcionally" -> "functionally", "equivalient" -> "equivalent", double parenthesis "0x03))" -> "0x03)"
- Update IANA TODO to reference Setup Options and Message Parameters

Addresses part of #1407

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -790,8 +790,8 @@ application.
 The client can establish a connection to an MOQT server identified by a given
 URI by setting up a QUIC connection to the host and port identified by the
 `authority` section of the URI. The `authority`, `path-abempty` and `query`
-portions of the URI are also transmitted in SETUP parameters (see
-{{setup-params}}). If the port is omitted in the URI, a default port of 443 is
+portions of the URI are also transmitted in Setup Options (see
+{{setup-options}}). If the port is omitted in the URI, a default port of 443 is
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2108,103 +2097,110 @@ outstanding until the Largest Group increases.
...
-  Number of Parameters (vi64),
-  Setup Parameters (..) ...,
+  Number of Setup Options (vi64),
+  Setup Options (..) ...,
 }
 ~~~
 {: #moq-transport-setup-format title="MOQT Setup Messages"}
 
-The available Setup parameters are detailed in the next sections.
+Setup Options are serialized as Key-Value-Pairs {{moq-key-value-pair}}.
```
ianswett, line 2124: It feels like all this should go up above, maybe right above the Message Parameters section?

But we can move it later since that makes the diff easier.

ianswett (approved):

Issue #1463, opened 2026-02-05
Validating END_OF_GROUP bit in a subgroup header


sharmafb:
Section 10.4.2 of the specification says "The END_OF_GROUP bit (0x08) indicates that this subgroup contains the largest Object in the Group. When set to 1, the subscriber can infer the final Object in the Group when the data stream is terminated by a FIN. In this case, Objects that have the same Group ID and an Object ID larger than the last Object received on the stream do not exist."

Let's say that an endpoint sends a subgroup header with an END_OF_GROUP bit, terminates it with a FIN, but then sends a larger object on a different subgroup of the same group. How is the receiver supposed to handle this? The specification doesn't have clear language on this. Is this a PROTOCOL_VIOLATION?

afrind:
Malformed Track section has:

> An Object is received in a Group whose Object ID is larger than the final Object in the Group. The final Object in a Group is the Object with Status END_OF_GROUP or the last Object sent in a FETCH that requested the entire Group.

I guess we need to tweak that slightly to include EOG bit cases?

sharmafb:
Created PR #1464 for this.

PR #1464, opened 2026-02-05
Clarify language for malformed tracks in a subgroup with END_OF_GROUP

sharmafb:
Fixes: #1463

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -651,13 +651,13 @@ include:
    END_OF_GROUP, the last Object before a FIN in a Subgroup which has the
    END_OF_GROUP bit set, or the last Object sent in a FETCH that requested the
    entire Group.
-9. An Object is received on a Track whose Group and Object ID are larger than the
+8. An Object is received on a Track whose Group and Object ID are larger than the
    final Object in the Track.  The final Object in a Track is the Object with
    Status END_OF_TRACK or the last Object sent in a FETCH whose response indicated
    End of Track.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -649,14 +649,15 @@ include:
 7. An Object is received in a Group whose Object ID is larger than the final
    Object in the Group.  The final Object in a Group is the Object with Status
    END_OF_GROUP or the last Object sent in a FETCH that requested the entire
-   Group.
-8. An Object is received on a Track whose Group and Object ID are larger than the
+   Group or the last object before a FIN in a subgroup which has the END_OF_GROUP
+   bit set.
```
sandarsh: typo: No point 8.

draft-ietf-moq-transport.md
```diff
@@ -649,7 +649,8 @@ include:
 7. An Object is received in a Group whose Object ID is larger than the final
    Object in the Group.  The final Object in a Group is the Object with Status
    END_OF_GROUP or the last Object sent in a FETCH that requested the entire
-   Group.
+   Group or the last object before a FIN in a subgroup which has the END_OF_GROUP
```
afrind: I think we should make it:

> The final Object in a Group is the Object with Status END_OF_GROUP, the last Object before a FIN in a Subgroup which has the END_OF_GROUP bit set, or the last Object sent in a FETCH that requested the entire Group.
   Group or

sharmafb: Addressed comment.

PR #1471, opened 2026-02-08
Fix treatment of Existing Request ID

### LINKED ISSUE #1467: Inconsistency on Message Parameters
In Sec 9.2 

Each message parameter definition indicates the message types in which it can appear. If it appears in some other type of message, it MUST be ignored.

In 9.11

The receiver MUST close the session with PROTOCOL_VIOLATION if the sender specifies an invalid Existing Request ID, or if the parameters included in the REQUEST_UPDATE are invalid for the type of request being modified.

I hope 9.11 is an oversight, because I've written a lot of code assuming I can just ignore irrelevant parameters.

In theory, you could be totally accepting according to message type but super-strict within a type based on the other type that it's modifying, and then both texts are correct. But that seems really perverse.

martinduke:
Fixes #1467.

In another section of the draft, parameters not suitable for a message are ignored, not an error.

Also, if the local host has recently killed the request (eg PUBLISH_DONE or PUBLISH_NAMESPACE_CANCEL or even REQUEST_ERROR), it's silly to make it retain state to potentially send a PROTOCOL_ERROR. If it can't match an Existing Request ID, just send REQUEST_ERROR unless it happens to have the state or otherwise detect it (e.g., Existing Request ID > Request ID).

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -477,7 +477,7 @@ possible states:
 
 Whenever the publisher communicates that certain objects do not exist, this
 fact is expressed as a contiguous range of non-existent objects and
-by include extension headers indicating the group/object gaps; MOQT
+by including Properties indicating the group/object gaps; MOQT
 implementers should take that into account when selecting appropriate data
 structures.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2535,9 +2535,10 @@ REQUEST_UPDATE Message {
 
 * Existing Request ID: The Request ID of the request this message is
   updating.  This MUST match the Request ID of an existing request.  The
-  receiver MUST close the session with `PROTOCOL_VIOLATION` if the sender
-  specifies an invalid Existing Request ID, or if the parameters included
-  in the REQUEST_UPDATE are invalid for the type of request being modified.
+  receiver SHOULD close the session with `PROTOCOL_VIOLATION` if it can verify
+  that the Existing Request ID does not corresponding to an existing request
+  or one that was recently closed locally. If it cannot verify this but cannot
+  match the request, it MUST send REQUEST_ERROR.
```
```suggestion
  or one that was recently closed locally. If it cannot verify this and cannot
  match the request, it MUST send REQUEST_ERROR.
```

afrind (approved):
seems fine

afrind (changes requested):
Oops, I missed the part about allowing bogus parameters in REQUEST_UPDATE, I think that should still be a PROTOCOL_VIOLATION.

afrind:
I hijacked your PR @martinduke to restore the PROTOCOL_VIOLATION, but kept request ID changes.

draft-ietf-moq-transport.md
```diff
@@ -2551,11 +2551,14 @@ REQUEST_UPDATE Message {
...
-  specifies an invalid Existing Request ID, or if the parameters included
-  in the REQUEST_UPDATE are invalid for the type of request being modified.
-
-* Parameters: The parameters are defined in {{message-params}}.
+  receiver SHOULD close the session with `PROTOCOL_VIOLATION` if it can verify
+  that the Existing Request ID does not corresponding to an existing request
+  or one that was recently closed locally. If it cannot verify this and cannot
+  match the request, it MUST send REQUEST_ERROR.
+
+* Parameters: The parameters are defined in {{message-params}}. The receiver MUST
```
ianswett: We say this elsewhere, and there might be parameters that specify more specific errors, so I'd prefer to not add this sentence.

afrind: Let's make sure it's in Message Parameters

draft-ietf-moq-transport.md
```diff
@@ -2551,11 +2551,14 @@ REQUEST_UPDATE Message {
...
 * Existing Request ID: The Request ID of the request this message is
   updating.  This MUST match the Request ID of an existing request.  The
-  receiver MUST close the session with `PROTOCOL_VIOLATION` if the sender
-  specifies an invalid Existing Request ID, or if the parameters included
-  in the REQUEST_UPDATE are invalid for the type of request being modified.
-
-* Parameters: The parameters are defined in {{message-params}}.
+  receiver SHOULD close the session with `PROTOCOL_VIOLATION` if it can verify
+  that the Existing Request ID does not corresponding to an existing request
+  or one that was recently closed locally. If it cannot verify this and cannot
```
```suggestion
  that the Existing Request ID does not correspond to an existing request
  or one that was recently closed locally. If it cannot verify this and cannot
```

PR #1472, opened 2026-02-08
Allow zero-element namespaces

### LINKED ISSUE #1457: Track Namespaces can have zero fields now
A NAMESPACE has a track_namespace_suffix field. If it's announcing the whole namespace, the suffix will have zero elements.

Section 2.4.1 is very clear that if Number of Track Namespace Fields is zero, it's a PROTOCOL_VIOLATION. 

That's not true anymore. Either track_namespace_suffix needs an explicit exception, or the rules have to be relaxed.

martinduke:
Fixes #1457

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1271,7 +1271,7 @@ If an endpoint accepts its own PUBLISH, this behaves as self-subscription descri
 in {{subscriptions}}.
 
 A SUBSCRIBE_NAMESPACE with zero Track Namespace fields indicates the sender is
-interested in all tracks or namespaces from the receiver.
+interested in all tracks and/or namespaces from the receiver.
 
 The subscriber sends SUBSCRIBE_NAMESPACE on a new bidirectional stream and the
 publisher MUST send a single REQUEST_OK or REQUEST_ERROR as the first message on the


afrind:
Hmm, do we want to allow tracks that exist in the "global" (0 tuple namespace), or just allow them in SUB_NS?

draft-ietf-moq-transport.md
```diff
@@ -1270,6 +1270,9 @@ Namespaces under the SUBSCRIBE_NAMESPACE prefix to the endpoint that sent them.
 If an endpoint accepts its own PUBLISH, this behaves as self-subscription described
 in {{subscriptions}}.
 
+A SUBSCRIBE_NAMESPACE with zero Track Namespace fields indicates the sender is
+interested in all tracks or namespaces from the receiver.
```
```suggestion
interested in all tracks and/or namespaces from the receiver.
```

vasilvv:
My only minor annoyance with this is that it kind of makes it ambiguous whether an empty string track namespace encodes `{}` or `{""}`.

afrind:
> My only minor annoyance with this is that it kind of makes it ambiguous whether an empty string track namespace encodes `{}` or `{""}`.

Empty elements are disallowed already. Reminder for @vasilvv you need to add an empty element check to your implementation.

PR #1476, opened 2026-02-09
Delivery timeouts are both Track and Object Properties

### LINKED ISSUE #606: timeout properties got broken

afrind:
An Object Property on the first object in a subgroup overrides the Track-level value for that subgroup, for both OBJECT_DELIVERY_TIMEOUT and SUBGROUP_DELIVERY_TIMEOUT.
 
Fixes: #606

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1953,9 +1953,9 @@ DELIVERY_TIMEOUT, if present, MUST contain a value greater than 0.  If an
 endpoint receives a DELIVERY_TIMEOUT equal to 0 it MUST close the session
 with `PROTOCOL_VIOLATION`.
 
-If both the subscriber specifies this parameter and the Track has a
-DELIVERY_TIMEOUT extension, the endpoints use the min of
-the two values for the subscription.
+If both the subscriber specifies this parameter and the publisher specifies a
+DELIVERY_TIMEOUT extension (Track or Object), the endpoints use the min of
+the two values for each Object.
 
 Publishers can, at their discretion, discontinue forwarding Objects earlier than
 the negotiated DELIVERY TIMEOUT, subject to stream closure and ordering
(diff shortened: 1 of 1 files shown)


ianswett (approved):

suhasHere (approved):

afrind:
Victor asks if it's ok to go from zero to non-zero.

draft-ietf-moq-transport.md
```diff
@@ -1814,9 +1814,11 @@ are expressed in milliseconds; both are optional; a value of 0 means that
...
 
 The publisher communicates both timeout values as a Track Property; the
-subscriber communicates them as Message Parameters.  For each type of timeout,
-if both the publisher and the subscriber have a non-zero value, the smaller of
-the two is used.
+subscriber communicates them as Message Parameters.  Either timeout value can
+also be set as an Object Property on the first object in a subgroup, overriding
+the Track-level value for that subgroup.  For each type of timeout, if both the
+publisher and the subscriber have a non-zero value, the smaller of the two is
+used.
```
vasilvv: This is ambiguous now, since there are three values.  How about
```
Pub Timeout = Object Timeout if non-null, otherwise Track Property
Final TImeout = min(Pub Timeout, Sub Timeout)
```

afrind: Good point — reworded to make the precedence explicit: the publisher's value is the Object Property when present on the first object of the subgroup, and the Track Property otherwise; then if both the publisher's and subscriber's values are non-zero, the smaller is used. PTAL.

afrind: Claude went rogue and posted without my consent

vasilvv (approved):

Issue #1478, opened 2026-02-09
Some way to tell Peer that under operational load / dos


fluffy:
This is phrased poorly for place holder from Mike's session. Will suggest that need error of some time to indicate DOS.

PR #1479, opened 2026-02-09
Add EXCESSIVE_LOAD error code (0x9)

afrind:
Add a new error code to indicate the responder is overloaded, available in REQUEST_ERROR, PUBLISH_DONE, and Data Stream Reset code spaces. In REQUEST_ERROR, the sender SHOULD use the Retry Interval to indicate when the request can be retried.

Fixes: #1478

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -648,8 +648,9 @@ include:
    different final Objects.
 7. An Object is received in a Group whose Object ID is larger than the final
    Object in the Group.  The final Object in a Group is the Object with Status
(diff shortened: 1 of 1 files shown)


LPardue:
Why not call it EXCESSIVE_LOAD?

afrind:
> Why not call it EXCESSIVE_LOAD?

I'm happy to call it whatever.

LPardue:
I only mention that because it's what H3 uses. And is the less cool version of ENHANCE_YOUR_CALM

ianswett (approved):

Issue #268, opened 2023-10-09
Define and register the moqt:// URL 


fluffy:

hardie:
Do we want moq or moqt ?

wilaw:
Shorter is better when it comes to names. So I'd suggest moq://

hardie:
On Tue, Oct 10, 2023 at 5:10 PM Will Law ***@***.***> wrote:

> Shorter is better when it comes to names. So I'd suggest moq://
>
> BCP 35 (RFC 7595), section 3.8 has the guidance for selecting a scheme
name.  The reason we might prefer moqt is this:

   A scheme name is not a "protocol."  However, like a service name as
   defined in Section 5 of [RFC6335], it often identifies a particular
   protocol or application.  If a scheme name has a one-to-one
   correspondence with a service name, then the names SHOULD be the
   same.

The part of the system we're defining at this point is the transport, and
we expect there to be other parts of the overall MoQ ecosystem (the
mappings, the catalog formats, and so on).  Specifying that this refers to
the transport itself may be useful if any other part of the system would
also like to use URIs  as identifiers.

Neither of them seems likely to collide with other uses any time soon, so I
don't think we need to make the decision urgently, so it may be useful to
consider whether any other bits will want their own schemes before we
commit to this scheme name.

kixelated:
+1 to Ted. Let's use `moqt://` for the transport and save `moq://`, probably for the media layer.

fluffy:
I'm perfectly happy with either.

afrind:
Discussed in the context of MSF in Boulder: 

There's some appetite to have the moqt:// URL be able to handle either WebTransport or raw QUIC.  A client that supports both would send ALPNs like h3, moqt-17, moqt-16 in preference order.  The server chooses (though can have its own preference).  If h3 is selected, then the client issues Wt-Available-Protocols: moqt-17, moqt-16 in the CONNECT request.

Option to add "+" syntax, eg: moqt+quic and moqt+wt, which will constrain the ALPNs.  The downside risk is that if a server only supports quic and the client is vended a moqt-wt:// URL, the connection will fail with no fallback.

afrind:
Based on some feedback from Ted Hardie in the attached PR I gave this another think, and perhaps our design-on-the-fly in Boulder wasn't quite right.

Consider what if we _only_ supported the moqt:// flavor described here, and in the PR.  This instructs the client to try both WT and QUIC, if it supports both.

[...]

PR #1486, opened 2026-02-10
Unified moqt:// URI scheme for QUIC and WebTransport

afrind:
Addresses part of #268

Replace separate WebTransport (HTTPS URI) and QUIC (moqt URI) sections with a single moqt:// URI scheme that supports both transports.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -376,6 +376,9 @@ If a receiver understands a Type, and the following Value or Length/Value does
 not match the serialization defined by that Type, the receiver MUST close
 the session with error code `KEY_VALUE_FORMATTING_ERROR`.
 
+Key-Value-Pairs are always parsed with a known byte length, which bounds
+the sequence. The source of this length varies by context.
+
 ### Reason Phrase Structure {#reason-phrase}
 
 Reason Phrase provides a way for the sender to encode additional diagnostic
@@ -754,6 +757,10 @@ If a Relay supports a Property, it MAY be modified, added, removed, and/or
 cached, subject to the processing rules specified in the definition.
 
 Properties are serialized as Key-Value-Pairs (see {{moq-key-value-pair}}).
+Track Properties always appear as the final field in the messages that
+carry them; their length is the remaining bytes of the message after all
+preceding fields have been consumed. Object Properties ({{object-properties}})
+are preceded by an explicit length field.
 
 Property types are registered in the IANA table 'MOQ Properties'.
 See {{iana}}.
@@ -1118,7 +1125,7 @@ in order to obtain an accurate minimum RTT. Similarly, Reno halves it's congesti
 window upon detecting loss.  In both cases, the large reduction in sending rate might
 cause issues with latency sensitive applications.
 
-# Modularity
+# Extensibility
 
 MOQT defines all messages necessary to implement both simple publishing or
 subscribing endpoints as well as highly functional Relays.  Non-Relay endpoints
@@ -1513,7 +1520,7 @@ datagram (see {{data-streams}}).
 (groups with higher group ID are sent first).  The subscriber optionally
 communicates its group order preference in the SUBSCRIBE message; the
 publisher's preference is used if the subscriber did not express one (by
-setting Group Order field to value 0x0).  The group order of an existing
+omitting the Group Order parameter).  The group order of an existing
 subscription cannot be changed.
 
 ## Scheduling Algorithm
@@ -1974,7 +1981,8 @@ All Message Parameters MUST be defined in the negotiated version of MOQT or
 negotiated via Setup Options. An endpoint that receives an unknown Message
 Parameter MUST close the session with `PROTOCOL_VIOLATION`. Because the receiver
 has to understand every Message Parameter, there is no need for a mechanism to
-skip unknown parameters.
+skip unknown parameters. Because unknown parameters cannot be skipped, the block
+is bounded by a parameter count rather than a length.
 
 The Message Parameter types defined in this version of MOQT are listed below.
 
@@ -2341,7 +2349,8 @@ SETUP Message {
 ~~~
 {: #moq-transport-setup-format title="MOQT SETUP Message"}
 
-Setup Options are serialized as Key-Value-Pairs {{moq-key-value-pair}}.
+Setup Options are serialized as Key-Value-Pairs {{moq-key-value-pair}},
+spanning the entire message payload, bounded by the message Length field.
 Setup Options use a namespace that is constant across all MOQT versions,
 separate from Message Parameters.  Receivers MUST ignore unrecognized Setup
 Options.  Senders MUST NOT repeat the same Option Type in a message unless
@@ -2462,6 +2471,7 @@ GOAWAY Message {
   New Session URI Length (vi64),
   New Session URI (..),
   Timeout (vi64),
+  Request ID (vi64),
 }
 ~~~
 {: #moq-transport-goaway-format title="MOQT GOAWAY Message"}
@@ -2482,6 +2492,16 @@ GOAWAY Message {
   0 indicates the sender has no specific timeout, and the recipient SHOULD still
   close the session as quickly as possible. This is a hint; the sender of the
   GOAWAY MAY close the session before the indicated timeout has elapsed.
+
+* Request ID: The smallest peer Request ID that was not or might not have been
+  processed prior to sending the GOAWAY. If no requests have been processed,
+  this is 0 (at a server) or 1 (at a client). If the parity of the Request ID
+  does not match the receiver's parity, the endpoint MUST close the session with
+  `INVALID_REQUEST_ID`. Requests with a Request ID equal to or greater than the
+  indicated value, as well as any requests that arrive after the GOAWAY, MUST be
+  rejected with REQUEST_ERROR using error code GOING_AWAY. Requests with a
+  Request ID less than the indicated value were or might have been processed;
+  their status can be determined from the response on each request stream.
 
 ## REQUEST_OK {#message-request-ok}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -787,12 +779,45 @@ This protocol does not specify any semantics on the `path-abempty` and
...
+  `h3` in its TLS ClientHello. On a TCP+TLS connection, the client offers `h2`.
+  The client then establishes a WebTransport session as described in
+  {{webtransport}}.
+
+### WebTransport {#webtransport}
+
+When the client uses WebTransport, it sends an extended CONNECT request to the
+host and the path indicated by the URI, as described in
+({{WebTransport, Section 3}}). The client includes MOQT protocol identifiers in
+the WT-Available-Protocols header ({{WebTransport, Section 3.3}}).
```
LPardue, line 861: Maybe worth reminding people tbe scheme is still "https", as required in webtransport spec

MikeBishop: The client constructs an "https" URI in the following manner, then establishes a WebTransport session with the resource....

fluffy:
Probably worth having a read of https://www.rfc-editor.org/rfc/rfc7595.html and making sure we have all the details that are needed

draft-ietf-moq-transport.md
```diff
@@ -787,12 +779,62 @@ This protocol does not specify any semantics on the `path-abempty` and
...
+  WebTransport session as described in {{webtransport}}. On a TCP+TLS
+  connection, the client offers `h2` in its TLS ClientHello and establishes a
+  WebTransport session as described in {{webtransport}}.
+
+moqt+q:
+: The client MUST use native QUIC. The client offers only MOQT ALPNs
+  (e.g. `moqt/1`, `moqt/2`) in its TLS ClientHello. If ALPN negotiation
+  fails, the connection fails; the client MUST NOT fall back to WebTransport.
+
+TODO: consider fallback to QMUX
```
ianswett: Given where that is, I'd prefer to skip QMUX in this doc for now.

draft-ietf-moq-transport.md
```diff
@@ -787,12 +779,62 @@ This protocol does not specify any semantics on the `path-abempty` and
...
+: The client MUST use WebTransport. On a QUIC connection, the client offers
+  `h3` in its TLS ClientHello. On a TCP+TLS connection, the client offers `h2`.
+  The client then establishes a WebTransport session as described in
+  {{webtransport}}.
+
+### WebTransport {#webtransport}
+
+When the client uses WebTransport, it constructs an `https` URI from the `moqt`
+URI by replacing the scheme with `https` and removing any transport suffix
+(`+q` or `+wt`).  For example, `moqt+wt://example.com/path` becomes
```
ianswett: You can't use WebTransport if there's a +q suffix.

```suggestion
URI by replacing the scheme with `https`.
For example, `moqt+wt://example.com/path` becomes
```

ianswett (commented):
Looks good overall, though I'm not an expert in URI schemes.

wilaw (approved):
Thanks for implementing this. We shall use this as the basis for the MSF URL scheme and extend with a fragment definition for the track namespace and associated metadata.

draft-ietf-moq-transport.md
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -761,22 +761,14 @@ would be identified as "moqt-13".
 Note: Draft versions prior to -15 all used moq-00 ALPN, followed by version
 negotiation in the CLIENT_SETUP and SERVER_SETUP messages.
 
-### WebTransport
+### MOQT URI Scheme
```
suhasHere, line 814: Should this be MoQT instead of MOQT. Either is fine, i have see one of them more than other in the document

draft-ietf-moq-transport.md
```diff
@@ -4101,7 +4105,57 @@ TODO: fill out currently missing registries:
...
+"Uniform Resource Identifier (URI) Schemes" registry, per {{!RFC7595}}:
+
+### "moqt" URI Scheme Registration
+
+Scheme name: moqt
+
+Status: Permanent
+
+Applications/protocols that use this scheme name: Media over QUIC Transport
+(MOQT), as defined in this document.
```
```suggestion
Applications/protocols that use this scheme name: Media over QUIC Transport
(MOQT) over native QUIC or WebTransport, as defined in this document.
```

draft-ietf-moq-transport.md
```diff
@@ -787,12 +779,60 @@ This protocol does not specify any semantics on the `path-abempty` and
...
+When the client uses WebTransport, it constructs an `https` URI from the `moqt`
+URI by replacing the scheme with `https`.
+For example, `moqt+wt://example.com/path` becomes
+`https://example.com/path`. The client sends an extended CONNECT request to this
+URI to establish a WebTransport session, as described in
+({{WebTransport, Section 3}}). The client includes MOQT protocol identifiers in
+the WT-Available-Protocols header ({{WebTransport, Section 3.3}}).
+
+### Native QUIC {#native-quic}
+
```
suhasHere, line 864:
```suggestion
The client establishes a QUIC connection to the host and port identified by the
`authority` section of the URI.
```

suhasHere (approved):

draft-ietf-moq-transport.md
afrind: We decided MOQT is it a while ago.

draft-ietf-moq-transport.md
```diff
@@ -761,22 +761,14 @@ would be identified as "moqt-13".
...
-An MOQT server that is accessible via native QUIC can be identified by a
-URI with a "moqt" scheme.  The "moqt" URI scheme is defined as follows,
-using definitions from {{!RFC3986}}:
+An MOQT server is identified using a URI with the "moqt" scheme.  The "moqt"
+URI scheme is defined as follows, using definitions from {{!RFC3986}}:
 
 ~~~~~~~~~~~~~~~
-moqt-URI = "moqt" "://" authority path-abempty [ "?" query ]
+moqt-URI = "moqt" [ "+q" / "+wt" ] "://" authority
+            path-abempty [ "?" query ]
```
fluffy: I don't think we should define query to be one or more instances of either key or key=value separated with &

draft-ietf-moq-transport.md
fluffy, line 814: It seems like this section should also talk about the fragment and how it might be used

afrind: Let's land the basic scheme here that works and follow up to add fragment handling?  I'm not sure we have consensus yet that the MSF based fragment handling needs to go in MOQT?  We will open a new issue.

draft-ietf-moq-transport.md
afrind: I'm not sure I follow your comment - this is not defining query, and RFC3986 does not give any mandatory structure either.

draft-ietf-moq-transport.md
vasilvv: Fragments are not scheme-specific, thus it is my understanding is that we can't talk about that even if we want to.

vasilvv (approved):

afrind:
Here's a review I received from Ted Hardie via email:

So it looks you are registering three different URI schemes, "moqt", which you expect to handle either of the current two transports, and then a scheme for each specific transport , "moqt+wt", and "moqt+q".  This is not best practice, as you can see in the reviewer notes for similar approaches ([https://www.iana.org/assignments/uri-schemes/expert-notes/coap-expert-notes](https://urldefense.com/v3/__https://www.iana.org/assignments/uri-schemes/expert-notes/coap-expert-notes__;!!Bt8RZUm9aw!_5-hIbdMPJadMK_ZcDZ0oCOeRyYFaTLD2qRIAUQIYoKkIzr5oWH52B0EYTpqXT7QYZfy3_9OZ75G$) is one).  You may also want to review [https://datatracker.ietf.org/doc/html/rfc8323#section-8](https://urldefense.com/v3/__https://datatracker.ietf.org/doc/html/rfc8323*section-8__;Iw!!Bt8RZUm9aw!_5-hIbdMPJadMK_ZcDZ0oCOeRyYFaTLD2qRIAUQIYoKkIzr5oWH52B0EYTpqXT7QYZfy37wv3L_C$), which describes why CoAP went that way; in essence, they were willing to see the distinct URI schemes as referencing fully distinct origins/resources.  Is that what you expect to?  Or do you expect that 

moqt://Alans_authority/some_resource
moqt+wt://Alans_authority/some_resource 

will reference the same resource since the moqt URI scheme implies support of web transport?  If so, it may not work as you expect by default.  Note that using +something after a scheme name has no semantics; the default comparison is by full scheme name, and going this way expects URI parsers to special case this. 

One way around this is to push that information out of the scheme.  It could have been encoded in a port had WS and QUIC chosen unique ports, but that ship has sort of sailed.  Another way to do it would be to make the query section mandatory, and specify that it must contain a parameter (from Cullen's comment, likely a key value pair) identifying the transports.  So

moqt://Alans_authority/some_resource?transport=wt; otherkey=othervalue; 

for a resource at Alans_authority, which supports only webtransport, and for which there is another parameter represented by a key/value pair.

In general, this looks a bit underspecified, because you just adopted the core syntax and didn't modify it.  But if you want any semantics beyond identifying resources, I think you need a bit more to explain syntactically how those semantics (like parameters) would be expressed.

draft-ietf-moq-transport.md
fluffy: Right - I am think we have learned some things since 3986. We it clear at this point it is good to be able to support more than one query param and best not to rely on vague unspecified conventions as a way to do it. Defining it as  [ "?" query ] does not provide that and thus I think it would be better to define this from the start to support multiple key values pairs 

Something like 

query= "?" kvp *("&" kvp)
kvp = name [ "=" value ]

draft-ietf-moq-transport.md
fluffy: I agree they are not scheme specific, but that does not mean we should not talk about them, just not as part of the URI. How a an applications that use the resource the URI refers to and how the fragment interacts with that seems fine for a spec to talk about.

afrind: See #1571 which is targeted on top of this PR to address fragments

afrind:
Discussed at 125 - vibe was to remove +q +wt options, which has been done.  I moved fragment discussion to #1571.  I made it clear that path and query need to be convertible to https in case HTTP is negotiated.

I've got the necessary approvals to merge here, can someone give another re-review?

Issue #1023, opened 2025-05-12
Subgroups and DELIVERY_TIMEOUT can result in pathological FETCH


afrind:
Consider this setup:

Publisher -> Ingest Relay -> Egress Relay -> Subscribers

The Publisher has one track, with two subgroups: base (id=0) and enhancement (id=1).  The base layer is higher priority.  The object IDs are in decode order, such that sg=0 has even objects and sg=1 has odd objects.  There is a DELIVERY_TIMEOUT between the ingest relay and the Publisher for this track.

During congestion, there is sufficient bandwidth such that sg=0 makes it to the ingest relay, but sg=1 gets reset.

As a result, the egress relay cache contains only even objects for this track (0, 2, 4, ...)

A new subscriber joins this track and issues a Joining FETCH for this track.

This causes the relay to make a FETCH request to the original publisher either:

1) one request for every odd object
2) a request for the entire group up to the current object

As #945 describes, the original publisher may not even have sufficient bandwidth to handle any FETCH requests in parallel with current live data.  If the relay can't get object 1, it can't serve object 2 either.

I think the fundamental problem is the interleaving of mixed priority/reliability data in the FETCH response (eg: does the Joining FETCH need all subgroups in order to start playing?).

Johan511:
>  interleaving of mixed priority/reliability data in the FETCH response
In this case won't the problem be solved by having the base and enhancement layers in different tracks and all objects in the track will have the same publisher priority?

Also, at some point, didn't MOQ mention that the same object should not be sent more than once over a given connection?
If we had that property, it would you would simply need another fetch message instead of "one request for every odd object".

afrind:
Brainstorming a few options:

1. Do nothing - applications with this structure and usage pattern will cause upstream roundtrips and higher join latency.

2. Recommend different usage - applications like this can use multiple tracks instead of multiple subgroups - keeping the even object IDs in one track and odds in the other, and include object ID gap markers.  Then there are 2 independent joining fetches, one that succeeds immediately and one that may have to go back to origin.

3. Add a different kind of FETCH - maybe joining FETCH can filter only for subgroup=0?  Caches need some additional logic to record if two objects in the subgroup with an ID gap came on the same stream with no interruption.

4. Disallow caching when DELIVERY_TIMEOUT is present (suggested by @mzanaty).  This means a Joining FETCH will go back to the original publisher.

[...]

PR #1490, opened 2026-02-11
FILL_TIMEOUT parameter

afrind:
Specifies the subscriber's max wait to fill a gap in a FETCH range before getting Unknown.

Addresses part of: #1023


draft-ietf-moq-transport.md
```diff
@@ -1993,6 +1993,29 @@ successfully delivered within the timeout period before sending any data
...
+
+The FILL_TIMEOUT parameter (Parameter Type 0x0A) MAY appear in a FETCH message.
+
+It is the duration in milliseconds a relay SHOULD wait for an upstream source to
+provide Objects that are not immediately available before reporting them as
+Unknown gaps in the FETCH response. When a relay encounters a range of Objects
+within the requested range that are not immediately available and have unknown
+status, it issues an upstream FETCH to retrieve them. The Fill Timeout applies
+independently to each such upstream FETCH, not an overall timeout for the
+entire request. If an upstream FETCH does not complete within the Fill Timeout,
```
suhasHere: Fill timeout should not be individualized to relay upstream fetches , as it adds more time and expectation for the end subscribers. The API intent from end subscriber is to say i want my fetch to find answers in 2 seconds, if not return with gaps.

draft-ietf-moq-transport.md
```diff
@@ -1993,6 +1993,29 @@ successfully delivered within the timeout period before sending any data
...
+the relay reports the corresponding Objects as gaps and continues processing
+subsequent Objects in the range.
+
+A value of 0 indicates the subscriber only wants Objects that are immediately
+available; the relay MUST NOT wait for upstream delivery and MUST report any
+unavailable Objects as gaps.
```
suhasHere: Shouldn't in this case, the unavailable Objects  are actually Unknown status , since it is skip behavior

sharmafb (approved):

draft-ietf-moq-transport.md
ianswett: I think it's easier to reason about the behavior if it's per upstream FETCH.  If a Subscriber wants to have a deadline for the entire FETCH, it can set a timer locally.

draft-ietf-moq-transport.md
```diff
@@ -1993,6 +1993,29 @@ successfully delivered within the timeout period before sending any data
 for that Object, taking into account priorities, congestion control, and
 any other relevant information.
 
+#### FILL TIMEOUT Parameter {#fill-timeout}
+
+The FILL_TIMEOUT parameter (Parameter Type 0x0A) MAY appear in a FETCH message.
+
+It is the duration in milliseconds a relay SHOULD wait for an upstream source to
```
```suggestion
It is the maximum duration in milliseconds a relay SHOULD wait for an upstream source to
```

draft-ietf-moq-transport.md
```diff
@@ -1993,6 +1993,29 @@ successfully delivered within the timeout period before sending any data
...
+provide Objects that are not immediately available before reporting them as
+Unknown gaps in the FETCH response. When a relay encounters a range of Objects
+within the requested range that are not immediately available and have unknown
+status, it issues an upstream FETCH to retrieve them. The Fill Timeout applies
+independently to each such upstream FETCH, not an overall timeout for the
+entire request. If an upstream FETCH does not complete within the Fill Timeout,
+the relay reports the corresponding Objects as gaps and continues processing
```
```suggestion
the relay reports the corresponding Objects as Unknown gaps and continues delivering
```

draft-ietf-moq-transport.md
```suggestion
available; the relay MUST report any unavailable Objects as Unknown gaps.
```

draft-ietf-moq-transport.md
```diff
@@ -1993,6 +1993,29 @@ successfully delivered within the timeout period before sending any data
...
+subsequent Objects in the range.
+
+A value of 0 indicates the subscriber only wants Objects that are immediately
+available; the relay MUST NOT wait for upstream delivery and MUST report any
+unavailable Objects as gaps.
+
+If the Fill Timeout parameter is absent, the relay waits for an implementation
+specific duration before reporting gaps. If the subscriber specifies a Fill
```
```suggestion
specific duration before reporting Unknown gaps. If the subscriber specifies a Fill
```

ianswett (commented):
Gaps in FETCH indicate DNE, so I think we want to be clear that this is about Unknown gaps.

draft-ietf-moq-transport.md
suhasHere: I am fine with it. However, my concern is api semantics here. As a end subscriber, if i am saying my fill_timeout is 2 seconds , then i would adjust my playout decisions based on 2 seconds. If the relay splits my fetch into 10 upstream requests and in worst case, has all my answers in 20 seconds, it is of no use for my player ( as it is 18s late )

vasilvv (changes requested):
I don't understand the text of this PR.  I thought the reason FILL_TIMEOUT exists was to let the relay to stall a FETCH until an object arrives via upstream SUBSCRIBE; the text talks about upstream FETCH, though, so I have no idea what happens here.

afrind:
> I don't understand the text of this PR. I thought the reason FILL_TIMEOUT exists was to let the relay to stall a FETCH until an object arrives via upstream SUBSCRIBE; the text talks about upstream FETCH, though, so I have no idea what happens here.

Interesting.  That is not what I intended this PR to mean.  That sounds like a different behavior that should be controlled separately if we want it:

This PR: when sending a FETCH upstream, how long to wait before giving up and getting "unknown".  If 0, don't even attempt.
Your Idea: when encountering any gap in the cache, how long to wait before _attempting_ to FETCH upstream.

afrind:
Discussed 3/30 -- the timeout should be a total fill budget.

martinduke:
Individual comment:

I would like to park this until REWIND gets worked out. I would like to preserve the pure semantics of FETCH if REWIND is available for delay-sensitive information, and I definitely don't want two ways to do the same thing.

afrind:
I think this has utility independent of rewind?  Certainly the relay/cache will have some timeout for upstream fills, this is just giving the subscriber some control over it.

afrind:
@vasilvv : Not sure if you still have questions about how this interacts with subscribe.  Without rewind, a SUBSCRIBE has no need of FILL_TIMEOUT.  Objects are delivered to subscribers as they arrive at the relay, not retrieved by FETCH.

draft-ietf-moq-transport.md
```diff
@@ -2256,7 +2256,34 @@ successfully delivered within the timeout period before sending any data
 for that Object, taking into account priorities, congestion control, and
 any other relevant information.
 
+<<<<<<< fill-timeout
```
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -2256,7 +2256,34 @@ successfully delivered within the timeout period before sending any data
...
+
+If the Fill Timeout parameter is absent, the relay waits for an implementation
+specific duration before reporting Unknown gaps. If the subscriber specifies a Fill
+Timeout larger than the relay is willing to wait, the relay MAY use a shorter
+timeout without informing the subscriber.
+
+#### SUBSCRIBER PRIORITY Parameter {#subscriber-priority}
+=======
 ### RENDEZVOUS TIMEOUT Parameter {#rendezvous-timeout}
+>>>>>>> main
```
ianswett: ?

draft-ietf-moq-transport.md
```diff
@@ -2256,7 +2256,34 @@ successfully delivered within the timeout period before sending any data
...
+this request. If the budget is exhausted, the relay reports any remaining
+unavailable Objects as Unknown gaps and continues delivering available Objects
+in the range.
+
+A value of 0 indicates the subscriber only wants Objects that are immediately
+available; the relay MUST NOT wait for upstream delivery and MUST report any
+unavailable Objects as Unknown gaps.
+
+If the Fill Timeout parameter is absent, the relay waits for an implementation
+specific duration before reporting Unknown gaps. If the subscriber specifies a Fill
```
ianswett, line 2277: If the parameter is absent, I'd expect the publisher to no have unknown gaps and instead fail the fetch partway through with a timeout?  But maybe that's not what we want?

ianswett (approved):
One small comment and you still have some merge conflict leftovers

draft-ietf-moq-transport.md
suhasHere: I tend to agree with Ian. Not specifying the fill timeout parameter is an indication to do your best and return me the results.

draft-ietf-moq-transport.md
```diff
@@ -2256,6 +2256,29 @@ successfully delivered within the timeout period before sending any data
 for that Object, taking into account priorities, congestion control, and
 any other relevant information.
 
+### FILL TIMEOUT Parameter {#fill-timeout}
+
+The FILL_TIMEOUT parameter (Parameter Type 0x0A) MAY appear in a FETCH message.
+
+It is the maximum total duration in milliseconds a relay SHOULD spend waiting
```
suhasHere, line 2262: I wonder if the duration should be in seconds as it makes more sense for fetch requests

suhasHere (approved):
Minor nits and clarification on missing timeout parameter

draft-ietf-moq-transport.md
afrind: I think subsecond fill could make sense in some scenarios.  I don't think we care about saving 1-2 bytes in this varint control msg.

draft-ietf-moq-transport.md
afrind: Are you saying the same thing?  To me it sounds like Ian wants:

Ian: No Fill Timeout -> *error* after relay internal upstream timeout
Suhas: No Fill Timeout -> *unknown* after relay internal upstream timeout [current text]

I tend to think Unknown is better.  The subscriber can cancel if it sees an unknown and didn't want one.

ianswett: I'm fine with Unknown, based on our conversation.

vasilvv (approved):
I am extremely skeptical of the internal timeout business in this PR, but overall this looks like a step in the right direction.

Issue #1428, opened 2026-01-22
What happens when a client subscribes to a namespace that is too large?


vasilvv:
I imagine that at some point the namespace can reach the number of active publishers (1000? 10000? 100000?) where it is too large for the relay network to handle.  We should have some story for what happens in that scenario.

suhasHere:
I really need to see a case where we have numbers in that range.  Also nothing stops from a subscribing  to 100000 tracks. 

Having said that, we need to be able to have a way to subscribe namespace to be rejected if the resource consumption issues arise. I would expect the same thing to happen on individual subscription too.

May we adding this in securiy consideration sections is good and perhaps add a error code in SubscribeError/SubscribeNamespace bidi response ?

afrind:
We don't have much in our security section, but we have this:

> A Relay SHOULD have mechanisms to prevent malicious endpoints from flooding it with PUBLISH_NAMESPACE or SUBSCRIBE_NAMESPACE requests that could bloat data structures. It could use the advertised MAX_REQUEST_ID to limit the number of such requests, or could have application-specific policies that can reject incoming PUBLISH_NAMESPACE or SUBSCRIBE_NAMESPACE requests that cause the state maintenance for the session to be excessive.

> A Relay can use authorization rules in order to prevent subscriptions closer to the root of a large prefix tree. Otherwise, if an entity sends a relay a SUBSCRIBE_NAMESPACE message with a short prefix, it can cause the relay to send a large volume of PUBLISH_NAMESPACE messages. As churn continues in the tree of prefixes, the relay would have to continue to send PUBLISH_NAMESPACE/PUBLISH_NAMESPACE_DONE messages to the entity that had sent the SUBSCRIBE_NAMESPACE.

So to some degree, it's expected that this be handled by auth.

Let's assume that there are 100k tracks published under one namespace and a client issues SUBSCRIBE_NS with publish option.  This will likely run out of Request IDs unless the client has very generous limits.  Perhaps the relay also needs a way to protect itself from opening too many (eg: client has MAX_REQUEST_ID 100000), but the relay won't do that.  See also #1262.

afrind:
This may be a duplicate - @vasilvv to investigate.

martinduke:
Elsewhere I proposed that we have a TOO_FAR_BEHIND error. Namespaces that the relay already has will be in a tree structure. If the whole tree cannot fit in the initial bidi stream credit, I believe it will be difficult to keep track of that state.

[...]

PR #1496, opened 2026-02-12
Add NAMESPACE_TOO_LARGE error and stream reset for large namespaces

afrind:
Add NAMESPACE_TOO_LARGE (0x31) REQUEST_ERROR code so publishers can reject a SUBSCRIBE_NAMESPACE when the prefix matches more publishers than the publisher is willing to enumerate.

Allow publishers to reset the SUBSCRIBE_NAMESPACE response stream when blocked by flow control, so subscribers that fall too far behind can be terminated gracefully.

Fixes: #1428

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -124,7 +124,7 @@ rapidly detect and respond to congestion.
 
 The parallel nature of QUIC streams can provide improvements in the face
 of loss. A goal of MOQT is to design a streaming protocol to leverage
-the transmission benefits afforded by parallel QUIC streams as well
+the transmission benefits afforded by parallel QUIC streams as well as
 exercising options for flexible loss recovery.
 
 ### Convergence
@@ -259,7 +259,7 @@ after the first 0 and subsequent bytes, if any, represent the integer value,
 encoded in network byte order.
 
 Integers are encoded in 1, 2, 3, 4, 5, 6, 8, or 9 bytes and can encode up to 64
-bit unsinged integers. The following table summarizes the encoding properties.
+bit unsigned integers. The following table summarizes the encoding properties.
 
 |--------------|----------------|-------------|------------------------|
 | Leading Bits | Length (bytes) | Usable Bits | Range                  |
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2435,6 +2435,10 @@ PREFIX_OVERLAP:
 : In response to SUBSCRIBE_NAMESPACE, the namespace prefix overlaps with another
 SUBSCRIBE_NAMESPACE in the same session.
 
+NAMESPACE_TOO_LARGE:
+: In response to SUBSCRIBE_NAMESPACE, the namespace prefix matches more
+publishers than the publisher is willing to enumerate.
```
sharmafb: More publishers than the RELAY is willing to enumerate?

```suggestion
publishers than the relay is willing to enumerate.
```

draft-ietf-moq-transport.md
```diff
@@ -2435,6 +2435,10 @@ PREFIX_OVERLAP:
 : In response to SUBSCRIBE_NAMESPACE, the namespace prefix overlaps with another
 SUBSCRIBE_NAMESPACE in the same session.
 
+NAMESPACE_TOO_LARGE:
+: In response to SUBSCRIBE_NAMESPACE, the namespace prefix matches more
+publishers than the relay is willing to enumerate.
```
suhasHere, line 2489: I am not sure this is helpful. Since when the subscribe namespace is received, the relay has no clear idea on number of publishers that match the prefix.

May be the better option is
- Setup a upper limit as a parameter in subscribe namespace and react based on it
- Dynamically, send NamespaceDone with error code too large after hitting a negotiated upper limit or something/

suhasHere (changes requested):

draft-ietf-moq-transport.md
afrind: I don't see harm in having this code when the relay knows up front it's too large.  To handle the dynamic case, you can reset with TOO_FAR_BEHIND as well.

If you want negotiation to control relay behavior, make a new PR?

suhasHere: how does the relay know upfront ?

suhasHere: we discussed and having this will be nice to have relay send it in cases where it has more context at the point of SubscribeNamespace Request

PR #1505, opened 2026-02-13
Fix Request ID validation for SUBSCRIBE_NAMESPACE

### LINKED ISSUE #1459: Request ID validation is broken in draft-16
In Section 9.1 it says

The client's Request ID starts at 0 and are even and the server's Request ID starts at 1 and are odd. The Request ID increments by 2 with each FETCH, SUBSCRIBE, REQUEST_UPDATE, SUBSCRIBE_NAMESPACE, PUBLISH, PUBLISH_NAMESPACE or TRACK_STATUS request. Other messages with a Request ID field reference the Request ID of another message for correlation. If an endpoint receives a Request ID that is not valid for the peer, or a new request with a Request ID that is not the next in sequence or exceeds the received MAX_REQUEST_ID, it MUST close the session with INVALID_REQUEST_ID.

But now there's no way to verify the order in which request IDs arrive, because SUBSCRIBE_NAMESPACE is on a different stream. The simplest fix is probably to just let the request ID in SUB_NS to be literally anything, and it's those on the control stream that have to be in order.

ianswett:
I tried to avoid changing the wire image, so I kept an 'Unused ID' varint in the SUBSCRIBE_NAMESPACE message.

Fixes #1459

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1771,13 +1771,13 @@ ongoing requests, and supports the endpoint's ability to limit the concurrency
 and frequency of requests.  Request IDs for one endpoint increment independently
 from those sent by the peer endpoint.  The client's Request ID starts at 0 and
 are even and the server's Request ID starts at 1 and are odd.  The Request ID
-increments by 2 with each FETCH, SUBSCRIBE, REQUEST_UPDATE,
-SUBSCRIBE_NAMESPACE, PUBLISH, PUBLISH_NAMESPACE or TRACK_STATUS request.
+increments by 2 with each FETCH, SUBSCRIBE, REQUEST_UPDATE, PUBLISH,
+PUBLISH_NAMESPACE or TRACK_STATUS request.
 Other messages with a Request ID field reference the Request ID of another
 message for correlation. If an endpoint receives a Request ID that is not valid
-for the peer, or a new request with a Request ID that is not the next in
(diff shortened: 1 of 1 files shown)


ianswett:
Good point.  Is the better fix to just say that the Request IDs need to be monotonically increasing?

draft-ietf-moq-transport.md
```diff
@@ -3176,7 +3176,7 @@ SUBSCRIBE_NAMESPACE Message {
 ~~~
 {: #moq-transport-subscribe-ns-format title="MOQT SUBSCRIBE_NAMESPACE Message"}
 
-* Request ID: See {{request-id}}.
+* Unused ID: Variable length integer unused in this draft.
```
suhasHere, line 3179: i am ok to remove it altogether than using unused id.

afrind: I think we're going to end up adding it back with bidi streams anyway, so we can express ordering.  Let's leave the ID on the wire in the request (doesn't need to be in the response).

suhasHere: +1 , let's not change this and keep it as-is for now

afrind:
The simple fix here is leave the sender side as is (start at 0/1, increment by 2) and relax the receiver side, e.g.: 

> because SUBSCRIBE_NAMESPACE can be received out of order with other control messages, receivers cannot assert that control messages always increment by exactly 2.

PR #1509, opened 2026-02-16
Add support for mandatory to understand track extensions

### LINKED ISSUE #1480: Have a range of mandatory to implement track extensions
Allocate a set space of 2 byte or larger track extensions that if a relay receives a track with this extension, and the relay does not understand the extension, it will not forward the track.

suhasHere:
Fixes #1480

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -798,8 +798,14 @@ understand, it MUST NOT process or forward that track:
   error code UNSUPPORTED_EXTENSION to the downstream fetch requester.  If the
   relay has already forwarded data on a fetch stream, it MUST reset the stream.
 
-A publisher SHOULD NOT send a Mandatory Track Property to a subscriber that is
-known not to support it.
+A publisher that knows a subscriber does not support a Mandatory Track Property
+SHOULD take the following action:
+
+* For SUBSCRIBE: respond with REQUEST_ERROR with error code UNSUPPORTED_EXTENSION.
+
+* For FETCH: respond with REQUEST_ERROR with error code UNSUPPORTED_EXTENSION.
+
+* For PUBLISH: do not publish the track to that subscriber.
 
 # Sessions {#session}
 


draft-ietf-moq-transport.md
```diff
@@ -728,6 +729,36 @@ Properties are serialized as Key-Value-Pairs (see {{moq-key-value-pair}}).
 Property types are registered in the IANA table 'MOQ Properties'.
 See {{iana}}.
 
+### Mandatory Track Properties {#mandatory-track-properties}
+
+Property types in the range 0x40-0x7F are designated as Mandatory Track
```
afrind: That's a huge range - it's half of the 1 byte space, and includes application reserved (the last 8).  Since these are sent with the Track and not the Object - let's put these all in 3+ byte land?

draft-ietf-moq-transport.md
```diff
@@ -728,6 +729,36 @@ Properties are serialized as Key-Value-Pairs (see {{moq-key-value-pair}}).
...
+Properties. These properties have special handling rules that differ from other
+Properties.
+
+When a relay receives Track Properties (in PUBLISH, SUBSCRIBE_OK, or FETCH_OK
+messages) containing a Mandatory Track Property type that it does not
+understand, the relay MUST NOT forward that track. Instead, the relay SHOULD:
+
+1. For PUBLISH messages: respond with PUBLISH_DONE containing an appropriate
```
afrind: I think we should describe the upstream behavior which applies to all subscribers -- send REQUEST_ERROR.  For downstream behavior (relays) - probably we don't want to even send the PUBLISH, rather than PUBLISH+PUBLISH_DONE?

draft-ietf-moq-transport.md
```diff
@@ -728,6 +729,36 @@ Properties are serialized as Key-Value-Pairs (see {{moq-key-value-pair}}).
 Property types are registered in the IANA table 'MOQ Properties'.
 See {{iana}}.
 
+### Mandatory Track Properties {#mandatory-track-properties}
+
+Property types in the range 0x40-0x7F are designated as Mandatory Track
+Properties. These properties have special handling rules that differ from other
+Properties.
+
+When a relay receives Track Properties (in PUBLISH, SUBSCRIBE_OK, or FETCH_OK
```
afrind: Why is this only for relays and not all subscribers?

draft-ietf-moq-transport.md
```diff
@@ -728,6 +729,36 @@ Properties are serialized as Key-Value-Pairs (see {{moq-key-value-pair}}).
...
+messages) containing a Mandatory Track Property type that it does not
+understand, the relay MUST NOT forward that track. Instead, the relay SHOULD:
+
+1. For PUBLISH messages: respond with PUBLISH_DONE containing an appropriate
+   error code indicating the property is not supported.
+
+2. For SUBSCRIBE_OK messages: not forward the SUBSCRIBE_OK to downstream
+   subscribers. The relay MAY send a REQUEST_ERROR with code NOT_SUPPORTED (0x3)
+   to the original subscriber.
```
afrind: Same comment - what's the upstream behavior?  Your only recourse is to UNSUBSCRIBE (or 1389 cancel the stream).  For relays, there may be a pending downstream SUBSCRIBE that can receive a REQUEST_ERROR.  NOT_SUPPORTED may not be the best error code, might need a new one "UNSUPPORTED_EXTENSION" or something.

draft-ietf-moq-transport.md
```diff
@@ -728,6 +729,36 @@ Properties are serialized as Key-Value-Pairs (see {{moq-key-value-pair}}).
...
+
+3. For FETCH_OK messages: not forward the fetch response. The relay MAY respond
+   with REQUEST_ERROR with code NOT_SUPPORTED (0x3) to the fetch requester.
```
afrind: Same, cancel upstream fetch.

The fetch data stream may already be open with objects sent (though probably not) but we need to specify both REQUEST_ERROR and reset behavior.

draft-ietf-moq-transport.md
```diff
@@ -728,6 +729,36 @@ Properties are serialized as Key-Value-Pairs (see {{moq-key-value-pair}}).
...
+
+This behavior ensures that publishers can rely on mandatory properties being
+understood by all relays in the forwarding path. If negotiation of property
```
afrind: > can rely on ...

This doesn't really mean it can be relied on, and in fact the publisher may never know that there's a relay in the chain the doesn't support.  Pub -> R1 -> R2 -> Sub -- if R2 doesn't support the extension, the publisher probably will never know.

So it's really more about preventing tracks with mandatory extensions from being mishandled.

draft-ietf-moq-transport.md
```diff
@@ -728,6 +729,36 @@ Properties are serialized as Key-Value-Pairs (see {{moq-key-value-pair}}).
...
+support is not possible end-to-end, this range provides a fail-safe mechanism
+where unknown mandatory properties cause tracks to not be forwarded rather than
+being forwarded incorrectly.
+
+Mandatory Track Properties only apply to Track scope properties. Object scope
+properties do not have a mandatory range. Implementations MUST NOT register
+types in the range 0x40-0x7F for Object Properties.
```
afrind: I think right there there's a single space?  You get a code point that's either track, object or both.  We can make unique spaces, but that's a different change.

For now, just say that any MTI code point needs to be Track only.

afrind (approved):

draft-ietf-moq-transport.md
```diff
@@ -769,6 +770,34 @@ applications using the same code point in these ranges may assign different
...
+  (see {{request-cancellation}}).  If the subscriber is a relay with pending
+  downstream subscribers, it MUST send REQUEST_ERROR with error code
+  UNSUPPORTED_EXTENSION to the downstream subscribers.
+
+* For FETCH_OK messages: the subscriber MUST cancel the fetch
+  (see {{request-cancellation}}).  If the subscriber is a relay and has not yet
+  sent a FETCH_OK or REQUEST_ERROR downstream, it MUST send REQUEST_ERROR with
+  error code UNSUPPORTED_EXTENSION to the downstream fetch requester.  If the
+  relay has already forwarded data on a fetch stream, it MUST reset the stream.
+
```
afrind, line 800:
```suggestion

A publisher SHOULD NOT send a Mandatory Track Property to a subscriber that is
known not to support it.

```

draft-ietf-moq-transport.md
```diff
@@ -769,6 +770,34 @@ applications using the same code point in these ranges may assign different
 meanings; the interpretation depends on the track or application
 context known to the publisher and subscriber.
 
+### Mandatory Track Properties {#mandatory-track-properties}
+
+Property types in the range 0x4000-0x40FF are designated as Mandatory Track
```
afrind: Don't carve the range, instead, add a MANDATORY_PROPERTIES property, that lists the MTI properties.

draft-ietf-moq-transport.md
```diff
@@ -2528,6 +2557,10 @@ EXCESSIVE_LOAD:
 : The responder is overloaded and cannot process the request at this time. The
 sender SHOULD use the Retry Interval to indicate when the request can be retried.
 
+UNSUPPORTED_EXTENSION:
```
afrind, line 2569: This should also be a reset stream code.

draft-ietf-moq-transport.md
afrind: Or not, but make the range much larger.

draft-ietf-moq-transport.md
ianswett: #1581

draft-ietf-moq-transport.md
```diff
@@ -769,6 +770,37 @@ applications using the same code point in these ranges may assign different
...
+  UNSUPPORTED_EXTENSION to the downstream subscribers.
+
+* For FETCH_OK messages: the subscriber MUST cancel the fetch
+  (see {{request-cancellation}}).  If the subscriber is a relay and has not yet
+  sent a FETCH_OK or REQUEST_ERROR downstream, it MUST send REQUEST_ERROR with
+  error code UNSUPPORTED_EXTENSION to the downstream fetch requester.  If the
+  relay has already forwarded data on a fetch stream, it MUST reset the stream.
+
+A publisher SHOULD NOT send a Mandatory Track Property to a subscriber that is
+known not to support it.
```
ianswett: This could be read that the publisher should removed the mandatory track property, but I don't think that's the intent.


```suggestion
A publisher SHOULD NOT allow Subscriptions or Fetches for Tracks with a
Mandatory Track Property that a subscriber is known not to support.
```

PR #1512, opened 2026-02-17
Constrain encoding/parsing of track namespace and names to be more strict 

### LINKED ISSUE #1501: Track name parsing
We define a format for serializing the track namespace, and I can easily implement that, but there are some questions I have about the reverse tranformation:
- The document specifies using lower hex (.e.g `.2e`) to encode the `.` chara.  Would `.2E` be valid?
- If a symbol is allowed as is (e.g. `a`), should I accept the hex representation of it (e.g. `.61` for `a`)?

I can see arguments either way, though I think if we ban both of those, our encoding is 1:1.

suhasHere:
Fixes #1501

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -414,6 +414,30 @@ The goal of this format is to have a format that is both filename and
 URL safe. It allows many common names to be rendered in an easily human
 readable form while still supporting binary values.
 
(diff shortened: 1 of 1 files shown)


afrind:
Oops, I also wrote #1514.  We can pick one

ianswett (approved):

PR #1514, opened 2026-02-17
Clarify track name text format parsing rules

afrind:
Add text specifying that when parsing the recommended text format for track names, implementations must accept both upper and lowercase hex digits and must accept hex-encoded bytes even when they could be represented literally. This makes encoding canonical but parsing permissive.

Fixes: #1501

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -422,6 +422,12 @@ example.2enet-team2-project_x--report
   Track name: report
 ~~~
 
+When parsing this format, implementations MUST accept both upper and lowercase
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -422,6 +422,12 @@ example.2enet-team2-project_x--report
   Track name: report
 ~~~
 
+When parsing this format, implementations MUST accept both upper and lowercase
+hex digits (e.g., `.2E` and `.2e` are equivalent). Implementations MUST also
```
sharmafb, line 426: I think this conflicts with #1512

PR #1518, opened 2026-03-02
Filters with reduced scope, no location or group filter

mzanaty:
Rename existing subscription filter to subscription location filter to disambiguate from new range filters and track filter.
Add range filters for subgroup ID, object ID, priority, and property under subscribe section.
Add track filter under subscribe namespace section.
Add setup options and parameters for new filters.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1596,6 +1596,17 @@ Objects received with a location lower than the largest received in that track
 are ignored for track filter state updates; however, if the track is already
 selected, they pass the filter.
 
+### Relay Resource Protection in Large Namespaces
+
+Relays SHOULD aggregate and propagate filters upstream on subscriptions,
+especially namespace subscriptions including track selection filters,
+to conserve and protect their resources from excessive load.  They MAY
+also impose limits on the number of publishers in a namespace, by rejecting
+or closing namespace subscriptions with the error NAMESPACE_TOO_LARGE, or
+CONFLICTING_FILTERS if a track selection filter has non-uniform parameter
+values across a large number of subscribers, or PREFIX_OVERLAP if different
+subscribers force an aggregated upstream subscription to overlap.
+
 # Priorities {#priorities}
 
 MoQ priorities allow a subscriber and original publisher to influence
@@ -2791,6 +2802,11 @@ publishers than the relay is willing to enumerate.
 INVALID_JOINING_REQUEST_ID:
 : In response to a Joining FETCH, the referenced Request ID is not an
 `Established` Subscription.
+
+CONFLICTING_FILTERS:
+: In response to SUBSCRIBE_NAMESPACE, the filter parameters conflict among
+too many subscribers to aggregate the subscription upstream or otherwise
+efficiently service it.
 
 ## SUBSCRIBE {#message-subscribe-req}
 
@@ -4666,6 +4682,7 @@ the length field.
 | PREFIX_OVERLAP             | 0x30 | {{message-request-error}} |
 | NAMESPACE_TOO_LARGE        | 0x31 | {{message-request-error}} |
 | INVALID_JOINING_REQUEST_ID | 0x32 | {{message-request-error}} |
+| CONFLICTING_FILTERS        | 0x33 | {{message-request-error}} |
 | Reserved for greasing      | 0x7f * N + 0x9D | {{grease}} |
 
 ### PUBLISH_DONE Codes {#iana-publish-done}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3925,9 +3925,11 @@ Object in the Subgroup if at least one of the following is true:
    stream.
  * The Object was received on the same upstream Subgroup stream as the
    previously sent Object on the downstream Subgroup stream, with no other
-   Objects in between.
+   Objects in between, unless the intervening Objects did not pass the
+   subscriber's filters.
  * It determined all Object IDs between the current and previous Object IDs
-   on the Subgroup stream belong to different Subgroups or do not exist.
+   on the Subgroup stream belong to different Subgroups or do not exist,
+   or do not pass the subscriber's filters.
 
 If the relay does not know if an Object is the next Object, it MUST reset the
 Subgroup stream and open a new one to forward it.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1297,32 +1297,37 @@ Range Filters are parameters in subscriptions or fetches that tell a publisher
 to filter tracks and objects according to subscriber criteria which are
 allowed ranges of integer values in Track and Object Properties and other
 object header fields (Subgroup ID, Object ID, and Publisher Priority).
-
-There are four Range Filter parameter types, 0x25-0x28, each with a Length
-prefix in bytes and similar structures and behavior as summarized below.
+There are four Range Filter parameter types, 0x25-0x28, as shown below.
 
 ~~~
-SUBGROUP_FILTER { Type=0x25, Length, AndSet, Range... }
-  OBJECT_FILTER { Type=0x26, Length, AndSet, Range... }
-PRIORITY_FILTER { Type=0x27, Length, AndSet, Range... }
-PROPERTY_FILTER { Type=0x28, Length, AndSet, Property Type, Range... }
-          Range { Start (..), [End (..)] }
+SUBGROUP_FILTER { Type=0x25, Length, Set, Range... }
+OBJECTID_FILTER { Type=0x26, Length, Set, Range... }
+PRIORITY_FILTER { Type=0x27, Length, Set, Range... }
+PROPERTY_FILTER { Type=0x28, Length, Set, Property Type, Range... }
+          Range { Start, End }
 ~~~
 
-Each Range Filter is a Length (vi64) prefixed sequence of AndSet (8 bits) followed
-by Start/End inclusive Range pairs.
-Start is delta encoded from the prior Range's End (or from 0
-for the first Range), and End is delta encoded from the current Range's Start.
-The final End in a sequence of Ranges MAY be omitted to indicate no end.
+Each Range Filter is a sequence of Start/End (vi64) inclusive Range pairs
+prefixed with a Length (vi64) in bytes and a Set (8 bits).
+The Property Filter includes an additional prefix for Property Type (vi64).
+The final End in a sequence of Ranges can be omitted to indicate no end.
+
+Start is delta encoded from the prior Range's End or from 0
+for the first Range, and End is delta encoded from the current Range's Start.
 For example, to express ranges 3-5 and 10-15: the first Start is 3
 (delta from 0), the first End is 2 (5 minus 3), the second Start is 5
 (10 minus 5), and the second End is 5 (15 minus 10).
-All filter parameters with the same AndSet value are combined using logical
-"and" operations, then all the resulting sets are combined using logical "or"
-operations.
 
-These parameters MAY appear in a FETCH, SUBSCRIBE, SUBSCRIBE_NAMESPACE,
-PUBLISH_OK, or REQUEST_UPDATE (from a subscriber) message.
+All filter parameters with the same Set value are combined using logical
+"AND" operations, then all the resulting Sets are combined using logical
+"OR" operations.  The final result is Set=0 OR Set=1 OR ... Set=255,
+where each Set=i is the AND of filters with Set=i.
+
+These parameters MAY appear multiple times in a FETCH, SUBSCRIBE,
+SUBSCRIBE_NAMESPACE, PUBLISH_OK, or REQUEST_UPDATE (from a subscriber)
+message.  If the same combination of Parameter Type, Property Type
+(only in the Property Filter), and Set repeat in any message,
+an endpoint MUST close the session with a `PROTOCOL_VIOLATION`.
 
 A filter parameter with zero length indicates no filter, which can
 be used in REQUEST_UPDATE to remove the filter.  If a filter parameter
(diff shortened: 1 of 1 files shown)


suhasHere (approved):
Still looks good to me and easier to review in a new PR

mzanaty:
If you're wondering why it's crickets here, the discussion is at:
https://github.com/mzanaty/moq-transport/issues

RichLogan (approved):
I've spent the last couple of days implementing this into our media client, specifically for the use-case of migrating some existing client side and metadata track driven active speaker switching logic to relay track filters. The setup here is an activity metric in a extension for audio/video tracks, a track filter watching that extension, and some FETCH and New Group Request magic to ease the media join.  This is actually working quite well, and the switching latencies achievable are looking very good (even on fully deselected paths). I do support this to be part of core transport for helping to better map conferencing flows to MoQ.

TimEvens (approved):
We have an implementation that works as documented.

draft-ietf-moq-transport.md
```diff
@@ -1241,6 +1241,53 @@ delivered will the Group ID in 'Start Location' plus the 'End Group Delta'.
...
+
+~~~
+SUBGROUP_FILTER { Type=0x25, Length, Range... }
+  OBJECT_FILTER { Type=0x26, Length, Range... }
+PRIORITY_FILTER { Type=0x27, Length, Range... }
+PROPERTY_FILTER { Type=0x28, Length, Property Type, Range... }
+          Range { Start (..), [End (..)] }
+~~~
+
+Each Range Filter is a Length (vi64) prefixed sequence of Start/End inclusive
```
gloinul: The length definition is missing here. Is this in bytes, or number of range tuples, or varints making up the range?

mzanaty: Good catch. Length is in bytes, previously inherited from Parameters (for odd types), but now each Parameter must define its own, so we must add this.

draft-ietf-moq-transport.md
```diff
@@ -1291,6 +1291,59 @@ close the session with a `PROTOCOL_VIOLATION`.
...
 session with `PROTOCOL_VIOLATION`.
 
+### Range Filters
+
+Range Filters are parameters in subscriptions or fetches that tell a publisher
+to filter tracks and objects according to subscriber criteria which are
+allowed ranges of integer values in Track and Object Properties and other
+object header fields (Subgroup ID, Object ID, and Publisher Priority).
+
+There are four Range Filter parameter types, 0x25-0x28, each with a Length
```
afrind: These need to all be odd?

draft-ietf-moq-transport.md
```diff
@@ -1291,6 +1291,59 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+  OBJECT_FILTER { Type=0x26, Length, AndSet, Range... }
+PRIORITY_FILTER { Type=0x27, Length, AndSet, Range... }
+PROPERTY_FILTER { Type=0x28, Length, AndSet, Property Type, Range... }
+          Range { Start (..), [End (..)] }
+~~~
+
+Each Range Filter is a Length (vi64) prefixed sequence of AndSet (8 bits) followed
+by Start/End inclusive Range pairs.
+Start is delta encoded from the prior Range's End (or from 0
+for the first Range), and End is delta encoded from the current Range's Start.
```
afrind: Any delta encoding that exceeds 2^64-1 is a PROTOCOL_VIOLATION

draft-ietf-moq-transport.md
```diff
@@ -1291,6 +1291,59 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+The final End in a sequence of Ranges MAY be omitted to indicate no end.
```
afrind: Use when instead of MAY?

draft-ietf-moq-transport.md
```diff
@@ -1291,6 +1291,59 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+For example, to express ranges 3-5 and 10-15: the first Start is 3
+(delta from 0), the first End is 2 (5 minus 3), the second Start is 5
+(10 minus 5), and the second End is 5 (15 minus 10).
+All filter parameters with the same AndSet value are combined using logical
```
afrind: You will kill me but we should delta encode the AndSet also to enforce the conditions in an AndSet are consecutive on the wire.

draft-ietf-moq-transport.md
```diff
@@ -1291,6 +1291,59 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+
+Range Filters are only allowed if the setup option MAX_FILTER_RANGES
+is non-zero, which limits the total number of Ranges allowed concurrently
+in all Range Filter parameters for a given subscription or fetch.
+If this limit is exceeded, an endpoint MUST close the session with a
+`PROTOCOL_VIOLATION`.
+
+### Combining Filters
+
+All filter types are combined using logical "and" operations
```
afrind: Does forcing And between filter types leave a relay without multiple subscription capability similarly exposed -- eg one subscriber asks for Ext Foo=1 and another asks for Ext Bar=1 - the relay has to subscribe to all objects and filter locally?

draft-ietf-moq-transport.md
```diff
@@ -1442,6 +1495,121 @@ has accepted a PUBLISH_NAMESPACE with a namespace that exactly matches the
...
+## Filtering Namespaces
+
+Range Filters {{range-filters}} and Track Filters {{track-filters}} can be used
+in SUBSCRIBE_NAMESPACE (or REQUEST_UPDATE for it) to filter tracks and objects
+in a namespace.
+
+### Track Filters
+
+The TRACK_FILTER parameter {{track-filter}} selects a specified number of tracks
+within a namespace with the highest Property Values for a specified Track or
```
afrind, line 1524: highest Property values: sounds like Monopoly lol

draft-ietf-moq-transport.md
```diff
@@ -1442,6 +1495,121 @@ has accepted a PUBLISH_NAMESPACE with a namespace that exactly matches the
...
+value for Property Type, where N = MaxTracksSelected.  The publisher
+MUST send a PUBLISH message for each newly selected track.  For tracks
+with the same value, the earliest delivered object wins the tie
+breaker, so a selected track remains selected until another track
+publishes a higher value that demotes it out of the top N, or Timeout
+elapses before the track delivers an object that remains in the top N,
+either of which deselect the track.  The publisher MUST send a
+REQUEST_UPDATE message with Forward=0 when a track is deselected.
+The publisher MUST send a REQUEST_UPDATE message with Forward=1 when a
+deselected track is reselected, which also updates the Joining Location.
```
afrind: Could use a section reference.

draft-ietf-moq-transport.md
```diff
@@ -1442,6 +1495,121 @@ has accepted a PUBLISH_NAMESPACE with a namespace that exactly matches the
...
+
+Recently deselected tracks SHOULD be kept in a list to avoid more PUBLISH
+messages in case a deselected track is reselected.  A relay SHOULD limit
+this list size if it consumes excessive resources by sending PUBLISH_DONE
```
afrind: s/if it consumes/to prevent consuming/

draft-ietf-moq-transport.md
```diff
@@ -1442,6 +1495,121 @@ has accepted a PUBLISH_NAMESPACE with a namespace that exactly matches the
...
+
+If a track filter for a namespace overlaps with a direct subscription
+to a track name in the same namespace, it is considered to pass the
+track filter whether or not it is counted in the top N list, and MUST NOT
+be subject to track filter state changes or actions.
+
+The track filter evaluates both Track and Object Properties.
+If a track has a Track Property of the specified Property Type, its value
+is used for filtering both the PUBLISH message and any Objects from that track
+that lack their own value for that Property Type.  If the Track Property value
```
afrind: This feels strange here.  If this is true, it should be true for range filters without track filters, no?

draft-ietf-moq-transport.md
```diff
@@ -1442,6 +1495,121 @@ has accepted a PUBLISH_NAMESPACE with a namespace that exactly matches the
...
+are ignored for track filter state updates; however, if the track is already
+selected, they pass the filter.
+
+### Relay Resource Protection in Large Namespaces
+
+Relays SHOULD aggregate and propagate filters upstream on subscriptions,
+especially namespace subscriptions including track selection filters,
+to conserve and protect their resources from excessive load.  They MAY
+also impose limits on the number of publishers in a namespace, by rejecting
+or closing namespace subscriptions with the error NAMESPACE_TOO_LARGE, or
```
afrind, line 1621: Mentioned this in the issue, but I think we need to tighten the bounds here in terms of what relays need to support, or what subscribers can expect.

draft-ietf-moq-transport.md
afrind: scratch that - params no longer play the even/odd game

draft-ietf-moq-transport.md
```diff
@@ -3696,9 +3930,11 @@ Object in the Subgroup if at least one of the following is true:
    stream.
  * The Object was received on the same upstream Subgroup stream as the
```
afrind, line 3933: This stream has be be not filtered as well.  If there's an upstream gap caused by a filter, the downstream needs to be in two streams, unless it was the _same_ filter as downstream

draft-ietf-moq-transport.md
```diff
@@ -1442,6 +1495,121 @@ has accepted a PUBLISH_NAMESPACE with a namespace that exactly matches the
...
+        +--------------+
+
+          Objects FAIL                     Objects PASS
+           the filter                       the filter
+~~~
+
+A track is selected if it publishes an object with the top N highest
+value for Property Type, where N = MaxTracksSelected.  The publisher
+MUST send a PUBLISH message for each newly selected track.  For tracks
+with the same value, the earliest delivered object wins the tie
```
```suggestion
with the same value, the Track with the earliest delivered Object with the higher value wins the tie
```

draft-ietf-moq-transport.md
mzanaty: As we discussed, this is not possible since Parameter Types are delta encoded, forcing all filters of the same Parameter Type to be encoded together.

draft-ietf-moq-transport.md
mzanaty: No, the new sets were added to allow relays to aggregate filters. An example showing the and/or combinations using multiple sets will likely help.

draft-ietf-moq-transport.md
mzanaty: Race to IANA to register the Boardwalk property...

draft-ietf-moq-transport.md
mzanaty: Added a reference to the section on Joining Fetch Range Calculation.

draft-ietf-moq-transport.md
mzanaty: Done

draft-ietf-moq-transport.md
mzanaty: Adding "Track with the" is ok, but the end is wrong, since there is no higher value in a tie with the same value.

draft-ietf-moq-transport.md
mzanaty: Added

draft-ietf-moq-transport.md
mzanaty: s/ MAY / can

draft-ietf-moq-transport.md
mzanaty: Moved to earlier section on range filters.

draft-ietf-moq-transport.md
```diff
@@ -1291,6 +1291,76 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+### Range Filters
+
+Range Filters are parameters in subscriptions or fetches that tell a publisher
+to filter tracks and objects according to subscriber criteria which are
+allowed ranges of integer values in Track and Object Properties and other
+object header fields (Subgroup ID, Object ID, and Publisher Priority).
+There are four Range Filter parameter types, 0x25-0x28, as shown below.
+
+~~~
+SUBGROUP_FILTER { Type=0x25, Length, Set, Range... }
```
yuyou, line 1303: how to filter a subgroup(s) for a track? The Start in the Range does not have a SubGroup ID, as the [Location Structure](https://www.ietf.org/archive/id/draft-ietf-moq-transport-18.html#name-location-structure) is 

```
Location {
  Group (vi64),
  Object (vi64)
}
```

mzanaty: The range in a subgroup filter is the subgroup ID (e.g. 0) not Location. A separate Subscription / Location filter is needed to select locations.

mzanaty:
Range filters were moved to PR#1765. They will be removed from here once that merges.

draft-ietf-moq-transport.md
yuyou: Sure, I am not asking for a Location Filter. Just suggest any potential fixes to the design, e.g. the next line #1307 where the spec says:
 `Range { Start, End }`
In the case of subgroup_filter, the Range must be subgroup ID.

draft-ietf-moq-transport.md
```diff
@@ -1291,6 +1291,76 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+OBJECTID_FILTER { Type=0x26, Length, Set, Range... }
+PRIORITY_FILTER { Type=0x27, Length, Set, Range... }
+PROPERTY_FILTER { Type=0x28, Length, Set, Property Type, Range... }
+          Range { Start, End }
```
yuyou, line 1307: In the case of video scrubbing, what if I like to get some frames every n objects, instead of the 1st object in every group? e.g. in a VOD example, the player wants to get some snapshots every n seconds, regardless of the GOP size. would it be useful to have an optional "step" value provided to the `Range`?  

e.g.
Range { Start, End [, Step] }

afrind:
@mzanaty Should we close this PR now that Range Filters have merged, or do you plan to continue merging and iterating here?

mzanaty:
Range Filters were moved to #1765 and merged.
Location Filters were moved to #1809 and merge is pending.
Top Track Filters were moved to #1830 and merge is pending.
Closing this now.

PR #1526, opened 2026-03-04
Replace vi64 with v64

vasilvv:
I think I suggested this when we were discussing the original PR in the editors/authors code, and no one objected, but it got lost somewhere before landing.


ianswett:
We decided to do the naming change separately, since it's purely editorial.

I have no opinion on this, I'm fine with either.

afrind:
We will ask in the next virtual interim

suhasHere (approved):

afrind:
Discussed at 125 and there was not sufficient energy to make a change

Issue #1529, opened 2026-03-06
Subscribe Namespace and Subscribe Overlap


suhasHere:
Due to the introduction of rendezvous timeout for Subscribe,  we have a situation where a relay might have the below scenario: 

Subscriber sends 
SUBSCRIBE to ns:           (a,b,c) name: d 
SUBSCRIBE_NS to ns:    (a,b,c)

Publisher sends
PUBLISH to.                  ns: (a,b,c) name: d 

Now the relay sees  2 possible way to react to the incoming publish
1. Forward PUBLISH as a response to SUBSCRIBE_NAMESPACE due to prefix match
2. Send SUBSCRIBE_OK to fulfill the SUBSCRIBE

There are few options we can deal in this situation
1. Publisher Wins  ( so option 1 above). This 
2. Subscriber Wins ( so option 2 above). 
3. Add a parameter to SubscribeNamespace called `excludedSuffixes': [], a list of suffix that the subscriber wants not to be bothered about when sending Publish.

I am inclined to option 2 since it let's the end subscriber influence the choice. 

I also like option 3 since it allows to attain such functionality generically. 

Thoughts ?

afrind:
This is similar to a publisher and subscriber racing -- one sending PUBLISH A and the other sending SUBSCRIBE A.  The current draft text says publisher wins, and subscriber get REQUEST_ERROR.  I think it was somewhat arbitrary that we picked publisher wins.

I'm also inclined towards 2 in this case.  If you resolve the subscribe first, then at least my implementation logic would skip sending the PUBLISH, since this subscriber is already subscribed to the track.  Are we ok with the inconsistency with the above, or should we flip that to subscriber wins also?  It is slightly different - in the above the publisher _already_ sent the publish when the subscribe arrives.

I'd prefer not to add more machinery as in 3 at this time as long as 2 solves the problem.

suhasHere:
excludeSuffixes mimics Namespace model and allows one to not know about its own namespace, if needed.

PR #1533, opened 2026-03-06
SUBSCRIBE takes precedence over SUBSCRIBE_NAMESPACE at relay

afrind:
When a relay receives a PUBLISH for a track that it is already holding a downstream SUBSCRIBE for (awaiting a publisher via rendezvous timeout), proceed with the SUBSCRIBE rather than forwarding the PUBLISH via SUBSCRIBE_NAMESPACE.

Fixes: #1529

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1738,14 +1738,10 @@ then the Relay MUST use Forward=1 when subscribing upstream.
 
 When a relay receives an incoming PUBLISH message, it MUST send a PUBLISH
 request to each subscriber that has subscribed (via SUBSCRIBE_NAMESPACE)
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1740,6 +1740,13 @@ When a relay receives an incoming PUBLISH message, it MUST send a PUBLISH
 request to each subscriber that has subscribed (via SUBSCRIBE_NAMESPACE)
 to the Track's namespace or prefix thereof.
 
+When a relay receives an incoming PUBLISH message for a Track for which
```
ianswett: There's no mention of SUBSCRIBE_NAMESPACE in the first sentence.  Can you meld this into the prior sentence, so the transition makes sense?

PR #1534, opened 2026-03-06
Add REDIRECT for request errors and established subscriptions

### LINKED ISSUE #1481: Do we need a way to move / goaway for individual track

afrind:
- Add a Redirect structure (Connect URI + Full Track Name) and REDIRECT error code to REQUEST_ERROR, 
allowing publishers/relays to redirect new requests (SUBSCRIBE, FETCH, TRACK_STATUS,
SUBSCRIBE_NAMESPACE) to a different location
- Add a standalone REDIRECT message for redirecting established subscriptions (SUBSCRIBE, PUBLISH,
SUBSCRIBE_NAMESPACE, PUBLISH_NAMESPACE) without tearing down the session
- Servers receiving a non-empty Connect URI MUST close with PROTOCOL_VIOLATION; non-empty Track Name
on namespace-scoped requests is also PROTOCOL_VIOLATION

This is in two separate commits so we can easily remove the second behavior if not warranted.

Fixes: #1481

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2737,15 +2737,15 @@ Redirect {
   receives a Redirect with a non-zero Connect URI Length it MUST close the
   session with a `PROTOCOL_VIOLATION`.
 
-* Track Namespace: The Track Namespace to use for the redirected request. If
-  both Track Namespace and Track Name have zero length, the redirected request
-  uses the same Full Track Name as the original request.
+* Track Namespace and Track Name: The Track Namespace and Track Name to use
+  for the redirected request. If both have zero length, the redirected request
+  uses the same values as the original request. Otherwise, Track Namespace and
+  Track Name are the literal values for the redirected request.
 
-* Track Name: The Track Name to use for the redirected request. Track Name
-  is not meaningful for namespace-scoped requests (SUBSCRIBE_NAMESPACE,
-  PUBLISH_NAMESPACE) and MUST be empty; an endpoint that receives a non-empty
-  Track Name in a Redirect for a namespace-scoped request MUST close the session
-  with a `PROTOCOL_VIOLATION`.
+  Track Name is not meaningful for namespace-scoped requests
+  (SUBSCRIBE_NAMESPACE, PUBLISH_NAMESPACE) and MUST be empty; an endpoint that
+  receives a non-empty Track Name in a Redirect for a namespace-scoped request
+  MUST close the session with a `PROTOCOL_VIOLATION`.
 
 ### REQUEST_ERROR Message Format
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2923,6 +2923,8 @@ indicate that the peer can re-establish the request on a different session.
 REDIRECT MAY be sent on streams for SUBSCRIBE, PUBLISH, SUBSCRIBE_NAMESPACE,
 and PUBLISH_NAMESPACE. An endpoint MUST close the session with a
 `PROTOCOL_VIOLATION` if it receives a REDIRECT on any other stream type.
+If an endpoint wants to drain an entire Session, it SHOULD use GOAWAY rather
+than sending REDIRECT on each Request.
 
 REDIRECT does not terminate the current request. The sender MAY continue to
 serve the request after sending REDIRECT. For subscriptions, a publisher MAY
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2737,7 +2737,9 @@ Redirect {
   receives a Redirect with a non-zero Connect URI Length it MUST close the
   session with a `PROTOCOL_VIOLATION`.
 
-* Track Namespace: The Track Namespace to use for the redirected request.
+* Track Namespace: The Track Namespace to use for the redirected request. If
+  both Track Namespace and Track Name have zero length, the redirected request
+  uses the same Full Track Name as the original request.
 
 * Track Name: The Track Name to use for the redirected request. Track Name
   is not meaningful for namespace-scoped requests (SUBSCRIBE_NAMESPACE,
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2919,7 +2919,7 @@ EXCESSIVE_LOAD (0x9):
 ## REDIRECT {#message-redirect}
 
 An endpoint sends a REDIRECT message on a bidirectional request stream to
-indicate that the peer should re-establish the request at a different location.
+indicate that the peer can re-establish the request on a different session.
 REDIRECT MAY be sent on streams for SUBSCRIBE, PUBLISH, SUBSCRIBE_NAMESPACE,
 and PUBLISH_NAMESPACE. An endpoint MUST close the session with a
 `PROTOCOL_VIOLATION` if it receives a REDIRECT on any other stream type.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2570,7 +2570,8 @@ REDIRECT:
 location specified in the Redirect structure. The requester SHOULD establish a
 new session to the provided URI (if present) and retry the request using the
 Full Track Name from the Redirect (if present). This error code can appear in
-response to SUBSCRIBE, FETCH, TRACK_STATUS, and SUBSCRIBE_NAMESPACE.
+response to SUBSCRIBE, FETCH, TRACK_STATUS, PUBLISH_NAMESPACE and
+SUBSCRIBE_NAMESPACE.
 
 Below are errors for use by the publisher. They can appear in response to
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -393,6 +393,34 @@ Reason Phrase {
...
+  Connect URI (..),
+  Track Namespace (..),
+  Track Name Length (vi64),
+  Track Name (..),
+}
+~~~
+
+* Connect URI: The URI to connect to for this track. If the length is
+  zero, the requester SHOULD use the current session's URI. If a server
+  receives a Redirect with a non-zero Connect URI Length it MUST close the
```
ianswett: Is this always true?  If two relays contact each other, then one is the client and one is the server, but I can imagine either wanting to do a redirect?

draft-ietf-moq-transport.md
```diff
@@ -2531,6 +2565,13 @@ DUPLICATE_SUBSCRIPTION (0x19):
 : The PUBLISH or SUBSCRIBE request attempted to create a subscription to a Track
 with the same role as an existing subscription.
 
+REDIRECT:
+: The request cannot be fulfilled by this endpoint, but could succeed at the
+location specified in the Redirect structure. The requester SHOULD establish a
+new session to the provided URI (if present) and retry the request using the
+Full Track Name from the Redirect (if present). This error code can appear in
+response to SUBSCRIBE, FETCH, TRACK_STATUS, and SUBSCRIBE_NAMESPACE.
```
ianswett: Above you mention PUBLISH_NAMESPACE, does it apply to PUBLISH_NAMESPACE?  I assume not?

ianswett (approved):

draft-ietf-moq-transport.md
afrind: I copied this from GOAWAY.  Your use case kind of makes sense and kind of doesn't.

draft-ietf-moq-transport.md
afrind: I think it should?

```suggestion
response to SUBSCRIBE, FETCH, TRACK_STATUS, PUBLISH_NAMESPACE and
SUBSCRIBE_NAMESPACE.
```

ianswett: As long as we're consistent, I don't have a strong opinion.

draft-ietf-moq-transport.md
```diff
@@ -2874,6 +2916,32 @@ UPDATE_FAILED (0x8):
 EXCESSIVE_LOAD (0x9):
 : The publisher is overloaded and is terminating the subscription.
 
+## REDIRECT {#message-redirect}
+
+An endpoint sends a REDIRECT message on a bidirectional request stream to
+indicate that the peer should re-establish the request at a different location.
```
```suggestion
indicate that the peer can re-establish the request on a different session.
```

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -2874,6 +2916,32 @@ UPDATE_FAILED (0x8):
 EXCESSIVE_LOAD (0x9):
 : The publisher is overloaded and is terminating the subscription.
 
+## REDIRECT {#message-redirect}
+
+An endpoint sends a REDIRECT message on a bidirectional request stream to
+indicate that the peer can re-establish the request on a different session.
+REDIRECT MAY be sent on streams for SUBSCRIBE, PUBLISH, SUBSCRIBE_NAMESPACE,
+and PUBLISH_NAMESPACE. An endpoint MUST close the session with a
+`PROTOCOL_VIOLATION` if it receives a REDIRECT on any other stream type.
```
```suggestion
`PROTOCOL_VIOLATION` if it receives a REDIRECT on any other stream type.
If an endpoint wants to drain an entire Session, it SHOULD use GOAWAY rather
than sending REDIRECT on each Request.
```

afrind:
Discussed 3/30: Mo suggested it might need a timeout, but on further reflection no one had the appetite for it.  This should be good to go.

vasilvv:
Noting my current concerns about this:
- I suspect we almost certainly want different messages from redirecting a request in lieu of OK/ERROR, vs redirecting an active subscription.
- Who can use the latter form of the redirect? The publisher? The request initiator?
- Is there any special handling required if I redirect a PUBLISH, but I already started sending objects?

There's probably more but I've not thought of those yet.

afrind:
> I suspect we almost certainly want different messages from redirecting a request in lieu of OK/ERROR, vs redirecting an active subscription.

That is already true in this PR:

Redirecting a request -> SUBSCRIBE_ERROR, code=REDIRECT + Redirect structure
Redirecting an active subscription -> REDIRECT message (contains redirect structure).

> Who can use the latter form of the redirect? The publisher? The request initiator?

The request initiator can _always_ redirect their _own_ request by cancelling the request and making a new one.  No action is required by the receiver.  The useful cases are:

* Publisher redirecting a subscriber (go subscribe here instead please)
* Subscriber redirecting a publisher (useful for client-publishers -> server subscriber says "please publish here instead")

I don't think we need a client publisher who initiates the request to ask the server to SUBSCRIBE to it at some other location.

So I think the answer is "the entity responding to a request can send a REQUEST_ERROR+REDIRECT or REDIRECT message".

> Is there any special handling required if I redirect a PUBLISH, but I already started sending objects?

I think you mean:

C->S: PUBLISH
C->S: Objects
S->C: ERROR+Redirect

?

In this case, we can add language that no objects will be processed by the server and the client can resend them.

C->S: PUBLISH
S->C: PUBLISH OK
C->S: Objects
...
S->C: REDIRECT

This is the more interesting case, since the client might like to know which of its objects were received and processed by the server.  We could add a field to REDIRECT which is the largest object received/processed.

> There's probably more but I've not thought of those yet.

This is not very actionable.

afrind:
Discussed in author/editor call:

Remove REDIRECT message from this PR
Use GOAWAY on a bidi stream to mean what REDIRECT did

draft-ietf-moq-transport.md
```diff
@@ -2714,16 +2714,47 @@ REQUEST_OK Message {
...
+  receives a Redirect with a non-zero Connect URI Length it MUST close the
+  session with a `PROTOCOL_VIOLATION`.
+
+* Track Namespace: The Track Namespace to use for the redirected request.
+
+* Track Name: The Track Name to use for the redirected request. Track Name
+  is not meaningful for namespace-scoped requests (SUBSCRIBE_NAMESPACE,
+  PUBLISH_NAMESPACE) and MUST be empty; an endpoint that receives a non-empty
+  Track Name in a Redirect for a namespace-scoped request MUST close the session
+  with a `PROTOCOL_VIOLATION`.
```
suhasHere: I am not sure how a relay would know the right FullTrackName which is application scoped.  I

suhasHere (commented):
The idea seems to be fine. But I am not 100% sure we should add track infomration to this.

vasilvv:
This overall looks good, but we do need text on relay behavior (forwarding and caching).

draft-ietf-moq-transport.md
afrind: @suhasHere this is done via configuration rules typically, it's not in-band

afrind:
> This overall looks good, but we do need text on relay behavior (forwarding and caching).

@vasilvv Do you remember what we agreed to say?  Cacheable up to retry interval?

vasilvv (approved):

draft-ietf-moq-transport.md
suhasHere: Then we should make this an optional field ?

draft-ietf-moq-transport.md
```diff
@@ -2714,16 +2714,47 @@ REQUEST_OK Message {
...
+A Redirect provides a way for an endpoint to direct the peer to retry a
+request at a different URI and/or for a different Full Track Name.
+
+~~~
+Redirect {
+  Connect URI Length (vi64),
+  Connect URI (..),
+  Track Namespace (..),
+  Track Name Length (vi64),
+  Track Name (..),
```
suhasHere, line 2731:
```suggestion
  [Track Namespace (..),
  Track Name Length (vi64),
  Track Name (..)],
```

suhasHere (approved):

draft-ietf-moq-transport.md
afrind: I left them required, but said that 0 length Track Namespace / Name means "use the same one in the original request"

draft-ietf-moq-transport.md
afrind: Done, see above

draft-ietf-moq-transport.md
```diff
@@ -2476,28 +2672,83 @@ GOAWAY Message {
...
+}
+~~~
+
+* Connect URI: The URI to connect to for this track. If the length is
+  zero, the requester SHOULD use the current session's URI. If a server
+  receives a Redirect with a non-zero Connect URI Length it MUST close the
+  session with a `PROTOCOL_VIOLATION`.
+
+* Track Namespace: The Track Namespace to use for the redirected request. If
+  both Track Namespace and Track Name have zero length, the redirected request
```
ianswett: What if only one is zero length?

Issue #1498, opened 2026-02-12
Does a REQUEST_OK (FETCH_OK) need to contain Largest Object?


ianswett:
For a cache that has enough Objects to serve the FETCH, does the FETCH really need to go upstream?

One can easily imagine cases when the FETCH could be completed quickly, but finding out Largest Object is relatively slow.

This is partially motivated by the fact that it'd be nice to be able to return REQUEST_OK as the start of a response stream and then follow it with the Objects being Fetched.

afrind:
It's called "largest object" but it's actually not always the largest object - it's more like the "fetch endpoint".  It's only something other than what you specified as Fetch End if Fetch End spans the true largest object.  Maybe the right behavior is to make this optional and only include it when the range spans the end.

PR #1536, opened 2026-03-06
Clarify FETCH_OK End Location semantics

afrind:
Reword End Location description to clarify it conveys the end of the requested range, not the Largest Object on the track. It equals the FETCH request End Location unless the range extends beyond published data. Remove relay-specific text from FETCH_OK in favor of the existing relay FETCH handling section.

Fixes: #1498

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3041,7 +3041,7 @@ Fetch specifies an inclusive range of Objects starting at Start Location and
 ending at End Location. End Location MUST specify the same or a larger Location
 than Start Location for Standalone and Absolute Joining Fetches.
 
-Objects that are not yet published will not be retrieved by a FETCH.  If the
+Objects larger than the Largest Object will not be retrieved by a FETCH.  If the
 requested End Location exceeds the Largest available Object, the actual end of
 the FETCH response is indicated in the FETCH_OK End Location.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -3041,10 +3041,9 @@ Fetch specifies an inclusive range of Objects starting at Start Location and
 ending at End Location. End Location MUST specify the same or a larger Location
 than Start Location for Standalone and Absolute Joining Fetches.
 
-Objects that are not yet published will not be retrieved by a FETCH.  The
-Largest available Object in the requested range is indicated in the FETCH_OK,
-and is the last Object a fetch will return if the End Location have not yet been
-published.
+Objects that are not yet published will not be retrieved by a FETCH.  If the
```
ianswett: Unrelated to this PR, but I think this isn't quite true.


```suggestion
Objects larger than the Largest Object will not be retrieved by a FETCH.  If the
```

draft-ietf-moq-transport.md
```diff
@@ -3083,22 +3082,16 @@ FETCH_OK Message {
 * End Of Track: 1 if all Objects have been published on this Track, and
   the End Location is the final Object in the Track, 0 if not.
 
-* End Location: The largest object covered by the FETCH response.
-  The End Location is determined as follows:
+* End Location: The end of the range covered by the FETCH response.
```
ianswett: Can we note if this is inclusive?  I think it is?

ianswett (approved):
I do think it'd be nicer to use Parameters for this, since it'd allow us to get rid of FETCH_OK, but either way this text is an improvement.

draft-ietf-moq-transport.md
afrind: Hmm, with publisher publishing random groups I think "not yet published" is more correct?

draft-ietf-moq-transport.md
afrind: It depends on what the definition of is is.

Since in _the request_ the end is encoded as Object + 1, and we say it's included here, then one could say it's _exclusive_ of what you receive on the wire.  If it's 0, then it means "you asked for and got the entire group".

draft-ietf-moq-transport.md
ianswett: Maybe, but the reason I made this suggestion is it makes more sense in the context of the following sentence.

draft-ietf-moq-transport.md
afrind: I rewrote it for even more clarity.

draft-ietf-moq-transport.md
afrind: You are right, FETCH only goes to Largest Object and non-temporal publishers get the shaft.

Issue #1517, opened 2026-03-01
varint encoding using min size possible


fluffy:
we have 

** To reduce unnecessary use of bandwidth, variable length integers SHOULD be encoded using the least number of bytes possible to **

This seems to be something that can contributes to bug by allowing something that is seldom tested because most software does the SHOULD. 

I think this would be better as a MUST. I do not understand when it would be reasonable for software not to do it.

afrind:
I think we copied this language from RFC 9000. Unsure if anyone remembers the thought process there?

I agree with you that now there’s 9 ways to say “0” and that may not be great.

It does create an extra error check on the decode path - today I can look at the first byte, wait for that many more bytes to arrive, and I’m guaranteed to decode a valid integer, which is very likely going into a uint64_t.

LPardue:
https://www.rfc-editor.org/rfc/rfc9000.html#section-16 says

> Values do not need to be encoded on the minimum number of bytes necessary, with the sole exception of the Frame Type field; see [Section 12.4](https://www.rfc-editor.org/rfc/rfc9000.html#frames).

Which states

> The Frame Type field uses a variable-length integer encoding (see [Section 16](https://www.rfc-editor.org/rfc/rfc9000.html#integer-encoding)), with one exception. To ensure simple and efficient implementations of frame parsing, a frame type MUST use the shortest possible encoding. For frame types defined in this document, this means a single-byte encoding, even though it is possible to encode these values as a two-, four-, or eight-byte variable-length integer. For instance, though 0x4001 is a legitimate two-byte encoding for a variable-length integer with a value of 1, PING frames are always encoded as a single byte with the value 0x01. This rule applies to all current and future QUIC frame types. An endpoint MAY treat the receipt of a frame type that uses a longer encoding than necessary as a connection error of type PROTOCOL_VIOLATION.


There was a fairly length discussion about the topic of minimal encoding in https://github.com/quicwg/base-drafts/issues/2299. If you want to make it a MUST then surveying implemeneters for feasibility first would make sense.

kixelated:
I've purposely used more bytes than needed with QUIC VarInts. Specifically, encoding the packet size as 0 with two bytes. After the packet is fully encoded, then I update it with the correct size, which might not actually need two bytes. 

I think MoQ should keep the SHOULD.

afrind:
Control messages us 16 bit length not varint - or you mean something else?

kixelated:
[...]

PR #1538, opened 2026-03-06
Require minimal varint encoding

afrind:
Change variable length integer encoding from SHOULD to MUST use the minimum number of bytes. Non-minimal encoding is a PROTOCOL_VIOLATION. No current use case justifies non-minimal encoding, and allowing it creates unnecessary attack surface and implementation complexity.

Replace the non-minimal example (0x8025 = 37) with a valid 2-byte example (0x81ee = 494).

Fixes: #1517

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -288,7 +288,7 @@ The following table contains some example encodings:
 | Byte Sequence        | Decimal Value              |
 |----------------------|----------------------------|
 | 0x25                 | 37                         |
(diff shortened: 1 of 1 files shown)


vasilvv:
Related issue: #1543

suhasHere (approved):

afrind:
We should talk about this with #1543.  If we don't do this PR _and_ allow 7 byte varints, then the _only_ error condition for a varint is "need more data"

afrind:
Discussed at 125 and decided not to do this.

PR #1540, opened 2026-03-07
Allow coalescing REQUEST_UPDATE processing

### LINKED ISSUE #1433: Can multiple SUBSCRIBE_UPDATES be collapsed into one, and then applied?
When discussing potential DoS/Resource issues in MoQ with @englishm I wondered if we could merge a sequence of SUBSCRIBE_UPDATE messages into one, and then apply that one message to reduce churn and resource consumption, whether it's an attack or just a noisy subscriber (ie: re-prioritizes every few ms with no hysteresis)?

I believe that all existing Parameters have the property that applying a sequence of SUBSCRIBE_UPDATE messages is equivalent to applying one SUBSCRIBE_UPDATE where a later Parameter value overwrites a prior value.

For example, for sequence of updates to the Subscriber Priority or Forward value, only the latest value matters.

If we can state this MUST always be true in the draft, I think it makes parameters easier to reason about, but importantly for DoS, it allows a publisher that's behind on SUBSCRIBE_UPDATEs to enact only one SUBSCRIBE_UPDATE to catch up.

afrind:
Clarify that a receiver of multiple REQUEST_UPDATE messages may coalesce their processing by applying only the cumulative result, since parameter values are last-one-wins. Each REQUEST_UPDATE still requires its own REQUEST_OK or REQUEST_ERROR response.

Fixes: #1433

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2696,6 +2696,13 @@ When a subscription
 update is unsuccessful, the publisher MUST also terminate the subscription with
 PUBLISH_DONE with error code `UPDATE_FAILED`.
 
+A receiver of multiple REQUEST_UPDATE messages on the same stream MAY
+coalesce their processing by applying only the cumulative result.
+Parameter values from later REQUEST_UPDATE messages override values
+from earlier ones. The receiver MUST still send a REQUEST_OK or
+REQUEST_ERROR in response to each REQUEST_UPDATE, but it is not
+required to process intermediate states individually.
+
 ## PUBLISH {#message-publish}
 
 The publisher sends PUBLISH as the first message on a new bidirectional stream
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2707,9 +2707,12 @@ the bidi stream.
 A receiver of multiple REQUEST_UPDATE messages on the same stream MAY
 coalesce their processing by applying only the cumulative result.
 Parameter values from later REQUEST_UPDATE messages override values
-from earlier ones. The receiver MUST still send a REQUEST_OK or
-REQUEST_ERROR in response to each REQUEST_UPDATE, but it is not
-required to process intermediate states individually.
+from earlier ones. The receiver MUST still send a REQUEST_OK for
+each successful update, but it is not required to process
+intermediate states individually. If the coalesced REQUEST_UPDATE
+results in REQUEST_ERROR, only a single REQUEST_ERROR will be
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2696,6 +2696,13 @@ When a subscription
 update is unsuccessful, the publisher MUST also terminate the subscription with
 PUBLISH_DONE with error code `UPDATE_FAILED`.
 
+A receiver of multiple REQUEST_UPDATE messages on the same stream MAY
+coalesce their processing by applying only the cumulative result.
+Parameter values from later REQUEST_UPDATE messages override values
+from earlier ones. The receiver MUST still send a REQUEST_OK or
+REQUEST_ERROR in response to each REQUEST_UPDATE, but it is not
+required to process intermediate states individually.
```
ianswett: Q: If it only processes the last one, and that's an ERROR, does it send ERROR for all of them with the same error code even through really it was the last one that failed?

I wonder if it'd be better to just send one REQUEST_OK or REQUEST_ERROR back with the largest Request ID processed?

afrind: Discussed in call:

We need to require updates be processed in order regardless
REQ_OK/ERROR doesn't have an ID in it anymore.
I think you should only get one ERROR, you will not know which REQUEST_UDPATE triggered the problem if the receiver coalesced.  Tough cookies.

draft-ietf-moq-transport.md
```diff
@@ -2704,6 +2704,13 @@ a FETCH, the publisher MUST reset the FETCH data stream. When a REQUEST_UPDATE
 fails for a SUBSCRIBE_NAMESPACE or PUBLISH_NAMESPACE, the responder MUST close
 the bidi stream.
 
+A receiver of multiple REQUEST_UPDATE messages on the same stream MAY
+coalesce their processing by applying only the cumulative result.
+Parameter values from later REQUEST_UPDATE messages override values
+from earlier ones. The receiver MUST still send a REQUEST_OK or
+REQUEST_ERROR in response to each REQUEST_UPDATE, but it is not
+required to process intermediate states individually.
```
```suggestion
from earlier ones. The receiver MUST still send a REQUEST_OK for
each successful update, but it is not required to process
intermediate states individually. If the coalesced REQUEST_UPDATE
results in REQUEST_ERROR, only a single REQUEST_ERROR will be
sent and the sender of the REQUEST_UPDATEs will not always be
able to determine which caused an error.
```

vasilvv (approved):

ianswett (approved):

suhasHere (approved):

PR #1542, opened 2026-03-07
Split SUBSCRIBE_NAMESPACE into SUBSCRIBE_NAMESPACE and SUBSCRIBE_TRACKS

### LINKED ISSUE #1458: Edge Cases around updating SUBSCRIBE_NAMESPACE
Now that SUBSCRIBE_NAMESPACE has an option, there is not much flexibility to control the option type.

Say I want namespaces for "Foo" but PUBLISH and NAMESPACE for "Foo/bar". I could express this as two separate SUBSCRIBE_NAMESPACEs but I can't have overlapping prefixes, so there's no legal way to do it.

Say I start out just wanting NAMESPACE for "Foo" but later want PUBLISH as well. I can't send another message because that would be overlapping prefixes. I believe the only to proceed is kill the stream and start over. This is fraught because I have no way of guaranteeing that the Reset arrives before the new SN unless I wait for an RTT, and even then assumes the QUIC API gives me packet granularity.

And of course, a RESET/SUBSCRIBE pattern introduces latency in getting stuff while waiting.

Some possibilities:
- we could just get rid of the option altogether, as it's got serious problems.
- we could make the option a parameter, which would at least solve the second problem but not the first one.
- we could lift the "no overlapping prefixes" rule, which has its own can of worms
- there could be entirely separate namespace trees for PUBLISH and NAMESPACE. At that point, it's basically a different message and should probably be split as such.

afrind:
Replace the single SUBSCRIBE_NAMESPACE message (0x11) with two separate
message types: SUBSCRIBE_NAMESPACE (0x50) for namespace discovery
(NAMESPACE/NAMESPACE_DONE) and SUBSCRIBE_TRACKS (0x51) for track
subscriptions (PUBLISH). This removes the SUBSCRIBE_NAMESPACE_OPTIONS
parameter and the BOTH mode entirely — behavior is now determined by
message type.

The two types have independent overlap spaces, so a subscriber may hold
a SUBSCRIBE_NAMESPACE and SUBSCRIBE_TRACKS with the same prefix
simultaneously.

Add TRACK_NAMESPACE_PREFIX parameter (0x34) to allow updating the Track
Namespace Prefix of an established SUBSCRIBE_NAMESPACE or SUBSCRIBE_TRACKS
via REQUEST_UPDATE, avoiding the need to tear down and restart a stream
when the prefix needs to change.

Fixes: #1458


draft-ietf-moq-transport.md
```diff
@@ -2241,6 +2241,24 @@ If the parameter is omitted from REQUEST_UPDATE, the value for the
 subscription remains unchanged.  If the parameter is omitted from any other
 message, the default value is 1.
 
+### SUBSCRIBE NAMESPACE OPTIONS Parameter {#subscribe-options}
```
ianswett: nit: Can you call this link subscribe-namespace-options?


```suggestion
### SUBSCRIBE NAMESPACE OPTIONS Parameter {#subscribe-namespace-options}
```

draft-ietf-moq-transport.md
```diff
@@ -4285,6 +4300,7 @@ TODO: register the URI scheme and the ALPN
 | 0x21 | SUBSCRIPTION_FILTER | {{subscription-filter}} |
 | 0x22 | GROUP_ORDER | {{group-order}} |
 | 0x32 | NEW_GROUP_REQUEST | {{new-group-request}} |
+| 0x33 | SUBSCRIBE_NAMESPACE_OPTIONS | {{subscribe-options}} |
```
```suggestion
| 0x33 | SUBSCRIBE_NAMESPACE_OPTIONS | {{subscribe-namespace-options}} |
```

ianswett (approved):

afrind:
Discussed 3/30: 

1) We want to _remove_ this option and bake it into the message type: no more "BOTH".
2) We want to allow updating the prefix via REQUEST_UPDATE, rather than relaxing overlap restrictions

draft-ietf-moq-transport.md
```diff
@@ -1427,34 +1427,41 @@ The syntax of these messages is described in {{message}}.
...
-bidirectional stream in response to a SUBSCRIBE_NAMESPACE. The subscriber
-SHOULD close the session with a protocol error if it detects receiving more
-than one.
+SUBSCRIBE_NAMESPACE or SUBSCRIBE_TRACKS to publishers/relays it has established
+a session with.
+
+SUBSCRIBE_NAMESPACE requests namespace discovery: the publisher sends relevant
+NAMESPACE and NAMESPACE_DONE messages for namespaces matching the prefix,
+including echoing back Track Namespaces under the prefix that have been published
+to it.
```
suhasHere, line 1556: I think this has gone backwards in direction and i am leaning towards the definition we had so far. 
The mode, where Subscribe is for FTN and SubscribeNamespace is for Namespace and both end up getting data seems natural.
SubcribeNamespace is wild card subscription and we want all track under it namespace to arrive to us. This change of semantics feels more mental load in explaining concepts personally.

draft-ietf-moq-transport.md
```diff
@@ -1427,34 +1427,41 @@ The syntax of these messages is described in {{message}}.
...
+
+SUBSCRIBE_TRACKS requests track subscriptions: the publisher sends PUBLISH
```
suhasHere, line 1558: How about call it Discover_Names, which is its intent. Its for discovery purposes only and doesn't establish subscription(s)

draft-ietf-moq-transport.md
```diff
@@ -1427,34 +1427,41 @@ The syntax of these messages is described in {{message}}.
...
+this behaves as self-subscription described in {{subscriptions}}.
+
+Either message with zero Track Namespace fields indicates the sender is
+interested in all namespaces or all tracks from the receiver, respectively.
+
+The subscriber sends SUBSCRIBE_NAMESPACE or SUBSCRIBE_TRACKS on a new
+bidirectional stream and the publisher MUST send a single REQUEST_OK or
+REQUEST_ERROR as the first message on the bidirectional stream in response.
+The subscriber SHOULD close the session with a protocol error if it detects
+receiving more than one.
```
ianswett: I think this sentence is potentially confusing as worded and we're already saying it has to be the first message on the response stream, so I'd rather just drop it.

```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -1427,34 +1427,41 @@ The syntax of these messages is described in {{message}}.
...
 
 If a Subscription cannot be created because there is no available Request ID,
```
ianswett: No available bidirectional Streams.

Looks like this didn't get updated in the bidi stream PR.

draft-ietf-moq-transport.md
```diff
@@ -1876,55 +1883,57 @@ control stream ({{session-init}}), and Request indicates a bidirectional
...
+|--------|------------------------------------------------|----------------|
+| 0x16   | FETCH ({{message-fetch}})                      | Request, First |
+|--------|------------------------------------------------|----------------|
+| 0x18   | FETCH_OK ({{message-fetch-ok}})                | Request        |
+|--------|------------------------------------------------|----------------|
+| 0xD    | TRACK_STATUS ({{message-track-status}})        | Request, First |
+|--------|------------------------------------------------|----------------|
+| 0x6    | PUBLISH_NAMESPACE ({{message-pub-ns}})         | Request, First |
+|--------|------------------------------------------------|----------------|
+| 0x50   | SUBSCRIBE_NAMESPACE ({{message-subscribe-ns}}) | Request, First |
```
ianswett, line 2039: Did you mean to renumber these?  If so, did you want to note that 0x11 was used for the old SUBSCRIBE_NAMESPACE?

draft-ietf-moq-transport.md
afrind: > The mode, where Subscribe is for FTN and SubscribeNamespace is for Namespace and both end up getting data seems natural.

You can easily achieve this by sending SUBSCRIBE_NAMESPACE a/ + SUBSCRIBE_TRACKS a/

> SubcribeNamespace is wild card subscription and we want all track under it namespace to arrive to us.

Is it just a naming problem?  What you are asking for is called SUBSCRIBE_TRACKS here but I am not overly hung up on names.

> This change of semantics feels more mental load in explaining concepts personally.

I was skeptical at first, but I actually found it somewhat refreshing - for example PUBLISH_BLOCKED is now scoped exclusively to SUBSCRIBE_TRACKS.

draft-ietf-moq-transport.md
afrind: You seem to have it backwards -- SUBSCRIBE_NAMESPACE would be DISCOVER_NAMES[PACE].  SUBSCRIBE_TRACKS gets you PUBLISH.

draft-ietf-moq-transport.md
```diff
@@ -2785,6 +2808,19 @@ a FETCH, the publisher MUST reset the FETCH data stream. When a REQUEST_UPDATE
...
+
+A subscriber can update the Track Namespace Prefix of an established
+SUBSCRIBE_NAMESPACE or SUBSCRIBE_TRACKS by including the
+TRACK_NAMESPACE_PREFIX parameter ({{track-namespace-prefix-param}}) in a
+REQUEST_UPDATE.  The overlap restriction applies independently per type: the
+new prefix MUST NOT share a common prefix with any other active
+SUBSCRIBE_NAMESPACE (for a SUBSCRIBE_NAMESPACE update) or SUBSCRIBE_TRACKS
+(for a SUBSCRIBE_TRACKS update) in the same session.  If the update is
+accepted, the publisher adjusts which NAMESPACE, NAMESPACE_DONE, or PUBLISH
+messages it sends to reflect the new prefix.  Messages sent before the
```
ianswett: 'reflect' seems odd here.  How about having a separate sentence for SUBSCRIBE_NAMESPACE and SUBSCRIBE_TRACKS and be more specific about what happens.  For SUBSCRIBE_TRACKS, we should be clear about what happens to existing subscriptions (I think they remain open unless cancelled?) and whether you get a PUBLISH for existing Tracks in the namespace, or only when a new Track is published in that namespace.

draft-ietf-moq-transport.md
```diff
@@ -3367,12 +3389,71 @@ stream, it SHOULD treat this as though each active namespace received a
...
+
+* Parameters: The parameters are defined in {{message-params}}.
+
+The publisher will respond with REQUEST_OK or REQUEST_ERROR on the response half
+of the stream. If the subscriber receives any frame other than a REQUEST_OK or a
+REQUEST_ERROR as the first frame on the response half of the stream, then it
+MUST close the session with a PROTOCOL_VIOLATION. If the SUBSCRIBE_TRACKS is
+successful, the publisher will send PUBLISH messages on new bidirectional streams
+for tracks within matching namespaces. If it is an error, the stream will be
+immediately closed via FIN.
```
```suggestion
closed via FIN after REQUEST_ERROR is sent.
```

I realize this text happens elsewhere, but I think this is a bit clearer?  I can do it as a separate PR if that's better.  The text about REQUEST_OK and REQUEST_ERROR is repeated a lot.

draft-ietf-moq-transport.md
```diff
@@ -3367,12 +3389,71 @@ stream, it SHOULD treat this as though each active namespace received a
...
+Within a session, if a publisher receives a SUBSCRIBE_TRACKS with a
+Track Namespace Prefix that shares a common prefix with an established
+SUBSCRIBE_TRACKS, it MUST respond with REQUEST_ERROR with error code
+`PREFIX_OVERLAP`.  SUBSCRIBE_TRACKS and SUBSCRIBE_NAMESPACE have independent
+overlap spaces (see {{message-subscribe-ns}}).
+
+The publisher MUST ensure the subscriber is authorized to perform this
+namespace subscription.
+
+SUBSCRIBE_TRACKS is not required for a publisher to send PUBLISH messages to
```
ianswett, line 3626: Do we even need this paragraph?

ianswett (commented):
Some detailed comments, but looks good overall.

draft-ietf-moq-transport.md
afrind: I changed it on purpose since the functionality was different.

draft-ietf-moq-transport.md
afrind: It was there before and now appears in both sections, but needs some updates.

draft-ietf-moq-transport.md
```diff
@@ -3367,12 +3387,71 @@ stream, it SHOULD treat this as though each active namespace received a
...
+a subscriber.  It is useful in applications or relays where subscribers are
```
```suggestion
a subscriber.  It is useful for subscribers are
```

draft-ietf-moq-transport.md
```diff
@@ -1427,34 +1427,39 @@ The syntax of these messages is described in {{message}}.
...
+this behaves as self-subscription described in {{subscriptions}}.
+
+Either message with zero Track Namespace fields indicates the sender is
+interested in all namespaces or all tracks from the receiver, respectively.
+
+The subscriber sends SUBSCRIBE_NAMESPACE or SUBSCRIBE_TRACKS on a new
+bidirectional stream and the publisher MUST send a single REQUEST_OK or
+REQUEST_ERROR as the first message on the bidirectional stream in response.
 
 If a Subscription cannot be created because there is no available Request ID,
```
```suggestion
If a Subscription cannot be created because there are no available bidirectional streams,
```

draft-ietf-moq-transport.md
```diff
@@ -2785,6 +2806,19 @@ a FETCH, the publisher MUST reset the FETCH data stream. When a REQUEST_UPDATE
...
+A subscriber can update the Track Namespace Prefix of an established
+SUBSCRIBE_NAMESPACE or SUBSCRIBE_TRACKS by including the
+TRACK_NAMESPACE_PREFIX parameter ({{track-namespace-prefix-param}}) in a
+REQUEST_UPDATE.  The overlap restriction applies independently per type: the
+new prefix MUST NOT share a common prefix with any other active
+SUBSCRIBE_NAMESPACE (for a SUBSCRIBE_NAMESPACE update) or SUBSCRIBE_TRACKS
+(for a SUBSCRIBE_TRACKS update) in the same session.  If the update is
+accepted, the publisher adjusts which NAMESPACE, NAMESPACE_DONE, or PUBLISH
+messages it sends to reflect the new prefix.  Messages sent before the
+REQUEST_OK do not reflect the new prefix.
```
```suggestion
(for a SUBSCRIBE_TRACKS update) in the same session.  If the update is
accepted, NAMESPACE and NAMESPACE_DONE messages following the
REQUEST_OK will contain Track Namespace suffixes relative to the
updated prefix.  Updating the prefix of a SUBSCRIBE_TRACKS has
no effect on existing subscriptions.  If the subscriber is no longer
interested it can cancel the corresponding bidirectional stream.
```

draft-ietf-moq-transport.md
```diff
@@ -1427,34 +1427,39 @@ The syntax of these messages is described in {{message}}.
...
 forward the result to the application, so the application can decide which other
 publishers to contact, if any.
 
-A SUBSCRIBE_NAMESPACE can be cancelled by closing the stream with
-either a FIN or by resetting it. Cancelling does not prohibit original publishers
+A SUBSCRIBE_NAMESPACE or SUBSCRIBE_TRACKS can be cancelled by closing the
+stream with either a FIN or RESET_STREAM. Cancelling does not prohibit original publishers
 from sending further PUBLISH_NAMESPACE or PUBLISH messages, but relays MUST NOT
 send any further PUBLISH messages to a client without knowing the client is
 interested in and authorized to receive the content.
```
afrind, line 1584:
```suggestion
A SUBSCRIBE_NAMESPACE or SUBSCRIBE_TRACKS can be cancelled by closing the
stream with either a FIN or RESET_STREAM. Cancelling SUBSCRIBE_TRACKS does not prohibit original publishers
from sending further PUBLISH messages, but relays MUST NOT
send any further PUBLISH messages to a client without knowing the client is
interested in and authorized to receive the content.
```

draft-ietf-moq-transport.md
```diff
@@ -1469,7 +1474,7 @@ in a namespace without having received a PUBLISH_NAMESPACE for it.
 If a publisher is authoritative for a given namespace, or is a relay that has
 received an authorized PUBLISH_NAMESPACE for that namespace from an upstream
 publisher, it MUST send a PUBLISH_NAMESPACE to any subscriber that has
```
```suggestion
publisher, it MUST send a NAMESPACE message to any subscriber that has
```

draft-ietf-moq-transport.md
```diff
@@ -3317,43 +3349,31 @@ SUBSCRIBE_NAMESPACE Message {
...
+Within a session, if a publisher receives a SUBSCRIBE_NAMESPACE with a
+Track Namespace Prefix that shares a common prefix with an established
+SUBSCRIBE_NAMESPACE, it MUST respond with REQUEST_ERROR with error code
+`PREFIX_OVERLAP`.  SUBSCRIBE_NAMESPACE and SUBSCRIBE_TRACKS have independent
+overlap spaces (see {{message-subscribe-tracks}}).
 
 The publisher MUST ensure the subscriber is authorized to perform this
 namespace subscription.
 
 SUBSCRIBE_NAMESPACE is not required for a publisher to send PUBLISH_NAMESPACE
```
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -3317,43 +3349,31 @@ SUBSCRIBE_NAMESPACE Message {
...
-
-If the FORWARD parameter ({{forward-parameter}}) is present in this message and
-equal to 0, PUBLISH messages resulting from this SUBSCRIBE_NAMESPACE will set
-the FORWARD parameter to 0. If the FORWARD parameter is equal to 1 or omitted
-from this message, PUBLISH messages resulting from this SUBSCRIBE_NAMESPACE will
-set the FORWARD parameter to 1, or indicate that value by omitting the parameter
-(see {{subscriptions}}).
+messages to a subscriber.  It is useful in applications or relays where
+subscribers are only interested in or authorized to access a subset of
+available namespaces.
```
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -1511,35 +1511,40 @@ The syntax of these messages is described in {{message}}.
...
-than one.
-
-If a Subscription cannot be created because there is no available Request ID,
-the Publisher sends a PUBLISH_BLOCKED message on the response stream to indicate
-the Full Track Name of the Subscription that could not be established. The Publisher
-MUST NOT send a PUBLISH for a Track after PUBLISH_BLOCKED has been sent.  The subscriber can instead issue a SUBSCRIBE to establish a subscription to that track.
+SUBSCRIBE_NAMESPACE or SUBSCRIBE_TRACKS to publishers/relays it has established
+a session with.
+
+SUBSCRIBE_NAMESPACE requests namespace discovery: the publisher sends relevant
```
suhasHere, line 1553:
```suggestion
SUBSCRIBE_NAMESPACES requests namespace discovery: the publisher sends relevant
```

draft-ietf-moq-transport.md
```diff
@@ -1511,35 +1511,40 @@ The syntax of these messages is described in {{message}}.
...
+NAMESPACE and NAMESPACE_DONE messages for namespaces matching the prefix,
+including echoing back Track Namespaces under the prefix that have been published
+to it.
+
+SUBSCRIBE_TRACKS requests track subscriptions: the publisher sends PUBLISH
+messages for tracks within matching namespaces, including echoing back published
+Tracks under the prefix to the subscriber. If an endpoint accepts its own PUBLISH,
+this behaves as self-subscription described in {{subscriptions}}.
```
suhasHere: I think we don't allow for the tracks to be echoed by default (sub-ns with self track commit)

draft-ietf-moq-transport.md
```diff
@@ -1511,35 +1511,40 @@ The syntax of these messages is described in {{message}}.
...
+
+Either message with zero Track Namespace fields indicates the sender is
+interested in all namespaces or all tracks from the receiver, respectively.
+
+The subscriber sends SUBSCRIBE_NAMESPACE or SUBSCRIBE_TRACKS on a new
+bidirectional stream and the publisher MUST send a single REQUEST_OK or
+REQUEST_ERROR as the first message on the bidirectional stream in response.
```
suhasHere, line 1567: should we also add a note to say the namespace, namespace done, publish, publish done messages all get sent on the same bidirectional stream estabalisining the request ?

draft-ietf-moq-transport.md
```diff
@@ -1960,55 +1965,57 @@ control stream ({{session-init}}), and Request indicates a bidirectional
...
+| 0xE    | NAMESPACE_DONE ({{message-namespace-done}})    | Request        |
+|--------|------------------------------------------------|----------------|
+| 0xF    | PUBLISH_BLOCKED ({{message-publish-blocked}})  | Request        |
+|--------|------------------------------------------------|----------------|
+| 0x2    | REQUEST_UPDATE ({{message-request-update}})    | Request        |
+|--------|------------------------------------------------|----------------|
+| 0x7    | REQUEST_OK ({{message-request-ok}})            | Request        |
+|--------|------------------------------------------------|----------------|
+| 0x5    | REQUEST_ERROR ({{message-request-error}})      | Request        |
+|--------|------------------------------------------------|----------------|
```
suhasHere, line 2054: may be you need to add one line that says what does "First" mean here ?

draft-ietf-moq-transport.md
```diff
@@ -2903,6 +2924,21 @@ results in REQUEST_ERROR, only a single REQUEST_ERROR will be
...
 able to determine which caused an error.
 
+### Updating Namespace Subscriptions
+
+A subscriber can update the Track Namespace Prefix of an established
+SUBSCRIBE_NAMESPACE or SUBSCRIBE_TRACKS by including the
+TRACK_NAMESPACE_PREFIX parameter ({{track-namespace-prefix-param}}) in a
+REQUEST_UPDATE.  The overlap restriction applies independently per type: the
+new prefix MUST NOT share a common prefix with any other active
+SUBSCRIBE_NAMESPACE (for a SUBSCRIBE_NAMESPACE update) or SUBSCRIBE_TRACKS
```
suhasHere, line 3024: clarifying question:
if i do a sub-ns to a/b
and then do req_update to a/b/c ( more focussed namespace), is it an error ?

draft-ietf-moq-transport.md
```diff
@@ -3435,44 +3469,27 @@ SUBSCRIBE_NAMESPACE Message {
...
+`PREFIX_OVERLAP`.  SUBSCRIBE_NAMESPACE and SUBSCRIBE_TRACKS have independent
+overlap spaces (see {{message-subscribe-tracks}}).
 
 The publisher MUST ensure the subscriber is authorized to perform this
 namespace subscription.
 
-SUBSCRIBE_NAMESPACE is not required for a publisher to send PUBLISH_NAMESPACE
-or PUBLISH messages to a subscriber.  It is useful in applications or relays
-where subscribers are only interested in or authorized to access a subset of
-available namespaces and tracks.
```
suhasHere, line 3538: why did we remove this ? I think this still holds true right ?

suhasHere (commented):
this is looking pretty close to be done. Thanks for the work. 

I do have one comment on how the prefix_overlap error is framed on updated.

vasilvv (approved):

draft-ietf-moq-transport.md
afrind: 🤷 I can spend 45 seconds asking in the interim

draft-ietf-moq-transport.md
afrind: I think this whole paragraph is out of place.  The mechanics like that are covered in the message section.

draft-ietf-moq-transport.md
afrind: @suhasHere seem immediately above the table  

> Messages marked "First" MUST be the first message on a new request stream.

draft-ietf-moq-transport.md
afrind: It is not an error.  It is only an error if the new namespace overlaps with a different sub_ns

draft-ietf-moq-transport.md
afrind: I think @ianswett wanted to remove it. It's not normative, and I think he felt like it wasn't adding that much value.

draft-ietf-moq-transport.md
```suggestion
messages for tracks within matching namespaces, excluding tracks published
by the subscriber.
```

afrind: It was removed in #1596, I updated here to match

draft-ietf-moq-transport.md
afrind: Claude thinks the text is fine:

> you could add a brief clarifying clause like "excluding the subscription being updated" after "any other active SUBSCRIBE_NAMESPACE" to make it explicit. But I'd lean toward leaving it as-is since "any other" already implies the distinction.

draft-ietf-moq-transport.md
```diff
@@ -2800,12 +2818,14 @@ UNINTERESTED:
 Errors below can only be used in response to one message type.
 
 PREFIX_OVERLAP:
-: In response to SUBSCRIBE_NAMESPACE, the namespace prefix overlaps with another
-SUBSCRIBE_NAMESPACE in the same session.
+: In response to SUBSCRIBE_NAMESPACE or SUBSCRIBE_TRACKS, the namespace prefix
+overlaps with another subscription of the same type in the same session.
```
```suggestion
shares a common prefix with another subscription of the same type in the same session.
```

draft-ietf-moq-transport.md
```diff
@@ -2497,6 +2504,15 @@ to the Largest Group, it does not send a NEW_GROUP_REQUEST upstream.
 After sending a NEW_GROUP_REQUEST upstream, the request is considered
 outstanding until the Largest Group increases.
 
+### TRACK_NAMESPACE_PREFIX Parameter {#track-namespace-prefix-param}
+
+The TRACK_NAMESPACE_PREFIX parameter (Parameter Type 0x34) uses the Track
+Namespace encoding described in {{track-name}}.  It MAY appear in REQUEST_UPDATE
+for a SUBSCRIBE_NAMESPACE or SUBSCRIBE_TRACKS request.  It updates the Track
+Namespace Prefix for that subscription.  If the new prefix would overlap with
```
```suggestion
Namespace Prefix for that subscription.  If the new prefix would share a common prefix with
```

suhasHere (approved):

suhasHere (approved):

PR #1544, opened 2026-03-08
Improve Startup Latency and 0-RTT

### LINKED ISSUE #83: Improve startup latency

### LINKED ISSUE #420: Explain security of 0-rtt setup
Note in the draft that raq QUIC can to 0-RTT moq setup but webtransport does not suppoir this. Add text to security section of why this is safe.

ianswett:
Add sections on how to reduce startup latency and 0-RTT.

Fixes #420
Fixes #83


draft-ietf-moq-transport.md
```diff
@@ -888,6 +888,67 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+
+As such, a client or server can initiate a MoQ session, subscribe, and
+start publishing Objects all in parallel.
+
+Prior to receiving the peer's SETUP message, it's unknown what extensions
+a peer will support. Message Parameters requiring negotiation SHOULD NOT
+be used prior to receiving the peer's SETUP message unless the application
+requires the extension or the endpoint knows the peer supports the
+extension. If an unsupported Message Parameter is used, the peer will be
+unable to process it and the session will be terminated. See {message-params}
```
```suggestion
unable to process it and the session will be terminated. See {message-params}.
```

draft-ietf-moq-transport.md
```diff
@@ -888,6 +888,67 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+0-RTT with HTTP in more detail. If 0-RTT is used with an existing or future
+version of WebTransport, the following would apply to it as well as QUIC.
+
+0-RTT is not forward secure, so MoQ users need to consider that when
+deciding if 0-RTT is appropriate.
+
+MoQ Messages and Objects as defined in this draft are safe to replay in most
+circumstances.
+
+* TRACK_STATUS gets the Largest Object and Track Properties, but does not
```
afrind, line 1043: It actually doesn't get Track Properties right now :|

draft-ietf-moq-transport.md
```diff
@@ -888,6 +888,67 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+  change the state of a Track or any Object in the Track.
+* SUBSCRIBE requests Objects be delivered, but does not change the Objects
+  being requested.
+* PUBLISH initiates a Subscription and Objects can be immediately sent to
+  the Subscriber. Sending the same Objects multiple times does not change
+  the Track contents, because MoQ ignores duplicate Objects.
```
afrind: MoQ -> MOQT?

Also, I'm not sure it's true that MOQT ignores duplicate objects?  Maybe you mean to say it's relatively easy for the application to detect duplicate objects?

draft-ietf-moq-transport.md
```diff
@@ -888,6 +888,67 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+* SUBSCRIBE_NAMESPACE requests a list of namespaces and the establishment
+  of new subscriptions, but does not change the available Namespaces,
+  Tracks, or Objects contained within a Track.
+* PUBLISH_NAMESPACE indicates that Subscriptions should be sent to the
+  Publisher. If a Subscription was sent to the replaying endpoint, it
+  would fail because the endpoint cannot complete the handshake.
+
```
afrind, line 1057: All of these mutate the state of the relay (or relay network).  Replaying a 0-RTT subscribe to every host in every pop could cause problems?  Do we need to mention things like this?

draft-ietf-moq-transport.md
```diff
@@ -888,6 +888,67 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+Some potential side effects of replay are:
+
+* The NEW_GROUP_REQUEST Parameter {{new-group-request}} can cause a new
```
afrind: NGR is deduplicated by the value.  A replayed NGR should only reach the OP once?

afrind (commented):
Some of the text is slightly duplicative of what we have in extension negotiation I think?

Someone smarter than me needs to review the 0-RTT section.

draft-ietf-moq-transport.md
ianswett: Hmmm should we fix that?  It seems like it should since the intent was to be identical to SUBSCRIBE, but not cause any Objects to be sent.  Or we can finally remove it, I don't care.

draft-ietf-moq-transport.md
```diff
@@ -888,6 +888,67 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+  change the state of a Track or any Object in the Track.
+* SUBSCRIBE requests Objects be delivered, but does not change the Objects
+  being requested.
+* PUBLISH initiates a Subscription and Objects can be immediately sent to
+  the Subscriber. Sending the same Objects multiple times does not change
+  the Track contents, because MoQ ignores duplicate Objects.
+* SUBSCRIBE_NAMESPACE requests a list of namespaces and the establishment
+  of new subscriptions, but does not change the available Namespaces,
+  Tracks, or Objects contained within a Track.
+* PUBLISH_NAMESPACE indicates that Subscriptions should be sent to the
```
afrind: Lowercase should

draft-ietf-moq-transport.md
ianswett: If you are a Subscriber to two Publishers for the same Track, I believe the current draft says that you ignore whichever arrives second, modulo potential Object properties?

Thoughts on how to say that?

draft-ietf-moq-transport.md
```suggestion
* PUBLISH_NAMESPACE requests Subscriptions for Tracks in the namespace be sent to that
```

draft-ietf-moq-transport.md
ianswett: Interesting point.  Spurious PUBLISH_NAMESPACEs could cause churn in the available namespaces?  Though if we're only talking replay, then presumably someone else already published that namespace successfully, so this would just be an extra available Publisher?

draft-ietf-moq-transport.md
ianswett: Good point, removed.

draft-ietf-moq-transport.md
```diff
@@ -1026,6 +1029,55 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+MoQ Messages and Objects as defined in this draft are safe to replay in most
+circumstances.
+
+* TRACK_STATUS gets the Largest Object and Track Properties, but does not
+  change the state of a Track or any Object in the Track.
+* SUBSCRIBE requests Objects be delivered, but does not change the Objects
+  being requested.
+* PUBLISH initiates a Subscription and Objects can be immediately sent to
+  the Subscriber. Sending the same Objects multiple times does not change
+  the Track contents, because MoQ ignores duplicate Objects.
```
```suggestion
  the Subscriber. Processing the same Objects multiple times is
  idempotent, as the subscriber or relay can identify and discard
  duplicates based on the Group ID and Object ID.
```

draft-ietf-moq-transport.md
```diff
@@ -1015,6 +1015,9 @@ begin with any other message type unless negotiated. If they do, the peer MUST
 close the Session with a `PROTOCOL_VIOLATION`. Objects are sent on unidirectional
 streams.
 
+As such, a client or server can initiate a MoQ session, subscribe, and
```
```suggestion
As such, a client or server can initiate a MoQT session, subscribe, and
```

draft-ietf-moq-transport.md
```diff
@@ -1026,6 +1029,56 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+QUIC supports 0-RTT ({{Section 2.3 of ?RFC8446}}), but Webtransport over QUIC
+is not expected to use 0-RTT, because initializing a WebTransport session
+uses CONNECT, which is not a safe method. {{?RFC8470}} describes the use of
+0-RTT with HTTP in more detail. If 0-RTT is used with an existing or future
+version of WebTransport, the following would apply to it as well as QUIC.
+
+0-RTT is not forward secure, so MoQ users need to consider that when
+deciding if 0-RTT is appropriate.
+
+MoQ Messages and Objects as defined in this draft are safe to replay in most
```
```suggestion
0-RTT is not forward secure, so MoQT users need to consider that when
deciding if 0-RTT is appropriate.

MoQT Messages and Objects as defined in this draft are safe to replay in most
```

draft-ietf-moq-transport.md
```diff
@@ -1026,6 +1029,56 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+  to that Publisher. If a Subscription was sent to the replaying endpoint, it
+  would fail because the endpoint cannot complete the handshake.
+
+Some potential side effects of replay are:
+
+* Publishing Objects that were previously published could cause those
+  Objects to be distributed to active Subscriptions if the relays do
+  not identify them as already having been published.
+
+Replays could increase load on the MoQ network. For relay to client
```
```suggestion
Replays could increase load on the MoQT network. For relay to client
```

draft-ietf-moq-transport.md
```diff
@@ -1026,6 +1029,56 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+  not identify them as already having been published.
+
+Replays could increase load on the MoQT network. For relay to client
+traffic, this is no worse than 0-RTT in HTTP/3, since it'd be limited by
+the amplification factor until address validation. However, it could cause
+the relay to initiate new upstream Subscriptions. For a SUBSCRIBE_NAMESPACE
+that requested Subscriptions in the Namespace, sending that upstream could
+cause the Relay to receive a number of new Subscriptions on the replaying
+client's behalf.
+
```
ianswett, line 1072:
```suggestion
client's behalf.

Relays SHOULD implement rate-limiting or validation for 0-RTT requests that trigger upstream subscriptions to mitigate resource exhaustion from replayed packets.

```

draft-ietf-moq-transport.md
```diff
@@ -1026,6 +1029,58 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+Relays SHOULD implement rate-limiting or validation for 0-RTT requests that trigger upstream subscriptions to mitigate resource exhaustion from replayed packets.
```
```suggestion
Relays SHOULD implement rate-limiting or validation for 0-RTT requests
that trigger upstream subscriptions to mitigate resource exhaustion from
replayed packets.
```

draft-ietf-moq-transport.md
```diff
@@ -1015,6 +1015,9 @@ begin with any other message type unless negotiated. If they do, the peer MUST
 close the Session with a `PROTOCOL_VIOLATION`. Objects are sent on unidirectional
 streams.
 
+As such, a client or server can initiate a MoQT session, subscribe, and
```
afrind: server's can't initiate a session?

Also MOQT

draft-ietf-moq-transport.md
```diff
@@ -1026,6 +1029,60 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+  would fail because the endpoint cannot complete the handshake.
+
+Some potential side effects of replay are:
+
+* Publishing Objects that were previously published could cause those
+  Objects to be distributed to active Subscriptions if the relays do
+  not identify them as already having been published.
+
+Replays could increase load on the MoQT network. For relay to client
+traffic, this is no worse than 0-RTT in HTTP/3, since it'd be limited by
```
afrind: 'it'd be' is a little colloquial

draft-ietf-moq-transport.md
```diff
@@ -1026,6 +1029,60 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+
+### 0-RTT
+
+QUIC supports 0-RTT ({{Section 2.3 of ?RFC8446}}), but WebTransport over QUIC
+is not expected to use 0-RTT, because initializing a WebTransport session
+uses CONNECT, which is not a safe method. {{?RFC8470}} describes the use of
+0-RTT with HTTP in more detail. If 0-RTT is used with an existing or future
+version of WebTransport, the following would apply to it as well as QUIC.
+
+0-RTT is not forward secure, so MoQT users need to consider that when
```
```suggestion
0-RTT is not forward secure, so MOQT users need to consider that when
```

draft-ietf-moq-transport.md
```diff
@@ -1026,6 +1029,60 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+deciding if 0-RTT is appropriate.
+
+MoQT Messages and Objects as defined in this draft are safe to replay in most
```
```suggestion
MOQT Messages and Objects as defined in this draft are safe to replay in most
```

draft-ietf-moq-transport.md
```diff
@@ -1026,6 +1029,60 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+  to that Publisher. If a Subscription was sent to the replaying endpoint, it
+  would fail because the endpoint cannot complete the handshake.
+
+Some potential side effects of replay are:
+
+* Publishing Objects that were previously published could cause those
+  Objects to be distributed to active Subscriptions if the relays do
+  not identify them as already having been published.
+
+Replays could increase load on the MoQT network. For relay to client
```
```suggestion
Replays could increase load on the MOQT network. For relay to client
```

afrind (commented):
I would still feel more comfortable with a security expert's review

draft-ietf-moq-transport.md
```suggestion
As such, a client can initiate a MOQT session, subscribe, and
```

draft-ietf-moq-transport.md
```suggestion
traffic, this is no worse than 0-RTT in HTTP/3, since the server is limited by
```

afrind (approved):

sharmafb (approved):

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -1026,6 +1029,60 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+
+### 0-RTT
+
+QUIC supports 0-RTT ({{Section 2.3 of ?RFC8446}}), but WebTransport over QUIC
+is not expected to use 0-RTT, because initializing a WebTransport session
+uses CONNECT, which is not a safe method. {{?RFC8470}} describes the use of
+0-RTT with HTTP in more detail. If 0-RTT is used with an existing or future
+version of WebTransport, the following would apply to it as well as QUIC.
+
+0-RTT is not forward secure, so MOQT users need to consider that when
```
vasilvv: I don't understand what forward secrecy has anything to do with the text of this section.

vasilvv (changes requested):
Overall LGTM, except for one paragraph that does not make sense to me.

draft-ietf-moq-transport.md
ianswett: I think I took that from HTTP/3 or the early data draft?  Should I remove this?

vasilvv: I don't see text like that in RFC 8470.  Let's just remove it?

draft-ietf-moq-transport.md
```diff
@@ -1019,6 +1022,60 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+deciding if 0-RTT is appropriate.
+
```
```suggestion
```

draft-ietf-moq-transport.md
ianswett: Removed

draft-ietf-moq-transport.md
```diff
@@ -1026,6 +1029,60 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+be used prior to receiving the peer's SETUP message unless the application
+requires the extension or the endpoint knows the peer supports the
+extension. If an unsupported Message Parameter is used, the peer will be
+unable to process it and the session will be terminated. See {{message-params}}.
+
+### 0-RTT
+
+QUIC supports 0-RTT ({{Section 2.3 of ?RFC8446}}), but WebTransport over QUIC
+is not expected to use 0-RTT, because initializing a WebTransport session
+uses CONNECT, which is not a safe method. {{?RFC8470}} describes the use of
```
martinthomson, line 1036: This introductory sentence is a bit weird.

I think that what you are saying is:

> QUIC 0-RTT provides the option for a client to initiate transactions immediately after attempting to establish a connection, without waiting for a response.

And:

> 0-RTT is only available to MOQT when using QUIC directly; a WebTransport session cannot be instantiated in the 0-RTT phase of the QUIC connection that it runs over, though it is possible to use WebTransport streams concurrent with WebTransport setup on an existing connection.

And then the RFC8470 business.

draft-ietf-moq-transport.md
martinthomson: The main thing you likely want to focus on is replay risk, yeah.  Though I would want to retain a note about post-compromise security.  That is, 0-RTT data does not remain confidential and authenticated if the static secret used to protect is is compromised.  For that, I would stick a reference to RFC 8446 (bis) in.  But keep the emphasis on replay.

draft-ietf-moq-transport.md
```diff
@@ -1026,6 +1029,60 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+deciding if 0-RTT is appropriate.
+
+MOQT Messages and Objects as defined in this draft are safe to replay in most
+circumstances.
+
+* TRACK_STATUS gets the Largest Object and Track Properties, but does not
+  change the state of a Track or any Object in the Track.
+* SUBSCRIBE requests Objects be delivered, but does not change the Objects
+  being requested.
+* PUBLISH initiates a Subscription and Objects can be immediately sent to
```
martinthomson:
```suggestion
* PUBLISH initiates a Subscription. Objects can be immediately sent to
```

The "and" here reads poorly.  Probably more so if you are less familiar with MOQT.  Reword according to your preferences, of course.

draft-ietf-moq-transport.md
```diff
@@ -1026,6 +1029,60 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+is not expected to use 0-RTT, because initializing a WebTransport session
+uses CONNECT, which is not a safe method. {{?RFC8470}} describes the use of
+0-RTT with HTTP in more detail. If 0-RTT is used with an existing or future
+version of WebTransport, the following would apply to it as well as QUIC.
+
+0-RTT is not forward secure, so MOQT users need to consider that when
+deciding if 0-RTT is appropriate.
+
+MOQT Messages and Objects as defined in this draft are safe to replay in most
+circumstances.
```
martinthomson, line 1041: In MOQT, is the only state that might be altered by a request on the server?  It might be worth noting that, because 0-RTT messages that commit resources of a client (I authorize a payment of $10 being the canonical one) are also implicated by replay.

From what I can see, this list seems fine, but I'm not read up on MOQT.  You should be extra careful to check that you are not making assumptions about what it means to deliver objects though.  In theory, HTTP GET doesn't have side effects, but it often does.  By the same token, HTTP POST is assumed to have effects, but that is not always the case, so it can be safe to send in 0-RTT if both parties understand what is going on (DoH being a good example of that).  In our analysis in RFC 8470, we had to be very precise in wording things because of that mismatch between expectations and reality.

draft-ietf-moq-transport.md
```diff
@@ -1026,6 +1029,60 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+  Tracks, or Objects contained within a Track.
+* PUBLISH_NAMESPACE requests that Subscriptions under the namespace be sent
+  to that Publisher. If a Subscription was sent to the replaying endpoint, it
+  would fail because the endpoint cannot complete the handshake.
+
+Some potential side effects of replay are:
+
+* Publishing Objects that were previously published could cause those
+  Objects to be distributed to active Subscriptions if the relays do
+  not identify them as already having been published.
```
martinthomson: As above, this is one of those examples of where server-side state is altered by a request.  If this happened in 0-RTT, followed by an action that reversed the decision, then a replay could undo that reversal.  That's an example of where a series of actions is not idempotent, even where each individual action is idempotent.  Again, focus on the state change that the 0-RTT triggers: if it is non-consequential, then it is OK to recommend that servers accept the request, but they will have to make their own decisions about what is non-consequential.

draft-ietf-moq-transport.md
```diff
@@ -1026,6 +1029,60 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+traffic, this is no worse than 0-RTT in HTTP/3, since the server is limited by
+the amplification factor until address validation. However, it could cause
+the relay to initiate new upstream Subscriptions. For a SUBSCRIBE_NAMESPACE
+that requested Subscriptions in the Namespace, sending that upstream could
+cause the Relay to receive a number of new Subscriptions on the replaying
+client's behalf.
+
+Relays SHOULD implement rate-limiting or validation for 0-RTT requests
+that trigger upstream subscriptions to mitigate resource exhaustion from
+replayed packets.
```
martinthomson: Is this it?

In RFC 8470, we recommended that servers do one of two things:

1. Accept the risk and process the request.
2. Defer processing and wait until the handshake is complete (this is unlikely to be feasible at scale, but it's worth noting that many handshakes complete and you can often commit less costly actions before making a final call).
3. Reject the request.

I strongly recommend that you provide a "try again" style of rejection into these things so that servers can make their own decisions about the risk associated with requests that appear in 0-RTT.

draft-ietf-moq-transport.md
ianswett: Technically I could 0-RTT a WebTransport session if I declare my request as safe in the client library.  For example, there are cases clients do that today for gRPC requests, which are POST.  But I agree my effort to allude to that without saying it is making the text more awkward.

draft-ietf-moq-transport.md
ianswett: You can't unpublish an Object, though it can time out from the cache, so this would make it temporarily available again.

draft-ietf-moq-transport.md
```diff
@@ -1019,6 +1022,57 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+  Tracks, or Objects contained within a Track.
+* PUBLISH_NAMESPACE requests that Subscriptions under the namespace be sent
+  to that Publisher. If a Subscription was sent to the replaying endpoint, it
+  would fail because the endpoint cannot complete the handshake.
+
+Some potential side effects of replay are:
+
+* Publishing Objects that were previously published could cause those
+  Objects to be distributed to active Subscriptions if the relays do
+  not identify them as already having been published.
```
```suggestion
  not identify them as already having been published. This re-distribution could
  also make them available in cache again after they previously had expired.
```

draft-ietf-moq-transport.md
```diff
@@ -1019,6 +1022,58 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+* PUBLISH_NAMESPACE requests that Subscriptions under the namespace be sent
+  to that Publisher. If a Subscription was sent to the replaying endpoint, it
+  would fail because the endpoint cannot complete the handshake.
+
+Some potential side effects of replay are:
+
+* Publishing Objects that were previously published could cause those
+  Objects to be distributed to active Subscriptions if the relays do
+  not identify them as already having been published. This re-distribution could
+  also make them available in cache again after they previously had expired.
```
```suggestion
  also make them available in cache again after they previously expired.
```

draft-ietf-moq-transport.md
```diff
@@ -1019,6 +1022,58 @@ A control stream MUST NOT be closed at the underlying transport layer during the
...
+traffic, this is no worse than 0-RTT in HTTP/3, since the server is limited by
+the amplification factor until address validation. However, it could cause
+the relay to initiate new upstream Subscriptions. For a SUBSCRIBE_NAMESPACE
+that requested Subscriptions in the Namespace, sending that upstream could
+cause the Relay to receive a number of new Subscriptions on the replaying
+client's behalf.
+
+Relays SHOULD implement rate-limiting or validation for 0-RTT requests
+that trigger upstream subscriptions to mitigate resource exhaustion from
+replayed packets.
```
ianswett: Is this better?

```suggestion
Relays MAY defer initiating upstream subscriptions until the handshake is complete
or reject 0-RTT entirely to mitigate resource exhaustion from replayed packets.
```

draft-ietf-moq-transport.md
ianswett: Good point. I proposed a rewrite below

draft-ietf-moq-transport.md
martinthomson: Do you have a specific response code that a relay could use so that the client can know that this was something that can be retried?  Or is it always possible to retry a subscription?  Surely there are classes of rejection that are permanent and others that are temporary.

vasilvv (approved):

Issue #1547, opened 2026-03-09
SUBGROUP_HEADER's type scope comfirm


xiaosen83:
In Section 3.4, it says there are 3 types of unidirectional steams, and the scrop of SUBGROUP_HEADER is "0x10-0x1D", But in the defination of SUBGROUP_HEADER in 10.4.2, it includes  also 0x30..0x35 / 0x38..0x3D.

Section 3.4
ID | Type
-- | --
0x05 | FETCH_HEADER (Section 10.4.4)
**0x10-0x1D** | SUBGROUP_HEADER (Section 10.4.2)
0x2F00 | SETUP (Section 9.4)

Section 10.4.2
SUBGROUP_HEADER {
  Type (i) = **0x10..0x15 / 0x18..0x1D / 0x30..0x35 / 0x38..0x3D,**
  Track Alias (vi64),
  Group ID (vi64),
  [Subgroup ID (vi64),]
  [Publisher Priority (8),]
}

afrind:
This is correct:

`Type (i) = 0x10..0x15 / 0x18..0x1D / 0x30..0x35 / 0x38..0x3D,`

I made a PR to make it clear in the other table.

PR #1548, opened 2026-03-11
Fix SUBGROUP_HEADER type range in stream type table

afrind:
The line is grotesquely long, but unsure how to fix it.

Fixes: #1547


draft-ietf-moq-transport.md
```diff
@@ -911,7 +911,7 @@ the type of the stream.
 |------------:|:------------------------------------------------|
 | 0x05        | FETCH_HEADER  ({{fetch-header}})                |
 |-------------|-------------------------------------------------|
-| 0x10-0x1D   | SUBGROUP_HEADER  ({{subgroup-header}})          |
+| 0x10-0x15 / 0x18-0x1D / 0x30-0x35 / 0x38-0x3D | SUBGROUP_HEADER  ({{subgroup-header}}) |
```
ianswett, line 914: You could put it on 4 lines?

afrind: I think that will make 4 separate rows in the rendered table

Issue #1519, opened 2026-03-02
Improve design of requests blocking on other requests


vasilvv:
We need to at least add the text about how request ID dependencies interact with timeouts and stream being reset, also add a note that one cannot rely on those because of this.

I also have an alternative design I would like us to consider, quoting from #1389:
> * A message `WAIT_FOR x` causes the receiver to stop reading control stream until `x` is triggered, or timeout occurs.
> * A message `UNBLOCK x` causes all `WAIT_FOR x` messages to unblock.
> * Messages can occur on any control stream.
> * We need some text about state tracking.

The main advantage is that it decouples the blocking from request processing, and you can move it to the parser level (where it arguably belongs).

afrind:
Consolidating from #1603:

From Cullen's post to the mailing list regarding use-cases:

---
1) Swap tracks. In a video conference, a subscriber is subscribe to track for Alice and Bob’s video and is watching Alice with Bob paused, but wants to pause Alice and unpause the track with Bob. It’s pretty common to want to ensure to pause the current one before unpausing the new track to not have congestion.

2) client side ABR

3) a stream being paused and unpaused in rapid succession and requests reorders on the wire resulting in the opposite of the desired state.
---

I'll note that #3 is already handled because REQUEST_UPDATE for a given subscription happens on a single control stream.

Could someone add more color on the specific ordering requirements for "Client Side ABR"?  Is that a special case of 1 (Swap tracks) when the tracks have the same content?  To what extent would SWITCH handle both?

ianswett:
Some approaches for ABR:
1) Subscribe to the new Track at a specific group and also end the current subscription. (could be downswitch or changing audio tracks or ?)
2) Subscribe to the new Track and ensure you actually have enough bandwidth before ending the current subscription. (upswitch)
3) Use SWITCH or DTS

For 1, I think the potential use case is "I really want the Subscribe to succeed before I end the existing subscription"?  But if it fails completely, do we still execute the REQUEST_UPDATE to set forward=0?  What if the control message succeeds quickly, but the actual data is slow to arrive?  You definitely have an active subscription.  No content, but you know what Largest Object is.

For the "pause before unpause" the case that I believe is potentially problematic is they're in two separate packets and the pause gets lost. If you send them in a single packet, I believe you'd achieve the desired outcome?  ie: It's still ok if the two messages were processed in parallel.  It might take longer to start sending the new track than pause the existing subscription anyway.

afrind:
[...]

PR #1559, opened 2026-03-15
Add Request ID to GOAWAY

### LINKED ISSUE #1549: Draft-17: Handle redirection before session establishment on the control stream
In draft-ietf-moq-transport-17, https://www.ietf.org/archive/id/draft-ietf-moq-transport-17.html#name-unidirectional-stream-types
the first message on the control stream MUST be a SETUP(Type=0x2F00) message.

However, consider the following scenario:
A server receives the client's SETUP request and determines that the session itself is valid, but the client should connect to another server instead (for example, due to high CPU load, load balancing, or deployment topology).
In the current specification, the server need send two messages:

  1. SETUP
  2. GOAWAY

This results: client set to an established state when receive the server SETUP, and then immediately close the session when receive the GOAWAY. This behavior introduce unnecessary state transitions and complexity in implementations.

In https://www.ietf.org/archive/id/draft-ietf-moq-transport-17.html#name-termination:
Section "Termination" allows the server to close the session if a specific error occurs (for example UNAUTHORIZED). However, in the scenario described above, the SETUP itself is valid and successful, but the server prefers the client to connect to another server.

Proposed change:
1. allow send GOAWAY as the first message on the control stream
2. Alternatively, add new SETUP Option to indicate a new URI(like GOAWAY)

Would the working group consider the proposed change?

afrind:
This mirrors HTTP and allows the receiver (clients, mostly) to know if the peer definitely didn't act on a given request.  This can be factored into the decision about whether or how to retry the request, and can fast-fail requests in flight.

Fixes: #1549

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2435,6 +2435,7 @@ GOAWAY Message {
   New Session URI Length (vi64),
   New Session URI (..),
   Timeout (vi64),
+  Request ID (vi64),
 }
 ~~~
 {: #moq-transport-goaway-format title="MOQT GOAWAY Message"}
@@ -2455,6 +2456,16 @@ GOAWAY Message {
   0 indicates the sender has no specific timeout, and the recipient SHOULD still
   close the session as quickly as possible. This is a hint; the sender of the
   GOAWAY MAY close the session before the indicated timeout has elapsed.
+
+* Request ID: The smallest peer Request ID that was not or might not have been
+  processed prior to sending the GOAWAY. If no requests have been processed,
+  this is 0 (at a server) or 1 (at a client). If the parity of the Request ID
+  does not match the receiver's parity, the endpoint MUST close the session with
+  `INVALID_REQUEST_ID`. Requests with a Request ID equal to or greater than the
+  indicated value, as well as any requests that arrive after the GOAWAY, MUST be
+  rejected with REQUEST_ERROR using error code GOING_AWAY. Requests with a
+  Request ID less than the indicated value were or might have been processed;
+  their status can be determined from the response on each request stream.
 
 ## REQUEST_OK {#message-request-ok}
 


draft-ietf-moq-transport.md
```diff
@@ -2456,6 +2457,16 @@ GOAWAY Message {
   close the session as quickly as possible. This is a hint; the sender of the
   GOAWAY MAY close the session before the indicated timeout has elapsed.
 
+* Request ID: The smallest peer Request ID that was not or might not have been
+  processed prior to sending the GOAWAY. If no requests have been processed,
```
ianswett, line 2461:
```suggestion
* Request ID: The smallest peer Request ID that was not
processed prior to sending the GOAWAY. If no requests have been processed,
```

ianswett (approved):

draft-ietf-moq-transport.md
afrind: This came straight from HTTP.  I originally had "that had not initiated processing", since processed implies it also _finished_.

vasilvv:
I'm a bit sad since this further increases the use of request IDs, which I was hoping to get rid of in #1519, but this does sound like a worthwhile feature.

afrind:
I also need them in MOQPACK: https://datatracker.ietf.org/doc/draft-frindell-moq-moqpack/

vasilvv (approved):

suhasHere (approved):

draft-ietf-moq-transport.md
afrind: HTTP cargo cult FTW

ianswett:
Thinking about this more, I'm not actually sure this is very useful in practice?  The only case I can think of is if you sent a GOAWAY while a SUBSCRIBE or other REQUEST was in flight, the Subscriber would find out slightly faster that its SUBSCRIBE was going to be rejected, but if GOAWAY is for graceful shutdowns, I wouldn't expect a sub-RTT optimization to be worthwhile?

PR #1560, opened 2026-03-15
Close session when delta encoding wraps

afrind:
Add overflow checks for Message Parameter Type Delta, End Group Delta, and Object ID Delta fields. When delta encoding results in a value greater than 2^64 - 1, the endpoint MUST close the session with PROTOCOL_VIOLATION, consistent with the existing Key-Value-Pair check.

Fixes: #1552


afrind:
Marking editorial since it was clearly a problem before and this text existed for other delta encoding.

PR #1561, opened 2026-03-15
Don't close the Session for unknown errors

### LINKED ISSUE #1175: Handling of non-defined error codes

### LINKED ISSUE #1231: Are error codes best-effort or mandatory?
I have the impression that there is currently a lack of consensus on whether error codes, or errors in general, may/should be explicitly handled. See related discussion in #1175 and #1162. There should be an explicit discussion about this.

Should recoverable, i.e., explicitly handleable and/or retryable states, be expressed using
a) an error and a specific code, or
b) should those states always be expressed using a different mechanism?

See [#1036 FETCH_ERROR + No Objects](https://github.com/moq-wg/moq-transport/issues/1036) as an example of this.

I don’t have a strong personal preference, but I believe the WG should reach a clear consensus and update the draft accordingly. The current draft is, in my view, not entirely clear on this point, as shown by the number of recent comments about error handling.

I want to note that once the spec defines a set of error codes, it is inevitable, in my opinion, that implementations in the wild will match on them and build specific handling around them.

ianswett:
Fixes #1231 
Fixes #1175

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1104,9 +1104,6 @@ tasks.  For example, a limited media player could operate using only SUBSCRIBE
 related messages.  Limited endpoints SHOULD respond to any unsupported messages
 with the appropriate `NOT_SUPPORTED` error code, rather than ignoring them.
 
-Endpoints MUST NOT close the session when receiving an unknown error code for
-REQUEST_ERROR or SUBSCRIBE_DONE.
-
 Relays MUST implement all MOQT messages defined in this document, as well as
 processing rules described in {{relays-moq}}.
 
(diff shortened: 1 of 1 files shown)


afrind:
I think we already say this in the new grease section?

draft-ietf-moq-transport.md
```diff
@@ -1095,7 +1095,7 @@ in order to obtain an accurate minimum RTT. Similarly, Reno halves it's congesti
 window upon detecting loss.  In both cases, the large reduction in sending rate might
 cause issues with latency sensitive applications.
 
-# Modularity
+# Extensibility
```
afrind, line 1098: This actually is about modularity (what can a moqt v1 client choose not to implement), but meh.

ianswett: I wasn't sure whether modularity was well understood, but I don't have a strong opinion.

sharmafb (approved):

suhasHere (approved):

Issue #1507, opened 2026-02-14
FEATURE REQUEST: Mechanism to get the sender's bitrate


kixelated:
For client-side ABR, it's extremely helpful to get the sender's estimated bitrate straight from the congestion controller. Otherwise, the subscriber has to guess based on media timestamps and is oblivious to any QUIC-level probing.

I think we should have an equivalent to CMSD. A periodic report from the publisher as to the current network conditions. This would be on a per-hop basis, so it needs to be in MoqTransport and not a higher layer.

@wilaw @gwendalsimon @acbegen

kixelated:
I'm implementing this in moq-lite and it's difficult to determine frequently these updates should be sent. Maybe it could be opt-and configurable via a PROBE request, used to indicate if/what bitrate the server (publisher) should probe to.

@vasilvv @ianswett if you have any ideas. This is probably stand-alone enough for an extension.

wilaw:
+1 to having this, to assisting clients with making ABR decisions. 

One solution is for the client to issue a PROBE message. This is necessary if your throughput estimate is application limited (you are playing 32 kbps audio and you want to know if you can play 4k video). The PROBE response could also return the throughput (goodput) estimate. 

Another option would be for the relay to append an extension to each object it sends to that client, each holding the current throughput estimate for the connection. This would give the client a near constant stream of throughput estimates, without having to ask for each one.

It could be a flag on SUBSCRIBE/FETCH  and could be turned on/off using REQUEST_UPDATE.

suhasHere:
Something like this is useful. I would propose a feedback track 
" ./well-known/<connection-specific-id>/feedback"
 at each relay and define payload types such as BitRateEstimation and a mechanisms of its periodicity and so on. 

Such a design can be extended to get other send side / relay side measurements that might be useful for end-subscribers and probably between relays too.

acbegen:
We had implemented this before the Seattle interim and even showed a demo, if anybody recalls. We had a probe track that a subscriber could join to get updates. The carriage of the information is not really the issue here. As suggested above, each object can carry a tiny extension for this purpose (so no need for another track). The bigger issue is determining the value to report in this extension or the new track. 

@kixelated what info are you suggesting to put in there? Something obtained from the QUIC layer or something you measure through probing (which is not trivial)?

Also, I don't think we need this carriage per hop between the original publisher and end subscriber. The last mile is sufficient for this (just like we use with CMSD), as subscriber-side ABR should occur only there.

[...]

PR #1562, opened 2026-03-15
RFC: Add Session-Level Tracks reserved namespace

afrind:
Reserve the `.session` namespace tuple[0] for session-level tracks that are published by the MOQT session itself, not the application. This enables extending transport functionality using existing subscription and object delivery machinery. Relays MUST NOT forward these subscriptions, and unrecognized session-level tracks MUST be rejected with NOT_SUPPORTED.

Establishes an IANA registry for session-level track names under the `.session` namespace with Specification Required registration policy.

Useful for #1507

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -862,15 +862,19 @@ New versions of MOQT MUST specify which existing extensions can be used with
 that version. New extensions MUST specify the existing versions with which they
 can be used.
 
+### Reserved Namespaces {#reserved-namespaces}
+
+MOQT reserves all Track Namespace values whose first tuple field begins with
+a period (0x2e, `.`) for use by MOQT and its extensions.
+
 ### Session-Level Tracks {#session-level-tracks}
 
-MOQT reserves the namespace tuple field value `.session` (the bytes 0x2e,
-0x73, 0x65, 0x73, 0x73, 0x69, 0x6f, 0x6e) in the first position of the
-Track Namespace for session-level tracks. Session-level tracks are published
-by the MOQT session itself, not by the application. They provide a mechanism
-for extending MOQT transport functionality using existing subscription and
-object delivery machinery, without defining new control messages or stream
-types.
+MOQT defines the `.session` namespace (the bytes 0x2e, 0x73, 0x65, 0x73,
+0x73, 0x69, 0x6f, 0x6e) in the first position of the Track Namespace for
+session-level tracks. Session-level tracks are published by the MOQT session
+itself, not by the application. They provide a mechanism for extending MOQT
+transport functionality using existing subscription and object delivery
+machinery, without defining new control messages or stream types.
 
 Applications MUST NOT publish tracks with a Track Namespace whose first
 field is `.session`. Relays MUST NOT forward subscriptions for session-level
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -4407,7 +4407,8 @@ Each registration must include:
 
 | Field | Description |
 |:------|:------------|
-| Track Name | The track name (bytes) under the `.session` namespace |
+| Track Namespace | The track namespace under the `.session` namespace, can be empty |
+| Track Name | The track name (bytes) within the full namespace |
 | Description | Brief description of the track's purpose |
 | Change Controller | Who may update the registration |
 | Specification | Reference to the defining specification |
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -865,27 +865,41 @@ can be used.
 ### Reserved Namespaces {#reserved-namespaces}
 
 MOQT reserves all Track Namespace values whose first tuple field begins with
-a period (0x2e, `.`) for use by MOQT and its extensions.
+a period (0x2e, `.`). These namespaces MUST NOT be used unless their meaning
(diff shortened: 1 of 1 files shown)


ianswett (approved):

sharmafb (approved):

draft-ietf-moq-transport.md
```diff
@@ -4373,6 +4396,24 @@ the length field.
...
+This document establishes a registry for session-level track names
+under the `.session` namespace (see {{session-level-tracks}}). The
+registration policy is Specification Required (per {{!RFC8126,
+Section 4.6}}).
+
+Each registration must include:
+
+| Field | Description |
+|:------|:------------|
+| Track Name | The track name (bytes) under the `.session` namespace |
```
suhasHere: We also need namespace tuples

draft-ietf-moq-transport.md
```diff
@@ -862,6 +862,28 @@ New versions of MOQT MUST specify which existing extensions can be used with
...
+field is `.session`. Relays MUST NOT forward subscriptions for session-level
+tracks to other sessions.
+
+An endpoint that receives a SUBSCRIBE or FETCH for an unrecognized
+session-level track MUST reject the request with REQUEST_ERROR using error
+code NOT_SUPPORTED rather than passing it to the application.
+
+The track names available under the `.session` namespace are defined by
+extensions to this specification and registered with IANA (see
+{{iana-session-level-tracks}}).
```
suhasHere, line 903: we need a way for publisher to reject this track ?

afrind: See the preceding paragraph:

> An endpoint that receives a SUBSCRIBE or FETCH for an unrecognized
session-level track MUST reject the request with REQUEST_ERROR using error
code NOT_SUPPORTED rather than passing it to the application.

draft-ietf-moq-transport.md
```suggestion
| Track Namespace | The track namespace under the `.session` namespace, can be empty |
| Track Name | The track name (bytes) within the full namespace |
```

afrind:
Discussed 3/30:

1) reserve all tuples starting with . for MOQT use
2) reserve .session for session specific - and MOQT MUST error without delivering to application if not understood
3) IANA registrations might be heavy handed - supported tracks could be negotiated via Setup Options

draft-ietf-moq-transport.md
```diff
@@ -862,6 +862,32 @@ New versions of MOQT MUST specify which existing extensions can be used with
 that version. New extensions MUST specify the existing versions with which they
 can be used.
 
+### Reserved Namespaces {#reserved-namespaces}
+
+MOQT reserves all Track Namespace values whose first tuple field begins with
+a period (0x2e, `.`) for use by MOQT and its extensions.
+
+### Session-Level Tracks {#session-level-tracks}
+
```
suhasHere, line 880: This is fine, but the spec needs to specify what to expect from such a track 

Say a [.,session], "" ftn

afrind: We talked about it in the call but couldn’t think what to say. Essentially we’re telling people not to use these names, because we might define them in the future.

@suhasHere do you have an alternate suggestion?

draft-ietf-moq-transport.md
```diff
@@ -862,6 +862,32 @@ New versions of MOQT MUST specify which existing extensions can be used with
...
+transport functionality using existing subscription and object delivery
+machinery, without defining new control messages or stream types.
+
+Applications MUST NOT publish tracks with a Track Namespace whose first
+field is `.session`. Relays MUST NOT forward subscriptions for session-level
+tracks to other sessions.
+
+An endpoint that receives a SUBSCRIBE or FETCH for an unrecognized
+session-level track MUST reject the request with REQUEST_ERROR using error
+code NOT_SUPPORTED rather than passing it to the application.
```
ianswett: I think DOES_NOT_EXIST is a better error code?

```suggestion
code DOES_NOT_EXIST rather than passing it to the application.
```

ianswett (approved):
One nit, but still LG

draft-ietf-moq-transport.md
```diff
@@ -862,6 +862,32 @@ New versions of MOQT MUST specify which existing extensions can be used with
 that version. New extensions MUST specify the existing versions with which they
 can be used.
 
+### Reserved Namespaces {#reserved-namespaces}
+
+MOQT reserves all Track Namespace values whose first tuple field begins with
+a period (0x2e, `.`) for use by MOQT and its extensions.
```
vasilvv: nit: it can also be used by the applications, as long as they reserve it.

vasilvv (approved):
lgtm modulo a nit

vasilvv:
Notes based on editors' call.

I assume we are reserving any track namespace where the first element is `.session`.  What does this means for the track namespace that is exactly `.session` (one element)?

Also, do we want to caveat anything for the track namespaces where the first element is exactly `.`?  It is technically reservable from the current text.

afrind:
> Notes based on editors' call.
> 
> I assume we are reserving any track namespace where the first element is `.session`. What does this means for the track namespace that is exactly `.session` (one element)?

People can reserve track names in .session with IANA.

> Also, do we want to caveat anything for the track namespaces where the first element is exactly `.`? It is technically reservable from the current text.

Also reserve bare '.'namespace and all subnamespaces/tracks to mean nothing/forbidden.

suhasHere (approved):

Issue #1453, opened 2026-02-02
Send Rate parameter


wilaw:
Standard congestion control algorithms download data as fast as possible. Since throughput between client and server is typically much higher than the encoded rate of the content (Netflix report an average of 13x higher), this leads to bursty downloads at either the Object or Group . Bustiness leads to buffer bloat and packet loss. Instead of downloading data as fast as possible, a better goal for CC is to download as fast as needed. In the ABR content, the "as needed" refers to a rate which prevents the player's ABR algorithm from switching down and simultaneously allows it to switch up. In their testing, a rate of 3.5x encoded birtate satisfied these goals

Delivering "as fast as needed" produced performance benefits (in Netflix study presented at MHV26 Conference by TY) in throughput, retrans and RTT. 

![Image](https://github.com/user-attachments/assets/4b0c56fd-0ab8-40d2-b1dd-f7f0ba188b13)

To allow similar gains in MOQT, we should enable a new "send rate" parameter to accompany a SUBSCRIBE or FETCH. This parameter specifies the rate at which the client wishes to have the content sent by the relay.. The value of the parameter is the rate in kbps at which the content should be paced. A client can adjust the send rate using a REQUEST_UPDATE message. A value of 0 has special meaning and it disables any pacing.

LPardue:
The SCONE WG has investigated significant effort in defining how to communicate "throughput advice". I expect QUIC implementations will be able to support acting on the advice independent of the signal delivery mechanism. 

If MoQT wants to take advantage of a feature like this, keeping the model similar to SCONE is likely to reduce friction

afrind:
Do we like "kbps" or rather something more moq-ish like "milliseconds per group".

wilaw:
Per Lucas's [comment](https://github.com/moq-wg/moq-transport/issues/1453#issuecomment-3837637787) , SCONE defined a logarithmic scale with a 6-bit or 7-bit Rate Signal field. The bitrate is determined by a formula where the signal value (n) acts as an exponent. For the most common version (0xSCONE1):

Bitrate = b_min \times 10^{n/20}
b_min (Base Rate): 100 Kbps.\
n: The value provided in the SCONE packet header.
Range: This allows the protocol to signal everything from 100 Kbps (n=0) up to ~199.5 Gbps (n=126).

Rate Signal (n) | Calculated Bitrate
0 | 100 Kbps
20 | 1 Mbps
40 | 10 Mbps
60 | 100 Mbps
[...]

PR #1565, opened 2026-03-15
RFC: Add SEND_RATE parameter for paced FETCH delivery

afrind:
Add a SEND_RATE Message Parameter for FETCH and REQUEST_UPDATE that allows a subscriber to request the publisher pace Object delivery at a specified rate in kbps. This reduces bufferbloat and packet loss for VoD/catch-up content where the congestion controller would otherwise burst data far above the needed rate. A value of 0 or absence of the parameter means no pacing constraint.

Fixes: #1453

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2303,9 +2303,9 @@ The subscriber can update the send rate during an active fetch by
 sending a REQUEST_UPDATE with a new SEND_RATE value.
 
 The publisher SHOULD pace Object delivery to approximately the requested
-rate.  The publisher MAY exceed the requested rate in short bursts, for
-example at the beginning of a Group, to allow the subscriber to build
-an initial buffer.  The publisher MUST NOT use SEND_RATE to pace below
+rate.  The rate limit is a request only and is not guaranteed. The publisher MAY
+exceed the requested rate in short bursts.  The publisher MUST NOT use SEND_RATE
+to pace below
 the congestion controller's minimum rate; the congestion controller's
 behavior takes precedence.
 


draft-ietf-moq-transport.md
```diff
@@ -2290,6 +2290,31 @@ to the Largest Group, it does not send a NEW_GROUP_REQUEST upstream.
...
+constraint; the publisher delivers as fast as the congestion controller
+allows.  This is the default behavior when the parameter is absent.
+
+The subscriber can update the send rate during an active fetch by
+sending a REQUEST_UPDATE with a new SEND_RATE value.
+
+The publisher SHOULD pace Object delivery to approximately the requested
+rate.  The publisher MAY exceed the requested rate in short bursts, for
+example at the beginning of a Group, to allow the subscriber to build
+an initial buffer.  The publisher MUST NOT use SEND_RATE to pace below
```
wilaw: if the subscriber wants a different burst rate, it will ask for it. The publisher should never attempt to anticipate a buffering behavior on the subscriber. Exceeding the requested rate for short bursts should be allowed in order to simplify the senders rate-control implementation. 
```suggestion
rate.  The rate limit is a request only and is not guaranteed. The publisher MAY
exceed the requested rate in short bursts.  The publisher MUST NOT use SEND_RATE
to pace below
```

draft-ietf-moq-transport.md
```diff
@@ -2290,6 +2290,31 @@ to the Largest Group, it does not send a NEW_GROUP_REQUEST upstream.
 After sending a NEW_GROUP_REQUEST upstream, the request is considered
 outstanding until the Largest Group increases.
 
+### SEND RATE Parameter {#send-rate}
+
+The SEND_RATE parameter (Parameter Type 0x34) is a varint.  It MAY
+appear in a FETCH or REQUEST_UPDATE for a fetch.  The value is the rate
+in kilobits per second at which the publisher SHOULD pace the delivery
+of Objects in the response.  A value of 0 indicates no pacing
```
wilaw, line 2298: Do we need to define a timebase over which this rate is calculated?

wilaw (commented):
Do we need to define a timebase over which the rate is calculated?

draft-ietf-moq-transport.md
afrind: The timebase (e.g. for scone) is needed because of enforcement.  Since this is just a request I think the sender can just do their best to accommodate.

draft-ietf-moq-transport.md
```diff
@@ -2290,6 +2290,31 @@ to the Largest Group, it does not send a NEW_GROUP_REQUEST upstream.
...
+allows.  This is the default behavior when the parameter is absent.
+
+The subscriber can update the send rate during an active fetch by
+sending a REQUEST_UPDATE with a new SEND_RATE value.
+
+The publisher SHOULD pace Object delivery to approximately the requested
+rate.  The rate limit is a request only and is not guaranteed. The publisher MAY
+exceed the requested rate in short bursts.  The publisher MUST NOT use SEND_RATE
+to pace below
+the congestion controller's minimum rate; the congestion controller's
```
ianswett, line 2309: I'm not sure we need this sentence.

draft-ietf-moq-transport.md
ianswett, line 2298:
```suggestion
of Objects in the response stream. Because the rate applies to the unidirectional
Fetch response stream, it includes Object framing including Object Properties.
A value of 0 indicates no pacing
```

draft-ietf-moq-transport.md
```diff
@@ -2290,6 +2290,31 @@ to the Largest Group, it does not send a NEW_GROUP_REQUEST upstream.
...
+behavior takes precedence.
+
+Relay Handling:
+
+A Relay that receives a FETCH with SEND_RATE and needs to make an
+upstream FETCH to satisfy the request MAY use the SEND_RATE value when
```
ianswett, line 2315: I'm not sure how to say it, but I don't think we want to encourage copying the SEND_RATE verbatim.

martinduke:
I hate this. The whole reason we created FETCH was to provide an easy mechanism for flow control via the single stream. There is already a single, well-understood mechanism and layering another poorly understood mechanism on top of it is going to result in weird interactions and unintended consequences.

If we let SUBSCRIBE go into the past, that's different, but that's for SUBSCRIBE.

suhasHere:
My initial fetch proposal had DeliveryRate to address the same idea. However, we reverted it as was not very useful for someone trying to catch up to fill the buffers. I couldn't convince myself where it gets really used. I am open to understanding the use-case though

afrind:
>  I am open to understanding the use-case though

Read the linked issue?  In many cases, you don't want to fetch media significantly faster than you can consume/play it.  Yes the subscriber sort-of controls this via flow control release on a single stream.  Speculative Pre-fetch is another potential use case.

martinduke:
On 3/30 you remarked we might need this because of REWIND, but that is neither here nor there, as this is a FETCH parameter.

Maybe it comes back as a SUBSCRIBE parameter with REWIND, but I'm not yet convinced.

martinduke:
Why I think not even REWIND needs it:
- @ianswett made the point that we probably want to control this on the publisher side, and REWIND gives the publisher enough discretion to not blow up the subscriber.
- the subscriber can set a delivery timeout that prevents huge buildup of data

afrind:
3/30: No appetite to add this as FETCH param

PR #1566, opened 2026-03-15
Clarify Object existence and cross-source contradictions

### LINKED ISSUE #1376: Text on reordering and Objects going from DNE to Exists is incorrect
While writing #1342, it was observed that the following text doesn't seem correct anymore.

"Note that due to reordering, an implementation can receive an Object after
receiving an indication that the Object in question does not exist.  The
endpoint SHOULD NOT cache or forward the object in this case."

This text might not be needed?

afrind:
Clarify that all sources of Object non-existence (FETCH gaps, Object Status, Prior Gap Properties) are authoritative, and that the "does not exist" state is permanent for the endpoint that learned it.

Narrow malformed Track conditions 7 and 8 to apply within a single subscription or FETCH response. Contradictions across different sources (e.g., a FETCH gap followed by delivery via subscription) are expected due to timing and are not malformed — the Object is silently dropped.

Fixes: #1376

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -496,13 +496,19 @@ Objects within a Group are in ascending order by Object ID.
 From the perspective of a subscriber or a cache, an Object can be in three
 possible states:
 
-1. The Object is known to not exist. This state is permanent. MOQT has multiple
-   ways to communicate that a certain range of objects does not exist,
-   including the Object Status field, and the use of gaps in FETCH responses.
+1. The Object is known to not exist. This state is permanent for the endpoint
+   that learned it. MOQT has multiple ways to communicate that a certain range
+   of Objects does not exist, including the Object Status field, gaps in FETCH
+   responses, and Prior Gap Properties.  All of these are authoritative.
 2. The Object is known to exist. From this state, it can transition to not
    existing, but not vice versa.
-3. The state of the Object is unknown, either because it has not been yet
+3. The state of the Object is unknown, either because it has not yet been
    received, or it has not been produced yet.
+
+Due to timing, an endpoint can receive an Object after it has already recorded
+that the Object does not exist (e.g., via a FETCH gap from one source and later
+delivery via a subscription).  This is not a protocol error and the Track is not
+malformed.
 
 Whenever the publisher communicates that certain objects do not exist, this
 fact is expressed as a contiguous range of non-existent objects and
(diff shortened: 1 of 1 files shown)


ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -496,14 +496,20 @@ Objects within a Group are in ascending order by Object ID.
 From the perspective of a subscriber or a cache, an Object can be in three
 possible states:
 
-1. The Object is known to not exist. This state is permanent. MOQT has multiple
-   ways to communicate that a certain range of objects does not exist,
-   including the Object Status field, and the use of gaps in FETCH responses.
+1. The Object is known to not exist. This state is permanent for the endpoint
+   that learned it. MOQT has multiple ways to communicate that a certain range
```
vasilvv: > for the endpoint that learned it

This is redundant, since we already say above that this is perspective "of a subscriber or a cache".

draft-ietf-moq-transport.md
```diff
@@ -496,14 +496,20 @@ Objects within a Group are in ascending order by Object ID.
...
+   of Objects does not exist, including the Object Status field, gaps in FETCH
+   responses, and Prior Gap Properties.  All of these are authoritative.
```
vasilvv: > MOQT has multiple ways to communicate that a certain range [...]

This sentence is the same statement as the paragraph at the bottom of the section.  The authoritative part is new, though, and is useful.

draft-ietf-moq-transport.md
```diff
@@ -496,14 +496,20 @@ Objects within a Group are in ascending order by Object ID.
...
 2. The Object is known to exist. From this state, it can transition to not
    existing, but not vice versa.
-3. The state of the Object is unknown, either because it has not been yet
+3. The state of the Object is unknown, either because it has not yet been
    received, or it has not been produced yet.
 
+Due to timing, an endpoint can receive an Object after it has already recorded
```
vasilvv: "timing" is a bit vague, maybe "Since the objects can be delivered out of order" or something?

draft-ietf-moq-transport.md
```diff
@@ -675,15 +681,14 @@ include:
...
-   Status END_OF_TRACK or the last Object sent in a FETCH whose response indicated
-   End of Track.
+7. Within a single subscription, an Object is received in a Group whose Object
+   ID is larger than the final Object in the Group.  The final Object in a Group
+   is the Object with Status END_OF_GROUP, or the last Object before a FIN in a
+   Subgroup which has the END_OF_GROUP bit set.
+8. An Object is received on a Track whose Group and Object ID are larger than
+   the final Object in the Track.  The final Object in a Track is the Object
+   with Status END_OF_TRACK or the last Object sent in a FETCH whose response
+   indicated End of Track.
```
vasilvv, line 691: I don't remember what we discussed before about this, but isn't this somewhat in contradiction with requiring to ignore non-existing objects?

suhasHere (approved):

vasilvv (approved):

fluffy:
I don't think this is correct. 

I do not think there are any valid situations where a relay or subscriber can be get a definitive answer that an object does not exist and then later find out the object does exist. Similarly there is no situations where all the entities are correct following the protocol and get and object exists and later get a result that the object does not exist.

afrind:
> I do not think there are any valid situations where a relay or subscriber can be get a definitive answer that an object does not exist and then later find out the object does exist.

The scenario has been documented in at least one issue and wg presentations when it was discussed in the last 6 months but I'll add it here also.

The track has ephemeral objects that live for 500ms, then no longer exist.

Time T: Publisher publishes an object via subscription -- delayed in the network
Time T+600: The relay sees the object is missing, issues a FETCH.  The publisher has discarded the object -> signals non-existence
Time T+650: The original object arrives

>  Similarly there is no situations where all the entities are correct following the protocol and get and object exists and later get a result that the object does not exist.

Also incorrect.  See above -- a track with ephemeral objects meets the criteria.

PR #1567, opened 2026-03-16
An alternative approach to Mandatory Track Properties

### LINKED ISSUE #1480: Have a range of mandatory to implement track extensions

ianswett:
Fixes #1480 

An alternative to #1509

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -774,6 +774,23 @@ applications using the same code point in these ranges may assign different
 meanings; the interpretation depends on the track or application
 context known to the publisher and subscriber.
 
+## Mandatory Track Properties {#mandatory-track-properties}
+
+An extension can define Mandatory Track Properties with special
+handling rules that prevent Tracks from being forwarded to endpoints that
+do not understand the Properties.
+
+When a Publisher receives Track Properties (in PUBLISH, SUBSCRIBE_OK, or
+FETCH_OK messages) containing a Mandatory Track Property type that the
+Subscriber does not understand, it MUST NOT forward that track:
+
+* PUBLISH messages, including those as a result of SUBSCRIBE_NAMESPACE,
+  MUST NOT be sent.
+
+* A Publisher MUST respond to SUBSCRIBE and FETCH requests for the Track
+  with REQUEST_ERROR with error code UNSUPPORTED_EXTENSION if the
+  Subscriber does not indicate support for the Track Property.
+
 # Sessions {#session}
 
 ## Session establishment {#session-establishment}


draft-ietf-moq-transport.md
```diff
@@ -774,6 +774,23 @@ applications using the same code point in these ranges may assign different
 meanings; the interpretation depends on the track or application
 context known to the publisher and subscriber.
 
+## Mandatory Track Properties {#mandatory-track-properties}
+
+An extension can define Mandatory Track Properties with special
+handling rules that prevent Tracks from being forwarded to endpoints that
+do not understand the Properties.
+
+When a Publisher receives Track Properties (in PUBLISH, SUBSCRIBE_OK, or
```
afrind, line 783: Publishers don't receive properties, they send them.  Do you mean Relays?

draft-ietf-moq-transport.md
```diff
@@ -774,6 +774,23 @@ applications using the same code point in these ranges may assign different
...
+FETCH_OK messages) containing a Mandatory Track Property type that the
+Subscriber does not understand, it MUST NOT forward that track:
+
+* PUBLISH messages, including those as a result of SUBSCRIBE_NAMESPACE,
+  MUST NOT be sent.
```
afrind, line 788: A little awkward to skip sending the publish and there's no way for the subscriber to know.

draft-ietf-moq-transport.md
```diff
@@ -774,6 +774,23 @@ applications using the same code point in these ranges may assign different
 meanings; the interpretation depends on the track or application
 context known to the publisher and subscriber.
 
+## Mandatory Track Properties {#mandatory-track-properties}
+
+An extension can define Mandatory Track Properties with special
+handling rules that prevent Tracks from being forwarded to endpoints that
+do not understand the Properties.
```
afrind, line 781: This requires that all MTI properties need to have corresponding SETUP Options identifying them, which I guess is correct?  Maybe need to say it explicitly.

Also, what if a relay doesn't understand a property but the publisher screwed up and sent it anyways.  Now it will keep propagating until it hits a relay/subscriber that does understand.

draft-ietf-moq-transport.md
ianswett: Agreed.  Is this not an issue we'd want to deal with even if we reserved a range?

draft-ietf-moq-transport.md
ianswett: Good point that it'll keep getting propagated if there's a mistake.

It's possible that reserving a range is better, but it feels a bit odd to me to make the enforcement entirely driven by the receiver and not have the sender/publisher at least try to avoid Subscriptions where it's sending a MTI property to someone who doesn't understand it?

vasilvv:
I am extremely confused by what this PR is trying to do.  It says that mandatory properties exist, but does not actually tell you how to figure out what those are.

ianswett:
Closed in favor of an updated version of #1509

PR #1571, opened 2026-03-18
RFC: Add fragment identifier support for moqt URIs

afrind:
NOTE TO EDITORS: Target correct branch for merge

Define a fragment identifier syntax for moqt URIs where the fragment begins with a registered type identifier followed by a colon and a type-specific value. Fragments are processed client-side and not sent to the server. Add an IANA registry for MOQT URI Fragment Types.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -833,6 +833,9 @@ syntax as `https` URIs.
 
 ### Fragment Identifiers {#moqt-fragment}
 
+The media type for resources identified by `moqt` URIs is
+`application/moqt` (see {{iana-media-type}}).
+
 Fragment identifiers MAY be used with `moqt` URIs. The fragment is not
 transmitted to the server; it is processed locally by the client after
 establishing the MOQT session.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -4357,6 +4373,21 @@ Change controller: IETF
...
+This document establishes the "MOQT URI Fragment Types" registry. This
+registry governs fragment type identifiers used in `moqt` URI fragments
+as defined in {{moqt-fragment}}.
+
+New fragment type identifiers are registered using the Specification
+Required policy ({{!RFC8126, Section 4.6}}).
+
+Each entry in the registry contains the following fields:
+
+| Fragment Type | Description | Specification |
```
wilaw, line 4459: Is fragment type case sensitive?

Which characters are allowed?

draft-ietf-moq-transport.md
```diff
@@ -824,6 +824,27 @@ URI can be converted to an `https` URI by replacing the scheme (see
...
+
+Fragment identifiers MAY be used with `moqt` URIs. The fragment is not
+transmitted to the server; it is processed locally by the client after
+establishing the MOQT session.
+
+A `moqt` URI fragment MUST begin with a registered fragment type
+identifier, followed by a colon (`:`), followed by a type-specific value:
+
+~~~
+moqt://example.com/app#<type>:<value>
```
afrind, line 847: Let's define application/moqt MIME type, then we can make this restriction

Issue #1563, opened 2026-03-15
Track Properties are not conveyed in response to TRACK_STATUS


afrind:
This is kind of sad, since these are probably useful things one might use TRACK_STATUS for.

We could add "Number of Properties + Properties" to REQUEST_OK.  It would just be 0 in response to REQUEST_UPDATE, SUB_NS, PUB_NS.  If we parameterize FETCH_OK fields then we could reuse REQUEST_OK there and it would see use there too.

PR #1576, opened 2026-03-21
Add Track Properties to REQUEST_OK

afrind:
This allows Track Properties to be conveyed in response to TRACK_STATUS requests, which was previously not possible since REQUEST_OK did not include them.

Track Properties are populated in response to TRACK_STATUS messages and empty in response to REQUEST_UPDATE,
SUBSCRIBE_NAMESPACE, and PUBLISH_NAMESPACE.

Fixes: #1563

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2482,7 +2482,9 @@ REQUEST_OK Message {
   length of Track Properties is the remaining length of the message
   after parsing all previous fields. Track Properties are populated in
   response to TRACK_STATUS messages; they are empty in response to
-  REQUEST_UPDATE, SUBSCRIBE_NAMESPACE and PUBLISH_NAMESPACE.
+  REQUEST_UPDATE, SUBSCRIBE_NAMESPACE and PUBLISH_NAMESPACE.  If an
+  endpoint receives Track Properties in response to one of these messages
+  it MUST close the session with a `PROTOCOL_VIOLATION`.
 
 ## REQUEST_ERROR {#message-request-error}
 


vasilvv:
It seems a bit odd to me that we merged TRACK_STATUS_OK or whatever it was called into REQUEST_OK, because it was the same as other messages, but now there's a TRACK_STATUS-specific field in it that is not useful in 3 out of 4 scenarios you'd use that message in.

afrind:
> It seems a bit odd to me that we merged TRACK_STATUS_OK or whatever it was called into REQUEST_OK, because it was the same as other messages, but now there's a TRACK_STATUS-specific field in it that is not useful in 3 out of 4 scenarios you'd use that message in.

Mostly agreed.  As mentioned in the issue, there's a possible universe where we convert FETCH_OK to REQUEST_OK, in which case it's only two out of 4.  We could add a separate message, or maybe have two spellings of REQUEST_OK. Or maybe this is just fine?

ianswett:
I'm ok with this being unused for the time being.  I can imagine it might be used for some of the other three messages, so let's see how that plays out.

draft-ietf-moq-transport.md
```diff
@@ -2470,13 +2470,20 @@ REQUEST_OK Message {
...
 ~~~
 {: #moq-transport-request-ok format title="MOQT REQUEST_OK Message"}
 
 * Parameters: The parameters are defined in {{message-params}}.
 
+* Track Properties : A sequence of Properties. See {{properties}}. The
+  length of Track Properties is the remaining length of the message
+  after parsing all previous fields. Track Properties are populated in
+  response to TRACK_STATUS messages; they are empty in response to
+  REQUEST_UPDATE, SUBSCRIBE_NAMESPACE and PUBLISH_NAMESPACE.
```
vasilvv: What happens if the sender sets track properties on REQUEST_OK for any of those?

```suggestion
  after parsing all previous fields. Track Properties are populated in
  response to TRACK_STATUS messages; they are empty in response to
  REQUEST_UPDATE, SUBSCRIBE_NAMESPACE and PUBLISH_NAMESPACE.  If an
  endpoint receives Track Properties in response to one of these messages
  it MUST close the session with a `PROTOCOL_VIOLATION`.
```

PR #1586, opened 2026-04-01
Make Object ID and Group ID delta encoded in Fetch responses

### LINKED ISSUE #877: Pack the bits

### LINKED ISSUE #1345: Separate the list of reasons for malformed tracks into two lists

ianswett:
Fixes #877 
Fixes #1345


draft-ietf-moq-transport.md
```diff
@@ -667,35 +667,28 @@ to MOQT constraints. Such a Track is considered malformed.  Some example
 conditions that constitute a malformed track when detected by a receiver
 include:
 
-1. An Object is received in a FETCH response with the same Group ID as the
-   previous Object, but whose Object ID is not strictly larger than the previous
-   object.
-2.  In a FETCH response, an Object with a particular Subgroup ID is received, but its
+1.  In a FETCH response, an Object with a particular Subgroup ID is received, but its
```
ianswett: Q: Is this really Fetch specific?  Can't this happen with Subscribe as well?

draft-ietf-moq-transport.md
```diff
@@ -3946,15 +3939,27 @@ the corresponding condition is true.
 
 Bitmask | Condition if set | Condition if not set (0)
 --------|------------------|---------------------
-0x04 | Object ID field is present | Object ID is the prior Object's ID plus one
-0x08 | Group ID field is present | Group ID is the prior Object's Group ID
+0x04 | Object ID Delta is present | Object ID is the prior Object's ID plus one
```
afrind, line 4042: Do we need to make it clear that bit 0x4=0 also works with Ascending/Descending?  The problem with delta encoding groups in general is that the FETCH parser _doesn't know_ the asc/dsc order in-line, it needs to be supplied this info from the request, which is annoying.

draft-ietf-moq-transport.md
```diff
@@ -3946,15 +3939,27 @@ the corresponding condition is true.
...
+0x08 | Group ID Delta is present | Group ID is the prior Object's Group ID
 0x10 | Priority field is present | Priority is the prior Object's Priority
 0x20 | Properties field is present | Properties field is not present
 0x40 | Datagram: ignore the two least significant bits | Decode the Subgroup ID as indicated by the two least significant bits
 
 If the first Object in the FETCH response uses a flag that references fields in
 the prior Object, the Subscriber MUST close the session with a
-`PROTOCOL_VIOLATION`.
+`PROTOCOL_VIOLATION`. The first Object therefore MUST include a Group ID Delta
+and Object ID Delta, and these values are the absolute Group ID and Object ID.
```
afrind: Might read better as the first sentence.

draft-ietf-moq-transport.md
```diff
@@ -3946,15 +3939,27 @@ the corresponding condition is true.
...
+
+If the Group ID Delta field is present on an Object other than the first, the
```
afrind, line 4053: Do we need normative text about what happens when computed group is > MAX or < 0?

draft-ietf-moq-transport.md
afrind: One can publish the same subgroup ID in two different streams and give it two different priorities - so yes.

afrind:
Does this fix 1345?  Seems like there's still one list.

ianswett:
> Does this fix 1345? Seems like there's still one list.

It removes all but one of the FETCH specific reasons for a malformed Track, and I don't think there's a reason to have a separate list for one case.

draft-ietf-moq-transport.md
```suggestion
1.  An Object with a particular Subgroup ID is received, but its
```

draft-ietf-moq-transport.md
ianswett: Object Ids are always ascending, so I don't think there's a problem with 0x4=0?

draft-ietf-moq-transport.md
```diff
@@ -3956,15 +3949,27 @@ the corresponding condition is true.
...
+The first Object MUST include a Group ID Delta and Object ID Delta, and
+these values are the absolute Group ID and Object ID. If the first Object in
+the FETCH response uses a flag that references fields in the prior Object,
+the Subscriber MUST close the session with a `PROTOCOL_VIOLATION`.
+
+If the Group ID Delta field is present on an Object other than the first, the
+Group ID is computed from the Group ID Delta and the prior Object's Group ID.
+If the Group Order is Ascending, the Group ID is the prior Object's Group ID
+plus the Group ID Delta + 1.  If the Group Order is Descending, the Group ID is
+the prior Object's Group ID minus the (Group ID Delta + 1).
```
```suggestion
the prior Object's Group ID minus the (Group ID Delta + 1). If the computed
Group ID would be less than 0 or greater than 2^64-1, the Subscriber MUST
close the Session with a 'PROTOCOL_VIOLATION'.
```

draft-ietf-moq-transport.md
```diff
@@ -3956,15 +3949,29 @@ the corresponding condition is true.
...
+If the Group Order is Ascending, the Group ID is the prior Object's Group ID
+plus the Group ID Delta + 1.  If the Group Order is Descending, the Group ID is
+the prior Object's Group ID minus the (Group ID Delta + 1). If the computed
+Group ID would be less than 0 or greater than 2^64-1, the Subscriber MUST
+close the Session with a 'PROTOCOL_VIOLATION'.
+
+If the Object ID Delta field is present on an Object other than the first in the
+Group, the Object ID is the prior Object's ID plus the Object ID Delta.  When the
+Group ID Delta field is present and therefore the Group ID changes, the Object ID
+is the absolute value of Object ID Delta if present, and 0 otherwise.
```
```suggestion
is the absolute value of Object ID Delta if present, and 0 otherwise. If the computed
Object ID would be greater than 2^64-1, the Subscriber MUST close the Session
with error 'PROTOCOL_VIOLATION'.
```

draft-ietf-moq-transport.md
ianswett: Yes, added.

draft-ietf-moq-transport.md
```diff
@@ -3956,15 +3949,31 @@ the corresponding condition is true.
...
+the FETCH response uses a flag that references fields in the prior Object,
+the Subscriber MUST close the session with a `PROTOCOL_VIOLATION`.
+
+If the Group ID Delta field is present on an Object other than the first, the
+Group ID is computed from the Group ID Delta and the prior Object's Group ID.
+If the Group Order is Ascending, the Group ID is the prior Object's Group ID
+plus the Group ID Delta + 1.  If the Group Order is Descending, the Group ID is
+the prior Object's Group ID minus the (Group ID Delta + 1). If the computed
+Group ID would be less than 0 or greater than 2^64-1, the Subscriber MUST
+close the Session with a 'PROTOCOL_VIOLATION'.
```
```suggestion
close the Session with error 'PROTOCOL_VIOLATION'.
```

draft-ietf-moq-transport.md
```diff
@@ -4031,15 +4024,31 @@ the corresponding condition is true.
...
+If the Group ID Delta field is present on an Object other than the first, the
+Group ID is computed from the Group ID Delta and the prior Object's Group ID.
+If the Group Order is Ascending, the Group ID is the prior Object's Group ID
+plus the Group ID Delta + 1.  If the Group Order is Descending, the Group ID is
+the prior Object's Group ID minus the (Group ID Delta + 1). If the computed
+Group ID would be less than 0 or greater than 2^64-1, the Subscriber MUST
+close the Session with error 'PROTOCOL_VIOLATION'.
+
+If the Object ID Delta field is present on an Object other than the first in the
+Group, the Object ID is the prior Object's ID plus the Object ID Delta.  When the
```
afrind: First object in the group is ambiguous - eg if the group starts at object 0 but the fetch start is object 10.  I think this paragraph could be tighter in explaining the cases.

Group ID Delta Present -> Object ID Delta = [delta from 0]
Group ID Delta Not Present -> Object ID Delta [delta from last object ID in the group]

draft-ietf-moq-transport.md
afrind: I think I meant 0x08 then?  But the current text looks right for Ascending/Descending.  The change is still annoying at least in my implementation.

afrind (approved):

vasilvv (approved):

draft-ietf-moq-transport.md
```diff
@@ -4046,15 +4039,31 @@ the corresponding condition is true.
...
+Group ID Delta field is present and therefore the Group ID changes, the Object ID
+is the absolute value of Object ID Delta if present, and 0 otherwise. If the computed
```
```suggestion
Group ID Delta field is present and therefore the Group ID changes, the Object ID
is the absolute value of Object ID Delta. If the computed
```

draft-ietf-moq-transport.md
```diff
@@ -4046,15 +4039,31 @@ the corresponding condition is true.
...
+If the Group ID Delta field is present on an Object other than the first, the
+Group ID is computed from the Group ID Delta and the prior Object's Group ID.
+If the Group Order is Ascending, the Group ID is the prior Object's Group ID
+plus the Group ID Delta + 1.  If the Group Order is Descending, the Group ID is
+the prior Object's Group ID minus the (Group ID Delta + 1). If the computed
+Group ID would be less than 0 or greater than 2^64-1, the Subscriber MUST
+close the Session with error 'PROTOCOL_VIOLATION'.
+
+If the Object ID Delta field is present on an Object other than the first in the
+Group, the Object ID is the prior Object's ID plus the Object ID Delta.  When the
```
```suggestion
If there is a prior Object in the Group and the Object ID Delta field is present,
the Object ID is the prior Object's ID plus the Object ID Delta.  When the
```

draft-ietf-moq-transport.md
ianswett: PTAL a the suggestions below to see if they're correct?

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -4046,15 +4039,31 @@ the corresponding condition is true.
...
+Group ID Delta field is present and therefore the Group ID changes, the Object ID
+is the absolute value of Object ID Delta if present, and 0 otherwise. If the computed
+Object ID would be greater than 2^64-1, the Subscriber MUST close the Session
+with error 'PROTOCOL_VIOLATION'.
```
```suggestion
When the Group ID Delta field is present, the Object ID is the value of Object ID Delta if
present. When the Group ID Delta field is not present, the Object ID is the prior Object's ID
plus the Object ID Delta if present. If Object ID Delta is not present, the Object ID is the
prior Object's ID plus one, regardless of which group it belongs to. If the computed Object ID
would be greater than 2^64-1, the Subscriber MUST close the Session with error
'PROTOCOL_VIOLATION'.
```

PR #1591, opened 2026-04-02
RFC: Add flow control for Subscriptions

### LINKED ISSUE #869: Difficult to limit resource consumption of a Subscription

ianswett:
Adds MAX_SUB_STREAMS and MAX_SUB_BYTES for flow control and introduces SUBGROUP_RESET.

I did not add Setup Options to configure defaults for new Subscriptions, but I can if people like this.

Fixes #869.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1957,6 +1957,8 @@ new request stream.
 |--------|-----------------------------------------------|----------------|
 | 0xF    | PUBLISH_BLOCKED ({{message-publish-blocked}}) | Request        |
 |--------|-----------------------------------------------|----------------|
+| 0x1F   | SUBGROUP_RESET ({{message-subgroup-reset}})   | Control        |
+|--------|-----------------------------------------------|----------------|
 | 0x2    | REQUEST_UPDATE ({{message-request-update}})   | Request        |
 |--------|-----------------------------------------------|----------------|
 | 0x7    | REQUEST_OK ({{message-request-ok}})           | Request        |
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2768,6 +2808,40 @@ SUBSCRIBE_OK Message {
 
 * Track Properties : A sequence of Properties. See {{properties}}.
 
+## SUBGROUP_RESET {#message-subgroup-reset}
+
+A publisher sends a `SUBGROUP_RESET` message on the Subscription's control
+stream to inform the subscriber of the total number of bytes sent on a subgroup
```
afrind, line 3337: Is this in serialized bytes including SUBGROUP_HEADER, Object metadata and payload, or just payload?  Does it include WebTransport framing?

afrind (commented):
High level comments:

What is the expected interaction on the publisher side when out of flow control?  I assume it's - buffer and deliver the highest priority stuff you can when flow control reopens?

Do you need BLOCKED messages also?

If you add this, is it sufficient to just allow all SUBSCRIBE in the past and remove FETCH?  You would lose implicit gap detection capabilities.

draft-ietf-moq-transport.md
```diff
@@ -2458,6 +2460,44 @@ If the parameter is omitted from REQUEST_UPDATE, the value for the
...
+The MAX_SUB_STREAMS parameter (Parameter Type 0x33) is a varint. It MAY appear
+in SUBSCRIBE, PUBLISH_OK, or REQUEST_UPDATE for a Subscription. It sets or updates
+a limit on the total number of subgroup streams for the Subscription in order
+to provide flow control.
+
+A Publisher MUST NOT open more subgroup streams for that Subscription than the
+limit allows, and the Subscriber SHOULD close the Session with PROTOCOL_VIOLATION
+if detected. The Subscriber SHOULD increase this limit before the Publisher runs
+out of streams by sending a REQUEST_UPDATE with the Parameter. In REQUEST_UPDATE,
+this Parameter specifies the number of additional streams that can be opened.
```
ArcherDTY, line 2792: If the same parameter name has a different meaning in REQUEST_UPDATE, why not just give it a new parameter name? This can avoid confusion.

fluffy (commented):
I think that max-sub-streams is a good idea and not sure about bytes

afrind:
We're going to progress this as an extension draft for now.

PR #1593, opened 2026-04-02
RFC: Allow framing single Objects without Subgroup ID

### LINKED ISSUE #1405: Single Object Subgroups don't need a Subgroup ID

ianswett:
Allows single-object subgroups to be sent on streams with the same header as OBJECT_DATAGRAM, without explicitly encoding a Subgroup ID or an Object Length.

By introducing OBJECT_STREAM and allowing it to use the identical framing structure as OBJECT_DATAGRAM, single-object streams no longer require a Subgroup ID or an explicit Object Payload Length (the payload naturally consists of the remainder of the stream). To safely accomplish this, FETCH_HEADER's stream type identifier was moved from 0x05 to 0x50 to avoid conflicts with the Datagram format type IDs 0x00..0x0F and 0x20..0x2F.

An object received in an OBJECT_STREAM message retains an Object Forwarding Preference of Subgroup. This optimizes both wire space and simplifies single-object per-stream use cases.

Fixes #1405


draft-ietf-moq-transport.md
```diff
@@ -1007,7 +1007,9 @@ the type of the stream.
 |-------------|-------------------------------------------------|
 | ID          | Type                                            |
 |------------:|:------------------------------------------------|
-| 0x05        | FETCH_HEADER  ({{fetch-header}})                |
+| 0x50        | FETCH_HEADER  ({{fetch-header}})                |
+|-------------|-------------------------------------------------|
+| 0x00-0x0F / 0x20-0x2D | OBJECT_STREAM  ({{object-datagram}})    |
```
```suggestion
| 0x00-0x0F / 0x20-0x2D / 0x24..0x25 / 0x28..0x29 / 0x2C..0x2D, | OBJECT_STREAM  ({{object-datagram}})    |
```

afrind:
>  FETCH_HEADER's stream type identifier was moved from 0x05 to 0x50 to avoid conflicts with the Datagram format type IDs 0x00..0x0F and 0x20..0x2F.

This doesn't make any sense.  Datagrams and FETCH_HEADER never appear in the same context.

I question the high level utility here.  There's already a way to omit the subgroup id in single-object subgroups -- set the mode bits to '01' and the sg-id == object id.

> The SUBGROUP_ID_MODE field (bits 1-2, mask 0x06) is a two-bit field that determines the encoding of the Subgroup ID. To extract this value, perform a bitwise AND with mask 0x06 and right-shift by 1 bit:
>
> 0b00: The Subgroup ID field is absent and the Subgroup ID is 0.
>
> 0b01: The Subgroup ID field is absent and the Subgroup ID is the Object ID of the first Object transmitted in this Subgroup.
>
> 0b10: The Subgroup ID field is present in the header.
> 
> 0b11: Reserved for future use.

Removing the length -- this is the draft-02 OBJECT message.  Is it really worth it?

afrind:
If you really hate the length, you could defined 0b11 to mean sg-id == obj-id and there is no explicit length.

draft-ietf-moq-transport.md
```diff
@@ -1007,7 +1007,9 @@ the type of the stream.
 |-------------|-------------------------------------------------|
 | ID          | Type                                            |
 |------------:|:------------------------------------------------|
-| 0x05        | FETCH_HEADER  ({{fetch-header}})                |
+| 0x50        | FETCH_HEADER  ({{fetch-header}})                |
+|-------------|-------------------------------------------------|
+| 0x00-0x0F / 0x20-0x2D / 0x24..0x25 / 0x28..0x29 / 0x2C..0x2D, | OBJECT_STREAM  ({{object-datagram}})    |
```
suhasHere, line 1051: it would be nice to flip this around

PR #1595, opened 2026-04-03
Allow 7-byte varint and non-minimal encodings

### LINKED ISSUE #1517: varint encoding using min size possible

### LINKED ISSUE #1543: Invalid varints (implementation experience tracking)
I've started implementing some of the draft-17 varint syntax, and wanted to note some of the experience / thoughts I've had while doing this.

The good is that there is no longer an invalid value case when serializing uint64 as a varint.  This simplifies serialization logic.

The annoying but not really new part is that now it is much easier to hit integer overflows in all of the delta-encoded fields.

The main downside occurs on the parsing side.  The output of a varint parsing is now a ternary state:
1. Varint parsing is successful.
2. Varint parsing is unsuccessful because there is not enough input data.
3. Varint parsing is unsuccessful because the varint is invalid.

In a lot of cases state 2 and state 3 lead to the same result (e.g. when parsing the contents of a control message), meaning those can be handled the same way.  In other cases, failing to distinguish between the states 2 and 3 leads to a deadlock (i.e. if you naively replace the old varint parser with the new one, and then someone writes an invalid varint on the data stream, it is permanently stuck).  @martinthomson has [pointed out](https://github.com/moq-wg/moq-transport/pull/1016#discussion_r2743708235) this on the original PR.

Given all of this, I'd like to hear from other implementers about how much complexity they've encountered integrating this change.

afrind:
Per discussion at IETF 125 and on-list

Fixes: #1517
Fixes: #1543

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -263,7 +263,7 @@ first byte to indicate the length of the encoding in bytes. The remaining bits
 after the first 0 and subsequent bytes, if any, represent the integer value,
 encoded in network byte order.
 
-Integers are encoded in 1, 2, 3, 4, 5, 6, 8, or 9 bytes and can encode up to 64
+Integers are encoded in 1 to 9 bytes and can encode up to 64
 bit unsigned integers. The following table summarizes the encoding properties.
 
 |--------------|----------------|-------------|------------------------|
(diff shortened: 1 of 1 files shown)


ianswett (approved):

sandarsh (approved):
LGTM

suhasHere (approved):

PR #1604, opened 2026-04-10
Joining FETCH with subscription

### LINKED ISSUE #1602: Joining Fetch should be on the SUBSCRIBE/PUBLISH stream
This solves a bunch of problems:

- no need for an annoying buffer in case the FETCH arrives before SUBSCRIBE
- no need for matching request IDs, looking up the subscribe context, etc
- solves the race condition #1601, which I don't think the current solution completely does
- neatly handles the case where the Unsubscribe happens before SUBSCRIBE_OK but after Joining FETCH
- eliminates tons of error cases and therefore test cases

If the use case where SUBSCRIBE ends but joining FETCH is still going is important, then we don't want to do this. I suspect this is not a real use case.

I might do a PR once REWIND is out of my hands.

### LINKED ISSUE #1612: What happens to Joining FETCH if fwd changes to 0?
The spec says I can't send Joining FETCH if the subscription has fwd=0. What if I send it and then fwd changes to 0? Does that cancel the joining FETCH?

I don't really care what the answer is, but the spec should say what happens.

martinduke:
Fixes #1602
Fixes #1612

This turned out to be much spicier than expected. The weirdest thing is that the joining fetch shares parameter state with the subscribe. I can't see another way to do it without adding another field to REQUEST_UPDATE. Perhaps a REQUEST_UPDATE for joining Fetch always has a dependency on the original FETCH request, and nothing else, and any other listed dependency is for the SUBSCRIBE?

I think this would be cleaner if we separated out the three FETCH flavors into different message types, but that's a quasi-editorial change that makes this diff bigger than it needs to be.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3067,7 +3067,7 @@ A Joining FETCH shares parameters with the associated subscription. If a
 parameter can only be included in SUBSCRIBE or PUBLISH_OK, it only applies to
 the subscription. If it can only be included in a FETCH, it only applies to
 the FETCH. If it can apply to both, it applies to both. Therefore, a joining
-FETCH inherits the parameters from the associated subsccription. Similarly,
+FETCH inherits the parameters from the associated subscription. Similarly,
 any parameters included in a Joining FETCH therefore function as a
 REQUEST_UPDATE to the subscription if the parameters apply. Finally, any
 subsequent REQUEST_UPDATE applies to both.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1431,6 +1431,9 @@ groups. A publisher that does will begin the next group as soon as practical.
 
 ## Fetch State Management
 
+FETCH messages are sent on either their own Bidirectional stream or the
+bidirectional stream of an associated SUBSCRIBE (see {{joining-fetches}}).
+
 The publisher MUST send exactly one FETCH_OK or REQUEST_ERROR in response to a
 FETCH.
 
@@ -1440,6 +1443,9 @@ receives a FIN or is reset. If the data stream is already open,
 the subscriber wishing to cancel the FETCH MAY send STOP_SENDING for the
 data stream as well as the the bidi request stream. It MUST send STOP_SENDING
 for the bidi request stream.
+
+If on an associated SUBSCRIBE request stream, closing the stream terminates the
+FETCH request.
 
 The Publisher can destroy fetch state as soon as it has received a
 STOP_SENDING. It MUST reset the bidi request stream and unidirectional
@@ -1947,7 +1953,7 @@ new request stream.
 |--------|-----------------------------------------------|----------------|
 | 0xB    | PUBLISH_DONE ({{message-publish-done}})       | Request        |
 |--------|-----------------------------------------------|----------------|
-| 0x16   | FETCH ({{message-fetch}})                     | Request, First |
+| 0x16   | FETCH ({{message-fetch}})                     | Request, sometimes First |
 |--------|-----------------------------------------------|----------------|
 | 0x18   | FETCH_OK ({{message-fetch-ok}})               | Request        |
 |--------|-----------------------------------------------|----------------|
@@ -3015,18 +3021,10 @@ EXCESSIVE_LOAD (0x9):
 
 ## FETCH {#message-fetch}
 
-A subscriber sends FETCH as the first message on a new bidi stream to a
-publisher to request a range of already published objects within a track.
-
-There are three types of Fetch messages.
-
-Code | Fetch Type
-0x1 | Standalone Fetch
-0x2 | Relative Joining Fetch
-0x3 | Absolute Joining Fetch
-
-An endpoint that receives a Fetch Type other than 0x1, 0x2 or 0x3 MUST close
-the session with a `PROTOCOL_VIOLATION`.
+A subscriber can send a FETCH as the first message on a new bidi stream, in
+which case it is a "Standalone FETCH", or in an existing SUBSCRIBE or PUBLISH
+bidi stream, in which case it is a "Joining FETCH". Joining FETCH is further
+subdivided into "Relative Joining FETCH" or "Absolute Joining FETCH".
 
 ### Standalone Fetch
 
@@ -3056,12 +3054,8 @@ Standalone Fetch {
 
 ### Joining Fetches
 
-A Joining Fetch is associated with a Subscribe request by
-specifying the Request ID of a subscription in the `Established` or
-`Pending (subscriber)` state. Because Joining Fetch references an existing
-subscription, if that subscription has not yet been established, the Publisher
-receiving the Joining Fetch buffers the pending Joining Fetch until either
-the Subscription is established or the request times out.
+A Joining Fetch is associated with a Subscribe or Publish request that
+initiated the stream that sent it.
 
 A publisher receiving a Joining Fetch uses properties of the associated
 subscription to determine the Track Namespace, Track Name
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -3056,29 +3054,38 @@ Standalone Fetch {
...
 subscription to determine the Track Namespace, Track Name
 and End Location such that it is contiguous with the associated
 subscription.  The subscriber can set the Start Location to an absolute
 Location or a Location relative to the Largest group.
 
+A Joining FETCH shares parameters with the associated subscription. If a
+parameter can only be included in SUBSCRIBE or PUBLISH_OK, it only applies to
+the subscription. If it can only be included in a FETCH, it only applies to
+the FETCH. If it can apply to both, it applies to both. Therefore, a joining
+FETCH inherits the parameters from the associated subsccription. Similarly,
```
```suggestion
FETCH inherits the parameters from the associated subscription. Similarly,
```

draft-ietf-moq-transport.md
```diff
@@ -3056,29 +3054,38 @@ Standalone Fetch {
...
+any parameters included in a Joining FETCH therefore function as a
+REQUEST_UPDATE to the subscription if the parameters apply. Finally, any
+subsequent REQUEST_UPDATE applies to both.
+
+If a publisher responds to joining FETCH with REQUEST_ERROR, it does not
```
afrind, line 3075: How do you know if the REQUEST_ERROR is for Subscribe or Fetch?

afrind (commented):
I agree it's awkward.  You can no longer have a different subscriber priority for your fetch and subscription.

GROUP_ORDER needs to apply to both, though perhaps that's the only sensible choice.

AUTH_TOKEN also needs to be the same.

draft-ietf-moq-transport.md
martinduke: This happens with REQUEST_ERROR too, so I guess it has to be in order?

ianswett (commented):
I'll have to think more about the pros and cons of this, but Joining Fetch is still awkward.

draft-ietf-moq-transport.md
gwendalsimon: yes, the text could be revised.

On a SUBSCRIBE bidi: the subscriber sends SUBSCRIBE, then FETCH, in that order on the same stream. The publisher responds in order: SUBSCRIBE_OK (or REQUEST_ERROR for SUBSCRIBE), then FETCH_OK (or REQUEST_ERROR for FETCH). A REQUEST_ERROR that arrives before SUBSCRIBE_OK applies to the SUBSCRIBE and terminates the stream. A REQUEST_ERROR that arrives after SUBSCRIBE_OK applies to the Joining FETCH and does not terminate the stream.

On a PUBLISH bidi: the relay sends PUBLISH; the subscriber replies PUBLISH_OK, then FETCH. The relay's FETCH_OK or REQUEST_ERROR follows and unambiguously applies to the FETCH.

The revised text could be:
"On a bidi stream that carries both a SUBSCRIBE (or PUBLISH) and a Joining FETCH, the publisher MUST send the response to the SUBSCRIBE/PUBLISH before sending the response to the Joining FETCH. A REQUEST_ERROR sent before SUBSCRIBE_OK terminates the stream and applies to the SUBSCRIBE. A REQUEST_ERROR sent after SUBSCRIBE_OK applies only to the Joining FETCH and MUST NOT terminate the stream."

draft-ietf-moq-transport.md
```diff
@@ -3056,29 +3054,38 @@ Standalone Fetch {
...
+terminate the stream, and any parameter changes to the subscription are not
+applied.
+
+After sending FETCH_OK, if the joining FETCH is terminated for any reason not
+affecting the associated SUBSCRIBE, the endpoint MUST terminate the
+unidirectional Fetch stream but not the request stream.
```
gwendalsimon, line 3081: It covers the case where Joining FETCH terminates without affecting the associated SUBSCRIBE/PUBLISH. The reverse case is not addressed: what happens when the publisher closes the request stream (PUBLISH_DONE+ FIN on the PUBLISH bidi) while the Joining FETCH unidirectional data stream is still delivering objects?

The current wording "closing the stream terminates the FETCH request" means that if the relay sends PUBLISH_DONE + FIN before the FETCH data stream has finished, the subscriber loses any in-flight past-content objects.

We could say "A publisher MUST NOT send the final FIN on the request stream until the associated Joining FETCH unidirectional data stream has been closed (FIN or RESET_STREAM)."

This ensures the subscriber always receives the complete FETCH range before the subscription terminates.

gwendalsimon:
> I agree it's awkward. You can no longer have a different subscriber priority for your fetch and subscription.

I guess the most common use case for Publisher/Relay is to assign higher transmission priority to the FETCH data stream so the subscriber catches up to the live edge quickly, and lower priority to the live subgroup streams during that catch-up phase — then normalize once the FETCH data stream closes. A relay can implement this unilaterally by controlling the QUIC stream priorities of the streams it opens, without the subscriber specifying any parameter. So for this particular use case, the shared-parameter model does not block the right behavior. 

However, the problem occurs when a subscriber includes SUBSCRIBER_PRIORITY in a Joining FETCH and expects it to apply only to past-content delivery. This issues a REQUEST_UPDATE on the live subscription, changing live delivery priority. There is no way for the subscriber to include a parameter in Joining FETCH and have it apply only to FETCH. We could make that UPDATES happening during a dual FETCH/SUBSCRIBE must be ignored?

gwendalsimon:
I want to share a concrete use case related to this PR: the SWITCH proposal (PR #1378).

In the SWITCH design, after the subscriber sends a SWITCH control message identifying the current and target Tracks, the Relay opens a PUBLISH bidi for the target Track. The PUBLISH includes a SWITCH_TRANSITION parameter carrying G_switch (the transition GroupID) and the Live Edge Group ID. The range [G_switch, Live Edge Group ID) is the "catch-up range": the past objects the subscriber needs to receive before it is aligned with the live edge of the target Track. 

If this range is non-empty, I propose these past objects to be delivered on the same PUBLISH bidi (alongside live objects arriving on PUBLISH subgroup streams) with higher QUIC transmission priority than the live streams (see https://github.com/moq-wg/moq-transport/pull/1604#issuecomment-4254198238)

In practice this is achieved by having the Relay immediately follow the PUBLISH message on the relay-to-subscriber direction of the PUBLISH bidi with a FETCH_HEADER and deliver the catch-up objects inline, without any FETCH request from the subscriber.

In other words, #1378 is a relay-proactive variant of what this PR #1604 proposes: co-locating FETCH data on the PUBLISH bidi, but initiated by the Relay rather than by the Subscriber.

draft-ietf-moq-transport.md
afrind: @martinduke thoughts on Gwendal's proposal?

martinduke: this is a general problem with REQUEST_UPDATE, so it's orthogonal to this PR.

If there are REQUEST_UPDATES before the FETCH those have to be processed too.

draft-ietf-moq-transport.md
martinduke: I added text that killing SUBSCRIBE also kills the FETCH. I'm not sure how else to do it; there's no other way to turn off the SUBSCRIBE.

martinduke:
Now fixes #1612 as well

ianswett:
> I want to share a concrete use case related to this PR: the SWITCH proposal (PR #1378).
> 
...
> 
> In other words, #1378 is a relay-proactive variant of what this PR #1604 proposes: co-locating FETCH data on the PUBLISH bidi, but initiated by the Relay rather than by the Subscriber.

Thanks for the detailed reply, as well as your reply on the list, that was helpful for me.

I'd like to put the question of whether Objects are delivered on the control stream in your SWITCH proposal and this one as a separate question.  If we want to do that, we should do it for FETCH as well.

This conversation makes me think using a single message would be better.

ianswett:
> 
> However, the problem occurs when a subscriber includes SUBSCRIBER_PRIORITY in a Joining FETCH and expects it to apply only to past-content delivery. This issues a REQUEST_UPDATE on the live subscription, changing live delivery priority. There is no way for the subscriber to include a parameter in Joining FETCH and have it apply only to FETCH. We could make that UPDATES happening during a dual FETCH/SUBSCRIBE must be ignored?

I thought about this, and I don't think there's a compelling use case for updating the priority separately.  You're issuing a message to Join a Track.  Your interest in joining the Track may increase or decrease since the original request, but I can't imagine a case when you're much more interested in the FETCH data, but not more interested in the SUBSCRIBE data.

afrind:
> but I can't imagine a case when you're much more interested in the FETCH data, but not more interested in the SUBSCRIBE data.

Luke indicated in another issue a case where you want future SUBSCRIBE groups > FETCH current group > SUBSCRIBE current group.  But that also can't be expressed in any form in MOQ today.

zafergurel (approved):

zafergurel:
I accidentally approved the PR, sorry. And I could not take it back. :/

afrind:
Closed in favor of #1642

PR #1605, opened 2026-04-14
Split DELIVERY_TIMEOUT into two types of timeout

### LINKED ISSUE #667: DELIVERY_TIMEOUT is unimplementable
Sec 7.1.1.2:

> If an object in a subgroup exceeds the delivery timeout, the publisher MUST reset the underlying transport stream

To properly implement this, there should (conceptually) be a timer on every object in the stream. Unfortunately, since the Webtransport API does not inform the application when data is acknowledged, and there are no object ACKs, there is not way for MoQT to identify if a given object was delivered within the timeout.

Furthermore, even with full knowledge of ACKs, a lost ACK (at QUIC or MoQT layer) could cause a spurious reset. Putting some responsibility on the receiver to send STOP_SENDING could solve many of these problems, though the spec solely makes demands of the publisher, and the receiver does not know the original arrival time.

When implementing, I approximated it as follows:

1) Record the time of every object arrival in the queue.

2) When trying to send an object to the QUIC layer, reset the stream if delivery_timeout is already expired.

3) Set a timer when FIN is sent; reset the stream if still open when the timer fires.

This approach seems tractable, but does not cover the case where objects not sent with FIN are sent on time, but retransmissions cause delivery to exceed the timeout.

I don't have a specific proposal to solve this, although @vasilvv does.

vasilvv:
OBJECT_DELIVERY_TIMEOUT is intended to be semantically equivalent to the existing DELIVERY_TIMEOUT while being more precisely defined, while the new SUBGROUP_DELIVERY_TIMEOUT provides additional coverage for subgroups that have been fully queued but not yet fully delivered.

Fixes #667

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1662,6 +1662,70 @@ Implementations that have a default priority SHOULD set it to a value in
 the middle of the range (eg: 128) to allow non-default priorities to be
 set either higher or lower.
 
+# Delivery Timeouts and Data Reliability {#timeouts}
+
+Each MOQT subscription has two timeout values associated with it: a
+SUBGROUP_DELIVERY_TIMEOUT and an OBJECT_DELIVERY_TIMEOUT.  Both of those values
+are expressed in milliseconds; both are optional; a value of 0 means that
+there is no timeout set.
+
+The publisher communicates both timeout values as a Track Property; the
+subscriber communicates them as a Message Parameter.  For each type of timeout,
+if both the publisher and the subscriber have a non-zero value, the smaller of
+the two is used.
+
+If the OBJECT_DELIVERY_TIMEOUT is not zero, the MOQT implementation MUST retain
+the time at which the first byte of every object has been either received from
+the upstream subscription, or provided by the original publisher application.
+The actual mechanism by which the timeout works depends on the Object
+Forwarding Preference:
+
+- For subgroups, the implementation MUST check the time elapsed since the first
+  byte of the object before attempting to send it; if the time elapsed exceeds
+  OBJECT_DELIVERY_TIMEOUT, it MUST reset the underlying transport stream with
+  the reset stream code DELIVERY_TIMEOUT (see {{closing-subgroup-streams}}) and
+  SHOULD NOT attempt to open a new stream to deliver additional Objects in that
+  Subgroup.  In order for this timeout mechanism to be efficient, the
+  implementations SHOULD limit the amount of data buffered at the underlying
+  transport layer.
+- For datagrams, the implementation MUST drop the datagrams if the time elapsed
+  since the first byte exceeds OBJECT_DELIVERY_TIMEOUT.  Similar to subgroups,
+  implementations SHOULD either minimize datagram queueing, or use datagram
+  queueing mechanisms that support time bounds (such as the `outgoingMaxAge`
+  parameter in the W3C WebTransport API).
+
+If the Object Forwarding Preference is Subgroup and the value of
+SUBGROUP_DELIVERY_TIMEOUT is not zero, once the MOQT implementation becomes
+aware that all of the objects on the subgroup have been published (either by
+receiving a FIN from the upstream subscription, or, in case of the original
+publisher, through being notified of this fact by the application), it MUST
+start a timer that will reset the underlying subgroup stream after
+SUBGROUP_DELIVERY_TIMEOUT elapses.  The subgroup timeout timer MUST remain
+active until the underlying transport stream reaches "all data committed" state
+({{!I-D.ietf-webtrans-overview, Section 4.3}}); this ensures that MOQT can time
+out subgroups where all of the data has been sent, but the stream has not been
+fully closed due to the packet loss.
+
+For objects with Object Forwarding Preference set to Datagram, the
+SUBGROUP_DELIVERY_TIMEOUT acts the same way as OBJECT_DELIVERY_TIMEOUT; if both
+are non-zero, the smaller of the two is used.
+
+|          | `SUBGROUP_DELIVERY_TIMEOUT` | `OBJECT_DELIVERY_TIMEOUT` |
+|:---------|:----------------------------|:--------------------------|
+| Timeout starts | When the FIN for the subgroup is received | When the first byte of the object is received |
+| Timeout checked at | Via a timer until all data is acknowledged | When the object is sent to the underlying transport |
+| Action upon timeout | Reset for subgroups, drop for datagrams | Reset for subgroups, drop for datagrams |
+{: #timeout-comparison title="Comparison of the delivery timeout mechanisms" }
+
+Publishers can, at their discretion, discontinue forwarding Objects before
+either of the timeouts occurs, subject to stream closure and ordering
+constraints described in {{closing-subgroup-streams}}.  However, if none of the
+timeouts are set to a non-zero value, all Objects in the track matching the
+subscription filter are delivered as indicated by their Group Order and
+Priority.  If a subscriber fails to consume Objects at a sufficient rate,
+causing the publisher to exceed its resource limits, the publisher MAY
+terminate the subscription using PUBLISH_DONE with error `TOO_FAR_BEHIND`.
+
 # Relays {#relays-moq}
 
 Relays are leveraged to enable distribution scale in the MOQT
@@ -1752,7 +1816,7 @@ to identify its Track and find the current subscribers.  Each new Object
 belonging to the Track is forwarded to each subscriber, as allowed by the
 subscription's filter (see {{message-subscribe-req}}), and delivered according
 to the priority (see {{priorities}}) and delivery timeout (see
-{{delivery-timeout}}).
+{{timeouts}}).
 
 A relay MUST NOT reorder or drop objects received on a multi-object stream when
 forwarding to subscribers, unless it has application specific information.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1662,6 +1662,70 @@ Implementations that have a default priority SHOULD set it to a value in
...
+the two is used.
+
+If the OBJECT_DELIVERY_TIMEOUT is not zero, the MOQT implementation MUST retain
+the time at which the first byte of every object has been either received from
+the upstream subscription, or provided by the original publisher application.
+The actual mechanism by which the timeout works depends on the Object
+Forwarding Preference:
+
+- For subgroups, the implementation MUST check the time elapsed since the first
+  byte of the object before attempting to send it; if the time elapsed exceeds
```
afrind: Do we need to say something about retransmissions?  If the first TX happens before ODT but loss is detected after, the publisher can, MAY, SHOULD, MUST NOT reset?

draft-ietf-moq-transport.md
```diff
@@ -1662,6 +1662,70 @@ Implementations that have a default priority SHOULD set it to a value in
...
+  OBJECT_DELIVERY_TIMEOUT, it MUST reset the underlying transport stream with
+  the reset stream code DELIVERY_TIMEOUT (see {{closing-subgroup-streams}}) and
+  SHOULD NOT attempt to open a new stream to deliver additional Objects in that
+  Subgroup.  In order for this timeout mechanism to be efficient, the
+  implementations SHOULD limit the amount of data buffered at the underlying
+  transport layer.
```
afrind: Do you mean "effective" rather than "efficient"?  And if so, more elaboration about why limiting the buffering makes the timeout more effective?

draft-ietf-moq-transport.md
```diff
@@ -1662,6 +1662,70 @@ Implementations that have a default priority SHOULD set it to a value in
...
+- For datagrams, the implementation MUST drop the datagrams if the time elapsed
+  since the first byte exceeds OBJECT_DELIVERY_TIMEOUT.  Similar to subgroups,
+  implementations SHOULD either minimize datagram queueing, or use datagram
+  queueing mechanisms that support time bounds (such as the `outgoingMaxAge`
+  parameter in the W3C WebTransport API).
```
afrind, line 1842: I think it makes more sense to say - drop the datagram at the lowest possible layer -- preferably by passing the timeout value down.  The references to queuing muddles things because queues don't have a relationship to timeouts?

draft-ietf-moq-transport.md
vasilvv: Good question!  I assume that most QUIC stacks don't really expose that information, so I am not sure what to say about that.

draft-ietf-moq-transport.md
vasilvv: Well, at least in WebTransport, datagram queues have timeouts associated with them.  If your stack defers datagram queueing to the application, you'd have to deal with that check yourself.

draft-ietf-moq-transport.md
afrind: I think we should explicitly permit cancellation of retransmissions after delivery timeout, and even suggest that is optimal:

> Optimal implementations will also prevent retransmitting object data its delivery timeout has expired.

draft-ietf-moq-transport.md
afrind, line 1842:
```suggestion
  since the first byte exceeds OBJECT_DELIVERY_TIMEOUT.  
  Optimal implementations will evaluate the datagram's delivery timeout
  as late as possible before sending, after any internal queuing.
```

draft-ietf-moq-transport.md
```suggestion
  Subgroup.  Optimal implementations will evaluate the subgroup's delivery timeout
  as late as possible, after any internal queuing.
```

draft-ietf-moq-transport.md
```diff
@@ -1662,6 +1662,70 @@ Implementations that have a default priority SHOULD set it to a value in
...
 
+# Delivery Timeouts and Data Reliability {#timeouts}
+
+Each MOQT subscription has two timeout values associated with it: a
+SUBGROUP_DELIVERY_TIMEOUT and an OBJECT_DELIVERY_TIMEOUT.  Both of those values
+are expressed in milliseconds; both are optional; a value of 0 means that
+there is no timeout set.
+
+The publisher communicates both timeout values as a Track Property; the
+subscriber communicates them as a Message Parameter.  For each type of timeout,
```
```suggestion
subscriber communicates them as Message Parameters.  For each type of timeout,
```

draft-ietf-moq-transport.md
ianswett: MAY is probably pragmatic.  Or SHOULD if you support it?

draft-ietf-moq-transport.md
ianswett: Given Webtransport supports this, it seems like a reasonable example.

draft-ietf-moq-transport.md
```diff
@@ -1662,6 +1662,70 @@ Implementations that have a default priority SHOULD set it to a value in
...
+  implementations SHOULD either minimize datagram queueing, or use datagram
+  queueing mechanisms that support time bounds (such as the `outgoingMaxAge`
+  parameter in the W3C WebTransport API).
+
+If the Object Forwarding Preference is Subgroup and the value of
+SUBGROUP_DELIVERY_TIMEOUT is not zero, once the MOQT implementation becomes
+aware that all of the objects on the subgroup have been published (either by
+receiving a FIN from the upstream subscription, or, in case of the original
+publisher, through being notified of this fact by the application), it MUST
+start a timer that will reset the underlying subgroup stream after
```
ianswett: In implementation, one can typically avoid using a timer, so saying MUST her feels odd.

draft-ietf-moq-transport.md
```diff
@@ -1662,6 +1662,70 @@ Implementations that have a default priority SHOULD set it to a value in
 the middle of the range (eg: 128) to allow non-default priorities to be
 set either higher or lower.
 
+# Delivery Timeouts and Data Reliability {#timeouts}
```
```suggestion
# Delivery Timeouts and Data Reliability {#delivery-timeouts}
```

ianswett (commented):
I think this looks reasonable, but I don't intuitively understand why two timeouts are necessary.

afrind (commented):
oops flushing old comments

draft-ietf-moq-transport.md
vasilvv: I think you actually have to, but sure, I'll remove the timer mention.

draft-ietf-moq-transport.md
vasilvv: "Effective" is probably better.

vasilvv:
Addressed the comments.

draft-ietf-moq-transport.md
```diff
@@ -1662,6 +1662,71 @@ Implementations that have a default priority SHOULD set it to a value in
...
+are expressed in milliseconds; both are optional; a value of 0 means that
+there is no timeout set.
+
+The publisher communicates both timeout values as a Track Property; the
+subscriber communicates them as Message Parameters.  For each type of timeout,
+if both the publisher and the subscriber have a non-zero value, the smaller of
+the two is used.
+
+If the OBJECT_DELIVERY_TIMEOUT is not zero, the MOQT implementation MUST retain
+the time at which the first byte of every object has been either received from
```
afrind: Is this header byte or payload byte?

draft-ietf-moq-transport.md
```diff
@@ -1662,6 +1662,71 @@ Implementations that have a default priority SHOULD set it to a value in
...
+the upstream subscription, or provided by the original publisher application.
+The actual mechanism by which the timeout works depends on the Object
+Forwarding Preference:
+
+- For subgroups, the implementation MUST check the time elapsed since the first
+  byte of the object before attempting to send it; if the time elapsed exceeds
```
afrind: "send it" -> passing it to the underlying transport for transmission?

draft-ietf-moq-transport.md
```diff
@@ -1662,6 +1662,71 @@ Implementations that have a default priority SHOULD set it to a value in
...
+  OBJECT_DELIVERY_TIMEOUT, it MUST reset the underlying transport stream with
+  the reset stream code DELIVERY_TIMEOUT (see {{closing-subgroup-streams}}) and
+  SHOULD NOT attempt to open a new stream to deliver additional Objects in that
+  Subgroup.  This SHOULD apply to retransmissions if the underlying transport
+  implementation allows.  If not, in order for this timeout mechanism to be
```
afrind: The timeout should be re-checked before retransmitting an object if the underlying transport allows

draft-ietf-moq-transport.md
```diff
@@ -1662,6 +1662,71 @@ Implementations that have a default priority SHOULD set it to a value in
...
+  effective, the implementations SHOULD limit the amount of data buffered at
+  the underlying transport.
```
afrind: I'm still tripping on this buffering comment.  Maybe adding some explanation like:

Any delay between the timeout check and transmission of the object can result in delivering expired data.  If the underlying transport buffers or queues data before transmission, the implementation SHOULD limit the amount of data buffered there.

That said, there's a tension with priority vs delivery timeout -- if I've buffered only low-pri data, and higher-pri data arrives, should I pass it to the underlying transport immediately, or should I wait until transport space becomes availaible?

draft-ietf-moq-transport.md
afrind: Why don't streams have an "outgoingMaxAge" parameter -- e.g. all this queueing business is a band-aid over having deadline aware transports.  I recognize those aren't common.

It seems to implement either of these mechanisms, transports need to provide some callback when "transport buffers/queues are available".

This is all to say I prefer to keep the advice high level -- check the timeouts as late as you can.  That might mean delaying passing available objects to the underlying transport, which has other tradeoffs.

draft-ietf-moq-transport.md
vasilvv: Let's go with payload, since it's easier to define.

draft-ietf-moq-transport.md
vasilvv: Added the explanation.

> That said, there's a tension with priority vs delivery timeout -- if I've buffered only low-pri data, and higher-pri data arrives, should I pass it to the underlying transport immediately, or should I wait until transport space becomes availaible?

What would waiting give you?  At least in our implementation, the buffer is per-stream, meaning that you don't get anything from waiting.

draft-ietf-moq-transport.md
vasilvv: > Why don't streams have an "outgoingMaxAge" parameter -- e.g. all this queueing business is a band-aid over having deadline aware transports. I recognize those aren't common.

Datagrams are atomic application units, streams bytes are not, thus the difference; for contrast, SCTP is message-based, and [it has that feature too](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/maxPacketLifeTime).  Streams themselves are atomic, but since we give you an object to access a stream, you can reset it yourself. 

> This is all to say I prefer to keep the advice high level -- check the timeouts as late as you can. That might mean delaying passing available objects to the underlying transport, which has other tradeoffs.

I can see that this would not apply to implementation that don't have datagram queues.  The advice given here is a practicality.

draft-ietf-moq-transport.md
afrind: What you get from delaying is a later delivery timeout check.  In any case, it feels like you've written all of this with your particular implementation in mind, and it's not generalizing.

draft-ietf-moq-transport.md
```diff
@@ -1662,6 +1662,71 @@ Implementations that have a default priority SHOULD set it to a value in
...
+publisher, through being notified of this fact by the application), it MUST
+reset the underlying subgroup stream after SUBGROUP_DELIVERY_TIMEOUT elapses.
+The subgroup timeout timer MUST remain active until the underlying transport
+stream reaches "all data committed" state ({{!I-D.ietf-webtrans-overview,
+Section 4.3}}); this ensures that MOQT can time out subgroups where all of the
+data has been sent, but the stream has not been fully closed due to the packet
+loss.
+
+For objects with Object Forwarding Preference set to Datagram, the
+SUBGROUP_DELIVERY_TIMEOUT acts the same way as OBJECT_DELIVERY_TIMEOUT; if both
```
ianswett, line 1858: This is weird because it SUBGROUP_DELIVERY_TIMEOUT would be the end of the Subgroup has been received, but OBJECT_DELIVERY_TIMEOUT is the beginning of the Object.  My intuition is that it would be from when the end of the Object in the Datagram is received.

ianswett (approved):
Overall looks good, though I'm confused by one part.

draft-ietf-moq-transport.md
afrind: For datagram, doesn't the beginning and the end necessarily have to be the same time?

draft-ietf-moq-transport.md
```diff
@@ -2336,49 +2404,25 @@ REGISTER without waiting for a response.
...
-Subgroup.
-
-This parameter is intended to be specific to a
-subscription, so it SHOULD NOT be forwarded upstream by a relay that intends
-to serve multiple subscriptions for the same track.
-
-Publishers SHOULD consider whether the entire Object can likely be
-successfully delivered within the timeout period before sending any data
-for that Object, taking into account priorities, congestion control, and
-any other relevant information.
```
afrind, line 2381: Was the removal of this intentional?  I think we still want it

afrind (approved):
Probably needs another editorial pass

PR #1607, opened 2026-04-18
[Draft/RFC] Add a `Largest Available Group` filter.

vasilvv:
This is similar to other proposals such as [Rewind](https://datatracker.ietf.org/doc/draft-duke-moq-subscribe-rewind/), [CurrentGroupFill](https://github.com/afrind/moq-transport/pull/15), and [LargestGroup](https://mailarchive.ietf.org/arch/msg/moq/y5f5XZT005Y6ebrHYBOXtr4JojQ/). Distinct features of this approach:
* Only the current group is supported.
* It will always serve a complete group.
* It does not require any explicit backfill on the relay side.
* Probably really easy to implement.
* It has sensible semantics for cases when the groups are really large.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1424,6 +1424,27 @@ delivered will be the Group ID in `Start Location` plus the `End Group Delta`.
 If the resulting Group ID would be greater than 2^64 - 1, the endpoint MUST
 close the session with a `PROTOCOL_VIOLATION`.
 
+Largest Available Group (0x5): The filter Largest Avaiable Group provides a way
+to join the track on the newest group that is immediately available.
+Specifically:
+* If the original publisher receives a SUBSCRIBE, and it has all of the objects
+  for the current group available, it SHOULD serve the subscription as if the
+  Largest Object is `(Largest Object.Group ID, 0)`.
+* If the original publisher does not have all of the objects for the current
+  group available to serve, it MUST treat this filter as Next Group Start.
+* If a relay has an existing upstream subscription, it MUST follow the same
+  rules as the original publisher; since a relay does not have the knowledge
+  of the object ID structure that the original publisher has, it is defined to
+  "have all objects for the current group available" if it could serve a FETCH
+  for the range `[(Largest Object.Group ID, 0); (Largest Object.Group ID,
+  Largest Object.Object ID)]`.
+* If a relay does not have an existing upstream subscription, it MUST use the
+  Largest Avaiable Group filter when subscribing to the upstream publisher.
+
+TODO: we may need to tell the subscriber which of the scenarios above have
+occurred; consider porting `START_LOCATION` from
+<https://github.com/afrind/moq-transport/pull/15/changes>.
+
 An endpoint that receives a filter type other than the above MUST close the
 session with `PROTOCOL_VIOLATION`.
 


draft-ietf-moq-transport.md
```diff
@@ -1424,6 +1424,27 @@ delivered will be the Group ID in `Start Location` plus the `End Group Delta`.
...
 close the session with a `PROTOCOL_VIOLATION`.
 
+Largest Available Group (0x5): The filter Largest Avaiable Group provides a way
+to join the track on the newest group that is immediately available.
+Specifically:
+* If the original publisher receives a SUBSCRIBE, and it has all of the objects
+  for the current group available, it SHOULD serve the subscription as if the
+  Largest Object is `(Largest Object.Group ID, 0)`.
+* If the original publisher does not have all of the objects for the current
+  group available to serve, it MUST treat this filter as Next Group Start.
```
kixelated, line 1434: The MUST is a too strong and requiring a full cache is too narrow. IMO:

A relay MAY attempt to reconstruct subscription from a partial cache. An object MUST NOT be served until all prior objects within that sub-group have been served. This can be deduced by sequence numbers, upstream subscriptions (sub-groups are ordered), and/or upstream fetches (groups are ordered).

In the event of an incomplete sub-group, a relay MAY issue upstream FETCH(s) to populate any gaps. Otherwise it MUST skip the remainder of the sub-group.


@vasilvv note that it would always be valid to serve object 0, even if you have zero information about object 1. Likewise it's valid to serve sub-group 0 from cache even if sub-group 1 might have gaps.

Both of these are better than defaulting to NextGroup, but that's still an option. A relay could "skip" each sub-group if it has no information about the head and doesn't want to cache fill.

vasilvv: > A relay MAY attempt to reconstruct subscription from a partial cache. An object MUST NOT be served until all prior objects within that sub-group have been served. This can be deduced by sequence numbers, upstream subscriptions (sub-groups are ordered), and/or upstream fetches (groups are ordered).

The main problem here is that as far as I remember, MoQ does not currently have a mechanism to specify if an object is the first object of the subgroup, so you can't tell if you have a complete subgroup or not.

draft-ietf-moq-transport.md
```diff
@@ -1424,6 +1424,27 @@ delivered will be the Group ID in `Start Location` plus the `End Group Delta`.
 If the resulting Group ID would be greater than 2^64 - 1, the endpoint MUST
 close the session with a `PROTOCOL_VIOLATION`.
 
+Largest Available Group (0x5): The filter Largest Avaiable Group provides a way
```
sharmafb, line 1427: Typo: available

draft-ietf-moq-transport.md
```diff
@@ -1424,6 +1424,27 @@ delivered will be the Group ID in `Start Location` plus the `End Group Delta`.
...
+* If the original publisher does not have all of the objects for the current
+  group available to serve, it MUST treat this filter as Next Group Start.
+* If a relay has an existing upstream subscription, it MUST follow the same
+  rules as the original publisher; since a relay does not have the knowledge
+  of the object ID structure that the original publisher has, it is defined to
+  "have all objects for the current group available" if it could serve a FETCH
+  for the range `[(Largest Object.Group ID, 0); (Largest Object.Group ID,
+  Largest Object.Object ID)]`.
+* If a relay does not have an existing upstream subscription, it MUST use the
+  Largest Avaiable Group filter when subscribing to the upstream publisher.
```
sharmafb, line 1442: typo: Available

draft-ietf-moq-transport.md
ianswett: The first object in a subgroup starts the subgroup except in cases like "largest Object" today and when a range filter explicitly starts partway through a Group.  If you do either of these things, you don't know when it starts.

We could force the Subgroup ID to be the Object ID of the first Object, and then it'd be unambiguous.

draft-ietf-moq-transport.md
suhasHere: This breaks the deduplication property of subscription aggregation

draft-ietf-moq-transport.md
```diff
@@ -1424,6 +1424,27 @@ delivered will be the Group ID in `Start Location` plus the `End Group Delta`.
 If the resulting Group ID would be greater than 2^64 - 1, the endpoint MUST
 close the session with a `PROTOCOL_VIOLATION`.
 
+Largest Available Group (0x5): The filter Largest Avaiable Group provides a way
+to join the track on the newest group that is immediately available.
+Specifically:
+* If the original publisher receives a SUBSCRIBE, and it has all of the objects
+  for the current group available, it SHOULD serve the subscription as if the
+  Largest Object is `(Largest Object.Group ID, 0)`.
```
suhasHere, line 1432: I am not a big fan of incorrectly setting the largest group. This is a hack at the best IIRC. 
Say Sub1 asks for this filter and gets largest group/object 0,0
and Sub2 asks around the same time for latest group/object and gets 0, 35 , the system cannot rely on such information. 
This has implications on latency expectations, buffer depth calculations and may be some more.

suhasHere (changes requested):
I have few concerns of representing the state ( and implicitly assigning the start point). May be we need to be explicit about live edge location.

At a higher level, I am tying to understand what use-cases this would solve in comparison to NextGroup filter. May be I am missing something, love to learn more.

There are 2 possible type of groups today - Static GOPs and Dynamic GOPs for media. For static GOPs, if the duration is short, say 2s , then going to the beginning of the current group might be same effect as starting always at next group. For larger GOPs or dynamic GOPs, next group may be too far to have applications wait. I do feel NGR might be a good solution that end subscriber may make such a decision.

Having said that my original filter proposal couple of years ago did support "beginning of the current group" as well.  I would like to better understand the use-case .

kixelated:
> At a higher level, I am tying to understand what use-cases this would solve in comparison to NextGroup filter. May be I am missing something, love to learn more.

One concrete use-case is the catalog track. `NextGroup` will never resolve; you MUST do a JOINING FETCH or SUBSCRIBE LargestGroup. A NGR for catalogs is extremely wasteful; every existing subscriber will get a duplicate copy of the old catalog.

> For static GOPs, if the duration is short, say 2s , then going to the beginning of the current group might be same effect as starting always at next group. 

Suhas, at Twitch we heavily optimized to reduce TTV. It's not acceptable to wait a random amount of time from 0-2s waiting for the next group.

Let's take the median and say we're 1s into a 2s group. We want to use LargestGroup to race:

1. Downloading/decoding 1s+ of old media (at network speed) to catch up from behind.
2. Or wait 1s.

As a concrete example, let's say the media is 1.5Mb/s (360p) and our network can do 3Mb/s. It'll take something like 0.66s to start playback instead of 1s if we waited, plus the congestion controller will be warmed. 333ms faster startup time is HUGE, and this is just for this median case.

> For larger GOPs or dynamic GOPs, next group may be too far to have applications wait. I do feel NGR might be a good solution that end subscriber may make such a decision.

You have to remember that NGR will be ignored if the group is too new, and NextGroup can sit there doing nothing.

Ideally, you combine CurrentGroup + NGR. Again it's a race:

1. Download/decode x seconds of old media at network speed.
2. Or wait for a new group.

This will warm the congestion controller much like LargestObject, but you also have the ability to WIN THE RACE and have faster startup time. I think it's a strict improvement over the status quo.

draft-ietf-moq-transport.md
kixelated: Oh and I just re-read this. The "original publisher" is an important distinction; I thought this line applied to all publishers (ex. relay).

I think this is fine, although `MUST` is still very strong. I would open the door to serve whatever is available from cache instead of all-or-nothing.

suhasHere:
> One concrete use-case is the catalog track. `NextGroup` will never resolve; you MUST do a JOINING FETCH or SUBSCRIBE LargestGroup. A NGR for catalogs is extremely wasteful; every existing subscriber will get a duplicate copy of the old catalog.
> 
> >
NGR is not used for catalog typically. Also if new group generates the same catalog , it is application problem. Regardless, fetching existing catalog seems a fine solution.

vasilvv:
Closing in favor of #1642

PR #1608, opened 2026-04-23
Make Subgroup ID identical to first Object Id in the Subgroup

### LINKED ISSUE #1405: Single Object Subgroups don't need a Subgroup ID

ianswett:
Update the text on Subgroup header/object framing to mandate that the Subgroup ID must be the Object ID of the first Object in the Subgroup. Changed the Object ID Delta text so that the Object ID is the Subgroup ID + Object ID Delta if it's the first Object in the Subgroup stream.

* Allows a receiver of Objects in a Subgroup to know what the first Object ID in the Subgroup is.
* Makes prioritization between Subgroups and Datagrams well-defined.
* Saves ~1 byte of wire overhead per Subgroup

Fixes #1405 
Closes #1593

---
*PR created automatically by Jules for task [5092477403524565180](https://jules.google.com/task/5092477403524565180) started by @ianswett*

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -542,11 +542,11 @@ same Subgroup MUST NOT be sent on different streams, unless one of the streams
 was reset prematurely, or upstream conditions have forced objects from a Subgroup
 to be sent out of Object ID order.
 
-Original publishers assign each Subgroup a Subgroup ID, and do so as they see fit.  The
-scope of a Subgroup ID is a Group, so Subgroups from different Groups MAY share a Subgroup
-ID without implying any relationship between them. In general, publishers assign
-objects to subgroups in order to leverage the features of streams as described
-above.
+Original publishers MUST assign each Subgroup a Subgroup ID equal to the Object ID
+of the first Object in the Subgroup.  The scope of a Subgroup ID is a Group, so
+Subgroups from different Groups MAY share a Subgroup ID without implying any
+relationship between them. In general, publishers assign objects to subgroups in
+order to leverage the features of streams as described above.
 
 In general, if Object B is dependent on Object A, then delivery of B can follow
 A, i.e. A and B can be usefully delivered over a single stream.  If an Object is
@@ -3822,9 +3822,9 @@ following fields.
 The Object Status field is only sent if the Object Payload Length is zero.
 
 The Object ID Delta + 1 is added to the previous Object ID in the Subgroup
-stream if there was one.  The Object ID is the Object ID Delta if it's the first
-Object in the Subgroup stream. If the resulting Object ID would be greater
-than 2^64 - 1, the endpoint MUST close the session with a
+stream if there was one.  The Object ID is the Subgroup ID + Object ID Delta
+if it's the first Object in the Subgroup stream. If the resulting Object ID
+would be greater than 2^64 - 1, the endpoint MUST close the session with a
 `PROTOCOL_VIOLATION`. For example, a Subgroup of sequential Object IDs
 starting at 0 will have 0 for all Object ID Delta values. A consumer cannot
 infer information about the existence of Objects between the current and
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1654,8 +1654,7 @@ the objects SHOULD be selected as follows:
    and publisher priority and belong to the same group of the same track, the
    one with **the lowest Subgroup ID** (for objects with forwarding preference
    Subgroup), or **the lowest Object ID** (for objects with forwarding preference
-   Datagram) is scheduled to be sent first.  If the two objects have
-   different Forwarding Preferences the order is implementation dependent.
+   Datagram) is scheduled to be sent first.
 
 The definition of "scheduled to be sent first" in the algorithm is implementation
 dependent and is constrained by the prioritization interface of the underlying


draft-ietf-moq-transport.md
```diff
@@ -1654,8 +1654,7 @@ the objects SHOULD be selected as follows:
    and publisher priority and belong to the same group of the same track, the
    one with **the lowest Subgroup ID** (for objects with forwarding preference
    Subgroup), or **the lowest Object ID** (for objects with forwarding preference
-   Datagram) is scheduled to be sent first.  If the two objects have
```
afrind, line 1693: This is still relevant if you have a group with SG=0 and datagrams.  Unless you are saying that the datagram's object ID is compared to the subgroup ID, in which case you should be more explicit.

afrind (commented):
Do you need to add content to malformed tracks?

More text needed that says when you would have a object delta > 0 for the first object (eg Largest Object)

draft-ietf-moq-transport.md
ianswett: That's what I mean, so I guess I should be more explicit.

draft-ietf-moq-transport.md
```diff
@@ -542,11 +542,11 @@ same Subgroup MUST NOT be sent on different streams, unless one of the streams
 was reset prematurely, or upstream conditions have forced objects from a Subgroup
 to be sent out of Object ID order.
 
-Original publishers assign each Subgroup a Subgroup ID, and do so as they see fit.  The
-scope of a Subgroup ID is a Group, so Subgroups from different Groups MAY share a Subgroup
-ID without implying any relationship between them. In general, publishers assign
-objects to subgroups in order to leverage the features of streams as described
-above.
+Original publishers MUST assign each Subgroup a Subgroup ID equal to the Object ID
```
suhasHere, line 545:
```suggestion
Original publishers SHOULD assign each Subgroup a Subgroup ID equal to the Object ID
```

draft-ietf-moq-transport.md
ianswett: Actually, re-reading the text, isn't that what it says?

ianswett:
Feedback from the WG at today's interim was:
1) People agreed it was important to know what the start Object ID of the Subgroup (and possibly Group?)
2) People had different concerns about restricting the Subgroup ID to be starting Object ID at the Original Publisher.
3) There was some confusion about both this proposal and what is possible in today's Object model in terms of publishing Objects in a subgroup 'out of order'

yuyou:
Thanks for the PR. My thoughts are as follows:

I have some concerns that mandating the Subgroup ID to always be identical to the first Object ID might restrict publisher flexibility.

**The Problem**: Tying the Subgroup ID directly to the first Object ID prevents applications from maintaining structurally consistent Subgroup IDs across multiple Groups. Publishers often need the flexibility to keep the exact same Subgroup ID for a specific data stream component over the lifespan of a track, which this new rule would break when a new Group starts.

**Proposed Solution:** To solve the original problem of identifying the first Object ID
, may it be an alternative approach to explicitly signal the starting Object ID in the Subgroup header instead of tying it to the Subgroup ID field. By keeping the Subgroup ID logically decoupled from the Object ID, publishers can maintain consistent data structures.

afrind:
Discussed 4/27:

The working group didn't think this was the right approach, but agreed we need a way to know if a subgroup contains the beginning.

ianswett:
I looked at https://github.com/moq-wg/moq-transport/pull/1618 and I think this approach has some benefits:
1) It's required that one use it.
2) You can tell when you Don't have the first Object of a Subgroup.

These both seem like desirable properties, particularly the second one.

ianswett:
> Thanks for the PR. My thoughts are as follows:
> 
> I have some concerns that mandating the Subgroup ID to always be identical to the first Object ID might restrict publisher flexibility.
> 
> **The Problem**: Tying the Subgroup ID directly to the first Object ID prevents applications from maintaining structurally consistent Subgroup IDs across multiple Groups. Publishers often need the flexibility to keep the exact same Subgroup ID for a specific data stream component over the lifespan of a track, which this new rule would break when a new Group starts.

I agree this is a constraint, but I haven't come up with any cases it's a problem.  Do you have a specific use case for Subgroups where it is problematic?

For the "One subgroup per layer" example, using the First Object ID seems like a reasonable choice.
For the "One base subgroup plus put every b-frame (or equiv) in its own subgroup", this restriction also works out as intended.

> 
> **Proposed Solution:** To solve the original problem of identifying the first Object ID , may it be an alternative approach to explicitly signal the starting Object ID in the Subgroup header instead of tying it to the Subgroup ID field. By keeping the Subgroup ID logically decoupled from the Object ID, publishers can maintain consistent data structures.

See PR #1618

afrind:
Closing in favor of 1618

PR #1609, opened 2026-04-23
Joining Fetch forward state mismatch is a request error

### LINKED ISSUE #1601: Joining FETCH session errors are subject to a race condition
The draft says Joining FETCH to a forward = 0 subscription is a session error.

But the forward state and the FETCH are on different streams, so there's a race condition. If the REQUEST_UPDATE that carries forward = 1 is delayed, that can kill the whole session.

This cannot be a session error!

afrind:
The forward state and the FETCH are on different streams, so there is a race condition if a REQUEST_UPDATE carrying forward=1 is delayed. This cannot be a session error.

Fixes: #1601

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3136,8 +3136,9 @@ A Subscriber can use a Joining Fetch to, for example, fill a playback buffer
 with a certain number of groups prior to the live edge of a track.
 
 A Joining Fetch is only permitted when the associated subscription has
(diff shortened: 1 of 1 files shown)


martinduke:
LGTM if it's the way we want to fix it, instead of something bigger like #1604

ianswett (approved):

sharmafb (approved):

sandarsh (approved):

PR #1610, opened 2026-04-23
Define textual aliases for REQUEST_OK by request type

afrind:
Introduce shorthand names REQUEST_UPDATE_OK, TRACK_STATUS_OK, SUBSCRIBE_NAMESPACE_OK, and PUBLISH_NAMESPACE_OK to refer to a REQUEST_OK sent in response to the corresponding request type. This improves readability by eliminating verbose "REQUEST_OK (in response to X)" constructions throughout the document.

PR #1611, opened 2026-04-23
Remove PUBLISH_OK message type, make it a REQUEST_OK alias

### LINKED ISSUE #1598: Why PUBLISH_OK not REQUEST_OK?
Why do we have PUBLISH_OK? It conveys nothing beyond REQUEST_OK. Can we remove it?

afrind:
NOTE TO EDITORS -- retarget main branch after #1610 lands

PUBLISH_OK has the same wire format as REQUEST_OK (just Parameters, no Track Properties). Replace it with REQUEST_OK and define PUBLISH_OK as a textual shorthand, consistent with the other REQUEST_OK aliases.

This is a wire format change because the code point and also there is an empty Properties section.

Fixes: #1598


ianswett (approved):
Reminder to retarget this.

sharmafb (approved):

sandarsh (approved):

PR #1613, opened 2026-04-23
Add MAX_REQUEST_UPDATES setup option and TOO_MANY_REQUEST_UPDATES error

### LINKED ISSUE #1063: Rate limiting SUBSCRIBE_UPDATE
Most control messages now consume a request ID, which limits the number an endpoint can send without additional credit from the peer.  SUBSCRIBE_UPDATE doesn't use a new request ID -- it uses the one from SUBSCRIBE/PUBLISH, which means the subscriber can send an unlimited number.  This seems like a DoS vector.

afrind:
Add per-stream flow control for REQUEST_UPDATE messages via a new MAX_REQUEST_UPDATES Setup Option. Each REQUEST_OK or REQUEST_ERROR response restores one unit of capacity. Default is 1 if not present.

Fixes: #1063

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1193,6 +1193,11 @@ INVALID_AUTHORITY (0x19):
 MALFORMED_AUTHORITY (0x1A):
 : The AUTHORITY value is syntactically invalid.
 
+TOO_MANY_REQUEST_UPDATES (0x1B):
+: The endpoint received a REQUEST_UPDATE that exceeded the per-stream limit
+  communicated via the MAX_REQUEST_UPDATES Setup Option
+  ({{max-request-updates}}).
+
 An endpoint MAY choose to treat a subscription or request specific error as a
 session error under certain circumstances, closing the entire session in
 response to a condition with a single subscription or message. Implementations
@@ -2618,6 +2623,26 @@ advertising or other nonessential information. Implementations SHOULD NOT use
 the identifiers of other implementations to declare compatibility, as this
 undermines the usefulness of implementation identification for debugging.
 
+#### MAX_REQUEST_UPDATES {#max-request-updates}
+
+The MAX_REQUEST_UPDATES option (Option Type 0x08) communicates the maximum
+number of unacknowledged REQUEST_UPDATE messages per request stream that
+the endpoint is willing to receive.
+
+A REQUEST_UPDATE is considered outstanding from when it is sent until the
+sender receives the corresponding REQUEST_OK or REQUEST_ERROR response.
+The sender MUST NOT have more than MAX_REQUEST_UPDATES outstanding
+REQUEST_UPDATEs on any single request stream at a time. Each REQUEST_OK
+or REQUEST_ERROR response restores one unit of capacity on that stream.
+
+The value is encoded as a variable-length integer. If not present, the default
+value is 1. A value of 0 means the endpoint will not accept any
+REQUEST_UPDATEs.
+
+If an endpoint receives a REQUEST_UPDATE on a stream that already has
+MAX_REQUEST_UPDATES outstanding REQUEST_UPDATEs, it MUST close the session
+with `TOO_MANY_REQUEST_UPDATES`.
+
 
 ## GOAWAY {#message-goaway}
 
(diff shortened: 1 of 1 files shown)


martinduke:
What happens when REQUEST_UPDATE is aggregated as described in the draft?

martinduke:
Actually I take it back, this doesn't solve the problem at all. The problem isn't pending REQUEST_UPDATEs, it's any REQUEST_UPDATE at all.

If the sender sends 1,000 REQUEST_UPDATES, skipping a valid ID each time, and each is OKed, I still have credit to do more requests, but the receiver still has to store 1,000 request IDs in case there is a reference to them later.

martinduke:
OK, we chatted online and I get it now. Given the number of authorized streams, there's a cap on the maximum possible request ID assuming the peer isn't skipping request IDs, which it shouldn't.

So this does finitely bound the non-contiguous request ID table.

However, this PR is missing any text that endpoints have to check the request ID against this theoretical maximum. That's crucial, and a little tricky to write.

draft-ietf-moq-transport.md
```diff
@@ -2618,6 +2623,26 @@ advertising or other nonessential information. Implementations SHOULD NOT use
...
+
+The MAX_REQUEST_UPDATES option (Option Type 0x08) communicates the maximum
+number of unacknowledged REQUEST_UPDATE messages per request stream that
+the endpoint is willing to receive.
+
+A REQUEST_UPDATE is considered outstanding from when it is sent until the
+sender receives the corresponding REQUEST_OK or REQUEST_ERROR response.
+The sender MUST NOT have more than MAX_REQUEST_UPDATES outstanding
+REQUEST_UPDATEs on any single request stream at a time. Each REQUEST_OK
+or REQUEST_ERROR response restores one unit of capacity on that stream.
```
afrind: I think we need to mention

1) how this interacts with REQUEST_UPDATE coalescing -- the receiver can coalesce but the sender is charged one credit per control message
2) the sender can assume success and pipeline REQUEST_UPDATEs, somewhat defeating the rate-limiting mechanism here if the receiver has a naive receive -> process -> ok.
3) This overlaps with FIFO processing rules mentioned in #1653

vasilvv (approved):

draft-ietf-moq-transport.md
```diff
@@ -2618,6 +2623,26 @@ advertising or other nonessential information. Implementations SHOULD NOT use
...
+
+The value is encoded as a variable-length integer. If not present, the default
+value is 1. A value of 0 means the endpoint will not accept any
+REQUEST_UPDATEs.
```
suhasHere: From the issue discussion the default is infinite and that would be my preference too. From what i recall from London interim, we came to conclusion that DDoS work should drive some of these too.

suhasHere (commented):
This seems fine except for the default value.

draft-ietf-moq-transport.md
```diff
@@ -2618,6 +2623,26 @@ advertising or other nonessential information. Implementations SHOULD NOT use
...
+the endpoint is willing to receive.
+
+A REQUEST_UPDATE is considered outstanding from when it is sent until the
+sender receives the corresponding REQUEST_OK or REQUEST_ERROR response.
+The sender MUST NOT have more than MAX_REQUEST_UPDATES outstanding
+REQUEST_UPDATEs on any single request stream at a time. Each REQUEST_OK
+or REQUEST_ERROR response restores one unit of capacity on that stream.
+
+The value is encoded as a variable-length integer. If not present, the default
+value is 1. A value of 0 means the endpoint will not accept any
```
```suggestion
value is 2^64 - 1. A value of 0 means the endpoint will not accept any
```

draft-ietf-moq-transport.md
afrind: Update to 2^64 - 1

draft-ietf-moq-transport.md
```diff
@@ -2618,6 +2623,26 @@ advertising or other nonessential information. Implementations SHOULD NOT use
...
+the endpoint is willing to receive.
+
+A REQUEST_UPDATE is considered outstanding from when it is sent until the
+sender receives the corresponding REQUEST_OK or REQUEST_ERROR response.
+The sender MUST NOT have more than MAX_REQUEST_UPDATES outstanding
+REQUEST_UPDATEs on any single request stream at a time. Each REQUEST_OK
+or REQUEST_ERROR response restores one unit of capacity on that stream.
+
+The value is encoded as a variable-length integer. If not present, the default
+value is 2^64 - 1. A value of 0 means the endpoint will not accept any
```
ianswett: We really want to disallow request update?  I think that's going to break applications in a painful way, so I don't think it's a good idea.

ianswett (changes requested):
REQUEST_UPDATE feels like a feature MoQ applications are going to depend upon.  Not allowing it at all seems like asking for problems.

draft-ietf-moq-transport.md
```diff
@@ -2618,6 +2623,26 @@ advertising or other nonessential information. Implementations SHOULD NOT use
...
+REQUEST_UPDATEs.
```
```suggestion
The value is encoded as a variable-length integer. A value of 0 means the
endpoint does not limit REQUEST_UPDATE concurrency. If not present, the default
value is 0.
```

ianswett (approved):

suhasHere (approved):

PR #1615, opened 2026-04-27
Remove Required Request ID

### LINKED ISSUE #1603: What is the use case for required-request-id
I see that draft-17 has required-request-id in every type of request. I don't see the value of this.

REQUEST_UPDATE might need it to sequence forward modes, etc

FETCH needs it for joining FETCH, obviously

Other than that, the only dependencies are on request termination, which doesn't have a request ID.

If there are this many dependencies we shouldn't have separated out the control stream.

I propose we eliminate this field except for REQUEST_UPDATE and FETCH (unless #1602 makes the FETCH dependency obvious)

ianswett:
Fixes #1603

Removes 'Required Request ID'.  Does not remove Request ID, because it is used by Joining Fetch and GOAWAY.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1124,10 +1124,6 @@ INVALID_REQUEST_ID (0x4):
 : The endpoint received a Request ID with an incorrect least significant
   bit for the sender, or a duplicate Request ID. See {{request-id}}.
 
-INVALID_REQUIRED_REQUEST_ID (0x7):
-: The endpoint received a Required Request ID Delta that results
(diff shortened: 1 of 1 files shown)


vasilvv (approved):

afrind:
Removing RRID creates races between REQUEST_UPDATE FWD=1 and Joining FETCH (rejoining a paused subscription).  At least, #1609 is required, so it's a request rather than a session error.  I guess the subscriber will either need to wait for the REQUEST_OK and send a FETCH, or yolo and handle the INVALID_RANGE error by retrying.

afrind (approved):

suhasHere (approved):

PR #1617, opened 2026-04-28
Allow GOAWAY on request streams to migrate individual requests

### LINKED ISSUE #1481: Do we need a way to move / goaway for individual track

afrind:
GOAWAY can now appear on a request stream using the same wire format as the control-stream form, but without Request ID (optional, present only on control stream).  A client MUST send a zero-length URI in any GOAWAY. Upon receiving a per-request GOAWAY, the endpoint re-issues that request on a session at the specified URI and closes the old stream.

Fixes: #1481

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3606,7 +3606,6 @@ SUBSCRIBE_TRACKS Message {
   Type (vi64) = 0x51,
   Length (16),
   Request ID (vi64),
-  Required Request ID Delta (vi64),
   Track Namespace Prefix (..),
   Number of Parameters (vi64),
   Parameters (..) ...
@@ -3615,8 +3614,6 @@ SUBSCRIBE_TRACKS Message {
 {: #moq-transport-subscribe-tracks-format title="MOQT SUBSCRIBE_TRACKS Message"}
 
 * Request ID: See {{request-id}}.
-
-* Required Request ID Delta: See {{required-request-id}}.
 
 * Track Namespace Prefix: A Track Namespace structure as described in
   {{track-name}} with between 0 and 32 Track Namespace Fields.  This prefix is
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2643,37 +2651,41 @@ GOAWAY Message {
   New Session URI Length (vi64),
   New Session URI (..),
   Timeout (vi64),
-  Request ID (vi64),
+  [Request ID (vi64)],
```
ianswett, line 2695: I think we should just remove Request ID.  I really don't think it's useful.

draft-ietf-moq-transport.md
```diff
@@ -2643,37 +2651,41 @@ GOAWAY Message {
...
-  indicated value, as well as any requests that arrive after the GOAWAY, MUST be
-  rejected with REQUEST_ERROR using error code GOING_AWAY. Requests with a
-  Request ID less than the indicated value were or might have been processed;
-  their status can be determined from the response on each request stream.
+* Timeout: The time in milliseconds the sender will wait for graceful closure.
+  When sent on the control stream, the sender closes the session with
+  `GOAWAY_TIMEOUT` after the indicated timeout if there are still open requests.
+  When sent on a request stream, the sender SHOULD reset the stream with
+  `GOING_AWAY` after the indicated timeout.  A value of 0 indicates the sender has no
+  specific timeout, and the recipient SHOULD still migrate as quickly as
```
```suggestion
  specific timeout, but the recipient SHOULD migrate as quickly as
```

draft-ietf-moq-transport.md
```diff
@@ -2643,37 +2651,41 @@ GOAWAY Message {
...
+  possible. This is a hint; the sender of the GOAWAY MAY close the session or
+  reset the stream before the indicated timeout has elapsed.
```
```suggestion
  reset the request stream before the indicated timeout has elapsed.
```

ianswett (commented):
I think we should remove Request ID from GOAWAY entirely, since I don't think it has much practical value.

draft-ietf-moq-transport.md
ianswett: I wrote #1623

afrind: This is in a separate PR and whoever loses the merge race can address

ianswett (approved):

vasilvv (approved):

suhasHere (approved):

PR #1618, opened 2026-04-28
Add FIRST_OBJECT bit to SUBGROUP_HEADER type

afrind:
Add bit 6 (0x40) to signal that the subgroup contains the first object published in the subgroup by the original publisher. The type format expands from 0b00X1XXXX to 0b0XX1XXXX. All valid type values still fit in a 1-byte varint.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1083,7 +1083,7 @@ the type of the stream.
 |------------:|:------------------------------------------------|
 | 0x05        | FETCH_HEADER  ({{fetch-header}})                |
 |-------------|-------------------------------------------------|
-| 0x10-0x15 / 0x18-0x1D / 0x30-0x35 / 0x38-0x3D | SUBGROUP_HEADER  ({{subgroup-header}}) |
+| 0b0XX1XXXX  | SUBGROUP_HEADER  ({{subgroup-header}})          |
 |-------------|-------------------------------------------------|
 | 0x2F00      | SETUP ({{message-setup}})                       |
 |-------------|-------------------------------------------------|
@@ -3800,7 +3800,8 @@ flow control, while the sender waits for flow control to send the message.
 
 ~~~
 SUBGROUP_HEADER {
-  Type (i) = 0x10..0x15 / 0x18..0x1D / 0x30..0x35 / 0x38..0x3D,
+  Type (i) = 0x10..0x15 / 0x18..0x1D / 0x30..0x35 / 0x38..0x3D /
+             0x50..0x55 / 0x58..0x5D / 0x70..0x75 / 0x78..0x7D,
   Track Alias (vi64),
   Group ID (vi64),
   [Subgroup ID (vi64),]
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -3845,15 +3847,23 @@ and bit 5 determine which fields are present in the header:
   established the subscription. When set to 0, the Priority field is present in
   the Subgroup header.
 
+* The **FIRST_OBJECT** bit (0x40) indicates that this subgroup contains the
+  first object published in the subgroup by the original publisher. A relay
+  forwarding a subgroup MUST preserve the value of this bit from the upstream
+  publisher. A subscriber can use this signal to identify the subgroup that
```
ianswett: Should we require use of this?  Are there valid cases where it would be impractical to know what the first Object is?

ianswett (commented):
Ideally, I'd like this to be required, which was a perk of my other proposal to force the Subgroup ID==First Object ID

draft-ietf-moq-transport.md
afrind: An original publisher really ought to know if it's the beginning or not.  A relay should know then either because it just received this information or by caching it. 

That said, it's not catastrophic if it's not remembered, at least for the relay.  It's primary value is to know when it's safe to serve the beginning of a subgroup from cache vs going upstream.

suhasHere: if the subscriber/relay uses object status to make decision, this is at the same level. Whatever, we say there should be said here too or atleast they both should match

suhasHere (approved):
Looks fine to me

afrind:
One question came up: does this signal need to be replicated in FETCH?  The answer is that we could, but we should do that for both begin and end of subgroup.  It's unclear if there's application value.

draft-ietf-moq-transport.md
```diff
@@ -558,6 +558,11 @@ When assigning Objects to different Subgroups, the Original Publisher makes a
 reasonable tradeoff between having an optimal mapping of Object relationships in
 a Group and minimizing the number of streams used.
 
+When the Original Publisher opens a new subgroup, it MUST set the FIRST_OBJECT
+bit ({{subgroup-header}}) to indicate that the subgroup begins with the first
+object published in that subgroup. A relay forwarding a subgroup MUST preserve
+the value of this bit from the upstream publisher.
```
vasilvv: This does not sound like the correct behavior.  It can only do this if the first object ID on the downstream subscription is the same as the upstream subscription.

draft-ietf-moq-transport.md
```diff
@@ -3845,15 +3852,20 @@ and bit 5 determine which fields are present in the header:
   established the subscription. When set to 0, the Priority field is present in
   the Subgroup header.
 
+* The **FIRST_OBJECT** bit (0x40) indicates that this subgroup contains the
+  first object published in the subgroup by the original publisher.
```
```suggestion
* The **FIRST_OBJECT** bit (0x40) indicates that the first object in this
subgroup stream is the first object published in the subgroup by the original publisher.
```

draft-ietf-moq-transport.md
```diff
@@ -558,6 +558,11 @@ When assigning Objects to different Subgroups, the Original Publisher makes a
 reasonable tradeoff between having an optimal mapping of Object relationships in
 a Group and minimizing the number of streams used.
 
+When the Original Publisher opens a new subgroup, it MUST set the FIRST_OBJECT
+bit ({{subgroup-header}}) to indicate that the subgroup begins with the first
+object published in that subgroup. A relay forwarding a subgroup MUST preserve
```
```suggestion
bit ({{subgroup-header}}) to indicate that the subgroup begins with the first
object published in that subgroup stream. A relay forwarding a subgroup MUST preserve
```

ianswett (commented):
I think this is made more confusing because there's a 'Subgroup', which is a sequence of Objects produced by the original publisher and they're an Object model concept.  Then there's how Subgroups are delivered, which we also call Subgroups, but are really Subgroup Streams or something more specific.

draft-ietf-moq-transport.md
afrind: Agreed, let me make this more clear

```suggestion
When the Original Publisher opens a new subgroup, it MUST set the FIRST_OBJECT
bit ({{subgroup-header}}) to indicate that the first object in the subgroup
stream is the first object ever published in that subgroup. A relay forwarding a
subgroup that begins with the first object ever published in that subgroup MUST
set the FIRST_OBJECT bit.
```

ianswett (approved):

afrind:
Screenshot 2026-05-11 at 2 54 50 PM

Evidence of Victor approval

PR #1621, opened 2026-04-28
Forbid relays from lying about LARGEST_OBJECT

### LINKED ISSUE #1386: Can a publisher 'lie' about what Largest Object is? And does Largest Object need to exist?
I believe the answer is that its ok if the Publisher lies, as long as it's internally consistent within the Session and that there's no need for the Largest Object to actually exist. Previously we've said it's ok for the Publisher to start at the beginning of what is actually the Largest Object if it decides to, instead of waiting for the next largest to arrive, allowing it to immediately start sending data.

What I'm thinking is that if a Publisher/Relay knows that the Subscriber wants to start at the beginning of the Group, ideally explicitly (PR# 1362), it can say that the Largest Object is in the prior Group and give the Subscriber the entire Subscription via SUBSCRIBE, avoiding splitting the Subscription between FETCH and SUBSCRIBE streams and avoiding the Object Unknown issues entirely.

The only challenge I can see is a Schrodinger's cat sort of situation where if one did a TRACK_STATUS, followed by a SUBSCRIBE, it'd be weird to have Largest Object go backwards, but if they're on separate streams, there might be cases that'd happen anyway.

afrind:
If we want to serve cached objects in response to SUBSCRIBE, lying is not the correct approach.

Fixes: #1386

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2446,11 +2446,11 @@ have been published on this Track the Publisher MUST include this parameter.
 If omitted from a message, the sending endpoint has not published or received
 any Objects in the Track.
 
-A relay MUST NOT populate the LARGEST_OBJECT parameter from the contents
-of its cache. A relay MUST set LARGEST_OBJECT to the largest of the
-LARGEST_OBJECT values received from the upstream publisher in SUBSCRIBE_OK,
-PUBLISH, or REQUEST_UPDATE_OK, or the Location of an Object received on
-the upstream subscription, whichever is largest.
+A relay MUST set LARGEST_OBJECT to the largest of the following:
+
+1. Any LARGEST_OBJECT value received from the upstream publisher in SUBSCRIBE_OK,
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -2445,6 +2446,12 @@ have been published on this Track the Publisher MUST include this parameter.
 If omitted from a message, the sending endpoint has not published or received
 any Objects in the Track.
 
+A relay MUST NOT populate the LARGEST_OBJECT parameter from the contents
+of its cache. A relay MUST set LARGEST_OBJECT to the largest of the
```
ianswett: I think this first sentence is potentially confusing and the second sentence is sufficient.

```suggestion
A relay MUST set LARGEST_OBJECT to the largest of the
```

vasilvv (approved):

ianswett (approved):

PR #1627, opened 2026-05-03
SUBSCRIBE with Joining Fetch

### LINKED ISSUE #1039: Simplifying joining at the latest available join point.

### LINKED ISSUE #1313: Joining FETCH as a separate control message creates edge cases and feature gaps

### LINKED ISSUE #1602: Joining Fetch should be on the SUBSCRIBE/PUBLISH stream

### LINKED ISSUE #1612: What happens to Joining FETCH if fwd changes to 0?

ianswett:
A different take on #1604 that adds two new modes to SUBSCRIBE instead of allowing Joining FETCH to be sent on the SUBSCRIBE stream.

Notes:
* Might be able to removed Request ID once #1615 lands.
* Could use some more text on what happens to the FETCH_HEADER stream if the Subscription is cancelled
* Could use clarification on how prioritization of Objects is intended to work.

Fixes #1039
Fixes #1313 
Fixes #1602
Fixes #1612

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1343,8 +1343,6 @@ Forward State in either PUBLISH or SUBSCRIBE.  The subscriber can send PUBLISH_O
 or REQUEST_UPDATE to update the Forward State. Control messages, such as
 PUBLISH_DONE ({{message-publish-done}}) are sent regardless of the forward state.
 
-
-
 Either endpoint can initiate a subscription to a track without exchanging any
 prior messages other than SETUP.  Relays MUST NOT send any PUBLISH messages
 without knowing the client is interested in and authorized to receive the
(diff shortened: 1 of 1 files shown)


gwendalsimon:
Passing an absolute or relative Group start in SUBSCRIBE is easy. Removing the current Joining FETCH section is easy too. That is not where the WG has been stuck. Some points have blocked consensus for months:
- Placement of catch-up Objects. Are they carried on a separate unidirectional stream, or in-band on the parent bidi? With an extra stream, boundary signaling, cancellation, and lifecycle are still split across streams.
- Cache gaps. If an Object is missing from cache, does the relay wait for it, with possible HoL blocking, or skip it, which makes catch-up unreliable? This needs an explicit rule. FILL_TIMEOUT looks like the right vehicle.
- Continuity model. Are past Objects required to be delivered before live Objects, to preserve continuity, or can past and live be sent concurrently? These are different semantics. 
- Drain behavior. If continuity is preserved by serving catch-up first, then live Objects accumulate while catch-up is being sent. Then the relay behavior must be defined: how backlog is drained, what “caught up” means, and what to do if drain does not converge (TOO_FAR_AWAY and TIMEOUT_DELIVERY seem good vehicles but they introduce gaps).
- Scope. This PR defines a SUBSCRIBE-side trigger. That may be fine, but it should not be confused with defining the catch-up transport itself.

In short, this PR simplifies the syntax. It does not yet define the semantics.

martinduke (changes requested):
FILL_TIMEOUT has to be allowed in a SUBSCRIBE message as well, but is only meaningful with a Join filter. I don't think DELIVERY_TIMEOUT can work with a FETCH stream.

ianswett:
Closing in favor of #1642

PR #1628, opened 2026-05-11
Add QMux framing for moqt-18 over TLS+TCP

### LINKED ISSUE #1626: Version negotiation for QMUX
We have an idea of how version negotiation works for MoQ-over-HTTP/3 and how it works for MoQ-over-QUIC, but do we know how it's going to work for MoQ-over-QMUX?

afrind:
When the moqt-18 ALPN is negotiated over TLS+TCP, the underlying framing uses QMux version 1.

Fixes: #1626

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -829,11 +829,6 @@ connection is established. The QUIC DATAGRAM extension ({{!RFC9221}})
 MUST be supported and negotiated in the QUIC connection used for MOQT,
 which is already a requirement for WebTransport over HTTP/3.
 
-When the moqt-18 ALPN is negotiated over TLS+TCP, the underlying
-framing uses QMux version 1 {{!QMUX=I-D.ietf-quic-qmux-01}}.
-
-\[\[RFC editor: please remove the previous paragraph before publication.]]
-
 MOQT uses ALPN in QUIC and "WT-Available-Protocols" in WebTransport
 ({{WebTransport, Section 3.3}}) to perform version negotiation.
 
@@ -842,7 +837,13 @@ MOQT uses ALPN in QUIC and "WT-Available-Protocols" in WebTransport
 The ALPN value {{!RFC7301}} for the final version of this specification
 is `moqt`.  ALPNs used to identify IETF drafts are created by appending
 the draft number to "moqt-". For example, draft-ietf-moq-transport-13
-would be identified as "moqt-13".
+would be identified as "moqt-13".  When using QMux {{QMUX}} over
+TLS+TCP, ALPNs are created by appending the draft number to "moqt-qx-".
+For example, draft-ietf-moq-transport-18 over QMux would be identified
+as "moqt-qx-18".
+
+When the moqt-qx-18 ALPN is negotiated over TLS+TCP, the underlying
+framing uses QMux version 1 {{!QMUX=I-D.ietf-quic-qmux-01}}.
 
 Note: Draft versions prior to -15 all used moq-00 ALPN, followed by version
 negotiation in the SETUP messages.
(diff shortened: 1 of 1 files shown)


LPardue:
This would appear to violate https://quicwg.org/qmux/draft-ietf-quic-qmux.html#section-8.1-2

> ALPN protocol identifiers identify the application protocol in use. Application protocols that use QMux over TLS MUST designate their ALPN identifier and specify that they use QMux version 1. The identifier for a mapping to QMux MUST be different from the mapping of the same protocol to QUIC, to retain compatibility with Service Binding and Parameter Specification via the DNS [[SVCB](https://quicwg.org/qmux/draft-ietf-quic-qmux.html#SVCB)].

vasilvv:
I am confused.  Why does this draft require QMux-specific text?

afrind:
> I am confused. Why does this draft require QMux-specific text?

We need to write down the version mapping from MOQT to QMux somewhere, but if you feel strongly we can do it in another draft, but that seems like overkill to me.

vasilvv:
Are you saying using QMux with an application protocol that is already defined to work over QUIC requires QMux integration?  This sounds like a serious design issue in QMux.

afrind:
> Are you saying using QMux with an application protocol that is already defined to work over QUIC requires QMux integration? This sounds like a serious design issue in QMux.

Application protocols that run over QUIC have to specify the versions of QUIC they run over.  QMux is no different.  And this may be more of a "while there are draft versions of qmux in flight" problem, like it was when we said "h3-24 runs on quic-14" or whatever.

suhasHere:
I do have this draft: https://datatracker.ietf.org/doc/draft-nandakumar-moq-qmux-moqt/ to discuss more specific stuff and there are edge cases, esp in enterprises , where we need to handle TCP to UDP flows too.

I don't feel we need need to hurry in adding text here and should have some discussions ? Also would love to see if there are feedback on the above draft too.

ianswett (approved):

LPardue:
The technical approach to uses ALPN looks correct now. I can't add much opinion on if it should be in this draft or not though.

draft-ietf-moq-transport.md
```diff
@@ -841,7 +837,13 @@ MOQT uses ALPN in QUIC and "WT-Available-Protocols" in WebTransport
...
 is `moqt`.  ALPNs used to identify IETF drafts are created by appending
 the draft number to "moqt-". For example, draft-ietf-moq-transport-13
-would be identified as "moqt-13".
+would be identified as "moqt-13".  When using QMux {{QMUX}} over
+TLS+TCP, ALPNs are created by appending the draft number to "moqt-qx-".
+For example, draft-ietf-moq-transport-18 over QMux would be identified
+as "moqt-qx-18".
+
+When the moqt-qx-18 ALPN is negotiated over TLS+TCP, the underlying
+framing uses QMux version 1 {{!QMUX=I-D.ietf-quic-qmux-01}}.
```
ianswett, line 846: nit: Is this QMux version 1 or draft 1?

draft-ietf-moq-transport.md
```diff
@@ -906,8 +908,10 @@ and `h3` that it supports in its TLS ClientHello, in preference order. If the
 server selects an MOQT ALPN, the session proceeds as described in
 {{native-quic}}. If the server selects `h3`, the client establishes a
 WebTransport session as described in {{webtransport}}. On a TCP+TLS
-connection, the client offers `h2` in its TLS ClientHello and establishes a
-WebTransport session as described in {{webtransport}}.
+connection, the client offers MOQT QMux ALPNs (e.g. `moqt-qx-18`) and/or
```
ianswett, line 911: nit: I'd probably put these in the reverse order, given I expect h2 is a lot more common.

ianswett (approved):
Some nits, but looks good.

gloinul:
As Chair I do have questions about adding this in the draft. It does create a normative reference for a work that isn't particular mature yet to my understanding. So the QMUX was discussed in Shenzhen but I don't remember a consensus call on adding this functionality.

gloinul:
As communicated on the mailing list in: https://mailarchive.ietf.org/arch/msg/moq/G3WnGnXMcTmTJGi-CD29mDOwScI/ 
This is outside of the current charter and shall not be merged in a WG document. It has to live in an individual draft for now.

PR #1638, opened 2026-06-01
Add Sender-Side Track Switching (SSTS)

### LINKED ISSUE #259: Sender-side ABR
The basic problem is: the publisher provides two versions of the same track, let's say, `video/360p` and `video/720p`; the subscriber wants to receive only one of those, and it best quality it can receive in a timely manner.  It makes more sense to make this switch at the sender, since the sender's congestion controller ultimately decides how much data the sender is going to send short-term.

@kixelated proposed before to allow the sender to subscribe to a list of tracks in order of preference, so something like `SUBSCRIBE_REQUEST video/720p video/360p` would result in 720p video being sent preferentially when available.  I like this approach, though we do need to specify more details first.  For instance, what are the conditions under which the sender is allowed to switch?  I think those should be group boundaries (and we should also require alignment between tracks if you want to subscribe to those as an alternative).

wilaw:
Adds the Sender-Side Track Switching (SSTS) feature, detailing subscriber operations, relay behavior, and bandwidth allocation algorithm.

Adds a new SWITCHING-SET-ASSIGNMENT parameter. 

Fixes #259

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2299,7 +2299,7 @@ Relays prioritize forwarded Objects as described in {{priorities}}.
 ## Sender Side Track Switching {#sender-side-track-switching}
 
 Sender Side Track Switching (SSTS) is a subscriber-initiated and controlled behavior in which
-relays dynamically select which track to forward from a switching set based on
+a publisher dynamically selects which track to forward from a switching set based on
 various algorithms. Each algorithm defines a set of attributes which are passed in the
 SWITCHING-SET-ASSIGNMENT parameter {{switching-set-assignment-param}} along with a
 complimentary set of rules for subscriber behavior and relay behavior.
@@ -2308,7 +2308,7 @@ This specification defines a default algorithm - type 0. Other algorithms are re
 "SSTS Algorithms" registry {{iana-ssts-algorithms}}.
 
 During SETUP, a relay communicates which SSTS algorithms it supports by passing the
-SSTS_ALGORITHMS option {{ssts-algorithms}}.
+SSTS_ALGORITHMS Setup Option {{ssts-algorithms}}.
 
 ### General behaviors for all SSTS algorithms {#ssts-general-requirements}
 
@@ -2316,10 +2316,10 @@ The subscriber is responsible for grouping tracks into switching sets based on a
 knowledge. A switching set is a collection of tracks representing the same content encoded at
 different throughput levels, typically from a single source. Tracks within a switching set are
 time-aligned at certain group boundaries, allowing the relay to switch between tracks at these
-boundaries without initiating reception errors at the client. The relay selects exactly one track
-per switching set to forward at any given time.
+boundaries while ensuring the subscriber receives uninterrupted content from the set. The relay
+selects exactly one track per switching set to forward at any given time.
 
-Subscribers can create switching sets through two methods. Both support single or multiple
+Subscribers can create switching sets through three methods. All support single or multiple
 switching sets and result in identical relay behavior:
 
 * Individual SUBSCRIBE: - the subscriber sends a separate SUBSCRIBE message for each track,
@@ -2329,12 +2329,14 @@ switching sets and result in identical relay behavior:
   track, the relay will issue a PUBLISH message. The subscriber assigns tracks to switching sets
   by appending the SWITCHING-SET-ASSIGNMENT parameter to the PUBLISH_OK message.
 
-In both cases, tracks are grouped into a switching set by specifying the same switching set ID.
+* PUBLISH: - the publisher sends a PUBLISH message. The subscriber assigns tracks to switching sets
+  by appending the SWITCHING-SET-ASSIGNMENT parameter to the PUBLISH_OK message.
+
+In all cases, tracks are grouped into a switching set by specifying the same switching set ID.
 
 ### Default switching algorithm
 
-This specification defines a default SSTS algorithm and adds it to the "MOQT SSTS Algorithms"
-registry {{iana-ssts-algorithms}} with a type of 0.
+This specification defines a default SSTS algorithm with a type of 0.
 
 #### SWITCHING-SET-ASSIGNMENT fields
 This algorithm extends the base definition of the SWITCHING-SET-ASSIGNMENT parameter
@@ -2344,39 +2346,40 @@ SWITCHING-SET-ASSIGNMENT {
   Switching set ID (vi64),
   Algorithm ID (vi64),
   Throughput threshold (vi64),
-  Set throughput fraction (vi64),
+  Set throughput weight (vi64),
   Activate switching (vi64),
   Set rank (8)
 }
 
 * Throughput threshold: Minimum throughput (kbps) required to select this track.
 
-* Set throughput fraction: Relative weight for bandwidth allocation, expressed as an
-  integer 1 <= N <= 10. Each set receives bandwidth proportional to its fraction:
-  `target = B_total × fraction / sum_F`. Fractions are relative weights, not absolute
-  percentages; for example, fractions of 6, 4, 3 (sum_F=13) allocate 46%, 31%, 23%
+* Set throughput weight: Relative weight for bandwidth allocation, expressed as an
+  integer 1 <= N <= 10. Each set receives bandwidth proportional to its weight:
+  `target = B_total × weight / sum_F`. These are relative weights, not absolute
+  percentages; for example, weights of 6, 4, 3 (sum = 13) allocate 46%, 31%, 23%
   respectively. This allows sets to be added or removed without requiring other sets to
-  update their fractions. When multiple subscriptions in the same switching set specify
-  different fraction values, the relay MUST use the value from the most recently received
+  update their weights. When multiple subscriptions in the same switching set specify
+  different weight values, the publisher MUST use the value from the most recently received
   message for that set.
 
 * Activate switching: Integer, when set to 0, pauses SSTS switching for this set. When set
   to N, the relay activates or resumes switching as soon as the number of tracks assigned to
-  the switching set is >= N.  Activation takes effect at the next group boundary.
+  the switching set is >= N.  Activation takes effect when an Object is received or published
+  on a Group larger than previously largest Group. When multiple subscriptions in the same
+  switching set specify different activate values, the publisher MUST use the value from the
+  most recently received message for that set.
 
-* Set rank: Degradation priority when bandwidth is constrained, expressed as
-  an 8-bit unsigned integer (1-255). Default is 1. Values outside this range MUST result in
-  a protocol error. Lower values indicate higher priority (protected from degradation).
-  When bandwidth is sufficient, all sets receive their fraction-based allocation. When
-  bandwidth is constrained, lower-ranked sets (higher numeric value) degrade first: they
-  select lower-throughput tracks or receive no bandwidth, while higher-ranked sets maintain
-  their target allocation. See {{allocation-algorithm}} for details.
+* Set rank: Degradation priority when bandwidth is constrained, expressed as an 8-bit unsigned
+  integer (0-255). Default is 0. Lower values indicate higher priority (protected from degradation).
+  See {{allocation-algorithm}} for details. When multiple subscriptions in the same switching set
+  specify different rank values, the publisher MUST use the value from the most recently received
+  message for that set.
 
 #### Subscriber behavior
 
 The subscriber follows the general rules {#ssts-general-requirements} for switching set establishment.
 
-The subscriber sets activate = N, where N is the number of tracks that will be assigned to that
+The subscriber sets activate switching = N, where N is the number of tracks that will be assigned to that
 switching set.
 
 To modify an established switching set, the subscriber can
@@ -2387,58 +2390,61 @@ To modify an established switching set, the subscriber can
 * Pause SSTS: Send REQUEST_UPDATE for any track assigned to that set with a SWITCHING-SET-ASSIGNMENT
   parameter defining activate = 0.
 * Resume SSTS: Send REQUEST_UPDATE for any track assigned to that set with a SWITCHING-SET-ASSIGNMENT
-  parameter defining activate = N, where N is the number of tracks assigned to that switching set.
+  parameter defining activate switching = N, where N is the number of tracks assigned to that switching set.
 
-#### Relay behavior
+#### Publisher behavior
 
-When the relay receives a subscription with SWITCHING-SET-ASSIGNMENT:
+When the publisher receives a subscription with SWITCHING-SET-ASSIGNMENT:
 
 1. Add the subscription to the specified switching set, creating the set if needed.
-2. Set Forward state to 0 for the new subscription.
-3. Store 'throughput' as a property of the subscription.
-4. Store 'fraction', 'rank' and 'activate' as properties of the set.
-5. If the number of tracks assigned to the set is >= the activate switching value, then begin
-   active track selection by applying the bandwidth allocation algorithm {{allocation-algorithm}} at the
-   next group boundary.
+2. Set Forward state to 0 for the new subscription, irrespective of the forward state received from the
+   SUBSCRIBE or PUBLISH_OK.
+4. Store 'throughput threshold' as a property of the subscription.
+5. Store 'Set throughput weight', 'Set rank' and 'Activate switching' as properties of the set.
+6. If the number of tracks assigned to the set with active subscriptions >= the activate switching value,
+   then begin active track selection by applying the bandwidth allocation algorithm {{allocation-algorithm}}
+   when an Object is received or published on a Group larger than previously largest Group.
 
-If the relay receives a PUBLISH_DONE message, or an UNSUBSCRIBE for a subscription that was
+If the publisher receives a PUBLISH_DONE message, or an UNSUBSCRIBE for a subscription that was
 previously added to a switching set, then it must remove that subscription from the switching set
-and continue to process the switching across the remaining subscriptions within that set.
+and continue to process the switching across the remaining subscriptions within that set. The value of
+'activate switching' MUST be decremented by one to enable the swictching to remain active. 
 
 If all tracks are removed from a previously established switching set, then that set is
 considered deleted and is removed from the bandwidth allocation algorithm.
 
-Relays SHOULD maintain a forward=1 upstream state on all tracks within a switching set, irrespective
-of their downstream forwarding state.
+Publishers SHOULD maintain a forward=1 upstream state (if any) on all tracks within a switching set,
+irrespective of their downstream forwarding state.
 
 #### Bandwidth Allocation {#allocation-algorithm}
 
-The relay maintains:
+The publisher maintains:
 
 - `B_total`: Estimated downstream bandwidth capacity for the subscriber connection. The
-  relay MUST maintain a bandwidth estimate for each downstream subscriber. The timebase of this
-  estimate SHOULD be at least the Group duration of the track. The estimate is obtained
-  periodically from the QUIC stack (e.g., congestion window pacing rate, smoothed RTT)
+  publisher maintains a bandwidth estimate for each downstream subscriber. The timebase of this
+  estimate SHOULD be at least the Group duration of the track, if that is known or can be estimated
+  by the publisher, or several seconds if it is unknown. The estimate is obtained
+  periodically from the transport stack (e.g., congestion window pacing rate, smoothed RTT)
   and MAY be supplemented by external sources or application-level feedback. The exact mechanism
-  is not defined by this specification and MAY vary between implementations.
-- `sum_F`: Sum of all set fractions, updated incrementally as subscriptions are added or removed
-- 'set.fraction': for each switching set, the switching set fraction, as defined by the set
-  throughput fraction of the SWITCHING-SET-ASSIGNMENT {{switching-set-assignment-param}} parameter.
-- 'set.rank': for each switching set, the switching set rank, as defined by the optional set
+  is not defined by this algorithm and might vary between implementations.
+- `sum_W`: Sum of all set weights updated incrementally as subscriptions are added or removed
+- 'set.weight': for each switching set, the switching set weight, as defined by the set
+  throughput weight of the SWITCHING-SET-ASSIGNMENT {{switching-set-assignment-param}} parameter.
+- 'set.rank': for each switching set, the switching set rank, as defined by the set
   rank field of the SWITCHING-SET-ASSIGNMENT {{switching-set-assignment-param}} parameter.
 
-On a periodic update interval or at a minimum of a new Group boundary, the relay executes the
-following algorithm:
+On a periodic update interval or at a minimum when an object is received/published on a group
+larger than previously largest group, the relay executes the following algorithm:
 
 ~~~
 B_remaining  = B_total
-for each set in ascending rank order:
-  set.target = B_remaining × set.fraction / sum_F
+for each set in ascending set.rank order:
+  set.target = B_remaining × set.weight / sum_W
   set.allocated = min(set.target, B_remaining)
   set.selected = track in set with highest throughput_threshold where track.throughput_threshold <= set.allocated
   B_remaining -= set.selected.throughput_threshold
 for each track in a switching set:
-  set forward state = track.selected
+  set forward state = (track == set.selected)
 ~~~
 
 The rank ordering ensures higher-priority sets receive their target allocation first; lower-priority sets
@@ -3116,8 +3122,7 @@ a sequence of varints. Returning an empty sequence is acceptable and indicates
 that SSTS is not supported. Algorithms are registered in the SSTS-Algorithms
 {{iana-ssts-algorithms}} registry.
 
-This option is optional. The default value is any empty list which prohibits
-the use of SSTS.
+The default value is any empty list which prohibits the use of SSTS.
 
 #### MAX FILTER RANGES
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2113,22 +2113,28 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
 combine, split, or otherwise modify object payloads.  A relay SHOULD
 prioritize sending Objects based on {{priorities}}.
 
-## Dynamic Track Switching {#dynamic-track-switching}
+## Sender Side Track Switching {#sender-side-track-switching}
 
-Dynamic Track Switching (DTS) is a subscriber-initiated and controlled behavior in which
+Sender Side Track Switching (SSTS) is a subscriber-initiated and controlled behavior in which
 relays dynamically select which track to forward from a switching set based on
-available downstream bandwidth.
+various algorithms. Each algorithm defines a set of attributes which are passed in the
+SWITCHING-SET-ASSIGNMENT parameter {{TBD}} along with a complimentary set of rules for subscriber
+operations and relay behavior.
 
-The subscriber is responsible for grouping tracks into switching sets based on
-application-level knowledge. A switching set is a collection of tracks representing
-the same content encoded at different throughput levels, typically from a single source.
-Tracks within a switching set are time-aligned at group boundaries, allowing the relay
-to switch between tracks at group boundaries without initiating reception errors at
-the client. The relay selects exactly one track per switching set to forward at any
-given time.
+This specification defines a default algorithm - type 0. Other algorithms are referenced in the
+"MOQT SSTS Algorithms" registry {{iana-ssts-algorithms}}.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2113,97 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
...
+* Resume DTS: Send REQUEST_UPDATE with a SWITCHING-SET-ASSIGNMENT parameter defining activate = 1
+
+### Relay behavior
+
+When the relay receives a subscription with SWITCHING-SET-ASSIGNMENT:
+
+1. Add subscription to the specified switching set, creating the set if needed.
+2. Set Forward state to 0 for the new subscription
+3. Store the Set throughput fraction and rank as properties of the set
+4. If Activate switching = 1, begin active track selection
```
```suggestion
4. If Activate switching = 1, begin active track selection, forwarding objects for selected track.
```

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2113,97 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
...
+
+- `B_total`: Estimated downstream bandwidth capacity for the subscriber connection. The
+  relay MUST maintain a bandwidth estimate for each downstream subscriber. The timebase of this
+  estimate SHOULD be at least the Group duration of the track. The estimate is obtained
+  periodically from the QUIC stack (e.g., congestion window pacing rate, smoothed RTT)
+  and MAY be supplemented by external sources or application-level feedback. The exact mechanism
+  is not defined by this specification and MAY vary between implementations. 
+- `sum_F`: Sum of all set fractions, updated incrementally as subscriptions are added or removed
+- 'set.fraction': for each switching set, the switching set fraction, as defined by the set
+  throughput fraction of the SWITCHING-SET-ASSIGNMENT {{switching-set-assignment-param}} parameter.
```
suhasHere: i think. it also maintains set.rank for a giving switching set

suhasHere (approved):
This looks good to me.

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2113,97 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
...
```
wilaw:
```suggestion
- 'set.fraction': for each switching set, the switching set fraction, as defined by the set
  throughput fraction of the SWITCHING-SET-ASSIGNMENT {{switching-set-assignment-param}} parameter.
  - 'set.rank': for each switching set, the switching set rank, as defined by the optional set
  rank field of the SWITCHING-SET-ASSIGNMENT {{switching-set-assignment-param}} parameter.
```

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2113,97 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
...
+
+Subscribers can create switching sets through two methods. Both support single or multiple
+switching sets and result in identical relay behavior:
+
+* Individual SUBSCRIBE: - the subscriber sends a separate SUBSCRIBE message for each track,
+  and appends the SWITCHING-SET-ASSIGNMENT parameter to assign the track to a switching set.
+
+* SUBSCRIBE_NAMESPACE: - the subscriber sends a SUBSCRIBE_NAMESPACE message. For each matching
+  track, the relay will issue a PUBLISH message. The subscriber assigns tracks to switching sets
+  by appending the SWITCHING-SET-ASSIGNMENT parameter to the PUBLISH_OK message.
```
sharmafb: Bit of a pedantic comment, but might SUBSCRIBE_TRACKS be a better fit for this, since the subscriber is going to receive a PUBLISH in response?

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2113,97 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
...
+
+In both cases, tracks are grouped into a switching set by specifying the same switching set ID.
+The subscriber sets activate=0 for all tracks except the last one in each set, then sets
+activate=1 on the final track to signal that the set is complete and the relay should
+begin active track selection.
```
sharmafb: "Final track" seems a little suspicious. Since the messages are in different streams, they can arrive in any order, right?

draft-ietf-moq-transport.md
```diff
@@ -2605,6 +2696,52 @@ Namespace Prefix for that subscription.  If the new prefix would share a common
...
+switching set and specifies bandwidth allocation and optional ranking values.
+
+~~~
+SWITCHING-SET-ASSIGNMENT {
+  Switching set ID (vi64),
+  Throughput threshold (vi64),
+  Set throughput fraction (vi64),
+  Activate switching (1),
+  [Set rank (8)]
+}
```
sharmafb, line 2994: I don't think this is safely parseable? How would a parser know if "Set rank" is present or not?

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2113,97 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
...
+- 'set.fraction': for each switching set, the switching set fraction, as defined by the set
+  throughput fraction of the SWITCHING-SET-ASSIGNMENT {{switching-set-assignment-param}} parameter.
+
+On a periodic update interval or at a minimum of a new Group boundary, the relay executes the
+following algorithm:
+
+~~~
+B_remaining  = B_total
+for each set in ascending rank order:
+  set.target = B_remaining × set.fraction / sum_F
```
sharmafb: This doesn't seem right to me. Shouldn't you use B_total here instead of B_remaining here?

draft-ietf-moq-transport.md
yuyou: Good catch. Ranking is not part of our PoC implementation. We always carry the rank field alongside the other properties instead of treating it as optional, and it is currently ignored by our Relay server.

Since ranking is difficult to test and does not seem necessary for subscribers to indicate, we have not implemented any ranking logic.

draft-ietf-moq-transport.md
wilaw: That is true. However, the out-of-order of stream delivery is likely to be in the order of 1/2 RTT at most, and this is likely << the Group duration, which is usually the GOP length for media applications. So it was thought to be acceptable. 

We did also consider a separate message to turn on/off the switching behavior, however even that message, itself being on a separate stream, is not guaranteed to arrive after the last track has been assigned to the switching set. 

One solution here is to change the activate flag from binary to a number, where 0 means turn off all switching and N means start switching once you have N tracks assigned to the set. Now the arrival order doesn't matter.

draft-ietf-moq-transport.md
wilaw: I think line 2194 is correct. Switching sets are allocated bandwidth based on their rank order. The highest ranked set has ALL of the available bandwidth allocated to it.  If it doesn't consume all that bandwidth, then what is left over (B_remaining) is allocated to the next ranked set. The algorithm is initialized by setting B_remaining to B_total, which effectively gives the highest ranked set access to B_total. All other sets are allocated something less than that.

draft-ietf-moq-transport.md
wilaw: Good catch. The initial draft of this PR was written prior to SUBSCRIBE_TRACKS being added to MOQT.
Fixed with https://github.com/moq-wg/moq-transport/pull/1638/changes/cf10fd3fb597f2070e659149958d96e1bc4f0446

draft-ietf-moq-transport.md
wilaw: Fixed with https://github.com/moq-wg/moq-transport/pull/1638/changes/bd4b664066077870621940a0e514ae8782a49040

draft-ietf-moq-transport.md
sharmafb: I like your idea of changing the activate flag from binary to a number, rather than making probabilistic statements.

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2113,99 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
...
+  set.allocated = min(set.target, B_remaining)
+  set.selected = track in set with highest throughput where track.throughput <= set.allocated
+  B_remaining -= set.selected.throughput
```
acbegen: In 2198 and 2199, it would be clearer to write track.thoughput-threshold as that is the info coming from the Switching-Set-Assignment.

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2113,99 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
 combine, split, or otherwise modify object payloads.  A relay SHOULD
 prioritize sending Objects based on {{priorities}}.
 
+## Dynamic Track Switching {#dynamic-track-switching}
```
acbegen: I feel like "Relay-Side Track Switching" is a more appropriate term here. If "Dynamic" is important, we can also keep it but adding "relay-side" is critical, IMO.
If that is changed, then we could consider renaming SWITCHING-SET-ASSIGNMENT to RS-SWITCHING-SET-ASSIGNMENT.

draft-ietf-moq-transport.md
wilaw: Changing Activate to a Number was implemented in https://github.com/moq-wg/moq-transport/pull/1638/changes/ed3164105d551cc28e1f98422d54f7fae9f6c23e

draft-ietf-moq-transport.md
wilaw: Note that DTS can be implemented by any publisher, not only Relays. Would you prefer Publisher Track Switching?

acbegen: Are you saying an original publisher could implement this? I don't think it makes sense.

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2113,103 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
...
+switching set.
+
+### Subscriber operations
+
+To modify an established switching set, the subscriber can
+
+* Add a track to an existing set: send SUBSCRIBE or PUBLISH_OK with a SWITCHING-SET-ASSIGNMENT parameter
+  referencing an existing set.
+* Remove a track from a set: unsubscribe from that track.
+* Pause DTS: Send REQUEST_UPDATE for any track assigned to that set with a SWITCHING-SET-ASSIGNMENT
```
yuyou: Based on the discussion in [issue #11](https://github.com/wilaw/dts4moq/issues/11), sending a REQUEST_UPDATE to an individual subscription appears to be semantically more appropriate, as it effectively modifies the state of a specific subscription within a switching set rather than affecting the entire set.

With this approach, tracks can be enabled or disabled individually without re-subscribing. 

> Remove a track from a set: unsubscribe from that track.

Current re-subscription design may introduce object delivery synchronization challenges and could delay track activation until the next group boundary.

> Resume DTS: 

A Set becomens activate when there's at least one activated track.

> Pause a DTS:

A Set becomes deactivated when all tracks are deactivated.  

It might be beneficial to consider supporting this fine-grained level of activation control in the design.

gwendalsimon:
In the relay behavior section, all non-selected switching-set tracks have fwd=0 downstream. Per §9.2 (discussed in #1650), relays SHOULD set the Forward flag to 1 upstream regardless of the downstream value. However, this PR says nothing about upstream Forward State for non-selected tracks. The selection algorithm picks the best fitting track based on its configured throughput threshold _with no guard on upstream data availability_. What if a newly selected track has no upstream data flowing?

One option: "_When the algorithm selects a track for which upstream data is not yet available, the relay MUST forward using the highest-throughput available track and activate the upstream subscription for the selected track, switching at the next group boundary once data flows._" (note that this assumes the upstream subscription can be established before the next group boundary, which is not guaranteed with short groups and high upstream RTT)

Alternatively, this PR could require the relay to maintain fwd=1 upstream for all switching-set tracks, and rely on MAX_DTS_CONCURRENT_TRACKS to bound the DoS exposure.

gwendalsimon:
I have some (D)DoS concerns with the PR. I have two independent (but related) comments.

`MAX_DTS_CONCURRENT_TRACKS` bounds the number of tracks in a switching set, but not their aggregate throughput. A subscriber can assign a full ABR ladder to a single switching set. For example, Apple's HLS HEVC authoring spec defines 12 tiers from 145 kbps to 16.8 Mbps (~54 Mbps total). Per §9.2, the relay maintains upstream fwd=1 for all 12 tracks while the algorithm selects only the lowest tier downstream. A malicious subscriber throttling its network to minimum receives 145 kbps while the relay ingests 54 Mbps. Ingest-to-egress ratio: 372×

There is also a cache cost that is easy to overlook: the relay must buffer at least one full group per track to switch at group boundaries. At a 2-second group duration, the 12-tier ladder requires ~14 MB of cache per session, independent of what is actually delivered downstream.

As discussed in my earlier comment (https://github.com/moq-wg/moq-transport/pull/1638#issuecomment-4647283499), the spec does not require the relay to maintain upstream fwd=1 for all switching-set tracks. A relay aware of the subscriber's downstream bandwidth could limit upstream fwd=1 to tracks at or near that bandwidth. For a client consistently at 145 kbps, the top tiers could stay at fwd=0. However, this is not robust against malicious subscribers who continuously vary their apparent bandwidth.

gwendalsimon:
This second (D)DoS concern is related to but independent from my previous comment (https://github.com/moq-wg/moq-transport/pull/1638#issuecomment-4647768418).

`MAX_DTS_CONCURRENT_TRACKS` is scoped per session (one QUIC connection or one WebTransport session). Multiple WebTransport sessions can be established over a single QUIC connection, and the MOQT spec sets no limit on how many a client may open.

Consider a relay configured with `MAX_DTS_CONCURRENT_TRACKS` = 3, which is the most conservative possible setting, allowing only a minimal +1/-1 switching set. A single attacker opens 10,000 WebTransport sessions over one QUIC connection. With a 100 Mbps home connection, the per-session bandwidth estimate drops to ~10 kbps. Since it is below the lowest track in the ladder, the algorithm selects nothing and delivers zero bits downstream. The attacker's cost is only the one-time control-plane burst to establish sessions. Sustained bandwidth is negligible. Yet the relay must maintain state for 30,000 subscriptions and execute the allocation algorithm 10,000 times per group boundary. If each session targets a different channel, subscription aggregation does not apply and the relay must cache independently per session. For the three top Apple HLS HEVC tiers, we are talking around 100 GB of live group cache at the relay. If the Relay is smarter with its upstream `fwd` flag management, it still has computation cost and state management.

Should the PR consider whether DTS limits belong at the client level rather than the session level? Multiple WebTransport sessions over the same QUIC connection are identifiable as the same source. It would limit the risk of DoS. It would not fix potential DDoS though.

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2113,103 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
...
+
+On a periodic update interval or at a minimum of a new Group boundary, the relay executes the
+following algorithm:
+
+~~~
+B_remaining  = B_total
+for each set in ascending rank order:
+  set.target = B_remaining × set.fraction / sum_F
+  set.allocated = min(set.target, B_remaining)
+  set.selected = track in set with highest throughput where track.throughput <= set.allocated
```
wilaw:
```suggestion
  set.selected = track in set with highest throughput where track.throughput-threshold <= set.allocated
```

wilaw:
>  Multiple WebTransport sessions can be established over a single QUIC connection, and the MOQT spec sets no limit on how many a client may open.

@gwendalsimon  - is this a DTS issue or a MOQT issue? I think its more the latter. I agree with the DDOS risk you describe. If the scare resource on the server is bandwidth more than connections, then we should probably consider replacing MAX_DTS_CONCURRENT_TRACKS with MAX_DTS_CONCURRENT_THROUGHPUT, or else have noth of those together as negotiated settings. 


>  A single attacker opens 10,000 WebTransport sessions over one QUIC connection. With a 100 Mbps home connection, the per-session bandwidth estimate drops to ~10 kbps. Since it is below the lowest track in the ladder, the algorithm selects nothing and delivers zero bits downstream. The attacker's cost is only the one-time control-plane burst to establish sessions. Sustained bandwidth is negligible.

A similar risk situation occurs with Http/3 requests. However, both endpoints negotiate a MAX_STREAMS limit during the QUIC handshake, and HTTP/3 layers its own SETTINGS_MAX_CONCURRENT_STREAMS on top of that. A typical server default is 100–1000 concurrent streams. We could apply something similar for MOQT?

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2113,103 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
 combine, split, or otherwise modify object payloads.  A relay SHOULD
 prioritize sending Objects based on {{priorities}}.
 
+## Dynamic Track Switching {#dynamic-track-switching}
+
+Dynamic Track Switching (DTS) is a subscriber-initiated and controlled behavior in which
+relays dynamically select which track to forward from a switching set based on
+available downstream bandwidth.
```
ArcherDTY: If relay can get the information/status of subscribers, I think it's also possible to select tracks based on `available downstream bandwidth` + `subscriber buffer length`. So, could this description of 'based on
available downstream bandwidth' cause unnecessary restrictions?

draft-ietf-moq-transport.md
```diff
@@ -2605,6 +2702,54 @@ Namespace Prefix for that subscription.  If the new prefix would share a common
...
+
+* Switching set ID: Integer identifying the switching set. A track MUST only be assigned
+  to one switching set at a time. If a subscription attempts to assign a track that is
+  already assigned to a different switching set, the relay MUST reject the subscription
+  with a Parameter Error. If a subscription attempts to assign a track, when the number of
+  tracks already assigned to switching sets within the session is
+  >= MAX_DTS_CONCURRENT_TRACKS {{max-dts-concurrent-tracks}}, then the relay MUST reject
+  the subscription with a Parameter Error.
+
+* Throughput threshold: Minimum throughput (kbps) required to select this track.
```
ArcherDTY: This parameter has a decisive impact on the sender side ABR. Will it be decided by the subscriber or the relay itself? 
If it's subscriber decided, it will be confusing to me. I feel this parameter is more suitable as a part of ABR algorithm. If a subscriber directly specifies it, it would limit the ABR designs. 
If this `throughput threshold` actually represents the encoding bitrate of the corresponding track, it might be more appropriate.

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2113,103 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
...
+considered deleted and is removed from the bandwidth allocation algorithm.
+
+### Bandwidth Allocation {#allocation-algorithm}
+
+The relay maintains:
+
+- `B_total`: Estimated downstream bandwidth capacity for the subscriber connection. The
+  relay MUST maintain a bandwidth estimate for each downstream subscriber. The timebase of this
+  estimate SHOULD be at least the Group duration of the track. The estimate is obtained
+  periodically from the QUIC stack (e.g., congestion window pacing rate, smoothed RTT)
```
ArcherDTY: If the algorithm 0 is mandatory to implement, do you need to specify a bandwidth estimation algorithm here?

draft-ietf-moq-transport.md
```diff
@@ -5131,6 +5313,19 @@ This document does not define any initial entries.
...
 
+### SSTS-Algorithms {#iana-ssts-algorithms}
+
+This document establishes a registry for SSTS algorithms. The
+registration policy is Specification Required (per {{!RFC8126,
+Section 4.6}}).
+
+| Type | Name       | Specification |
+|-----:|:-----------|:--------------|
+| 0x0  | Default  | {{authorization-token}} |
```
wilaw:
```suggestion
| 0x0  | Default  | this |
```

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2113,156 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
...
+## Sender Side Track Switching {#sender-side-track-switching}
+
+Sender Side Track Switching (SSTS) is a subscriber-initiated and controlled behavior in which
+relays dynamically select which track to forward from a switching set based on
+various algorithms. Each algorithm defines a set of attributes which are passed in the
+SWITCHING-SET-ASSIGNMENT parameter {{switching-set-assignment-param}} along with a
+complimentary set of rules for subscriber behavior and relay behavior.
+
+This specification defines a default algorithm - type 0. Other algorithms are referenced in the
+"MOQT SSTS Algorithms" registry {{iana-ssts-algorithms}}.
```
wilaw:
```suggestion
"SSTS Algorithms" registry {{iana-ssts-algorithms}}.
```

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2113,156 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
...
+This algorithm extends the base definition of the SWITCHING-SET-ASSIGNMENT parameter
+{{switching-set-assignment-param}} to add the following fields:
+
+SWITCHING-SET-ASSIGNMENT {
+  Switching set ID (vi64),
+  Algorithm ID (vi64),
+  Throughput threshold (vi64),
+  Set throughput fraction (vi64),
+  Activate switching (vi64),
+  [Set rank (8)]
```
wilaw:
```suggestion
  Set rank (8)
```

draft-ietf-moq-transport.md
```diff
@@ -2113,6 +2113,156 @@ A relay MUST treat the object payload as opaque.  A relay MUST NOT
...
+  respectively. This allows sets to be added or removed without requiring other sets to
+  update their fractions. When multiple subscriptions in the same switching set specify
+  different fraction values, the relay MUST use the value from the most recently received
+  message for that set.
+
+* Activate switching: Integer, when set to 0, pauses SSTS switching for this set. When set
+  to N, the relay activates or resumes switching as soon as the number of tracks assigned to
+  the switching set is >= N.  Activation takes effect at the next group boundary.
+
+* Set rank (optional): Degradation priority when bandwidth is constrained, expressed as
```
wilaw:
```suggestion
* Set rank: Degradation priority when bandwidth is constrained, expressed as
```

draft-ietf-moq-transport.md
wilaw: Commit https://github.com/moq-wg/moq-transport/pull/1638/changes/70237e6c860f07f8d40fe1a07cdbfea52dc50994 changed name to Sender Side Track Switching, per London interim decision

draft-ietf-moq-transport.md
wilaw: We raised this issue at the London interim, and the room consensus was to leave the current design, which requires a unsubscribe to remove a track from a switching set. The use-case of continuing to subscribe to a track after removing it from a switching set was not evident.

At the same meeting we agree to make the SSTS extensible over different algorithms, so the good newsis that  in the future it will be possible to design a different version of SSTS which does use the REQUEST_UPDATE to remove a track from a set, if that proves to be the most expedient solution.

draft-ietf-moq-transport.md
wilaw: Agreed. Commit https://github.com/moq-wg/moq-transport/pull/1638/changes/70237e6c860f07f8d40fe1a07cdbfea52dc50994 changed the language to 

> Sender Side Track Switching (SSTS) is a subscriber-initiated and controlled behavior in which relays dynamically select which track to forward from a switching set based on various algorithms.

draft-ietf-moq-transport.md
wilaw: This is subscriber decided value, because the subscriber has the knowledge of the bitrate of the track (say 1Mbps) and the throughput level it is comfortable switching to that track (perhaps 1.5Mbps). The relay knows neither of these values. 

If throughput threshold was communicated as the encoded bitrate of the track, then either different relays would choose different safety margins, or else the safety margin would need to be communicated as an additional parameter, which is less efficient than communicating the product of the two which is what we send now. Different clients will want different safety margins at different times (typically the closer you are to the live edge, the higher your margin), so baking a fixed value into the relay would not be optimal.

draft-ietf-moq-transport.md
wilaw: In prior meetings we debated the efficacy of specifying one bandwidth estimation algorithm that all relays must follow. Bandwidth estimation is a black art and is dependent on congestion control algorithms, competing traffic, MTU, application demand etc. The design team could not agree on a single algorithm which would work across all future implementations. We therefore decided to leave it in its (intentionally) loosely-defined state for algorithm 0, which is that the relay should make it best effort to estimate the bandwidth available. This will have the consequence that different relay implementations will switch at slightly different times. We considered that acceptable for the default algorithm. if implementation experience proves that the variance between relays in untenable, then we can design algorithm 1 which would implement a fixed-estimation scheme.

afrind:
I've reviewed the latest text here: https://docs.google.com/document/d/1S2WlpVFzXICjYFDDd7rI5yHpc_Zhp_aWnXgUvXixTy4/edit?tab=t.0

You will need to request access.

Since it's more than 30 comments, I figured you didn't want me to leave it as PR comments.

yuyou:
It would be beneficial to define supplemental algorithms compatible with the implementation, incorporating metrics beyond throughput. Potential metrics for the switching algorithm include throughput, ECN for L4S (CE marks), Round-Trip Time (RTT), buffer occupancy, and queuing delay.

draft-ietf-moq-transport.md
ArcherDTY: Sounds good.

wilaw:
> I've reviewed the latest text here: https://docs.google.com/document/d/1S2WlpVFzXICjYFDDd7rI5yHpc_Zhp_aWnXgUvXixTy4/edit?tab=t.0
> 
@afrind All issues listed there addressed with commit https://github.com/moq-wg/moq-transport/pull/1638/changes/c1ddc56e68ec3c71bd57b79100793b5e6295e53f

draft-ietf-moq-transport.md
```diff
@@ -2296,6 +2296,162 @@ combine, split, or otherwise modify object payloads.
...
+On a periodic update interval or at a minimum when an object is received/published on a group
+larger than previously largest group, the relay executes the following algorithm:
+
+~~~
+B_remaining  = B_total
+for each set in ascending set.rank order:
+  set.target = B_remaining × set.weight / sum_W
+  set.allocated = min(set.target, B_remaining)
+  set.selected = track in set with highest throughput_threshold where track.throughput_threshold <= set.allocated
+  B_remaining -= set.selected.throughput_threshold
```
ArcherDTY, line 2445: This part of the code looks kind of weird.
Let's think about the logic after the first loop: `B_remaining` represents the remaining available bandwidth after removing the throughput of the selected set. When the next loop starts, `sum_W` doesn't take out the weight of the selected set. It means that, since the second loop, `set.weight / sum_W` no longer reflects the proportion of that set's weight among all remaining sets.
Maybe adding another line `sum_W -= set.weight` could solve the problem? Or adding a similar `sum_W_remaining` variable might also work.

wilaw: @ArcherDTY  - this PR has been refactored into an [external draft ](https://wilaw.github.io/moqt-ssts/draft-wilaw-moq-moqt-ssts.html)which implements SSTS as a MOQT extension. The draft contains an updated version of this algorithm, shown below, which I think addresses your concern.  Please review and add any comments to the [issues ](https://github.com/wilaw/moqt-ssts/issues)associated with that draft. 

```
B_remaining = B_total
for each distinct set.rank value r present among active switching sets, in ascending order:
  active = { switching sets with set.rank == r }
  tier_pool = B_remaining
  repeat:
    sum_W_active = sum of set.weight for set in active
    for each set in active:
      set.target = tier_pool × set.weight / sum_W_active
      set.selected = null
      if ( set.target >= lowest track.throughput_threshold in set )
          set.selected = track in set with highest throughput_threshold
                         where track.throughput_threshold <= set.target
    saturated = { set in active : set.selected == highest-throughput_threshold
                                   track available in set }
    for each set in saturated:
      tier_pool -= (set.selected != null) ? set.selected.throughput_threshold : 0
      remove set from active
  until ( active is empty OR saturated is empty )
  for each set with set.rank == r:
    B_remaining -= (set.selected != null) ? set.selected.throughput_threshold : 0
for each track in a switching set:
  set forward state = (track == set.selected)
```

wilaw:
Based on feedback received during presentation at IETF #126, this PR has been refactored into an [external draft ](https://wilaw.github.io/moqt-ssts/draft-wilaw-moq-moqt-ssts.html)which implements SSTS as a MOQT extension.  Please review and add any comments to the [issues ](https://github.com/wilaw/moqt-ssts/issues)associated with that draft. 

Closing this PR for now. We can reopen in the future if decide a MOQT extension is not the best approach.

PR #1640, opened 2026-06-01
Add EXPIRES to most request types

### LINKED ISSUE #1639: EXPIRES should be allowed in more REQUEST_OK sub-types.
IIUC EXPIRES is basically saying "this is how long your auth token is good for." If that's the case, this parameter is useful for any response for a request that contains an auth token, which is all of them.

I can squint and say that FETCH_OK is an exception, because all the objects are in the past and therefore the token can't expire. That's a design question I don't have the answer to.

Similarly, TRACK_STATUS is ~ instantaneous so maybe EXPIRES doesn't make sense.

The EXPIRES text also mentions PUBLISH_DONE, which is not applicable if the parameter is applicable to a wide range of request types, so that language would need cleanup.

martinduke:
Fixes #1639

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2510,7 +2510,8 @@ unfiltered.  If omitted from REQUEST_UPDATE, the value is unchanged.
 ### EXPIRES Parameter {#expires}
 
 The EXPIRES parameter (Parameter Type 0x8) is a varint. It MAY appear in
-SUBSCRIBE_OK, PUBLISH, PUBLISH_OK, or REQUEST_UPDATE_OK. It encodes the time
+SUBSCRIBE_OK, PUBLISH, PUBLISH_OK, SUBSCRIBE_NAMESPACE_OK, SUBSCRIBE_TRACKS_OK,
+PUBLISH_NAMESPACE_OK, or REQUEST_UPDATE_OK. It encodes the time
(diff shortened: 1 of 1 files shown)


afrind (approved):

draft-ietf-moq-transport.md
```diff
@@ -2510,7 +2510,8 @@ unfiltered.  If omitted from REQUEST_UPDATE, the value is unchanged.
 ### EXPIRES Parameter {#expires}
 
 The EXPIRES parameter (Parameter Type 0x8) is a varint. It MAY appear in
-SUBSCRIBE_OK, PUBLISH, PUBLISH_OK, or REQUEST_UPDATE_OK. It encodes the time
+SUBSCRIBE_OK, PUBLISH, PUBLISH_OK, SUBSCRIBE_NAMESPACE_OK, SUBSCRIBE_TRACKS_OK,
```
ianswett, line 2513: It's a bit odd that PUBLISH is the only case where it can appear in either the message or the OK.  I'm not sure its a problem, just odd.

ianswett (approved):

draft-ietf-moq-transport.md
afrind: The publisher is the one that can define the lifetime of a subscription, and the relay is the one who knows auth token lifetimes.

PR #1642, opened 2026-06-02
Replace Joining Fetch with Subscription Fill, add Current Group

### LINKED ISSUE #1039: Simplifying joining at the latest available join point.

### LINKED ISSUE #1313: Joining FETCH as a separate control message creates edge cases and feature gaps

### LINKED ISSUE #1358: Subscribing to the start of the current Group could be optimized

### LINKED ISSUE #1614: (JOINING) FETCH + SUBSCRIBE prioritization
Making a separate issue instead of hijacking #1358

When starting a subscription for a real-time use-case, the goal is to always deliver the most recent GoP and cancel/deprioritize any older GoPs.

We do this via `SUBSCRIBE order=desc`, and of course we need to issue two requests to start the subscription: 
```
SUBSCRIBE track=video order=desc filter=LatestObject
JOINING FETCH track=video
```

However, as it stands in the current draft, it's implementation specific how to prioritize these two requests. Ideally, we want the `JOINING FETCH` first, otherwise we're wasting bandwidth/RAM on undecodable frames from `SUBSCRIBE`, hurting TTV. We could explicitly request this:

```
SUBSCRIBE track=video order=desc filter=LatestObject priority=5
JOINING FETCH track=video priority=6
```

However, this doesn't work properly when NextGroup starts. The higher priority `JOINING FETCH` takes precedence over `order=DESC`. That means we're stuck downloading all of the old segment up until the original `LargestObject` before we can skip to the next group.

---

Here's a concrete example. The largest object is 1.5s into a 2s GoP (group 6), so it's almost done. Let's say the the media is encoded at 3Mb/s and our network speed is 4.5Mb/s.

It takes 1s to download the higher priority JOINING FETCH (1.5s of media). However, at this point the live playhead is already 0.5s into the next group (group 7). Now SUBSCRIBE is the highest priority request and skips to the next group, delivering the first 0.5s (in 0.33s) and then delivering the next 0.33s that was generated in the meantime (in 0.22s or whatever), etc. IDK of the exact math but it would take something like 1.5s to catch up to the live playhead.

But if we did a LargestGroup SUBSCRIBE, it would only take 0.5s to catch up to the live playhead. We would spend the first 0.5s downloading the largest group (6), but the instant there's a new group, the SUBSCRIBE immediately prioritizes the next group (order=desc) instead.

---

Basically, we need `order=DESC` support for JOINING FETCH. Either some way of prioritizing between the SUBSCRIBE + JOINING FETCH, or cancelling the JOINING FETCH if the next group starts (kinda gross), or add back the LargestGroup filter (pls).

afrind:
Remove Joining Fetch and instead let SUBSCRIBE specify fill filter types that retrieve past objects.
The publisher opens a fill fetch stream (FETCH_HEADER with the subscription's Request ID) for objects
before the current group, while delivering current and future objects via subscribe subgroups and
datagrams.

Changes:
- Remove Joining Fetch types, Joining Fetches section, Joining Fetch Range Calculation, Joining
   Location, and INVALID_JOINING_REQUEST_ID error code
- Inline Standalone Fetch fields directly into the FETCH message wire format
- Add new subscription filter types: AbsoluteStartFill (0x5), AbsoluteRangeFill (0x6), CurrentGroup
  (0x7), RelativeStartFill (0x8)
- Add Current Group Delivery section with publisher subgroup ordering and relay object accounting
  requirements, applicable to any filter covering the current group
- Add Fill Semantics section defining fill fetch stream behavior
- Add FillDescending (0x3) group order: descending for fill, ascending for subscribe
- Update LARGEST_OBJECT to serve as the fill boundary in SUBSCRIBE_OK
- Extend FILL_TIMEOUT to apply in SUBSCRIBE messages with fill filter types
- Add fill fetch stream cancellation semantics and relay guidance
- REQUEST_UPDATE with a fill filter type opens a new fill fetch stream using the REQUEST_UPDATE's
  Request ID
- Disallow fill filter types and CurrentGroup in PUBLISH_OK (stale LARGEST_OBJECT); recommend forward=0
   + REQUEST_UPDATE pattern instead
- Clarify PUBLISH forward=1 starts delivery at Largest Object

TODO: Resolve AbsoluteRangeFill behavior when end group is before current group.

Fixes #1039
Fixes #1313 
Fixes #1358
Fixes #1614

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1139,6 +1139,9 @@ EXPIRED_AUTH_TOKEN (0x7):
 EXCESSIVE_LOAD (0x9):
 : The endpoint is overloaded and is resetting this stream.
 
+CURRENT_GROUP_UNAVAILABLE (0xA):
+: The publisher is unable to complete the subscriber's request for the current group.
+
 MALFORMED_TRACK (0x12):
 : A relay publisher detected that the track was malformed (see
   {{malformed-tracks}}).
@@ -5162,6 +5165,7 @@ This document does not define any initial entries.
 | UNKNOWN_OBJECT_STATUS | 0x6  | {{stream-reset-codes}}   |
 | EXPIRED_AUTH_TOKEN    | 0x7  | {{stream-reset-codes}}   |
 | EXCESSIVE_LOAD        | 0x9  | {{stream-reset-codes}}   |
+| CURRENT_GROUP_UNAVAILABLE | 0xA | {{stream-reset-codes}} |
 | MALFORMED_TRACK       | 0x12 | {{stream-reset-codes}}   |
 | Reserved for greasing | 0x7f * N + 0x9D | {{grease}} |
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1577,7 +1577,7 @@ and forward delivery.
 The fill range cannot be expanded via REQUEST_UPDATE. To retrieve additional
 past objects, use a standalone FETCH.
 
-FILL_TIMEOUT (see {{delivery-timeout}}) applies to fill fetch streams in the
+FILL_TIMEOUT (see {{delivery-timeouts}}) applies to fill fetch streams in the
 same way it applies to standalone fetches.
 
 ### Joining an Ongoing Track
@@ -2498,7 +2498,9 @@ for the same track.
 
 ### FILL TIMEOUT Parameter {#fill-timeout}
 
-The FILL_TIMEOUT parameter (Parameter Type 0x0A) MAY appear in a FETCH message.
+The FILL_TIMEOUT parameter (Parameter Type 0x0A) MAY appear in a FETCH or
+SUBSCRIBE message. When present in a SUBSCRIBE with a fill filter type, it
+applies to the fill fetch stream.
 
 It is the maximum total duration in milliseconds a relay SHOULD spend waiting
 for upstream sources to provide Objects that are not immediately available
@@ -2563,7 +2565,7 @@ SUBSCRIBE, PUBLISH_OK, or FETCH.
 
 Its value indicates how to prioritize Objects from different groups within
 the same subscription (see {{priorities}}), or how to order Groups in a Fetch
-response (see {{fetch-handling}}). The allowed values are Ascending (0x1),
+response (see {{message-fetch}}). The allowed values are Ascending (0x1),
 Descending (0x2), or Inside Out (0x3). Inside Out orders fill fetch stream
 objects in descending order and subscribe stream objects in ascending order;
 when used with a subscription that has no fill, it degenerates to Ascending;
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1528,27 +1528,29 @@ close the session with a `PROTOCOL_VIOLATION`.
 AbsoluteStartFill (0x5): The filter Start Location is specified explicitly.
 Objects from Start Location up to but not including `{Largest Object.Group, 0}`
 are delivered on a fill fetch stream (see {{fill-semantics}}). Objects from
-`{Largest Object.Group, 0}` onward are delivered on subscribe streams. There is
-no End Group - the subscription is open ended.
+`{Largest Object.Group, 0}` onward are delivered using subscribe subgroups and
+datagrams. There is no End Group - the subscription is open ended.
 
 AbsoluteRangeFill (0x6): The filter Start Location and End Group Delta are
 specified explicitly. Objects from Start Location up to but not including
 `{Largest Object.Group, 0}` are delivered on a fill fetch stream (see
 {{fill-semantics}}). Objects from `{Largest Object.Group, 0}` onward are
-delivered using subscribe subgroups and datagrams. If the End Group is before `Largest Object.Group`,
-only fill delivery occurs.
+delivered using subscribe subgroups and datagrams. If the End Group is before
+`Largest Object.Group`, only fill delivery occurs.
 TODO: Determine behavior when end is before current group — does the
 subscription auto-close, or remain open for possible REQUEST_UPDATE?
 
-RelativeStart (0x7): A single varint N specifying how many groups before
-`Largest Object.Group` to start. The Start Location is
-`{Largest Object.Group - N, 0}`. When N > 0, objects from Start Location up to
-but not including `{Largest Object.Group, 0}` are delivered on a fill fetch
-stream (see {{fill-semantics}}), and objects from `{Largest Object.Group, 0}`
-onward are delivered using subscribe subgroups and datagrams. When N = 0, no fill fetch stream is
-opened; cached objects from `{Largest Object.Group, 0}` are delivered on
-subscribe streams, transitioning to live delivery on the same streams.
+CurrentGroup (0x7): The filter Start Location is `{Largest Object.Group, 0}`.
+If no content has been delivered yet, the filter Start Location is {0, 0}.
 There is no End Group - the subscription is open ended.
+
+RelativeStartFill (0x8): A single varint N. The Start Location is
+`{Largest Object.Group - (N + 1), 0}`. Objects from Start Location up to but
+not including `{Largest Object.Group, 0}` are delivered on a fill fetch stream
+(see {{fill-semantics}}), and objects from `{Largest Object.Group, 0}` onward
+are delivered using subscribe subgroups and datagrams. If N + 1 is greater than
+Largest Object.Group, the Start Location is `{0, 0}`. There is no End Group -
+the subscription is open ended.
 
 An endpoint that receives a filter type other than the above MUST close the
 session with `PROTOCOL_VIOLATION`.
@@ -1558,24 +1560,54 @@ If the publisher cannot satisfy the requested Subscription Filter (see
 it SHOULD send a REQUEST_ERROR with code `INVALID_RANGE`.  A publisher MUST
 NOT send objects from outside the requested range.
 
+### Current Group Delivery {#current-group-delivery}
+
+When a subscription's filter includes the current group — as with
+CurrentGroup, RelativeStartFill, AbsoluteStartFill with a Start Location at
+or before the current group, or AbsoluteRange that crosses the current group
+— the publisher delivers both cached and newly published Objects for that
+group via subscribe subgroups and datagrams. For subsequent Groups, Objects
+are delivered normally as they are published.
+
+The publisher MUST begin each subgroup with the lowest Object ID published by
+the Original Publisher in that subgroup. If the publisher is unable to deliver
+any subgroups or datagrams for the current Group (e.g., an empty cache and
+the Original Publisher no longer has the data), the publisher SHOULD send a
+REQUEST_ERROR with code `INVALID_RANGE`.
+
+Relays MUST obtain all Objects in the current Group that are not already
+cached. A relay MUST NOT deliver an Object in a subgroup unless all prior
+Objects in that subgroup are accounted for — either already received,
+confirmed non-existent via gap properties, or fetched from upstream. An
+existing upstream subscription that started before the current Group will
+typically deliver these Objects without additional action, though temporary
+gaps can exist until they arrive. A relay without an active upstream
+subscription can issue one with CurrentGroup; otherwise it uses FETCH.
+
 ### Fill Semantics {#fill-semantics}
 
-Fill filter types (AbsoluteStartFill, AbsoluteRangeFill, and RelativeStart
-with N > 0) cause the publisher to open a unidirectional stream with a
+Fill filter types (AbsoluteStartFill, AbsoluteRangeFill, and
+RelativeStartFill) cause the publisher to open a unidirectional stream with a
 FETCH_HEADER (see {{fetch-header}}) using the subscription's Request ID to
 deliver objects before the current group. This is called a fill fetch stream.
 
 The fill fetch stream carries objects from the fill Start Location up to but
 not including `{Largest Object.Group, 0}`, where `Largest Object` is the value
-communicated in SUBSCRIBE_OK. Subscribe streams carry objects from
-`{Largest Object.Group, 0}` onward.
+communicated in SUBSCRIBE_OK or REQUEST_UPDATE_OK. For RelativeStartFill, both
+sides compute the fill Start Location as `{Largest Object.Group - (N + 1), 0}`
+using the LARGEST_OBJECT from the response. Subscribe subgroups and datagrams
+carry objects from `{Largest Object.Group, 0}` onward.
 
-The fill fetch stream and subscription share the same Request ID, subscriber
-priority, and authorization. Setting Forward State to 0 suppresses both fill
-and forward delivery.
+The fill fetch stream and subscription share all subscription parameters,
+including Request ID, subscriber priority, and authorization. Setting Forward
+State to 0 implicitly cancels the fill
+fetch stream and suppresses forward delivery. The publisher MUST reset any
+open fill fetch streams when Forward State transitions to 0.
 
-The fill range cannot be expanded via REQUEST_UPDATE. To retrieve additional
-past objects, use a standalone FETCH.
+A REQUEST_UPDATE containing a SUBSCRIPTION_FILTER with a fill filter type
+opens a new fill fetch stream. The fill fetch stream uses the Request ID from
+the REQUEST_UPDATE. The REQUEST_UPDATE_OK will include LARGEST_OBJECT
+establishing the fill boundary for the new fill fetch stream.
 
 FILL_TIMEOUT (see {{delivery-timeouts}}) applies to fill fetch streams in the
 same way it applies to standalone fetches.
@@ -1588,13 +1620,16 @@ request an existing Group or wait for a future Group.  Different applications
 will have different approaches for when to begin a new Group.
 
 To join a Track at a past Group, the subscriber sends a SUBSCRIBE with a fill
-filter type: AbsoluteStartFill, AbsoluteRangeFill, or RelativeStart with N > 0.
+filter type: AbsoluteStartFill, AbsoluteRangeFill, or RelativeStartFill.
+The publisher responds with SUBSCRIBE_OK including LARGEST_OBJECT.
+Largest Object.Group indicates the boundary between fill and live delivery. If
+Objects prior to the current group are requested, the publisher opens a fill
+fetch stream for objects before `{Largest Object.Group, 0}` and delivers
+current and future objects using subscribe subgroups and datagrams (see
+{{fill-semantics}}).
+
 To join a Track at the current Group, the subscriber sends a SUBSCRIBE with
-filter type RelativeStart and N = 0.
-The publisher responds with SUBSCRIBE_OK including LARGEST_OBJECT. Largest Object.Group
-indicates the boundary between fill and live delivery. The publisher opens a fill fetch
-stream for objects before `{Largest Object.Group, 0}` and delivers current and
-future objects using subscribe subgroups and datagrams (see {{fill-semantics}}).
+Filter Type `CurrentGroup`.
 
 To join a Track at the next Group, the subscriber sends a SUBSCRIBE with
 Filter Type `Next Group Start`.
@@ -1604,7 +1639,7 @@ Filter Type `Next Group Start`.
 While some publishers will deterministically create new Groups, other
 applications might want to only begin a new Group when needed.  A subscriber
 joining a Track might detect that it is more efficient to request the Original
-Publisher create a new group than to subscribe using RelativeStart N = 0.  Publishers indicate a
+Publisher create a new group than to subscribe using CurrentGroup.  Publishers indicate a
 Track supports dynamic group creation using the DYNAMIC_GROUPS parameter
 ({{dynamic-groups}}).
 
@@ -1652,10 +1687,10 @@ sending STOP_SENDING on the fill fetch stream. The subscription continues
 to deliver objects using subscribe subgroups and datagrams.
 
 Cancelling the subscription (STOP_SENDING on the bidi stream) cancels both
-the fill fetch stream and forward delivery.  The publish MUST reset any open
+the fill fetch stream and forward delivery.  The publisher MUST reset any open
 fill fetch streams associated with a cancelled subscription.
 
-The fill fetch stream is closed with a FIN when all past objects up to the
+The fill fetch stream is closed with a FIN after all past objects up to the
 fill boundary (LARGEST_OBJECT from SUBSCRIBE_OK) have been delivered.
 
 # Namespace Discovery {#track-discovery}
@@ -1798,7 +1833,7 @@ datagram (see {{data-streams}}).
 
 `Group Order` is a property of an individual subscription.  It can be
 'Ascending' (groups with lower group ID are sent first), 'Descending'
-(groups with higher group ID are sent first), or 'Inside Out' (fill
+(groups with higher group ID are sent first), or 'FillDescending' (fill
 fetch stream objects are sent in descending order while subscribe stream
 objects are sent in ascending order).  The subscriber optionally
 communicates its group order preference in the SUBSCRIBE message; the
@@ -1819,7 +1854,7 @@ the objects SHOULD be selected as follows:
 2. If two objects in response to the same request have the same subscriber
    and publisher priority, but belong to two different groups of the same track,
    **the group order** of the associated subscription is used to
-   decide the one that is scheduled to be sent first. For Inside Out group
+   decide the one that is scheduled to be sent first. For FillDescending group
    order, subscribe-delivered objects are preferred over fill-delivered objects;
    within subscribe objects, ascending group order is used; within fill objects,
    descending group order is used.
@@ -2120,11 +2155,10 @@ When a Relay needs to make an upstream FETCH request, it determines the
 available publishers using the same matching rules as SUBSCRIBE. When more than
 one publisher is available, the Relay MAY send the FETCH to any of them.
 
-When a Relay receives a downstream SUBSCRIBE with a fill filter type, it MAY
-serve the fill portion from its cache. If the cache does not contain the
-requested objects, the Relay MAY issue upstream standalone FETCHes to retrieve
-them. The fill fetch stream opened to the downstream subscriber uses the
-subscription's Request ID.
+When a Relay receives a downstream SUBSCRIBE with a fill filter type or
+CurrentGroup, it serves the fill or current group portion from its cache and
+retrieves any missing objects upstream using SUBSCRIBE or standalone FETCHes
+(see {{current-group-delivery}} and {{fill-semantics}}).
 
 When a Relay receives an authorized SUBSCRIBE for a Track with one or more
 `Established` upstream subscriptions, it MUST reply with SUBSCRIBE_OK.  If the
@@ -2502,9 +2536,9 @@ for the same track.
 ### FILL TIMEOUT Parameter {#fill-timeout}
 
 The FILL_TIMEOUT parameter (Parameter Type 0x0A) MAY appear in a FETCH,
-SUBSCRIBE, PUBLISH_OK or REQUEST_UPDATE (for a SUBSCRIBE or PUBLISH) message.
-When present in a non-FETCH request with a fill filter type, it
-applies to the fill fetch stream.
+SUBSCRIBE, or REQUEST_UPDATE (for a subscription) message. When present in a
+SUBSCRIBE or REQUEST_UPDATE with a fill filter type, it applies to the fill
+fetch stream.
 
 It is the maximum total duration in milliseconds a relay SHOULD spend waiting
 for upstream sources to provide Objects that are not immediately available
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1526,6 +1528,39 @@ delivered will be the Group ID in `Start Location` plus the `End Group Delta`.
 If the resulting Group ID would be greater than 2^64 - 1, the endpoint MUST
 close the session with a `PROTOCOL_VIOLATION`.
 
+AbsoluteStartFill (0x5): Start Location is present. The filter Start Location
+is specified explicitly.
+Objects from Start Location up to but not including `{Largest Object.Group, 0}`
```
sharmafb, line 1533: Maybe I'm missing something here, but why isn't this {Largest Object.Group, Largest Object.Object}? Wouldn't we end up with a gap from {Largest Object.Group, Largest Object.Object + 1} to {Largest Object.Group + 1, 0}?

afrind: See the next sentence:

> Objects from `{Largest Object.Group, 0}` onward are delivered using subscribe subgroups and
datagrams.

sharmafb: Ah, I see. Thanks for the clarification.

draft-ietf-moq-transport.md
```diff
@@ -1526,6 +1528,39 @@ delivered will be the Group ID in `Start Location` plus the `End Group Delta`.
 If the resulting Group ID would be greater than 2^64 - 1, the endpoint MUST
 close the session with a `PROTOCOL_VIOLATION`.
 
+AbsoluteStartFill (0x5): Start Location is present. The filter Start Location
```
ianswett: nit: The above filters don't say "Start Location is present". I think we can probably drop this, but if we want it, it'd be nice to be parallel with AbsoluteStart and AbsoluteRange


```suggestion
AbsoluteStartFill (0x5): The filter Start Location
```

draft-ietf-moq-transport.md
```diff
@@ -1526,6 +1528,39 @@ delivered will be the Group ID in `Start Location` plus the `End Group Delta`.
...
+is specified explicitly.
+Objects from Start Location up to but not including `{Largest Object.Group, 0}`
+are delivered on a fill fetch stream (see {{fill-semantics}}). Objects from
+`{Largest Object.Group, 0}` onward are delivered using subscribe subgroups and
+datagrams. There is no End Group - the subscription is open ended. If the Start
+Location is at or after `{Largest Object.Group, 0}`, the fill range is empty
+and no fill fetch stream is opened.
+
+AbsoluteRangeFill (0x6): Start Location and End Group Delta are present. The
```
```suggestion
AbsoluteRangeFill (0x6): The
```

draft-ietf-moq-transport.md
```diff
@@ -1526,6 +1528,39 @@ delivered will be the Group ID in `Start Location` plus the `End Group Delta`.
...
+filter Start Location and End Group Delta are
+specified explicitly. Objects from Start Location up to but not including
+`{Largest Object.Group, 0}` are delivered on a fill fetch stream (see
+{{fill-semantics}}). Objects from `{Largest Object.Group, 0}` onward are
+delivered using subscribe subgroups and datagrams. If the End Group is before
+`Largest Object.Group`, only fill delivery occurs.
+TODO: Determine behavior when end is before current group — does the
+subscription auto-close, or remain open for possible REQUEST_UPDATE?
```
ianswett, line 1548: I think it should remain open.  I'm not sure how important it is, but you could get delayed objects/subgroups via Subscribe for Groups before the current group, correct?  That combined with a FILL_TIMEOUT that skips Objects that aren't in cache might be useful?

draft-ietf-moq-transport.md
```diff
@@ -1534,16 +1569,81 @@ If the publisher cannot satisfy the requested Subscription Filter (see
 it SHOULD send a REQUEST_ERROR with code `INVALID_RANGE`.  A publisher MUST
 NOT send objects from outside the requested range.
 
+### Current Group Delivery {#current-group-delivery}
+
+When a subscription's filter includes the current group — as with
+CurrentGroup, RelativeStartFill, AbsoluteStartFill with a Start Location at
+or before the current group, or AbsoluteRange that crosses the current group
```
ianswett: The behavior of AbsoluteRange and AbsoluteStart should be the same, correct?  And I thought for both of them we're not changing the existing behavior of them being passthroughs?

draft-ietf-moq-transport.md
```diff
@@ -1534,16 +1569,81 @@ If the publisher cannot satisfy the requested Subscription Filter (see
...
+There is no guarantee the subscriber will receive any subgroups or datagrams
+for the Current Group due to DELIVERY_TIMEOUT, MAX_CACHE_DURATION,
+or lack of data availability.  If the publisher opens a subgroup for the Current
+Group it is unable to complete, it can reset it with CURRENT_GROUP_UNAVAILABLE.
+
+### Fill Semantics {#fill-semantics}
+
+Fill filter types (AbsoluteStartFill, AbsoluteRangeFill, and
+RelativeStartFill) cause the publisher to open a unidirectional stream with a
+FETCH_HEADER (see {{fetch-header}}) using the subscription's Request ID to
```
ianswett: Subscription's or REQUEST_UPDATE's Request ID?

draft-ietf-moq-transport.md
```diff
@@ -1534,16 +1569,81 @@ If the publisher cannot satisfy the requested Subscription Filter (see
...
+deliver objects before the current group. This is called a fill fetch stream.
+
+The fill fetch stream carries objects from the fill Start Location up to but
+not including `{Largest Object.Group, 0}`, where `Largest Object` is the value
+communicated in SUBSCRIBE_OK or REQUEST_UPDATE_OK. For RelativeStartFill, both
+sides compute the fill Start Location as `{Largest Object.Group - (N + 1), 0}`
+using the LARGEST_OBJECT from the response. Subscribe subgroups and datagrams
+carry objects from `{Largest Object.Group, 0}` onward. If the fill Start
```
```suggestion
carry objects from `{Largest Object.Group, 0}` onward. If the Start
```

draft-ietf-moq-transport.md
```diff
@@ -1534,16 +1569,81 @@ If the publisher cannot satisfy the requested Subscription Filter (see
...
+To join a Track at a past Group, the subscriber sends a SUBSCRIBE with a fill
+filter type: AbsoluteStartFill, AbsoluteRangeFill, or RelativeStartFill.
+The publisher responds with SUBSCRIBE_OK including LARGEST_OBJECT.
+Largest Object.Group indicates the boundary between fill and live delivery. If
+Objects prior to the current group are requested, the publisher opens a fill
+fetch stream for objects before `{Largest Object.Group, 0}` and delivers
+current and future objects using subscribe subgroups and datagrams (see
+{{fill-semantics}}).
+
+To join a Track at the current Group, the subscriber sends a SUBSCRIBE with
```
ianswett: nit: I'd probably put this paragraph first because its simpler.

draft-ietf-moq-transport.md
```diff
@@ -1593,6 +1693,18 @@ Since a relay can start delivering FETCH Objects from cache before determining
 the result of the request, some Objects could be received even if the FETCH
 results in error.
 
+## Fill Fetch Stream Cancellation
+
+A fill fetch stream can be cancelled independently of the subscription by
+sending STOP_SENDING on the fill fetch stream. The subscription continues
```
```suggestion
sending STOP_SENDING. The subscription continues
```

draft-ietf-moq-transport.md
```diff
@@ -1593,6 +1693,18 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+to deliver objects using subscribe subgroups and datagrams.
+
+Cancelling the subscription (STOP_SENDING on the bidi stream) cancels both
+the fill fetch stream and forward delivery.  The publisher MUST reset any open
+fill fetch streams associated with a cancelled subscription.
+
+The fill fetch stream is closed with a FIN after all past objects up to the
+fill boundary (LARGEST_OBJECT from SUBSCRIBE_OK) have been delivered.
```
ianswett: It's {Largest Object.Group, 0}, not LARGEST_OBJECT, correct?

draft-ietf-moq-transport.md
```diff
@@ -1534,16 +1569,81 @@ If the publisher cannot satisfy the requested Subscription Filter (see
...
+fill fetch stream carries the Request ID of the message that initiated
+it: the SUBSCRIBE Request ID for the initial fill, or the REQUEST_UPDATE
+Request ID for a subsequent fill. Setting Forward
+State to 0 implicitly cancels any fill
+fetch stream and suppresses forward delivery. The publisher MUST reset any
+open fill fetch streams when Forward State transitions to 0.
+
+A REQUEST_UPDATE containing a SUBSCRIPTION_FILTER with a fill filter type
+opens a new fill fetch stream. The fill fetch stream uses the Request ID from
+the REQUEST_UPDATE. The REQUEST_UPDATE_OK will include LARGEST_OBJECT
```
```suggestion
A REQUEST_UPDATE containing a SUBSCRIPTION_FILTER with a fill filter type
opens a new fill fetch stream with the the Request ID from
```

ianswett: Do we also need to mention forward=1?

draft-ietf-moq-transport.md
```diff
@@ -2051,6 +2168,11 @@ When a Relay needs to make an upstream FETCH request, it determines the
 available publishers using the same matching rules as SUBSCRIBE. When more than
 one publisher is available, the Relay MAY send the FETCH to any of them.
 
+When a Relay receives a downstream SUBSCRIBE with a fill filter type or
```
```suggestion
When a Relay receives a SUBSCRIBE with a fill filter type or
```

draft-ietf-moq-transport.md
```diff
@@ -2196,8 +2318,8 @@ the length of the Message Payload, the receiver MUST close the session with a
 ## Request ID {#request-id}
 
 Request ID is included in request messages and is used to identify
-requests across messages. For example, Joining Fetch references
-the Request ID of a SUBSCRIBE.
+requests across messages. For example, fetch streams reference
+the Request ID of a SUBSCRIBE, PUBLISH or FETCH.
```
```suggestion
the Request ID of a SUBSCRIBE, PUBLISH, FETCH, or REQUEST_UPDATE.
```

draft-ietf-moq-transport.md
```diff
@@ -2426,7 +2548,10 @@ for the same track.
 
 ### FILL TIMEOUT Parameter {#fill-timeout}
 
-The FILL_TIMEOUT parameter (Parameter Type 0x0A) MAY appear in a FETCH message.
+The FILL_TIMEOUT parameter (Parameter Type 0x0A) MAY appear in a FETCH,
+SUBSCRIBE, or REQUEST_UPDATE (for a subscription) message. When present in a
+SUBSCRIBE or REQUEST_UPDATE with a fill filter type, it applies to the fill
```
ianswett, line 2556: If it's present in a subscription that doesn't have a fill filter type, is it ignored?  If so, we should probably say that, because I can also imagine people treating it as an error.

draft-ietf-moq-transport.md
```diff
@@ -2504,6 +2633,17 @@ The SUBSCRIPTION_FILTER parameter (Parameter Type 0x21) uses length-prefixed
 encoding. It MAY appear in a SUBSCRIBE, PUBLISH_OK or REQUEST_UPDATE (for a
 subscription) message. It is a Subscription Filter (see {{subscription-filters}}).
 
+Fill filter types (AbsoluteStartFill, AbsoluteRangeFill, RelativeStartFill)
+and CurrentGroup MUST NOT appear in PUBLISH_OK. A publisher that receives a
+PUBLISH_OK with one of these filter types MUST close the session with
+`PROTOCOL_VIOLATION`. The LARGEST_OBJECT in PUBLISH may be stale by the time
+PUBLISH_OK is processed, making the fill or start boundary unreliable.
```
ianswett: What's a start boundary?


```suggestion
PUBLISH_OK is processed, making the fill boundary unreliable.
```

draft-ietf-moq-transport.md
```diff
@@ -2504,6 +2633,17 @@ The SUBSCRIPTION_FILTER parameter (Parameter Type 0x21) uses length-prefixed
...
+To fill-join a track initiated via PUBLISH, the
+subscriber SHOULD respond with PUBLISH_OK with Forward State 0, then send
```
ianswett, line 2646: Can you not go from Forward State 1 and LargestGroup or LargestObject directly to RelativeStartFill for example?

draft-ietf-moq-transport.md
```diff
@@ -4186,7 +4222,9 @@ as defined in {{stream-reset-codes}}.
 ### Fetch Header {#fetch-header}
 
 When a stream begins with `FETCH_HEADER`, all objects on the stream belong to the
-track requested in the Fetch message identified by `Request ID`.
+track requested in the Fetch message identified by `Request ID`. A FETCH_HEADER
+can also be opened by a publisher for fill delivery, using the Request ID of
+a subscription with a fill filter type (see {{fill-semantics}}).
```
```suggestion
track requested in the message identified by `Request ID`.
```

ianswett (commented):
Looks good, though I had a number of detailed suggestions.

draft-ietf-moq-transport.md
afrind: Start Location both refers to an optional wire format field, and a concept (where the filter starts), but it does read a little awkwardly.

draft-ietf-moq-transport.md
afrind: I agree we should leave it open.

draft-ietf-moq-transport.md
afrind: Should be AbsoluteRangeFill

```suggestion
or before the current group, or AbsoluteRangeFill that crosses the current group
```

draft-ietf-moq-transport.md
```diff
@@ -1534,16 +1569,81 @@ If the publisher cannot satisfy the requested Subscription Filter (see
...
+for the Current Group due to DELIVERY_TIMEOUT, MAX_CACHE_DURATION,
+or lack of data availability.  If the publisher opens a subgroup for the Current
+Group it is unable to complete, it can reset it with CURRENT_GROUP_UNAVAILABLE.
+
+### Fill Semantics {#fill-semantics}
+
+Fill filter types (AbsoluteStartFill, AbsoluteRangeFill, and
+RelativeStartFill) cause the publisher to open a unidirectional stream with a
+FETCH_HEADER (see {{fetch-header}}) using the subscription's Request ID to
+deliver objects before the current group. This is called a fill fetch stream.
```
```suggestion
FETCH_HEADER (see {{fetch-header}}) to deliver objects before the current
group. This is called a fill fetch stream.  The fill fetch stream uses either
subscription's Request ID, or the Request ID of the REQUEST_UPDATE that
triggered the fill.
```

draft-ietf-moq-transport.md
afrind: We say elsewhere that you can send forward=0 with a fill type, but it implicitly cancels the fill.

draft-ietf-moq-transport.md
afrind: Good catch

draft-ietf-moq-transport.md
afrind: In this current version of this PR, Forward State=1 is always Largest Object, and the correct way to join it is 

PUB_OK [fwd=0/1, doesn't matter?]
REQUEST_UPDATE fwd=1, filter=[CurrentGroup | RelativePreviousFill(1) | ... ]

It's awkward, and the solution is to allow the filter type in SUBSCRIBE_TRACKS (e.g. CurrentGroup).  A free publish (e.g. from a client) is awkward since there's no context for the receiver where the publisher is starting from.

draft-ietf-moq-transport.md
```diff
@@ -1534,16 +1569,84 @@ If the publisher cannot satisfy the requested Subscription Filter (see
...
+including subscriber priority and authorization. The FETCH_HEADER on the
+fill fetch stream carries the Request ID of the message that initiated
+it: the SUBSCRIBE Request ID for the initial fill, or the REQUEST_UPDATE
+Request ID for a subsequent fill. Setting Forward
+State to 0 implicitly cancels any fill
+fetch stream and suppresses forward delivery. The publisher MUST reset any
+open fill fetch streams when Forward State transitions to 0.
+
+A REQUEST_UPDATE containing a SUBSCRIPTION_FILTER with a fill filter type
+opens a new fill fetch stream. The fill fetch stream uses the Request ID from
```
sharmafb, line 1623: nit: Duplicated text

draft-ietf-moq-transport.md
martinduke: It should behave exactly like AbsoluteRange. I can see that AbsoluteRange filter is not very specific about what happens in that case. I suppose that if I had already abandoned EndGroup (due to DELIVERY_TIMEOUT or whatever) I'd probably send PUBLISH_DONE immediately. You can't really send REQUEST_ERROR if there's a fill stream.

draft-ietf-moq-transport.md
```diff
@@ -1526,6 +1528,39 @@ delivered will be the Group ID in `Start Location` plus the `End Group Delta`.
...
+There is no End Group - the subscription is open ended.
+
+RelativeStartFill (0x8): Relative Previous is present. The value N in
+Relative Previous determines the Start Location:
+`{Largest Object.Group - (N + 1), 0}`. Objects from Start Location up to but
+not including `{Largest Object.Group, 0}` are delivered on a fill fetch stream
+(see {{fill-semantics}}), and objects from `{Largest Object.Group, 0}` onward
+are delivered using subscribe subgroups and datagrams. If N + 1 is greater than
+Largest Object.Group, the Start Location is `{0, 0}`. There is no End Group -
+the subscription is open ended.
```
martinduke, line 1562: You spelled it wrong -- why not collapse these into 
RelativeStartFill and get rid of the N+1?

martinduke (approved):
I like this.

draft-ietf-moq-transport.md
```diff
@@ -1534,16 +1569,84 @@ If the publisher cannot satisfy the requested Subscription Filter (see
...
+
+A REQUEST_UPDATE containing a SUBSCRIPTION_FILTER with a fill filter type
+opens a new fill fetch stream. The fill fetch stream uses the Request ID from
+A REQUEST_UPDATE containing a SUBSCRIPTION_FILTER with a fill filter type
+opens a new fill fetch stream with the the Request ID from
+establishing the fill boundary for the new fill fetch stream.
+
+FILL_TIMEOUT (see {{delivery-timeouts}}) applies to fill fetch streams in the
+same way it applies to standalone fetches.
+
```
yanpzhou, line 1630: For a subscription with fill filter type, if the current group and the subsequent data are able to deliver, but the fill fetch stream has error like time out, should it be successful subscription or not?

draft-ietf-moq-transport.md
afrind: A relay can't always close an AbsoluteRange subscription once the end has been published.

1) it's possible there's some object that passes the range that is delayed in the network
2) the publisher could be publishing out of order
3) the subscriber may want to update the range

All of these point towards "leave it open".  We also discovered this neat trick that helps VOD -- #1316 

This design allows:

SUBSCRIBE AbsRangeFill [0-10]
REQUEST_UPDATE [11-20]
REQUEST_UPDATE[21-30]

Each will spawn a fetch stream.  REQUEST_UPDATE is lighter weight than issuing a FETCH for each.

draft-ietf-moq-transport.md
afrind: I started with that, but opted for a specific Current Group enum for clarity.  I'm mostly ambivalent.

draft-ietf-moq-transport.md
afrind: This is an excellent question.  We said that cancelling the subscription cancels the FETCH, so I would think any REQUEST_ERROR would apply to both as well.

If one of the halves fails after SUBSCRIBE_OK, I'm unsure.  I'm somewhat inclined to let them be independent at that point.

We might need to add text that clarifies when the SUBSCRIBE_OK can be sent.  It might also be nice if the publisher opened the fill fetch stream and then reset it on error, even if no objects are sent.  The subscriber needs a timer anyways, since the fill-fetch might never come, and it's always good to have a fast-fail mechanism.

draft-ietf-moq-transport.md
afrind: I was chatting with @mzanaty about this.

In PUBLISH fwd=1, Largest Object is _not_ stale, all the objects are arriving and standalone FETCH will work fine.
In PUBLISH fwd=0, Largest Object can be stale.  The subscriber needs the correct Largest Object value in order to properly interpret the FIN on the fill-fetch stream.  So I think the advice to use REQUEST_UPDATE stands.

gwendalsimon:
The earlier Joining FETCH dissent discussions were often about relay behavior, described in fairly vague terms. The current SUBSCRIBE / FETCH dichotomy has the merit of being simple to implement, with clear roles and management.

To check this, I worked through a complete diagram of the relay's behavior on a SUBSCRIBE with a fill filter, aiming to capture every case and flag where the spec lacks precision. The result is the diagram below, from which I derived four points where the text could be improved. None of them blocking imho.
 
```mermaid
flowchart TD
    %% ===== Phase 1: Admission =====
    A0(["SUBSCRIBE received (relay acts as publisher to the subscriber)"]) --> A1{"Filter type?"}
    A1 -->|"unknown / unsupported"| E1["PROTOCOL_VIOLATION"]:::err
    A1 -->|"0x1–0x4 (non-fill)"| B0["Baseline subscribe path (non-fill filter)"]:::ok
    A1 -->|"0x5 / 0x6 / 0x8 (fill)  or  0x7 (CurrentGroup)"| A2{"Forward State?"}
    A2 -->|"0"| E2["fwd=0: no forward delivery;<br/>a fill stream, if any, is cancelled (fill is a no-op)"]
    A2 -->|"1"| U0{"Established upstream subscription?"}

    %% ===== Phase 2: Upstream & boundary =====
    U0 -->|"no"| U1["Establish an upstream subscription for current group + live<br/>(CurrentGroup suggested — 1589); older fill via upstream FETCH (2564)."]:::gap
    U0 -->|"yes (aggregated)"| U2["Reuse the upstream subscription"]
    U1 --> U3["Largest Object L known.<br/>Boundary = {L.Group, 0} (start of the current group)"]
    U2 --> U3
    U3 --> V1{"Publisher can satisfy the filter?<br/>(e.g. AbsoluteRangeFill End Group not entirely in the past — 1567)"}
    V1 -->|"no"| E3["INVALID_RANGE"]:::err
    V1 -->|"yes"| OK0["Send SUBSCRIBE_OK (LARGEST_OBJECT = L)"]:::ok

    %% ===== After SUBSCRIBE_OK: subscribe path ALWAYS; fill path only for fill filters =====
    OK0 -->|"0x5 / 0x6 / 0x7 / 0x8 → current group {L.Group,0} onward + live (subscribe subgroups)"| P0{"Current-group prior subgroup objects<br/>received or known-absent?"}
    OK0 -->|"0x5 / 0x6 / 0x8 only → past groups [Start, {L.Group,0}) (fill fetch stream)"| S0["Compute fill Start from filter type<br/>(0x5 explicit · 0x6 explicit + bounded range · 0x8 N+1 groups back)"]

    %% ----- Fill path (fill filters 0x5/0x6/0x8, on the fill fetch stream) -----
    S0 --> SE{"Fill range empty?<br/>(Start ≥ {L.Group,0})"}
    SE -->|"yes"| SEN["No fill fetch stream opened (1537); live only — done"]:::ok
    SE -->|"no"| RA["Open fill fetch stream (FETCH_HEADER, subscription's Request ID)"]
    RA --> C0{"Cache coverage of [Start,{L.Group,0}) ?"}
    C0 -->|"all objects cached"| CA["Deliver all from cache in group order (descending if FillDescending)"]:::ok
    C0 -->|"partial"| CB{"First object (range Start) cached?"}
    C0 -->|"none (empty cache, or non-caching relay)"| FU
    CB -->|"yes — start present"| CB1["Deliver cached prefix; pause at the first missing object (3479–3482)"]
    CB -->|"no — start missing"| CB2["Cannot start; FETCH the missing prefix upstream first"]
    CA --> D0
    CB1 --> FU
    CB2 --> FU
    FU["Pause + upstream FETCH for the missing / unknown-status objects,<br/>bounded by FILL_TIMEOUT (3480, 2564)"] --> X0{"Upstream result?"}
    X0 -->|"objects returned"| D0
    X0 -->|"FILL_TIMEOUT budget exhausted"| FUG["Report remaining as Unknown gaps (0x10C); continue (2566)"]:::ok
    X0 -->|"upstream FETCH fails"| E5["REQUEST_ERROR; reset fill stream (3483)"]:::err
    E5 --> EM["Effect on the subscribe path (current + live) is unclear<br/>for a publisher-side fill reset — to clarify."]:::gap
    FUG --> D0
    D0["FIN fill fetch stream — fill delivered"]:::ok

    %% ----- Subscribe path (current group + live; concurrent with fill) -----
    P0 -->|"yes"| L0["Deliver current group from {L.Group,0} (cached + newly published),<br/>then live onward, on subscribe subgroups/datagrams (ascending)"]:::ok
    P0 -->|"no: prior objects missing"| CGU["FETCH the missing current-group objects upstream<br/>(or an upstream CurrentGroup subscription if none — 1589)"]:::gap
    CGU -->|"completed"| L0
    CGU -->|"cannot complete"| E7["CURRENT_GROUP_UNAVAILABLE (resets this subgroup)"]:::err
    E7 --> EK["Subscription continues with future (live) groups (1578). To discuss: a cold relay /<br/>no-cache live Publisher yields CURRENT_GROUP_UNAVAILABLE + a fill REQUEST_ERROR, yet<br/>future objects keep flowing — two failures + a live subscription is valid but odd."]:::gap

    %% (Scope: relay behaviour at SUBSCRIBE reception — follow-up lifecycle
    %%  REQUEST_UPDATE / STOP_SENDING is out of scope; delivery is the terminal.)

    %% ===== Legend (self-contained) =====
    subgraph Legend
      direction LR
      Lok["Well-defined behaviour"]:::ok
      Lerr["Defined error / stream reset"]:::err
      Lgap["To be discussed further"]:::gap
    end

    classDef err fill:#fdd,stroke:#c00,color:#900;
    classDef gap fill:#ffe8cc,stroke:#e8820c,stroke-dasharray:4 3,color:#7a4a00;
    classDef ok fill:#e6f4ea,stroke:#137333,color:#0b5a26;
```

I hope it helps

gwendalsimon:
**U1** in https://github.com/moq-wg/moq-transport/pull/1642#issuecomment-4663419565 No Established Subscription.

I'm not sure why the fill filter is not forwarded upstream. It would be simpler. The upstream sourcing (when the relay must FETCH vs SUBSCRIBE) could be specified more precisely, unless I miss something

gwendalsimon:
**CGU** in https://github.com/moq-wg/moq-transport/pull/1642#issuecomment-4663419565 (current-group prefix missing from cache)

When the relay completes the current group via an upstream FETCH, the objects arrive on a single FETCH stream (FETCH semantics), but the relay must forward them downstream on subgroup streams. FETCH objects do carry Subgroup ID, so this is mechanically possible, but the reassembly (from single FETCH stream to multiple subgroup streams, preserving per-subgroup order and completeness) is not described. Should it be specified?

gwendalsimon:
**EM** in https://github.com/moq-wg/moq-transport/pull/1642#issuecomment-4663419565 (fill fetch fails upstream)

When a publisher-side fill fetch fails (`REQUEST_ERROR`), the effect on the subscribe path (current + live) is unclear. Since the fill stream shares the subscription's Request ID, please confirm a fill failure does not tear down the subscription.

gwendalsimon:
**EK** in https://github.com/moq-wg/moq-transport/pull/1642#issuecomment-4663419565 current group cannot complete

A case where the cold relay and no-cache live Publisher can return both `CURRENT_GROUP_UNAVAILABLE` and a fill `REQUEST_ERROR`, yet future/live objects keep flowing. This composite (two failure signals plus a live-only subscription) is a valid use-case but it seems counterintuitive. Should we make explicit that it is possible?

vasilvv:
Per @afrind request, adding a comment to remind him that he should add ability to fill-fetch to the largest object.

To elaborate on the use case:

If all objects on the enhancement layers are independent, the subscriber can subscribe at the largest object, and then do a fill-fetch up to that object for subgroup=0 only.  This is helpful, since the skipped enhancement items in the past are not useful (they are neither displayed, nor are useful for decoding of the further objects), and skipping them is ~free bandwidth savings.

afrind:
This has been superseded by #1673 - please direct any comments there.

Leaving this open for reference.

afrind:
I think 1673 has been open long enough, and this pr will always be here.  Closing.

PR #1654, opened 2026-06-09
Editorial: Move generic URI-scheme prose out of Fragment Identifiers; reorder Session establishment

michalhosna:
The Fragment Identifiers subsection contained text generic to the whole moqt URI scheme.

  Two commits:

  1. Fix the misplacement: move every generic line (media type included) back under MOQT URI Scheme, leaving Fragment Identifiers with only fragment-specific text. Pure relocation, no wording changes.
  2. Move Fragment Identifiers to the end: the fragment is the last component of a URI, so it reads more naturally after the transport subsections. Pure move.


afrind (commented):
I sort of liked having the syntax grouped together.  Could the same problem be addressed by keeping the original order, but adding a new subheading like ### Dereferencing a MOQT URI?

michalhosna:
It's definetly smaller diff. Fine by me.

afrind (approved):

Issue #1503, opened 2026-02-12
Can use PUBLISH_NAMESPACE in lieu of NAMESPACE? Is this a security issue?


martinduke:
If a subscriber has an open SUBSCRIBE_NAMESPACE, can a publisher send PUBLISH_NAMESPACE on the control stream (or PUBLISH_NAMESPACE stream in the editors' draft) instead of NAMESPACE on the SUB_NS stream if the namespace is within the SUB_NS prefix?

This seems like annoying behavior, but there's an edge case where the SUBSCRIBE_NAMESPACE or its REQUEST_OK is in flight and a PUBLISH_NAMESPACE is in flight as well.

Also, in principle I might not want to have PN remain in effect even if the SUB_NS goes away (as closing that stream is an implicit NS_DONE). Don't know if this is a real use case or not.

So I'm afraid the answer may have to be yes even though I very much want it to be no.

martinduke:
Relatedly: how does AUTH work in SUB_NS streams?

If I SUBSCRIBE_NS * to Alan, and then he sends NAMESPACE for us-govt, do I just have to accept that as truth? Is the ability to send AUTH a reason to use PUBLISH_NAMESPACE, or do we have to add parameters to NAMESPACE? And if we add to NAMESPACE, does that mean we need REQUEST_OK/REQUEST_ERROR

martinduke:
After offline discussion in Boulder, it sounds like using SUB_NS contains an underlying assumption that the publisher will be authorized to publish that prefix. That is logically coherent, but if so there should be text in the SUB_NS section and/or security considerations to indicate when it is safe to send SUB_NS.

afrind:
See #799 for related content that has a similar conclusion - if you send subscribe namespace to someone, you implicitly trust that they are authoritative for the tracks in that namespace, so separate auth is not required.

A relay has no reason to send a publish_namespace in draft-16. If we’re talking about direct publisher to subscriber connection, then yes, it’s possible the ships pass in the night.

afrind:
Closed by #1656

PR #1656, opened 2026-06-10
Address concerns around PubNs and SubNs Authz 

### LINKED ISSUE #799: How can end subscriber verify an authorization claim in an ANNOUNCE
So in -10 the ANNOUNCE section (8.18) states:

The receiver verifies the publisher is authorized to publish tracks under this namespace.

So with the proposed authorization model how can End Subscriber verify this authroization? I can see that Relay's part of a CDN will have the necessary keys to verify the upstream being either part of the same CDN and context, or that an original publisher is authorized for this track name space.

suhasHere:
Fixes #799 and #1503

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -4680,6 +4680,33 @@ which resources the endpoint providing the token is authorized to
 perform and access. Relays will verify the
 token to ensure that the request is authorized.
 
+### Authorization Verification by Role
+
+The following describes how authorization verification applies to Original
+Publishers, End Subscribers, and Relays (as defined in {{terms}}):
+
+Authorization verification differs by role in the delivery chain:
+
+- Original Publishers are verified by Relays using AUTHORIZATION TOKEN
+  on PUBLISH_NAMESPACE and PUBLISH messages. A Relay MUST NOT forward
+  content from an Original Publisher that has not been authorized.
+
+- End Subscribers that send SUBSCRIBE_NAMESPACE or SUBSCRIBE_TRACKS
+  implicitly trust the Relay to have verified Original Publishers for
+  namespaces within the requested prefix. The application provides the
+  End Subscriber with the Relay identity and namespace; sending
+  SUBSCRIBE_NAMESPACE to that Relay constitutes delegation of publisher
+  verification.
+
+- Relay-to-Relay authorization is deployment-specific and outside the
+  scope of this document.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1660,6 +1665,14 @@ publisher, it MUST send a NAMESPACE message to any subscriber that has
 sent SUBSCRIBE_NAMESPACE for that namespace, or a prefix of that
 namespace. A publisher MAY send the PUBLISH_NAMESPACE to any other subscriber.
 
+A subscriber MAY receive a PUBLISH_NAMESPACE on the control stream for a
+namespace that falls within an active SUBSCRIBE_NAMESPACE prefix. This can
+occur when SUBSCRIBE_NAMESPACE or its response is in flight at the same time
```
afrind: I think can works better than MAY, since it's not something the subscriber chooses to do.

```suggestion
A subscriber can receive a PUBLISH_NAMESPACE on the control stream for a
namespace that falls within an active SUBSCRIBE_NAMESPACE prefix. This
occurs when SUBSCRIBE_NAMESPACE or its response is in flight at the same time
```

draft-ietf-moq-transport.md
```diff
@@ -4680,6 +4693,33 @@ which resources the endpoint providing the token is authorized to
 perform and access. Relays will verify the
 token to ensure that the request is authorized.
 
+### Authorization Verification by Role
```
afrind: Maybe we should drop this section for now, pending the Auth design team output, but I think the first two paragraphs address the issues by themselves?

vasilvv (approved):

draft-ietf-moq-transport.md
```diff
@@ -1660,6 +1665,14 @@ publisher, it MUST send a NAMESPACE message to any subscriber that has
 sent SUBSCRIBE_NAMESPACE for that namespace, or a prefix of that
 namespace. A publisher MAY send the PUBLISH_NAMESPACE to any other subscriber.
 
+A subscriber can receive a PUBLISH_NAMESPACE on the control stream for a
```
ianswett: There isn't a single control stream anymore

```suggestion
A subscriber can receive a PUBLISH_NAMESPACE on a request stream for a
```

```suggestion
A subscriber can receive a PUBLISH_NAMESPACE for a
```

ianswett (approved):

afrind (approved):

Issue #1354, opened 2025-11-04
Why do we need a dedicated SWITCH message?


acbegen:
To date, numerous discussions have been held to understand and devise meaningful solutions for track switching in MOQT. The following is the list of open issues tagged with the ABR label.

- #259 
The challenge involves a publisher providing multiple quality versions of the same content (e.g., 360p and 720p video), where the subscriber needs to receive only the best quality version they can handle. The argument is that the publisher knows what it can send and should decide what to send.

- #370 
Previously, we implemented this through a probing track and whenever the subscriber needed a probing test, it would subscribe to this track.

- #471 
In client-side ABR streaming, the decision to switch to a higher quality stream can lead to excessive bandwidth usage, potentially causing network congestion and degrading the overall user experience. To address this, the proposed solution introduces a "Less than Best Effort" (LBE) mechanism. With LBE, higher-quality (but lower-priority) content is transmitted only if the congestion window (CWND) remains below a specified threshold, such as 1.5 times the Bandwidth-Delay Product (BDP). This approach ensures that LBE data does not compete with or crowd out higher-priority traffic, thereby maintaining network stability and prioritizing essential data delivery.

- #475 
There's no effective way to keep two subscriptions with identical priority in sync, which becomes problematic when one subscription experiences delays (such as during transcoding), potentially causing them to drift out of sync and impact the quality of experience, especially for low-latency use cases.

- #534 
The author proposes allowing for at least padding unidirectional streams, which can then be used for probing.

- #1101 
The issue raises concerns about how to achieve a smooth, gapless switch between two synchronized media tracks in MOQT, such as switching video quality (ABR) or audio languages, using the existing SUBSCRIBE and SUBSCRIBE_UPDATE methods, possibly with track priorities. The goal is to avoid both interruptions and duplicate delivery of the same content during the switch, particularly for live streaming. Previous attempts to address this issue with track priorities have led to conflicting approaches, highlighting ambiguities and limitations in the current protocol. The discussion aims to identify and address these shortcomings and initiate work on protocol improvements to enable reliable and seamless track switching.

**What we have now:**

As advocates of client-side ABR and server-assisted ABR techniques, we implemented the following:
- The first ABR demo that used client-side information, such as client bandwidth measured on the client side and some congestion signals, such as late-arriving MOQ groups. 
- Bandwidth probing from the relay to assist the client in doing ABR.

[...]

PR #1673, opened 2026-06-14
Replace Joining FETCH with fill fetch streams

afrind:
A revision of #1642, this PR allows any range of objects passed to subscription to be satisfied.  The portion of the range before Largest Object comes on a "fill fetch stream" -- a unidirectional stream beginning with a FETCH_HEADER.

- Renamed AbsoluteStart/Range filters to AbsoluteStart/RangeFill, and added a new filter: RelativeStartFill.
- Defined the semantics of fill-fetch and the streams are opened and closed
- Added a FILL_PARAMETERS parameter, which carries the fill-specific versions of SUBSCRIBER_PRIORITY, GROUP_ORDER, etc
- Updated the priority algorithm to explain how to order objects on fill-fetch vs the live portion of a subscription
  - Different Groups, same Group Order -> object's group determines
  - Different Groups and Different Group Order -> live wins
  - Same Group -> fill-fetch wins
- Explained how PUBLISH_OK and REQUEST_UPDATE work with new filter types
- Removed the Joining version of FETCH and removed the "standalone" monikker
- Minor updates to PUBLISH_DONE to include fill-fetch streams in stream count, and allow the publisher to delay the message
- INVALID_RANGE now only applies to FETCH

Fixes: #1039
Fixes: #1313
Fixes: #1614

PR #1674, opened 2026-06-14
Track Switching via the SWITCH_FROM parameter

### LINKED ISSUE #1519: Improve design of requests blocking on other requests

afrind:
Introduce a new parameter SWITCH_FROM, that can appear in SUBSCRIBE, PUBLISH_OK or REQUEST_UPDATE.  The parameter has a Mode enum which can control details of the switching algorithm.  The base algorithm and a single "Hard" mode that flips the Forward State are defined.  Subsequent PRs will introduce additional modes.

The parameter also has a "Publish Done" option which the subscriber uses to control whether the other track is closed or left open after Switch.

See #1354
Fixes #1519

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2222,7 +2222,7 @@ one publisher is available, the Relay MAY send the FETCH to any of them.
 
 When a Relay receives a SUBSCRIBE with a fill filter type, it serves the fill
 portion from its cache and retrieves any missing objects upstream using
-SUBSCRIBE or FETCHes (see {{fill-semantics}}).
+a SUBSCRIBE with fill or FETCHes (see {{fill-semantics}}).
 
 When a Relay receives an authorized SUBSCRIBE for a Track with one or more
 `Established` upstream subscriptions, it MUST reply with SUBSCRIBE_OK.  If the
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1542,7 +1542,8 @@ A publisher MUST NOT send objects from outside the requested range.
 
 Fill filter types (AbsoluteStartFill, AbsoluteRangeFill, and
 RelativeStartFill) cause the publisher to open a unidirectional stream beginning
-with a FETCH_HEADER (see {{fetch-header}}). This is called a fill fetch stream.
+with a FETCH_HEADER (see {{fetch-header}}) and delivered as a FETCH response (see
+{{message-fetch}}). This is called a fill fetch stream.
 
 The fill fetch stream carries the fill range: the objects from the fill Start
 Location up to and including the fill boundary. The fill boundary
@@ -1572,7 +1573,9 @@ fill fetch stream.
 
 An object delivered on the fill fetch stream is **fill-delivered**. Because the
 subscription filter necessarily overlaps the fill range, an object can be both
-fill-delivered and subscription-delivered.
+fill-delivered and subscription-delivered. A subscriber can avoid duplicate
+delivery by setting FILL_TIMEOUT to 0 in FILL_PARAMETERS (see {{fill-timeout}}):
+the fill fetch stream then carries only objects that are immediately available.
 
 #### Opening and Closing Fill Fetch Streams
 
@@ -2636,10 +2639,11 @@ The following parameters MAY appear inside FILL_PARAMETERS:
 A parameter that is omitted from FILL_PARAMETERS takes the value it has for the
 subscription; FILL_PARAMETERS therefore carries only the settings that
 differ. An endpoint that receives a parameter inside FILL_PARAMETERS that is not
-permitted above MUST close the session with `PROTOCOL_VIOLATION`.
+listed above MUST close the session with `PROTOCOL_VIOLATION`.
 
-FILL_PARAMETERS is meaningful only with a fill filter type. If it is present
-without a fill filter type, it is ignored.
+FILL_PARAMETERS is meaningful only with a fill filter type. An endpoint that
+receives FILL_PARAMETERS in a message that specifies a non-fill filter type MUST
+close the session with `PROTOCOL_VIOLATION`.
 
 ### EXPIRES Parameter {#expires}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1687,8 +1687,8 @@ On receiving a message containing SWITCH_FROM, the publisher:
    subscription is in Forward State 0.
 
 2. Sets the activating subscription to Forward State 1 and applies the
-   SUBSCRIPTION_FILTER. This ensures objects after the Current Group are not
-   missed.
+   current SUBSCRIPTION_FILTER. This ensures objects after the Current Group
+   are not missed.
 
 3. Waits until it is ready to publish an object from the SUBSCRIPTION_FILTER's
    Start Location's Group (the Start Group; see {{subscription-filters}}),
@@ -1720,7 +1720,7 @@ Start Group on the activating track, it MUST respond with REQUEST_ERROR `TIMEOUT
 
 ### Relay Handling of SWITCH_FROM {#relay-switch-from}
 
-When a relay receives a message containing SWITCH_FROM, it MUST NOT forward the
+When a relay receives a message containing SWITCH_FROM, it SHOULD NOT forward the
 SWITCH_FROM parameter upstream. The relay handles the switch locally, applying
 the start group computation from {{track-switching}} using its locally observed
 state for the activating track and servicing any fill fetch stream from its cache
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1667,6 +1667,63 @@ Since a relay can start delivering FETCH Objects from cache before determining
 the result of the request, some Objects could be received even if the FETCH
 results in error.
 
+## Track Switching {#track-switching}
+
+A subscriber can atomically stop delivery on one subscription (the suspend
+subscription) and start delivery on another (the resume subscription) by
+including the SWITCH_FROM parameter ({{switch-from}}) in a SUBSCRIBE or
+REQUEST_UPDATE on the resume subscription's request stream.  This enables track
+switching (e.g., ABR quality changes, alternate camera angles, or meeting
+participants) without tearing down subscriptions or issuing standalone fetches.
+
+On receiving a message containing SWITCH_FROM, the publisher:
+
+1. Validates that Switch From Request ID identifies an existing subscription
+   and is not the same as the resume subscription's Request ID. If not,
+   responds with REQUEST_ERROR `INVALID_SWITCH`. The resume subscription
+   MUST be transitioning from Forward State 0 to 1 as part of the switch; if
+   it already has Forward State 1, or the message carrying SWITCH_FROM has
+   FORWARD = 0, the publisher responds with REQUEST_ERROR `INVALID_SWITCH`.
+
+2. Sets resume to Forward State 1 and applies the SUBSCRIPTION_FILTER. This
+   ensures objects after the Current Group are not missed.
+
+3. Waits until it is ready to publish an object from the SUBSCRIPTION_FILTER's
+   Start Location's Group (the Start Group; see {{subscription-filters}}),
+   computed from the resume track at the time the request is received, while
+   continuing to deliver objects on the suspend subscription.  When GROUP_ORDER
+   is Descending and Start Group is before the Current Group, the publisher
+   waits for the first object in the largest fill group instead.
+
+4. Stops delivery on the suspend subscription:
+
+   * Mode Hard (0x0): sets Forward State 0 on the suspend subscription. If
+     Publish Done is 1, the publisher also sends PUBLISH_DONE.
+
+   In all modes, the publisher also resets any outstanding suspend data for
+   groups greater than or equal to the Start Group; objects already in flight
+   can still be received by the subscriber.
+
+   If Publish Done is 0, the suspend subscription remains established.
+
+5. Begins delivery of resume subscription from Start Group. If the
+   SUBSCRIPTION_FILTER is a fill filter type, opens a fill fetch stream using
+   the resume subscription's Request ID (see {{fill-semantics}}), if necessary.
+
+6. Responds with SUBSCRIBE_OK or REQUEST_OK as appropriate, including
+   LARGEST_OBJECT if the SUBSCRIPTION_FILTER is a fill filter type.
+
+If the publisher times out waiting to be ready to publish an object from the
+Start Group on the resume track, it MUST respond with REQUEST_ERROR `TIMEOUT`.
+
+### Relay Handling of SWITCH_FROM {#relay-switch-from}
+
+When a relay receives a message containing SWITCH_FROM, it MUST NOT forward the
+SWITCH_FROM parameter upstream. The relay handles the switch locally, applying
+the start group computation from {{track-switching}} using its locally observed
+state for the resume track and servicing any fill fetch stream from its cache
+and upstream sources as described in {{fill-semantics}}.
+
 # Namespace Discovery {#track-discovery}
 
 Discovery of MOQT servers is always done out-of-band. Namespace discovery can be
@@ -2595,6 +2652,37 @@ close the session with `PROTOCOL_VIOLATION`.
 If omitted from SUBSCRIBE, the publisher's preference from
 the Track is used. If omitted from FETCH, the receiver uses Ascending (0x1).
 
+### SWITCH_FROM Parameter {#switch-from}
+
+The SWITCH_FROM parameter (Parameter Type 0x24) consists of a varint Switch
+From Request ID, a varint Mode, and a single byte of flags. It MAY appear in a
+SUBSCRIBE or REQUEST_UPDATE (for a subscription) message.
+
+~~~
+SWITCH_FROM {
+  Switch From Request ID (vi64),
+  Mode (vi64),
+  Publish Done (1),
+  Reserved (7),
+}
+~~~
+
+* Switch From Request ID: The Request ID of the subscription to suspend.
+
+* Mode: A vi64 enum selecting how the suspend subscription is stopped. The
+  following mode is defined: Hard (0x0). An endpoint that receives a Mode value
+  that is not defined MUST close the session with `PROTOCOL_VIOLATION`. See
+  {{track-switching}}.
+
+* Publish Done: If 1, the publisher sends PUBLISH_DONE on the suspend
+  subscription as described in {{track-switching}}.
+
+* Reserved: MUST be 0. An endpoint that receives a non-zero value MUST
+  close the session with `PROTOCOL_VIOLATION`.
+
+When present, the enclosing SUBSCRIBE or REQUEST_UPDATE initiates a track
+switch as described in {{track-switching}}.
+
 ### SUBSCRIPTION FILTER Parameter {#subscription-filter}
 
 The SUBSCRIPTION_FILTER parameter (Parameter Type 0x21) uses length-prefixed
@@ -3095,6 +3183,13 @@ INVALID_RANGE:
 MALFORMED_TRACK:
 : In response to a FETCH, a relay publisher detected the track was
 malformed (see {{malformed-tracks}}).
+
+INVALID_SWITCH:
+: In response to a SUBSCRIBE or REQUEST_UPDATE carrying SWITCH_FROM, the track
+switch cannot be performed: the Switch From Request ID does not identify an
+existing subscription, is the same as the resume subscription's Request ID, the
+resume subscription is not transitioning from Forward State 0 to 1, or the
+message carrying SWITCH_FROM has FORWARD = 0 (see {{track-switching}}).
 
 The following are errors for use by the subscriber. They can appear in response
 to PUBLISH or PUBLISH_NAMESPACE, unless otherwise noted.
@@ -5014,6 +5109,7 @@ Setup Options SHOULD request a provisional registration.
 | 0x21 | SUBSCRIPTION_FILTER | {{subscription-filter}} |
 | 0x22 | GROUP_ORDER | {{group-order}} |
 | 0x23 | FILL_PARAMETERS | {{fill-parameters}} |
+| 0x24 | SWITCH_FROM | {{switch-from}} |
 | 0x32 | NEW_GROUP_REQUEST | {{new-group-request}} |
 | 0x34 | TRACK_NAMESPACE_PREFIX | {{track-namespace-prefix-param}} |
 
@@ -5133,6 +5229,7 @@ This document does not define any initial entries.
 | UNINTERESTED               | 0x20 | {{message-request-error}} |
 | PREFIX_OVERLAP             | 0x30 | {{message-request-error}} |
 | NAMESPACE_TOO_LARGE        | 0x31 | {{message-request-error}} |
+| INVALID_SWITCH             | 0x32 | {{message-request-error}} |
 | UNSUPPORTED_EXTENSION      | 0x33 | {{message-request-error}} |
 | REDIRECT                   | 0x34 | {{message-request-error}} |
 | Reserved for greasing      | 0x7f * N + 0x9D | {{grease}} |
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1639,9 +1639,8 @@ fill filter type) while Forward State is 1.
   one whose Start Location does not add objects beyond an earlier fill (for
   example, one that only decreases the End Group), does not open a new fill fetch
   stream.
-- When Forward State transitions to 0, or the subscription is cancelled with
-  STOP_SENDING on the bidi stream, the publisher MUST reset any open fill fetch
-  streams.
+- When the subscription is cancelled with STOP_SENDING on the bidi stream, the
+  publisher MUST reset any open fill fetch streams.
 
 The publisher signals that the fill is complete by closing the stream with a FIN
 once all objects up to and including the fill boundary have been
@@ -3322,8 +3321,9 @@ REQUEST_UPDATE Message {
 
 When a subscriber decreases the Start Location of the Subscription Filter
 (see {{subscription-filters}}), the Start Location can be smaller than the Track's
-Largest Location, similar to a new Subscription. The publisher delivers the new
-fill range by opening a new fill fetch stream (see {{fill-semantics}}).
+Largest Location, similar to a new Subscription. If a fill filter is specified,
+the publisher delivers the new fill range by opening a new fill fetch stream
+(see {{fill-semantics}}).
 
 When a subscriber increases the End Location, the Largest Object at
 the publisher might already be larger than the previous End Location. This will
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1697,7 +1697,7 @@ On receiving a message containing SWITCH_FROM, the publisher:
    computed from the activating track at the time the request is received, while
    continuing to deliver objects on the suspending subscription.  When GROUP_ORDER
    is Descending and Start Group is before the Current Group, the publisher
-   waits for the first object in the largest fill group instead.
+   waits for the first object in the largest group in the fill range instead.
 
 4. Stops delivery on the suspending subscription:
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1593,7 +1593,7 @@ RelativeStartFill) cause the publisher to open a unidirectional stream beginning
 with a FETCH_HEADER (see {{fetch-header}}) and delivered as a FETCH response (see
 {{message-fetch}}). This is called a fill fetch stream.
 
-The fill fetch stream carries the fill range: the objects from the fill Start
+The fill fetch stream carries the objects from the fill Start
 Location up to and including the fill boundary. The fill boundary
 is the Largest Object, or the End Group when one is specified with a Location
 less than the Largest Object. The subscriber
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1669,12 +1669,12 @@ results in error.
 
 ## Track Switching {#track-switching}
 
-A subscriber can atomically stop delivery on one subscription (the suspending
-subscription) and start delivery on another (the activating subscription) by
+A subscriber can atomically stop delivery on one subscription (the _suspending
+subscription_) and start delivery on another, possibly new subscription (the _activating subscription_) by
 including the SWITCH_FROM parameter ({{switch-from}}) in a SUBSCRIBE or
 REQUEST_UPDATE on the activating subscription's request stream.  This enables
-track switching (e.g., ABR quality changes, alternate camera angles, or meeting
-participants) without tearing down subscriptions or issuing standalone fetches.
+subscriber track switching (e.g., ABR quality changes, alternate camera angles, or meetings
+participants) without coordinating multiple messages.
 
 On receiving a message containing SWITCH_FROM, the publisher:
 
@@ -1683,12 +1683,14 @@ On receiving a message containing SWITCH_FROM, the publisher:
    responds with REQUEST_ERROR `INVALID_SWITCH`. The activating subscription
    handles Forward State transitions as part of the switch; if the message
    carrying SWITCH_FROM has a FORWARD parameter, the publisher responds with
-   REQUEST_ERROR `INVALID_SWITCH`.  Note it is not an error if the pausing
+   REQUEST_ERROR `INVALID_SWITCH`.  Note it is not an error if the suspending
    subscription is in Forward State 0.
 
 2. Sets the activating subscription to Forward State 1 and applies the
-   current SUBSCRIPTION_FILTER. This ensures objects after the Current Group
-   are not missed.
+   the most recent SUBSCRIPTION_FILTER parameter for this subscription (which
+   can be in the same message carrying SWITCH_FROM). This ensures objects
+   in Groups greater than or equal to the Largest Object's Group
+   are not missed during the transition.
 
 3. Waits until it is ready to publish an object from the SUBSCRIPTION_FILTER's
    Start Location's Group (the Start Group; see {{subscription-filters}}),
@@ -1703,16 +1705,16 @@ On receiving a message containing SWITCH_FROM, the publisher:
      Publish Done is 1, the publisher also sends PUBLISH_DONE with code
      SUBSCRIPTION_ENDED.
 
-   In all modes, the publisher also resets any outstanding streams on the suspending
+   In all modes, the publisher also resets any outstanding streams (including fill-fetch streams) on the suspending
    subscription; objects already in flight can still be received by the subscriber.
 
    If Publish Done is 0, the suspending subscription remains established.
 
 5. Begins delivery of activating subscription from Start Group. If the
    SUBSCRIPTION_FILTER is a fill filter type, opens a fill fetch stream using
-   the resuming subscription's Request ID (see {{fill-semantics}}), if necessary.
+   the activating subscription's Request ID (see {{fill-semantics}}).
 
-6. Responds with SUBSCRIBE_OK or REQUEST_OK as appropriate, including
+6. Responds on the activating subscription's control stream with SUBSCRIBE_OK or REQUEST_OK as appropriate, including
    LARGEST_OBJECT if the SUBSCRIPTION_FILTER is a fill filter type.
 
 If the publisher times out waiting to publish an object from the
@@ -1720,11 +1722,11 @@ Start Group on the activating track, it MUST respond with REQUEST_ERROR `TIMEOUT
 
 ### Relay Handling of SWITCH_FROM {#relay-switch-from}
 
-When a relay receives a message containing SWITCH_FROM, it SHOULD NOT forward the
-SWITCH_FROM parameter upstream. The relay handles the switch locally, applying
-the start group computation from {{track-switching}} using its locally observed
-state for the activating track and servicing any fill fetch stream from its cache
-and upstream sources as described in {{fill-semantics}}.
+Relays ordinarily handle the switch locally, applying
+the start group computation from {{track-switching}} using locally observed
+state for the activating track and servicing any fill fetch stream from cache
+and upstream sources as described in {{fill-semantics}}.  When a relay performs
+the switch operation, it MUST NOT forward the SWITCH_FROM parameter upstream.
 
 # Namespace Discovery {#track-discovery}
 
@@ -3189,9 +3191,7 @@ malformed (see {{malformed-tracks}}).
 
 INVALID_SWITCH:
 : In response to a SUBSCRIBE or REQUEST_UPDATE carrying SWITCH_FROM, the track
-switch cannot be performed: the Switch From Request ID does not identify an
-existing subscription, is the same as the activating subscription's Request ID, or the
-request contained the FORWARD parameter (see {{track-switching}}).
+switch cannot be performed (see {{track-switching}}).
 
 The following are errors for use by the subscriber. They can appear in response
 to PUBLISH or PUBLISH_NAMESPACE, unless otherwise noted.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2671,10 +2671,11 @@ SWITCH_FROM {
 
 * Switch From Request ID: The Request ID of the subscription to suspend.
 
-* Mode: A vi64 enum selecting how the suspending subscription is stopped. The
-  following mode is defined: Hard (0x0). An endpoint that receives a Mode value
-  that is not defined MUST close the session with `PROTOCOL_VIOLATION`. See
-  {{track-switching}}.
+* Mode: A vi64 enum selecting how the suspending subscription is stopped. Modes
+  are registered in the IANA table "MOQT SWITCH_FROM Modes"
+  ({{iana-switch-from-modes}}); the following mode is defined: Hard (0x0). An
+  endpoint that receives a Mode value that is not defined MUST close the session
+  with `PROTOCOL_VIOLATION`. See {{track-switching}}.
 
 * Publish Done: If 1, the publisher sends PUBLISH_DONE on the suspend
   subscription as described in {{track-switching}}.
@@ -5115,6 +5116,16 @@ Setup Options SHOULD request a provisional registration.
 | 0x34 | TRACK_NAMESPACE_PREFIX | {{track-namespace-prefix-param}} |
 
 * Message Parameters - List which params can be repeated in the table.
+
+## SWITCH_FROM Modes {#iana-switch-from-modes}
+
+This document establishes the "MOQT SWITCH_FROM Modes" registry, governing the
+Mode field of the SWITCH_FROM parameter (see {{switch-from}}). The registration
+policy is First Come First Served (per {{!RFC8126, Section 4.4}}).
+
+| Mode | Name     | Specification |
+|-----:|:---------|:--------------|
+| 0x0  | Hard     | {{switch-from}} |
 
 ## Properties {#iana-properties}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1411,9 +1411,7 @@ PUBLISH_DONE ({{message-publish-done}}) are sent regardless of the forward state
 
 A publisher MUST save the Largest Location communicated in SUBSCRIBE_OK, PUBLISH
 or REQUEST_UPDATE_OK that changes the Forward State
-from 0 to 1.  This value is called the Joining Location and can be used in a
-Joining FETCH (see {{joining-fetches}}) while the subscription is in the
-`Established` state.
+from 0 to 1.
 
 Either endpoint can initiate a subscription to a track without exchanging any
 prior messages other than SETUP.  Relays MUST NOT send any PUBLISH messages
@@ -1492,6 +1490,7 @@ Subscription Filter {
   Filter Type (vi64),
   [Start Location (Location),]
   [End Group Delta (vi64),]
+  [Relative Previous (vi64),]
 }
 ~~~
 
@@ -1509,30 +1508,96 @@ Next Group Start (0x1): The filter Start Location is `{Largest Object.Group + 1,
 0}` and `Largest Object` is communicated in SUBSCRIBE_OK. If no content has been
 delivered yet, the filter Start Location is {0, 0}.  There is no End Group -
 the subscription is open ended. For scenarios where the subscriber intends to
-start from more than one group in the future, it can use an AbsoluteStart filter
-instead.
+start from more than one group in the future, it can use an AbsoluteStartFill
+filter instead.
 
-AbsoluteStart (0x3): The filter Start Location is specified explicitly. The
-specified `Start Location` MAY be less than the `Largest Object` observed at the
-publisher. There is no End Group - the subscription is open ended.  An
-AbsoluteStart filter with `Start` = {0, 0} is equivalent to an unfiltered
-subscription.
+AbsoluteStartFill (0x3): The filter Start Location is specified explicitly.
+There is no End Group - the subscription is open ended. The subscription opens a
+fill fetch stream for already-published objects (see {{fill-semantics}}).
 
-AbsoluteRange (0x4): The filter Start Location and End Group are specified
-explicitly. The specified `Start Location` MAY be less than the `Largest Object`
-observed at the publisher. If the specified `End Group Delta` is zero, the
-remainder of that Group passes the filter. Otherwise, the last Group ID to be
-delivered will be the Group ID in `Start Location` plus the `End Group Delta`.
-If the resulting Group ID would be greater than 2^64 - 1, the endpoint MUST
-close the session with a `PROTOCOL_VIOLATION`.
+AbsoluteRangeFill (0x4): The filter Start Location and End Group Delta are
+specified explicitly. If the specified `End Group Delta` is zero, the remainder
+of the Start Location's Group passes the filter. Otherwise, the last Group ID to
+be delivered is the Group ID in `Start Location` plus the `End Group Delta`. If
+the resulting Group ID would be greater than 2^64 - 1, the endpoint MUST close
+the session with a `PROTOCOL_VIOLATION`. The subscription opens a fill fetch
+stream for already-published objects (see {{fill-semantics}}).
+
+RelativeStartFill (0x5): Relative Previous is present. The value N in
+Relative Previous determines the Start Location: `{Largest Object.Group - N, 0}`
+(N = 0 starts at Largest Object.Group). If N is greater than Largest
+Object.Group, the Start Location is `{0, 0}`. There is no End Group - the
+subscription is open ended. The subscription opens a fill fetch stream for
+already-published objects (see {{fill-semantics}}).
 
 An endpoint that receives a filter type other than the above MUST close the
 session with `PROTOCOL_VIOLATION`.
 
-If the publisher cannot satisfy the requested Subscription Filter (see
-{{subscription-filter}}) or if the entire End Group has already been published
-it SHOULD send a REQUEST_ERROR with code `INVALID_RANGE`.  A publisher MUST
-NOT send objects from outside the requested range.
+A publisher MUST NOT send objects from outside the requested range. Every
+subscription filter type yields a satisfiable range (see {{fill-semantics}}).
+
+### Fill Semantics {#fill-semantics}
+
+Fill filter types (AbsoluteStartFill, AbsoluteRangeFill, and
+RelativeStartFill) cause the publisher to open a unidirectional stream beginning
+with a FETCH_HEADER (see {{fetch-header}}) to deliver already-published objects.
+This is called a fill fetch stream.
+
+The fill fetch stream carries objects from the fill Start Location up to and
+including the fill boundary. The fill boundary is the Largest Object, or the End
+Group if one is specified and it precedes the Largest Object. The subscriber
+learns the Largest Object from the `LARGEST_OBJECT` parameter in SUBSCRIBE_OK or
+REQUEST_UPDATE_OK. For RelativeStartFill, both sides compute the fill Start
+Location as `{Largest Object.Group - N, 0}`. For any fill filter type, if the
+Start Location is after the Largest Object the fill range is empty, the
+publisher does not open a fill fetch stream, and the subscription proceeds
+live-only. If the End Group precedes the Largest Object, only the fill fetch
+stream delivers objects and the subscription remains open (for example, for a
+later REQUEST_UPDATE).
+
+The fill fetch stream inherits the subscription's parameters, including
+subscriber priority and authorization; a FILL_PARAMETERS parameter (see
+{{fill-parameters}}) MAY override individual parameters for the fill fetch
+stream. FILL_TIMEOUT (see {{fill-timeout}}) applies to fill fetch streams in the
+same way it applies to a FETCH.
+
+The FETCH_HEADER on the fill fetch stream carries the Request ID of the message
+that initiated it: the SUBSCRIBE Request ID for the initial fill, or the
+REQUEST_UPDATE Request ID for a subsequent fill. A subscription can have
+multiple fill fetch streams open at once, each identified by its Request ID;
+opening a new fill fetch stream does not implicitly cancel any previously opened
+fill fetch stream.
+
+Because the subscription filter necessarily overlaps the fill fetch range, an
+object can be delivered both on the fill fetch stream and via subgroups or
+datagrams.
+
+#### Opening and Closing Fill Fetch Streams
+
+A publisher opens a fill fetch stream when it processes a SUBSCRIBE or
+REQUEST_UPDATE that specifies a fill filter type (a SUBSCRIPTION_FILTER with a
+fill filter type) while Forward State is 1.
+
+- A fill filter specified while Forward State is 0 opens no fill fetch stream.
+  Transitioning to Forward State 1 without re-specifying the fill filter does
+  not open one either.
+- A REQUEST_UPDATE that does not specify a fill filter type, or that specifies
+  one whose Start Location does not add objects beyond an earlier fill (for
+  example, one that only changes the End Group), does not open a new fill fetch
+  stream.
+- When Forward State transitions to 0, or the subscription is cancelled with
+  STOP_SENDING on the bidi stream, the publisher MUST reset any open fill fetch
+  streams.
+
+The publisher signals that the fill is complete by closing the stream with a FIN
+once all objects up to and including the fill boundary have been
+delivered. Because there is no REQUEST_ERROR associated with a fill fetch
+stream, the publisher signals a fill failure by resetting the stream; it can
+open a fill fetch stream and reset it immediately after the FETCH_HEADER if
+necessary. A subscriber can cancel a fill fetch stream independently by sending
+STOP_SENDING on it. Resetting or cancelling a fill fetch stream, by either
+endpoint, does not affect the subscription, which continues to deliver objects
+using subscribe subgroups and datagrams.
 
 ### Joining an Ongoing Track
 
@@ -1541,9 +1606,18 @@ is a join point, so in order for a subscriber to join a Track, it needs to
 request an existing Group or wait for a future Group.  Different applications
 will have different approaches for when to begin a new Group.
 
-To join a Track at a past Group, the subscriber sends a SUBSCRIBE, PUBLISH_OK or
-REQUEST_UPDATE with Forward State 1 followed by a Joining FETCH (see
-{{joining-fetches}}) for the intended start Group, which can be relative.
+To join a Track immediately, the subscriber sends a SUBSCRIBE with Filter Type
+`Largest Object`. Delivery begins with the next Object and can begin mid-group.
+
+To join a Track at the current Group, the subscriber sends a SUBSCRIBE with
+Filter Type `RelativeStartFill` and Relative Previous N = 0, which fills the
+current Group from its start.
+
+To join a Track at a past Group, the subscriber sends a SUBSCRIBE with a fill
+filter type (AbsoluteStartFill, AbsoluteRangeFill, or RelativeStartFill). The
+publisher fills already-published objects and delivers newer objects live (see
+{{fill-semantics}}).
+
 To join a Track at the next Group, the subscriber sends a SUBSCRIBE with
 Filter Type `Next Group Start`.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,63 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+REQUEST_UPDATE on the resume subscription's request stream.  This enables track
+switching (e.g., ABR quality changes, alternate camera angles, or meeting
+participants) without tearing down subscriptions or issuing standalone fetches.
+
+On receiving a message containing SWITCH_FROM, the publisher:
+
+1. Validates that Switch From Request ID identifies an existing subscription
+   and is not the same as the resume subscription's Request ID. If not,
+   responds with REQUEST_ERROR `INVALID_SWITCH`. The resume subscription
+   MUST be transitioning from Forward State 0 to 1 as part of the switch; if
```
sharmafb: I think you might want to rephrase this as "The resume subscription MUST either be a new subscription with a Forward State of 1 or be transitioning from Forward State 0 to 1 as part of the switch". The existing sentence applies for REQUEST_UPDATE, but not for new subscriptions.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,63 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   it already has Forward State 1, or the message carrying SWITCH_FROM has
+   FORWARD = 0, the publisher responds with REQUEST_ERROR `INVALID_SWITCH`.
```
sharmafb: I think you might want to update the wording under the "Updating Subscriptions" section of the spec, which says: "When a REQUEST_UPDATE is unsuccessful, the publisher MUST also terminate the subscription by sending a PUBLISH_DONE with error code UPDATE_FAILED". This kind of contradicts that section.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,63 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   Start Location's Group (the Start Group; see {{subscription-filters}}),
+   computed from the resume track at the time the request is received, while
+   continuing to deliver objects on the suspend subscription.  When GROUP_ORDER
+   is Descending and Start Group is before the Current Group, the publisher
+   waits for the first object in the largest fill group instead.
+
+4. Stops delivery on the suspend subscription:
+
+   * Mode Hard (0x0): sets Forward State 0 on the suspend subscription. If
+     Publish Done is 1, the publisher also sends PUBLISH_DONE.
```
sharmafb: Which status code should be used in the PUBLISH_DONE message? How about we define a new status code SWITCHING_AWAY?

sharmafb (commented):
Flushing some initial comments.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,63 @@ Since a relay can start delivering FETCH Objects from cache before determining
 the result of the request, some Objects could be received even if the FETCH
 results in error.
 
+## Track Switching {#track-switching}
+
+A subscriber can atomically stop delivery on one subscription (the suspend
+subscription) and start delivery on another (the resume subscription) by
```
ianswett: nit: resume is potentially confusing, because sometimes it is a new subscription.

Maybe 'join' or 'joining' or 'starting' subscription?

draft-ietf-moq-transport.md
ianswett: Do you mean MUST, or just that the SWITCH_FROM causes the forward state to change?

Also, from a design perspective I think it'd be ideal for the publisher to send a REQUEST_UPDATE with the new end of the filter.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,63 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+On receiving a message containing SWITCH_FROM, the publisher:
+
+1. Validates that Switch From Request ID identifies an existing subscription
+   and is not the same as the resume subscription's Request ID. If not,
+   responds with REQUEST_ERROR `INVALID_SWITCH`. The resume subscription
+   MUST be transitioning from Forward State 0 to 1 as part of the switch; if
+   it already has Forward State 1, or the message carrying SWITCH_FROM has
+   FORWARD = 0, the publisher responds with REQUEST_ERROR `INVALID_SWITCH`.
+
+2. Sets resume to Forward State 1 and applies the SUBSCRIPTION_FILTER. This
```
```suggestion
2. Sets the resume Subscription to Forward State 1 and applies the SUBSCRIPTION_FILTER. This
```

draft-ietf-moq-transport.md
ianswett: Does this change in Forward state get communicated via REQUEST_UPDATE?

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,63 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   is Descending and Start Group is before the Current Group, the publisher
+   waits for the first object in the largest fill group instead.
+
+4. Stops delivery on the suspend subscription:
+
+   * Mode Hard (0x0): sets Forward State 0 on the suspend subscription. If
+     Publish Done is 1, the publisher also sends PUBLISH_DONE.
+
+   In all modes, the publisher also resets any outstanding suspend data for
+   groups greater than or equal to the Start Group; objects already in flight
```
ianswett: This is the Start Group specified in the new subscription?

```suggestion
   In all modes, the publisher also resets any outstanding data streams for
   groups greater than or equal to the Start Group; objects already in flight
```

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,63 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   If Publish Done is 0, the suspend subscription remains established.
+
+5. Begins delivery of resume subscription from Start Group. If the
+   SUBSCRIPTION_FILTER is a fill filter type, opens a fill fetch stream using
+   the resume subscription's Request ID (see {{fill-semantics}}), if necessary.
+
+6. Responds with SUBSCRIBE_OK or REQUEST_OK as appropriate, including
+   LARGEST_OBJECT if the SUBSCRIPTION_FILTER is a fill filter type.
+
+If the publisher times out waiting to be ready to publish an object from the
```
ianswett: to be ready?

draft-ietf-moq-transport.md
```diff
@@ -2595,6 +2652,37 @@ close the session with `PROTOCOL_VIOLATION`.
...
+  Publish Done (1),
+  Reserved (7),
+}
+~~~
+
+* Switch From Request ID: The Request ID of the subscription to suspend.
+
+* Mode: A vi64 enum selecting how the suspend subscription is stopped. The
+  following mode is defined: Hard (0x0). An endpoint that receives a Mode value
+  that is not defined MUST close the session with `PROTOCOL_VIOLATION`. See
```
```suggestion
  that is not defined or negotiated MUST close the session with `PROTOCOL_VIOLATION`. See
```

draft-ietf-moq-transport.md
```diff
@@ -2595,6 +2652,37 @@ close the session with `PROTOCOL_VIOLATION`.
...
+The SWITCH_FROM parameter (Parameter Type 0x24) consists of a varint Switch
+From Request ID, a varint Mode, and a single byte of flags. It MAY appear in a
+SUBSCRIBE or REQUEST_UPDATE (for a subscription) message.
+
+~~~
+SWITCH_FROM {
+  Switch From Request ID (vi64),
+  Mode (vi64),
+  Publish Done (1),
+  Reserved (7),
```
ianswett: Should we call it 'Reserved bits' or 'Reserved flags' to be clearer?

draft-ietf-moq-transport.md
```diff
@@ -3096,6 +3184,13 @@ MALFORMED_TRACK:
 : In response to a FETCH, a relay publisher detected the track was
 malformed (see {{malformed-tracks}}).
 
+INVALID_SWITCH:
+: In response to a SUBSCRIBE or REQUEST_UPDATE carrying SWITCH_FROM, the track
+switch cannot be performed: the Switch From Request ID does not identify an
+existing subscription, is the same as the resume subscription's Request ID, the
+resume subscription is not transitioning from Forward State 0 to 1, or the
```
ianswett: These last two cases seem identical to me?

draft-ietf-moq-transport.md
afrind: The subscription in SWITCH_FROM has to start in 0 and end in 1.  We could just say Forward is ignored if it accompanies a SWITCH_FROM, and the resume subscription is immediately paused if it's in Forward=1 when SWITCH_FROM is received.  Is that better?

> Also, from a design perspective I think it'd be ideal for the publisher to send a REQUEST_UPDATE with the new end of the filter.

The new _end_ of the _pause subscription's_ filter?  It's a little odd, since that's not really a request update, it's a "state update" (eg it does not solicit an OK or ERROR, or actually update anything).  If we really need that, we should add a new message (top-N can use this also when it toggles Forward?).

draft-ietf-moq-transport.md
afrind: See Ian's point above, we could also remove this as an error condition.

draft-ietf-moq-transport.md
afrind: > Does this change in Forward state get communicated via REQUEST_UPDATE?

Not presently. See above.

draft-ietf-moq-transport.md
afrind: > This is the Start Group specified in the new subscription?

The computed start group in the "resume" subscription, yes.

afrind: Also I realized that this cancellation behavior may be incorrect for Hard switch -- we should cancel _all_ outstanding streams?

draft-ietf-moq-transport.md
```suggestion
If the publisher times out waiting to publish an object from the
```

draft-ietf-moq-transport.md
afrind: RFC 9000 uses Reserved Bits

```suggestion
  Reserved Bits (7),
```

draft-ietf-moq-transport.md
afrind: Isn't "or negotiated" implied by 

> Extensions can define new Message types, new Parameters, new Properties, or new framing for Streams and Datagrams.

Or maybe that sentence needs "Enum values"?

draft-ietf-moq-transport.md
afrind: They are not exactly the same.  But maybe it's moot if we decide to just ignore Forward?

draft-ietf-moq-transport.md
ianswett: It is, but somewhere we should make it clear this can be extended.

afrind: See #1799

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+1. Validates that Switch From Request ID identifies an existing subscription
+   and is not the same as the activating subscription's Request ID. If not,
+   responds with REQUEST_ERROR `INVALID_SWITCH`. The activating subscription
+   handles Forward State transitions as part of the switch; if the message
+   carrying SWITCH_FROM has a FORWARD parameter, the publisher responds with
+   REQUEST_ERROR `INVALID_SWITCH`.  Note it is not an error if the pausing
+   subscription is in Forward State 0.
+
+2. Sets the activating subscription to Forward State 1 and applies the
+   SUBSCRIPTION_FILTER. This ensures objects after the Current Group are not
```
fluffy: at this point in the text, it's not clear to me what SUBSCRIPTION_FILTER we are talking about. I think we just need bit more clarification text, perhaps above this numbered list, to explain the filter part.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   missed.
+
+3. Waits until it is ready to publish an object from the SUBSCRIPTION_FILTER's
+   Start Location's Group (the Start Group; see {{subscription-filters}}),
+   computed from the activating track at the time the request is received, while
+   continuing to deliver objects on the suspending subscription.  When GROUP_ORDER
+   is Descending and Start Group is before the Current Group, the publisher
+   waits for the first object in the largest fill group instead.
```
fluffy: Could we just remove the support for Descending here? Any use case that needs it ? Do it in later PR ? My issue is thinking about how implementations are going to buffer waiting for largetst fill group sounds painful. I don't think about Descending much so I may have this wrong but it does not seem like this case even makes any sense in hard mode. Can it be in a different mode.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+
+6. Responds with SUBSCRIBE_OK or REQUEST_OK as appropriate, including
+   LARGEST_OBJECT if the SUBSCRIPTION_FILTER is a fill filter type.
+
+If the publisher times out waiting to publish an object from the
+Start Group on the activating track, it MUST respond with REQUEST_ERROR `TIMEOUT`.
+
+### Relay Handling of SWITCH_FROM {#relay-switch-from}
+
+When a relay receives a message containing SWITCH_FROM, it MUST NOT forward the
```
fluffy: I think we need a SHOULD NOT instead of MUST NOT. We do the same thing for other things where a local wifi relay that is one the downstream side of a congested WAN link probably would want to forward this upstream.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+## Track Switching {#track-switching}
+
+A subscriber can atomically stop delivery on one subscription (the suspending
+subscription) and start delivery on another (the activating subscription) by
+including the SWITCH_FROM parameter ({{switch-from}}) in a SUBSCRIBE or
+REQUEST_UPDATE on the activating subscription's request stream.  This enables
+track switching (e.g., ABR quality changes, alternate camera angles, or meeting
+participants) without tearing down subscriptions or issuing standalone fetches.
+
+On receiving a message containing SWITCH_FROM, the publisher:
```
fluffy, line 1679: Are all publishers required to support this or just relays ?  I think it would be best to have it like it is that everything needs to support this but I can imaging lots of end user clients that are original publishers uploading video might not want to support this. Also - I am happy if we merge this PR without solving this and just raise this as an open issue that can be dealt with later.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
 results in error.
 
+## Track Switching {#track-switching}
+
+A subscriber can atomically stop delivery on one subscription (the suspending
+subscription) and start delivery on another (the activating subscription) by
+including the SWITCH_FROM parameter ({{switch-from}}) in a SUBSCRIBE or
+REQUEST_UPDATE on the activating subscription's request stream.  This enables
+track switching (e.g., ABR quality changes, alternate camera angles, or meeting
+participants) without tearing down subscriptions or issuing standalone fetches.
```
fluffy: I think it would be best to define the IANA table for types of switches in PR and have just one entry in it of hard but that sets up the structure for later PRs.

fluffy (approved):
There is some small clarification changes I think we should make but I'm also fine with merging as is and doing the rest later. I think the basic mechanism proposed here is fine and glad to see it merged in to main draft.  If we do merge without fixing the things I brought up, I would ask we opened issue for the ones the editors agree with.

fluffy:
This should be rebased onto main

draft-ietf-moq-transport.md
afrind: The nature of Descending is that it makes one's head hurt.  Anyone whose head is not hurting just hasn't tried to implement it yet.

I don't the the utility of Descending mode is different in Hard vs Soft switch, since it's about the activating subscription "Do I have the first requested object ready yet".  I don't think it's any more complicated than Descending support in general -- if you have a cache, then you likely have the object (it's from a more recent group) and if you don't you can issue a Descending FETCH upstream.  There shouldn't be a "buffer incoming objects" case?

draft-ietf-moq-transport.md
```suggestion
When a relay receives a message containing SWITCH_FROM, it SHOULD NOT forward the
```

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   and is not the same as the activating subscription's Request ID. If not,
+   responds with REQUEST_ERROR `INVALID_SWITCH`. The activating subscription
+   handles Forward State transitions as part of the switch; if the message
+   carrying SWITCH_FROM has a FORWARD parameter, the publisher responds with
+   REQUEST_ERROR `INVALID_SWITCH`.  Note it is not an error if the pausing
+   subscription is in Forward State 0.
+
+2. Sets the activating subscription to Forward State 1 and applies the
+   SUBSCRIPTION_FILTER. This ensures objects after the Current Group are not
+   missed.
```
```suggestion
2. Sets the activating subscription to Forward State 1 and applies the
   current SUBSCRIPTION_FILTER. This ensures objects after the Current Group
   are not missed.
```

draft-ietf-moq-transport.md
afrind: Discussed on 7/6 - Descending needs to be supported.

draft-ietf-moq-transport.md
afrind: Discussed 7/6 - no one raised a significant objection to all publishers having to support this.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
 the result of the request, some Objects could be received even if the FETCH
 results in error.
 
+## Track Switching {#track-switching}
+
+A subscriber can atomically stop delivery on one subscription (the suspending
+subscription) and start delivery on another (the activating subscription) by
```
vasilvv: nit: add italics for "suspending"/"activating" terms.

Actually, shouldn't it be "suspended" and not "suspending"?  "Suspending" implies it's the subscription that has caused suspension, not the one being suspended.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+3. Waits until it is ready to publish an object from the SUBSCRIPTION_FILTER's
+   Start Location's Group (the Start Group; see {{subscription-filters}}),
+   computed from the activating track at the time the request is received, while
+   continuing to deliver objects on the suspending subscription.  When GROUP_ORDER
+   is Descending and Start Group is before the Current Group, the publisher
+   waits for the first object in the largest fill group instead.
+
+4. Stops delivery on the suspending subscription:
+
+   * Mode Hard (0x0): sets Forward State 0 on the suspending subscription. If
```
vasilvv, line 1704: "Immediate"?  "Hard" seems like a vague term in absence of other modes.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   and is not the same as the activating subscription's Request ID. If not,
+   responds with REQUEST_ERROR `INVALID_SWITCH`. The activating subscription
+   handles Forward State transitions as part of the switch; if the message
+   carrying SWITCH_FROM has a FORWARD parameter, the publisher responds with
+   REQUEST_ERROR `INVALID_SWITCH`.  Note it is not an error if the pausing
+   subscription is in Forward State 0.
+
+2. Sets the activating subscription to Forward State 1 and applies the
+   current SUBSCRIPTION_FILTER. This ensures objects after the Current Group
+   are not missed.
```
vasilvv: Since this already depending on fill-fetch, maybe "applies the current filter" -> "creates fill-fetch streams as appopriate"?  Otherwise it sounds ambiguous as to what we're actually doing.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   subscription; objects already in flight can still be received by the subscriber.
+
+   If Publish Done is 0, the suspending subscription remains established.
+
+5. Begins delivery of activating subscription from Start Group. If the
+   SUBSCRIPTION_FILTER is a fill filter type, opens a fill fetch stream using
+   the resuming subscription's Request ID (see {{fill-semantics}}), if necessary.
+
+6. Responds with SUBSCRIBE_OK or REQUEST_OK as appropriate, including
+   LARGEST_OBJECT if the SUBSCRIPTION_FILTER is a fill filter type.
```
vasilvv, line 1718: I assume that for modes that do not immediately Forward=0 the old subscription, steps 5 and 6 would come before step 4.

vasilvv (commented):
This seems to overall look good; some comments, plus a note that this depends on text in #1763.

draft-ietf-moq-transport.md
afrind: Italics is a good call. At the beginning of the switch, the suspending stream is active and the activating stream is not, so I think "-ing" is correct.

draft-ietf-moq-transport.md
```diff
@@ -2595,6 +2654,38 @@ close the session with `PROTOCOL_VIOLATION`.
 If omitted from SUBSCRIBE, the publisher's preference from
 the Track is used. If omitted from FETCH, the receiver uses Ascending (0x1).
 
+### SWITCH_FROM Parameter {#switch-from}
+
+The SWITCH_FROM parameter (Parameter Type 0x24) consists of a varint Switch
```
DenizUgur, line 2661: Shouldn't the parameter type be an odd value?

afrind: Parameters no longer use the even/odd distinction for varint/length prefixed.  Each parameter has a bespoke format that must be known in order to parse it.

DenizUgur: I missed that change, my bad. I was still thinking of message parameters as KVPs.

afrind:
@suhasHere : we are blocked on your review.  Please review ASAP.

draft-ietf-moq-transport.md
suhasHere: we need terminology on what is suspending subscriptions

draft-ietf-moq-transport.md
suhasHere: I do think, it will be helpful for a publisher to reject SWITHC_FROM even after saying yes initially. We need to define a way to do it. I can think of upstream goaway subscription doesn't has the quality to be switched_to anymore.

draft-ietf-moq-transport.md
suhasHere: I also think, this doesn't handle the case on what happens to in-progress fill stream or when the new subscription cannot be switched because fill failed . It would be good to enumerate the edge/error cases.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+participants) without tearing down subscriptions or issuing standalone fetches.
+
+On receiving a message containing SWITCH_FROM, the publisher:
+
+1. Validates that Switch From Request ID identifies an existing subscription
+   and is not the same as the activating subscription's Request ID. If not,
+   responds with REQUEST_ERROR `INVALID_SWITCH`. The activating subscription
+   handles Forward State transitions as part of the switch; if the message
+   carrying SWITCH_FROM has a FORWARD parameter, the publisher responds with
+   REQUEST_ERROR `INVALID_SWITCH`.  Note it is not an error if the pausing
```
suhasHere: s/pausing/suspending ?

draft-ietf-moq-transport.md
```diff
@@ -3096,6 +3187,12 @@ MALFORMED_TRACK:
 : In response to a FETCH, a relay publisher detected the track was
 malformed (see {{malformed-tracks}}).
 
+INVALID_SWITCH:
+: In response to a SUBSCRIBE or REQUEST_UPDATE carrying SWITCH_FROM, the track
+switch cannot be performed: the Switch From Request ID does not identify an
+existing subscription, is the same as the activating subscription's Request ID, or the
+request contained the FORWARD parameter (see {{track-switching}}).
```
suhasHere: nit: grammar seems correction here

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+track switching (e.g., ABR quality changes, alternate camera angles, or meeting
+participants) without tearing down subscriptions or issuing standalone fetches.
+
+On receiving a message containing SWITCH_FROM, the publisher:
+
+1. Validates that Switch From Request ID identifies an existing subscription
+   and is not the same as the activating subscription's Request ID. If not,
+   responds with REQUEST_ERROR `INVALID_SWITCH`. The activating subscription
+   handles Forward State transitions as part of the switch; if the message
+   carrying SWITCH_FROM has a FORWARD parameter, the publisher responds with
```
gwendalsimon, line 1685: "The activating subscription handles Forward State transitions as part of the switch" is vague — "handles" does not say what actually happens. Please state it directly.

I am also not sure FORWARD should be rejected at all. The suspending subscription is allowed to already be in Forward State 0, so why is Forward State 0 not allowed on the activating subscription? With `FORWARD=0` and Publish Done=1, the switch would stop the old track and leave a paused subscription on the new track — a silent switch from one track to another, without starting delivery yet. That seems like a valid case. Can we clarify why FORWARD is forbidden, and whether `FORWARD=0` on the activating subscription should be allowed?

draft-ietf-moq-transport.md
gwendalsimon: I agree this is ambiguous. "current" is not defined, and "applies" does not say what happens in practice. If the intent is to create fill-fetch streams, it would be better to say it explicitly, as you suggest.

Separately, the sentence "This ensures objects after the Current Group are not missed" is hard to understand. First, what is the Current Group? It is written here as a defined term, but it is not defined in the draft. Second, "missed" is unclear — do we mean the objects are not sent to the subscriber?

draft-ietf-moq-transport.md
```diff
@@ -2595,6 +2654,38 @@ close the session with `PROTOCOL_VIOLATION`.
...
+  are registered in the IANA table "MOQT SWITCH_FROM Modes"
+  ({{iana-switch-from-modes}}); the following mode is defined: Hard (0x0). An
+  endpoint that receives a Mode value that is not defined MUST close the session
+  with `PROTOCOL_VIOLATION`. See {{track-switching}}.
+
+* Publish Done: If 1, the publisher sends PUBLISH_DONE on the suspend
+  subscription as described in {{track-switching}}.
+
+* Reserved Bits: MUST be 0. An endpoint that receives a non-zero value MUST
+  close the session with `PROTOCOL_VIOLATION`.
```
suhasHere, line 2686: may be we can remove this field as we are defining anything .If needed, we can add this back as it is in parameter.

draft-ietf-moq-transport.md
gwendalsimon: I agree "Hard" is not clear. I am not sure "Immediate" is the best replacement either, because it describes *when* the switch happens, not what happens at the junction between the two tracks.

The defining property of this mode is that it does not align the two tracks. The suspending track is stopped as soon as the activating track is ready. Its last object can be before the activating track's first object (a gap), or after it (an overlap that the subscriber may need to drop from its buffer).

Because the essence is alignment, not timing, I would suggest "Unaligned". The second mode is already proposed in #1675 (Soft Mode); an "Unaligned" / "Aligned" pair would describe the two modes more clearly than "Hard" / "Soft".

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+4. Stops delivery on the suspending subscription:
+
+   * Mode Hard (0x0): sets Forward State 0 on the suspending subscription. If
+     Publish Done is 1, the publisher also sends PUBLISH_DONE with code
+     SUBSCRIPTION_ENDED.
+
+   In all modes, the publisher also resets any outstanding streams on the suspending
+   subscription; objects already in flight can still be received by the subscriber.
+
+   If Publish Done is 0, the suspending subscription remains established.
```
gwendalsimon, line 1711: The handling of Publish Done is split across two indentation levels, which makes step 4 hard to follow.

The "Publish Done is 1" case is nested under the Mode Hard bullet, so it reads as specific to Hard mode. The "Publish Done is 0" case is a separate paragraph at the step level. Between them there is an unrelated "In all modes" paragraph about resetting streams.

Publish Done is a property of the SWITCH_FROM parameter, not of the mode. It would be clearer to describe it once, at a single indentation level, independent of the mode, and keep the mode bullets for mode-specific behavior only (for Hard, setting Forward State 0).

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
 the result of the request, some Objects could be received even if the FETCH
 results in error.
 
+## Track Switching {#track-switching}
+
+A subscriber can atomically stop delivery on one subscription (the suspending
```
```suggestion
A subscriber can automically stop delivery on one subscription (the suspending
```

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+     SUBSCRIPTION_ENDED.
+
+   In all modes, the publisher also resets any outstanding streams on the suspending
+   subscription; objects already in flight can still be received by the subscriber.
+
+   If Publish Done is 0, the suspending subscription remains established.
+
+5. Begins delivery of activating subscription from Start Group. If the
+   SUBSCRIPTION_FILTER is a fill filter type, opens a fill fetch stream using
+   the resuming subscription's Request ID (see {{fill-semantics}}), if necessary.
```
gwendalsimon: `resuming` should be `activating`, to match the term used in the rest of this section.

```suggestion
   the activating subscription's Request ID (see {{fill-semantics}}).
```

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
 the result of the request, some Objects could be received even if the FETCH
 results in error.
 
+## Track Switching {#track-switching}
+
+A subscriber can atomically stop delivery on one subscription (the suspending
+subscription) and start delivery on another (the activating subscription) by
+including the SWITCH_FROM parameter ({{switch-from}}) in a SUBSCRIBE or
+REQUEST_UPDATE on the activating subscription's request stream.  This enables
+track switching (e.g., ABR quality changes, alternate camera angles, or meeting
```
suhasHere: This doesn'tt allow all forms of ABR .. may be we should call it Client Side ABR ?

draft-ietf-moq-transport.md
suhasHere: IIUC, we tear down the subscription when PUBLISH_DONE is set to true, isn't it ?

suhasHere: also I might be missing the context with standalone fetches here. We should expand a bit more for reader to understand ( and yes not in full details :-) )

draft-ietf-moq-transport.md
suhasHere: Agree with @gwendalsimon  and not been able to convince myself on the reasoning. May be I am missing some context here.

draft-ietf-moq-transport.md
gwendalsimon: "the largest fill group" is not defined. The fill concepts this builds on (fill range, Largest Object, Group Order) are defined in #1673, but "fill group" is not defined there either.

In Descending order the highest group is delivered first (#1673), so I assume "largest fill group" means the highest group in the fill range, that is, the group of the Largest Object. It would be clearer to express this using the terms from #1673 rather than introduce a new term here.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+
+1. Validates that Switch From Request ID identifies an existing subscription
+   and is not the same as the activating subscription's Request ID. If not,
+   responds with REQUEST_ERROR `INVALID_SWITCH`. The activating subscription
+   handles Forward State transitions as part of the switch; if the message
+   carrying SWITCH_FROM has a FORWARD parameter, the publisher responds with
+   REQUEST_ERROR `INVALID_SWITCH`.  Note it is not an error if the pausing
+   subscription is in Forward State 0.
+
+2. Sets the activating subscription to Forward State 1 and applies the
```
suhasHere, line 1689: This is assuming we have an activating subscription already with forward = 0. A Switch is possible even for a new subscription.

An activating subscription can still have forward=0.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   current SUBSCRIPTION_FILTER. This ensures objects after the Current Group
```
suhasHere: we need to clarify on what is "current" here ? Is it the latest object filter ?

draft-ietf-moq-transport.md
suhasHere: I may be misunderstood here. I felt no one had opinion on it . Apologies if my recollection is wrong here. Unfortunately, I am not seeing the meeting minutes to confirm either.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   computed from the activating track at the time the request is received, while
+   continuing to deliver objects on the suspending subscription.  When GROUP_ORDER
+   is Descending and Start Group is before the Current Group, the publisher
+   waits for the first object in the largest fill group instead.
+
+4. Stops delivery on the suspending subscription:
+
+   * Mode Hard (0x0): sets Forward State 0 on the suspending subscription. If
+     Publish Done is 1, the publisher also sends PUBLISH_DONE with code
+     SUBSCRIPTION_ENDED.
```
suhasHere, line 1706: may be good to add a reason as SWITCHED ?

draft-ietf-moq-transport.md
suhasHere, line 1704: We need to explain when the relay decides to it. Is it at the point of first object being delivered on the activating subscription to ensure data flow has started ? or is it at the point of receiving ack for bytes sent  on the beginning of the object.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+
+   In all modes, the publisher also resets any outstanding streams on the suspending
+   subscription; objects already in flight can still be received by the subscriber.
```
suhasHere, line 1709: isn't pausing or sending publish_done as implcit call to reset any ongoing streams ?
I wonder if it is safe for resetting the stream before the current object is completed ? I suspect there will be glitch or bad end user experience ? Perhaps this is a question to @gwendalsimon

suhasHere: we also need to tell what happens to fill streams too in this case.

draft-ietf-moq-transport.md
suhasHere: resuming subscription is a new term. I think we are referring to following 4 variants 
activating , suspending, pausing , resuming . Probably defining these at the beginning might be useful in general

suhasHere: Not sure I understand "if necessary" here. Why is it optional ? I suspect, if there is a fill parameter, the relay MUST setup fill fetch for the requested range, isn't it ?

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+
+   If Publish Done is 0, the suspending subscription remains established.
+
+5. Begins delivery of activating subscription from Start Group. If the
+   SUBSCRIPTION_FILTER is a fill filter type, opens a fill fetch stream using
+   the resuming subscription's Request ID (see {{fill-semantics}}), if necessary.
+
+6. Responds with SUBSCRIBE_OK or REQUEST_OK as appropriate, including
```
suhasHere: Responds with SUBSCRIBE_OK or REQUEST_OK to who ? activating subscription ? If so, its too much of time spent to respond I believe.

draft-ietf-moq-transport.md
suhasHere: +1 needs some clarification here

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   LARGEST_OBJECT if the SUBSCRIPTION_FILTER is a fill filter type.
+
+If the publisher times out waiting to publish an object from the
+Start Group on the activating track, it MUST respond with REQUEST_ERROR `TIMEOUT`.
```
suhasHere, line 1721: This is changing the behavior on how subscriptions are defined today. Today' we don't wait for object to eb ready to publish before responding to SUBSCRIBE. We wait for a response from upstream and object may or may not come anytime soon. 

I would like to learn more on why is this is mandated differently here.

suhasHere, line 1721: may be we already have this . What timeout is this ? randevouz timeout ?

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+
+### Relay Handling of SWITCH_FROM {#relay-switch-from}
+
+When a relay receives a message containing SWITCH_FROM, it SHOULD NOT forward the
+SWITCH_FROM parameter upstream. The relay handles the switch locally, applying
```
suhasHere: SHOULD_NOT and relay handles locally are contradicting.

suhasHere (commented):
Thanks for the PR. I think this needs more work on 
 - specifying the behavior on handoff
 - interaction with fills between both the subscriptions
 - unifying terminology 
 - edge cases
 - forward state clarification

draft-ietf-moq-transport.md
afrind: automically is not a word?  atomic (as a single operation) is what I want.

draft-ietf-moq-transport.md
afrind: > A subscriber can atomically stop delivery on one subscription (the suspending subscription) and start delivery on another (the activating subscription) 

I'm not sure what else you want me to say, I find it clear enough.

draft-ietf-moq-transport.md
```suggestion
subscriber track switching (e.g., ABR quality changes, alternate camera angles, or meetings
```

draft-ietf-moq-transport.md
```suggestion
A subscriber can atomically stop delivery on one subscription (the _suspending
subscription_) and start delivery on another, possibly new subscription (the _activating subscription_) by
```

draft-ietf-moq-transport.md
```suggestion
participants) without coordinating multiple messages.
```

draft-ietf-moq-transport.md
afrind: > I do think, it will be helpful for a publisher to reject SWITHC_FROM even after saying yes initially.

There isn't really a "saying yes initially" -- the _final_ step in the algorithm is sending REQUEST_OK (or SUBSCRIBE_OK). If the switch becomes impossible before it completes, send REQUEST_ERROR. The quality being "switched to" is the track this parameter is operating on -- so ERROR/PUBLISH_DONE is a fine way to signal that.

draft-ietf-moq-transport.md
afrind: > The suspending subscription is allowed to already be in Forward State 0, so why is Forward State 0 not allowed on the activating subscription? 

The action of a SWITCH_FROM is not necessary if the activating subscription is going to forward = 0 -- there is nothing to coordinate, and logic about "when there's an object to send on the activating subscription" doesn't make any sense.  If folks strongly feel there is a use case we can allow it, but I think it will add a bunch of text for a corner case that was not discussed as a use case.

draft-ietf-moq-transport.md
```suggestion
   REQUEST_ERROR `INVALID_SWITCH`.  Note it is not an error if the suspending
```

draft-ietf-moq-transport.md
afrind: > This is assuming we have an activating subscription already with forward = 0. A Switch is possible even for a new subscription.

If the SWITCH_FROM is accompanying a SUBSCRIBE, the same logic applies.  I added "possibly new subscription" to the opening paragraph to further establish that.

> An activating subscription can still have forward=0.

This PR does not allow that currently - SWITCH_FROM always sets the new subscription to Forward=1.

draft-ietf-moq-transport.md
afrind: The "current" subscription filter is either the SUBSCRIPTION_FILTER param in the same message as SWITCH_FROM, or the most recently sent SUBSCRIPTION_FILTER.

```suggestion
   the most recent SUBSCRIPTION_FILTER parameter for this subscription (which
   can be in the same message carrying SWITCH_FROM). This ensures objects
   in Groups greater than or equal to the Largest Object's Group
```

draft-ietf-moq-transport.md
```suggestion
   are not missed during the transition.
```

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+2. Sets the activating subscription to Forward State 1 and applies the
+   current SUBSCRIPTION_FILTER. This ensures objects after the Current Group
+   are not missed.
+
+3. Waits until it is ready to publish an object from the SUBSCRIPTION_FILTER's
+   Start Location's Group (the Start Group; see {{subscription-filters}}),
+   computed from the activating track at the time the request is received, while
+   continuing to deliver objects on the suspending subscription.  When GROUP_ORDER
+   is Descending and Start Group is before the Current Group, the publisher
+   waits for the first object in the largest fill group instead.
```
```suggestion
   waits for the first object in the largest group in the fill range instead.
```

Updating to use 1763 "largest group in the fill range".

draft-ietf-moq-transport.md
afrind: > I may be misunderstood here. I felt no one had opinion on it . Apologies if my recollection is wrong here. Unfortunately, I am not seeing the meeting minutes to confirm either.

I rewatched the video.  No one expressed a verbal opinion besides me.  There was text in Zulip stating "Descending is required to support live".  I said I was inclined to leave it in.  No one objected.

draft-ietf-moq-transport.md
afrind: I like "Unaligned" and "Aligned". Hard is Unaligned because it doesn't require aligned group numbers between the tracks in order to be effective (though it can still work if they are).  Soft mode mostly makes sense when the tracks have aligned groups, since it uses a group number from one Track to set the end filter on another.

afrind: > We need to explain when the relay decides to it.

This is step 4 in an algorithm.  It happens as soon as Step 3 (`Waits until it is ready to publish an object ...`) is complete.

> or is it at the point of receiving ack for bytes sent on the beginning of the object.

There's nowhere else in MOQ where we require the timings of Acks so it's definitely not that.

draft-ietf-moq-transport.md
afrind: I looked at this before and decided not to add a new code at this time, but now I don't remember exactly why.

draft-ietf-moq-transport.md
afrind: > isn't pausing or sending publish_done as implcit call to reset any ongoing streams ?

That text does not appear anywhere in the draft.  PUBLISH_DONE to the contrary has explicit text about waiting for open streams to complete so cancellation is clearly not for that case.

Pausing by itself does not trigger stream resets on open streams, but the publisher has to reset them if it gets another object (since skipping publishing it would create a gap for a paused subscriber).  An implementation is free to proactively reset on pause.  In any case, this is a separate issue.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+
+4. Stops delivery on the suspending subscription:
+
+   * Mode Hard (0x0): sets Forward State 0 on the suspending subscription. If
+     Publish Done is 1, the publisher also sends PUBLISH_DONE with code
+     SUBSCRIPTION_ENDED.
+
+   In all modes, the publisher also resets any outstanding streams on the suspending
```
```suggestion
   In all modes, the publisher also resets any outstanding streams (including fill-fetch streams) on the suspending
```

draft-ietf-moq-transport.md
afrind: The multiple nesting levels are an artifact of splitting this PR.  The not-indented section applies to both Hard and Soft modes.

> It would be clearer to describe it once, at a single indentation level, independent of the mode, and keep the mode bullets for mode-specific behavior only 

That's sort of how it's written now?  The 0 behavior is common but the 1 behavior is not.  I'm inclined to leave this more-or-less as is, and deal with the flow after Soft also lands (or if it fails to reach consensus/moves to an extension).

draft-ietf-moq-transport.md
afrind: I switched from pausing/resuming to suspending/activating and missed a couple.  The terminology is introduced at the beginning of the section.

draft-ietf-moq-transport.md
```suggestion
6. Responds on the activating subscription's control stream with SUBSCRIBE_OK or REQUEST_OK as appropriate, including
```

afrind: Because the response carries LARGEST_OBJECT, it cannot be sent until that value is known.  There is a case where you could send the _OK before the fill once Largest Object is known, but in part the signal is delayed because it's communicating the time-point when the suspending subscription is paused.

If we add e.g. SUBSCRIPTION_STATE_UPDATE to communicate that event, we can return the _OK once Largest Object is determined.

draft-ietf-moq-transport.md
afrind: > I assume that for modes that do not immediately Forward=0 the old subscription,

The Soft mode PR still has "4. Stops delivery on the suspending subscription" but goes on to detail that is accomplished by setting the end-filter.  So it's in the same immediacy as Hard (Forward=0).  The real questions are 1) do we want a separate control signal on the suspending subscription, and 2) can we then move step 6 earlier in either case?

draft-ietf-moq-transport.md
afrind: This is only in the context of SWITCH_FROM.  If you want uncoordinated suspend/activate - don't send SWITCH_FROM.

I mostly read your comment as wanting to get the _OK sooner, which I've responded to in the above comments.

afrind: RENDEZVOUS_TIMEOUT is how long the subscriber is willing to wait at a relay with no known publishers for a track.

This is different -- it's that the activating track didn't receive the first object in their requested range in a timely way, and hence the old track was never suspended.  The publisher may be known, but the fill-fetch took too long, or e.g. NextGroup was used and the publisher didn't mint a new group in a reasonable time.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,65 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+If the publisher times out waiting to publish an object from the
+Start Group on the activating track, it MUST respond with REQUEST_ERROR `TIMEOUT`.
+
+### Relay Handling of SWITCH_FROM {#relay-switch-from}
+
+When a relay receives a message containing SWITCH_FROM, it SHOULD NOT forward the
+SWITCH_FROM parameter upstream. The relay handles the switch locally, applying
+the start group computation from {{track-switching}} using its locally observed
+state for the activating track and servicing any fill fetch stream from its cache
+and upstream sources as described in {{fill-semantics}}.
```
```suggestion
Relays ordinarily handle the switch locally, applying
the start group computation from {{track-switching}} using locally observed
state for the activating track and servicing any fill fetch stream from cache
and upstream sources as described in {{fill-semantics}}.  When a relay performs
the switch operation, it MUST NOT forward the SWITCH_FROM parameter upstream.
```

draft-ietf-moq-transport.md
afrind: It's needed for byte alignment

draft-ietf-moq-transport.md
```suggestion
: In response to a SUBSCRIBE or REQUEST_UPDATE carrying SWITCH_FROM, the track
switch cannot be performed (see {{track-switching}}).
```

draft-ietf-moq-transport.md
gwendalsimon: Yes, let's see after Soft (unaligned) mode lands

draft-ietf-moq-transport.md
gwendalsimon, line 1679: The publisher actions here are numbered 1 to 6, but the text does not say whether each step must wait for the previous one to complete before starting. For Hard mode this does not matter, but for Soft mode (#1675) the ordering is important: the suspending track must finish draining (step 4) before the activating track starts delivery (step 5).

Could we state whether these steps are strictly sequential? See the related comment on #1675: https://github.com/moq-wg/moq-transport/pull/1675#discussion_r3597470349

afrind: I replied in 1675 ([comment](https://github.com/moq-wg/moq-transport/pull/1675#discussion_r3608721773)).

Without improved end-of-group signaling from publisher to relay, forcing these steps to be sequential will introduce implementation specific delays.

draft-ietf-moq-transport.md
gwendalsimon: Can we have a SHOULD instead of a MUST for "the reserved bits at 0"? The reason is that it could be used for implementation-specific flags, which the Relay could use to run some optimized track transition (e.g. temporary priority management to accelerate the switch)

afrind: @gwendalsimon if someone wants to use these bits, they need to negotiate an extension first.  This is consistent with the rest of the document, where we tend to allocate a full varint for an enum with 2 or 3 fields, and say any other value is a protocol violation.  It's fine to define extensions, you just need to ensure the peer speaks it first.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,67 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   can be in the same message carrying SWITCH_FROM). This ensures objects
+   in Groups greater than or equal to the Largest Object's Group
+   are not missed during the transition.
+
+3. Waits until it is ready to publish an object from the SUBSCRIPTION_FILTER's
+   Start Location's Group (the Start Group; see {{subscription-filters}}),
+   computed from the activating track at the time the request is received, while
+   continuing to deliver objects on the suspending subscription.  When GROUP_ORDER
+   is Descending and Start Group is before the Current Group, the publisher
+   waits for the first object in the largest group in the fill range instead.
```
martinduke, line 1700: Maybe it's easier to switch on any object in the fill range?

wilaw: In the case where the payload is media and keyframes are located at group boundaries, wouldn't switching on 'any object' give you an un-decodeable object sequence? Since groups are defined agnostically as join points to tracks, shouldn't all switches occur at group boundaries?

draft-ietf-moq-transport.md
fluffy: let me put a very weak vote for Immediate but don't really care.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,67 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+
+5. Begins delivery of activating subscription from Start Group. If the
+   SUBSCRIPTION_FILTER is a fill filter type, opens a fill fetch stream using
+   the activating subscription's Request ID (see {{fill-semantics}}).
+
+6. Responds on the activating subscription's control stream with SUBSCRIBE_OK or REQUEST_OK as appropriate, including
+   LARGEST_OBJECT if the SUBSCRIPTION_FILTER is a fill filter type.
+
+If the publisher times out waiting to publish an object from the
+Start Group on the activating track, it MUST respond with REQUEST_ERROR `TIMEOUT`.
```
fluffy, line 1721: This does not seem part of this PR. I would asume this happens the same as it would if forwarding is changes from 0 to 1.

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,67 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+
+### Relay Handling of SWITCH_FROM {#relay-switch-from}
+
+Relays ordinarily handle the switch locally, applying
+the start group computation from {{track-switching}} using locally observed
+state for the activating track and servicing any fill fetch stream from cache
+and upstream sources as described in {{fill-semantics}}.  When a relay performs
+the switch operation, it MUST NOT forward the SWITCH_FROM parameter upstream.
```
fluffy, line 1729: So the stub relays will need to forward this upstream. I think this MUST NOT needs to be a SHOULD NOT with an explanation around that

draft-ietf-moq-transport.md
```diff
@@ -1667,6 +1667,67 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   LARGEST_OBJECT if the SUBSCRIPTION_FILTER is a fill filter type.
+
+If the publisher times out waiting to publish an object from the
+Start Group on the activating track, it MUST respond with REQUEST_ERROR `TIMEOUT`.
+
+### Relay Handling of SWITCH_FROM {#relay-switch-from}
+
+Relays ordinarily handle the switch locally, applying
+the start group computation from {{track-switching}} using locally observed
+state for the activating track and servicing any fill fetch stream from cache
```
fluffy, line 1727: This needs clear advice how it works for relays with no cache.

draft-ietf-moq-transport.md
```diff
@@ -3096,6 +3189,10 @@ MALFORMED_TRACK:
 : In response to a FETCH, a relay publisher detected the track was
 malformed (see {{malformed-tracks}}).
 
+INVALID_SWITCH:
+: In response to a SUBSCRIBE or REQUEST_UPDATE carrying SWITCH_FROM, the track
+switch cannot be performed (see {{track-switching}}).
```
fluffy, line 3194: how do we get the extra information of why the switch could not be done

fluffy (changes requested):
This looks good at broad level but I think a few small things needed.

draft-ietf-moq-transport.md
```diff
@@ -1867,6 +1867,78 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   that the stream delivers: the Group of the fill range's Start Location, or
+   the largest Group in the fill range when GROUP_ORDER inside FILL_PARAMETERS
+   is Descending.  Otherwise the Start Group is the Group of the Start Location
+   of the activating subscription's LOCATION_FILTER (see {{location-filters}}).
+
+5. Stops delivery on the suspending subscription:
+
+   * Mode Hard (0x0): sets Forward State 0 on the suspending subscription.
+
+   In all modes, the publisher also resets any outstanding streams (including
```
fluffy: are we sure this is correct. In some modes, If I switch at group 7, I still want to get all the data in group 6 on the old stream and then get group 7 on new stream. I may have lost track of how all the bits fit together on this.

draft-ietf-moq-transport.md
```diff
@@ -1867,6 +1867,78 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   fill fetch streams) on the suspending subscription; objects already in flight
+   can still be received by the subscriber.
+
+   The publisher then sends PUBLISH_STATE_NOTIFY ({{ps-notify}}) on the
+   suspending subscription's stream, reporting the state now in effect and
+   including LARGEST_OBJECT.  If Publish Done is 1, the publisher follows it
```
fluffy: need a bit of info of where Publish Done comes from as this is first time menetioned

draft-ietf-moq-transport.md
```diff
@@ -1867,6 +1867,78 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   with PUBLISH_DONE with code SWITCHED_AWAY; otherwise
+   the suspending subscription remains established.
+
+6. Begins delivery of activating subscription from Start Group, including any
```
fluffy: seems like might have happened much sooner

draft-ietf-moq-transport.md
```diff
@@ -1867,6 +1867,78 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   fill fetch stream (see {{fill-semantics}}), which uses the activating
+   subscription's Request ID.
+
+If the publisher cannot begin delivery on the activating subscription, it MUST
```
fluffy: this seems very wrong. What are we thinking would cause this. This would need an error at least.

draft-ietf-moq-transport.md
```diff
@@ -1867,6 +1867,78 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+leave the suspending subscription unchanged.
+
+### Relay Handling of SWITCH_FROM {#relay-switch-from}
+
+Relays ordinarily handle the switch locally, applying the start group
+computation from {{track-switching}} using locally observed state for the
+activating track and servicing any fill fetch stream from cache and upstream
+sources.  When a relay performs the switch operation, it MUST NOT forward the
```
fluffy: I think the MUST NOT needs to be a SHOULD NOT here. There are cases, particularly when relay has is a stub relay and has one subscriber, where sending this upstream might be the right thing to do.

draft-ietf-moq-transport.md
```diff
@@ -2850,6 +2922,38 @@ close the session with `PROTOCOL_VIOLATION`.
 If omitted from SUBSCRIBE or SUBSCRIBE_TRACKS, the publisher's preference from
 the Track is used. If omitted from FETCH, the receiver uses Ascending (0x1).
 
+### SWITCH_FROM Parameter {#switch-from}
+
+The SWITCH_FROM parameter (Parameter Type 0x24) consists of a varint Switch
```
fluffy: Is this sentence really adding much value ? Seems like just one more thing to keep in sync

draft-ietf-moq-transport.md
```diff
@@ -1867,6 +1867,78 @@ Since a relay can start delivering FETCH Objects from cache before determining
...
+   When the activating subscription opens a fill fetch stream (see
+   {{fill-semantics}}), the Start Group is the first Group of the fill range
+   that the stream delivers: the Group of the fill range's Start Location, or
+   the largest Group in the fill range when GROUP_ORDER inside FILL_PARAMETERS
+   is Descending.  Otherwise the Start Group is the Group of the Start Location
+   of the activating subscription's LOCATION_FILTER (see {{location-filters}}).
+
+5. Stops delivery on the suspending subscription:
+
+   * Mode Hard (0x0): sets Forward State 0 on the suspending subscription.
```
fluffy: The other modes also do this at some level. I think it would be easier to differenate from other modes if it said 

"immediately sets Forward State"

fluffy (changes requested):

draft-ietf-moq-transport.md
afrind: This PR needs to be updated with the new terminology (we no longer talk about forward state, but pause/unpause).  Some of the editorial awkwardness comes from how this was split into Hard and Soft PRs.  I think the combined text read better.  As long as we agree on the functionality, we can massage the final text correctly depending on whether we are supporting one or both in the base draft.

draft-ietf-moq-transport.md
afrind: Agreed that "Soft/Aligned" mode doesn't work like this, and it's gotten out of sync.  The original text said:

>  In all modes, the publisher also resets any outstanding suspend data for
>   **groups greater than or equal to the Start Group**; objects already in flight
>    can still be received by the subscriber.

draft-ietf-moq-transport.md
afrind: I'm not sure that's true because of the "from Start Group" qualifier.

The first three steps are:

1. Validate
2. Set the activating subscription to unpaused and apply the new filter
3. Send Subscribe OK

And I expect these to be more-or-less synchronous.

4. "Waits until it is ready to publish an object from the Start Group"
5. Pause old subscription
6. Begin delivery...

What can happen is an object from a group later than Start group can be forwarded while waiting for an object from Start Group.  That said, I expect that to be kind of an edge case, and step two explains it:

> This ensures objects in Groups greater than or equal to the Largest Object's Group
>  are not missed during the transition.

Though now that I think about it, this could leave the SWITCH in a hung state if the Start group is greater than largest object, but doesn't exist.  (e.g. Largest = 10, Start=11, but the publisher only publishes even groups).

draft-ietf-moq-transport.md
afrind: > This would need an error at least.

Agree and I think all the cases do generate errors.

> What are we thinking would cause this

Timeout waiting for the new track to show up is one, it says REQUEST_ERROR+TIMEOUT.
Another might be a new subscription or subscribe update that fails, this would get REQUEST_ERROR on the activating subscription.

If you want to unilaterally pause the old subscription this is the wrong vehicle --> Just send REQUEST_UPDATE Foward=0.  SWITCH_FROM is for coordinated switching, where you want A-then-B.  I'm thinking e.g. I want to swap out one hollywood square from Alice to Bob: you don't want to stop sending Alice until you have Bob media, right?  If you spelled Bob wrong or auth failed to get that track, pausing Alice seems incorrect. But I suspect there's some other scenario where maybe you would?

draft-ietf-moq-transport.md
afrind: I agree with your use case and I don't think the text precludes it.  It says the relay MUST NOT forward when **it** is performing the SWITCH.  If the stub relay isn't switching, forwarding upstream is allowed and correct.  Or is there some scenario where you want both relays to execute the switch algorithm?

draft-ietf-moq-transport.md
afrind: There are 3 distinct error cases in the algorithm:

"From ID" doesn't refer to an active subscription
"From ID" is the same ID as "To ID"
Request contains both Forward=0 and SWITCH_FROM -- though it sounds like some folks want to make this case not an error.

If we really want 2 or 3 distinct codes, we can add them.

PR #1675, opened 2026-06-14
SWITCH_FROM Soft Mode

afrind:
This introduces a "Soft" switch mode that assumes group alignment between the suspend and resume tracks.  The switch is executed by updating the filter of the suspend subscription to end at Start Group - 1.  This allows graceful draining, preventing a possible gap between the aligned groups of the two tracks.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1700,6 +1700,17 @@ On receiving a message containing SWITCH_FROM, the publisher:
    * Mode Hard (0x0): sets Forward State 0 on the suspend subscription. If
      Publish Done is 1, the publisher also sends PUBLISH_DONE.
 
+   * Mode Soft (0x1): when Start Group is greater than zero, the publisher
+     updates the suspend subscription's End Group to Start Group - 1 and
+     delivery continues until that group is reached; otherwise the publisher
+     stops delivery on the suspend subscription immediately, as in Hard mode.
+     Outstanding fill fetch streams on the suspend subscription are not
+     cancelled. This mode is most useful when the suspend and resume tracks
+     are group-aligned (i.e., share group boundaries), ensuring a clean
+     handoff between tracks. If Publish Done is 1, the publisher sends
+     PUBLISH_DONE after sending the final Object in Start Group - 1, if
+     signaled by the publisher, or after an implementation-specific timeout.
+
    In all modes, the publisher also resets any outstanding suspend data for
    groups greater than or equal to the Start Group; objects already in flight
    can still be received by the subscriber.
@@ -2670,9 +2681,9 @@ SWITCH_FROM {
 * Switch From Request ID: The Request ID of the subscription to suspend.
 
 * Mode: A vi64 enum selecting how the suspend subscription is stopped. The
-  following mode is defined: Hard (0x0). An endpoint that receives a Mode value
-  that is not defined MUST close the session with `PROTOCOL_VIOLATION`. See
-  {{track-switching}}.
+  following modes are defined: Hard (0x0) and Soft (0x1). An endpoint that
+  receives a Mode value that is not defined MUST close the session with
+  `PROTOCOL_VIOLATION`. See {{track-switching}}.
 
 * Publish Done: If 1, the publisher sends PUBLISH_DONE on the suspend
   subscription as described in {{track-switching}}.


sharmafb (approved):

draft-ietf-moq-transport.md
```diff
@@ -1700,6 +1700,17 @@ On receiving a message containing SWITCH_FROM, the publisher:
    * Mode Hard (0x0): sets Forward State 0 on the suspend subscription. If
      Publish Done is 1, the publisher also sends PUBLISH_DONE.
 
+   * Mode Soft (0x1): when Start Group is greater than zero, the publisher
+     updates the suspend subscription's End Group to Start Group - 1 and
```
ianswett, line 1704: is this communicated in a REQUEST_UPDATE?  Because I think it should be.  If the peer makes a change to your subscription without telling you, that's bound to cause state inconsistencies.

ianswett (commented):
I'd like a REQUEST_UPDATE to be sent, but otherwise looks good.

draft-ietf-moq-transport.md
afrind: It's not, see comments on #1674.  SUBSCRIPTION_STATE_UDPATED?

draft-ietf-moq-transport.md
```diff
@@ -2670,9 +2681,9 @@ SWITCH_FROM {
 * Switch From Request ID: The Request ID of the subscription to suspend.
 
 * Mode: A vi64 enum selecting how the suspend subscription is stopped. The
```
fluffy, line 2683: Need an IANA  table for this so we have extensibility as discussed in WG meeting on this.

draft-ietf-moq-transport.md
```diff
@@ -1700,6 +1700,17 @@ On receiving a message containing SWITCH_FROM, the publisher:
...
 
+   * Mode Soft (0x1): when Start Group is greater than zero, the publisher
+     updates the suspend subscription's End Group to Start Group - 1 and
+     delivery continues until that group is reached; otherwise the publisher
+     stops delivery on the suspend subscription immediately, as in Hard mode.
+     Outstanding fill fetch streams on the suspend subscription are not
+     cancelled. This mode is most useful when the suspend and resume tracks
+     are group-aligned (i.e., share group boundaries), ensuring a clean
+     handoff between tracks. If Publish Done is 1, the publisher sends
+     PUBLISH_DONE after sending the final Object in Start Group - 1, if
```
fluffy, line 1711: This seems a bit off. Say I have multiple sub groups and I have sent the last object in the group on one stream but I still have data to be sent on other streams.  It's more like after having sent all the objects in the group minus the ones that would be filtered.

draft-ietf-moq-transport.md
```diff
@@ -1700,6 +1700,17 @@ On receiving a message containing SWITCH_FROM, the publisher:
...
+     signaled by the publisher, or after an implementation-specific timeout.
```
fluffy, line 1712: I'm very against timeouts that the impact the client and the client can not discover the time or know it happened. So a few changes to propose here. Do it in a way the client knows, or can set, the timeout. Use an error if the timeout happens instead of treating it as all OK.

fluffy (changes requested):

draft-ietf-moq-transport.md
afrind: PUBLISH_DONE can safely be sent as soon as all subgroups in the group have been opened or otherwise accounted for in the Stream Count field.  The receiver knows upon receiving PUBLISH_DONE how many subgroups are still in flight, so it gets a clean handoff. We can adjust the text here to reflect that.

draft-ietf-moq-transport.md
afrind: I’m ok with having a PUBLISH_DONE code that indicates that it was a) triggered by soft switch and b) timed out waiting to see an end of group object (which are not required or guaranteed to arrive) — SOFT_SWITCH_GROUP_END_TIMEOUT?

I also am ok with the client hinting the max time they are willing to wait, but relays can always have policy that takes precedence (no we’re not waiting 2 days for EOG). So both endpoints can state a max (in theory 2 maxes - one for pub and one for sub), but it’s an awful lot of machinery for a corner case, and this is not the only timeout in the protocol that is not advertised.

draft-ietf-moq-transport.md
afrind: Probably the table belongs in the previous PR and this just adds an entry.

draft-ietf-moq-transport.md
afrind: Thinking about this more, it's very difficult for a relay to know with certainty there are no more objects coming from upstream in a given _group_.  When upstream sends a PUBLISH_DONE, it has a StreamCount, but if the upstream has an ongoing subscription, that signal won't come.  The end-of-group marker helps, but only for cases where the relay can trivially scoreboard the object IDs in the group.  I think this is a retread of #641.

This is to say, that the relay's implementation almost certainly needs to have timers, so it likely makes sense to have the subscriber at least hint the timeout.

draft-ietf-moq-transport.md
gwendalsimon, line 1712: The text does not say what a relay does when the subscriber lags behind live.

Example: the relay is delivering Group 32 to the subscriber. Live is at Group 35 on both the suspending and activating tracks. Start Group is 34. Does the relay send 34 and 35 from the activating track immediately, without waiting for the suspending track to finish 32 and 33? That would not be the suitable behavior (two simultaneous tracks).

I think steps 4 and 5 (in #1674) must be strictly sequential. The relay should first finish the suspending track up to Start Group - 1 (33), and only then start the activating track from Start Group (34). During the drain it keeps groups 34, 35 (and any that arrive), so that at handoff it starts the activating track from 34 with a fill-fetch, not from live.

Could we reinforce this ordering, and state that the activating track resumes from Start Group and not from live? The ordering text probably belongs in #1674.

gwendalsimon:
A few terminology points across this PR.

1. This branch still uses "suspend" and "resume", but #1674 has renamed these to "suspending" and "activating".

2. On the mode names: I proposed in [#1674 (r3588983169)](https://github.com/moq-wg/moq-transport/pull/1674#discussion_r3588983169) to rename Hard to "Unaligned", because the defining property is alignment, not timing. For the same reason, "Soft" would be better named "Aligned". Another possible pair is "Continuous" (Soft) and "Discontinuous" (Hard).

draft-ietf-moq-transport.md
```diff
@@ -1700,6 +1700,17 @@ On receiving a message containing SWITCH_FROM, the publisher:
    * Mode Hard (0x0): sets Forward State 0 on the suspend subscription. If
      Publish Done is 1, the publisher also sends PUBLISH_DONE.
 
+   * Mode Soft (0x1): when Start Group is greater than zero, the publisher
+     updates the suspend subscription's End Group to Start Group - 1 and
+     delivery continues until that group is reached; otherwise the publisher
+     stops delivery on the suspend subscription immediately, as in Hard mode.
```
gwendalsimon, line 1706: I do not understand why Start Group = 0 is treated like Hard mode.

I see one use case for Start Group = 0: the subscriber does not know where it is relative to live on the suspending track (for example it is receiving an old group while the live edge is far ahead), or the suspending track is at Forward State 0. In that case Start Group = 0 means "I do not know live; take the current group of the activating track as the Start Group."

But if the subscriber asks for Soft mode, it wants the suspending track to continue up to Start Group - 1 and then switch to the activating track. That is very different from stopping immediately. So treating Start Group = 0 as Hard does not make sense to me.

afrind: I needed to set some behavior for cases when `Start Group - 1` was invalid.  When the tracks are aligned, and the activating track is asking to start _before_ 0, I assumed it meant that everything in the old track should be cancelled (e.g. - the same as hard mode).

> I see one use case for Start Group = 0: the subscriber does not know where it is relative to live on the suspending track 

I disagree that is the use case for Start Group = 0. 

With fill fetch, setting Start Group = 0 will generate a fill fetch from the beginning; this is not what you want.  The subscriber should only specify one when they know an absolute group to start from, and this can be done when behind live (eg, live=35, you can SWITCH_FROM with start=33).  If the sub doesn't know a group to start from, they likely want RelativePrevFill(N) -- note RPF(>0) can also result in a computed Start Group < 0, which has the same behavior -- or NextGroup.

draft-ietf-moq-transport.md
afrind: I see the problem you are describing, but I don't want to make steps 4 and 5 sequential in order to address it.  In MOQ, it is computationally impractical in the general case for a relay to know when it has received all objects in a group from the publisher.  So by forcing serial operation, we are necessarily introducing delay (the relay's group timeout).

Instead, I think it's better to try to handle this via prioritization.  Since suspending and activating tracks are aligned, they can be considered the same track for prioritization purposes (assuming publisher and subscriber priorities are identical) and fall back to group ordering.  Then, in your example, the relay will queue up 34, but 32 and 33 will always have priority.  There's an annoying edge if the subscriptions have opposite group orders.

This is much easier to write than to implement for most folks however.  I'm open to other suggestions.

afrind: I suppose one mechanism is to add end-of-group signaling to the subscription control channel.  eg: 

```
END_OF_GROUP { 
  Group ID (vi64),
  StreamCount (vi64)
}
```

Then relay can use this to know when to move to the other group (when it has all the streams and they have all terminated).  However, it's a bit heavy-handed to require such a message (3 MOQ bytes on the control stream every group).

draft-ietf-moq-transport.md
fluffy: Lets think about a way to do it without timers.

draft-ietf-moq-transport.md
fluffy: ah, yes, agree

fluffy (changes requested):
I think we need some discussion on if timers are needed and if we can avoid them

draft-ietf-moq-transport.md
afrind: @fluffy I have given it a significant amount of thought -- see #641, #1832 #1833.  I think some progress is possible if we include _per group_ **stream** counts (and possibly datagram counts too) and publishers send these accurately.  e.g. could be a Parameter on PUBLISH_STATE_NOTIFY sent by every publisher like 

```
GROUP_COMPLETE {
  ID (vi64),
  Stream Count (vi64)
}
```

This is not to say it eliminates timers, since the publisher could be malicious or buggy, and the provided stream count is never reached.  But in the normal case, a relay can be certain that it has received all the data sent from upstream for a given group, and send it's own GROUP_COMPLETE (which could have different counts).

I'll note this problem likely impacts SSTS also.

PR #1698, opened 2026-06-14
Clarify FIN vs RST/STOP_SENDING semantics on request streams

### LINKED ISSUE #1634: [draft-18] What semantics do FIN or RST on a request stream carry?
The spec indicates that the way to cancel a request with bidi streams is to send a STOP_SENDING:

> 4.5: "The subscriber terminates a subscription in the Pending (Subscriber) or Established states by sending STOP_SENDING."

Of course the send side needs cleanup to, so there's:

> 3.6: "Implementations SHOULD cancel requests by abruptly terminating any directions of a stream that are still open by resetting or sending 
  STOP_SENDING."

What action should a receiver take, if any, when receiving only a FIN, or only a RST_STREAM, but no STOP_SENDING?

### LINKED ISSUE #1635: [draft-18] Should a subscriber treat FIN/RST on a bidi stream as equivalent to PUBLISH_DONE?
PUBLISH_DONE conveys error information as well as stream count.

Seems like FIN without PUBLISH_DONE is entirely avoidable and could be a PROTOCOL_VIOLATION?

### LINKED ISSUE #1643: [draft-18] Is it a request or session error to FIN a response bidi stream without a control message first?
eg: publisher responds to FETCH with FIN but not FETCH_OK/ERROR?

If I read HTTP/3 correctly, this is just a stream error there.

afrind:
Add a Graceful Request Stream Closure section: a FIN ends one direction and is not a cancellation; an endpoint MUST send required messages (response message, and PUBLISH_DONE for an Established subscription) before FIN. Cancellation is via RESET_STREAM/STOP_SENDING. Remove now-redundant per-site text and fix the SUBSCRIBE_NAMESPACE/SUBSCRIBE_TRACKS and Stream Cancellation wording.

Fixes: #1634
Fixes: #1635
Fixes: #1643

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1121,7 +1121,7 @@ requester SHOULD then send a FIN on its direction, gracefully closing the stream
 An endpoint that receives a FIN before all required messages have arrived treats
 the request as failed.
 
-An endpoint SHOULD send a FIN immediately after a message when it has nothing
+An endpoint SHOULD send a FIN promptly after a message when it has nothing
 further to send on that direction and will not need to respond to a future
 REQUEST_UPDATE. A requester, with the exception of the sender of PUBLISH,
 MAY FIN immediately after sending a message if it will not send a
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1089,20 +1089,42 @@ client's behalf.
 Relays MAY defer initiating upstream subscriptions until the handshake is complete
 or reject 0-RTT entirely to mitigate resource exhaustion from replayed packets.
 
+### Graceful Request Stream Closure {#graceful-request-closure}
+
+A request stream is bidirectional and each direction is closed independently,
+either gracefully with a FIN or abruptly with RESET_STREAM.
+
+A FIN only indicates that an endpoint will send no further messages in that
+direction; it is not a request cancellation. An endpoint MUST NOT send a FIN on
+a direction of a request stream until it has sent all required messages on that
+direction for its request type. In particular, an endpoint sending a response to
+a request MUST send the corresponding response message, and the publisher of an
+`Established` subscription MUST send PUBLISH_DONE, before sending a FIN. A FIN
+sent by the responder after its response and any subsequent messages for the
+request signals that the request is complete; if it has not already done so, the
+requester SHOULD then send a FIN on its direction, gracefully closing the stream.
+An endpoint that receives a FIN before all required messages have arrived treats
+the request as failed.
+
+An endpoint SHOULD send a FIN immediately after a message when it has nothing
+further to send on that direction and will not need to respond to a future
+REQUEST_UPDATE. A requester, with the exception of the sender of PUBLISH,
+MAY FIN immediately after sending a message if it will not send a
+REQUEST_UPDATE.
+
 ### Request Cancellation and Rejection {#request-cancellation}
 
 Once a request stream has been opened, the request MAY be cancelled by either
 endpoint. Senders cancel requests if the response is no longer of interest;
 Receivers cancel requests if they are unable to or choose not to respond.
+Implementations cancel a request by abruptly terminating any directions of the
+stream that are still open, using RESET_STREAM for a direction they are sending
+and STOP_SENDING for a direction they are receiving. An endpoint that has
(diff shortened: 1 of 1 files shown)


sharmafb (approved):

draft-ietf-moq-transport.md
```diff
@@ -1104,20 +1104,42 @@ Subscriptions on the replaying client's behalf.
...
+direction; it is not a request cancellation. An endpoint MUST NOT send a FIN on
+a direction of a request stream until it has sent all required messages on that
+direction for its request type. In particular, an endpoint sending a response to
+a request MUST send the corresponding response message, and the publisher of an
+`Established` subscription MUST send PUBLISH_DONE, before sending a FIN. A FIN
+sent by the responder after its response and any subsequent messages for the
+request signals that the request is complete; if it has not already done so, the
+requester SHOULD then send a FIN on its direction, gracefully closing the stream.
+An endpoint that receives a FIN before all required messages have arrived treats
+the request as failed.
```
ianswett, line 1122: Are we leaving this ambiguous?

ianswett (approved):

draft-ietf-moq-transport.md
afrind: It's not really ambiguous here?  "request as failed" means it's a request-ish error.  Of course, any endpoint can upgrade any request error to a session error at any time for any reason.  Sending REQUEST_ERROR is not always possible based on request state machines.

vasilvv (approved):

draft-ietf-moq-transport.md
```diff
@@ -1104,20 +1104,42 @@ Subscriptions on the replaying client's behalf.
 Relays MAY defer initiating upstream subscriptions until the handshake is complete
 or reject 0-RTT entirely to mitigate resource exhaustion from replayed packets.
 
+### Graceful Request Stream Closure {#graceful-request-closure}
+
+A request stream is bidirectional and each direction is closed independently,
+either gracefully with a FIN or abruptly with RESET_STREAM.
```
fluffy, line 1110: is it worth mentioning "stop ending" or "reset at" here ?

draft-ietf-moq-transport.md
```diff
@@ -1104,20 +1104,42 @@ Subscriptions on the replaying client's behalf.
...
+direction for its request type. In particular, an endpoint sending a response to
+a request MUST send the corresponding response message, and the publisher of an
+`Established` subscription MUST send PUBLISH_DONE, before sending a FIN. A FIN
+sent by the responder after its response and any subsequent messages for the
+request signals that the request is complete; if it has not already done so, the
+requester SHOULD then send a FIN on its direction, gracefully closing the stream.
+An endpoint that receives a FIN before all required messages have arrived treats
+the request as failed.
+
+An endpoint SHOULD send a FIN immediately after a message when it has nothing
```
fluffy: I would remove "immediately" as implementations may want to clean this up on a time based sweep.

draft-ietf-moq-transport.md
```diff
@@ -1104,20 +1104,42 @@ Subscriptions on the replaying client's behalf.
...
+further to send on that direction and will not need to respond to a future
```
fluffy, line 1125: how does an relay know it will not need to respond to updates ? I agree with this but wondering how I would implement it.

draft-ietf-moq-transport.md
afrind: STOP_SENDING does not close any stream directions from the QUIC state machine's perspective.  It asks the peer to reset.

draft-ietf-moq-transport.md
afrind: How about SHOULD send a FIN _promptly_?  It's already a SHOULD so if someone wants to delay for good reasons, they can.

draft-ietf-moq-transport.md
afrind: For non-subscriptions, a relay never knows that, so it would never FIN first (it can RESET).  The exception is a subscription, after it sends PUBLISH_DONE.

suhasHere (approved):

draft-ietf-moq-transport.md
```suggestion
An endpoint SHOULD send a FIN promptly after a message when it has nothing
```

PR #1708, opened 2026-06-14
Clarify Group and Subgroup terminology and object model

### LINKED ISSUE #1665: Group do not need to be temporal ordered.
The draft defines a group as "A temporal sequence of objects."
I think we should remove temporal or replace it with "ordered"

If we take a single frame of video and break it into a single group that has a bunch of objects for each fragment, all the pixels were captured at same time so it is not temporal.

### LINKED ISSUE #1666: terminology section should define Sub Group

### LINKED ISSUE #1670: order of objects
The draft says

Objects within a Group are in ascending order by Object ID.

I don't really know what this means. What part of the draft would be less clear if we just removed this line.

### LINKED ISSUE #1671: Clarify that object ID gaps in subgroups and datagrams do not imply non-existence
The draft has the text 

> The Object is known to not exist. This state is permanent. All
> signals that an Object does not exist are authoritative.

I think the above is correct. 

The draft also has contradicting text bellow that I think is wrong/out of date. 

> Since Objects can be delivered out of order, an endpoint can receive
> an Object after it has already recorded that the Object does not
> exist (e.g., via a FETCH gap from one source and later delivery via a
> subscription). This is not a protocol error and the Track is not
> malformed.

I think we need some text along the lines that says object can arrive out of order and thus just the fact that objects 4 and 6 are received does not mean 5 does not exist.

afrind:
Define Subgroup in the Terminology section and align the Group definition with the Object Data Model, which does not describe a Group as a temporal sequence. Also remove the redundant statement that objects within a Group are in ascending Object ID order.

Clarify that object ID gaps in subgroups and datagrams do not imply non-existence.

Fixes: #1665
Fixes: #1666
Fixes: #1670
Fixes: #1671

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -223,8 +223,13 @@ Congestion:
 
 Group:
 
-: A temporal sequence of objects. A group represents a join point in a
-  track. See ({{model-group}}).
+: A collection of objects within a track. A group represents a join point
+  in a track. See ({{model-group}}).
+
+Subgroup:
+
+: A sequence of one or more objects from the same group, sent on a single
+  transport stream whenever possible. See ({{model-subgroup}}).
 
 Object:
 
@@ -495,8 +500,6 @@ responsible for the content of the object payload. This includes the
 underlying encoding, compression, any end-to-end encryption, or
 authentication. A relay MUST NOT combine, split, or otherwise modify object
 payloads.
-
-Objects within a Group are in ascending order by Object ID.
 
 From the perspective of a subscriber or a cache, an Object can be in three
 possible states:
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -511,6 +511,11 @@ possible states:
 3. The state of the Object is unknown, either because it has not yet been
    received, or it has not been produced yet.
 
+A gap in the observed Object IDs does not by itself convey any information about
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

sharmafb (approved):

draft-ietf-moq-transport.md
```diff
@@ -223,8 +223,13 @@ Congestion:
...
 
-: A temporal sequence of objects. A group represents a join point in a
-  track. See ({{model-group}}).
+: A collection of objects within a track. A group represents a join point
+  in a track. See ({{model-group}}).
+
+Subgroup:
+
+: A sequence of one or more objects from the same group, sent on a single
+  transport stream whenever possible. See ({{model-subgroup}}).
```
vasilvv, line 232: I think this should explicitly call out the fact that ojbects in a subgroup are (1) in order, (2) in order of their Object IDs.

vasilvv (approved):

draft-ietf-moq-transport.md
ianswett: That's stated clearly in {model-subgroup}

afrind: The first sentence of model subgroup is:

> A subgroup is a sequence of one or more objects from the same group ([Section 2.3](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#model-group)) in ascending order by Object ID.

So I don't think we need to say it here again.

PR #1734, opened 2026-06-14
Fix punctuation typo in Prior Object ID Gap section

fluffy:
Just fixed up such that meaning could not be confused.  I know there are lots of place we need to clean up but this one seemed like it could lead to confusion.


sharmafb (approved):

suhasHere (approved):

afrind (approved):

afrind:
I believe the other Prior Gap property has the same text, may want to update there too.

PR #1735, opened 2026-06-14
Discuss upstream filter aggregation tradeoffs

### LINKED ISSUE #1826: §9.4 justifies subscription aggregation with a widening restriction I can't find in the draft
## Summary

**The problem.** §9.4 tells relays that aggregate subscriptions they can use a Largest Object Location filter upstream, "Because MOQT restricts widening a subscription" — but §10.9.1 describes widening in both directions as ordinary behaviour, and I can't find a restriction anywhere in the draft. (see [The two passages](#the-two-passages), below)

**Consequences.** The clause is the stated *justification*, not an aside: it is why that upstream filter is presented as safe even when downstream subscribers have "disparate filters". A Largest Object filter starts after everything already published, so if widening is in fact permitted, the premise fails and §9.4's recommended strategy is left without a stated basis — while §5.1.2 explicitly permits a subscriber to ask for earlier Objects. An implementer reading §9.4 on its own gets a strategy plus a reason to trust it, and the reason doesn't hold. (see [Why it matters](#why-it-matters), below)

**Possible solutions.** Either drop the clause and state the actual reason the filter is safe — presumably that §5.1.5's Joining FETCH is how downstream subscribers obtain earlier Objects, which would also be worth a pointer from §9.4 — or, if the restriction is intended, make §10.9.1 and §10.2.9 say so, since they currently read as permitting exactly what §9.4 calls restricted. (see [Suggested resolutions](#suggested-resolutions), below)

I may simply have missed the rule; if so, that's the answer to this issue and a pointer would be welcome.

---

## The two passages

§9.4 (Subscriber Interactions):

> Relays MAY aggregate authorized subscriptions for a given Track when multiple subscribers request the same Track. [...] **Because MOQT restricts widening a subscription**, relays that aggregate upstream subscriptions can subscribe using the Largest Object Location filter to avoid churn as downstream subscribers with disparate filters subscribe and unsubscribe from a Track.

§10.9.1 (Updating Subscriptions), on the same operation:

> When a subscriber **decreases** the Start Location of the Location Filter (see Section 5.1.2), the Start Location **can be smaller than the Track's Largest Location**, similar to a new Subscription. FETCH can be used to retrieve any necessary Objects smaller than the current Largest Location.
>
> When a subscriber **increases the End Location**, the Largest Object at the publisher might already be larger than the previous End Location. This will create a gap in the subscription. The REQUEST_UPDATE_OK will include the LARGEST_OBJECT parameter, and the subscriber can issue a FETCH to retrieve the omitted Objects, if any.

Both of those are widening, and §10.9.1 describes each as ordinary — down to naming the remedy for the resulting gap.

Two further places I looked and found no restriction:

- §10.2.9 says LOCATION_FILTER "MAY appear in a SUBSCRIBE, PUBLISH_OK or REQUEST_UPDATE (for a subscription) message", with no constraint on which direction its Start Location or End Group Delta may move.
- Searching the whole document for *widen* / *narrow* returns only the §9.4 clause above and §10.9.1's paragraph about **narrowing** ("When a subscriber narrows their subscription (increase the Start Location and/or decrease the End Group), it might still receive Objects outside the new range…"), which constrains nothing about widening.

So as far as I can tell the clause is either stale — describing a restriction that was removed when REQUEST_UPDATE gained the §10.9.1 semantics — or it refers to something I've missed.

## Why it matters

A Largest Object Location filter resolves at the publisher processing the message, so the aggregated upstream subscription starts after everything already published, and the shared cache §9.4 offers as the remedy only ever holds what that subscription delivered.

If widening is permitted, the premise doesn't hold, and §9.4 gives no guidance on what an aggregating relay should do for a subscriber whose filter starts earlier — which §5.1.2 explicitly permits:

> AbsoluteStart (0x3): The filter Start Location is specified explicitly. The specified Start Location MAY be less than the Largest Object observed at the publisher.

§10.9.1 names FETCH for the analogous REQUEST_UPDATE case; §9.4 doesn't mention it. I'd rather keep that as a separate question and settle the factual point here first.

## Suggested resolutions

Whichever matches the intent:

1. **The restriction doesn't exist** — drop the clause, and if the Largest Object Location filter is still the right default for aggregation, state the actual reason (e.g. that §5.1.5's Joining FETCH is how downstream subscribers are expected to obtain earlier Objects). A pointer from §9.4 to FETCH would also close the gap with §10.9.1.
2. **The restriction is intended** — then §10.9.1's first two paragraphs and §10.2.9 need to say so, since they currently read as permitting exactly what §9.4 says is restricted.

sharmafb:
Draft-14 forbade the widening of subscriptions, but draft-18 doesn't. I'm making changes to a section of text that mentions that widening subscriptions is not allowed.

Fixes: #1826

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1980,7 +1980,8 @@ multiple subscribers request the same Track. Subscription aggregation
 allows relays to make only a single upstream subscription for the
 Track. The published content received from the upstream subscription
 request is cached and shared among the pending subscribers.
-Because MOQT restricts widening a subscription, relays that
+Because changing the range of an established subscription is not retroactive
+and can create gaps that require FETCH requests to fill, relays that
 aggregate upstream subscriptions can subscribe using the Largest Object
 filter to avoid churn as downstream subscribers with disparate filters
 subscribe and unsubscribe from a Track.
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2309,13 +2309,13 @@ allows relays to make only a single upstream subscription for the
 Track. The published content received from the upstream subscription
 request is cached and shared among the pending subscribers.
 Because changing the range of an established subscription is not retroactive
-and can create gaps that require FETCH requests to fill, relays that
-aggregate upstream subscriptions can subscribe using the Largest Object
-Location filter to avoid churn as downstream subscribers with disparate filters
-subscribe and unsubscribe from a Track. Aggregating subscriptions can also
-help relays conserve resources especially with disparate filters or
-SUBSCRIBE_TRACKS in a namespace with a large number of Tracks
-(see {{large-namespaces}}).
+and can create gaps that require a fill fetch stream (see {{fill-semantics}})
+or a FETCH to fill, relays that aggregate upstream subscriptions can subscribe
+using a Location Filter that starts at the Next Object to avoid churn as
+downstream subscribers with disparate filters subscribe and unsubscribe from
+a Track. Aggregating subscriptions can also help relays conserve resources
+especially with disparate filters or SUBSCRIBE_TRACKS in a namespace with a
+large number of Tracks (see {{large-namespaces}}).
 
 A subscriber remains subscribed to a Track at a Relay until it unsubscribes, the
 upstream publisher terminates the subscription, or the subscription expires (see
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2308,14 +2308,14 @@ multiple subscribers request the same Track. Subscription aggregation
 allows relays to make only a single upstream subscription for the
 Track. The published content received from the upstream subscription
 request is cached and shared among the pending subscribers.
-Because changing the range of an established subscription is not retroactive
-and can create gaps that require a fill fetch stream (see {{fill-semantics}})
-or a FETCH to fill, relays that aggregate upstream subscriptions can subscribe
-using a Location Filter that starts at the Next Object to avoid churn as
-downstream subscribers with disparate filters subscribe and unsubscribe from
-a Track. Aggregating subscriptions can also help relays conserve resources
-especially with disparate filters or SUBSCRIBE_TRACKS in a namespace with a
-large number of Tracks (see {{large-namespaces}}).
+Relays that aggregate subscriptions MAY combine filters from downstream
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1980,7 +1980,8 @@ multiple subscribers request the same Track. Subscription aggregation
 allows relays to make only a single upstream subscription for the
 Track. The published content received from the upstream subscription
 request is cached and shared among the pending subscribers.
-Because MOQT restricts widening a subscription, relays that
+Because changing the range of an established subscription is not retroactive
+and can create gaps that require FETCH requests to fill, relays that
```
afrind, line 1984: I'm not sure this is the reason?

sharmafb: Should I just remove that sentence altogether?

afrind: Let's hold off here until some of the other in-flight work lands.  #1673 and #1765 change what's possible here.

draft-ietf-moq-transport.md
```diff
@@ -2308,13 +2308,14 @@ multiple subscribers request the same Track. Subscription aggregation
...
-SUBSCRIBE_TRACKS in a namespace with a large number of Tracks
-(see {{large-namespaces}}).
+Because changing the range of an established subscription is not retroactive
+and can create gaps that require a fill fetch stream (see {{fill-semantics}})
+or a FETCH to fill, relays that aggregate upstream subscriptions can subscribe
+using a Location Filter that starts at the Next Object to avoid churn as
+downstream subscribers with disparate filters subscribe and unsubscribe from
+a Track. Aggregating subscriptions can also help relays conserve resources
+especially with disparate filters or SUBSCRIBE_TRACKS in a namespace with a
+large number of Tracks (see {{large-namespaces}}).
```
```suggestion
Relays that aggregate subscriptions MAY combine filters from downstream
subscribers on the upstream subscription, up to the peer's MAX_FILTER_RANGES.
If adding filters to an upstream subscription is not possible, the relay can either
remove some or all filters, or make additional subscriptions to the same Track.
Multiple subscriptions to the same Track with non-disjoint filter sets will result in
duplicate objects arriving at the relay.  Using wider upstream filters can protect the
relay from churn as subscribers with disparate filters subscribe and unsubscribe from
a Track, at the cost of receiving more objects.
```

suhasHere (approved):

vasilvv (approved):

ianswett (approved):

PR #1754, opened 2026-06-15
Drop out-of-scope deployment speculation about mutual TLS

### LINKED ISSUE #1738: between relay identification
The draft has 

> Mutual TLS is expected to be widely used for node level
> identification between relays, especially within one organization.

that seems true but out of scope for this draft.

afrind:
Remove the prediction that mutual TLS will be "widely used between relays, especially within one organization," which describes deployment expectations outside the scope of this document. Retain the substantive point that mutual TLS provides only node-level authentication, not authorization.

Fixes: #1738

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -4663,12 +4663,10 @@ to identify media content, user patterns and media stream origin.
 MOQT supports authorization via mutual TLS for node-level identification
 and token-based schemes for fine-grained access control.
 
-Mutual TLS is expected to be widely used for node level identification
-between relays, especially within one organization. However, in some
-deployments mutual TLS can also be used for end subscribers or
-original publishers. However, as only node level authentication is
-provided, what a particular identified node is allowed to do is not
-provided at TLS level.
+Mutual TLS provides node-level identification and can be used between
+relays as well as for end subscribers or original publishers. As only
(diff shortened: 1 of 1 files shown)


sharmafb (approved):

vasilvv (commented):
The term "node-level" does not seem entirely correct to me.  If I want to use mTLS for authenticating clients, it is entirely possible for me to issue a cert to every user.

draft-ietf-moq-transport.md
```diff
@@ -4677,12 +4677,10 @@ to identify media content, user patterns and media stream origin.
...
-Mutual TLS is expected to be widely used for node level identification
-between relays, especially within one organization. However, in some
-deployments mutual TLS can also be used for end subscribers or
-original publishers. However, as only node level authentication is
-provided, what a particular identified node is allowed to do is not
-provided at TLS level.
+Mutual TLS provides node-level identification and can be used between
+relays as well as for end subscribers or original publishers. As only
+node-level authentication is provided, what a particular identified node
+is allowed to do is not provided at the TLS level.
```
ianswett, line 4696: I'm not sure this is true in all cases?

ianswett (commented):
Are we better off just deleting this paragraph?

afrind:
Superseded by #1786

PR #1755, opened 2026-06-15
Recommend Secure Objects for confidentiality from relays

### LINKED ISSUE #1711: Protecting end to end content from relays
Security considerations should mention that as part of "confidential operation by default" , one way to protect content from relays is to use Secure Objets. 

(from https://datatracker.ietf.org/doc/statement-iab-statement-on-internet-confidentiality/ )

afrind:
Note in Media Security that, consistent with confidential operation by default, content that must not be exposed to relays can be protected with end-to-end object encryption such as Secure Objects, leaving relays only the metadata needed for forwarding.

I wonder if an overview draft pointing at both is better than creating a circular reference, but let's do this for now at least.

Fixes: #1711

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -4710,6 +4710,12 @@ access to media objects, object sizes and traffic patterns enable
 analysis of content. Track namespace and track name can also be
 analyzed and correlated between end subscribers by relays.
 
+Consistent with the principle of confidential operation by default,
+content that must not be exposed to relays can be kept confidential by
+applying end-to-end object encryption, for example using Secure Objects
+({{I-D.ietf-moq-secure-objects}}), so that relays retain access only to
+the metadata required for forwarding.
+
 The end-to-end media security is handled by mechanisms external to this
 specification. They need to provide source authenticity and
 confidentiality. MOQT's object model does enable both the object
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -4747,17 +4747,11 @@ Consistent with the principle of confidential operation by default,
 content that must not be exposed to relays can be kept confidential by
 applying end-to-end object encryption, for example using Secure Objects
 ({{I-D.ietf-moq-secure-objects}}), so that relays retain access only to
-the metadata required for forwarding.
-
-The end-to-end media security is handled by mechanisms external to this
-specification. They need to provide source authenticity and
-confidentiality. MOQT's object model does enable both the object
-data itself as well as Object Properties to be confidentiality and
-integrity protected. MOQT also supports Object Properties being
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -4710,6 +4710,12 @@ access to media objects, object sizes and traffic patterns enable
 analysis of content. Track namespace and track name can also be
 analyzed and correlated between end subscribers by relays.
 
+Consistent with the principle of confidential operation by default,
+content that must not be exposed to relays can be kept confidential by
+applying end-to-end object encryption, for example using Secure Objects
+({{I-D.ietf-moq-secure-objects}}), so that relays retain access only to
+the metadata required for forwarding.
```
vasilvv: This feels a bit redundant with the existing two paragraphs below?

afrind: Agreed.

ianswett: yes, we already reference secure objects below as well. This section could use some improvement, but more text is not what it needs, so I'm inclined to either do nothing or try to make some tweaks to the existing text?

fluffy: From an IESG review point of view, this text seems  better. This says here is one thing you can do, the other are a bit more like "we punt the problem". 

I think we should take a run and perhaps merging this text and next to remove redudnacy.

fluffy (approved):

suhasHere:
@afrind @ianswett @vasilvv - fixed the redundant text,

draft-ietf-moq-transport.md
```diff
@@ -4743,15 +4743,15 @@ access to media objects, object sizes and traffic patterns enable
...
- - An E2EE scheme based on SFRAME: {{I-D.ietf-moq-secure-objects}}.
+Consistent with the principle of confidential operation by default,
+content that must not be exposed to relays can be kept confidential by
+applying end-to-end object encryption, for example using Secure Objects
+({{I-D.ietf-moq-secure-objects}}), so that relays retain access only to
+the metadata required for forwarding. Such end-to-end security
+mechanisms are external to this specification and additionally provide
+source authenticity. MOQT's object model enables both the object data
+and Object Properties to be confidentiality and integrity protected, or
+integrity protected only.
```
afrind, line 4753:
```suggestion
Consistent with the principle of confidential operation by default,
publishers can apply end-to-end object encryption, for example using Secure Objects
({{I-D.ietf-moq-secure-objects}}), so that relays retain access only to
the metadata required for forwarding. Such end-to-end security
mechanisms are external to this specification and additionally provide
source authenticity. MOQT's object model enables both the object data
and Object Properties to be confidentiality and integrity protected, or
integrity protected only.
```

PR #1759, opened 2026-06-15
Recommend Immutable Properties for relay-visible, unmodifiable data

### LINKED ISSUE #1729: Immutable relay properties
In section 11.2.1.2. Object Properties

The draft should say that Object Properties that are not meant to be changed by a relay SHOULD be serialized into a Immutable Properties

afrind:
The Object Properties section did not guide authors toward Immutable Properties for data that relays should see but not change. Add a SHOULD pointing at Immutable Properties for that case.

Fixes: #1729

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3865,8 +3865,9 @@ to MOQT Object distribution. Any Object metadata never intended to be accessed
 by the transport or Relays SHOULD be serialized as part of the Object payload
 and not as an Object Property.
 
-Object Properties that are intended to be visible to relays but not modified by
-them SHOULD be serialized within Immutable Properties ({{immutable-properties}}).
+Object Properties set by the Original Publisher that are intended to be visible
+to relays, but not modified by them, SHOULD be placed in Immutable Properties
+({{immutable-properties}}), which enables end-to-end authentication schemes.
 
 Object Properties are serialized as a length in bytes followed by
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -3865,6 +3865,9 @@ to MOQT Object distribution. Any Object metadata never intended to be accessed
 by the transport or Relays SHOULD be serialized as part of the Object payload
 and not as an Object Property.
 
+Object Properties that are intended to be visible to relays but not modified by
+them SHOULD be serialized within Immutable Properties ({{immutable-properties}}).
```
```suggestion
Object Properties that are set by the original publisher and intended to be relay-visible but relay-unmodifiable SHOULD be placed in Immutable Properties ({{immutable-properties}}) and SHOULD be authenticated end-to-end.
```

suhasHere: I do feel without the latter part the sentence is not enforceable

afrind: I wrote a slightly different version.

```suggestion
Object Properties set by the Original Publisher that are intended to be visible
to relays, but not modified by them, SHOULD be placed in Immutable Properties
({{immutable-properties}}), which enables end-to-end authentication schemes.
```

suhasHere (approved):

sharmafb (approved):

vasilvv (approved):

ianswett (approved):

ianswett:
The new text looks good, but it feels like it applies to Track properties as well?

PR #1760, opened 2026-06-15
Make Object Status payload rule extensible via IANA registry

### LINKED ISSUE #1727: Extensibility of object with non zero status code
The draft has 

> Any object
> with a status code other than zero MUST have an empty payload.

I think this would be better left as extensible with IANA table

afrind:
The blanket rule that any non-zero status MUST have an empty payload prevented future status codes from carrying a payload. Add an Object Status IANA registry with a Payload column and defer the constraint to it, so extensions can define status codes that permit payloads.

Fixes: #1727

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3848,7 +3848,10 @@ All of those SHOULD be cached.
 
 Any other value SHOULD be treated as a protocol error and the session SHOULD
 be closed with a `PROTOCOL_VIOLATION` ({{session-termination}}).
-Any object with a status code other than zero MUST have an empty payload.
+An Object MUST have an empty payload unless its Object Status value is
+registered as permitting a payload in the Object Status registry
+({{iana-object-status}}). Of the values defined in this document, only Normal
+(0x0) permits a payload.
 
 #### Object Properties {#object-properties}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -5047,6 +5050,22 @@ the length field.
...
+
+This document establishes a registry for Object Status values (see
+{{object-status}}). The "Payload" column indicates whether an Object with that
+status is permitted to carry a non-empty payload.
+
+| Code | Name | Payload | Specification |
+|-----:|:-----|:--------|:--------------|
+| 0x0 | Normal | Yes | {{object-status}} |
+| 0x3 | End of Group | No | {{object-status}} |
+| 0x4 | End of Track | No | {{object-status}} |
```
suhasHere, line 5063: s/{{object-status}}/TBD/

since TBD will be replaced by RFCXXXX

suhasHere (approved):

draft-ietf-moq-transport.md
afrind: I think other tables here use the section in the current document, which is ok.

sharmafb (approved):

ianswett (approved):

vasilvv (approved):

PR #1761, opened 2026-06-15
Add IANA registry for extension/version compatibility

### LINKED ISSUE #1681: IANA table to track what extentions are required in each moq version
Draft say

> New versions of MOQT MUST specify which existing extensions can be
> used with that version. New extensions MUST specify the existing
> versions with which they can be used.

need an IANA table to do this.

afrind:
The requirement that versions list usable extensions and extensions list usable versions had no place to record the relationships. Add a MOQT Extensions registry with an Applicable Versions column and reference it from the extensibility section.

Fixes: #1681


afrind:
Per @vasilvv this is not really for IANA.  See discussion in issue, closing

PR #1762, opened 2026-06-15
Remove relay exception for reordering or dropping objects

### LINKED ISSUE #1705: OK to break things text seems bad
Draft has 

> A relay MUST NOT reorder or drop objects received on a multi-object
> stream when forwarding to subscribers, unless it has application
> specific information.

I think this should be changed to.

> A relay MUST NOT reorder or drop objects received on a multi-object
> stream when forwarding to subscribers.

Sure an extension like filter might change say more but that is not "application specific information" that is an extension.  This sort of text is what NAT vendors use to justify doing awful things to traffic.

afrind:
The "unless it has application specific information" carve-out invited relays to justify reordering or dropping objects on a multi-object stream. Extensions that legitimately change forwarding behavior do so as negotiated extensions, not as unspecified application-specific behavior.

Fixes: #1705

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1974,7 +1974,7 @@ to the priority (see {{priorities}}) and delivery timeout (see
 {{delivery-timeouts}}).
 
 A relay MUST NOT reorder or drop objects received on a multi-object stream when
-forwarding to subscribers, unless it has application specific information.
+forwarding to subscribers.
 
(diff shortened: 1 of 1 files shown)


suhasHere (approved):
I see why we added this a while ago, but given all the new proposals this has less relevant

sharmafb (approved):

ianswett (approved):

vasilvv (approved):

PR #1763, opened 2026-06-15
Remove stray relay prioritization SHOULD

### LINKED ISSUE #1716: Downgrading sections of text
when the spec says things like 

> A relay SHOULD
> prioritize sending Objects based on {{priorities}}.

Does that make every MUST in the  {{priorities}} section a SHOULD ?

I don't think we should write it this way - I think it would be better to just remove the above sentence.

afrind:
"A relay SHOULD prioritize sending Objects based on {{priorities}}" in Relay Object Handling restated normative requirements from the Priorities section as a weaker SHOULD, muddying which requirements apply. Remove it; the Priorities section is authoritative.

Fixes: #1716

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2116,8 +2116,7 @@ A relay MUST NOT modify Object fields when forwarding, except for
 Object Properties as specified in {{properties}}.
 
 A relay MUST treat the object payload as opaque.  A relay MUST NOT
-combine, split, or otherwise modify object payloads.  A relay SHOULD
-prioritize sending Objects based on {{priorities}}.
+combine, split, or otherwise modify object payloads.
 
 # Control Messages {#message}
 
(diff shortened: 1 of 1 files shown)


vasilvv (approved):

draft-ietf-moq-transport.md
```diff
@@ -2116,8 +2116,7 @@ A relay MUST NOT modify Object fields when forwarding, except for
 Object Properties as specified in {{properties}}.
 
 A relay MUST treat the object payload as opaque.  A relay MUST NOT
-combine, split, or otherwise modify object payloads.  A relay SHOULD
-prioritize sending Objects based on {{priorities}}.
+combine, split, or otherwise modify object payloads.
```
suhasHere, line 2119: thinking instead fo removing it, will it help to say "Rules defined in {{priorities}} is followed when sending objects" ?

afrind, line 2119:
```suggestion
combine, split, or otherwise modify object payloads.

Relays prioritize forwarded Objects as described in {{priorities}}.
```

sharmafb (approved):

ianswett (approved):

vasilvv (approved):

PR #1765, opened 2026-06-16
Add Range Filters

mzanaty:
Add Range Filters:
SUBGROUP_FILTER
OBJECTID_FILTER
PRIORITY_FILTER
PROPERTY_FILTER
(No Location Filter nor Top N Track Filter)
Also rename current Subscription Filters to Subscription Location Filters.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -594,7 +594,7 @@ by doing a Fetch upstream, if necessary.
 
 Applications that cannot produce Group IDs that increase with time are limited
 to the subset of MOQT that does not compare group IDs. Subscribers to these
-Tracks SHOULD NOT use range filters which span multiple Groups in FETCH or
+Tracks SHOULD NOT use Location filters which span multiple Groups in FETCH or
 SUBSCRIBE.  SUBSCRIBE and FETCH delivery use Group Order, so they could have
 an unexpected delivery order if Group IDs do not increase with time.
 
@@ -1472,27 +1472,27 @@ A REQUEST_ERROR indicates no objects will be delivered, and both endpoints can
 immediately destroy relevant state. Objects MUST NOT be sent for requests that
 end with an error.
 
-### Subscription Filters {#subscription-filters}
+### Subscription Location Filters {#subscription-filters}
 
-Subscribers can specify a filter on a subscription indicating to the publisher
+Subscribers can specify a Location filter on a subscription indicating to the publisher
 which Objects to send.  Subscriptions without a filter pass all Objects
 published or received via upstream subscriptions.
 
-All filters have a Start Location and an optional End Group Delta.  Only objects
+All Location filters have a Start Location and an optional End Group Delta.  Only objects
 published or received via a subscription having Locations greater than or
 equal to Start Location and strictly less than or equal to the End Group (when
 present) pass the filter.
 
-Some filters are defined to be relative to the `Largest Object`. The `Largest
+Some Location filters are defined to be relative to the `Largest Object`. The `Largest
 Object` is the Object with the largest Location ({{location-structure}}) in the
 Track from the perspective of the publisher processing the message. Largest
 Object updates when the first byte of an Object with a Location larger than the
 previous value is published or received through a subscription.
 
-A Subscription Filter has the following structure:
+A Subscription Location Filter has the following structure:
 
 ~~~
-Subscription Filter {
+Subscription Location Filter {
   Filter Type (vi64),
   [Start Location (Location),]
   [End Group Delta (vi64),]
@@ -1533,10 +1533,79 @@ close the session with a `PROTOCOL_VIOLATION`.
 An endpoint that receives a filter type other than the above MUST close the
 session with `PROTOCOL_VIOLATION`.
 
-If the publisher cannot satisfy the requested Subscription Filter (see
+If the publisher cannot satisfy the requested Subscription Location Filter (see
 {{subscription-filter}}) or if the entire End Group has already been published
 it SHOULD send a REQUEST_ERROR with code `INVALID_RANGE`.  A publisher MUST
 NOT send objects from outside the requested range.
+
+### Range Filters
+
+Range Filters are parameters in subscriptions or fetches that tell a publisher
+to filter tracks and objects according to subscriber criteria which are
+allowed ranges of integer values in Track and Object Properties and other
+object header fields (Subgroup ID, Object ID, and Publisher Priority).
+There are four Range Filter parameter types, 0x25-0x28, as shown below.
+
+~~~
+SUBGROUP_FILTER { Type=0x25, Length, Set, Range... }
+OBJECTID_FILTER { Type=0x26, Length, Set, Range... }
+PRIORITY_FILTER { Type=0x27, Length, Set, Range... }
+PROPERTY_FILTER { Type=0x28, Length, Set, Property Type, Range... }
+          Range { Start, End }
+~~~
+
+Each Range Filter is a sequence of Start/End (vi64) inclusive Range pairs
+prefixed with a Length (vi64) in bytes and a Set (8 bits).
+The Property Filter includes an additional prefix for Property Type (vi64).
+The final End in a sequence of Ranges can be omitted to indicate no end.
+
+Start is delta encoded from the prior Range's End or from 0
+for the first Range, and End is delta encoded from the current Range's Start.
+Any delta encoding that exceeds 2^64-1 is a `PROTOCOL_VIOLATION`.
+For example, to express ranges 3-5 and 10-15: the first Start is 3
+(delta from 0), the first End is 2 (5 minus 3), the second Start is 5
+(10 minus 5), and the second End is 5 (15 minus 10).
+
+All filter parameters with the same Set value are combined using logical
+"AND" operations, then all the resulting Sets are combined using logical
+"OR" operations.  The final result is Set=0 OR Set=1 OR ... Set=255,
+where each Set=i is the AND of filters with Set=i.
+
+These parameters MAY appear multiple times in a FETCH, SUBSCRIBE,
+SUBSCRIBE_TRACKS, PUBLISH_OK, or REQUEST_UPDATE (from a subscriber)
+message.  If the same combination of Parameter Type, Property Type
+(only in the Property Filter), and Set repeat in any message,
+an endpoint MUST close the session with a `PROTOCOL_VIOLATION`.
+
+Length can be 0 or 1 in REQUEST_UPDATE to remove a filter parameter
+for all Sets or a specified Set, respectively.  If a filter parameter
+is omitted from REQUEST_UPDATE, the value is unchanged.  If omitted
+from other messages, the default is no filter.
+
+Range Filters are only allowed if the setup option MAX_FILTER_RANGES
+is non-zero, which limits the total number of Ranges allowed concurrently
+in all Range Filter parameters for a given subscription or fetch.
+If this limit is exceeded, an endpoint MUST close the session with a
+`PROTOCOL_VIOLATION`.
+
+The Property Filter evaluates both Track and Object Properties.
+If a track has a Track Property of the specified Property Type,
+its value is used for filtering both the PUBLISH message (when used with
+SUBSCRIBE_TRACKS) and any Objects from that track that lack their own
+value for that Property Type.  If the Track Property value
+does not pass the filter, no Objects from that track are delivered.
+
+### Combining Filters
+
+All filter types are combined using logical "AND" operations
+to further restrict which tracks and objects pass all filter criteria.
+This includes all Range Filters {{range-filters}} and Subscription Location
+Filters {{subscription-filters}}, which can be evaluated in any order.
+The Forward parameter is also a type of filter.
+
+~~~
+Pass = Forward AND Location Filter AND Range Filters
+~~~
 
 ### Joining an Ongoing Track
 
@@ -1693,6 +1762,21 @@ has accepted a PUBLISH_NAMESPACE with a namespace that exactly matches the
 namespace for that track, it SHOULD only request it from the senders of those
 PUBLISH_NAMESPACE messages.
 
+## Filtering Namespaces
+
+Range Filters {{range-filters}} can be used in SUBSCRIBE_TRACKS (or
+REQUEST_UPDATE for it) to filter tracks and objects in a namespace.
+
+### Relay Resource Protection in Large Namespaces
+
+Relays SHOULD aggregate and propagate filters upstream on subscriptions,
+especially namespace subscriptions,
+to conserve and protect their resources from excessive load.  They MAY
+also impose limits on the number of publishers in a namespace, by rejecting
+or closing namespace subscriptions with the error NAMESPACE_TOO_LARGE, or
+CONFLICTING_FILTERS if too many disjoint filters are requested on downstream
+subscriptions across a large number of subscribers, or PREFIX_OVERLAP if different
+subscribers force an aggregated upstream subscription to overlap.
 
 # Priorities {#priorities}
 
@@ -2507,14 +2591,39 @@ close the session with `PROTOCOL_VIOLATION`.
 If omitted from SUBSCRIBE, the publisher's preference from
 the Track is used. If omitted from FETCH, the receiver uses Ascending (0x1).
 
-### SUBSCRIPTION FILTER Parameter {#subscription-filter}
+### SUBSCRIPTION LOCATION FILTER Parameter {#subscription-filter}
 
-The SUBSCRIPTION_FILTER parameter (Parameter Type 0x21) uses length-prefixed
+The SUBSCRIPTION_LOCATION_FILTER parameter (Parameter Type 0x21) uses length-prefixed
 encoding. It MAY appear in a SUBSCRIBE, PUBLISH_OK or REQUEST_UPDATE (for a
-subscription) message. It is a Subscription Filter (see {{subscription-filters}}).
+subscription) message. It is a Subscription Location Filter (see {{subscription-filters}}).
 
 If omitted from SUBSCRIBE or PUBLISH_OK, the subscription is
 unfiltered.  If omitted from REQUEST_UPDATE, the value is unchanged.
+
+### SUBGROUP FILTER
+
+The SUBGROUP_FILTER parameter (Type 0x25) selects objects with specified
+Ranges of Subgroup ID.  See {{range-filters}}.
+
+### OBJECTID FILTER
+
+The OBJECTID_FILTER parameter (Type 0x26) selects objects with specified
+Ranges of Object ID.  See {{range-filters}}.
+
+### PRIORITY FILTER
+
+The PRIORITY_FILTER parameter (Type 0x27) selects objects with specified
+Ranges of Publisher Priority.  See {{range-filters}}.
+If a decoded value exceeds 255, the endpoint MUST close the session with a
+`PROTOCOL_VIOLATION` since Publisher Priority is an 8-bit field.
+
+### PROPERTY FILTER
+
+The PROPERTY_FILTER parameter (Type 0x28) selects tracks or objects with
+specified Ranges of Property Value for a specified Track or Object Property
+Type which MUST be even, i.e. a single integer value
+(see {{moq-key-value-pair}}), otherwise the endpoint MUST close the
+session with a `PROTOCOL_VIOLATION`. See {{range-filters}}.
 
 ### EXPIRES Parameter {#expires}
 
@@ -2729,6 +2838,13 @@ advertising or other nonessential information. Implementations SHOULD NOT use
 the identifiers of other implementations to declare compatibility, as this
 undermines the usefulness of implementation identification for debugging.
 
+#### MAX FILTER RANGES
+
+The MAX_FILTER_RANGES option (Type 0x06) limits the peer's total number of Ranges
+(Start/End pairs) allowed concurrently in all Range filter {{range-filters}}
+parameters for a given subscription or fetch.  The default value is 0, so if not
+specified, the peer MUST NOT send any such filter parameters.  If this limit is
+exceeded, an endpoint MUST close the session with a `PROTOCOL_VIOLATION`.
 
 ## GOAWAY {#message-goaway}
 
@@ -2995,6 +3111,11 @@ INVALID_JOINING_REQUEST_ID:
 : In response to a Joining FETCH, the referenced Request ID is not an
 `Established` Subscription.
 
+CONFLICTING_FILTERS:
+: In response to SUBSCRIBE_TRACKS, the filter parameters conflict among
+too many subscribers to aggregate the subscription upstream or otherwise
+efficiently service it.
+
 ## SUBSCRIBE {#message-subscribe-req}
 
 A subscription causes the publisher to send newly published objects for a track.
@@ -3091,7 +3212,7 @@ REQUEST_UPDATE Message {
 
 ### Updating Subscriptions
 
-When a subscriber decreases the Start Location of the Subscription Filter
+When a subscriber decreases the Start Location of the Subscription Location Filter
 (see {{subscription-filters}}), the Start Location can be smaller than the Track's
 Largest Location, similar to a new Subscription. FETCH can be used to retrieve
 any necessary Objects smaller than the current Largest Location.
@@ -3266,7 +3387,7 @@ TRACK_ENDED (0x2):
 : The track is no longer being published.
 
 SUBSCRIPTION_ENDED (0x3):
-: The publisher reached the end of an associated subscription filter.
+: The publisher reached the end of an associated subscription location filter.
 
 GOING_AWAY (0x4):
 : The subscriber or publisher issued a GOAWAY message.
@@ -3438,7 +3559,7 @@ The publisher responding to a FETCH is
 responsible for delivering all available Objects in the requested range in the
 requested order (see {{group-order}}). The Objects in the response are delivered on a single
 unidirectional stream. Any gaps in the Group and Object IDs in the response
-stream indicate objects that do not exist.  For Ascending Group Order this
+stream indicate objects that do not exist unless filters were requested.  For Ascending Group Order this
 includes ranges between the first requested object and the first object in the
 stream; between objects in the stream; and between the last object in the
 stream and the Largest Group/Object indicated in FETCH_OK, so long as the fetch
@@ -4147,9 +4268,11 @@ Object in the Subgroup if at least one of the following is true:
    stream.
  * The Object was received on the same upstream Subgroup stream as the
    previously sent Object on the downstream Subgroup stream, with no other
-   Objects in between.
+   Objects in between, unless the intervening Objects did not pass the
+   subscriber's filters.
  * It determined all Object IDs between the current and previous Object IDs
-   on the Subgroup stream belong to different Subgroups or do not exist.
+   on the Subgroup stream belong to different Subgroups or do not exist,
+   or do not pass the subscriber's filters.
 
 If the relay does not know if an Object is the next Object, it MUST reset the
 Subgroup stream and open a new one to forward it.
@@ -4995,8 +5118,12 @@ Setup Options SHOULD request a provisional registration.
 | 0x0A | FILL_TIMEOUT | {{fill-timeout}} |
 | 0x10 | FORWARD | {{forward-parameter}} |
 | 0x20 | SUBSCRIBER_PRIORITY | {{subscriber-priority}} |
-| 0x21 | SUBSCRIPTION_FILTER | {{subscription-filter}} |
+| 0x21 | SUBSCRIPTION_LOCATION_FILTER | {{subscription-filter}} |
 | 0x22 | GROUP_ORDER | {{group-order}} |
+| 0x25 | SUBGROUP_FILTER | {{subgroup-filter}} |
+| 0x26 | OBJECTID_FILTER | {{objectid-filter}} |
+| 0x27 | PRIORITY_FILTER | {{priority-filter}} |
+| 0x28 | PROPERTY_FILTER | {{property-filter}} |
 | 0x32 | NEW_GROUP_REQUEST | {{new-group-request}} |
 | 0x34 | TRACK_NAMESPACE_PREFIX | {{track-namespace-prefix-param}} |
 
@@ -5119,6 +5246,7 @@ This document does not define any initial entries.
 | INVALID_JOINING_REQUEST_ID | 0x32 | {{message-request-error}} |
 | UNSUPPORTED_EXTENSION      | 0x33 | {{message-request-error}} |
 | REDIRECT                   | 0x34 | {{message-request-error}} |
+| CONFLICTING_FILTERS        | 0x35 | {{message-request-error}} |
 | Reserved for greasing      | 0x7f * N + 0x9D | {{grease}} |
 
 ### PUBLISH_DONE Codes {#iana-publish-done}
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1583,13 +1583,13 @@ If the publisher cannot satisfy the requested Location Filter (see
 it SHOULD send a REQUEST_ERROR with code `INVALID_RANGE`.  A publisher MUST
 NOT send objects from outside the requested range.
 
-### Range Filters
+### Range Filters {#range-filters}
 
-Range Filters are parameters in subscriptions or fetches that tell a publisher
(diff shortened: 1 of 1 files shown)


mzanaty:
This contains only Range Filters from PR#1518. So there will be separate PRs for Range Filters, Top N Track Filters, and Location Filters.

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,80 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+All filter parameters with the same Set value are combined using logical
+"AND" operations, then all the resulting Sets are combined using logical
+"OR" operations.  The final result is Set=0 OR Set=1 OR ... Set=255,
+where each Set=i is the AND of filters with Set=i.
+
+These parameters MAY appear multiple times in a FETCH, SUBSCRIBE,
+SUBSCRIBE_TRACKS, PUBLISH_OK, or REQUEST_UPDATE (from a subscriber)
+message.  If the same combination of Parameter Type, Property Type
+(only in the Property Filter), and Set repeat in any message,
+an endpoint MUST close the session with a `PROTOCOL_VIOLATION`.
```
suhasHere: wonder if this should be request error instead

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,80 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+object header fields (Subgroup ID, Object ID, and Publisher Priority).
+There are four Range Filter parameter types, 0x25-0x28, as shown below.
+
+~~~
+SUBGROUP_FILTER { Type=0x25, Length, Set, Range... }
+OBJECTID_FILTER { Type=0x26, Length, Set, Range... }
+PRIORITY_FILTER { Type=0x27, Length, Set, Range... }
+PROPERTY_FILTER { Type=0x28, Length, Set, Property Type, Range... }
+          Range { Start, End }
+~~~
```
suhasHere, line 1601: may be rename Set to SetId ?

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,80 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+
+All filter types are combined using logical "AND" operations
+to further restrict which tracks and objects pass all filter criteria.
+This includes all Range Filters {{range-filters}} and Subscription Location
+Filters {{subscription-filters}}, which can be evaluated in any order.
+The Forward parameter is also a type of filter.
+
+~~~
+Pass = Forward AND Location Filter AND Range Filters
+~~~
```
suhasHere, line 1661: may be add one line to explain here that forward decision in this case is evaluated to true if Pass is evaluated as truthy above,

draft-ietf-moq-transport.md
```diff
@@ -2507,15 +2591,40 @@ close the session with `PROTOCOL_VIOLATION`.
...
+If a decoded value exceeds 255, the endpoint MUST close the session with a
+`PROTOCOL_VIOLATION` since Publisher Priority is an 8-bit field.
+
+### PROPERTY FILTER
+
+The PROPERTY_FILTER parameter (Type 0x28) selects tracks or objects with
+specified Ranges of Property Value for a specified Track or Object Property
+Type which MUST be even, i.e. a single integer value
+(see {{moq-key-value-pair}}), otherwise the endpoint MUST close the
+session with a `PROTOCOL_VIOLATION`. See {{range-filters}}.
```
suhasHere: same nit, should it be request error - Malformed Filter Param ?

suhasHere (approved):
minor nits , but overall this looks well written.

vasilvv:
> Also rename current Subscription Filters to Subscription Location Filters.

Do we need this in this PR, or should we do that in the actual location filters PR?

(I'll read the rest of the text later, hopefully today)

draft-ietf-moq-transport.md
afrind: +1 for SetId

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,80 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+~~~
+
+Each Range Filter is a sequence of Start/End (vi64) inclusive Range pairs
+prefixed with a Length (vi64) in bytes and a Set (8 bits).
+The Property Filter includes an additional prefix for Property Type (vi64).
+The final End in a sequence of Ranges can be omitted to indicate no end.
+
+Start is delta encoded from the prior Range's End or from 0
+for the first Range, and End is delta encoded from the current Range's Start.
+Any delta encoding that exceeds 2^64-1 is a `PROTOCOL_VIOLATION`.
```
```suggestion
Any delta encoding that results in a value that exceeds 2^64-1 is a `PROTOCOL_VIOLATION`.
```

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,80 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+(delta from 0), the first End is 2 (5 minus 3), the second Start is 5
+(10 minus 5), and the second End is 5 (15 minus 10).
+
+All filter parameters with the same Set value are combined using logical
+"AND" operations, then all the resulting Sets are combined using logical
+"OR" operations.  The final result is Set=0 OR Set=1 OR ... Set=255,
+where each Set=i is the AND of filters with Set=i.
+
+These parameters MAY appear multiple times in a FETCH, SUBSCRIBE,
+SUBSCRIBE_TRACKS, PUBLISH_OK, or REQUEST_UPDATE (from a subscriber)
```
afrind: I assume you cannot REQUEST_UPDATE the filters in FETCH.

```suggestion
SUBSCRIBE_TRACKS, PUBLISH_OK, or REQUEST_UPDATE (on a subscription, from the subscriber only)
```

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,80 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+### Range Filters
+
+Range Filters are parameters in subscriptions or fetches that tell a publisher
+to filter tracks and objects according to subscriber criteria which are
+allowed ranges of integer values in Track and Object Properties and other
+object header fields (Subgroup ID, Object ID, and Publisher Priority).
+There are four Range Filter parameter types, 0x25-0x28, as shown below.
+
+~~~
+SUBGROUP_FILTER { Type=0x25, Length, Set, Range... }
```
afrind: Since Set and Range are optional, I think you need to enclose them in []

I'd spell this like:

```
Filter Range {
  Start (vi64),
  [End (vi64)]
}

Range Filter {
  Type (vi64),
  Length (vi64),
  [Set ID (8)],
  [Property Type (vi64)],
  [Range (..) ..]
}
```

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,80 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+in all Range Filter parameters for a given subscription or fetch.
+If this limit is exceeded, an endpoint MUST close the session with a
+`PROTOCOL_VIOLATION`.
+
+The Property Filter evaluates both Track and Object Properties.
+If a track has a Track Property of the specified Property Type,
+its value is used for filtering both the PUBLISH message (when used with
+SUBSCRIBE_TRACKS) and any Objects from that track that lack their own
+value for that Property Type.  If the Track Property value
+does not pass the filter, no Objects from that track are delivered.
```
afrind: I think this could use an example, and clarification of what happens when a property is missing from the Track Properties and/or Object Properties.

SUBSCRIBE_TRACKS Ext Foo > 10

If the PUBLISH has a Foo <= 10, the subscriber does not get the PUBLISH.
If the PUBLISH has no Foo, the subscriber DOES get the PUBLISH, but only objects with Foo > 10.  Subscriber does not get Objects with no Foo(?)

draft-ietf-moq-transport.md
```diff
@@ -1693,6 +1762,21 @@ has accepted a PUBLISH_NAMESPACE with a namespace that exactly matches the
...
 PUBLISH_NAMESPACE messages.
 
+## Filtering Namespaces
+
+Range Filters {{range-filters}} can be used in SUBSCRIBE_TRACKS (or
+REQUEST_UPDATE for it) to filter tracks and objects in a namespace.
+
+### Relay Resource Protection in Large Namespaces
+
+Relays SHOULD aggregate and propagate filters upstream on subscriptions,
```
afrind, line 1847: There's a related line in REQUEST_UPDATE that needs to be updated now that we have this capability.  Something about using Largest Object when subscribing upstream.

draft-ietf-moq-transport.md
```diff
@@ -2729,6 +2838,13 @@ advertising or other nonessential information. Implementations SHOULD NOT use
 the identifiers of other implementations to declare compatibility, as this
 undermines the usefulness of implementation identification for debugging.
 
+#### MAX FILTER RANGES
+
+The MAX_FILTER_RANGES option (Type 0x06) limits the peer's total number of Ranges
+(Start/End pairs) allowed concurrently in all Range filter {{range-filters}}
+parameters for a given subscription or fetch.  The default value is 0, so if not
+specified, the peer MUST NOT send any such filter parameters.  If this limit is
+exceeded, an endpoint MUST close the session with a `PROTOCOL_VIOLATION`.
```
afrind: Probably warrants a TOO_MANY_FILTERS error

draft-ietf-moq-transport.md
mzanaty: Agree we need to review all protocol errors and possibly lower the action severity, but following the current pattern here.

draft-ietf-moq-transport.md
mzanaty: Set seems clearer.

draft-ietf-moq-transport.md
mzanaty: "Forward decision" may be confused with the Forward flag. How about:
Objects MUST pass all filters to be sent.

draft-ietf-moq-transport.md
mzanaty: Same as earlier. Agree but in a cohesive pass to change the pattern consistently throughout the spec.

mzanaty:
Victor, We need to disambiguate Subscription Filter which sounds like any filter on a sub, hence renamed to include Location which is actually what it filters. Even if we never adopt a general Location filter, I think this change is needed to avoid confusion among various filters.

draft-ietf-moq-transport.md
```diff
@@ -1693,6 +1762,21 @@ has accepted a PUBLISH_NAMESPACE with a namespace that exactly matches the
...
+### Relay Resource Protection in Large Namespaces
+
+Relays SHOULD aggregate and propagate filters upstream on subscriptions,
+especially namespace subscriptions,
+to conserve and protect their resources from excessive load.  They MAY
+also impose limits on the number of publishers in a namespace, by rejecting
+or closing namespace subscriptions with the error NAMESPACE_TOO_LARGE, or
+CONFLICTING_FILTERS if too many disjoint filters are requested on downstream
+subscriptions across a large number of subscribers, or PREFIX_OVERLAP if different
+subscribers force an aggregated upstream subscription to overlap.
```
afrind, line 1854: This is not something a subscriber probably can control so there isn't much point in returning this error to them.

Further, relays don't normally send SUBSCRIBE_TRACKS/NS upstream, except to other _relays_ which is out of scope.  I think we can remove this?

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,80 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+allowed ranges of integer values in Track and Object Properties and other
+object header fields (Subgroup ID, Object ID, and Publisher Priority).
+There are four Range Filter parameter types, 0x25-0x28, as shown below.
+
+~~~
+SUBGROUP_FILTER { Type=0x25, Length, Set, Range... }
+OBJECTID_FILTER { Type=0x26, Length, Set, Range... }
+PRIORITY_FILTER { Type=0x27, Length, Set, Range... }
+PROPERTY_FILTER { Type=0x28, Length, Set, Property Type, Range... }
+          Range { Start, End }
```
yuyou: In the case of video scrubbing, what if I like to get some frames every n objects, instead of the 1st object in every group? e.g. in a VOD example, the player wants to get some snapshots every n seconds, regardless of the GOP size. would it be useful to have an optional "step" value provided to the Range?

e.g.
Range { Start, End [, Step] }

draft-ietf-moq-transport.md
afrind: If a track property is acting as a default for all objects, which they can override, then it can cause a subscriber to not even get the PUBLISH for the track.

eg: Track Property Foo = 5  [everything is 5 unless overriden]

draft-ietf-moq-transport.md
sharmafb: +1 for SetId :)

draft-ietf-moq-transport.md
```diff
@@ -2729,6 +2838,13 @@ advertising or other nonessential information. Implementations SHOULD NOT use
 the identifiers of other implementations to declare compatibility, as this
 undermines the usefulness of implementation identification for debugging.
 
+#### MAX FILTER RANGES
```
sharmafb, line 2941: Please include this in the IANA setup options table in section 15.4

draft-ietf-moq-transport.md
mzanaty: For video scrubbing, random access only works at key frames. Otherwise you must decode all frames prior to the desired frame to render it. So apps can't jump to arbitrary objects for scrubbing; they must jump to key frames, which are intended to be the first object in a group. If a track wants to be explicit about all its random access points, rather than relying on the first object in each group, it can mark each with a property, e.g. FrameType=1 (Random Access Point), that can be used in the property filter.

draft-ietf-moq-transport.md
mzanaty: Good catch, new table in -18.

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,80 @@ close the session with a `PROTOCOL_VIOLATION`.
...
-If the publisher cannot satisfy the requested Subscription Filter (see
+If the publisher cannot satisfy the requested Subscription Location Filter (see
 {{subscription-filter}}) or if the entire End Group has already been published
 it SHOULD send a REQUEST_ERROR with code `INVALID_RANGE`.  A publisher MUST
 NOT send objects from outside the requested range.
 
+### Range Filters
+
+Range Filters are parameters in subscriptions or fetches that tell a publisher
+to filter tracks and objects according to subscriber criteria which are
```
vasilvv: `subscriber` -> `subscriber-provided`

draft-ietf-moq-transport.md
```diff
@@ -2507,15 +2591,40 @@ close the session with `PROTOCOL_VIOLATION`.
...
+
+The PRIORITY_FILTER parameter (Type 0x27) selects objects with specified
+Ranges of Publisher Priority.  See {{range-filters}}.
+If a decoded value exceeds 255, the endpoint MUST close the session with a
+`PROTOCOL_VIOLATION` since Publisher Priority is an 8-bit field.
+
+### PROPERTY FILTER
+
+The PROPERTY_FILTER parameter (Type 0x28) selects tracks or objects with
+specified Ranges of Property Value for a specified Track or Object Property
```
vasilvv: > for a specified Track or Object Property

So is it a track property or is it an object property?  What does it mean to filter an individual track subscription by a track property?

draft-ietf-moq-transport.md
```diff
@@ -2729,6 +2838,13 @@ advertising or other nonessential information. Implementations SHOULD NOT use
 the identifiers of other implementations to declare compatibility, as this
 undermines the usefulness of implementation identification for debugging.
 
+#### MAX FILTER RANGES
+
+The MAX_FILTER_RANGES option (Type 0x06) limits the peer's total number of Ranges
+(Start/End pairs) allowed concurrently in all Range filter {{range-filters}}
```
vasilvv, line 2944: "Concurrently" sounds weird.  "In total across all Range filter"?

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,80 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+for the first Range, and End is delta encoded from the current Range's Start.
+Any delta encoding that results in a value that exceeds 2^64-1 is a `PROTOCOL_VIOLATION`.
+For example, to express ranges 3-5 and 10-15: the first Start is 3
+(delta from 0), the first End is 2 (5 minus 3), the second Start is 5
+(10 minus 5), and the second End is 5 (15 minus 10).
+
+All filter parameters with the same Set value are combined using logical
+"AND" operations, then all the resulting Sets are combined using logical
+"OR" operations.  The final result is Set=0 OR Set=1 OR ... Set=255,
+where each Set=i is the AND of filters with Set=i.
```
vasilvv: I am confused.  Last time I remember us discussing this, we only had "AND" operators, and if you needed "OR", you had to normalize.  Is there a new motivation for those?

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,80 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+from other messages, the default is no filter.
+
+Range Filters are only allowed if the setup option MAX_FILTER_RANGES
+is non-zero, which limits the total number of Ranges allowed concurrently
+in all Range Filter parameters for a given subscription or fetch.
+If this limit is exceeded, an endpoint MUST close the session with a
+`PROTOCOL_VIOLATION`.
+
+The Property Filter evaluates both Track and Object Properties.
+If a track has a Track Property of the specified Property Type,
```
vasilvv: I think we need two separate filters for Track Properties vs Object Properties.  Conflating those could lead to counterintuitive results and restrict things unnecessarily.

Also, for Range filters that filter objects, there needs to be text that those are propagated from SUBSCRIBE_TRACKS to individual subscriptions.

draft-ietf-moq-transport.md
vasilvv, line 2941: How are those calculated for SUBSCRIBE_TRACKS?  Is it total for track filters + object filters, or do they count against separate limits?  Do we need to define separate parameters?

vasilvv (commented):
This looks reasonable overall, but I think the text currently conflates filtering tracks and filtering objects in a way that is unhelpful.

Also I find the AND/OR functionality adds a lot of complexity, and I am not sure why, so let's discuss if we need that first.

draft-ietf-moq-transport.md
```diff
@@ -1472,27 +1472,27 @@ A REQUEST_ERROR indicates no objects will be delivered, and both endpoints can
 immediately destroy relevant state. Objects MUST NOT be sent for requests that
 end with an error.
 
-### Subscription Filters {#subscription-filters}
+### Subscription Location Filters {#subscription-filters}
```
ianswett: It'd be good to update the reference to be #subscription-location-filters or #location-filters

draft-ietf-moq-transport.md
```suggestion
to filter tracks (for SUBSCRIBE_TRACKS) and objects according to subscriber criteria which are
```

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,80 @@ close the session with a `PROTOCOL_VIOLATION`.
...
 it SHOULD send a REQUEST_ERROR with code `INVALID_RANGE`.  A publisher MUST
 NOT send objects from outside the requested range.
 
+### Range Filters
+
+Range Filters are parameters in subscriptions or fetches that tell a publisher
+to filter tracks and objects according to subscriber criteria which are
+allowed ranges of integer values in Track and Object Properties and other
+object header fields (Subgroup ID, Object ID, and Publisher Priority).
+There are four Range Filter parameter types, 0x25-0x28, as shown below.
```
```suggestion
Range Filters are parameters in subscriptions or fetches that tell a publisher
to filter tracks and objects according to subscriber criteria. Filters specify
allowed ranges of integer values for Track and Object Properties and other
object header fields (Subgroup ID, Object ID, and Publisher Priority). Filters on
Properties apply to Track Properties as well as Object Properties when used in SUBSCRIBE_TRACKs. There are four Range Filter parameter types,
0x25-0x28, as shown below.
```

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,80 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+
+~~~
```
ianswett, line 1594: Organization request: Put all the generic text about Filter at the top, and then talk about each Filter separately, with the framing followed by the descrption.

draft-ietf-moq-transport.md
ianswett: I agree that splitting them would make this more straightforward to reason about.  Cleverness is not a feature here.

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,80 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+"OR" operations.  The final result is Set=0 OR Set=1 OR ... Set=255,
+where each Set=i is the AND of filters with Set=i.
+
+These parameters MAY appear multiple times in a FETCH, SUBSCRIBE,
+SUBSCRIBE_TRACKS, PUBLISH_OK, or REQUEST_UPDATE (on a subscription, from the subscriber only)
+message.  If the same combination of Parameter Type, Property Type
+(only in the Property Filter), and Set repeat in any message,
+an endpoint MUST close the session with a `PROTOCOL_VIOLATION`.
+
+Length can be 0 or 1 in REQUEST_UPDATE to remove a filter parameter
```
ianswett: Why 0 or 1?  I would expect if you want to remove it it should be 0?

draft-ietf-moq-transport.md
```diff
@@ -1693,6 +1762,21 @@ has accepted a PUBLISH_NAMESPACE with a namespace that exactly matches the
 namespace for that track, it SHOULD only request it from the senders of those
 PUBLISH_NAMESPACE messages.
 
+## Filtering Namespaces
```
```suggestion
### Filtering SUBSCRIBE_TRACKS
```

draft-ietf-moq-transport.md
```diff
@@ -2995,6 +3111,11 @@ INVALID_JOINING_REQUEST_ID:
 : In response to a Joining FETCH, the referenced Request ID is not an
 `Established` Subscription.
 
+CONFLICTING_FILTERS:
```
ianswett, line 3229: That's a bummer of a response code.  Any idea of what to do in that case?

draft-ietf-moq-transport.md
```diff
@@ -2507,15 +2591,40 @@ close the session with `PROTOCOL_VIOLATION`.
...
+
+The OBJECTID_FILTER parameter (Type 0x26) selects objects with specified
+Ranges of Object ID.  See {{range-filters}}.
+
+### PRIORITY FILTER
+
+The PRIORITY_FILTER parameter (Type 0x27) selects objects with specified
+Ranges of Publisher Priority.  See {{range-filters}}.
+If a decoded value exceeds 255, the endpoint MUST close the session with a
+`PROTOCOL_VIOLATION` since Publisher Priority is an 8-bit field.
```
ianswett: I would state this up above, because the PRIORITY_FILTER is more complex than a single number.

```suggestion
```

draft-ietf-moq-transport.md
ianswett: I wouldn't specify any errors here, since this is complex enough to parse and its explained above.

draft-ietf-moq-transport.md
mzanaty: Split into Track vs Object Property Filters per meeting.

draft-ietf-moq-transport.md
mzanaty: Split into Track vs Object Property Filters per meeting.

draft-ietf-moq-transport.md
mzanaty: Clarified and agreed in meeting.

draft-ietf-moq-transport.md
mzanaty: Done.

draft-ietf-moq-transport.md
mzanaty: Separated so counts are clear.

draft-ietf-moq-transport.md
mzanaty: Changed to only allow 0 to remove or non-zero to replace. Can't replace a single Set anymore.

draft-ietf-moq-transport.md
mzanaty: Added new errpr.

draft-ietf-moq-transport.md
mzanaty: SetID

draft-ietf-moq-transport.md
mzanaty: added new error.

draft-ietf-moq-transport.md
mzanaty: added new error.

draft-ietf-moq-transport.md
mzanaty: That's an existing not new error.

draft-ietf-moq-transport.md
mzanaty: done

draft-ietf-moq-transport.md
mzanaty: changed to location filters globally

draft-ietf-moq-transport.md
mzanaty: The common generic text is first, then only aspects specific to a filter (Track Property Filter and Object Property Filter) follow at the end. Please reflow however editors want after merge.

draft-ietf-moq-transport.md
mzanaty: done

draft-ietf-moq-transport.md
mzanaty: Retry without the filter.

draft-ietf-moq-transport.md
mzanaty: It only operates on Publisher Priority as stated in the top of the Range Filters section.

draft-ietf-moq-transport.md
mzanaty: changed all errors to request_error with a new code for invalid_filter.

draft-ietf-moq-transport.md
mzanaty: Not sure what you mean. Can you suggest which text should update to what?

fluffy (approved):
I agree with a bunch of the small nit level suggestions for the PR but I think that with those, this PR is ready to merge.

draft-ietf-moq-transport.md
afrind: This is the text I'm thinking of.

> Relays MAY aggregate authorized subscriptions for a given Track when multiple subscribers request the same Track. Subscription aggregation allows relays to make only a single upstream subscription for the Track. The published content received from the upstream subscription request is cached and shared among the pending subscribers. Because MOQT restricts widening a subscription, relays that aggregate upstream subscriptions can subscribe using the Largest Object filter to avoid churn as downstream subscribers with disparate filters subscribe and unsubscribe from a Track.

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,88 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+The Track Property filter parameter MAY appear multiple times in a
+SUBSCRIBE_TRACKS message or REQUEST_UPDATE for it.
+All other filter parameters MAY appear multiple times in a FETCH, SUBSCRIBE,
+SUBSCRIBE_TRACKS, PUBLISH_OK, or REQUEST_UPDATE (on a subscription, from the subscriber only)
+message.  If the same combination of Parameter Type, SetID, and Property Type
+(only in the Track and Object Property Filters) repeat in any message,
+an endpoint MUST reject this with REQUEST_ERROR with error code INVALID_FILTER.
+
+In REQUEST_UPDATE, Length can be 0 to remove a filter parameter or non-zero
+to replace that entire filter parameter including all sets and Property Types.
```
afrind, line 1630: Let's quickly discuss whether we want to use 0 to signal removal of a single filter, or if we want the subscriber to just update the full set.  I haven't thought deeply about the data structures of filters, but it's possible this delta-removal complicates things more often than the functionality will be used.

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,88 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+with error code INVALID_FILTER.
+
+The Track Property Filter can be used in SUBSCRIBE_TRACKS to filter PUBLISH
+messages with required Track Property types and values.  PUBLISH messages
+which pass the filter will be forwarded while those which do not pass it
+will not be forwarded nor will any Objects.
+
+The Object Property Filter can be used to filter Objects with required
+Object Property types and values.  Track Properties are not evaluated
+by the Object Property Filter.
```
afrind: I think this would be more clear if we say it only examines properties in Object headers.  The properties themselves are just integers.  For unknown (app defined) properties, whether they are Track or Object is only determined by where you find them.

draft-ietf-moq-transport.md
```diff
@@ -1533,11 +1533,88 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+
+### Combining Filters
+
+All filter types are combined using logical "AND" operations
+to further restrict which tracks and objects pass all filter criteria.
+This includes all Range Filters {{range-filters}} and Location
+Filters {{location-filters}}, which can be evaluated in any order.
+The Forward parameter is also a type of filter.  Objects MUST pass all
```
afrind: nit: Should we say "forward state"?

afrind: nit: It's odd to make Objects the subject of a MUST.  It would be better to make the publisher the actor "The publisher MUST NOT forward objects unless they pass all filters" or "The publisher MUST forward only objects that pass all filters"

draft-ietf-moq-transport.md
```diff
@@ -1693,6 +1770,24 @@ has accepted a PUBLISH_NAMESPACE with a namespace that exactly matches the
 namespace for that track, it SHOULD only request it from the senders of those
 PUBLISH_NAMESPACE messages.
 
+## Filtering SUBSCRIBE_TRACKS
+
+Range Filters {{range-filters}} can be used in SUBSCRIBE_TRACKS (or
+REQUEST_UPDATE for it) to filter tracks in a namesapce using the
```
```suggestion
REQUEST_UPDATE for it) to filter tracks in a namespace using the
```

draft-ietf-moq-transport.md
```diff
@@ -1693,6 +1770,24 @@ has accepted a PUBLISH_NAMESPACE with a namespace that exactly matches the
...
+Track Property Filter, and filter objects published in those tracks
+using all other Range Filters which get applied to the resulting
+subscriptions established by SUBSCRIBE_TRACKS.
+
+### Relay Resource Protection in Large Namespaces
+
+Relays SHOULD aggregate and propagate filters upstream on subscriptions,
+especially namespace subscriptions,
+to conserve and protect their resources from excessive load.  They MAY
```
afrind, line 1849: I'll put this on a slide today just to get head-nods from the wg

afrind:
@mzanaty : can you rebase/resolve conflicts, address the final nits and the one outstanding comment about relay handling?  Then I'm ready to merge.

draft-ietf-moq-transport.md
```diff
@@ -1578,11 +1578,89 @@ close the session with a `PROTOCOL_VIOLATION`.
...
+Any delta encoding that results in a value that exceeds 2^64-1
+MUST be rejected with REQUEST_ERROR with error code INVALID_FILTER.
+For example, to express ranges 3-5 and 10-15: the first Start is 3
+(delta from 0), the first End is 2 (5 minus 3), the second Start is 5
+(10 minus 5), and the second End is 5 (15 minus 10).
+
+All filter parameters with the same SetID value are combined using logical
+"AND" operations, then all the resulting sets are combined using logical
+"OR" operations.  The final result is SetID=0 OR SetID=1 OR ... SetID=255,
+where each SetID=i is the AND of filters with SetID=i.
```
vasilvv, line 1619: > The final result is SetID=0 OR SetID=1 OR ... SetID=255

`SetID=0` is a weird way to spell "all filters with SetID 0 pass".

Also, it should clarify that if no filters are present, objects always pass.

vasilvv (approved):

mzanaty:
> @mzanaty : can you rebase/resolve conflicts, address the final nits and the one outstanding comment about relay handling? Then I'm ready to merge.

Done

afrind (approved):

ianswett (approved):

Issue #1646, opened 2026-06-03
Application-specific property codepoints conflict


sharmafb:
Section 2.5 reserves 0x38 to 0x3F and 0x3800 to 0x3FFF for application use, but section 15.8 reserves 0x78 to 0x7F.

PR #1767, opened 2026-06-16
Fix section 15.8 1-byte application-specific reservation

sharmafb:
Section 2.5 reserves 0x38 to 0x3F and 0x3800 to 0x3FFF for application use, but section 15.8 reserves 0x78 to 0x7F.

This addresses #1646

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -5033,7 +5033,7 @@ the length field.
 
 * MOQ Properties - we wish to define the following registration policies:
   - 0x00 to 0x77: Standards Action or IESG Approval (1-byte encoding)
-  - 0x78 to 0x7F: Reserved for application-specific use (1-byte encoding,
+  - 0x38 to 0x3F: Reserved for application-specific use (1-byte encoding,
     no registration permitted)
   - 0x80 to 0x37FF: Specification Required (2-byte encoding)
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -5033,7 +5033,7 @@ the length field.
 
 * MOQ Properties - we wish to define the following registration policies:
   - 0x00 to 0x77: Standards Action or IESG Approval (1-byte encoding)
-  - 0x78 to 0x7F: Reserved for application-specific use (1-byte encoding,
+  - 0x38 to 0x3F: Reserved for application-specific use (1-byte encoding,
```
suhasHere, line 5036: what would the issue with 0x78 - 0x7f

afrind: I think 78-7f is right?  It's intended to be the last 8 code points?

sharmafb: The issue is the contradiction with section 2.5 of the specification, which reserves 0x38 to 0x3F for application use. We'd likely also have to modify the preceding 0x00 to 0x77 range.

afrind: I think what happened is we set both to 38-3F when variants were max 63. When we changed to max 127 we updated one and not the other. So I think we want to change section 2.5 and leave this (including the 00-77)

PR #1770, opened 2026-06-17
Make Default Publisher Priority an updatable parameter

### LINKED ISSUE #1270: Should Publish_update be able to update default publisher priority
At some point we might want to consider if this is needed

michalhosna:
Treat `DEFAULT_PUBLISHER_PRIORITY` as compression only hop-by-hop parameter. 

Fixes: #1270

See discussiong there.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -678,27 +678,24 @@ to MOQT constraints. Such a Track is considered malformed.  Some example
 conditions that constitute a malformed track when detected by a receiver
 include:
 
-1.  An Object with a particular Subgroup ID is received, but its
-     Publisher Priority is different from that of the previous Object with the same
-     Subgroup ID.
-2. An Object is received whose Object ID is larger than the final Object in the
+1. An Object is received whose Object ID is larger than the final Object in the
    Subgroup.  The final Object in a Subgroup is the last Object received on a
    Subgroup stream before a FIN.
-3. A Subgroup is received over multiple transport streams terminated by FIN with
+2. A Subgroup is received over multiple transport streams terminated by FIN with
    different final Objects.
-4. An Object is received in a Group whose Object
+3. An Object is received in a Group whose Object
    ID is larger than the final Object in the Group.  The final Object in a Group
    is the Object with Status END_OF_GROUP, or the last Object before a FIN in a
    Subgroup which has the END_OF_GROUP bit set.  If the end of a Group is
    implicitly determined via a gap in a FETCH response, the final Object in the
    Group remains unknown.
-5. An Object is received whose Group and Object ID are larger than
+4. An Object is received whose Group and Object ID are larger than
    the final Object in the Track.  The final Object in a Track is the Object
    with Status END_OF_TRACK or the last Object sent in a FETCH whose response
    indicated End of Track.
-6. The same Object is received more than once with different Payload or
+5. The same Object is received more than once with different Payload or
    other immutable properties.
-7. An Object is received with a different Forwarding Preference than previously
+6. An Object is received with a different Forwarding Preference than previously
    observed.
 
 The above list of conditions is not considered exhaustive.
@@ -1736,9 +1733,17 @@ scheduled.
 
 `Publisher Priority` is a priority number associated with an individual
 schedulable object.  A default for the subscription is specified in the
-DEFAULT_PUBLISHER_PRIORITY Track Property ({{publisher-priority}}). Publisher
+DEFAULT_PUBLISHER_PRIORITY parameter ({{publisher-priority}}). Publisher
 priority can also be set per subgroup or datagram in the subgroup header or
-datagram (see {{data-streams}}), which overrides the default.
+datagram (see {{data-streams}}), which MUST take precedence over the default.
+
+DEFAULT_PUBLISHER_PRIORITY can be updated using REQUEST_UPDATE. Changing
+the default is best-effort and not synchronized with the data plane. When
+an object omits the Publisher Priority, a subscriber resolves it using
+the default in effect when the object was received and SHOULD cache
+the resolved value, so a later change does not alter already-cached
+objects. If conflicting Publisher Priority values are observed for the
+same object, a subscriber SHOULD use the lower priority.
 
 `Group Order` is a property of an individual subscription.  It can be either
 'Ascending' (groups with lower group ID are sent first), or 'Descending'
@@ -1916,7 +1921,7 @@ fields that can be updated are the following:
    to the constraints of the specific property.
 
 An endpoint that receives a duplicate Object with a different Forwarding
-Preference, Subgroup ID, Priority or Payload MUST treat the track as Malformed.
+Preference, Subgroup ID, or Payload MUST treat the track as Malformed.
 
 For ranges of objects that do not exist, relays MAY change the representation
 of a missing range to a semantically equivalent one.  For instance, a relay may
@@ -2500,6 +2505,17 @@ session. Lower numbers get higher priority. See {{priorities}}.
 If omitted from SUBSCRIBE, PUBLISH_OK or FETCH, the publisher uses
 the value 128.
 
+### DEFAULT PUBLISHER PRIORITY Parameter {#publisher-priority}
+
+The DEFAULT_PUBLISHER_PRIORITY parameter (Parameter Type 0x0E)
+is a uint8. It MAY appear in a PUBLISH, SUBSCRIBE_OK, FETCH_OK, or
+REQUEST_UPDATE message. It is the default Publisher Priority (see
+{{priorities}}) for Subgroups and Datagrams of the subscription that
+do not carry an explicit Publisher Priority. Lower numbers get higher
+priority.
+
+If omitted, the default Publisher Priority is 128.
+
 ### GROUP ORDER Parameter {#group-order}
 
 The GROUP_ORDER parameter (Parameter Type 0x22) is a uint8. It MAY appear in a
@@ -3068,6 +3084,8 @@ The sender of a request (SUBSCRIBE, PUBLISH, FETCH, PUBLISH_NAMESPACE,
 SUBSCRIBE_NAMESPACE, SUBSCRIBE_TRACKS) can later send a REQUEST_UPDATE on the
 same bidi stream as the request to modify it.  A subscriber can also send
 REQUEST_UPDATE to modify parameters of a subscription established with PUBLISH.
+The publisher can also send REQUEST_UPDATEto modify parameters set
+using SUBSCRIBE_OK.
 
 The receiver of a REQUEST_UPDATE MUST respond with exactly one REQUEST_OK
 or REQUEST_ERROR message indicating if the update was successful, unless it
@@ -3952,8 +3970,8 @@ which fields are present in the datagram:
 
 * The **DEFAULT_PRIORITY** bit (0x08) indicates when the Priority field is
   present. When set to 1, the Priority field is omitted and this Object inherits
-  the Publisher Priority specified in the control message that established the
-  subscription. When set to 0, the Priority field is present.
+  the subscription's current default Publisher Priority (see
+  {{publisher-priority}}). When set to 0, the Priority field is present.
 
 * The **STATUS** bit (0x20) indicates whether the datagram contains an Object
   Status or Object Payload. When set to 1, the Object Status field is present
@@ -4057,8 +4075,8 @@ in the header:
 
 * The **DEFAULT_PRIORITY** bit (0x20) indicates when the Priority field is
   present. When set to 1, the Priority field is omitted and this Subgroup
-  inherits the Publisher Priority specified in the control message that
-  established the subscription. When set to 0, the Priority field is present in
+  inherits the subscription's current default Publisher Priority (see
+  {{publisher-priority}}). When set to 0, the Priority field is present in
   the Subgroup header.
 
 * The **FIRST_OBJECT** bit (0x40) indicates that the first object in this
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1736,9 +1733,17 @@ scheduled.
 
 `Publisher Priority` is a priority number associated with an individual
 schedulable object.  A default for the subscription is specified in the
-DEFAULT_PUBLISHER_PRIORITY Track Property ({{publisher-priority}}). Publisher
+DEFAULT_PUBLISHER_PRIORITY parameter ({{publisher-priority}}). Publisher
 priority can also be set per subgroup or datagram in the subgroup header or
-datagram (see {{data-streams}}), which overrides the default.
+datagram (see {{data-streams}}), which MUST take precedence over the default.
+
+DEFAULT_PUBLISHER_PRIORITY can be updated using REQUEST_UPDATE. Changing
```
afrind: Is this better in the DEFAULT_PUBLISHER_PRIORITY section, or here in priorities?

draft-ietf-moq-transport.md
```diff
@@ -1736,9 +1733,17 @@ scheduled.
...
+the default is best-effort and not synchronized with the data plane. When
+an object omits the Publisher Priority, a subscriber resolves it using
+the default in effect when the object was received and SHOULD cache
+the resolved value, so a later change does not alter already-cached
+objects. If conflicting Publisher Priority values are observed for the
+same object, a subscriber SHOULD use the lower priority.
```
afrind: is it clearer if we say "lower priority value"?

This paragraph contains important text, I'm just not sure where it should live.

draft-ietf-moq-transport.md
```diff
@@ -2500,6 +2505,17 @@ session. Lower numbers get higher priority. See {{priorities}}.
 If omitted from SUBSCRIBE, PUBLISH_OK or FETCH, the publisher uses
 the value 128.
 
+### DEFAULT PUBLISHER PRIORITY Parameter {#publisher-priority}
+
+The DEFAULT_PUBLISHER_PRIORITY parameter (Parameter Type 0x0E)
+is a uint8. It MAY appear in a PUBLISH, SUBSCRIBE_OK, FETCH_OK, or
+REQUEST_UPDATE message. It is the default Publisher Priority (see
```
```suggestion
REQUEST_UPDATE message sent by the publisher on a subscription. It is the default Publisher Priority (see
```

draft-ietf-moq-transport.md
```diff
@@ -2500,6 +2505,17 @@ session. Lower numbers get higher priority. See {{priorities}}.
...
+{{priorities}}) for Subgroups and Datagrams of the subscription that
+do not carry an explicit Publisher Priority. Lower numbers get higher
+priority.
```
afrind: I don't think we need to say how priorities work here, there's a ref to the section.

```suggestion
do not carry an explicit Publisher Priority.
```

draft-ietf-moq-transport.md
```diff
@@ -3068,6 +3084,8 @@ The sender of a request (SUBSCRIBE, PUBLISH, FETCH, PUBLISH_NAMESPACE,
 SUBSCRIBE_NAMESPACE, SUBSCRIBE_TRACKS) can later send a REQUEST_UPDATE on the
 same bidi stream as the request to modify it.  A subscriber can also send
 REQUEST_UPDATE to modify parameters of a subscription established with PUBLISH.
+The publisher can also send REQUEST_UPDATEto modify parameters set
```
```suggestion
The publisher can also send REQUEST_UPDATE to modify parameters set
```

draft-ietf-moq-transport.md
```diff
@@ -3068,6 +3084,8 @@ The sender of a request (SUBSCRIBE, PUBLISH, FETCH, PUBLISH_NAMESPACE,
...
+using SUBSCRIBE_OK.
```
```suggestion
using PUBLISH or SUBSCRIBE_OK.
```

draft-ietf-moq-transport.md
```diff
@@ -5220,6 +5227,8 @@ Issue and pull request numbers are listed with a leading octothorp.
 * Exclude your own tracks from SUBSCRIBE_NAMESPACE (#1596)
 * Add Session-Level Tracks reserved namespace (#1562)
 * Allow coalescing REQUEST_UPDATE processing (#1540)
+* Make Default Publisher Priority an updatable parameter; stop malforming on
+  priority change (#1270)
```
afrind: We write the comprehensive release notes when we publish a new draft version.  The previous text is immutable.

```suggestion
```

draft-ietf-moq-transport.md
afrind: Do we need to explain the possible race conditions and mitigations, or let people figure it out?

afrind:
It's a little sad we moved this from param to property and back again.

Do we need text about forwarding this parameter downstream, or caching the value seen in PUBLISH/SUB_OK?

draft-ietf-moq-transport.md
michalhosna: I am not sure what "on a subscription" brings/defines here?

draft-ietf-moq-transport.md
michalhosna: This is covered above by

> sender of a request (SUBSCRIBE, PUBLISH, FETCH, PUBLISH_NAMESPACE,
SUBSCRIBE_NAMESPACE, SUBSCRIBE_TRACKS) can later send a REQUEST_UPDATE

This whole paragraph doesn't read great, but I don't have a good suggestion how to improve it.

draft-ietf-moq-transport.md
ArcherDTY: Expect more explanations about the race conditions. SHOULD is already a pretty strict requirement. Isn't there any other possibility here?

michalhosna: **Placement.** I think this section is the right place to explain all the priorities handling. And the wire section is already quite heavy on intermingled logic.
 
**Races.** I did not try to implement synchronization based on https://github.com/moq-wg/moq-transport/issues/1270#issuecomment-4719569704
 
> Based on the discussions last week, I didn't hear folks asking for synchronization - sounds like the ones who want this feature were fine with some racy edge cases.
 
Also I did not find any easy way to spell this out without much more machinery required for a feature that seems niche.
 
> Do we need to explain the possible race conditions and *mitigations*
 
The REQUEST_UPDATE may arrive too early or too late.
 
I do not think there is currently a mitigation for "too early" without some kind of location in the update message, but that's more a rewrite of the whole REQUEST_UPDATE. Or a general look at priorities.
 
I think spelling out very explicitly "Changing the default is best-effort and not synchronized with the data plane" is the right level of calling out in normative text. It warns the reader that there is raciness, but enumerating all possible behaviors seems out-of-place for normative text describing this feature.
 
Of course this will be a different story if consensus emerges that we're unhappy with the raciness and want a robust solution.

draft-ietf-moq-transport.md
afrind: It just scopes where this can occur.  The publisher can't send it in a REQUEST_UPDATE message to any other kind of request (eg PUB_NS)

draft-ietf-moq-transport.md
ianswett: We could allow the sender to put Largest Object in the REQUEST_UPDATE and say that if its present, the priority is effective as of Largest Object?

draft-ietf-moq-transport.md
ianswett: Priorities is already large, so I'd prefer to keep this in another section.

draft-ietf-moq-transport.md
```diff
@@ -3068,6 +3083,8 @@ The sender of a request (SUBSCRIBE, PUBLISH, FETCH, PUBLISH_NAMESPACE,
 SUBSCRIBE_NAMESPACE, SUBSCRIBE_TRACKS) can later send a REQUEST_UPDATE on the
 same bidi stream as the request to modify it.  A subscriber can also send
 REQUEST_UPDATE to modify parameters of a subscription established with PUBLISH.
+The publisher can also send REQUEST_UPDATE to modify parameters set
+using SUBSCRIBE_OK.
```
```suggestion
same bidi stream as the request to modify it.  A subscription can be modified by
either the Publisher or Subscriber using REQUEST_UPDATE, subject to the
constraints of the specific Parameter ({{message-params}}).
```

ianswett (commented):
I think this is a good direction, some comments.

This is the first time we're updating a subscription from the Publisher perspective, but I think Switch, DTS, and Top-N will all need a way to update Subscription state.

draft-ietf-moq-transport.md
```diff
@@ -1736,9 +1733,17 @@ scheduled.
...
 priority can also be set per subgroup or datagram in the subgroup header or
-datagram (see {{data-streams}}), which overrides the default.
+datagram (see {{data-streams}}), which MUST take precedence over the default.
+
+DEFAULT_PUBLISHER_PRIORITY can be updated using REQUEST_UPDATE. Changing
+the default is best-effort and not synchronized with the data plane. When
+an object omits the Publisher Priority, a subscriber resolves it using
+the default in effect when the object was received and SHOULD cache
+the resolved value, so a later change does not alter already-cached
+objects. If conflicting Publisher Priority values are observed for the
```
fluffy: I'm a bit concerned about how this is detected and interaction with caches. Implementations tend to drop duplicate objects without much looking into them.  It might be better to say use which ever is seen first than lowest. I also don't want to imply there is any requirement for a relay to detect duplicates.

draft-ietf-moq-transport.md
fluffy: If applications care about the race, they could include priority in the object during the time of transition where the race can occur.

fluffy (commented):
I think a bit more to sort out around the race condition on this but FWIW ... I think applications can live with some race conditions on this.

draft-ietf-moq-transport.md
michalhosna: Under "9.1. Caching Relays" we currently have 

> An endpoint that receives a duplicate Object with a different Forwarding Preference, Subgroup ID, Priority or Payload *MUST* treat the track as Malformed.

so this should not be something new. For non-caching relays, or generally in cases where we do not know that this is even conflicting value because we do not have the "old" object the "If conflicting Publisher Priority values are **observed**" doesn't apply anyway. 

The point of that statement was to have some recommended behavior, but I am happy to drop that.

draft-ietf-moq-transport.md
```diff
@@ -678,27 +678,24 @@ to MOQT constraints. Such a Track is considered malformed.  Some example
 conditions that constitute a malformed track when detected by a receiver
 include:
 
-1.  An Object with a particular Subgroup ID is received, but its
-     Publisher Priority is different from that of the previous Object with the same
-     Subgroup ID.
```
suhasHere, line 683: this is still true when the publisher priority wasn't updated.

afrind: This opens an interesting quesiton:

If I started a subgroup with no priority (inherit default) and the default changed mid-subgroup, do newly received objects in that subgroup get the new priority?  If the answer is _yes_, then deleting the above _is correct_.  

If I'm a relay, should I reprioritize the matching outgoing subgroups?  If so, we need to state that.

draft-ietf-moq-transport.md
michalhosna: > If applications care about the race, they could include priority in the object during the time of transition where the race can occur.

This covers the "too late" direction, but not REQUEST_UPDATE arriving *too early*.

When the publisher decides to change the priority, it can start stamping explicit priority on objects from that point on and send REQUEST_UPDATE. But objects sent before that decision, already in flight/submited to the QUIC stack can't be retroactively stamped. 

Since the bidi request stream is prioritized above data-plane, I’d expect this to be the most common form of the race.

> We could allow the sender to put Largest Object in the REQUEST_UPDATE and say that if its present, the priority is effective as of Largest Object?

Should this be a generic `REQUEST_UPDATE` synchronization primitive? Would we want to allow it only for this specific case?

draft-ietf-moq-transport.md
michalhosna: > @suhasHere this is still true when the publisher priority wasn't updated.

It may be hard to tell if a publisher priority update happened, especially in multi-hop (even just P->R->S) or multi-publisher cases. 

Also it's now effectively a hop-by-hop compression parameter (maybe it always was?), so a Relay may decide to send an explicit per-object/per-subgroup priority instead of updating the default param. If we accept that priority changes are racy/best-effort, I think we have to drop any priority-related malformation rules, as the publisher now doesn't know what the subscriber (or anyone downstream) thinks the priority is for any given object that was sent without an explicit priority.

> @afrind If I'm a relay, should I reprioritize the matching outgoing subgroups? If so, we need to state that.

Good point, I hate this. 
How do you then expose this downstream for the next hops? 

I don't think we should actually allow effective mid-subgroup publisher priority changes in the current prioritization model. Priority should be fixed once you receive the start of a subgroup, and changes apply only to not-yet-started subgroups. This should be document in the document. 

I could see a mid-subgroup priority change being useful, but I think we would need a bigger update to the prioritization model. But reading the discussion here, we may need that anyway :/

draft-ietf-moq-transport.md
ianswett: It looks like SUBSCRIPTION_STATE_NOTIFY #1820 will land soon, in which case you could include Priority and the Largest Object/First Object with the new Priority there.  If we had that, I'm not sure we need to include it in the Objects as well?

afrind:
Discussed 8/10: there was very little enthusiasm about doing more than what's currently in this PR to solve races, but also very little enthusiasm to do this at all.  The chairs sent out an [email](https://mailarchive.ietf.org/arch/msg/moq/gqei2X0Z9pmxOjirJwYYdILVba8/) on the list:

> No one on the call would assert that this compression was a critical feature. However, the editors feel that some proponents were not present, so I've initiated this thread to allow any proponents to state the case that this is worth the logical and conceptual problems that it introduces.

CC: @fluffy

PR #1771, opened 2026-06-17
Specify relay processing rules for known Track Properties

### LINKED ISSUE #1663: Known Properties have no defined processing rules
Known Track Properties with no defined processing rules can be mutated/dropped by relays.

Per 2.5 Properties:
https://github.com/moq-wg/moq-transport/blob/935b9e8df4feb5dbef209295ddca9c15c5590293/draft-ietf-moq-transport.md?plain=1#L770-L771

Only `IMMUTABLE_PROPERTIES`, `PRIOR_GROUP_ID_GAP` and `PRIOR_OBJECT_ID_GAP` define rules at the moment.

Related: #1270

michalhosna:
I would call this a minimal straw-man proposal.

I am very much not sure which way we want to go with making all track properties immutable (immutable by relays, or immutable in time).

Fixes #1663

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -777,8 +777,11 @@ the most recent set SHOULD replace any cached values, removing any unknown
 values not present in the new set.  Relays MUST NOT attempt to merge sets
 of unknown properties received in different messages.
 
-If a Relay supports a Property, it MAY be modified, added, removed, and/or
-cached, subject to the processing rules specified in the definition.
+If a Relay supports a Property, it MUST follow the processing rules in the
+Property's definition.  Unless those rules permit otherwise, a Relay MUST NOT
+modify, add, or remove the Property; it MUST forward the Property to downstream
+subscribers and MUST cache it with the Track or Object if the Track or Object is
+cached.
 
 Properties are serialized as Key-Value-Pairs (see {{moq-key-value-pair}}).
 Track Properties always appear as the final field in the messages that


draft-ietf-moq-transport.md
```diff
@@ -4501,6 +4513,9 @@ the session with `PROTOCOL_VIOLATION`.
 
 If omitted, the value is 0.
 
+A Relay MAY set this value to 0, or remove the Property entirely (an omitted
+value means 0)
+
```
afrind: This doesn't seem quite right.  This is a property of the original publisher a relay probably shouldn't change it.  Relays have to implement NGR coalescing.

draft-ietf-moq-transport.md
```diff
@@ -4467,6 +4476,9 @@ handles a downstream request that includes those Objects re-requests them.
 If MAX_CACHE_DURATION is not sent by the publisher, the Objects
 can be cached until implementation constraints cause them to be evicted.
 
+MAX_CACHE_DURATION is an upper bound: a Relay MAY serve Objects from cache for a
+shorter time, or forward a smaller value, but MUST NOT increase it.
```
afrind: MAY server for a shorter time -- this is fine.  Caching is not a guarantee, and resources are finite.  But I'm not sure we should allow relays to shorten this value?

draft-ietf-moq-transport.md
```diff
@@ -4444,13 +4447,19 @@ It is a varint.  Its semantics are defined in {{delivery-timeouts}}.  As an
 Object Property on the first object in a subgroup, it overrides the Track-level
 value for that subgroup; it is ignored on any other object in the subgroup.
 
+The timeout is an upper bound: a Relay MAY enforce or forward a shorter timeout,
+but MUST NOT lengthen it.
```
afrind: I think enforcing a shorter timeout is discussed in delivery-timeout -- eg if the publisher asks for 1 day, that might not be honored.

Unsure a relay would every mess with the forwarded value though, as this could effect application logic.  In order to do so, the relay would need some out of band information, at which point it no longer needs to be a capital R Relay.

michalhosna:
Lets keep this clean leave discussion of the specific rules for some other time. 
This imho still improves the current state.

ianswett (approved):

vasilvv (approved):

afrind (approved):

PR #1773, opened 2026-06-18
Add specific protocol violation error codes organized by protocol area

### LINKED ISSUE #1664: Can not differenate protocol violations
There are many things that cause a protocol violation. It is too hard to know what caused it when a client gets this error. We should provide information on which error in the spec caused this. 

Note this may give extra info to attackers about the implications. That seems less relevant in a world where all the relays are open source and put their name and version in the setup message.

suhasHere:
Fixes #1664

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1210,10 +1210,9 @@ SHOULD treat it as a protocol violation in the corresponding area.
 
 ### Data Stream Errors (0x20-0x2F) {#data-stream-errors}
 
-INVALID_OBJECT_FORMAT (0x20):
-: An Object contained an invalid status value, invalid datagram or stream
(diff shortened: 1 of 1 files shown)


afrind:
I think this is a generally good direction, modulo some nit-picking of the specific codes, but we need to park it until the draft settles a bit more.

suhasHere:
@afrind  I am fine with parking it . But I do feel we need to start thinking this direction now and continue doing the same too . I would love to gather as much feedback and keep this PR as running notepad until we are ready to unpark ?

afrind:
> @afrind I am fine with parking it . But I do feel we need to start thinking this direction now and continue doing the same too . I would love to gather as much feedback and keep this PR as running notepad until we are ready to unpark ?

Agreed this is right directionally and keep the PR open to collect ideas.

PR #1775, opened 2026-06-18
Clarify multiple concurrent subscriptions per Track

### LINKED ISSUE #1633: Should we allow more than one concurrent subscription per Track
There's been a PR open on this for a while #1451 - but no specific tracking issue.  The editors have had slides on the topic since IETF 125, but we've never had enough agenda time to discuss.

I'm copying the text of the slides here, which includes proposals.

---
Primary Use Case: Preventing Relay DOS via “needle-in-a-haystack”

Subscriber 1: SUB t=huge track, range={100000,100000}-{100000,100001}
Subscriber 2: SUB t=huge track, range={200000,200000}-{200000,200001}

Relay has to either open 2 upstream sessions with narrow filters, or 1 session that is unfiltered or under-filtered.

Problem exists today but #1518 filters will create more opportunities
	Filters has been updated with “AndSets”

---
Secondary Use Case:

	SUBSCRIBE to a Track
	Cancel the Subscription by closing the request stream
		The RESET_STREAM is lost or delayed
	SUBSCRIBE to the Track again (for whatever reason)
		Peer closes the Session

Today, can avoid closing the Session by waiting for the Request Stream to fully close, but that’s slow and somewhat complex.

---
Considerations:

Wire Deduplication or Double Work
- Do subscriptions with overlapping filters receive the object once, or once per subscription?
- If it’s only once, both the publisher and subscriber both have to filter

Deduplicating multiple subscriptions is an illusion created by the publisher coalescing subscriptions and doing a union of the filters

The same effect could be created by expressing the union in a single subscription

Multiple subscriptions can have conflicting params that affect delivery
	Group Order, Priority, Delivery Timeout

---
Proposal 1a:

Allow multiple subscriptions, each with a unique Track Alias; no deduplication.
Subscriber/relay is responsible for minimizing overlaps/duplication and mapping correctly from upstreams to downstreams.

Proposal 1b:

Allow multiple subscriptions, all get the same Track Alias
Publisher coalesces these as a single subscription with a union of filters
Conflicting parameters ¯\_(ツ)_/¯

Proposal 2:

Keep single subscription per track
Allow unions of all filter filter types in control messages
Can’t specify conflicting parameters (Delivery Timeout, etc) 

Proposal 3:

Status Quo

Editors’ Recommendation: 1a

suhasHere:
Fixes #1633

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1433,18 +1433,16 @@ forwarded back to the endpoint, subject to priority and congestion response
 rules.
 
 An endpoint MAY have multiple concurrent subscriptions to the same Track,
-each identified by a unique Request ID. The publisher assigns their Track
-Aliases according to the TRACK_ALIAS_ASSIGNMENT_POLICY (see
-{{track-alias-assignment-policy}}); by default each subscription gets a unique
-Track Alias, letting the subscriber associate Objects with a specific
-subscription.
+each identified by a unique Request ID. A publisher MAY assign the same or
+different Track Aliases to these subscriptions.
 
 When an Object matches the filters of multiple subscriptions to the same Track,
-the publisher MUST send it exactly once per distinct Track Alias among them.
-Subscribers SHOULD avoid overlapping filters across subscriptions to the same
-Track. When subscriptions share a Track Alias, the subscriber must re-apply each
-subscription's filter to determine which subscriptions a received Object belongs
-to.
+the publisher MUST send the Object once for each matching subscription, even
+when those subscriptions share the same Track Alias. Because subscriptions can
+share a Track Alias, the subscriber re-applies each subscription's filter to
+determine which subscription a received Object belongs to. Subscribers SHOULD
+avoid overlapping filters across subscriptions to the same Track, as they are
+responsible for deduplicating any resulting duplicate Objects.
 
 A publisher SHOULD begin sending incomplete objects when available to avoid
 incurring additional latency.
@@ -2730,18 +2728,6 @@ advertising or other nonessential information. Implementations SHOULD NOT use
 the identifiers of other implementations to declare compatibility, as this
 undermines the usefulness of implementation identification for debugging.
 
-#### TRACK ALIAS ASSIGNMENT POLICY {#track-alias-assignment-policy}
-
-A subscriber sends the TRACK_ALIAS_ASSIGNMENT_POLICY option (Option Type 0x09),
-a variable-length integer, to control how the publisher assigns Track Aliases
-(see {{track-alias}}) to its subscriptions to a Track. With the default value of
-0, the publisher MUST assign a unique Track Alias to each subscription. With
-value 1, the publisher MAY assign one Track Alias to multiple subscriptions to
-the same Track. A publisher MUST treat any other value as the default. When
-sending a PUBLISH before receiving the peer's SETUP, a publisher uses the
-default value.
-
-
 ## GOAWAY {#message-goaway}
 
 An endpoint sends a `GOAWAY` message on its control stream to inform the peer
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1432,18 +1432,16 @@ subscriptions initiated by other endpoints, and all published Objects will be
...
-If a publisher receives a SUBSCRIBE request for a Track with an existing
-subscription in `Pending (publisher)` state, it MUST fail that request with
-a `DUPLICATE_SUBSCRIPTION` error. If a subscriber receives a PUBLISH for a Track
-with a subscription in the `Pending (Subscriber)` state, it MUST ensure the
-subscription it initiated transitions to the `Terminated` state before sending
-PUBLISH_OK.
+An endpoint MAY have multiple concurrent subscriptions to the same Track,
+each identified by a unique Request ID. A publisher MAY assign the same or
+different Track Aliases to these subscriptions. Using different Track Aliases
+allows the subscriber to associate Objects with a specific subscription.
```
afrind: This is what we said, but the publisher doesn't really know whether the subscriber wants this or not, and the subscribers behavior depends on the answer.

S->P: SUB track1, id=0
S->P: SUB track1, id=1
P->S: SUB_OK, id=0, alias=1
P->S: SUB_OK, id=1, alias=1  --> delayed in network
P->S: Object, Alias=1 --> delivered only to first subscription

It's actually worse if the publisher uses PUBLISH -- the subscriber doesn't know there's another control channel coming.

draft-ietf-moq-transport.md
```diff
@@ -1432,18 +1432,16 @@ subscriptions initiated by other endpoints, and all published Objects will be
...
+
+When an Object matches the filters of multiple subscriptions to the same Track,
+the publisher MUST deliver the Object on each matching subscription.
```
afrind: I think you need to say that the publisher needs to send one object per matching subscription but only one per alias.

draft-ietf-moq-transport.md
```diff
@@ -1432,18 +1432,16 @@ subscriptions initiated by other endpoints, and all published Objects will be
...
+Subscribers SHOULD avoid overlapping filters across subscriptions to the same
+Track, as they are responsible for deduplicating any resulting duplicate
+Objects.
```
afrind: This isn't quite right.  It can result either in duplicate objects when distinct aliases are used, or having to re-apply their filters to sort objects to matching subscriptions when common ones are.

afrind (commented):
Having read this, giving the publisher sole freedom to reuse a duplicate alias or not is a huge hazard - the subscribers need to handle both cases, and they lack sufficient information to do so.

I think the subscriber should opt into alias re-use via a SETUP param.  It can't be a SUBSCRIBE param since it applies to PUBLISH also.

afrind (changes requested):

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1432,18 +1432,19 @@ subscriptions initiated by other endpoints, and all published Objects will be
...
-PUBLISH_OK.
+An endpoint MAY have multiple concurrent subscriptions to the same Track,
+each identified by a unique Request ID. The publisher assigns their Track
+Aliases according to the TRACK_ALIAS_ASSIGNMENT_POLICY (see
+{{track-alias-assignment-policy}}); by default each subscription gets a unique
+Track Alias, letting the subscriber associate Objects with a specific
+subscription.
+
+When an Object matches the filters of multiple subscriptions to the same Track,
+the publisher MUST send it exactly once per distinct Track Alias among them.
```
ianswett: Once per Track Alias?  Doesn't that mean the behavior and implementation is substantially different between reusing a Track Alias and using the same one?  It seems like we're basically allowing either of the behaviors we discussed in London.

afrind (approved):

ianswett (commented):
I think I'm ok with this, but adding this Option seems to mean we're allowing either behavior discussed in London?

afrind:
Discussed 7/6:

1) remove setup option
2) publisher sends an object for every subscription, even if duplicate aliases are chosen

vasilvv (approved):

PR #1776, opened 2026-06-18
Add guidance on missing data timeout scenarios and bounds 

### LINKED ISSUE #1743: Errors from missing data timeouts
The draft says

> Implementations are advised to use timeouts to prevent resource
> exhaustion attacks by a peer that does not send expected data within
> an expected time. Each implementation is expected to set its own
> timeouts.

I think we need more guidance on how to set these and the bounds as this sounds like it likely to lead to high unpredictable results in deployments.

suhasHere:
fixes #1743

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -4753,7 +4753,45 @@ priority, after reaching a resource limit.
 
 Implementations are advised to use timeouts to prevent resource
 exhaustion attacks by a peer that does not send expected data within
-an expected time.  Each implementation is expected to set its own timeouts.
+an expected time.  Each implementation is expected to set its own timeouts
+and SHOULD make them configurable to accommodate different deployment
+scenarios (e.g., interactive conferencing vs. file delivery).
+
+The following scenarios require missing-data timeouts:
+
+Control Message Response:
+: After sending a control message that expects a response (e.g.,
+  SUBSCRIBE expects SUBSCRIBE_OK or SUBSCRIBE_ERROR), the sender SHOULD
+  start a timer. If the response is not received within the timeout, the
+  sender SHOULD close the session with `CONTROL_MESSAGE_TIMEOUT`. This
+  timeout SHOULD be at least 30 seconds to account for relay fan-out and
+  upstream latency.
+
+Data Stream Header:
+: When an endpoint receives a new data stream, it SHOULD start a timer
+  for receiving the stream header (SUBGROUP_HEADER). If the header is not
+  received within the timeout, the endpoint SHOULD close the session with
+  `DATA_STREAM_TIMEOUT`. A timeout of at least 5 seconds is RECOMMENDED,
+  as stream headers are small and should arrive promptly.
+
+Object Data on Open Subgroup:
+: When an endpoint is waiting for the next object header or a FIN on an
+  open subgroup stream, it SHOULD apply a timeout. If the timeout fires,
+  the endpoint MAY send STOP_SENDING on that stream or terminate the
+  subscription, rather than closing the entire session. This timeout
+  SHOULD be at least as large as any negotiated delivery timeout for the
+  track. In the absence of a negotiated delivery timeout, a value of at
+  least 30 seconds is RECOMMENDED for live media; applications with known
+  cadence (e.g., fixed frame rate video) MAY use shorter values.
+
+GOAWAY Response:
+: After sending a GOAWAY, the sender SHOULD close the session with
+  `GOAWAY_TIMEOUT` if the peer does not close the session within the
+  timeout indicated in the GOAWAY message.
+
+When a negotiated delivery timeout ({{delivery-timeouts}}) is in effect
+for a subscription, implementation-level missing-data timeouts for that
+subscription SHOULD NOT be shorter than the delivery timeout.
 
 ### Idle Connection Handling
 


draft-ietf-moq-transport.md
```diff
@@ -4753,7 +4753,45 @@ priority, after reaching a resource limit.
...
+scenarios (e.g., interactive conferencing vs. file delivery).
+
+The following scenarios require missing-data timeouts:
+
+Control Message Response:
+: After sending a control message that expects a response (e.g.,
+  SUBSCRIBE expects SUBSCRIBE_OK or SUBSCRIBE_ERROR), the sender SHOULD
+  start a timer. If the response is not received within the timeout, the
+  sender SHOULD close the session with `CONTROL_MESSAGE_TIMEOUT`. This
+  timeout SHOULD be at least 30 seconds to account for relay fan-out and
```
afrind, line 4767: SHOULD be 30 seconds sounds bananas.  I would remove the specific number.

draft-ietf-moq-transport.md
```diff
@@ -4753,7 +4753,45 @@ priority, after reaching a resource limit.
...
+  upstream latency.
+
+Data Stream Header:
```
afrind, line 4770: There's a second case here, which is receiving a data stream that references an alias that is unknown to the subscriber.

draft-ietf-moq-transport.md
```diff
@@ -4753,7 +4753,45 @@ priority, after reaching a resource limit.
...
+: When an endpoint receives a new data stream, it SHOULD start a timer
+  for receiving the stream header (SUBGROUP_HEADER). If the header is not
+  received within the timeout, the endpoint SHOULD close the session with
+  `DATA_STREAM_TIMEOUT`. A timeout of at least 5 seconds is RECOMMENDED,
```
afrind, line 4774: Why 5 and not 2 or 1?

draft-ietf-moq-transport.md
```diff
@@ -4753,7 +4753,45 @@ priority, after reaching a resource limit.
...
+: After sending a control message that expects a response (e.g.,
+  SUBSCRIBE expects SUBSCRIBE_OK or SUBSCRIBE_ERROR), the sender SHOULD
+  start a timer. If the response is not received within the timeout, the
+  sender SHOULD close the session with `CONTROL_MESSAGE_TIMEOUT`. This
+  timeout SHOULD be at least 30 seconds to account for relay fan-out and
+  upstream latency.
+
+Data Stream Header:
+: When an endpoint receives a new data stream, it SHOULD start a timer
+  for receiving the stream header (SUBGROUP_HEADER). If the header is not
```
afrind, line 4772:
```suggestion
  for receiving the complete stream header (SUBGROUP_HEADER). If the header is not
```

draft-ietf-moq-transport.md
```diff
@@ -4753,7 +4753,45 @@ priority, after reaching a resource limit.
...
+
+Object Data on Open Subgroup:
+: When an endpoint is waiting for the next object header or a FIN on an
+  open subgroup stream, it SHOULD apply a timeout. If the timeout fires,
+  the endpoint MAY send STOP_SENDING on that stream or terminate the
+  subscription, rather than closing the entire session. This timeout
+  SHOULD be at least as large as any negotiated delivery timeout for the
+  track. In the absence of a negotiated delivery timeout, a value of at
+  least 30 seconds is RECOMMENDED for live media; applications with known
+  cadence (e.g., fixed frame rate video) MAY use shorter values.
```
afrind, line 4785: Same

draft-ietf-moq-transport.md
```diff
@@ -4753,7 +4753,45 @@ priority, after reaching a resource limit.
...
+
+Data Stream Header:
+: When an endpoint receives a new data stream, it SHOULD start a timer
+  for receiving the stream header (SUBGROUP_HEADER). If the header is not
+  received within the timeout, the endpoint SHOULD close the session with
+  `DATA_STREAM_TIMEOUT`. A timeout of at least 5 seconds is RECOMMENDED,
+  as stream headers are small and should arrive promptly.
+
+Object Data on Open Subgroup:
+: When an endpoint is waiting for the next object header or a FIN on an
```
afrind, line 4778: What about waiting for the tail of an object with a multi-packet length?

draft-ietf-moq-transport.md
```diff
@@ -4753,7 +4753,45 @@ priority, after reaching a resource limit.
...
+  SHOULD be at least as large as any negotiated delivery timeout for the
+  track. In the absence of a negotiated delivery timeout, a value of at
+  least 30 seconds is RECOMMENDED for live media; applications with known
+  cadence (e.g., fixed frame rate video) MAY use shorter values.
+
+GOAWAY Response:
+: After sending a GOAWAY, the sender SHOULD close the session with
+  `GOAWAY_TIMEOUT` if the peer does not close the session within the
+  timeout indicated in the GOAWAY message.
+
```
afrind, line 4791: You should mention timeout awaiting streams based on PUBLISH_DONE stream count if you are enumerating all places where a timeout could apply.

How long to wait for a FETCH stream?

Another intersects with expires -- how long can a stateful subscription (SUB, SUB_NS, SUB_TRACKS, PUB_NS) without a REQUEST_UPDATE, etc.

afrind (commented):
I think we need to consider carefully if specific numbers are helping or hurting.

This guidance is both helpful in that it reminds you of places you need timeouts, and feels a little nanny-state.

afrind:
Part of a larger conversation about timeouts - which need to be advertised/negotiated in SETUP vs implementation specific, and we also need to discuss whether we want to recommend specific numbers vs just explaining tradeoffs.

PR #1777, opened 2026-06-18
Move GROUP_ORDER from PUBLISH_OK to SUBSCRIBE_TRACKS 

### LINKED ISSUE #1651: How does GROUP_ORDER interact with PUBLISH?
In SUBSCRIBE, the subscriber dictates the group order. If it doesn't specify, the Publisher's default wins.

Notably, GROUP_ORDER is not allowed in REQUEST_UPDATE, presumably because it would be disruptive to update mid-subscription. For one thing, it would involve re-prioritizing all streams in the subscription.

Given this, I don't understand how it's supposed to work with PUBLISH.

The Publisher has no subscriber input when it starts pushing objects so it uses its default group order. But then, a PUBLISH_OK arrives and it's supposed to change the group order?

This would be a lot cleaner if (1) GROUP_ORDER was not allowed in PUBLISH_OK or (2) it was allowed in REQUEST_UPDATE. As an implementer, I would prefer (1) as it's less painful.

I'm happy to write a PR if everyone agrees.

suhasHere:
Fixes #1651

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1739,9 +1739,9 @@ datagram (see {{data-streams}}).
 `Group Order` is a property of an individual subscription.  It can be either
 'Ascending' (groups with lower group ID are sent first), or 'Descending'
 (groups with higher group ID are sent first).  The subscriber optionally
-communicates its group order preference in the SUBSCRIBE message; the
-publisher's preference is used if the subscriber did not express one (by
-omitting the Group Order parameter).  The group order of an existing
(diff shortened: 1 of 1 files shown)


afrind (commented):
This needs to explain what happens for a PUBLISH that is not in response to SUBSCRIBE_TRACKS.

ianswett (approved):

martinduke (approved):

afrind (approved):

vasilvv (approved):

PR #1782, opened 2026-06-18
Update Forward handling for relays

### LINKED ISSUE #1650: Forward Handling SHOULD set Forward=1 upstream is too strong
> 9.2. Forward Handling
> 
> Relays SHOULD set the Forward flag to 1 when a new subscription needs to be sent upstream, regardless of the value of the Forward field from the downstream subscription. Subscriptions that are not forwarded consume resources from the publisher, so a publisher might deprioritize, reject, or close those subscriptions to ensure other subscriptions can be delivered.

Sending a subscription with fwd=0 causing the relay to open an upstream subscription with fwd=1 is another DOS vector.  In real production environments this is going to be a policy decision that involves factors like current load, the size of the subscription, the popularity of the Track, etc.

I propose we drop the normative SHOULD and instead explain that relays have discretion, and highlight some of the possible tradeoffs (faster join vs possible attacks).

afrind:
Replace "SHOULD set to 1" with pure discretion and more clear tradeoffs.

Fixes: #1650

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1958,11 +1958,18 @@ that specify otherwise.
 
 ## Forward Handling
 
-Relays SHOULD set the `Forward` flag to 1 when a new subscription needs to be
-sent upstream, regardless of the value of the `Forward` field from the
-downstream subscription. Subscriptions that are not forwarded consume resources
-from the publisher, so a publisher might deprioritize, reject, or close those
-subscriptions to ensure other subscriptions can be delivered.
+If one or more downstream subscribers to a track have Forward=1, the relay
+MUST set Forward=1 upstream in order to receive and forward the requested
+Objects. When no downstream subscriber has Forward=1, the relay chooses the
+upstream Forward value at its discretion, considering the following
+tradeoffs:
+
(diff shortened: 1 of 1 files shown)


ianswett (approved):

sharmafb (approved):

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1958,11 +1958,18 @@ that specify otherwise.
...
-Relays SHOULD set the `Forward` flag to 1 when a new subscription needs to be
-sent upstream, regardless of the value of the `Forward` field from the
-downstream subscription. Subscriptions that are not forwarded consume resources
-from the publisher, so a publisher might deprioritize, reject, or close those
-subscriptions to ensure other subscriptions can be delivered.
+If one or more downstream subscribers to a track have Forward=1, the relay
+MUST set Forward=1 upstream in order to receive and forward the requested
+Objects. When no downstream subscriber has Forward=1, the relay chooses the
+upstream Forward value at its discretion, considering the following
+tradeoffs:
```
```suggestion
upstream Forward value at its discretion, considering the following
tradeoffs and deployment considerations:
```

suhasHere (approved):

vasilvv (approved):

Issue #1745, opened 2026-06-15
Nothing good comes from an implementation string


fluffy:
The draft says 

> Privacy-conscious deployments MAY omit the MOQT_IMPLEMENTATION
> option entirely or send a generic value.

History would suggest that things will check that and do stuff based on it and then applications will decide to act they are something too big to break.  This then results in further difficulty to debug or fix the root causes.

LPardue:
+1

afrind:
There's no question its adding value to interop at the current stage.  Let's throw it in the pile with Reason Phrase and make a final call later.

PR #1783, opened 2026-06-18
Restrict reason phrase and implementation name to safe ASCII

### LINKED ISSUE #1234: Restrict strings that will be placed in log to be safe

### LINKED ISSUE #1668: security consideration for utf-8 messages to logging
Provide guidance to make utf-8 data safe when sending to logging

afrind:
Replace UTF-8 encoding of the Reason Phrase Value and MOQT_IMPLEMENTATION option with a restriction to printable ASCII characters (0x20-0x7E), and remove the UTF-8 example from the Track Namespace/Track Name encoding discussion. Drops the now-unused RFC3629 reference.

Note, both #273 and #1745 track possible removal of these fields closer to the end.

Fixes: #1234
Fixes: #1668

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -407,9 +407,10 @@ Reason Phrase {
   close the session with a `PROTOCOL_VIOLATION`
 
 * Reason Phrase Value: Additional diagnostic information about the error condition.
-  The reason phrase value is encoded as UTF-8 string and does not carry information,
-  such as language tags, that would aid comprehension by any entity other than
-  the one that created the text.
+  The reason phrase value contains only printable ASCII characters, in the range
+  0x20 to 0x7E inclusive, and does not carry information, such as language tags,
+  that would aid comprehension by any entity other than the one that created the
+  text.
 
 ## Representing Namespace and Track Names {#namespace-name-format}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -407,9 +407,10 @@ Reason Phrase {
   close the session with a `PROTOCOL_VIOLATION`
 
 * Reason Phrase Value: Additional diagnostic information about the error condition.
-  The reason phrase value is encoded as UTF-8 string and does not carry information,
-  such as language tags, that would aid comprehension by any entity other than
-  the one that created the text.
+  The reason phrase value contains only printable ASCII characters, in the range
+  0x20 to 0x7E inclusive, and does not carry information, such as language tags,
```
ianswett, line 411: What happens if there are values outside this range?

afrind: 💣💣💣💣💣

Do we need to say PROTOCOL_VIOLATION for bad reason phrase?  Or recommend receivers validate before printing?

vasilvv:
I'd rather we add a sentence to the security considerations, than add a new normative check into the spec.

ianswett:
We discussed this today, and I don't want to fail the session because of non-ASCII characters, but otherwise I'm fine with whatever text is acceptable to the WG.

afrind:
Based on the concept of "don't create a new error condition", I'll reverse course here, keep UTF-8 for these (which might get removed) and add a note to security considerations.

draft-ietf-moq-transport.md
fluffy: I don't think we really need to cover the what happens if outside this range  - people can do whatever they do when they receive attacks from little bobby drop tables https://xkcd.com/327/. But it makes it clear not send stuff outside this range that you expect the receiver to be able to do anything useful with.

fluffy (approved):
FWIW, if we don't do this now, I think we will be doing it in IETF Last Call.

afrind:
Closing this PR. Based on this [comment](https://github.com/moq-wg/moq-transport/pull/1783#issuecomment-4846932100)

PR #1784, opened 2026-06-18
Unexpected REQUEST_UPDATE is a session error

### LINKED ISSUE #1769: Is an unexpected REQUEST_UPDATE a request or session error?
The sender of a request can send `REQUEST_UPDATE` to modify it, and there is the additional case of a subscriber sending a `REQUEST_UPDATE` in the other direction when updating the subscribe-side params of a `PUBLISH` that it has accepted, if I understand correctly? 

I'm assuming that in any other case receiving a `REQUEST_UPDATE` on a request stream would be a `PROTOCOL_VIOLATION` (rather than a `REQUEST_ERROR` response)? Not sure if the draft should be more explicit on that point if so.

afrind:
Fixes: #1769

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3082,6 +3082,9 @@ SUBSCRIBE_NAMESPACE, SUBSCRIBE_TRACKS) can later send a REQUEST_UPDATE on the
 same bidi stream as the request to modify it.  A subscriber can also send
 REQUEST_UPDATE to modify parameters of a subscription established with PUBLISH.
 
+An endpoint that receives a REQUEST_UPDATE other than in the two cases above
+MUST close the session with a `PROTOCOL_VIOLATION`.
+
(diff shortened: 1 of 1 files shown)


ianswett (approved):
I think this is fine, but it feels like a Publisher should be able to update a Subscription, even if they didn't initiate it?  DTS and TopN filter seem like they'll need that?

afrind:
> a Publisher should be able to update a Subscription

The publisher can, when they initiated (via PUBLISH).  e.g. they can update AUTH_TOKEN, but the spec _today_ has no other parameters in PUBLISH or SUBSCRIBE_OK.  That could change with #1770, and if so, we'd need to update it here.  Top-N also uses REQUEST_UPDATE to notify the subscriber of forward changes triggered by top-N.  That might change shape (e.g. splitting publisher and subscriber forward states?), but those are always on PUBLISH initiated subscriptions anyways.

vasilvv (approved):

sandarsh (approved):

sharmafb (approved):

PR #1788, opened 2026-06-27
SUBSCRIBE_TRACKS Parameters copied to PUBLISH

### LINKED ISSUE #1660: SUBSCRIBE_TRACKS should contain SUBSCRIBE parameters
In London, we agreed that SUBSCRIBE_TRACKS can/should contain parameters that only apply to SUBSCRIBE, and resulting PUBLISH messages should act as if those parameters had been included in a SUBSCRIBE.

ianswett:
Specifes that any Subscription Parameters can be used on SUBSCRIBE_TRACKS and those will be the default for resulting PUBLISHes.

Assumes #1777 will also land, so doesn't touch that text.

Fixes #1660

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3737,7 +3737,9 @@ SUBSCRIBE_TRACKS Message {
   receives a Track Namespace Prefix consisting of greater than 32 Track
   Namespace Fields, it MUST close the session with a `PROTOCOL_VIOLATION`.
 
-* Parameters: The parameters are defined in {{message-params}}.
+* Parameters: The parameters are defined in {{message-params}}, though they
+  are handled differently from the same Parameters on Subscriptions, as outlined
+  below.
 
 The publisher will respond with REQUEST_OK or REQUEST_ERROR on the response half
 of the stream. If the subscriber receives any message other than a REQUEST_OK or a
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -3760,12 +3762,21 @@ SUBSCRIBE_TRACKS is not required for a publisher to send PUBLISH messages to
...
-from this message, PUBLISH messages resulting from this SUBSCRIBE_TRACKS will
-set the FORWARD parameter to 1, or indicate that value by omitting the parameter
-(see {{subscriptions}}).
+### Parameters on SUBSCRIBE_TRACKS
+
+Any Parameter that can be specified on a Subscription (ie: in SUBSCRIBE) is valid
+in SUBSCRIBE_TRACKS, unless otherwise specified. These parameters are copied
+over as the default Subscription parameters when a PUBLISH is sent as a result of
+SUBSCRIBE_TRACKS. The Parmeters MUST still be included with the resulting PUBLISHes
+to ensure there is no ambiguity about the state of the Subscription.
```
afrind: This is strange, because a bunch of those parameters are not valid in PUBLISH (eg SUBSCRIBER_PRIORITY).

ianswett: Good point on Subscriber Priority. Which other ones aren't valid?

Also see #1790

afrind: SUBSCRIPTION_FILTER
SUBSCRIBER_PRIORITY
*_DELIVERY_TIMEOUT
NEW_GROUP_REQUEST
RENDEZVOUS_TIMEOUT (though it doesn't make any sense in sub_tracks)

We just changed GROUP_ORDER to be allowed in #1777, but didn't update the "allowed messages".

AUTH_TOKEN is also nonsensical.

ianswett: I think we want to allow the Publisher to update the SUBSCRIPTION_FILTER for Switch/DTS/etc so it can communicate when the Subscription now ends.

Wouldn't RENDEZVOUS_TIMEOUT potentially make sense if we want to join any ongoing tracks at the current or a recent Group?  In which case, we'd put a Fill filter in the SUB_TRACKS.

Putting NGR in SUB_TRACKS seems odd, so maybe that shouldn't be allowed?

ianswett:
Discussed at virtual interim today, plan is to not communicate the Parameters for this PR, and then follow up and decide how we want to communicate them.

There was some interest in a Parameter in PUBLISH that referenced the Request ID of a SUB_TRACKS, so you knew what it came from, including the Subscribe parameters, but that's a new idea and we'll file a new issue for that.

draft-ietf-moq-transport.md
```diff
@@ -3821,6 +3823,14 @@ SUBSCRIBE_TRACKS is not required for a publisher to send PUBLISH messages to
 a subscriber.  It is useful for subscribers that are
 only interested in or authorized to access a subset of available tracks.
 
+### Parameters on SUBSCRIBE_TRACKS
+
+Any Parameter that can be specified on a Subscription (ie: in SUBSCRIBE) is valid
+in SUBSCRIBE_TRACKS, unless otherwise specified. These parameters are copied
+over as the default Subscription parameters when a PUBLISH is sent as a result of
```
afrind, line 3825:
```suggestion
in SUBSCRIBE_TRACKS, unless otherwise specified. These parameters are used by
the publisher as the initial Subscription parameters when a PUBLISH is sent as a result of
```

afrind (approved):
Left a wording suggestion, but LGTM

vasilvv (approved):

suhasHere (approved):

draft-ietf-moq-transport.md
ianswett: See #1815

PR #1790, opened 2026-06-27
Subscription parameters appear in REQUEST_UPDATE, not PUBLISH_OK

### LINKED ISSUE #1796: PUBLISH_OK parameters can have unpredictable results
When writing a PR to fix #1660 and reviewing #1673, I realized that changing parameters of a Subscription was only safe for some parameters (ie: Priority) and had unpredictable behavior for Forward and Location Filters.  In particular, anything that changes what Objects are sent for a Subscription (ie: Forward or Filters) has a stale Largest Object, as pointed out in #1673, so its unclear what Objects a Subscriber expects to get.

Additionally, a Subscriber can immediately send a REQUEST_UPDATE after the PUBLISH_OK, so there's no real need to update any Subscription parameters in PUBLISH_OK.  As a result, this doesn't change the functionality of MoQ at all, just the message sequence.

I wrote #1790 with this in mind.

ianswett:
Fixes #1796 

PUBLISH_OK no longer mutates the state of the Subscription.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2592,7 +2592,7 @@ that alias has not received a response.
 ### SUBGROUP_DELIVERY_TIMEOUT Parameter {#subgroup-delivery-timeout}
 
 The SUBGROUP_DELIVERY_TIMEOUT parameter (Parameter Type 0x06) is a varint. It
-MAY appear in a PUBLISH, SUBSCRIBE, or REQUEST_UPDATE message.  Its
+MAY appear in a SUBSCRIBE or REQUEST_UPDATE message.  Its
 semantics are defined in {{delivery-timeouts}}.
 
 This parameter is intended to be specific to a subscription, so it SHOULD NOT
@@ -2602,7 +2602,7 @@ for the same track.
 ### OBJECT_DELIVERY_TIMEOUT Parameter {#object-delivery-timeout}
 
 The OBJECT_DELIVERY_TIMEOUT parameter (Parameter Type 0x02) is a varint. It
-MAY appear in a PUBLISH, SUBSCRIBE, or REQUEST_UPDATE message.  Its
+MAY appear in a SUBSCRIBE or REQUEST_UPDATE message.  Its
 semantics are defined in {{delivery-timeouts}}.
 
 This parameter is intended to be specific to a subscription, so it SHOULD NOT
@@ -2661,13 +2661,12 @@ If RENDEZVOUS_TIMEOUT is absent, the default is 0.
 ### SUBSCRIBER PRIORITY Parameter {#subscriber-priority}
 
 The SUBSCRIBER_PRIORITY parameter (Parameter Type 0x20) is a uint8. It MAY
-appear in a SUBSCRIBE, FETCH, REQUEST_UPDATE (for a subscription or FETCH),
-or PUBLISH message. It is an integer expressing the priority of a
+appear in a SUBSCRIBE, FETCH, or REQUEST_UPDATE (for a subscription or FETCH).
+It is an integer expressing the priority of a
 subscription relative to other subscriptions and fetch responses in the same
 session. Lower numbers get higher priority. See {{priorities}}.
 
-If omitted from SUBSCRIBE, PUBLISH or FETCH, the publisher uses
-the value 128.
+If omitted from SUBSCRIBE or FETCH, the publisher uses the value 128.
 
 ### GROUP ORDER Parameter {#group-order}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2131,7 +2131,8 @@ send a SUBSCRIBE to the publisher that sent the PUBLISH_NAMESPACE for each
 matching subscription.  When it receives an authorized PUBLISH message for a
 Track that has `Established` downstream subscriptions, it MUST respond with
 PUBLISH_OK.  If at least one downstream subscriber for the Track has
-Forward State=1, the Relay MUST use Forward State=1 in the reply.
+Forward State=1, the Relay MUST change the Forward State to 1 with
+REQUEST_UPDATE.
 
 If a Session is closed due to an unknown or invalid control message or Object,
 the Relay MUST NOT propagate that message or Object to another Session, because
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2055,10 +2055,10 @@ to the old relay can be cancelled (see {{request-cancellation}}).
 There are two ways to publish through a relay:
 
 1. Send a PUBLISH message for a specific Track to the relay. The relay MAY
-respond with PUBLISH_OK in Forward State=0 until there are known subscribers for
-new Tracks.
+update the Subscription with REQUEST_UPDATE in Forward State=0 until there are
+known subscribers for new Tracks.
 
-2. Send a PUBLISH_NAMESPACE message for a Track Namespace to the relay. This
+3. Send a PUBLISH_NAMESPACE message for a Track Namespace to the relay. This
```
sharmafb: nit: Should be (2)

draft-ietf-moq-transport.md
sharmafb: Isn't this text stale now, since the PUBLISH_OK can't have a forward state in it?

draft-ietf-moq-transport.md
```suggestion
2. Send a PUBLISH_NAMESPACE message for a Track Namespace to the relay. This
```

draft-ietf-moq-transport.md
ianswett: Good point, I updated it.

sharmafb:
Allowing SUBGROUP_DELIVERY_TIMEOUT/OBJECT_DELIVERY_TIMEOUT in PUBLISH seems wrong.

draft-ietf-moq-transport.md
```diff
@@ -1495,7 +1495,7 @@ incurring additional latency.
 Publishers MAY start sending Objects on PUBLISH-initiated subscriptions before
 receiving a PUBLISH_OK response to reduce latency.  Doing so can consume
 unnecessary resources in cases where the Subscriber rejects the subscription
-with REQUEST_ERROR or sets Forward State=0 in PUBLISH_OK. It can also result in
+with REQUEST_ERROR. It can also result in
```
```suggestion
with REQUEST_ERROR or sets Forward=0 in REQUEST_UPDATE. It can also result in
```

draft-ietf-moq-transport.md
```diff
@@ -2685,10 +2685,10 @@ the Track is used. If omitted from FETCH, the receiver uses Ascending (0x1).
 ### LOCATION FILTER Parameter {#location-filter}
 
 The LOCATION_FILTER parameter (Parameter Type 0x21) uses length-prefixed
-encoding. It MAY appear in a SUBSCRIBE, PUBLISH_OK or REQUEST_UPDATE (for a
+encoding. It MAY appear in a SUBSCRIBE or REQUEST_UPDATE (for a
 subscription) message. It is a Location Filter (see {{location-filters}}).
 
-If omitted from SUBSCRIBE or PUBLISH_OK, the subscription is
+If omitted from SUBSCRIBE or PUBLISH, the subscription is
```
```suggestion
If omitted from SUBSCRIBE, the subscription is
```

ianswett:
I updated this PR to not change what Parameters can be put into PUBLISH and only specify that Parameters like the Location filter can't be put into PUBLISH_OK.

afrind:
This still removes a number of params from PUBLISH_OK besides forward/location -- timeouts and ngr?  Is that an oversight or are you still wanting to remove these?

ianswett:
> This still removes a number of params from PUBLISH_OK besides forward/location -- timeouts and ngr? Is that an oversight or are you still wanting to remove these?

Yes, my intent is that PUBLISH_OK no longer changes the Subscription and instead one does that with REQUEST_UPDATE. I can imagine there might be exceptions where a Parameter is more informative than anything else, but I don't think that's the case for any existing Parameters?

draft-ietf-moq-transport.md
```diff
@@ -2193,8 +2193,8 @@ to the old relay can be cancelled (see {{request-cancellation}}).
 There are two ways to publish through a relay:
 
 1. Send a PUBLISH message for a specific Track to the relay. The relay MAY
-respond with PUBLISH_OK in Forward State=0 until there are known subscribers for
-new Tracks.
+update the Subscription with REQUEST_UPDATE in Forward State=0 until there are
```
```suggestion
pause the Subscription with REQUEST_UPDATE in Forward State=0 until there are
```

afrind (approved):
Am I correctly reading now that no parameters are allowed in PUBLISH_OK?

ianswett:
> Am I correctly reading now that no parameters are allowed in PUBLISH_OK?

I just updated the PR so that's now true.  I updated the PR to not allow PUBLISH_OK to specify filters and not allow it to include 'Expires'.  All other uses of Expires are by the Publisher, so it was a bit odd that we ever allowed that, but let me know if there's a reason we need Expires in PUBLISH_OK.

suhasHere (approved):

afrind:
> All other uses of Expires are by the Publisher, so it was a bit odd that we ever allowed that, but let me know if there's a reason we need Expires in PUBLISH_OK.

I think we need to leave expires. It’s sent by any endpoint to tell the peer that they need to refresh the request to keep it alive.

eg if the client is a publisher, the relay can add EXPIRES to PUBLISH_OK to indicate auth validity

vasilvv (approved):

ianswett:
> > All other uses of Expires are by the Publisher, so it was a bit odd that we ever allowed that, but let me know if there's a reason we need Expires in PUBLISH_OK.
> 
> I think we need to leave expires. It’s sent by any endpoint to tell the peer that they need to refresh the request to keep it alive.
> 
> eg if the client is a publisher, the relay can add EXPIRES to PUBLISH_OK to indicate auth validity

I was thinking that we didn't allow REQUEST_UPDATE to be sent by the publisher, but in that case it looks like we do, so I restored it.

draft-ietf-moq-transport.md
```diff
@@ -2735,7 +2735,7 @@ REQUEST_ERROR with error code INVALID_FILTER. See {{range-filters}}.
 ### EXPIRES Parameter {#expires}
 
 The EXPIRES parameter (Parameter Type 0x8) is a varint. It MAY appear in
-SUBSCRIBE_OK, PUBLISH, PUBLISH_OK, SUBSCRIBE_NAMESPACE_OK, SUBSCRIBE_TRACKS_OK,
+SUBSCRIBE_OK, PUBLISH, SUBSCRIBE_NAMESPACE_OK, SUBSCRIBE_TRACKS_OK,
```
```suggestion
SUBSCRIBE_OK, PUBLISH, PUBLISH_OK, SUBSCRIBE_NAMESPACE_OK, SUBSCRIBE_TRACKS_OK,
```

PR #1805, opened 2026-07-07
Fix Redirect ambiguity with empty name and namespace

### LINKED ISSUE #1793: REDIRECT ambiguity with empty namespace and empty name
Redirect uses "zero namespace + empty track name" as "same as original", but that is also a valid Full Track Name. Should we instead add an explicit flag?

sharmafb:
Fixes #1793


afrind (approved):

afrind:
We discussed this in an Auth/Ed call and decided we want to remove the optionality of Trackname/Namespace.  I'll make a separate PR.

afrind:
Closing in favor of 1824

PR #1809, opened 2026-07-09
Location filter changed to match the design of other filters

mzanaty:


draft-ietf-moq-transport.md
```diff
@@ -1518,64 +1518,52 @@ end with an error.
...
-close the session with a `PROTOCOL_VIOLATION`.
+If Length is zero, the Start Location is `{Largest Object.Group, Largest Object.Object + 1}`,
+or {0,0} if no content has been delivered yet, and the subscription is open-ended.
+Note that due to network reordering or prioritization, relays can receive Objects with
+Locations smaller than `Largest Object` after the SUBSCRIBE is processed, but
+these Objects do not pass this filter.
 
-An endpoint that receives a filter type other than the above MUST close the
-session with `PROTOCOL_VIOLATION`.
+If only StartGroup is present, it is a relative number of groups prior to the next group,
```
suhasHere: Just to make sure it the range with StartGroup is current group +1 and walk backwards .

If it is not the current group, is by default we express to fill fetch ?

mzanaty: Fill fetch interaction is still TBD. It depends if people like a single range in Sub to Fill, or if they prefer separate Sub vs Fill ranges and parameters. This can work either way. But the semantics of what triggers a fill (single implicit range or separate explicit fill parameter) depend on the fill design.

draft-ietf-moq-transport.md
```diff
@@ -1518,64 +1518,52 @@ end with an error.
 
 ### Location Filters {#location-filters}
 
-Subscribers can specify a Location filter on a subscription indicating to the publisher
+Subscribers can specify a Location filter parameter on a subscription or fetch indicating to the publisher
```
sharmafb: FETCH already has mandatory Start Location / End Location fields. Does the parameter replace that range, intersect it, or is it invalid when both specify different ranges?

sharmafb: The existing LOCATION_FILTER Parameter section still only permits SUBSCRIBE, PUBLISH_OK, and subscription REQUEST_UPDATE. You may want to update that.

draft-ietf-moq-transport.md
```diff
@@ -1518,64 +1518,52 @@ end with an error.
...
-  [Start Location (Location),]
-  [End Group Delta (vi64),]
-}
+LOCATION_FILTER Parameter {
+  Parameter Type (vi64) = 0x21,
+  Length (vi64),
+  [StartGroup (vi64),]
+  [StartObject (vi64),]
+  [EndGroupDelta (vi64),]
+  [EndObject (vi64),]
```
sharmafb, line 1563: You may want to add some clarifying text on how this is decoded (two varints presumably mean absolute {StartGroup, StartObject}, three mean an end group with omitted EndObject, and four mean exact end object)

draft-ietf-moq-transport.md
suhasHere: makes sense @mzanaty My preference is to have them separate to keep the intent very clear.

draft-ietf-moq-transport.md
mzanaty: WIP. Plan is to replace fetch fields or fetch itself. Discussions ongoing.

mzanaty: WIP

draft-ietf-moq-transport.md
mzanaty: WIP

mzanaty: Agreed, updated.

draft-ietf-moq-transport.md
mzanaty: We're proposing separate location ranges in live Subscribe vs Fill parameter for cache fill/fetch. This will be detailed in PR#1673 not here, assuming we land 1809 before 1673.

draft-ietf-moq-transport.md
mzanaty: Replaced FETCH location fields with Location Filter parameter.

mzanaty: Added FETCH to allowed messages.

ianswett:
At the Vienna IETF, there was general support for this change, so it's just a matter of the editors/authors reviewing and approving it.

There was general agreement this should be landed before #1673

draft-ietf-moq-transport.md
```diff
@@ -1518,64 +1518,63 @@ end with an error.
...
-Object` is the Object with the largest Location ({{location-structure}}) in the
-Track from the perspective of the publisher processing the message. Largest
-Object updates when the first byte of an Object with a Location larger than the
+Subscribers can specify a Location filter parameter on a subscription or fetch
+indicating to the publisher which Objects to send.  Subscriptions without a
+filter pass all Objects published or received via upstream subscriptions.
+
+A Location filter specifies an inclusive range of Locations.  Only objects
+published or received via a subscription having Locations within the inclusive range
+pass the filter.
```
ianswett: Maybe this suggestion is too simple, but the existing text is a bit odd, especially since this now applies to fetch.

```suggestion
with Locations within the inclusive range pass the filter.
```

draft-ietf-moq-transport.md
```diff
@@ -1518,64 +1518,63 @@ end with an error.
...
+
+If only StartGroup is present, it is a relative number of groups prior to the next group,
+hence the start Location is `{Largest Object.Group + 1 - StartGroup, 0}`. For example:
+  * StartGroup=0 will start at the next group
+  * StartGroup=1 will start at the current group
+  * StartGroup=2 will start at 1 group prior to the current group
+  * StartGroup=N will start at N-1 groups prior to the current group
+
+If only StartGroup and StartObject are present and both 0, the start Location
+is the Next Object which is `{Largest Object.Group, Largest Object.Object + 1}`,
```
ianswett, line 1580: This works, but it feels overly clever and I can see it leading to programming errors.

I'm trying to think of a safer way to serialize this that isn't overly complex, but nothing is coming to mind right now.

draft-ietf-moq-transport.md
```diff
@@ -1518,64 +1518,63 @@ end with an error.
...
+or {0,0} if no content has been delivered yet.
+Note that due to network reordering or prioritization, relays can receive Objects with
+Locations smaller than `Largest Object` after the SUBSCRIBE is processed, but
+these Objects do not pass this filter.
```
ianswett, line 1585: I'm confused why this is relay specific?

Is there something special about Largest Object as a start?

Maybe we don't need this sentence?

draft-ietf-moq-transport.md
```diff
@@ -1518,64 +1518,63 @@ end with an error.
...
+
+Otherwise, all start/end group/object fields are absolute.  EndGroupDelta is delta
```
ianswett: You can't set the end Group/Object to Largest Object?  I'm not sure how useful that is, except the Joining Fetch type usecase?

draft-ietf-moq-transport.md
```diff
@@ -1518,64 +1518,63 @@ end with an error.
...
+encoded from StartGroup, but both the start and end groups are absolute, not
+relative to `Largest Object`.  If StartGroup + EndGroupDelta exceeds 2^64 - 1,
+the endpoint MUST close the session with a `PROTOCOL_VIOLATION`.
+
+EndGroupDelta and EndObject can be omitted for an open-ended subscription
+or to end a fetch at the `Largest Object`.
```
```suggestion
EndGroupDelta and EndObject can be omitted for an open-ended subscription.
When they are omitted from a Fetch, the EndGroup and EndObject are 'Largest Object'.
```

draft-ietf-moq-transport.md
```diff
@@ -2685,10 +2684,10 @@ the Track is used. If omitted from FETCH, the receiver uses Ascending (0x1).
 ### LOCATION FILTER Parameter {#location-filter}
 
 The LOCATION_FILTER parameter (Parameter Type 0x21) uses length-prefixed
-encoding. It MAY appear in a SUBSCRIBE, PUBLISH_OK or REQUEST_UPDATE (for a
+encoding. It MAY appear in a FETCH, SUBSCRIBE, PUBLISH_OK or REQUEST_UPDATE (for a
 subscription) message. It is a Location Filter (see {{location-filters}}).
 
-If omitted from SUBSCRIBE or PUBLISH_OK, the subscription is
+If omitted from FETCH, SUBSCRIBE or PUBLISH_OK, the subscription is
 unfiltered.  If omitted from REQUEST_UPDATE, the value is unchanged.
```
```suggestion
If omitted from FETCH, SUBSCRIBE or PUBLISH_OK, the subscription or Fetch
is unfiltered.  If omitted from REQUEST_UPDATE, the value is unchanged.
```

draft-ietf-moq-transport.md
```diff
@@ -3618,27 +3613,24 @@ Joining Fetch {
...
 subscription is used to calculate the end of a Joining Fetch, so the
 Objects retrieved by the FETCH and SUBSCRIBE are contiguous and non-overlapping.
 
 The publisher receiving a Joining Fetch sets the End Location to
-{Joining Location.Group, Joining Location.Object + 1} (see {{subscriptions}}.
-
-Note: the last Object included in the Joining FETCH response is the Object
-at the Joining Location.  The `+ 1` above indicates the equivalent Standalone
-Fetch encoding.
+`Largest Object` (see {{subscriptions}}.
```
ianswett: I think you want to keep the old text here.  The goal of a Joining Fetch is to set the End Location to the same as the start of the Subscription being joined.

ianswett (commented):
Some small comments.  It'd be nice to frame this in a way that doesn't rely on [0, 0] being a sentinel indicating Largest Object.

draft-ietf-moq-transport.md
```diff
@@ -1518,64 +1518,63 @@ end with an error.
...
-All Location filters have a Start Location and an optional End Group Delta.  Only objects
-published or received via a subscription having Locations greater than or
-equal to Start Location and strictly less than or equal to the End Group (when
-present) pass the filter.
-
-Some Location filters are defined to be relative to the `Largest Object`. The `Largest
-Object` is the Object with the largest Location ({{location-structure}}) in the
-Track from the perspective of the publisher processing the message. Largest
-Object updates when the first byte of an Object with a Location larger than the
+Subscribers can specify a Location filter parameter on a subscription or fetch
```
afrind: not sure we need to explicitly mention FETCH here, unless we're also going to add it to the next sentence "...without a filter"

draft-ietf-moq-transport.md
afrind: +1 - the filter on a subscription has a particular meaning with respect to "newly published / received" objects, vs fetch, which isn't defining a "filter" so much as the bounds of that request.

draft-ietf-moq-transport.md
```diff
@@ -1518,64 +1518,63 @@ end with an error.
...
+indicating to the publisher which Objects to send.  Subscriptions without a
+filter pass all Objects published or received via upstream subscriptions.
+
+A Location filter specifies an inclusive range of Locations.  Only objects
+published or received via a subscription having Locations within the inclusive range
+pass the filter.
+
+Some Location filters are defined to be relative to the `Largest Object` which is
+communicated in SUBSCRIBE_OK.  The `Largest Object` is the Object with the
```
afrind: LARGEST_OBJECT is not communicated in FETCH_OK.

draft-ietf-moq-transport.md
```diff
@@ -1518,64 +1518,63 @@ end with an error.
...
+largest Location ({{location-structure}}) in the
+Track from the perspective of the publisher processing the message.  `Largest
+Object` updates when the first byte of an Object with a Location larger than the
 previous value is published or received through a subscription.
 
-A Location Filter has the following structure:
+A Location filter parameter has the following length-prefixed structure:
```
afrind, line 1554: Odd to have this defined here, and not in the parameters section, though I appreciate it's an easier review. I suppose the editors will bat cleanup after this lands.

draft-ietf-moq-transport.md
```diff
@@ -1518,64 +1518,63 @@ end with an error.
...
-  [Start Location (Location),]
-  [End Group Delta (vi64),]
-}
+LOCATION_FILTER Parameter {
+  Parameter Type (vi64) = 0x21,
+  Length (vi64),
+  [StartGroup (vi64),]
+  [StartObject (vi64),]
+  [EndGroupDelta (vi64),]
+  [EndObject (vi64),]
```
afrind, line 1563: I'll not that we at one time _explicitly_ agreed to remove end object filter from subscriptions, and this unified syntax is adding it back.  It's passed multiple consensus calls.  I don't have an objection to it, but it's probably worth noting. I believe @martinduke filed the original issue that lead to its removal.

draft-ietf-moq-transport.md
```diff
@@ -1518,64 +1518,63 @@ end with an error.
...
-delivered will be the Group ID in `Start Location` plus the `End Group Delta`.
-If the resulting Group ID would be greater than 2^64 - 1, the endpoint MUST
-close the session with a `PROTOCOL_VIOLATION`.
-
-An endpoint that receives a filter type other than the above MUST close the
-session with `PROTOCOL_VIOLATION`.
+Length (in bytes) determines how many optional vi64 fields are present, 0 to 4.
+Length can be 0 for no filter, e.g. to remove the filter in REQUEST_UPDATE.
+Optional fields can be omitted consecutively from the end.  The options are:
+  * 1 vi64: StartGroup is present
```
afrind: I think numbering the options is confusing, because someone might interpret this as the number of bytes.  The length itself can vary fro 0 to 36.

draft-ietf-moq-transport.md
```diff
@@ -1518,64 +1518,63 @@ end with an error.
...
+  * 2 vi64: StartGroup, StartObject are present
+  * 3 vi64: StartGroup, StartObject, EndGroupDelta are present
+  * 4 vi64: StartGroup, StartObject, EndGroupDelta, EndObject are present
+
+If only StartGroup is present, it is a relative number of groups prior to the next group,
```
afrind: Should we say 'next group' or 'Largest Object.Group + 1', since the 'next' Group could imply temporality (next to arrive) and/or skip a large range.

draft-ietf-moq-transport.md
```diff
@@ -1518,64 +1518,63 @@ end with an error.
...
+hence the start Location is `{Largest Object.Group + 1 - StartGroup, 0}`. For example:
+  * StartGroup=0 will start at the next group
+  * StartGroup=1 will start at the current group
+  * StartGroup=2 will start at 1 group prior to the current group
+  * StartGroup=N will start at N-1 groups prior to the current group
```
afrind, line 1577: Need to explain what happens when LargestObject.Group == MAXINT or < N.

draft-ietf-moq-transport.md
afrind: Remove "Next Object" in favor of just the explicit math, which is unambiguous.  I share Ian's concern about the overly clever encoding, but probably overall simplification outweighs the concern?

I would add something like "because a an open-ended range beginning at 0,0 is equivalent to no filter..."

draft-ietf-moq-transport.md
afrind: I think this text has been around and maybe Mo just moved it?

It's not strictly relay specific (except the network reordering bit) - a relay publishing out of order could also hit the condition.  But for now, let's leave it alone.

draft-ietf-moq-transport.md
afrind: Isn't that the inverse of joining fetch?  I want my subscribe to _end_ at live head?  I don't think we need this.

draft-ietf-moq-transport.md
afrind: Largest object or end of track

draft-ietf-moq-transport.md
afrind: Again, not sure referring to fetch without location = unfiltered is the right framing.  Better to say, it includes the entire track from the beginning to end or largest object.

draft-ietf-moq-transport.md
```diff
@@ -3618,27 +3613,24 @@ Joining Fetch {
...
-* Joining Start : A relative or absolute value used to determine the Start
-  Location, described below.
+A Location Filter parameter (see {{location-filters}}) MUST be included
+to specify the joining start Location as a relative or absolute value,
+described below.
 
 #### Joining Fetch Range Calculation
 
-The Joining Location value from the corresponding
+The `Largest Object` value from the corresponding
```
afrind: Why did we lose joining location in this PR?  Isn't it still needed?

draft-ietf-moq-transport.md
```diff
@@ -3742,18 +3734,11 @@ FETCH_OK Message {
...
-  using the same encoding as the FETCH request End Location (the last
-  Object, plus 1; or 0 to indicate the entire Group).
-  This is the End Location from the FETCH request unless
-  the requested range extends beyond published data:
-   - If the requested FETCH End Location was beyond the Largest known (possibly
-     final) Object, End Location is {Largest.Group, Largest.Object + 1}
-  Where Fetch.End Location is either Fetch.Standalone.End Location or the computed
-  End Location described in {{joining-fetch-range-calculation}}.
-
+* End Location: The end of the range covered by the FETCH response.
```
afrind, line 3773: I think I had to write 3 PRs to get the above text to resolve all the issues people had with it.

It looks like you are removing a special syntax from FETCH which was `FETCH Start=10,0 End=10,0` => fetch all of group 10.  Now you need `FETCH Location Param={10,0,0,MAXINT}`?  I'm not overly upset but we should run by the wg?

draft-ietf-moq-transport.md
```diff
@@ -3742,18 +3734,11 @@ FETCH_OK Message {
...
+  This is the End Location from the FETCH request Location Filter unless
+  the requested range extends beyond published data.
   If End Location is smaller than the Start Location in the corresponding FETCH
-  the receiver MUST close the session with a `PROTOCOL_VIOLATION`.
+  request Location Filter, the receiver MUST close the session with a `PROTOCOL_VIOLATION`.
```
afrind: Since end is delta encoded, this is no longer possible, though the delta being larger than 2^64 is.  I think it's covered well enough elsewhere though.

afrind (changes requested):
Thanks @mzanaty.  In general I'm ok with the approach, but some updates are needed.

draft-ietf-moq-transport.md
mzanaty: I also dislike this but nothing simpler and clearer came to mind. Do we really even need this at all? It seems only joining fetch/fill cases need this to start sub and end fetch on head to avoid dupes. Any other use of this filter? If not, maybe we can remove it if joining fetch/fill can be expressed without it.

draft-ietf-moq-transport.md
ianswett: Fair point, no need for this.

draft-ietf-moq-transport.md
afrind: Auth/Ed vibe is that we probably don't want implicit joining fetch/fill-fetch behavior.

Add some warning guidance about the footgun here.

draft-ietf-moq-transport.md
afrind: Another option - make the end _exclusive_ and the way to fetch all of group 10 collapses to FETCH 10,0 11,0.

I had suggested this way back when but folks we unenthused as I recall.

draft-ietf-moq-transport.md
mzanaty: Existing text, not moved, just reflowed. I'm taking the above guidance as leave it alone.

draft-ietf-moq-transport.md
mzanaty: At the end of a track, largest object = end of track, right? So largest object already covers all cases. Please reopen if I'm missing something.

draft-ietf-moq-transport.md
mzanaty: Unfiltered (no location filter) = beginning to end, seems equivalent. Unfiltered seems like the right framing, similar to HTTP range requests, you get the entire resource/track.

draft-ietf-moq-transport.md
mzanaty: The old text explains the +1 hack in fetch locations which was needed to reserve Obj ID 0 to mean "all objects in the group". End Obj ID is now optional, so it can be omitted instead of reserving 0 to mean "all objects in the group". So the special 0 / +1 hack is gone.

draft-ietf-moq-transport.md
mzanaty: Updated "without a filter" too.

draft-ietf-moq-transport.md
mzanaty: If there is a specific structure that applies only to a specific protocol element, I think it is much clearer to define both at the same time, rather than define a general structure unlinked to any protocol element and force the reader to find the protocol element context in which it applies. But eds/auths can use their discretion here.

draft-ietf-moq-transport.md
mzanaty: I recall this, but believe it was during simpler days when we thought it was easy to determine an entire group was delivered, before we had subgroups, gaps, filters, and other complexities. If Martin feels the old restriction still has value and is still needed, please reopen. A motivation for a full end location is Ian's suggestion to replace Forward with a Location filter with the exact end location to stop, even if in the past.

draft-ietf-moq-transport.md
mzanaty: Agreed, while equivalent, there is no need for this change, so I will revert.

draft-ietf-moq-transport.md
mzanaty: The fetch end object ID=0 hack to mean "all objects in the group" (and the resulting +1 actual location confusion) is no longer needed since end object ID is optional and can be omitted for "all objects in the group".

draft-ietf-moq-transport.md
mzanaty: I think this means End Location in Fetch OK compared to Start Location in Fetch, so no delta encoding there.

draft-ietf-moq-transport.md
mzanaty: Agreed, will reword without numbers to avoid confusion with Length.

draft-ietf-moq-transport.md
mzanaty: Next Group/Object should be defined first in terms of Largest Object, then used in capitals later. Old text already had Next Group, so I think it's intuitive and good to keep.

draft-ietf-moq-transport.md
mzanaty: Should we clamp computed group < 0 to 0 instead of error? If the sub asks for 1 group prior, but head is 0, should we clamp to 0 or error? I'm inclined to clamp to 0, since this seems like valid common app behavior to ask for 1 prior group when blind of head.

draft-ietf-moq-transport.md
mzanaty: I will leave 0,0 for now but add Alan's suggested text. I have ideas to remove it once fill/fetch lands.

draft-ietf-moq-transport.md
mzanaty: Any guidance? If not, I will clamp to 0/maxint for under/overflow, rather than error. When blind of head, asking for Next Group or Prior Group do not seem like errors or protocol violations, so just clamp to valid.

mzanaty (commented):
All comments addressed.

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,64 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
-observed at the publisher. If the specified `End Group Delta` is zero, the
-remainder of that Group passes the filter. Otherwise, the last Group ID to be
-delivered will be the Group ID in `Start Location` plus the `End Group Delta`.
-If the resulting Group ID would be greater than 2^64 - 1, the endpoint MUST
-close the session with a `PROTOCOL_VIOLATION`.
-
-An endpoint that receives a filter type other than the above MUST close the
-session with `PROTOCOL_VIOLATION`.
+Length (in bytes) determines how many optional vi64 fields are present.
+Length can be 0 for no filter to remove the filter in REQUEST_UPDATE.
```
```suggestion
Length can be 0 to indicate no filter, for example to remove the filter in REQUEST_UPDATE.
```

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,64 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
+or {0,0} if no content has been delivered yet.  To start at absolute Location {0, 0}
+with no end, which is equivalent to unfiltered, do not include a Location Filter.
+Note that due to network reordering or prioritization, relays can receive Objects with
+Locations smaller than `Largest Object` after the SUBSCRIBE is processed, but
+these Objects do not pass this filter.
+
+If a relative start group results in a computed absolute group less than 0, the
+computed value is set to 0; if greater than 2^64 - 1, it is set to 2^64 - 1.
+
+Otherwise, all start/end group/object fields are absolute.  EndGroupDelta is delta
```
```suggestion
Otherwise, all fields are absolute.  EndGroupDelta is delta
```

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,64 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
+encoded from StartGroup, but both the start and end groups are absolute, not
```
afrind, line 1591: So there's no encoding of "5 groups before and 5 groups after largest object".  I think this is fine, since you can say "5 groups before + open ended" and then REQUEST_UPDATE to set your end.

draft-ietf-moq-transport.md
```diff
@@ -3650,8 +3650,9 @@ Joining Fetch {
   (subscriber)` states, it MUST return a REQUEST_ERROR with error code
   `INVALID_JOINING_REQUEST_ID`.
 
-* Joining Start : A relative or absolute value used to determine the Start
-  Location, described below.
+A Location Filter parameter (see {{location-filters}}) MUST be included
+to specify the joining start Location as a relative or absolute value,
+described below.
```
```suggestion
described below.  If a publisher receives a Joining Fetch with no Location Filter parameter,
it MUST close the session with `PROTOCOL_VIOLATION`.
```

draft-ietf-moq-transport.md
```diff
@@ -3660,17 +3661,16 @@ subscription is used to calculate the end of a Joining Fetch, so the
...
+the Joining Location (see {{subscriptions}}.
 
 Note: the last Object included in the Joining FETCH response is the Object
-at the Joining Location.  The `+ 1` above indicates the equivalent Standalone
-Fetch encoding.
+at the Joining Location.
 
 For a Relative Joining Fetch, the publisher sets the Start Location to
-{Joining Location.Group - Joining Start, 0}.
+{Joining Location.Group + 1 - StartGroup, 0}.
```
afrind: Is StartGroup ambiguous in this context?  There's an encoding on the wire, which could be 0, and then there's the resolved value.

draft-ietf-moq-transport.md
```diff
@@ -3729,12 +3729,12 @@ cached objects have been delivered before resetting the stream.
...
-Fetch specifies an inclusive range of Objects starting at Start Location and
-ending at End Location. End Location MUST specify the same or a larger Location
-than Start Location for Standalone and Absolute Joining Fetches.
+Fetch includes a Location Filter parameter
+which specifies an inclusive range of Objects starting at Start Location and
+ending at End Location.
 
-Objects larger than the Largest Object will not be retrieved by a FETCH.  If the
-requested End Location exceeds the Largest available Object, the actual end of
+Locations larger than the `Largest Object` will not be retrieved by a FETCH.  If the
```
```suggestion
Objects with Locations larger than the `Largest Object` will not be retrieved by a FETCH.  If the
```

draft-ietf-moq-transport.md
```diff
@@ -3774,16 +3774,9 @@ FETCH_OK Message {
...
-  This is the End Location from the FETCH request unless
-  the requested range extends beyond published data:
-   - If the requested FETCH End Location was beyond the Largest known (possibly
-     final) Object, End Location is {Largest.Group, Largest.Object + 1}
-  Where Fetch.End Location is either Fetch.Standalone.End Location or the computed
-  End Location described in {{joining-fetch-range-calculation}}.
-
+* End Location: The end of the range covered by the FETCH response.
+  This is the End Location from the FETCH request Location Filter unless
+  the requested range extends beyond published data.
```
```suggestion
  the requested range extends beyond Largest Object or the last Object in the Track.
```

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,64 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
+A Location filter specifies an inclusive range of Locations.  Only objects
+with Locations within the inclusive range pass the filter.
 
 Some Location filters are defined to be relative to the `Largest Object`. The `Largest
 Object` is the Object with the largest Location ({{location-structure}}) in the
 Track from the perspective of the publisher processing the message. Largest
 Object updates when the first byte of an Object with a Location larger than the
 previous value is published or received through a subscription.
+Next Object is defined as {Largest Object.Group, Largest Object.Object + 1}.
+Next Group is defined as {Largest Object.Group + 1, 0}.
```
afrind: I'm not sure these are needed.  Both references below explicitly give the Location math.

afrind (commented):
Modulo inline suggestions, this seems ready.

draft-ietf-moq-transport.md
mzanaty: Correct

draft-ietf-moq-transport.md
mzanaty: StartGroup is the structure/wire value. The resolved value is called Start Location, not StartGroup. So there should be no ambiguity.

draft-ietf-moq-transport.md
mzanaty: Removed

mzanaty (commented):
Ready for review/merge.

afrind (approved):

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,62 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
-observed at the publisher. If the specified `End Group Delta` is zero, the
-remainder of that Group passes the filter. Otherwise, the last Group ID to be
-delivered will be the Group ID in `Start Location` plus the `End Group Delta`.
-If the resulting Group ID would be greater than 2^64 - 1, the endpoint MUST
-close the session with a `PROTOCOL_VIOLATION`.
-
-An endpoint that receives a filter type other than the above MUST close the
-session with `PROTOCOL_VIOLATION`.
+Length (in bytes) determines how many optional vi64 fields are present.
+Length can be 0 to indicate no filter, for example to remove the filter in REQUEST_UPDATE.
```
```suggestion
A length of 0 indicates no filter, for example to remove the filter in REQUEST_UPDATE.
```

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,62 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
+Optional fields can be omitted consecutively from the end.
```
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,62 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
+  * If only one is present, it is StartGroup.
+  * If only two are present, they are StartGroup and StartObject.
+  * If only three are present, they are StartGroup, StartObject, and EndGroupDelta.
```
```suggestion
  * If only one field is present, it is StartGroup.
  * If only two fields are present, they are StartGroup and StartObject.
  * If only three fields are present, they are StartGroup, StartObject, and EndGroupDelta.
```

draft-ietf-moq-transport.md
```diff
@@ -3650,8 +3647,10 @@ Joining Fetch {
   (subscriber)` states, it MUST return a REQUEST_ERROR with error code
   `INVALID_JOINING_REQUEST_ID`.
 
-* Joining Start : A relative or absolute value used to determine the Start
-  Location, described below.
+A Location Filter parameter (see {{location-filters}}) MUST be included
```
ianswett: I think having no Location filter should start from the beginning of a Track, so there's no reason for this extra MUST.

draft-ietf-moq-transport.md
```diff
@@ -3774,16 +3772,9 @@ FETCH_OK Message {
...
-  This is the End Location from the FETCH request unless
-  the requested range extends beyond published data:
-   - If the requested FETCH End Location was beyond the Largest known (possibly
-     final) Object, End Location is {Largest.Group, Largest.Object + 1}
-  Where Fetch.End Location is either Fetch.Standalone.End Location or the computed
-  End Location described in {{joining-fetch-range-calculation}}.
-
+* End Location: The end of the range covered by the FETCH response.
+  This is the End Location from the FETCH request Location Filter unless
+  the requested range extends beyond Largest Object or the last Object in the Track.
```
vasilvv: > the End Location from the FETCH request Location Filter

Replace this with something along the lines of "the Location of the last object that matches the Location Filter at the time the FETCH request is processed".

draft-ietf-moq-transport.md
```diff
@@ -3660,17 +3659,16 @@ subscription is used to calculate the end of a Joining Fetch, so the
...
+the Joining Location (see {{subscriptions}}.
 
 Note: the last Object included in the Joining FETCH response is the Object
-at the Joining Location.  The `+ 1` above indicates the equivalent Standalone
-Fetch encoding.
+at the Joining Location.
 
 For a Relative Joining Fetch, the publisher sets the Start Location to
-{Joining Location.Group - Joining Start, 0}.
+{Joining Location.Group + 1 - StartGroup, 0}.
```
```suggestion
{Joining Location.Group + 1 - StartGroup, 0} or {0, 0} if the Start Location's group
would be less than zero.
```

ianswett (approved):
Overall, this is looking good.

draft-ietf-moq-transport.md
```diff
@@ -3729,12 +3727,12 @@ cached objects have been delivered before resetting the stream.
 
 The Object Forwarding Preference does not apply to fetches.
 
-Fetch specifies an inclusive range of Objects starting at Start Location and
-ending at End Location. End Location MUST specify the same or a larger Location
-than Start Location for Standalone and Absolute Joining Fetches.
+Fetch includes a Location Filter parameter
+which specifies an inclusive range of Objects starting at Start Location and
+ending at End Location.
```
vasilvv, line 3729: This doesn't seem to be correct anymore?  An absolute fetch can have any filter, or no filter at all.

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,62 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
+computed value is set to 0; if greater than 2^64 - 1, it is set to 2^64 - 1.
+
+Otherwise, all fields are absolute.  EndGroupDelta is delta
+encoded from StartGroup, but both the start and end groups are absolute, not
+relative to `Largest Object`.  If StartGroup + EndGroupDelta exceeds 2^64 - 1,
+the endpoint MUST close the session with a `PROTOCOL_VIOLATION`.
+
+EndGroupDelta and EndObject can be omitted for an open-ended subscription.
+When they are omitted from a Fetch, the EndGroup and EndObject are `Largest Object`.
+EndObject can be omitted to include all objects in the End Group.
```
```suggestion
When EndGroupDelta and EndObject are omitted from a subscription filter, the
subscription is open-ended. When they are omitted from a Fetch, the
EndGroup and EndObject are `Largest Object`.

When EndObject is omitted, the filter includes all objects in the End Group.
```

draft-ietf-moq-transport.md
```diff
@@ -3729,12 +3727,12 @@ cached objects have been delivered before resetting the stream.
...
-Fetch specifies an inclusive range of Objects starting at Start Location and
-ending at End Location. End Location MUST specify the same or a larger Location
-than Start Location for Standalone and Absolute Joining Fetches.
+Fetch includes a Location Filter parameter
+which specifies an inclusive range of Objects starting at Start Location and
+ending at End Location.
 
-Objects larger than the Largest Object will not be retrieved by a FETCH.  If the
-requested End Location exceeds the Largest available Object, the actual end of
+Objects with Locations larger than the `Largest Object` will not be retrieved by a FETCH.  If the
```
vasilvv: `Largest Object` at the moment request is processed?

draft-ietf-moq-transport.md
```diff
@@ -3729,12 +3727,12 @@ cached objects have been delivered before resetting the stream.
 
 The Object Forwarding Preference does not apply to fetches.
 
-Fetch specifies an inclusive range of Objects starting at Start Location and
-ending at End Location. End Location MUST specify the same or a larger Location
-than Start Location for Standalone and Absolute Joining Fetches.
+Fetch includes a Location Filter parameter
```
```suggestion
Fetch can include a Location Filter parameter (see {{location-filters}})
```

draft-ietf-moq-transport.md
```suggestion
Objects with Locations larger than the `Largest Object` at the time the request
is processed will not be retrieved by a FETCH.  If the
```

draft-ietf-moq-transport.md
```diff
@@ -3650,8 +3646,10 @@ Joining Fetch {
   (subscriber)` states, it MUST return a REQUEST_ERROR with error code
   `INVALID_JOINING_REQUEST_ID`.
 
-* Joining Start : A relative or absolute value used to determine the Start
-  Location, described below.
+A Location Filter parameter (see {{location-filters}}) MUST be included
+to specify the joining start Location as a relative or absolute value,
+described below.  If a publisher receives a Joining Fetch with no Location Filter parameter,
+it MUST close the session with `PROTOCOL_VIOLATION`.
```
```suggestion
```

draft-ietf-moq-transport.md
```suggestion
  the requested range extends beyond Largest Object at the time
  the request was processed, or the last Object in the Track.
```

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,63 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
+  * If only three fields are present, they are StartGroup, StartObject, and EndGroupDelta.
+
+If only StartGroup is present, it is a relative number of groups prior to the Next Group,
+hence the start Location is `{Largest Object.Group + 1 - StartGroup, 0}`. For example:
+  * StartGroup=0 will start at the Next Group
+  * StartGroup=1 will start at the current group
+  * StartGroup=2 will start at 1 group prior to the current group
+  * StartGroup=N will start at N-1 groups prior to the current group
+
+If only StartGroup and StartObject are present and both 0, the start Location
```
martinduke, line 1579: so how would I express the old NextGroupStart filter? I would think that would be (0, 0)!

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,63 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
-An endpoint that receives a filter type other than the above MUST close the
-session with `PROTOCOL_VIOLATION`.
+Length (in bytes) determines how many optional vi64 fields are present.
+A length of 0 indicates no filter, for example to remove the filter in REQUEST_UPDATE.
+  * If only one field is present, it is StartGroup.
+  * If only two fields are present, they are StartGroup and StartObject.
+  * If only three fields are present, they are StartGroup, StartObject, and EndGroupDelta.
+
+If only StartGroup is present, it is a relative number of groups prior to the Next Group,
+hence the start Location is `{Largest Object.Group + 1 - StartGroup, 0}`. For example:
```
martinduke, line 1573: This paragraph should explicit on what happens if there is no LargestObject, rather than relying on the examples below.

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,63 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
+computed value is set to 0; if greater than 2^64 - 1, it is set to 2^64 - 1.
+
+Otherwise, all fields are absolute.  EndGroupDelta is delta
+encoded from StartGroup, but both the start and end groups are absolute, not
+relative to `Largest Object`.  If StartGroup + EndGroupDelta exceeds 2^64 - 1,
+the endpoint MUST close the session with a `PROTOCOL_VIOLATION`.
+
+When EndGroupDelta and EndObject are omitted from a subscription filter, the
+subscription is open-ended. When they are omitted from a Fetch, the
+EndGroup and EndObject are `Largest Object`.
```
martinduke, line 1597: so IIUC, to Fetch from 0,0 to Largest Object, there is no filter at all? That would be a useful clarifying example.

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,63 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
+
```
martinduke, line 1598: Also, it would seem that we've eliminated the ability to start an open-ended subscription at an arbitrary point in the future? Unless we set EndGroup = UINT64_MAX? If so, that's another useful clarifying example.

draft-ietf-moq-transport.md
martinduke, line 3706: "in the requested range" is a weird thing to say here, since the struct above doesn't have any sort of range. Better to say that a FETCH delivers all previously published objects, subject to any parameters that might constrain the set.

draft-ietf-moq-transport.md
mzanaty: Just {0} so only StartGroup is present as stated immediately above.

draft-ietf-moq-transport.md
mzanaty: Nothing was eliminated. The old AbsoluteStart filter is now {X,Y} where absolute group X can be in the future or past. Just give a start with no end.

draft-ietf-moq-transport.md
mzanaty: Leaving to Ed's discretion.

draft-ietf-moq-transport.md
mzanaty: Leaving to Ed's discretion. Maybe "requested range in the Location Filter parameter if present".

mzanaty (commented):
Not sure any changes are needed, but leaving to Ed's discretion.

draft-ietf-moq-transport.md
suhasHere: I feel the filters applies only when there is a largest object, if not, we need to wait for largest object to be know to apply location filters. Not sure, if we need to state it explicitly. But i do, some clairifcaiton might not be harmful @afrind @ianswett  thoughts ?

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,63 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
-delivered will be the Group ID in `Start Location` plus the `End Group Delta`.
-If the resulting Group ID would be greater than 2^64 - 1, the endpoint MUST
-close the session with a `PROTOCOL_VIOLATION`.
-
-An endpoint that receives a filter type other than the above MUST close the
-session with `PROTOCOL_VIOLATION`.
+Length (in bytes) determines how many optional vi64 fields are present.
+A length of 0 indicates no filter, for example to remove the filter in REQUEST_UPDATE.
+  * If only one field is present, it is StartGroup.
+  * If only two fields are present, they are StartGroup and StartObject.
```
suhasHere, line 1569:
```suggestion
  * If only two fields are present, they are StartGroup followed by StartObject.
```

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,63 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
+  * If only three fields are present, they are StartGroup, StartObject, and EndGroupDelta.
```
suhasHere, line 1570:
```suggestion
  * If only three fields are present, they are StartGroup followed by StartObject, and EndGroupDelta.
```

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,63 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
+
+If only StartGroup is present, it is a relative number of groups prior to the Next Group,
+hence the start Location is `{Largest Object.Group + 1 - StartGroup, 0}`. For example:
+  * StartGroup=0 will start at the Next Group
+  * StartGroup=1 will start at the current group
+  * StartGroup=2 will start at 1 group prior to the current group
+  * StartGroup=N will start at N-1 groups prior to the current group
```
suhasHere, line 1577: I wonder if we need to clarify 
StartGroup=N will start at N-1 groups prior to the current group, for N > 0

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,63 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
+
+If only StartGroup and StartObject are present and both 0, the start Location
+is the Next Object which is `{Largest Object.Group, Largest Object.Object + 1}`,
+or {0,0} if no content has been delivered yet.  To start at absolute Location {0, 0}
+with no end, which is equivalent to unfiltered, do not include a Location Filter.
+Note that due to network reordering or prioritization, relays can receive Objects with
+Locations smaller than `Largest Object` after the SUBSCRIBE is processed, but
+these Objects do not pass this filter.
+
+If a relative start group results in a computed absolute group less than 0, the
```
suhasHere, line 1587: May be we don't needs this, since we are referring unsigned integers and it can't go less than 0 ?

draft-ietf-moq-transport.md
```diff
@@ -3774,16 +3770,10 @@ FETCH_OK Message {
...
-  Object, plus 1; or 0 to indicate the entire Group).
-  This is the End Location from the FETCH request unless
-  the requested range extends beyond published data:
-   - If the requested FETCH End Location was beyond the Largest known (possibly
-     final) Object, End Location is {Largest.Group, Largest.Object + 1}
-  Where Fetch.End Location is either Fetch.Standalone.End Location or the computed
-  End Location described in {{joining-fetch-range-calculation}}.
-
+* End Location: The end of the range covered by the FETCH response.
+  This is the End Location from the FETCH request Location Filter unless
```
```suggestion
  This is the End Location from the FETCH request Location Filter parameter unless
```

suhasHere (approved):
Thanks @mzanaty  for the PR.

this is looking fine overall. I do have few editorial clarifications into following categoreis
 
- Any computation going < 0
- I don't see a need for 4 examples of StartGroup is needed as the formula seems fine
- Optionally adding LargestLocation in filter text might not be a bad idea ..

draft-ietf-moq-transport.md
afrind: @suhasHere the computed value can be less than 0 if current group is e.g. 10 and StartGroup is 20.

draft-ietf-moq-transport.md
afrind: All N are greater than 0, not sure what you are saying

draft-ietf-moq-transport.md
afrind: @suhasHere that's an interesting idea.  I don't love the idea of deferring filter resolution.  I'd rather just default to Next Group or 0,0.  This is the case where no objects have been published yet.

draft-ietf-moq-transport.md
afrind: I think it's clear from the text and the diagram what the order is.

draft-ietf-moq-transport.md
afrind: same as above.

draft-ietf-moq-transport.md
```diff
@@ -1539,60 +1539,63 @@ Subscribers can specify a Location filter on a subscription indicating to the pu
...
 ~~~
-Location Filter {
-  Filter Type (vi64),
-  [Start Location (Location),]
-  [End Group Delta (vi64),]
-}
+LOCATION_FILTER Parameter {
+  Parameter Type (vi64) = 0x21,
+  Length (vi64),
+  [StartGroup (vi64),]
```
fluffy, line 1560: Start seems like a bad name. I think it more a count of Prior objects or groups that you add to current to get the StartGroup.

fluffy (changes requested):

draft-ietf-moq-transport.md
mzanaty: If no objects have been published, I think the correct resolution of any relative start is simply unfiltered. Nothing else makes sense, since there is no current group or prior groups, or any way to compute any relative group until an object is eventually published. Deferred resolution is an unnecessary burden that would anyway produce the wrong result. If you ask for Next Group when nothing was published, then finally object 5,0 is published, you would want to receive it, but deferred resolution would set Largest Object to 5,0 and Next Group to 6, so you would not get it. Unfiltered is what you likely want for a relative start if nothing published, to avoid the confusion in deferred resolution.

draft-ietf-moq-transport.md
mzanaty: The extra "for N>0" seems harmless, although perhaps superfluous. Should it be N>0 or N>1? Do you consider "0 groups prior to the current group" to be nonsensical or clearly mean current group? If the latter, we could also consider wording relative to next not current group to eliminate N-1: "StartGroup=N will start at N groups prior the Next Group".

draft-ietf-moq-transport.md
mzanaty: The computed value depends on an unknown Largest Object (only known by the peer), so relative start can go negative.

draft-ietf-moq-transport.md
mzanaty: Correct, for standalone fetch. Joining fetch says you MUST provide a filter, but Ian wants to remove this so it behaves just like standalone fetch: unfiltered up to Largest Object (or Track End, which I guess is also Largest Object when the track ends.) I agree with Ian, remove the MUST, and Martin, add clarifying text for this unfiltered case.

draft-ietf-moq-transport.md
mzanaty: StartGroup can be relative (if it's the only field present) or absolute (if more fields present). Need a name that works for both cases. Open to anything better.

draft-ietf-moq-transport.md
mzanaty: Agreed, MUST removed

vasilvv (approved):

PR #1813, opened 2026-07-15
Add an INCLUDE_PROPERTIES Parameter

### LINKED ISSUE #1644: Duplicate Track Properties on repeated SUBSCRIBE_OK / FETCH_OK
I don't think this is an issue worth fixing for v1, but I wanted to raise it.

Track properties are meant to be fixed for the lifetime of a track. They can be cached indefinitely, at least in theory.

However, each SUBSCRIBE_OK and FETCH_OK transmits a fresh copy of the Track Properties even if the subscriber already has that information cached. The most egregious example would be a FETCH for each group; we keep transmitting redundant track properties. 

This could get quite inefficient if track properties get large, like for folks using them as a replacement for the catalog. Imagine a FETCH for an audio group and each FETCH_OK contains the full codec/description/timescale/etc information.

ianswett:
Allows Subscribers to indicate whether they want the Track Properties.

Fixes #1644

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2824,6 +2824,16 @@ Namespace Prefix for that subscription.  If the new prefix would share a common 
 another active subscription of the same type in the same session, the receiver
 MUST respond with REQUEST_ERROR with error code `PREFIX_OVERLAP`.
 
+### INCLUDE_PROPERTIES Parameter {#include-properties-param}
+
+The INCLUDE_PROPERTIES parameter (Parameter Type 0x35) is a uint8. It MAY appear
+in SUBSCRIBE, TRACK_STATUS, FETCH or SUBSCRIBE_TRACKS. It specifies whether the
+OK message sent in response includes Track Properties or whether the resulting PUBLISH
+messages include Track Properties in the case of SUBSCRIBE_TRACKS. The allowed
+values are 0 (do not send Properties) or 1 (send Properties), and the default is 1.
+If an endpoint receives a value outside this range, it MUST close the session
+with `PROTOCOL_VIOLATION`.
+
 ## SETUP {#message-setup}
 
 The `SETUP` message is the first message each endpoint sends on its control
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2833,7 +2833,8 @@ MUST respond with REQUEST_ERROR with error code `PREFIX_OVERLAP`.
 The INCLUDE_PROPERTIES parameter (Parameter Type 0x35) is a uint8. It MAY appear
 in SUBSCRIBE, TRACK_STATUS, FETCH or SUBSCRIBE_TRACKS. It specifies whether the
 OK message sent in response includes Track Properties or whether the resulting PUBLISH
-messages include Track Properties in the case of SUBSCRIBE_TRACKS. The allowed
+messages include Track Properties in the case of SUBSCRIBE_TRACKS. The Track Properties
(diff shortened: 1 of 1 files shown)


afrind (approved):

draft-ietf-moq-transport.md
```diff
@@ -2824,6 +2824,16 @@ Namespace Prefix for that subscription.  If the new prefix would share a common
...
 
+### INCLUDE_PROPERTIES Parameter {#include-properties-param}
+
+The INCLUDE_PROPERTIES parameter (Parameter Type 0x35) is a uint8. It MAY appear
+in SUBSCRIBE, TRACK_STATUS, FETCH or SUBSCRIBE_TRACKS. It specifies whether the
+OK message sent in response includes Track Properties or whether the resulting PUBLISH
+messages include Track Properties in the case of SUBSCRIBE_TRACKS. The allowed
+values are 0 (do not send Properties) or 1 (send Properties), and the default is 1.
+If an endpoint receives a value outside this range, it MUST close the session
+with `PROTOCOL_VIOLATION`.
```
suhasHere: This should apply only to fetch and susbcribe_tracks, since same subscriber can issue multiple fetches to the same track and subscribe_tracks can have mulitples PUBLISHES ,

I don't see subscribe should be in the list

afrind: If I've already subscribed to a track in the past (eg switching back and forth between renditions), I don't need it's properties again.

suhasHere (approved):

draft-ietf-moq-transport.md
suhasHere: this makes sense

vasilvv (commented):
What does it mean to omit the properties?  The field is currently mandatory by syntax.  Does it mean just publish an empty block?

ianswett:
> What does it mean to omit the properties? The field is currently mandatory by syntax. Does it mean just publish an empty block?

Yes, it means to publish an empty block, not change the framing.

vasilvv:
> > What does it mean to omit the properties? The field is currently mandatory by syntax. Does it mean just publish an empty block?
> 
> Yes, it means to publish an empty block, not change the framing.

Please clarify that in the text.

ianswett:
> > > What does it mean to omit the properties? The field is currently mandatory by syntax. Does it mean just publish an empty block?
> > 
> > 
> > Yes, it means to publish an empty block, not change the framing.
> 
> Please clarify that in the text.

Done

draft-ietf-moq-transport.md
```diff
@@ -2835,6 +2835,17 @@ Namespace Prefix for that subscription.  If the new prefix would share a common
 another active subscription of the same type in the same session, the receiver
 MUST respond with REQUEST_ERROR with error code `PREFIX_OVERLAP`.
 
+### INCLUDE_PROPERTIES Parameter {#include-properties-param}
+
+The INCLUDE_PROPERTIES parameter (Parameter Type 0x35) is a uint8. It MAY appear
+in SUBSCRIBE, TRACK_STATUS, FETCH or SUBSCRIBE_TRACKS. It specifies whether the
+OK message sent in response includes Track Properties or whether the resulting PUBLISH
+messages include Track Properties in the case of SUBSCRIBE_TRACKS. The Track Properties
+are still present in the message, but they SHOULD be empty. The allowed
```
vasilvv: > The Track Properties are still present in the message, but they SHOULD be empty.

"If INCLUDE_PROPERTIES is 0, ...."

vasilvv (approved):
LGTM once the comment is addressed.

PR #1817, opened 2026-07-19
Define host resolution for moqt URIs

### LINKED ISSUE #1696: URI resolution
The draft says 

> Discovery of MOQT servers is always done out-of-band.

but I don't think this is quite the case. I think we need to describe how that works. If we were doing HTTP over again, I think we would use SRV records. I think the raw quic should use SRV as it gives us better load balancing capabilities. I realize websockets may be more limited.

afrind:
Specify that the host is resolved to network addresses (A/AAAA) and allow optional use of SVCB/HTTPS records to discover endpoints and supported ALPN protocols before connecting.

Fixes: #1696

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -931,6 +931,15 @@ TODO: Add URI scheme security considerations per RFC 7595 Section 3.7
 
 TODO: Add internationalization statement per RFC 7595 Section 3.6.
 
+The client resolves the `host` subcomponent of the `authority` to one or
+more network addresses, most commonly using DNS A and AAAA records.
+
+When SVCB-compatible records {{?RFC9460}} are published for the `authority`,
+a client MAY use them to learn the server's endpoints and supported ALPN
+protocols before connecting. A client using WebTransport resolves the
+`https` URI derived in {{webtransport}} using HTTPS resource records as for
+any `https` origin.
+
 If the port is omitted in the URI, a default port of 443 is used.
 
 The client MAY use either native QUIC or WebTransport. On a QUIC connection,
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -940,6 +940,7 @@ protocols before connecting. A client using WebTransport resolves the
 `https` URI derived in {{webtransport}} using HTTPS resource records as for
 any `https` origin.
 TODO: reference moqt SVCB record draft once available.
+
 If the port is omitted in the URI, a default port of 443 is used.
 
 The client MAY use either native QUIC or WebTransport. On a QUIC connection,
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -939,7 +939,7 @@ a client MAY use them to learn the server's endpoints and supported ALPN
 protocols before connecting. A client using WebTransport resolves the
 `https` URI derived in {{webtransport}} using HTTPS resource records as for
 any `https` origin.
-
+TODO: reference moqt SVCB record draft once available.
 If the port is omitted in the URI, a default port of 443 is used.
 
 The client MAY use either native QUIC or WebTransport. On a QUIC connection,
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -932,7 +932,7 @@ TODO: Add URI scheme security considerations per RFC 7595 Section 3.7
 TODO: Add internationalization statement per RFC 7595 Section 3.6.
 
 The client resolves the `host` subcomponent of the `authority` to one or
-more network addresses, most commonly using DNS A and AAAA records.
+more network addresses, most commonly using DNS A {{?RFC1035}} and AAAA {{?RFC3596}} records.
 
 When SVCB-compatible records {{?RFC9460}} are published for the `authority`,
 a client MAY use them to learn the server's endpoints and supported ALPN


draft-ietf-moq-transport.md
```diff
@@ -931,6 +931,15 @@ TODO: Add URI scheme security considerations per RFC 7595 Section 3.7
 
 TODO: Add internationalization statement per RFC 7595 Section 3.6.
 
+The client resolves the `host` subcomponent of the `authority` to one or
+more network addresses, most commonly using DNS A and AAAA records.
```
ianswett: Do we need a reference here?

ianswett (approved):

suhasHere:
This seems fine, but looking at original issue, it doesn't address the SRV record ?

afrind:
> This seems fine, but looking at original issue, it doesn't address the SRV record ?

True, but that should be defined in it's _own_ document, not this one.  Once there is a MOQ SRV record, we can reference it here.

bemasc:
This is a bit underspecified, as others have noted.  We should either expand this to encompass all the things that a "SVCB mapping document" requires (name structure, default ALPNs, automatically mandatory keys, etc.) or put that in a separate document and reference it here.

afrind:
> This is a bit underspecified, as others have noted. We should either expand this to encompass all the things that a "SVCB mapping document" requires (name structure, default ALPNs, automatically mandatory keys, etc.) or put that in a separate document and reference it here.

@bemasc - my thought was to define SVCB mapping in another doc, then reference it here.

suhasHere:
> @bemasc - my thought was to define SVCB mapping in another doc, then reference it here.

@fluffy  and I have been working on SVCB for moqt. We should have something to share soon for review. @afrind  if you think you can assign this issue/pr to me and I can update ?

bemasc:
OK.  I also put some text for this in a google doc somewhere so LMK if you want text.

afrind:
@suhasHere

My preference is that SVCB go in its own document.  If moq-wg adopts it we can reference in moq-transport, but I'd rather not ref an ID, or hold this PR based on that.  Not hard to add a ref as an editorial PR.

Below is the text @bemasc sent me.

MOQ SVCB Mapping

Service owners MAY publish SVCB records for the `authority` using Port-Prefix Naming ({{?RFC9460}}, Section 2.3).  Clients MAY use these records to learn the server's endpoints and relevant service parameters, such as the supported ALPN protocols, before connecting. A client using WebTransport resolves the `https` URI derived in {{webtransport}} using HTTPS resource records as for
any `https` origin.

The only default ALPN is “moqt”.  ((NOTE: This ALPN value is not assigned yet.  Testing of draft versions will therefore require use of the “no-default-alpn” SvcParam.))

All SvcParams typically applicable to QUIC or TLS are also applicable to these SVCB records.

Examples
;; A and AAAA records are omitted for brevity.

;; Minimal endpoint: Standard MOQT over QUIC.
_443._moqt.moq.example. IN SVCB 1 .

;; Testing endpoint for a draft version of MOQT with QUIC and QMUX
_443._moqt.moq.example. IN SVCB 1 . no-default-alpn alpn=moqt-20,moqt-20-qx

;; A complex deployment with WebTransport, aliases, ECH, and QUIC+TCP primary with TCP-only fallback
_443._moqt.moq.example. IN SVCB 0 moq-provider.example.
moq-provider.example. IN SVCB 1 . ech=ABC… alpn=moqt-qx
moq-provider.example. IN SVCB 2 backup.moq-provider.example. ech=ABC… no-default-alpn alpn=moqt-qx
moq.example. IN HTTPS 0 webtransport.moq-provider.example.
webtransport.moq-provider.example. IN HTTPS 1 moq-provider.example. ech=ABC… alpn=h2,h3

Mapping Summary

Mapped scheme:  "moqt"
Other affected schemes: None
RR type: SVCB (64)
Name prefix: _$PORT._moqt 
Automatically mandatory keys: port, no-default-alpn
SvcParam defaults:  alpn: ["moqt"]
Special behaviors:  Replaced by HTTPS record when using WebTransport
Keys that records must include: None

suhasHere:
@afrind I am ok with it , but want to highlight that this PR doesn't fully address the issue #1696 though , but only partly

suhasHere:
Thanks for sharing the text . Cullen and I got something similar written down as well . We will include the text from @bemsac

draft-ietf-moq-transport.md
```suggestion
more network addresses, most commonly using DNS A {{?RFC1035}} and AAAA {{?RFC3596}} records.
```

draft-ietf-moq-transport.md
```diff
@@ -931,6 +931,15 @@ TODO: Add URI scheme security considerations per RFC 7595 Section 3.7
...
 
+The client resolves the `host` subcomponent of the `authority` to one or
+more network addresses, most commonly using DNS A {{?RFC1035}} and AAAA {{?RFC3596}} records.
+
+When SVCB-compatible records {{?RFC9460}} are published for the `authority`,
+a client MAY use them to learn the server's endpoints and supported ALPN
+protocols before connecting. A client using WebTransport resolves the
+`https` URI derived in {{webtransport}} using HTTPS resource records as for
+any `https` origin.
+
```
```suggestion
TODO: reference moqt SVCB record draft once available.
```

draft-ietf-moq-transport.md
```diff
@@ -931,6 +931,15 @@ TODO: Add URI scheme security considerations per RFC 7595 Section 3.7
...
 
+The client resolves the `host` subcomponent of the `authority` to one or
+more network addresses, most commonly using DNS A {{?RFC1035}} and AAAA {{?RFC3596}} records.
+
+When SVCB-compatible records {{?RFC9460}} are published for the `authority`,
+a client MAY use them to learn the server's endpoints and supported ALPN
+protocols before connecting. A client using WebTransport resolves the
+`https` URI derived in {{webtransport}} using HTTPS resource records as for
+any `https` origin.
+TODO: reference moqt SVCB record draft once available.
```
afrind, line 942:
```suggestion
TODO: reference moqt SVCB record draft once available.

```

suhasHere (approved):

vasilvv (approved):

PR #1820, opened 2026-07-21
Add PUBLISH_STATE_NOTIFY message

afrind:
Add a new control message, PUBLISH_STATE_NOTIFY (0x22), that a publisher sends on a subscription's bidirectional stream to notify the subscriber that subscription state has changed for a reason other than a subscriber-sent REQUEST_UPDATE. Unlike REQUEST_UPDATE, it is a unilateral notification: no REQUEST_OK/REQUEST_ERROR response and not subject to MAX_REQUEST_UPDATES. The FORWARD and LOCATION_FILTER parameters are updated to permit appearing in this message.

No existing mechanism in the draft sends this message yet. It is intended to support:

- SWITCH_FROM (#1674, #1675)
- Top-N (#1518)
- Possibly Default Publisher Priority (#1770)


suhasHere (approved):

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -2338,6 +2338,8 @@ new request stream.
 |--------|------------------------------------------------|------------------|
 | 0x4    | SUBSCRIBE_OK ({{message-subscribe-ok}})        | Request          |
 |--------|------------------------------------------------|------------------|
+| 0x22   | SUBSCRIPTION_STATE_UPDATE ({{ss-update}})      | Request          |
```
fluffy: Mo made comment publish notifiy or something like that

fluffy (changes requested):
Add more to make this clear what can be changed and what a relay needs to do when it gets this information

afrind:
Discussed in Vienna: we'll look at renaming the message and ensure it's clear what can be included, and when, and take a look about possible control message storms.

ianswett:
Could/Should this include Largest Object, so we can tell what Object the Subscription state changed?

draft-ietf-moq-transport.md
```diff
@@ -2356,6 +2356,8 @@ new request stream.
 |--------|------------------------------------------------|------------------|
 | 0x4    | SUBSCRIBE_OK ({{message-subscribe-ok}})        | Request          |
 |--------|------------------------------------------------|------------------|
+| 0x22   | PUBLISH_NOTIFY ({{ss-update}})                 | Request          |
```
```suggestion
| 0x22   | PUBLISH_NOTIFY ({{message-publish-notify}})          | Request          |
```

draft-ietf-moq-transport.md
```diff
@@ -3412,6 +3419,49 @@ updated prefix.  Updating the prefix of a SUBSCRIBE_TRACKS has
 no effect on existing subscriptions.  If the subscriber is no longer
 interested it can cancel the corresponding bidirectional stream.
 
+## PUBLISH_NOTIFY {#ss-update}
+
+A publisher sends PUBLISH_NOTIFY on a subscription's bidirectional
+stream to notify the subscriber the state of the subscription has changed for a
```
```suggestion
stream to notify the subscriber that the state of the subscription has changed for a
```

draft-ietf-moq-transport.md
```diff
@@ -3412,6 +3419,49 @@ updated prefix.  Updating the prefix of a SUBSCRIBE_TRACKS has
...
+other request type, or from the subscriber, MUST close the session with a
+`PROTOCOL_VIOLATION`.
+
+A PUBLISH_NOTIFY carries the parameters whose values have changed.
+If a parameter is not present, its value is unchanged.  The semantics of each
+parameter, including whether it may appear in PUBLISH_NOTIFY, are
+defined by the parameter.
+
+PUBLISH_NOTIFY reports the current subscription state; it does not modify
+parameters chosen by the subscriber.  A publisher MUST NOT use PUBLISH_NOTIFY to
```
ianswett: I think this first sentence is potentially confusing, so I'd rather drop it.

```suggestion
A publisher MUST NOT use PUBLISH_NOTIFY to
```

draft-ietf-moq-transport.md
```diff
@@ -3412,6 +3419,49 @@ updated prefix.  Updating the prefix of a SUBSCRIBE_TRACKS has
...
+change the value of a subscription parameter that is under the subscriber's
+control.  Such parameters can only be changed by the subscriber via
```
ianswett: This is a bit vague to me.  Everything is under the subscriber's control, but there are cases its asking the Publisher to manage the state of the subscription.

ianswett (commented):
PUBLISH_NOTIFY makes me think that you're notifying a subscriber of a PUBLISH, as in one that's coming, not a change to subscription state.

I'm ok with the name, but there might be better options?  SUBSCRIPTION_UPDATED?

martinduke:
IIUC the sole purpose of this is to communicate unilateral changes to FORWARD. Under what circumstances would this happen? I can only think of Top-N.

afrind:
> IIUC the sole purpose of this is to communicate unilateral changes to FORWARD. Under what circumstances would this happen? I can only think of Top-N.

If you include informing the peer of the new end group of the Location filter, it’s also useful for SSTS and SWITCH_FROM

draft-ietf-moq-transport.md
```diff
@@ -3412,6 +3419,48 @@ updated prefix.  Updating the prefix of a SUBSCRIBE_TRACKS has
...
+
+A PUBLISH_NOTIFY carries the parameters whose values have changed.
+If a parameter is not present, its value is unchanged.  The semantics of each
+parameter, including whether it may appear in PUBLISH_NOTIFY, are
+defined by the parameter.
+
+A publisher MUST NOT use PUBLISH_NOTIFY to
+change the value of a subscription parameter that is under the subscriber's
+control.  Such parameters can only be changed by the subscriber via
+REQUEST_UPDATE ({{message-request-update}}).
```
```suggestion
A publisher MUST NOT use PUBLISH_NOTIFY to
change the value of a subscriber controlled subscription parameter
unless the subscriber requested the change.
```

draft-ietf-moq-transport.md
afrind: @ianswett see my suggestion below.  It probably needs an example pointing to something other than REQUEST_UPDATE (e.g. SWITCH_FROM or SSTS), but none of those are here yet. I didn't see how best to fix it.

fluffy (changes requested):
I still think this needs information about what a relay does when it receives this.  Am I missing something ?

afrind:
> I still think this needs information about what a relay does when it receives this. Am I missing something ?

This message is strictly informative, and is only sent when the subscriber requested some deferred action (eg: top-n deselected a track).  No specific action is required by the receiver.  I can add a line to this effect.

draft-ietf-moq-transport.md
```diff
@@ -3412,6 +3419,47 @@ updated prefix.  Updating the prefix of a SUBSCRIBE_TRACKS has
...
+defined by the parameter.
+
+A publisher MUST NOT use PUBLISH_NOTIFY to
+change the value of a subscriber controlled subscription parameter
+unless the subscriber requested the change.
+
+The publisher MUST include the LARGEST_OBJECT parameter ({{largest-param}}), if
+known, in PUBLISH_NOTIFY so the subscriber can determine the point in the
+Track at which the change took effect.
+
```
afrind, line 3451:
```suggestion

This message is informative and no action is required by the recipient.

```

vasilvv:
> PUBLISH_NOTIFY makes me think that you're notifying a subscriber of a PUBLISH, as in one that's coming, not a change to subscription state.

I have a similar concern. PUBLISH_STATE_NOTIFY?

draft-ietf-moq-transport.md
```diff
@@ -2702,11 +2704,14 @@ the Track is used. If omitted from FETCH, the receiver uses Ascending (0x1).
...
 The LOCATION_FILTER parameter (Parameter Type 0x21) uses length-prefixed
-encoding. It MAY appear in a SUBSCRIBE or REQUEST_UPDATE (for a
-subscription) message. It is a Location Filter (see {{location-filters}}).
+encoding. It MAY appear in a SUBSCRIBE, PUBLISH_OK, REQUEST_UPDATE (for a
+subscription) or PUBLISH_STATE_NOTIFY message. It is a Location Filter (see
+{{location-filters}}).
 
 If omitted from SUBSCRIBE, the subscription is
-unfiltered.  If omitted from REQUEST_UPDATE, the value is unchanged.
+unfiltered.  If omitted from REQUEST_UPDATE or PUBLISH_STATE_NOTIFY, the
```
vasilvv, line 2714: Since you added PUBLISH_OK above (was that intentional?), should it also be mentioned here?

vasilvv (approved):
LGTM but I do have one clarification question.

draft-ietf-moq-transport.md
afrind: Removed from both

draft-ietf-moq-transport.md
```diff
@@ -2786,18 +2792,20 @@ PUBLISH, or REQUEST_UPDATE_OK
 
 The FORWARD parameter (Parameter Type 0x10) is a uint8. It MAY appear in
 SUBSCRIBE, REQUEST_UPDATE (for a subscription or a SUBSCRIBE_TRACKS request),
-PUBLISH and SUBSCRIBE_TRACKS. It specifies the Forwarding State on
-affected subscriptions (see {{subscriptions}}). The allowed values are 0 (don't
-forward) or 1 (forward). If an endpoint receives a value outside this range, it
-MUST close the session with `PROTOCOL_VIOLATION`.
+PUBLISH, PUBLISH_OK, SUBSCRIBE_TRACKS and PUBLISH_STATE_NOTIFY. It
```
```suggestion
PUBLISH, SUBSCRIBE_TRACKS and PUBLISH_STATE_NOTIFY. It
```

PR #1822, opened 2026-07-22
Add Timed-Out gap status for expired Fill Timeout

### LINKED ISSUE #1720: Fill Timeout reduction by relay
The draft says 

> If
> the subscriber specifies a Fill Timeout larger than the relay is
> willing to wait, the relay MAY use a shorter timeout without
> informing the subscriber.

I don't like the idea that relay can use a shorter time, perhaps 0, without the subscriber knowing that happened. Could we design a way to inform what was used, or negotiate a lower bound that will not be shortened, or return an error when this happens. 

Clients will need this information to be able to sort out why a request failed and what to do about it.

afrind:
Report FETCH gaps caused by Fill Timeout expiry as a distinct End of Timed-Out Range (0x20C), separate from End of Unknown Range, so a subscriber can tell Objects the relay stopped waiting for apart from Objects whose status is genuinely unknown.

Fixes: #1720


draft-ietf-moq-transport.md
```diff
@@ -4574,14 +4559,14 @@ MUST ignore the bits.
...
 Properties are not present. All Objects with Locations between the last
 serialized Object, if any, and this Location, inclusive, either do not exist
-(when Serialization Flags is 0x8C) or are unknown (0x10C).  A publisher SHOULD
-NOT use `End of Non-Existent Range` in a FETCH response except to split a range
-of Objects that will not be serialized into those that are known not to exist
-and those with unknown status.
+(when Serialization Flags is 0x8C), are unknown (0x10C), or timed out (0x20C).  A
+publisher SHOULD NOT use `End of Non-Existent Range` in a FETCH response except to
+split a range of Objects that will not be serialized into those that are known
+not to exist and those with unknown status.
```
sharmafb: How about we change this to "distinguish a known non-existent subrange from adjacent unknown or timed-out subranges"?

ianswett: I agee that it seems like we should mention timed-out in this sentence.

draft-ietf-moq-transport.md
```diff
@@ -4558,14 +4559,14 @@ MUST ignore the bits.
...
```
```suggestion
not to exist and those with unknown or timed out status.
```

suhasHere (approved):

ianswett (approved):

sandarsh (approved):

sharmafb (approved):

Issue #1803, opened 2026-07-07
How do you know what Parameters a PUBLISH has?


ianswett:
Fixing #1660 allows SUBSCRIBE parameters on SUBSCRIBE_TRACKS, but it's also possible to update those parameters.

How does a Subscriber know what Parameters were used to create the subscription?  There are two cases, FORWARD and GROUP_ORDER where we echo that back in the PUBLISH today.

There's also the case when a PUBLISH arrives that is not associated with a SUBSCRIBE_TRACKS.  It seems somewhat important to know the effective values for other Parameters, such as SUBSCRIBER_PRIORITY, as well as FORWARD and GROUP_ORDER?

In the virtual interim on July 6th, some people were concerned about creating another error condition if a Publisher specifies a Parameter that doesn't match one in the original SUBSCRIBE_TRACKS.

Ideas I posted on #1660:
1) Assume that people can figure out what the initial parameters are, so omit them. (Status quo)
  - Pro: Easy, Status quo
  - Con: If you update SUB_TRACKS, then how do you know which params are in use? Doesn't tell you anything about bare PUBLISH
2) Reference the Request ID of the SUB_TRACKS or subsequent REQUEST_UPDATE to indicate which params are in use
  - Pro: Only one field, simple validation.
  - Con: Instead of being racy, now you need to store all params ever sent on a SUB_TRACKS for an unknown time.  Doesn't tell you anything about bare PUBLISH
3) Repeat all the SUBSCRIBE parameters in the PUBLISH
  - Pro: No new framing or codepoints. They also don't conflict, so there's no ambiguity.  Could be useful in bare PUBLISH?  Similar to what we do for FORWARD and GROUP_ORDER today.
  - Con: Some SUBSCRIBE parameters don't make that much sense in the PUBLISH and aren't allowed there today. Could conflict with the SUBSCRIBE_TRACKS, which creates another error condition.
4) Create a parameter that contains all the parameters the Subscription was initialized with (ie: SUBSCRIBE_PARAMS?)
  - Pro: Don't have to think about whether a Parameter might have a different meaning depending on which perspective it comes from (it probably shouldn't anyway, though?)
  - Con: Another nested parameter. Could conflict with the SUBSCRIBE_TRACKS, which creates another error condition.
5) Use a potential new frame (ie: SUBSCRIPTION_STATE_UPDATE) to communicate the Subscription state
 - Pro: Could re-use a mechanism we (probably) want for DTS and Switch
[...]

PR #1825, opened 2026-07-22
Replace FORWARD parameter with Location-Filter

### LINKED ISSUE #1352: SUBSCRIBE does not need a forward parameter if we have filters
Currently within SUBSCRIBE you can set a [forward parameter](https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#forward-parameter) of 0 (don't send me anything) or 1 ( send me everything, default).

If we implement FILTERS (see https://github.com/moq-wg/moq-transport/pull/1164 and also Filter design team results), then we can remove this flag as:

a forward flag of 0 is syntactically equivalent to a FILTER that blocks everything
a forward flag of 1 is syntactically equivalent to a FILTER that allows everything 

Forward state can be managed by SUBSCRIBE_UPDATE and adjusting the FILTER.

### LINKED ISSUE #1694: bikeshed forward state name
In talking with other implementors, the conversation gets confusing of when we are just talking about forwarding in general, which there is lots of , and when we are talking about the bool state of the forwarding flag.  

I think this would create less confusion if it was named "paused" 

This would help make clear what was going on when people say things like "stop forwarding" which is not clear if they mean end the subscription or set forwarding flag to zero.

suhasHere:
Fixes #1352 
Fixes #1694

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1788,7 +1788,7 @@ Object Property types and values.  It only filters Object Properties in
 the Object header, and does not evaluate Track Properties in PUBLISH
 messages.
 
-### Combining Filters
+### Combining Filters {#combining-filters}
 
 All filter types are combined using logical "AND" operations
 to further restrict which tracks and objects pass all filter criteria.
@@ -1802,22 +1802,29 @@ Pass = Location Filters AND Range Filters
 
 ### Pausing Subscriptions {#pausing-subscriptions}
 
-A subscription is paused by including a Location Filter where the End Location
-is strictly less than the Start Location, creating an empty range that no object
-can satisfy.  The recommended encoding is StartGroup=0, StartObject=1,
-EndGroupDelta=0, EndObject=0, representing the empty range [{0,1}, {0,0}].
+A subscription can be paused, which suppresses delivery of all Objects, by
+including a Location Filter whose End Location is strictly less than its Start
+Location: an empty range that no object can satisfy. The recommended encoding
+is StartGroup=0, StartObject=1, EndGroupDelta=0, EndObject=0, representing the
+empty range [{0,1}, {0,0}]. Any other filter that filters out all Objects
+(for example, a Location Filter whose range lies entirely before the Track's
+first Object) has the same effect on delivery.
 
-A paused subscription does not deliver any Objects, but control messages such
-as PUBLISH_DONE ({{message-publish-done}}) are still sent.
+When multiple filter SetIDs are in use, filters within a SetID are combined
+with AND while distinct SetIDs are combined with OR
+(see {{combining-filters}}). To pause a subscription, the pausing filter MUST
+be present in every SetID used by the subscription; otherwise Objects can still
+pass via an unpaused SetID.
 
 To pause a subscription, the endpoint includes a pausing Location Filter in
-SUBSCRIBE, PUBLISH, PUBLISH_OK, SUBSCRIBE_TRACKS, or REQUEST_UPDATE. To resume,
-the endpoint sends a REQUEST_UPDATE that either removes the Location Filter
-(Length=0) or replaces it with a non-pausing filter (see {{location-filters}}).
+SUBSCRIBE, PUBLISH, PUBLISH_OK, or REQUEST_UPDATE. To resume, the endpoint
+sends a REQUEST_UPDATE that either removes the Location Filter (Length=0) or
+replaces it with a non-pausing filter (see {{location-filters}}).
 
-In the case of a REQUEST_UPDATE for SUBSCRIBE_TRACKS, the paused state applies
-to future subscriptions that match the prefix. Existing subscriptions are
-unaffected.
+A pausing Location Filter can also appear in SUBSCRIBE_TRACKS; it does not
+pause any existing subscription. Instead, it sets the initial paused state for
+future subscriptions that match the prefix (see
+{{parameters-on-subscribe-tracks}}).
 
 ### Joining an Ongoing Track
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2039,8 +2039,8 @@ to the old relay can be cancelled (see {{request-cancellation}}).
 There are two ways to publish through a relay:
 
 1. Send a PUBLISH message for a specific Track to the relay. The relay MAY
-respond with PUBLISH_OK in a paused state (see {{pausing-subscriptions}}) until
-there are known subscribers for new Tracks.
+accept the PUBLISH and then pause the subscription via REQUEST_UPDATE (see
+{{pausing-subscriptions}}) until there are known subscribers for new Tracks.
 
 2. Send a PUBLISH_NAMESPACE message for a Track Namespace to the relay. This
 enables the relay to send SUBSCRIBE or FETCH messages to publishers for Tracks
@@ -2110,9 +2110,10 @@ When a relay receives an authorized PUBLISH_NAMESPACE for a namespace that
 matches one or more existing subscriptions to other upstream sessions, it MUST
 send a SUBSCRIBE to the publisher that sent the PUBLISH_NAMESPACE for each
 matching subscription.  When it receives an authorized PUBLISH message for a
-Track that has `Established` downstream subscriptions, it MUST respond with
-PUBLISH_OK.  If at least one downstream subscriber for the Track has
-a non-paused subscription, the Relay MUST NOT pause in the reply.
+Track that has `Established` downstream subscriptions, it MUST accept the
+PUBLISH.  If at least one downstream subscriber for the Track has a
+non-paused subscription, the Relay MUST NOT subsequently pause the upstream
+subscription.
 
 If a Session is closed due to an unknown or invalid control message or Object,
 the Relay MUST NOT propagate that message or Object to another Session, because
@@ -3911,13 +3912,13 @@ the Track is used. If omitted from FETCH, the receiver uses Ascending (0x1).
 ### LOCATION FILTER Parameter {#location-filter}
 
 The LOCATION_FILTER parameter (Parameter Type 0x21) uses length-prefixed
-encoding. It MAY appear in a FETCH, SUBSCRIBE, PUBLISH, PUBLISH_OK,
+encoding. It MAY appear in a FETCH, SUBSCRIBE, PUBLISH,
 SUBSCRIBE_TRACKS, REQUEST_UPDATE (for a subscription), or
 PUBLISH_STATE_NOTIFY message. It is a Location Filter
 (see {{location-filters}}).
 
 If omitted from FETCH or SUBSCRIBE, the fetch or subscription is
-unfiltered.  If omitted from PUBLISH, PUBLISH_OK, or SUBSCRIBE_TRACKS, the
+unfiltered.  If omitted from PUBLISH or SUBSCRIBE_TRACKS, the
 subscription is not paused.  If omitted from REQUEST_UPDATE or
 PUBLISH_STATE_NOTIFY, the value is unchanged.  When sent in
 PUBLISH_STATE_NOTIFY, it reports the Location Filter now in effect at the
@@ -3997,11 +3998,12 @@ FILL_PARAMETERS is not retained as subscription state. It applies only to the
 message that carries it, so the sticky-parameter rules in
 {{message-request-update}} do not apply to it.
 
-To fill-join a track initiated via PUBLISH, the subscriber SHOULD respond with
-PUBLISH_OK that pauses the subscription (see {{pausing-subscriptions}}), then
-send REQUEST_UPDATE that resumes the subscription and carries FILL_PARAMETERS.
+To fill-join a track initiated via PUBLISH, the subscriber SHOULD send
+REQUEST_UPDATE carrying FILL_PARAMETERS as soon as it accepts the PUBLISH.
 The REQUEST_UPDATE_OK will contain a fresh LARGEST_OBJECT establishing the
-correct fill range.
+correct fill range. Objects delivered on the subscription before the
+publisher processes the REQUEST_UPDATE are outside the fill range and are
+delivered as ordinary subscription Objects.
 
 ### EXPIRES Parameter {#expires}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -634,7 +634,7 @@ incurring additional latency.
 Publishers MAY start sending Objects on PUBLISH-initiated subscriptions before
 receiving a PUBLISH_OK response to reduce latency.  Doing so can consume
 unnecessary resources in cases where the Subscriber rejects the subscription
-with REQUEST_ERROR or pauses the subscription in PUBLISH_OK. It can also result in
+with REQUEST_ERROR or pauses the subscription in REQUEST_UPDATE. It can also result in
 the Subscriber dropping Objects if its buffering limits are exceeded (see
 {{datagrams}} and {{subgroup-header}}).
 
@@ -957,7 +957,7 @@ Pass = Location Filters AND Range Filters
 
 ### Pausing Subscriptions {#pausing-subscriptions}
 
-A subscription is paused by any filter that matches no Objects. The
+A subscription is paused if it contains a filter that cannot match any Objects. The
 recommended pausing filter is a Location Filter with StartGroup=0,
 StartObject=1, EndGroupDelta=0, EndObject=0, representing the empty range
 [{0,1}, {0,0}]. When multiple filter SetIDs are in use (see
@@ -965,9 +965,12 @@ StartObject=1, EndGroupDelta=0, EndObject=0, representing the empty range
 otherwise Objects can pass via an unpaused SetID.
 
 An endpoint pauses a subscription by including a pausing Location Filter in
-SUBSCRIBE, PUBLISH, PUBLISH_OK, or REQUEST_UPDATE, and resumes it by sending a
+SUBSCRIBE, PUBLISH, or REQUEST_UPDATE, and resumes it by sending a
 REQUEST_UPDATE that removes the Location Filter (Length=0) or replaces it with
 a non-pausing filter.
+
+Endpoints SHOULD NOT use Range Filters to pause a subscription, because Range
+Filters are optional and MAX_FILTER_RANGES can be zero.
 
 A pausing Location Filter in SUBSCRIBE_TRACKS does not pause any existing
 subscription; it sets the initial paused state for future subscriptions that
@@ -3102,7 +3105,7 @@ STOP_SENDING frame.
 
 A publisher that pauses the subscription (see {{pausing-subscriptions}})
 indicates that it will not transmit any objects until the subscriber resumes it
-via PUBLISH_OK or REQUEST_UPDATE. If the subscription is not paused, the
+via REQUEST_UPDATE. If the subscription is not paused, the
 publisher will start transmitting objects immediately, possibly before
 PUBLISH_OK. Delivery starts at the Next Object relative to the Largest Object
 at the time the publisher begins sending.
@@ -4042,10 +4045,10 @@ PUBLISH, or REQUEST_UPDATE_OK
 ### NEW GROUP REQUEST Parameter {#new-group-request}
 
 The NEW_GROUP_REQUEST parameter (Parameter Type 0x32) is a varint. It MAY appear
-in PUBLISH_OK, SUBSCRIBE or REQUEST_UPDATE for a subscription.  It represents the largest Group
+in SUBSCRIBE or REQUEST_UPDATE for a subscription.  It represents the largest Group
 ID in the Track known by the subscriber, plus 1. A value of 0 indicates that the
 subscriber has no Group information for the Track.  A subscriber MUST NOT send
-this parameter in PUBLISH_OK or REQUEST_UPDATE if the Track did not
+this parameter in REQUEST_UPDATE if the Track did not
 include the DYNAMIC_GROUPS Property with value 1.  A subscriber MAY
 include this parameter in SUBSCRIBE without foreknowledge of support.  If the
 original publisher does not support dynamic Groups, it ignores the parameter in that
@@ -5511,7 +5514,7 @@ Setup Options SHOULD request a provisional registration.
 | 0x08 | EXPIRES | {{expires}} |
 | 0x09 | LARGEST_OBJECT | {{largest-param}} |
 | 0x0A | FILL_TIMEOUT | {{fill-timeout}} |
-| 0x10 | Reserved | N/A |
+| 0x10 | Reserved (Formerly FORWARD) | N/A |
 | 0x20 | SUBSCRIBER_PRIORITY | {{subscriber-priority}} |
 | 0x21 | LOCATION_FILTER | {{location-filter}} |
 | 0x22 | GROUP_ORDER | {{group-order}} |
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1802,29 +1802,21 @@ Pass = Location Filters AND Range Filters
 
 ### Pausing Subscriptions {#pausing-subscriptions}
 
-A subscription can be paused, which suppresses delivery of all Objects, by
-including a Location Filter whose End Location is strictly less than its Start
-Location: an empty range that no object can satisfy. The recommended encoding
-is StartGroup=0, StartObject=1, EndGroupDelta=0, EndObject=0, representing the
-empty range [{0,1}, {0,0}]. Any other filter that filters out all Objects
-(for example, a Location Filter whose range lies entirely before the Track's
-first Object) has the same effect on delivery.
+A subscription is paused by any filter that matches no Objects. The
+recommended pausing filter is a Location Filter with StartGroup=0,
+StartObject=1, EndGroupDelta=0, EndObject=0, representing the empty range
+[{0,1}, {0,0}]. When multiple filter SetIDs are in use (see
+{{combining-filters}}), the pausing filter MUST be present in every SetID;
+otherwise Objects can pass via an unpaused SetID.
 
-When multiple filter SetIDs are in use, filters within a SetID are combined
-with AND while distinct SetIDs are combined with OR
-(see {{combining-filters}}). To pause a subscription, the pausing filter MUST
-be present in every SetID used by the subscription; otherwise Objects can still
-pass via an unpaused SetID.
+An endpoint pauses a subscription by including a pausing Location Filter in
+SUBSCRIBE, PUBLISH, PUBLISH_OK, or REQUEST_UPDATE, and resumes it by sending a
+REQUEST_UPDATE that removes the Location Filter (Length=0) or replaces it with
+a non-pausing filter.
 
-To pause a subscription, the endpoint includes a pausing Location Filter in
-SUBSCRIBE, PUBLISH, PUBLISH_OK, or REQUEST_UPDATE. To resume, the endpoint
-sends a REQUEST_UPDATE that either removes the Location Filter (Length=0) or
-replaces it with a non-pausing filter (see {{location-filters}}).
-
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1628,8 +1625,8 @@ where each SetID=i is the AND of filters with SetID=i.
 The Track Property filter parameter MAY appear multiple times in a
 SUBSCRIBE_TRACKS message or REQUEST_UPDATE for it.
 All other filter parameters MAY appear multiple times in a FETCH, SUBSCRIBE,
-SUBSCRIBE_TRACKS, PUBLISH_OK, or REQUEST_UPDATE (on a subscription, from the subscriber only)
-message.  If the same combination of Parameter Type, SetID, and Property Type
+PUBLISH, SUBSCRIBE_TRACKS, PUBLISH_OK, or REQUEST_UPDATE (on a subscription,
```
afrind: A little odd to allow all filter types in publish. The purpose of forward there is informative, or maybe this dovetails with Ian's proposal.

draft-ietf-moq-transport.md
```diff
@@ -1660,13 +1657,33 @@ All filter types are combined using logical "AND" operations
...
+### Pausing Subscriptions {#pausing-subscriptions}
+
+A subscription is paused by including a Range Filter with an empty range set.
+For example, an OBJECTID_FILTER (Parameter Type 0x26) containing only a SetID
+and no ranges matches no Object IDs, causing all Objects to be blocked. Any
+Range Filter type (0x25-0x27) can be used for this purpose; the result is the
+same since an empty range set passes nothing.
+
+A paused subscription does not deliver any Objects, but control messages such
+as PUBLISH_DONE ({{message-publish-done}}) are still sent.
```
afrind, line 1699: This is redundant with the sentence in subscriptions section, do we need both?

draft-ietf-moq-transport.md
```diff
@@ -1660,13 +1657,33 @@ All filter types are combined using logical "AND" operations
...
+
+To pause a subscription, the endpoint includes a Range Filter with an empty
+range set in SUBSCRIBE, PUBLISH, PUBLISH_OK, SUBSCRIBE_TRACKS, or
```
afrind: SUBSCRIBE_TRACKS doesn't pause a subscription, as described below

afrind (commented):
The interactions with joining fetch are a little more subtle now -- I need to scan all the filter types for an empty range, but maybe fill-fetch will remove that problem?

draft-ietf-moq-transport.md
```diff
@@ -1660,13 +1657,33 @@ All filter types are combined using logical "AND" operations
...
+A subscription is paused by including a Range Filter with an empty range set.
+For example, an OBJECTID_FILTER (Parameter Type 0x26) containing only a SetID
+and no ranges matches no Object IDs, causing all Objects to be blocked. Any
+Range Filter type (0x25-0x27) can be used for this purpose; the result is the
+same since an empty range set passes nothing.
+
+A paused subscription does not deliver any Objects, but control messages such
+as PUBLISH_DONE ({{message-publish-done}}) are still sent.
+
+To pause a subscription, the endpoint includes a Range Filter with an empty
```
sharmafb: I think things are actually more complicated than this with multiple filters, because filters within one SetID are ANDed, but different SetIDs are ORed. You may want to clarify that the filter should be present in every SetID in use.

draft-ietf-moq-transport.md
```diff
@@ -1660,13 +1657,33 @@ All filter types are combined using logical "AND" operations
...
+The publisher MUST forward only objects that pass all filters.
 
 ~~~
-Pass = Forward AND Location Filters AND Range Filters
+Pass = Location Filters AND Range Filters
 ~~~
 
+### Pausing Subscriptions {#pausing-subscriptions}
+
+A subscription is paused by including a Range Filter with an empty range set.
```
sharmafb: This creates a dependency on MAX_FILTER_RANGES being nonzero, right? MAX_FILTER_RANGES defaults to zero, so by default, someone wouldn't be able to pause a subscription.

draft-ietf-moq-transport.md
ianswett: Yes, we'll talk about this today, see #1803

draft-ietf-moq-transport.md
ianswett: You could use a Location Filter instead, which is what I'd recommend since MAX_FILTER_RANGES can be zero.

draft-ietf-moq-transport.md
```diff
@@ -1660,13 +1657,33 @@ All filter types are combined using logical "AND" operations
...
-The Forward parameter is also a type of filter.  The publisher MUST
-forward only objects that pass all filters.
+The publisher MUST forward only objects that pass all filters.
 
 ~~~
-Pass = Forward AND Location Filters AND Range Filters
+Pass = Location Filters AND Range Filters
 ~~~
 
+### Pausing Subscriptions {#pausing-subscriptions}
```
ianswett, line 1691: I like the addition of this section, since there are already multiple ways to pause a Subscription.

Does it make sense to say "A subscription can be paused by doing X, or by any other filter that filters out all Objects."

ianswett (commented):
Overall, this looks good, only one substantial suggestion, which is we shouldn't use Range filters as the example way to pause a subscription if they're optional.

ianswett:
Reading the Top-N filter PR, I think it's actually quite important we land this change, because it prevents us from being lazy when we pause a Track and saying we'll flip forward to 0.  Instead, we should typically should be specifying an End Location Filter or something similar to indicate the last Object the peer will receive.

draft-ietf-moq-transport.md
```diff
@@ -1517,7 +1514,7 @@ incurring additional latency.
 Publishers MAY start sending Objects on PUBLISH-initiated subscriptions before
 receiving a PUBLISH_OK response to reduce latency.  Doing so can consume
 unnecessary resources in cases where the Subscriber rejects the subscription
-with REQUEST_ERROR or sets Forward=0 in REQUEST_UPDATE. It can also result in
+with REQUEST_ERROR or pauses the subscription in PUBLISH_OK. It can also result in
```
```suggestion
with REQUEST_ERROR or pauses the subscription in REQUEST_UPDATE. It can also result in
```

draft-ietf-moq-transport.md
```diff
@@ -1784,19 +1788,36 @@ Object Property types and values.  It only filters Object Properties in
...
+The publisher MUST forward only objects that pass all filters.
 
 ~~~
-Pass = Forward AND Location Filters AND Range Filters
+Pass = Location Filters AND Range Filters
 ~~~
 
+### Pausing Subscriptions {#pausing-subscriptions}
+
+A subscription is paused by any filter that matches no Objects. The
```
```suggestion
A subscription is paused if it contains a filter that cannot match any Objects. The
```

draft-ietf-moq-transport.md
```diff
@@ -1784,19 +1788,36 @@ Object Property types and values.  It only filters Object Properties in
...
+recommended pausing filter is a Location Filter with StartGroup=0,
+StartObject=1, EndGroupDelta=0, EndObject=0, representing the empty range
+[{0,1}, {0,0}]. When multiple filter SetIDs are in use (see
+{{combining-filters}}), the pausing filter MUST be present in every SetID;
+otherwise Objects can pass via an unpaused SetID.
```
ianswett: Maybe we should advise that you shouldn't use Range Filters to pause a Track?

draft-ietf-moq-transport.md
```diff
@@ -1784,19 +1788,36 @@ Object Property types and values.  It only filters Object Properties in
...
+
+An endpoint pauses a subscription by including a pausing Location Filter in
+SUBSCRIBE, PUBLISH, PUBLISH_OK, or REQUEST_UPDATE, and resumes it by sending a
```
```suggestion
SUBSCRIBE, PUBLISH, or REQUEST_UPDATE, and resumes it by sending a
```

draft-ietf-moq-transport.md
```diff
@@ -5476,7 +5485,7 @@ Setup Options SHOULD request a provisional registration.
 | 0x08 | EXPIRES | {{expires}} |
 | 0x09 | LARGEST_OBJECT | {{largest-param}} |
 | 0x0A | FILL_TIMEOUT | {{fill-timeout}} |
-| 0x10 | FORWARD | {{forward-parameter}} |
+| 0x10 | Reserved | N/A |
```
```suggestion
| 0x10 | Reserved (Formerly FORWARD) | N/A |
```

draft-ietf-moq-transport.md
```diff
@@ -4926,7 +4935,7 @@ If omitted, the publisher's preference is Ascending (0x1).
 DYNAMIC_GROUPS (Property Type 0x30) is a Track Property.
 The allowed values are 0 or 1. When the value is 1, it indicates
 that the subscriber can request the Original Publisher to start a new Group
-by including the NEW_GROUP_REQUEST parameter in REQUEST_UPDATE
+by including the NEW_GROUP_REQUEST parameter in PUBLISH_OK or REQUEST_UPDATE
```
```suggestion
by including the NEW_GROUP_REQUEST parameter in REQUEST_UPDATE
```

draft-ietf-moq-transport.md
```diff
@@ -4139,15 +4147,16 @@ only interested in or authorized to access a subset of available tracks.
...
-FORWARD and GROUP_ORDER parameters as described below.  When omitted by
-the publisher in PUBLISH, the subscriber uses the default value for each.
-
-If the FORWARD parameter ({{forward-parameter}}) is present in this message and
-equal to 0, PUBLISH messages resulting from this SUBSCRIBE_TRACKS will set
-the FORWARD parameter to 0. If the FORWARD parameter is equal to 1 or omitted
-from this message, PUBLISH messages resulting from this SUBSCRIBE_TRACKS will
-set the FORWARD parameter to 1, or indicate that value by omitting the parameter
+SUBSCRIBE_TRACKS. These Parameters are explicitly communicated, with the
+exception of a pausing Location Filter and GROUP_ORDER as described below.
```
ianswett: I think the old text was wrong?

```suggestion
SUBSCRIBE_TRACKS. These Parameters are explicitly communicated in PUBLISH.
```

draft-ietf-moq-transport.md
```diff
@@ -3704,9 +3712,9 @@ send REQUEST_ERROR with error code `UNINTERESTED`, and abandon reading any
 publisher initiated streams associated with that subscription using a
 STOP_SENDING frame.
 
-A publisher that sends the FORWARD parameter ({{forward-parameter}}) equal to 0
-indicates that it will not transmit any objects until the subscriber sets the
-Forward State to 1. If the FORWARD parameter is omitted or equal to 1, the
+A publisher that pauses the subscription (see {{pausing-subscriptions}})
+indicates that it will not transmit any objects until the subscriber resumes it
+via PUBLISH_OK or REQUEST_UPDATE. If the subscription is not paused, the
```
```suggestion
via REQUEST_UPDATE. If the subscription is not paused, the
```

draft-ietf-moq-transport.md
```diff
@@ -2980,32 +3007,13 @@ A relay MUST set LARGEST_OBJECT to the largest of the following:
...
-
 ### NEW GROUP REQUEST Parameter {#new-group-request}
 
 The NEW_GROUP_REQUEST parameter (Parameter Type 0x32) is a varint. It MAY appear
-in SUBSCRIBE or REQUEST_UPDATE for a subscription.  It represents the largest Group
+in PUBLISH_OK, SUBSCRIBE or REQUEST_UPDATE for a subscription.  It represents the largest Group
 ID in the Track known by the subscriber, plus 1. A value of 0 indicates that the
 subscriber has no Group information for the Track.  A subscriber MUST NOT send
-this parameter in REQUEST_UPDATE if the Track did not
+this parameter in PUBLISH_OK or REQUEST_UPDATE if the Track did not
```
```suggestion
this parameter in REQUEST_UPDATE if the Track did not
```

draft-ietf-moq-transport.md
```diff
@@ -2980,32 +3007,13 @@ A relay MUST set LARGEST_OBJECT to the largest of the following:
...
-the value for the subscription remains unchanged.  If the parameter is omitted
-from any other message, the default value is 1.  When sent in
-PUBLISH_STATE_NOTIFY, it reports the Forwarding State now in effect at the
-publisher.
-
 ### NEW GROUP REQUEST Parameter {#new-group-request}
 
 The NEW_GROUP_REQUEST parameter (Parameter Type 0x32) is a varint. It MAY appear
-in SUBSCRIBE or REQUEST_UPDATE for a subscription.  It represents the largest Group
+in PUBLISH_OK, SUBSCRIBE or REQUEST_UPDATE for a subscription.  It represents the largest Group
```
```suggestion
in SUBSCRIBE or REQUEST_UPDATE for a subscription.  It represents the largest Group
```

ianswett (commented):
Mostly looks good, though you pulled in a few PUBLISH_OKs, where you can no longer change the FORWARD state.

draft-ietf-moq-transport.md
mzanaty: I agree. I would only describe pause in terms of (mandatory) location filters not (optional) range filters. It is impossible in general to determine if a range filter guarantees no matching objects, so the intent to pause is impossible for the publisher to discern. Also consider the mess it causes for upstream aggregation if each downstream subscriber tried to pause using different random range filters. Pause should be a well specified, deterministic mechanism that can be aggregated upstream unambiguously.

draft-ietf-moq-transport.md
suhasHere: Applied in eb20d309.

draft-ietf-moq-transport.md
suhasHere: Applied in eb20d309.

draft-ietf-moq-transport.md
suhasHere: Added a SHOULD NOT paragraph in eb20d309.

draft-ietf-moq-transport.md
suhasHere: Applied in eb20d309.

draft-ietf-moq-transport.md
suhasHere: Applied in eb20d309.

draft-ietf-moq-transport.md
suhasHere: Already applied in origin/main after rebase.

draft-ietf-moq-transport.md
suhasHere: Already applied in origin/main (#1869) after rebase.

draft-ietf-moq-transport.md
suhasHere: Applied in eb20d309.

draft-ietf-moq-transport.md
suhasHere: Applied in eb20d309.

draft-ietf-moq-transport.md
suhasHere: Applied in eb20d309.

draft-ietf-moq-transport.md
```diff
@@ -3636,7 +3663,7 @@ if found.
 The number of Message Parameters is not specifically limited, but the total
 length of a control message is limited to 2^16-1 bytes.
 
-Message Parameters in SUBSCRIBE and FETCH MUST NOT cause the
+Message Parameters in SUBSCRIBE, PUBLISH_OK and FETCH MUST NOT cause the
```
ianswett, line 3666:
```suggestion
Message Parameters in SUBSCRIBE and FETCH MUST NOT cause the
```

draft-ietf-moq-transport.md
```diff
@@ -3785,7 +3812,7 @@ that alias has not received a response.
 ### SUBGROUP_DELIVERY_TIMEOUT Parameter {#subgroup-delivery-timeout}
 
 The SUBGROUP_DELIVERY_TIMEOUT parameter (Parameter Type 0x06) is a varint. It
-MAY appear in a SUBSCRIBE, PUBLISH, or REQUEST_UPDATE message.  Its
+MAY appear in a SUBSCRIBE, PUBLISH, PUBLISH_OK, or REQUEST_UPDATE message.  Its
```
ianswett, line 3815:
```suggestion
MAY appear in a SUBSCRIBE, PUBLISH, or REQUEST_UPDATE message.  Its
```

draft-ietf-moq-transport.md
```diff
@@ -3795,7 +3822,7 @@ for the same track.
 ### OBJECT_DELIVERY_TIMEOUT Parameter {#object-delivery-timeout}
 
 The OBJECT_DELIVERY_TIMEOUT parameter (Parameter Type 0x02) is a varint. It
-MAY appear in a SUBSCRIBE, PUBLISH, or REQUEST_UPDATE message.  Its
+MAY appear in a SUBSCRIBE, PUBLISH, PUBLISH_OK, or REQUEST_UPDATE message.  Its
```
ianswett, line 3825:
```suggestion
MAY appear in a SUBSCRIBE, PUBLISH, or REQUEST_UPDATE message.  Its
```

draft-ietf-moq-transport.md
```diff
@@ -3857,12 +3884,13 @@ If RENDEZVOUS_TIMEOUT is absent, the default is 0.
 ### SUBSCRIBER PRIORITY Parameter {#subscriber-priority}
 
 The SUBSCRIBER_PRIORITY parameter (Parameter Type 0x20) is a uint8. It MAY
-appear in a SUBSCRIBE, PUBLISH, FETCH, or REQUEST_UPDATE
+appear in a SUBSCRIBE, PUBLISH, PUBLISH_OK, FETCH, or REQUEST_UPDATE
```
ianswett, line 3887:
```suggestion
appear in a SUBSCRIBE, PUBLISH, FETCH, or REQUEST_UPDATE
```

draft-ietf-moq-transport.md
```diff
@@ -3857,12 +3884,13 @@ If RENDEZVOUS_TIMEOUT is absent, the default is 0.
...
 (for a subscription or FETCH). It is an integer expressing the priority of a
 subscription relative to other subscriptions and fetch responses in the same
 session. Lower numbers get higher priority. See {{priorities}}.
 
-If omitted from SUBSCRIBE or FETCH, the publisher uses the value 128.
+If omitted from SUBSCRIBE, PUBLISH_OK or FETCH, the publisher uses
```
ianswett, line 3892:
```suggestion
If omitted from SUBSCRIBE or FETCH, the publisher uses
```

draft-ietf-moq-transport.md
```diff
@@ -3967,6 +3998,13 @@ FILL_PARAMETERS is not retained as subscription state. It applies only to the
 message that carries it, so the sticky-parameter rules in
 {{message-request-update}} do not apply to it.
 
+To fill-join a track initiated via PUBLISH, the subscriber SHOULD send
```
ianswett, line 4001: I think this paragraph got moved to 'Joining a Track' and updated?

ianswett (approved):
There are a few more PUBLISH_OKs that need to be removed, but looking good.

draft-ietf-moq-transport.md
```diff
@@ -760,6 +758,12 @@ When EndGroupDelta and EndObject are omitted from a subscription filter, the
 subscription is open-ended. When they are omitted from a Fetch, the
 EndGroup and EndObject are `Largest Object`.
 
+A Location Filter where the End Location is strictly less than the Start
+Location matches no objects and pauses the subscription (see
+{{pausing-subscriptions}}).  The recommended encoding for a paused filter is
```
afrind, line 763: recommended is a keyword?

draft-ietf-moq-transport.md
```diff
@@ -938,19 +943,39 @@ Object Property types and values.  It only filters Object Properties in
...
 
 ~~~
-Pass = Forward AND Location Filters AND Range Filters
+Pass = Location Filters AND Range Filters
 ~~~
 
+### Pausing Subscriptions {#pausing-subscriptions}
+
+A subscription is paused if it contains a filter that cannot match any Objects. The
+recommended pausing filter is a Location Filter with StartGroup=0,
```
afrind, line 961: recommeded is a keyword

draft-ietf-moq-transport.md
```diff
@@ -938,19 +943,39 @@ Object Property types and values.  It only filters Object Properties in
...
+StartObject=1, EndGroupDelta=0, EndObject=0, representing the empty range
```
afrind, line 962: This is a wire encoding detail?  Can you give it canonical coordinates {0,1} - {0,0}?

draft-ietf-moq-transport.md
```diff
@@ -938,19 +943,39 @@ Object Property types and values.  It only filters Object Properties in
...
+[{0,1}, {0,0}]. When multiple filter SetIDs are in use (see
+{{combining-filters}}), the pausing filter MUST be present in every SetID;
```
afrind, line 964: This is more a statement of fact than a normative requirement; we can drop MUST.

```suggestion
{{combining-filters}}), the subscriber adds the pausing filter to every SetID;
```

draft-ietf-moq-transport.md
```diff
@@ -973,6 +998,7 @@ which can be relative.  The publisher delivers the fill range on a fill fetch
 stream and subscription-delivered Objects in subgroups or
 datagrams (see {{fill-semantics}}).
 
+
```
afrind, line 1001: delete?

draft-ietf-moq-transport.md
```diff
@@ -2065,12 +2092,12 @@ a SUBSCRIBE with FILL_PARAMETERS or FETCHes (see {{fill-semantics}}).
...
-SUBSCRIBE has Forward State=1 and the upstream subscriptions are in Forward
-State=0, the Relay MUST send REQUEST_UPDATE with Forward=1 to all publishers.
+SUBSCRIBE is not paused and the upstream subscriptions are paused, the Relay
+MUST send REQUEST_UPDATE to resume all upstream subscriptions.
 If there are no `Established` upstream subscriptions for the requested Track, the Relay
 MUST send a SUBSCRIBE request to each publisher that has published the
-subscription's namespace or prefix thereof.  If the SUBSCRIBE has Forward=1,
-then the Relay MUST use Forward=1 when subscribing upstream.
+subscription's namespace or prefix thereof.  If the SUBSCRIBE is not paused,
+the Relay MUST NOT pause when subscribing upstream.
```
afrind, line 2100: This is just a little awkward -- you want to say that the upstream subscribe must not start the subscription paused.  I can live with this though.

draft-ietf-moq-transport.md
```diff
@@ -2083,10 +2110,10 @@ When a relay receives an authorized PUBLISH_NAMESPACE for a namespace that
 matches one or more existing subscriptions to other upstream sessions, it MUST
 send a SUBSCRIBE to the publisher that sent the PUBLISH_NAMESPACE for each
 matching subscription.  When it receives an authorized PUBLISH message for a
-Track that has `Established` downstream subscriptions, it MUST respond with
-PUBLISH_OK.  If at least one downstream subscriber for the Track has
-Forward State=1, the Relay MUST change the Forward State to 1 with
-REQUEST_UPDATE.
+Track that has `Established` downstream subscriptions, it MUST accept the
+PUBLISH.  If at least one downstream subscriber for the Track has a
+non-paused subscription, the Relay MUST NOT subsequently pause the upstream
```
afrind, line 2115: non-paused?

draft-ietf-moq-transport.md
```diff
@@ -5497,7 +5516,7 @@ Setup Options SHOULD request a provisional registration.
 | 0x08 | EXPIRES | {{expires}} |
 | 0x09 | LARGEST_OBJECT | {{largest-param}} |
 | 0x0A | FILL_TIMEOUT | {{fill-timeout}} |
-| 0x10 | FORWARD | {{forward-parameter}} |
+| 0x10 | Reserved (Formerly FORWARD) | N/A |
```
afrind, line 5519: Don't need to reserve the code point with ALPN?

yuyou:
The "paused"/"unpaused" state versus no object is forwarded due to the filter conditions, IMHO, are 2 different things. One could pause a subscription with filters; and also have a "unpaused" subscription but no object is forwarded due to the filters.  

Some server-side track switching, e.g. SSTS or SWITCH_FROM PR #1674, may depend on this un-/paused state without using any filters. 
I would vote for keeping the current spec design as it.

draft-ietf-moq-transport.md
fluffy: you do if the code point is used before the ALPN is changed. It's not like we are short on code points. I think we should minimize things that might make debug harder and clean up all the reserved ones and reorder all the other at some point we get closer to WGLC

suhasHere:
@yuyou the proposed solution wouldn't block any of the above other than mechanics to do so

You pause a subscription and unpause with the filters you need using the same location filter machinery than 2 different ways to do so

draft-ietf-moq-transport.md
```diff
@@ -670,11 +670,26 @@ not send Objects on a paused subscription, and does send them when it is not
...
+empty range `[{0,1}, {0,0}]`.  When multiple filter SetIDs are in use (see
+{{combining-filters}}), the pausing filter MUST be present in every SetID;
+otherwise Objects can pass via an unpaused SetID.
+
+An endpoint pauses a subscription by including a pausing Location Filter in
+SUBSCRIBE, PUBLISH, or REQUEST_UPDATE, and resumes it by sending a
+REQUEST_UPDATE that removes the Location Filter (Length=0) or replaces it with
+a non-pausing filter.
+
+Endpoints SHOULD NOT use Range Filters ({{range-filters}}) to pause a
```
fluffy, line 685: Its weird to say what not to do without saying what to do. Perhaps a pointer to the section that has the recommended way

draft-ietf-moq-transport.md
```diff
@@ -3876,10 +3890,23 @@ EndGroup and EndObject are `Largest Object`.
 
 When EndObject is omitted, the filter includes all objects in the End Group.
 
+A Location Filter where the End Location is strictly less than the Start
+Location matches no objects and pauses the subscription (see
+{{pausing-subscriptions}}).  The recommended encoding for a paused filter is
+StartGroup=0, StartObject=1, EndGroupDelta=0, EndObject=0, which represents the
+empty range `[{0,1}, {0,0}]`.
```
fluffy, line 3897: When a relay does the pause or unpause, I think we need to specify at a MUST level how it is done. The subscriber is going to get an message that says the filter was updated and will need to be able to understand what is going on

draft-ietf-moq-transport.md
```diff
@@ -670,11 +670,26 @@ not send Objects on a paused subscription, and does send them when it is not
...
-FORWARD parameter ({{forward-parameter}}) in PUBLISH or SUBSCRIBE. The
-subscriber can pause an `Established` subscription by sending REQUEST_UPDATE
-with FORWARD set to 0, or resume it by sending REQUEST_UPDATE with FORWARD set
-to 1.
+A subscription is paused if it carries a filter that cannot match any Objects.
+The recommended pausing filter is a Location Filter ({{location-filters}}) with
+StartGroup=0, StartObject=1, EndGroupDelta=0, EndObject=0, representing the
+empty range `[{0,1}, {0,0}]`.  When multiple filter SetIDs are in use (see
+{{combining-filters}}), the pausing filter MUST be present in every SetID;
+otherwise Objects can pass via an unpaused SetID.
```
fluffy, line 678: I think we have to get specific about what a relay does when it "unpauses"  - does it restore all tof the multiple filter IDs back to how they were before they were paused ?

fluffy (changes requested):
I'm missing the motivation for why we are doing this but can live with it. 

The big issue for me is when relays pause or unpause, how does the subscriber understand what happened? And when they unpause, what happens to the filters ? Do they go back to the state they were before ? 

I'm trying to work through my head where the relay is pausing and unpausing due to something like SSTS and at the same time the subscriber is changing what layers it wants of a layered codec. Both are updating the filter and I see no text that is going to keep the updates not conflicting. It also seems very race in that all the things are trying to change the same atomic state which is the filters instead of different things changing different state. 

The more I think about it, the less I am sure this works. Needs discussion.

fluffy:
See sept 21 meeting for what we decided to try next on this

PR #1830, opened 2026-07-24
Top Tracks Filter

mzanaty:
Added Top Tracks Filter (was Track Filter in PR#1508 / PR#1401).

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3068,7 +3068,7 @@ exceeded, an endpoint MUST reject this with REQUEST_ERROR with error code INVALI
 
 #### MAX TOP TRACKS
 
-The MAX_TOP_TRACKS option (Type 0x08) limits the peer's value of
+The MAX_TOP_TRACKS option (Type 0x09) limits the peer's value of
 MaxTopTracks in the TOP_TRACKS_FILTER {{top-tracks-filter}} parameter.
 The default value is 0, so if not specified, the peer MUST NOT send any
 TOP_TRACKS_FILTER parameter.  If this limit is exceeded, an endpoint MUST
@@ -5395,6 +5395,7 @@ This registry is initially empty.
 | 0x06 | MAX_FILTER_RANGES | {{max-filter-ranges}} |
 | 0x07 | MOQT_IMPLEMENTATION | {{moqt-implementation}} |
 | 0x08 | MAX_REQUEST_UPDATES | {{max-request-updates}} |
+| 0x09 | MAX_TOP_TRACKS | {{max-top-tracks}} |
 | 0x7f * N + 0x9D | Reserved for greasing | {{grease}} |
 
 Endpoints MUST ignore unknown Setup Options as specified in
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1658,13 +1658,15 @@ messages.
 
 All filter types are combined using logical "AND" operations
 to further restrict which tracks and objects pass all filter criteria.
-This includes all Range Filters {{range-filters}} and Location
-Filters {{location-filters}}, which can be evaluated in any order.
+This includes all Range Filters {{range-filters}}, Location Filters
+{{location-filters}}, and Top Tracks Filters {{top-tracks-filters}}.
+Top Tracks Filters always evaluate first before all other filters
+which can be evaluated in any order.
 The Forward parameter is also a type of filter.  The publisher MUST
 forward only objects that pass all filters.
 
 ~~~
-Pass = Forward AND Location Filters AND Range Filters
+Pass = Top Tracks Filter AND Forward AND Location Filter AND Range Filters
 ~~~
 
 ### Joining an Ongoing Track
@@ -1845,14 +1847,111 @@ PUBLISH_NAMESPACE messages.
 
 ## Filtering SUBSCRIBE_TRACKS
 
+The Top Tracks Filter {{top-tracks-filters}} can be used in SUBSCRIBE_TRACKS
+to filter Tracks in a namespace using a specified Track or Object Property.
 Range Filters {{range-filters}} can be used in SUBSCRIBE_TRACKS to filter
 Tracks in a namespace using the Track Property Filter. Objects published in
 the resulting Subscriptions can be filtered by any Range Filter.
 
+### Top Tracks Filters
+
+The TOP_TRACKS_FILTER parameter {{top-tracks-filter}} selects a specified
+number (N) of tracks within a namespace with the highest Property Values
+for a specified Track or Object Property Type which MUST be even, i.e.
+a single integer value (see {{moq-key-value-pair}}).
+
+It is encoded with a Length prefix in bytes which MAY be 0 to indicate
+no filter, which can be used to remove the filter in REQUEST_UPDATE.
+
+~~~
+TOP_TRACKS_FILTER Parameter {
+  Type (vi64) = 0x29,
+  Length (vi64),
+  Property Type (vi64),
+  MaxTopTracks (vi64),
+}
+~~~
+
+MaxTopTracks limits the number of tracks selected concurrently,
+which MUST NOT be 0 and MUST NOT exceed the MAX_TOP_TRACKS setup option
+value sent by the peer, either of which is a `PROTOCOL_VIOLATION`.
+
+TopTracksTimeout {{top-tracks-timeout}} is a Track Property
+that limits the number of milliseconds a selected track can remain
+selected without publishing an object with Property Type.
+Timeout in the next section refers to this Track Property.
+
+#### Track Selection
+
+Tracks are promoted to the selected state, demoted to the
+deselected state, or purged back to the unknown state as
+shown in the following diagram. Objects only pass the filter
+for tracks in the selected state, as shown below.
+
+~~~
+        +--------------+
+        +   UNKNOWN    |
+        +  name/alias  |
+        +--------------+
+             ^    |
+             |    |     Newly Selected    +------------+
+PUBLISH_DONE |    +---> PUBLISH FWD=1 --->|  SELECTED  |
+as needed to |             Promoted       |  in Top N  |
+limit state  |    +---> REQ UPD FWD=1 --->|   Tracks   |
+             |    |       Reselected      +------------+
+             |    |                             |
+        +--------------+                        |
+        |  DESELECTED  |<---- REQ UPD FWD=0 <---+
+        | out of Top N |     Demoted/Timeout
+        +--------------+
+
+          Objects FAIL                     Objects PASS
+           the filter                       the filter
+~~~
+
+A track is selected if it publishes an object with the top N highest
+value for Property Type, where N = MaxTopTracks.  The publisher
+MUST send a PUBLISH message for each newly selected track.  For tracks
+with the same value, the track with the earlier delivered object wins
+the tie breaker, so a selected track remains selected until another track
+publishes a higher value that demotes it out of the top N, or Timeout
+elapses before the track delivers an object that remains in the top N,
+either of which deselect the track.  The publisher MUST send a
+REQUEST_UPDATE message with Forward=0 when a track is deselected.
+The publisher MUST send a REQUEST_UPDATE message with Forward=1 when a
+deselected track is reselected, which also updates the Joining Location
+as described in {{joining-fetch-range-calculation}}.
+
+Recently deselected tracks SHOULD be kept in a list to avoid more PUBLISH
+messages in case a deselected track is reselected.  A relay SHOULD limit
+this list size to protect its resources by sending PUBLISH_DONE
+messages to purge state for the oldest deselected tracks, putting them
+back in the unknown state.  If a track is reselected after this, it is
+considered newly selected so the publisher MUST send a PUBLISH message
+again.  Relays SHOULD keep a list size large enough to avoid excessive
+control messages but small enough to avoid excessive old subscription state.
+
+If the TOP_TRACKS_FILTER parameter is updated, the publisher evaluates the
+new filter to determine the new list of top N selected tracks and follows
+the same rules above for newly selected tracks in the top N (send PUBLISH),
+reselected tracks in the top N (send REQUEST_UPDATE with Forward=1), and
+newly deselected tracks demoted out of the top N (send REQUEST_UPDATE
+with Forward=0).  It can also purge excessive old deselected tracks as
+needed (send PUBLISH_DONE).
+
+If a Top Tracks Filter for a namespace overlaps with a direct subscription
+to a track name in the same namespace, it is considered to pass the Top Tracks
+Filter whether or not it is counted in the top N list, and MUST NOT
+be subject to filter state changes or actions.
+
+Objects received with a location lower than the largest received in that track
+are ignored for Top Tracks Filter state updates; however, if the track is already
+selected, they pass the filter.
+
 ### Relay Resource Protection in Large Namespaces {#large-namespaces}
 
 Relays SHOULD aggregate and propagate filters upstream on subscriptions,
-especially namespace subscriptions,
+especially namespace subscriptions including Top Tracks Filters,
 to conserve and protect their resources from excessive load.  They MAY
 also impose limits on the number of publishers in a namespace, by rejecting
 or closing namespace subscriptions with the error NAMESPACE_TOO_LARGE, or
@@ -2732,6 +2831,16 @@ Type which MUST be even, i.e. a single integer value
 (see {{moq-key-value-pair}}), otherwise the endpoint MUST reject this with
 REQUEST_ERROR with error code INVALID_FILTER. See {{range-filters}}.
 
+### TOP TRACKS FILTER Parameter {#top-tracks-filter}
+
+The TOP_TRACKS_FILTER parameter (Type=0x30) MAY appear in a SUBSCRIBE_TRACKS
+message or REQUEST_UPDATE for it.  It selects a specified
+number of tracks within a namespace with the highest Property Values for a
+specified Track or Object Property Type which MUST be even, i.e. a single
+integer value (see {{moq-key-value-pair}}), otherwise the endpoint MUST
+close the session with a `PROTOCOL_VIOLATION`.
+See {{top-tracks-filters}} and {{max-top-tracks}}.
+
 ### EXPIRES Parameter {#expires}
 
 The EXPIRES parameter (Parameter Type 0x8) is a varint. It MAY appear in
@@ -2956,6 +3065,14 @@ The MAX_FILTER_RANGES option (Type 0x06) limits the peer's total number of Range
 parameters for a given subscription or fetch.  The default value is 0, so if not
 specified, the peer MUST NOT send any such filter parameters.  If this limit is
 exceeded, an endpoint MUST reject this with REQUEST_ERROR with error code INVALID_FILTER.
+
+#### MAX TOP TRACKS
+
+The MAX_TOP_TRACKS option (Type 0x08) limits the peer's value of
+MaxTopTracks in the TOP_TRACKS_FILTER {{top-tracks-filter}} parameter.
+The default value is 0, so if not specified, the peer MUST NOT send any
+TOP_TRACKS_FILTER parameter.  If this limit is exceeded, an endpoint MUST
+close the session with a `PROTOCOL_VIOLATION`.
 
 #### MAX_REQUEST_UPDATES {#max-request-updates}
 
@@ -4775,6 +4892,11 @@ the session with `PROTOCOL_VIOLATION`.
 
 If omitted, the value is 0.
 
+## TOP TRACKS TIMEOUT {#top-tracks-timeout}
+
+TOP_TRACKS_TIMEOUT (Property Type 0x32) is a Track Property.
+See {{top-tracks-filter}}. If omitted, the value is 1000 milliseconds.
+
 ## Immutable Properties
 
 Immutable Properties (Property Type 0xB) is a Track or Object Property that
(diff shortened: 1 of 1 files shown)


vasilvv:
High level thoughts:
- I think this version of the mechanism, like a lot of revisions of it before, does not allow for pre-computation of top-N.  We should design mechanism that allows relays to precompute top-N on publication whenever those are published.
- SELECTED/DESELECTED mechanism most likely should be shared with SSTS somehow.

mzanaty:
Can you elaborate on 'precompute top-N on publication whenever those are published' ?

vasilvv:
If I am a relay receiving three tracks from the original publisher, I should be able to precompute the top-N at that relay without knowing anything about subscribers.

I believe this is important because fundamentally, MOQT is a "small number of publishers to millions of subscribers" protocol, meaning that we almost never let information flow from the subscriber (relay) to the publisher (relay), and when we do (e.g. NGR), it's very carefully adjusted to scale.  The current design doesn't really allow that:
- You don't know what property the subscriber would compute top on, so you have to wait for the subscriber to tell you.
- Similarly, you don't know the N.
- One subscriber can subscribe for top-5, and then a later one can subscribe for top-10, causing a recompute on all relays in-between.

draft-ietf-moq-transport.md
```diff
@@ -1845,14 +1847,111 @@ PUBLISH_NAMESPACE messages.
...
+The Top Tracks Filter {{top-tracks-filters}} can be used in SUBSCRIBE_TRACKS
+to filter Tracks in a namespace using a specified Track or Object Property.
 Range Filters {{range-filters}} can be used in SUBSCRIBE_TRACKS to filter
 Tracks in a namespace using the Track Property Filter. Objects published in
 the resulting Subscriptions can be filtered by any Range Filter.
 
+### Top Tracks Filters
+
+The TOP_TRACKS_FILTER parameter {{top-tracks-filter}} selects a specified
+number (N) of tracks within a namespace with the highest Property Values
```
ianswett, line 1859: We have a number of examples where we say we want the lowest or even 'top 5' and 'bottom 5' Tracks.  How about adding a way to do lowest N as well?  It seems like a tiny increase in implementation complexity and avoids people having to publish the same property in two different ways, which I think is the current solution to that problem?

ianswett:
> High level thoughts:
> 
> * I think this version of the mechanism, like a lot of revisions of it before, does not allow for pre-computation of top-N.  We should design mechanism that allows relays to precompute top-N on publication whenever those are published.
> * SELECTED/DESELECTED mechanism most likely should be shared with SSTS somehow.

Yes, I think this should be updated once #1820 lands to indicate SELECTED/DESELECTED

mzanaty:
Subscribe Tracks is needed before a relay can filter. This conveys N and the property to inspect. We can force the property to be a fixed type (called Selector in early designs) to 'precompute' it, but people had use cases of different subscribers filtering a namespace with different properties. N is subscriber specific, so relays can't just infer that from published tracks. They must wait for it in Subscribe Tracks. Since that also conveys the property, I see no benefit in a fixed Selector property type.

mzanaty:
Publish Notify will indeed help to toggle Forward 1/0 when Selected/Deselected, cutting control traffic in half.

draft-ietf-moq-transport.md
```diff
@@ -1658,13 +1658,15 @@ messages.
 
 All filter types are combined using logical "AND" operations
```
ianswett, line 1659: Please separate out the Object filters from Track filters in this paragraph, because the only overlap is property filters and the way it's written now is confusing.

draft-ietf-moq-transport.md
```diff
@@ -1658,13 +1658,15 @@ messages.
...
+This includes all Range Filters {{range-filters}}, Location Filters
+{{location-filters}}, and Top Tracks Filters {{top-tracks-filters}}.
+Top Tracks Filters always evaluate first before all other filters
+which can be evaluated in any order.
 The Forward parameter is also a type of filter.  The publisher MUST
 forward only objects that pass all filters.
 
 ~~~
-Pass = Forward AND Location Filters AND Range Filters
+Pass = Top Tracks Filter AND Forward AND Location Filter AND Range Filters
```
ianswett, line 1669: This line is particularly confusing and I think incorrect, because the previous sentence talks about Objects and the Top Tracks filter applies to Tracks, not Objects.

draft-ietf-moq-transport.md
```diff
@@ -1845,14 +1847,111 @@ PUBLISH_NAMESPACE messages.
...
+to filter Tracks in a namespace using a specified Track or Object Property.
 Range Filters {{range-filters}} can be used in SUBSCRIBE_TRACKS to filter
 Tracks in a namespace using the Track Property Filter. Objects published in
 the resulting Subscriptions can be filtered by any Range Filter.
 
+### Top Tracks Filters
+
+The TOP_TRACKS_FILTER parameter {{top-tracks-filter}} selects a specified
+number (N) of tracks within a namespace with the highest Property Values
+for a specified Track or Object Property Type which MUST be even, i.e.
```
ianswett, line 1860: If a Property is both a Track and Object Property, the Object Property value overrides the Track Property value.

Using 'or' here is a bit confusing, because it could be that you're looking at both.  Please add a specific sentence about the case when you're looking at both.

draft-ietf-moq-transport.md
```diff
@@ -1845,14 +1847,111 @@ PUBLISH_NAMESPACE messages.
...
+a single integer value (see {{moq-key-value-pair}}).
+
+It is encoded with a Length prefix in bytes which MAY be 0 to indicate
+no filter, which can be used to remove the filter in REQUEST_UPDATE.
```
```suggestion
It is encoded with a Length prefix in bytes. A length of 0 indicates
no filter, which can be used to remove the filter in REQUEST_UPDATE.
```

draft-ietf-moq-transport.md
```diff
@@ -1845,14 +1847,111 @@ PUBLISH_NAMESPACE messages.
...
+  Property Type (vi64),
+  MaxTopTracks (vi64),
+}
+~~~
+
+MaxTopTracks limits the number of tracks selected concurrently,
+which MUST NOT be 0 and MUST NOT exceed the MAX_TOP_TRACKS setup option
+value sent by the peer, either of which is a `PROTOCOL_VIOLATION`.
+
+TopTracksTimeout {{top-tracks-timeout}} is a Track Property
```
ianswett, line 1879: I would have expected this to be a part of the Top Tracks Filter, not a Track Property.  Is the reason to make it a Track Property just because you expect the Publisher to know better than the Subscriber?

draft-ietf-moq-transport.md
```diff
@@ -1845,14 +1847,111 @@ PUBLISH_NAMESPACE messages.
...
+that limits the number of milliseconds a selected track can remain
+selected without publishing an object with Property Type.
+Timeout in the next section refers to this Track Property.
+
+#### Track Selection
+
+Tracks are promoted to the selected state, demoted to the
+deselected state, or purged back to the unknown state as
+shown in the following diagram. Objects only pass the filter
+for tracks in the selected state, as shown below.
```
ianswett, line 1889:
```suggestion
shown in the following diagram. Only Objects from selected
Tracks are published.
```

draft-ietf-moq-transport.md
```diff
@@ -1845,14 +1847,111 @@ PUBLISH_NAMESPACE messages.
...
+limit state  |    +---> REQ UPD FWD=1 --->|   Tracks   |
+             |    |       Reselected      +------------+
+             |    |                             |
+        +--------------+                        |
+        |  DESELECTED  |<---- REQ UPD FWD=0 <---+
+        | out of Top N |     Demoted/Timeout
+        +--------------+
+
+          Objects FAIL                     Objects PASS
+           the filter                       the filter
```
ianswett, line 1909:
```suggestion
```

draft-ietf-moq-transport.md
```diff
@@ -1845,14 +1847,111 @@ PUBLISH_NAMESPACE messages.
...
+
+A track is selected if it publishes an object with the top N highest
+value for Property Type, where N = MaxTopTracks.  The publisher
+MUST send a PUBLISH message for each newly selected track.  For tracks
+with the same value, the track with the earlier delivered object wins
+the tie breaker, so a selected track remains selected until another track
+publishes a higher value that demotes it out of the top N, or Timeout
+elapses before the track delivers an object that remains in the top N,
+either of which deselect the track.  The publisher MUST send a
+REQUEST_UPDATE message with Forward=0 when a track is deselected.
```
ianswett, line 1920: Presumably this will need updating once #1820 lands?  Also, wouldn't it commonly make more sense to add/set an End Location on the Location Filter, rather than setting Forward, so the subscriber knows when the Subscription was paused?

draft-ietf-moq-transport.md
```diff
@@ -1845,14 +1847,111 @@ PUBLISH_NAMESPACE messages.
...
+        |  DESELECTED  |<---- REQ UPD FWD=0 <---+
+        | out of Top N |     Demoted/Timeout
+        +--------------+
+
+          Objects FAIL                     Objects PASS
+           the filter                       the filter
+~~~
+
+A track is selected if it publishes an object with the top N highest
+value for Property Type, where N = MaxTopTracks.  The publisher
```
ianswett, line 1913:
```suggestion
A track is selected if it has published an object with one of the top
MaxTopTracks highest values for Property Type.  The publisher
```

draft-ietf-moq-transport.md
```diff
@@ -1845,14 +1847,111 @@ PUBLISH_NAMESPACE messages.
...
+elapses before the track delivers an object that remains in the top N,
+either of which deselect the track.  The publisher MUST send a
+REQUEST_UPDATE message with Forward=0 when a track is deselected.
+The publisher MUST send a REQUEST_UPDATE message with Forward=1 when a
+deselected track is reselected, which also updates the Joining Location
+as described in {{joining-fetch-range-calculation}}.
+
+Recently deselected tracks SHOULD be kept in a list to avoid more PUBLISH
+messages in case a deselected track is reselected.  A relay SHOULD limit
+this list size to protect its resources by sending PUBLISH_DONE
```
ianswett, line 1927: If the Track is already paused and there's no need to ensure all the data arrived, then you can also reset the stream for the Subscription.

draft-ietf-moq-transport.md
```diff
@@ -1845,14 +1847,111 @@ PUBLISH_NAMESPACE messages.
...
+messages to purge state for the oldest deselected tracks, putting them
+back in the unknown state.  If a track is reselected after this, it is
+considered newly selected so the publisher MUST send a PUBLISH message
+again.  Relays SHOULD keep a list size large enough to avoid excessive
+control messages but small enough to avoid excessive old subscription state.
+
+If the TOP_TRACKS_FILTER parameter is updated, the publisher evaluates the
```
ianswett, line 1934: I assumed you could update the MaxTopTracks, but can you update the Parameter as well?

draft-ietf-moq-transport.md
```diff
@@ -1845,14 +1847,111 @@ PUBLISH_NAMESPACE messages.
...
+new filter to determine the new list of top N selected tracks and follows
+the same rules above for newly selected tracks in the top N (send PUBLISH),
+reselected tracks in the top N (send REQUEST_UPDATE with Forward=1), and
+newly deselected tracks demoted out of the top N (send REQUEST_UPDATE
+with Forward=0).  It can also purge excessive old deselected tracks as
+needed (send PUBLISH_DONE).
+
+If a Top Tracks Filter for a namespace overlaps with a direct subscription
```
ianswett, line 1942:
```suggestion
If a Top Tracks Filter for a namespace overlaps with a SUBSCRIBE
```

draft-ietf-moq-transport.md
```diff
@@ -1845,14 +1847,111 @@ PUBLISH_NAMESPACE messages.
...
+to a track name in the same namespace, it is considered to pass the Top Tracks
+Filter whether or not it is counted in the top N list, and MUST NOT
+be subject to filter state changes or actions.
```
ianswett, line 1945:
```suggestion
to a track name in the same namespace, its state MUST NOT be modified
by the Top Tracks Filter and it does not change the Top N computation.
For example, if the Track is not in the Top N, then the Subscriber will receive a total of
N+1 Subscriptions from that Track Namespace.
```

draft-ietf-moq-transport.md
```diff
@@ -1845,14 +1847,111 @@ PUBLISH_NAMESPACE messages.
...
+
+Objects received with a location lower than the largest received in that track
+are ignored for Top Tracks Filter state updates; however, if the track is already
+selected, they pass the filter.
```
ianswett, line 1949: I don't know what this is saying?  Are you trying to be specific about which Objects will be received from a Subscription?  If so, I'd suggest specifying the Location Filter Start and End when a Track is selected/deselected.

draft-ietf-moq-transport.md
```diff
@@ -2732,6 +2831,16 @@ Type which MUST be even, i.e. a single integer value
 (see {{moq-key-value-pair}}), otherwise the endpoint MUST reject this with
 REQUEST_ERROR with error code INVALID_FILTER. See {{range-filters}}.
 
+### TOP TRACKS FILTER Parameter {#top-tracks-filter}
+
+The TOP_TRACKS_FILTER parameter (Type=0x30) MAY appear in a SUBSCRIBE_TRACKS
+message or REQUEST_UPDATE for it.  It selects a specified
+number of tracks within a namespace with the highest Property Values for a
+specified Track or Object Property Type which MUST be even, i.e. a single
```
```suggestion
specified Track or Object Property Type. This Property Type MUST be even, i.e. a single
```

draft-ietf-moq-transport.md
```diff
@@ -2732,6 +2831,16 @@ Type which MUST be even, i.e. a single integer value
...
+integer value (see {{moq-key-value-pair}}), otherwise the endpoint MUST
+close the session with a `PROTOCOL_VIOLATION`.
+See {{top-tracks-filters}} and {{max-top-tracks}}.
```
ianswett: This paragraph doesn't say anything about the MAX TOP TRACKs Setup Option, so I wouldn't reference it here.

```suggestion
See {{top-tracks-filters}}.
```

draft-ietf-moq-transport.md
```diff
@@ -2957,6 +3066,14 @@ parameters for a given subscription or fetch.  The default value is 0, so if not
 specified, the peer MUST NOT send any such filter parameters.  If this limit is
 exceeded, an endpoint MUST reject this with REQUEST_ERROR with error code INVALID_FILTER.
 
+#### MAX TOP TRACKS
+
+The MAX_TOP_TRACKS option (Type 0x09) limits the peer's value of
+MaxTopTracks in the TOP_TRACKS_FILTER {{top-tracks-filter}} parameter.
+The default value is 0, so if not specified, the peer MUST NOT send any
```
```suggestion
The default value is 0, so if not specified, the peer MUST NOT send the
```

draft-ietf-moq-transport.md
```diff
@@ -4775,6 +4892,11 @@ the session with `PROTOCOL_VIOLATION`.
 
 If omitted, the value is 0.
 
+## TOP TRACKS TIMEOUT {#top-tracks-timeout}
+
+TOP_TRACKS_TIMEOUT (Property Type 0x32) is a Track Property.
+See {{top-tracks-filter}}. If omitted, the value is 1000 milliseconds.
```
ianswett, line 4898: 1000ms seems like a short default timeout.  I'm not sure what the right default value is, but I'd expect is similar to the Group length?

ianswett (commented):
A few overall comments I didn't fix:
1) The filter is now Top Tracks, which seems like a good name, but the PR talks about Top N still a lot.  It'd be clearer to say 'Top Tracks' for the filter and 'MaxSelectedTracks' for the field name.  If 'MaxSelectedTracks' is too long, I'd be fine with calling it 'MaxTracks' or something shorter.
2) There's no mention of Group boundaries here, but in order to consume a Track, presumably you need to start at the beginning of a Group?  If Object 2 in a Group puts the Track in the Top Tracks, wouldn't you want to start at the beginning of the Group, not Object 2?

draft-ietf-moq-transport.md
suhasHere: I think it complicates upstream especially when subscribers need different views. My preference is to avoid it unless there is a strong use-case.

vasilvv:
One thing I have recently realized is that for top-N to work, the relay most likely has to subscribe to all the original publishers without pausing or filtering (otherwise it risks getting incorrect view of top-N).  The PR should explain that, if it doesn't already.

mzanaty:
1. For naming, do people prefer MaxSelectedTracks over MaxTopTracks? I'm agnostic, but included top since it's used in similar features of other protocols/tools. The setup option should match this, so if we change one we should sync both.
2. For group boundaries, many apps would often update the property only on a new group, but there is no restriction. If the receiver gets an object mid-group and wants to start from 0, it can fetch. If zero latency joining fetch is needed, I suppose we could allow Fill in Sub Tracks to always ask for Current Group. But I'm not sure allowing Fill (or Location Filters) in Sub Tracks is a good idea.

draft-ietf-moq-transport.md
mzanaty: We can add Bottom N as an option in a later PR if people want it. I haven't heard strong demand for this, but we can ask.

draft-ietf-moq-transport.md
mzanaty: Alan wanted some text to say that old (OOO) objects are ignored for Top filter evaluation. So if an old object arrives with a different property value than the latest object, the old value won't override the latest value. The last bit clarifies that while an old object is ignored for filter evaluation, it can still be forwarded if already selected. So ignored for updating filter state, but not necessarily dropped (not ignored for forwarding).

draft-ietf-moq-transport.md
mzanaty: Relays generally don't know group length, although they could estimate it from past groups forwarded. I don't expect this property to be omitted if a track supports Top Tracks, so the default is for a corner/error case when omitted, not a typical value used for most tracks.

yuyou:
@mzanaty @suhasHere 

I need some confirmation/clarification while implementing the Track Filter in our Relay. Appreciate if you could help me out.

One architectural question while trying to implement the Track Filter: does the session handling the SUBSCRIBE_TRACKS request with the Top Tracks Filter actually receive the object-level events for all tracks in the namespace so it can evaluate their property values? Our implementation, e.g.  tracks active tracks at the PUBLISH level, not per-object, and there's no relay-side monitoring of objects for tracks the namespace-subscriber never explicitly subscribed to. without sending PUBLISH_OK back, then the relay won't actually receive the object stream for that track. So if the relay wants to filter tracks based on property values, it needs to actively subscribe to every track in the namespace upstream to observe those properties — even tracks that haven't been selected yet — which could mean opening subscriptions for video quality, bitrate, and other metadata across all available tracks.

Of course, this may be applied to the tier 1 relays only which face upstream origin publishers  (see the figure).

```
                    Top-4 at each tier

       100             100             100             100
        |               |               |               |
       [P]             [P]             [P]             [P]
        \               /               \               /
         \             /                 \             /
          \           /                   \           /
           +---------+                     +---------+
                |                               |
               200                             200
                |                               |
               [R]                             [R]
                \                               /
                 \ (8, 4 for each)             / (8, 4 for each)
                  \                           /
                   +-------------------------+
                              |
                             8 (4 for each)
                             [R]
                              |
                              4
                              |
                             (S)

                         Tracks carried
```

mzanaty:
Yu, relays should propagate the Top Tracks filter upstream to ensure they only get N tracks on each direct ingress connection. If an edge relay has P publishers directly connected on separate ingress connections that match the subscribed namespace, it must subscribe to all P publishers to inspect their objects and select the right tracks.

draft-ietf-moq-transport.md
mzanaty: Agreed, updated.

draft-ietf-moq-transport.md
mzanaty: Agreed, updated.

draft-ietf-moq-transport.md
mzanaty: Agreed, updated.

draft-ietf-moq-transport.md
mzanaty: Yes, timeout was originally part of the filter parameters, but moved to a track property since it should be set by the publisher for all subscribers.

draft-ietf-moq-transport.md
mzanaty: This is for reclaiming resources even if no data streams, to purge all subscription state (alias, tokens, internals). The data streams should have terminated normally with FIN or RESET before the track was deselected. This reclaims resources long after deselection when a track is unlikely to be reselected.

draft-ietf-moq-transport.md
mzanaty: Assuming you mean update the property type not parameter, yes, there is currently no restriction. Do you think we should disallow updating the property type? I'm fine with restricting updates to MaxTopTracks (N).

yuyou:
just confirm, when 2+ subscribers subscribe to the track on the same relay, would the relay coalesce  those requests or not towards upstream? I feel the coalescing should not be applied, as neither OR or AND would work.

```
P1 -- (top n filter 1) -------> R1 -- (top n AND/OR top m). --> R0
                         |     
P2-- (top m filter 2) ---+
```

suhasHere:
> just confirm, when 2+ subscribers subscribe to the track on the same relay, would the relay coalesce those requests or not towards upstream? I feel the coalescing should not be applied, as neither OR or AND would work.
> 
> ```
> P1 -- (top n filter 1) -------> R1 -- (top n AND/OR top m). --> R0
>                          |     
> P2-- (top m filter 2) ---+
> ```
@yuyou  in this case , R1 would do top(X), where x = greaterOf(m, n)

mzanaty:
Yu, filter 1 and 2 can be aggregated upstream (using max of n,m) only if they have the same property type and namespace. If not, they can't be aggregated.

yuyou:
I will update our Relay server with the Top Track Filter initial implementation. It supports SETUP negotiation and filters like object propertiees.

The relay-side aggregation has not been done yet but it would be the next feature to explore.

draft-ietf-moq-transport.md
mzanaty: Updated to use PUBLISH_STATE_NOTIFY. If PR#1825 replaces Forward with Location Filter, I will do the same here after it lands.

PR #1831, opened 2026-07-27
Change the maximum Stream Count bound in PUBLISH_DONE to 2^64 - 1

vasilvv:
The current value (2^62-1) is a leftover from when we used QUIC varints.


suhasHere (approved):

afrind:
There actually can’t be a stream count higher than 2^60 - 1, because you are fundamentally limited by the number of uni QUIC streams the peer can open, which is 2^62 / 4.

I’m going to assert that if the number is in excess of 2^60, or even 2^40, it’s almost certainly a bug or an attack, and everyone’s stack will wait for the timer and deliver pub done. 60 streams per second will get you 2^31 per year.

That said, if we want to remove ceiling entirely, I don’t see any harm.

afrind (approved):

PR #1833, opened 2026-07-27
Remove PUBLISH_DONE reason SUBSCRIPTION_ENDED

### LINKED ISSUE #1832: Publish Done should not depend on subscriber location filter
Publish Done defines a reason:
SUBSCRIPTION_ENDED (0x3): The publisher reached the end of an associated location filter. 

This seems wrong, especially since filters can be updated. Only the original publisher should be able to permanently end a track regardless of subscriber filters.

mzanaty:
Fixes #1832

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1615,6 +1615,9 @@ A publisher MUST NOT send subscription-delivered objects from outside the
 requested range.  Because updating filters is asynchronous,
 subscribers can receive objects outside the current filter.
 
+A publisher does not end a subscription solely because the Largest Object advances
+past the end of the current Location Filter.
+
 Fill-delivered objects are governed by the Location filter in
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

ianswett (approved):
I think this is good, but it also goes back to what exactly is PUBLISH_DONE good for?  I guess there's the stream tracking to ensure you've received everything?

afrind:
PUBLISH_DONE still has utility, but this particular code isn't helpful given that one can update the filter range, and a relay can almost never tell with certainty that a subscriber's filter was fully satisfied.

I think Top-N and SWITCH_FROM can generate PUBLISH_DONE from a relay other than "track ended/error" but they can specify their own codes.

afrind (approved):

martinduke:
Can we add editorial language under EndGroup and location filters that the publisher will never send PUBLISH_DONE and the onus is on the subscriber to terminate these?

PR #1834, opened 2026-07-28
PUBLISH can contain Subscription Parameters

### LINKED ISSUE #1803: How do you know what Parameters a PUBLISH has?

ianswett:
Fixes #1803

I didn't add any normative language that they MUST be specified, otherwise the Subscriber will assume the default value for the Parameter, but maybe I should?


draft-ietf-moq-transport.md
```diff
@@ -3425,7 +3426,10 @@ PUBLISH Message {
 * Track Alias: The identifer used for this track in Subgroups or Datagrams (see
   {{track-alias}}).
 
-* Parameters: The parameters are defined in {{message-params}}.
+* Parameters: The parameters are defined in {{message-params}}. Parameters such
+  as FORWARD, GROUP_ORDER, SUBSCRIBER_PRIORITY, SUBGROUP_DELIVERY_TIMEOUT,
+  OBJECT_DELIVERY_TIMEOUT appear in the Parameters of a PUBLISH to inform the
+  Subscriber of the initial Subscription parameters.
```
```suggestion
  OBJECT_DELIVERY_TIMEOUT, and LOCATION FILTER can appear in the Parameters
  of a PUBLISH to inform the Subscriber of the initial Subscription parameters.
```

vasilvv (approved):

afrind (approved):

draft-ietf-moq-transport.md
```diff
@@ -3443,7 +3444,10 @@ PUBLISH Message {
 * Track Alias: The identifer used for this track in Subgroups or Datagrams (see
   {{track-alias}}).
 
-* Parameters: The parameters are defined in {{message-params}}.
+* Parameters: The parameters are defined in {{message-params}}. Parameters such
+  as FORWARD, GROUP_ORDER, SUBSCRIBER_PRIORITY, SUBGROUP_DELIVERY_TIMEOUT,
+  OBJECT_DELIVERY_TIMEOUT, and LOCATION FILTER can appear in the Parameters
+  of a PUBLISH to inform the Subscriber of the initial Subscription parameters.
```
afrind, line 3503:
```suggestion
  of a PUBLISH to inform the Subscriber of the initial Subscription parameters.
  If the PUBLISH is the result of a SUBSCRIBE_TRACKS, the parameters are handled
  as described in {{parameters-on-subscribe_tracks}}, otherwise, they represent
  the publisher's initial settings for the subscription, which the subscriber can
  change.
```

draft-ietf-moq-transport.md
```diff
@@ -4038,13 +4046,13 @@ SUBSCRIBE_TRACKS is not required for a publisher to send PUBLISH messages to
...
-### Parameters on SUBSCRIBE_TRACKS
+### Parameters on SUBSCRIBE_TRACKS {#parameters-on-subscribe-tracks}
 
 Any Parameter that can be specified on a Subscription (ie: in SUBSCRIBE) is valid
 in SUBSCRIBE_TRACKS, unless otherwise specified. These parameters are used by the
 publisher as the initial Subscription parameters when a PUBLISH is sent as a result of
-SUBSCRIBE_TRACKS. The Parameters are not explicitly communicated, with the
-exception of FORWARD and GROUP_ORDER as described below.
+SUBSCRIBE_TRACKS. The Parameters SHOULD be explicitly communicated, including the
+FORWARD and GROUP_ORDER parameters as described below.
```
afrind: Rather than SHOULD/MUST on the publisher, what if we turn it around? The subscriber assumes they are there.  If they are absent, it means the publisher used the default for that parameter.

```suggestion
SUBSCRIBE_TRACKS. These Parameters are explicitly communicated, including the
FORWARD and GROUP_ORDER parameters as described below.  When omitted by
the publisher in PUBLISH, the subscriber uses the default value for each.
```

ianswett (commented):
Thanks @afrind for the suggestion.  Sorry @suhasHere I accidentally deleted your suggestion, but PTAL at what's there now.

suhasHere (approved):

PR #1851, opened 2026-08-08
Clarify Range Filters section for readability (#1810)

### LINKED ISSUE #1810: New section on filters hard to process
I'm going through the changes in the new draft-19, and the whole section on filters (5.1.3) is very hard to process for me, let alone implement. The exact format of how to serialize each filter is, IMHO, not clear at all (at least to a non-native speaker like I am). The syntax is presented like this:

```
SUBGROUP_FILTER { Type=0x25, Length, [SetID], Range... }	
OBJECTID_FILTER { Type=0x26, Length, [SetID], Range... }	
PRIORITY_FILTER { Type=0x27, Length, [SetID], Range... }	
OBJECT_PROPERTY_FILTER { Type=0x28, Length, [SetID], [Property Type], Range... }	
TRACK_PROPERTY_FILTER  { Type=0x29, Length, [SetID], [Property Type], Range... }
Range { Start, [End] }
```

The text says:

> Each Range Filter is a sequence of Start/End (vi64) inclusive Rang pairs prefixed with a Length (vi64) in bytes and a SetID (8 bits).

but that doesn't help. Am I writing/parsing Length and SetID once, and then I can have multiple Ranges? Ot is the whole Length+SetID+Range that can be repeated multiple tiimes, over the same filter? Besides, SetID is presented in square brackets, that in the rest of the document usually means "it's optional": the text doesn't say what makes it optional, though, it only says it's an 8-bit value, not what rule to follow to add or skip it. Not knowing how many times SetID can appear in each filter parameter, it's also very unclear to me how the OR/AND part is supposed to work.

I'm sure this has been discussed at length, but I haven't followed any interim, so you can consider me a dumb reader that sees this text for the first time (which is what most MoQ implementers will be anyway), so I guess the text should be made clearer than it currently is. Not knowing what the actual meaning is, I can't provide suggestions at this stage on how to improve the text.

suhasHere:
Addresses part of #1810  - Set ID changes are pending
I do feel current text is still fine, but I did readjust some text to clarify the flow and also made purely editorial changes to keep filter type defined as multi-line

@mzanaty please see if this makes sense.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1730,16 +1730,14 @@ Range {
 ~~~
 
 Length (vi64) is the byte count of all fields after itself.  When Length
-is 0 (used in REQUEST_UPDATE to remove a filter), no further fields are
-is 0, there is no filter and no further fields are present.  This can
-be used in REQUEST_UPDATE to remove a filter. 
-The Object Property and Track Property Filters include a
-Property Type (vi64) which follows SetID.
+is 0, there is no filter and no further fields are present.  This can be
+used in REQUEST_UPDATE to remove a filter.  The Object Property and Track
+Property Filters include a Property Type (vi64) which follows SetID.
 
-Each Range is an inclusive Start/End pair.  End is optional in the last pair; if omitted
-it indicates the last Range is open-ended.  An object matches the filter if its
-value falls within any Range (i.e., Ranges are OR'd within a filter
-parameter).
+Each Range is an inclusive Start/End pair.  End is optional in the last
+pair; if omitted it indicates the last Range is open-ended.  An object
+matches the filter if its value falls within any Range (i.e., Ranges are
+OR'd within a filter parameter).
 
 Each Start is delta encoded from the prior Range's End (or from 0 for the
 first Range), and End is delta encoded from its own Start.  If adding the delta
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1609,31 +1609,66 @@ Object header fields (Subgroup ID, Object ID, and Publisher Priority).
 There are five Range Filter parameter types, 0x25-0x29, as shown below.
 
 ~~~
-SUBGROUP_FILTER { Type=0x25, Length, [SetID], Range... }
-OBJECTID_FILTER { Type=0x26, Length, [SetID], Range... }
-PRIORITY_FILTER { Type=0x27, Length, [SetID], Range... }
-OBJECT_PROPERTY_FILTER { Type=0x28, Length, [SetID], [Property Type], Range... }
-TRACK_PROPERTY_FILTER  { Type=0x29, Length, [SetID], [Property Type], Range... }
-Range { Start, [End] }
+SUBGROUP_FILTER {
+  Type (vi64) = 0x25,
+  Length (vi64),
+  [SetID (8)],
+  [Range (..) ...]
+}
+
+OBJECTID_FILTER {
+  Type (vi64) = 0x26,
+  Length (vi64),
+  [SetID (8)],
+  [Range (..) ...]
+}
+
+PRIORITY_FILTER {
+  Type (vi64) = 0x27,
+  Length (vi64),
+  [SetID (8)],
+  [Range (..) ...]
+}
+
+OBJECT_PROPERTY_FILTER {
+  Type (vi64) = 0x28,
+  Length (vi64),
+  [SetID (8)],
+  [Property Type (vi64)],
+  [Range (..) ...]
+}
+
+TRACK_PROPERTY_FILTER {
+  Type (vi64) = 0x29,
+  Length (vi64),
+  [SetID (8)],
+  [Property Type (vi64)],
+  [Range (..) ...]
+}
+
+Range {
+  Start (vi64),
+  [End (vi64)]
+}
 ~~~
 
-Each Range Filter is a sequence of Start/End (vi64) inclusive Range pairs
-prefixed with a Length (vi64) in bytes and a SetID (8 bits).
-The Track and Object Property Filters include an additional prefix for Property Type (vi64).
-The final End in a sequence of Ranges can be omitted to indicate no end.
+Length (vi64) is the byte count of all fields after itself.  When Length
+is 0 (used in REQUEST_UPDATE to remove a filter), no further fields are
+present.  The Object Property and Track Property Filters include a
+Property Type (vi64) between SetID and the Ranges.
 
-Start is delta encoded from the prior Range's End or from 0
-for the first Range, and End is delta encoded from the current Range's Start.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1609,31 +1609,66 @@ Object header fields (Subgroup ID, Object ID, and Publisher Priority).
...
+}
 ~~~
 
-Each Range Filter is a sequence of Start/End (vi64) inclusive Range pairs
-prefixed with a Length (vi64) in bytes and a SetID (8 bits).
-The Track and Object Property Filters include an additional prefix for Property Type (vi64).
-The final End in a sequence of Ranges can be omitted to indicate no end.
+Length (vi64) is the byte count of all fields after itself.  When Length
+is 0 (used in REQUEST_UPDATE to remove a filter), no further fields are
+present.  The Object Property and Track Property Filters include a
```
ianswett, line 1657: I tried to clarify, but I am not sure what filter is being removed in all cases?  Does specifying OBJECT_PROPERTY_FILTER with Length of 0 remove all Object property filters for all SetIDs?  Is that what we want?

```suggestion
is 0, there is no filter and no further fields are present.  This can
be used in REQUEST_UPDATE to remove a filter. 
The Object Property and Track Property Filters include a
```

draft-ietf-moq-transport.md
```diff
@@ -1609,31 +1609,66 @@ Object header fields (Subgroup ID, Object ID, and Publisher Priority).
...
-Start is delta encoded from the prior Range's End or from 0
-for the first Range, and End is delta encoded from the current Range's Start.
-Any delta encoding that results in a value that exceeds 2^64-1
-MUST be rejected with REQUEST_ERROR with error code INVALID_FILTER.
-For example, to express ranges 3-5 and 10-15: the first Start is 3
-(delta from 0), the first End is 2 (5 minus 3), the second Start is 5
-(10 minus 5), and the second End is 5 (15 minus 10).
+Start is delta encoded from the prior Range's End (or from 0 for the
+first Range), and End is delta encoded from its own Start.  A delta
+that exceeds 2^64-1 MUST be rejected with REQUEST_ERROR / INVALID_FILTER.
```
```suggestion
first Range), and End is delta encoded from its own Start.  If adding the delta
would exceed 2^64-1, the request MUST be rejected with `INVALID_FILTER`.
```

draft-ietf-moq-transport.md
```diff
@@ -1609,31 +1609,66 @@ Object header fields (Subgroup ID, Object ID, and Publisher Priority).
...
+For example, ranges 3-5 and 10-15 encode as: Start=3, End=2, Start=5, End=5.
 
-All filter parameters with the same SetID value are combined using logical
-"AND" operations, then all the resulting sets are combined using logical
-"OR" operations.  The final result is SetID=0 OR SetID=1 OR ... SetID=255,
-where each SetID=i is the AND of filters with SetID=i.
+Filter parameters with the same SetID are AND'd; distinct SetIDs are OR'd.
```
ianswett, line 1669: It'd be good to give an overview of the SetID concept in its own section before the framing, because it seems like an afterthought here.

draft-ietf-moq-transport.md
```diff
@@ -1643,10 +1678,9 @@ message.  If the same combination of Parameter Type, SetID, and Property Type
 (only in the Track and Object Property Filters) repeat in any message,
 an endpoint MUST reject this with REQUEST_ERROR with error code INVALID_FILTER.
 
-In REQUEST_UPDATE, Length can be 0 to remove a filter parameter or non-zero
-to replace that entire filter parameter including all sets and Property Types.
-If a filter parameter is omitted from REQUEST_UPDATE, the value is unchanged.
-If omitted from other messages, the default is no filter.
+In REQUEST_UPDATE, Length of 0 removes the filter; non-zero replaces it
+entirely.  If a filter parameter is omitted from REQUEST_UPDATE, it is
```
ianswett, line 1682: What about SetID?

ianswett (commented):
Thanks for the improvements.  Now that its clearer, I don't think the language about how to handle SetIDs is correct?  Or maybe its unclear?

draft-ietf-moq-transport.md
```diff
@@ -1609,31 +1609,66 @@ Object header fields (Subgroup ID, Object ID, and Publisher Priority).
...
 ~~~
 
-Each Range Filter is a sequence of Start/End (vi64) inclusive Range pairs
-prefixed with a Length (vi64) in bytes and a SetID (8 bits).
-The Track and Object Property Filters include an additional prefix for Property Type (vi64).
-The final End in a sequence of Ranges can be omitted to indicate no end.
+Length (vi64) is the byte count of all fields after itself.  When Length
+is 0 (used in REQUEST_UPDATE to remove a filter), no further fields are
+present.  The Object Property and Track Property Filters include a
+Property Type (vi64) between SetID and the Ranges.
```
afrind, line 1658: Awkward use of "between".  I think it may be fine to say Property Type is after SetID.

draft-ietf-moq-transport.md
```diff
@@ -1609,31 +1609,66 @@ Object header fields (Subgroup ID, Object ID, and Publisher Priority).
...
+
+Each Range is an inclusive Start/End pair.  The final End can be omitted
```
afrind, line 1660: "final End" is a little compact?  "End is required in all but the last pair"?

draft-ietf-moq-transport.md
```diff
@@ -1686,31 +1686,67 @@ Object header fields (Subgroup ID, Object ID, and Publisher Priority).
...
 ~~~
 
-Each Range Filter is a sequence of Start/End (vi64) inclusive Range pairs
-prefixed with a Length (vi64) in bytes and a SetID (8 bits).
-The Track and Object Property Filters include an additional prefix for Property Type (vi64).
-The final End in a sequence of Ranges can be omitted to indicate the Range is open-ended.
+Length (vi64) is the byte count of all fields after itself.  When Length
+is 0 (used in REQUEST_UPDATE to remove a filter), no further fields are
+present.  The Object Property and Track Property Filters include a
+Property Type (vi64) between SetID and the Ranges.
```
```suggestion
Property Type (vi64) which follows SetID.
```

draft-ietf-moq-transport.md
```diff
@@ -1686,31 +1686,67 @@ Object header fields (Subgroup ID, Object ID, and Publisher Priority).
...
+
+Each Range is an inclusive Start/End pair.  The final End can be omitted
+to indicate the Range is open-ended.  An object matches the filter if its
```
```suggestion
Each Range is an inclusive Start/End pair.  End is optional in the last pair; if omitted
it indicates the last Range is open-ended.  An object matches the filter if its
```

draft-ietf-moq-transport.md
afrind: Ian to follow up on Set ID clarifications

afrind (approved):

PR #1852, opened 2026-08-08
Restrict GOAWAY on request streams to the request receiver (#1655)

### LINKED ISSUE #1655: Tighter language around GOAWAYs on a request stream
Section 10.4 says "A GOAWAY MAY also be sent on a request stream to initiate migration of that individual request." and "A client MUST send a zero-length New Session URI in any GOAWAY, as clients cannot instruct servers to initiate connections."

We may want to add some further restrictions to this:
- Only the recipient of a SUBSCRIBE can send a GOAWAY for it.
- Only the sender of a PUBLISH can send a GOAWAY for it.
- Only the recipient of a FETCH can send a GOAWAY for it.

In general, only the endpoint that is sending the data for the track should be able to send a GOAWAY. I don't really see why the other endpoint would want to send a GOAWAY for it.

suhasHere:
Addresses #1655 by clarifying who can send `GOAWAY` on a request stream.

Note: this diverges from the original proposal in #1655. Issue #1655 proposed restricting request-stream `GOAWAY` to the **data sender** (publisher). Based on review feedback (thanks @afrind), the rule was generalized to the **request receiver** so it applies uniformly across request types — including the case where a relay acting as a subscriber wants to redirect an upstream publisher via `GOAWAY` on the request stream.

@sharmafb — flagging in case the broader rule is a concern relative to the original intent of #1655.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3227,6 +3227,22 @@ mechanism (e.g. FIN, stream reset, or PUBLISH_DONE). This allows, for example,
 a relay to move an individual subscription or publication to another relay
 without draining its entire session.
 
+The rule is symmetric across both roles and both request verbs — whoever
+received the request may ask the requester to re-issue it:
+
+| # | Requester        | Request   | GOAWAY sender (receiver) | New Session URI allowed? |
+|---|------------------|-----------|--------------------------|--------------------------|
+| 1 | client publisher | PUBLISH   | server (subscriber)      | yes                      |
+| 2 | client subscriber| SUBSCRIBE | server (publisher)       | yes                      |
+| 3 | server publisher | PUBLISH   | client (subscriber)      | no (empty, same session) |
+| 4 | server subscriber| SUBSCRIBE | client (publisher)       | no (empty, same session) |
+
+The same shape applies to FETCH, TRACK_STATUS, SUBSCRIBE_NAMESPACE, and
+PUBLISH_NAMESPACE. GOAWAY on a request stream never switches request verb —
+the requester re-issues the same verb it originally sent. Draining is handled
+by the existing per-request mechanisms (`PUBLISH_DONE`, unsubscribe, stream
+close), not by GOAWAY.
+
 The GOAWAY message does not impact subscription state. A subscriber
 SHOULD individually unsubscribe from each existing subscription, while a
 publisher MAY reject new requests after sending a GOAWAY.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -3011,13 +3011,25 @@ session migration ({{session-migration}}) with an optional URI.  A client MUST
...
-stream using the appropriate mechanism (e.g. FIN, stream reset, or PUBLISH_DONE).
-This allows, for example, moving the publishers and subscribers of a common set
-of tracks to a common relay without draining their entire session.
+The publisher MAY send a `GOAWAY` on a request stream to initiate migration of
+that individual request.  For subscriptions established via SUBSCRIBE or FETCH,
+the publisher is the responder; for subscriptions established via PUBLISH, the
+publisher is the requester.  In all cases, only the publisher (the endpoint
+sending objects for the track) MAY send a GOAWAY on the request stream.  An
+endpoint that receives a GOAWAY from the subscriber on a request stream MUST
+close the session with a `PROTOCOL_VIOLATION`.
```
afrind, line 3020: Do we want this restriction?  What about a case where the client is a publisher and the server is a relay, and it wants to move this track to another relay?  Isn't that a case where "the subscriber sends a GOAWAY" -- eg: go publish that track over here now.

draft-ietf-moq-transport.md
```diff
@@ -3011,13 +3011,25 @@ session migration ({{session-migration}}) with an optional URI.  A client MUST
...
+
+For SUBSCRIBE_NAMESPACE, SUBSCRIBE_TRACKS, and TRACK_STATUS, only the
+responder (the endpoint that received the request) MAY send a GOAWAY on the
```
afrind, line 3023: This is a publisher

draft-ietf-moq-transport.md
```diff
@@ -3011,13 +3011,25 @@ session migration ({{session-migration}}) with an optional URI.  A client MUST
...
+request stream.  For PUBLISH_NAMESPACE, only the requester (the publisher
```
afrind, line 3024: This doesn't make sense.  PUBLISH_NAMESPACE is sent from client to relay, and they are _unsolicited_.  Again I think the subscriber (relay) can send a GOAWAY asking "please advertise yourself as a publisher on this other relay".

draft-ietf-moq-transport.md
```diff
@@ -3011,13 +3011,25 @@ session migration ({{session-migration}}) with an optional URI.  A client MUST
...
+offering its namespace) MAY send a GOAWAY.
+
+Upon receiving a GOAWAY on a request stream, the receiver SHOULD re-issue
+that specific request on a session at the specified URI (or the current session
```
afrind, line 3028: See above, the PR currently has the initiator of non-subscription requests sending this message.

afrind (changes requested):
This needs a rethink

PR #1853, opened 2026-08-09
Rename application-specific to private use in IANA registry ranges 

### LINKED ISSUE #1828: Revise the name application-specfic to private to be umbaiguous for registry spaces
this came as part of msf discussion at IETF Vienna

suhasHere:
Fixes #1828

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -792,8 +792,8 @@ are preceded by an explicit length field.
 Property types are registered in the IANA table 'MOQ Properties'.
 See {{iana}}.
 
-Certain Property type ranges are reserved for application-specific
-use and will never be allocated by IANA in future MOQT specifications:
+Certain Property type ranges are reserved for private use and will
+never be allocated by IANA in future MOQT specifications:
 
 * 0x78 to 0x7F (1-byte encoding): 8 code points for applications with
   tight space constraints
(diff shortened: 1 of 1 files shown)


vasilvv:
I am not entirely sure this is consistent with how RFC 8126 defines Private Use:

> Private Use is for private or local use only, with the type and
   purpose defined by the local site.  No attempt is made to prevent
   multiple sites from using the same value in different (and
   incompatible) ways.  IANA does not record assignments from registries
   or ranges with this policy (and therefore there is no need for IANA
   to review them) and assignments are not generally useful for broad
   interoperability.

Now, the application-specific ranges are useful for interoperability; for example, MSF could define those, and those would be interoperable.

afrind:
> Now, the application-specific ranges are useful for interoperability; for example, MSF could define those, and those would be interoperable.

I think for the small number of code points we're carving out here, private use is correct.  e.g. IANA can assign 120 of the 1-byte values (including to MSF) but the last 8 are the wild west.

suhasHere:
> Now, the application-specific ranges are useful for interoperability; for example, MSF could define those, and those would be interoperable.

@vasilvv We did discuss this in length at IETF126 . The challenge is, application-specific, is not defined term though

afrind (approved):

vasilvv:
By "not defined", do you mean "not defined in terms of IANA"?  You don't really need IANA to define it, as far as IANA is concerned, the application-specific spaces are just defined in `draft-ietf-moq-transport`.  We kind of do similar thing with webtrans reset codes if I remember correctly.

PR #1855, opened 2026-08-14
Exclude URI query component from MOQT scope

### LINKED ISSUE #1835: Is query in URI scoping?
Is the draft's Scope section talking about `path` in RFC 3986 sense, or PATH option/HTTP `:path` sense? 

### Background:
- [RFC 3986 | Section 3](https://www.rfc-editor.org/info/rfc3986/#section-3) defines `path` and `query` as two distinct components
- HTTP/2 and HTTP/3 (and thereof WT) defines: The ":path" pseudo-header field includes the path and query parts of the target URI

> **10.3.1.2. PATH**
> When connecting to a server using a URI with the "moqt" scheme, the client MUST set the PATH option to the path-abempty portion of the URI; if query is present, the client MUST concatenate ?, followed by the query portion of the URI to the option.

afrind:
Name the RFC 3986 components explicitly in the scope-equality rule and state that the query component is not part of the scope.

Fixes: #1835

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -730,10 +730,15 @@ require the scope to include all of those CDNs.
 
 A single MOQT transport session is tied to the scope that is negotiated in the
 beginning of the session. Unless the application has additional information,
-two tracks are assumed to belong to the same scope if and only if the authority
-and the path values are equal. The authority and the path values are
-communicated through the SETUP message in case of raw QUIC, and through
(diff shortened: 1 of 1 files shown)


vasilvv (approved):
This does seem counterintuitive to me, but I do not see any strong reason to do this one way or another.

ianswett (approved):

suhasHere (approved):

PR #1856, opened 2026-08-20
Beware re-using a Track Alias

### LINKED ISSUE #1795: Track Aliases: allow re-use or not?
Section 11.1 allows Track Alias reuse when not simultaneous. However, I think we should outright disallow it. A datagram can arrive arbitrarily late and cause ambiguity, so I don't think it's ever safe to reuse.

ianswett:
For a different Track

Fixes #1795

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -4119,6 +4119,10 @@ Objects can arrive after a subscription has been cancelled.  Subscribers SHOULD
 retain sufficient state to quickly discard these unwanted Objects, rather than
 treating them as belonging to an unknown Track Alias.
 
+To avoid a protocol violation and to ensure the Subscriber knows which Track
+Objects are from, Publishers SHOULD NOT reuse a Track Alias within a session
+unless it is certain the prior Subscription has been completely closed and
+no Objects are scheduled to be sent or in flight.
 
 ## Objects {#message-object}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -4115,14 +4115,22 @@ PUBLISH or SUBSCRIBE_OK that uses the same Track Alias as a different Track
 with an `Established` subscription, it MUST close the session with error
 `DUPLICATE_TRACK_ALIAS`.
 
-Objects can arrive after a subscription has been cancelled.  Subscribers SHOULD
-retain sufficient state to quickly discard these unwanted Objects, rather than
-treating them as belonging to an unknown Track Alias.
+Objects can be sent before the Subscriber knows the Track Alias, requiring
+buffering Objects with an unknown Track Alias. If a Track Alias
+is used for two concurrent subscriptions to the same Track, the Subscriber needs
+Objects to be filtered by each Subscription's filters. Because
+a Publisher can re-use a Track Alias for the second Subscription to a Track,
+the Subscriber needs to act as though the Track Alias is reused until told
+otherwise, in order to avoid missing Objects.
 
 To avoid a protocol violation and to ensure the Subscriber knows which Track
-Objects are from, Publishers SHOULD NOT reuse a Track Alias within a session
-unless it is certain the prior Subscription has been completely closed and
-no Objects are scheduled to be sent or in flight.
+Objects are from, Publishers SHOULD NOT reuse a Track Alias for different Tracks
+within a session, unless it is certain the prior Subscription has been
+completely closed and no Objects are scheduled to be sent or in flight.
+
+Objects can arrive after a subscription has been cancelled.  Subscribers SHOULD
+retain sufficient state to quickly discard these unwanted Objects, rather than
+treating them as belonging to an unknown Track Alias. 
 
 ## Objects {#message-object}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -4130,7 +4130,8 @@ completely closed and no Objects are scheduled to be sent or in flight.
 
 Objects can arrive after a subscription has been cancelled.  Subscribers SHOULD
 retain sufficient state to quickly discard these unwanted Objects, rather than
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -4119,6 +4119,10 @@ Objects can arrive after a subscription has been cancelled.  Subscribers SHOULD
 retain sufficient state to quickly discard these unwanted Objects, rather than
 treating them as belonging to an unknown Track Alias.
 
+To avoid a protocol violation and to ensure the Subscriber knows which Track
+Objects are from, Publishers SHOULD NOT reuse a Track Alias within a session
```
afrind, line 4123: Did you intentionally leave off "for a different Track"?  e.g. the rules for reusing the alias for the same track also require you to know the previous subscription is done?

The draft says the publisher must not send objects outside the filter.  So if I 

SUBSCRIBE t1 10,0 - 11
SUB_OK Alias=X
SUBSCRIBE t1 12,0 - 13
SUB_OK Alias=X  // delayed in network
OBJECT AlIas=X 12,0 // receiver assumes this is part of the first t1 subscription, the second one never receives.

ianswett: The example you give is concerning, and I'm not sure how best to handle it except to highlight the issue and advise people handle it?

Popping up a level, I don't think a subscriber can enforce that MUST, because you could have a sequence of REQUEST_UPDATES and there are still Objects in flight that are from a prior filter?

draft-ietf-moq-transport.md
```diff
@@ -4115,10 +4115,22 @@ PUBLISH or SUBSCRIBE_OK that uses the same Track Alias as a different Track
...
+To avoid a protocol violation and to ensure the Subscriber knows which Track
+Objects are from, Publishers SHOULD NOT reuse a Track Alias for different Tracks
+within a session, unless it is certain the prior Subscription has been
+completely closed and no Objects are scheduled to be sent or in flight.
+
 Objects can arrive after a subscription has been cancelled.  Subscribers SHOULD
 retain sufficient state to quickly discard these unwanted Objects, rather than
-treating them as belonging to an unknown Track Alias.
-
+treating them as belonging to an unknown Track Alias.
```
```suggestion
treating them as belonging to an unknown Track Alias.

```

draft-ietf-moq-transport.md
ianswett: I took another stab at this.  Its sort of ugly, but I think the new text is more accurate.

draft-ietf-moq-transport.md
```diff
@@ -4115,6 +4115,19 @@ PUBLISH or SUBSCRIBE_OK that uses the same Track Alias as a different Track
 with an `Established` subscription, it MUST close the session with error
 `DUPLICATE_TRACK_ALIAS`.
 
+Objects can be sent before the Subscriber knows the Track Alias, requiring
+buffering Objects with an unknown Track Alias. If a Track Alias
+is used for two concurrent subscriptions to the same Track, the Subscriber needs
+Objects to be filtered by each Subscription's filters. Because
+a Publisher can re-use a Track Alias for the second Subscription to a Track,
+the Subscriber needs to act as though the Track Alias is reused until told
+otherwise, in order to avoid missing Objects.
```
```suggestion
buffering Objects with an unknown Track Alias. 
```

suhasHere (commented):
looks fine, with some simplification

draft-ietf-moq-transport.md
ianswett: Shouldn't I say something about this complexity?  It seems like a potential gotcha for client implementations and I wouldn't want all clients to have to re-discover this.

afrind: Buffering or filtering alone is insufficient to handle the case I suggested.  We can just warn?  Something like this 

> A subscriber that receives objects can deliver them immediately to subscription(s) known to use the associated alias. Reusing the same Track Alias for concurrent subscriptions the same Track can lead to missed delivery to a new subscription if the objects arrive before the control message establishing the shared Alias.

draft-ietf-moq-transport.md
```diff
@@ -4124,6 +4124,19 @@ PUBLISH or SUBSCRIBE_OK that uses the same Track Alias as a different Track
...
+
+To avoid a protocol violation and to ensure the Subscriber knows which Track
+Objects are from, Publishers SHOULD NOT reuse a Track Alias for different Tracks
```
vasilvv, line 4139: Does anyone actually want to use same alias for different tracks?  Can we just ban that?

ianswett: I believe the WG wanted to allow it, partially because preventing it is difficult in the general case.

draft-ietf-moq-transport.md
```diff
@@ -4124,6 +4124,19 @@ PUBLISH or SUBSCRIBE_OK that uses the same Track Alias as a different Track
 with an `Established` subscription, it MUST close the session with error
 `DUPLICATE_TRACK_ALIAS`.
 
+Objects can be sent before the Subscriber knows the Track Alias, requiring
+buffering Objects with an unknown Track Alias. If a Track Alias
+is used for two concurrent subscriptions to the same Track, the Subscriber needs
+Objects to be filtered by each Subscription's filters. Because
+a Publisher can re-use a Track Alias for the second Subscription to a Track,
+the Subscriber needs to act as though the Track Alias is reused until told
+otherwise, in order to avoid missing Objects.
```
```suggestion
buffering Objects with an unknown Track Alias. If a Track Alias
is used for two concurrent subscriptions to the same Track, an
Object that arrives with the Track Alias could be for either
Subscription. Reusing the same Track Alias for concurrent
subscriptions to the same Track can lead to missed delivery
if objects for the new subscription arrive
before the control message establishing the shared Alias.
The Subscriber can assume the Track Alias is reused until
told otherwise, in order to avoid missing Objects.
```

afrind (approved):

suhasHere (approved):

vasilvv (approved):

PR #1858, opened 2026-08-24
Add Parameters back to NAMESPACE

### LINKED ISSUE #1800: NAMESPACE doesn't allow parameters like PUBLISH_NAMESPACE
I realized when writing my extension that it can't work with SUBSCRIBE_NAMESPACE. I need some way to indicate the number of hops for a given NAMESPACE: https://github.com/moq-dev/drafts/blob/main/draft-lcurley-moq-relay-hops.md

This is also important because relays are expected to wait for a SUBSCRIBE_NAMESPACE, while clients are expected to send a PUBLISH_NAMESPACE. However this means any parameters from the original publisher can't be proxied by a relay.

Related to #1798

ianswett:
Fixes #1800

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3402,6 +3402,7 @@ NAMESPACE Message {
   namespace as defined in {{track-namespace-structure}} after removing
   namespace tuples included in 'Track Namespace Prefix'
   {{message-subscribe-ns}}.
+
 * Parameters: The parameters are defined in {{message-params}}.
 
 ## NAMESPACE_DONE {#message-namespace-done}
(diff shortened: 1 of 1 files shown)


afrind (approved):
Someone might wonder why we allow params here but none are defined.

draft-ietf-moq-transport.md
```diff
@@ -3397,6 +3402,8 @@ NAMESPACE Message {
   namespace as defined in {{track-namespace-structure}} after removing
   namespace tuples included in 'Track Namespace Prefix'
   {{message-subscribe-ns}}.
+
```
```suggestion

```

suhasHere (approved):

fluffy (approved):

PR #1865, opened 2026-08-31
Add a diagram illustrating the structure of a MOQT scope

vasilvv:


afrind (commented):
I'm not sure this diagram makes clear what is confusing about scope - the hierarchal nature of namespaces and tracks within a single scope seems pretty straightforward.  The important bit is that two scopes can use the same name to refer to different tracks.  What confuses people is "how do I define a scope".

PR #1866, opened 2026-08-31
Add a diagram illustrating Groups / Subgroups / Objects

vasilvv:


draft-ietf-moq-transport.md
```diff
@@ -579,6 +579,42 @@ stream is the first object ever published in that subgroup. A relay forwarding a
...
+:      ^       +-------+-------+-------+       :  +-------+-------------+
+:      |           ^                           :
+:      |           |                           :
+:      |           +-- FIRST_OBJECT = 1        :
+:      |                                       :
+:      +-- END_OF_GROUP = 1 in the subgroup    :
+:                                   header     :
+~~~
+{: #figure-subgroups title="Example of a Track with Objects layered into two Subgroups" }
+
```
suhasHere, line 617: not saying we need to use, but we have one more rendering we used in secure objects ( sans subgroups)

https://www.ietf.org/archive/id/draft-ietf-moq-secure-objects-01.html#section-2

vasilvv: This one is mostly to demonstrate multiple subgroups, so I think it's a bit different.

suhasHere: agreed, but I was wondering if we need a version that combines both ?

draft-ietf-moq-transport.md
```diff
@@ -579,6 +579,42 @@ stream is the first object ever published in that subgroup. A relay forwarding a
...
+:                                              :
+:              +-------+-------+-------+       :  +-------+-------------+
+: Subgroup 0   | {7,0} | {7,2} | {7,4} |       :  | {8,0} |     ...     |
+:      ^       +-------+-------+-------+       :  +-------+-------------+
+:      |           ^                           :
+:      |           |                           :
+:      |           +-- FIRST_OBJECT = 1        :
+:      |                                       :
+:      +-- END_OF_GROUP = 1 in the subgroup    :
+:                                   header     :
```
afrind, line 614: Including the subgroup bit encodings in the object model section runs counter to our new layout.  We could make this more abstract and keep it here, or move it to an appendix?

ianswett: I think this would be useful in an appendix.

PR #1887, opened 2026-09-01
Extract Track Namespace wire format to a common section

afrind:
Move the Track Namespace encoding, its field descriptions, and the size limits out of Track Naming and into a new Track Namespace Structure section under Notational Conventions and Common Structures. Track Naming keeps the model-level definition and cross-references the new section.

Retarget the Track Namespace, Track Namespace Prefix, and Track Namespace Suffix references in the control message sections at the new structure section. Track Name references continue to point at Track Naming, since each is paired with a namespace field in the same message and the wire section does not define a Track Name encoding.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3380,14 +3380,12 @@ SUBSCRIBE_NAMESPACE Message {
 * Request ID: See {{request-id}}.
 
 * Track Namespace Prefix: A Track Namespace structure as described in
-  {{track-namespace-structure}} with between 0 and 32 Track Namespace Fields.
-  This prefix is matched against track namespaces known to the publisher.  For
-  example, using the serialized format from {{namespace-name-format}}, if the
-  publisher is a relay that has received PUBLISH_NAMESPACE messages for
-  namespaces `example.2ecom-123-100` and `example.2ecom-123-200`, a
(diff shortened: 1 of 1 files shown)


ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -3489,10 +3486,11 @@ SUBSCRIBE_TRACKS Message {
...
 * Track Namespace Prefix: A Track Namespace structure as described in
-  {{track-name}} with between 0 and 32 Track Namespace Fields.  This prefix is
-  matched against track namespaces known to the publisher.  If an endpoint
-  receives a Track Namespace Prefix consisting of greater than 32 Track
-  Namespace Fields, it MUST close the session with a `PROTOCOL_VIOLATION`.
+  {{track-namespace-structure}} with between 0 and 32 Track Namespace Fields.
+  This prefix is matched against track namespaces known to the publisher.  If
+  an endpoint receives a Track Namespace Prefix consisting of greater than 32
+  Track Namespace Fields, it MUST close the session with a
+  `PROTOCOL_VIOLATION`.
```
suhasHere: i see the mention of 0 and 32 is repeated. may be push it all in one place

suhasHere (approved):

PR #1888, opened 2026-09-01
Move Location Filter wire encoding to its parameter section

afrind:
Move the LOCATION_FILTER diagram, the field presence rules keyed off Length, the relative StartGroup computation, the clamping rule, the EndGroupDelta delta encoding, and the end field omission rules out of Location Filters and into the LOCATION FILTER Parameter section. Location Filters keeps the semantics and cross-references the encoding.

Define Next Object and Next Group in the Largest Object section. Both terms are used throughout the semantics sections but were previously defined only in the Location Filter encoding text, which now refers to the definitions rather than restating them.

Retarget the FETCH and FILL PARAMETERS references at the parameter section. References from the semantics sections continue to point at Location Filters.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -751,6 +751,8 @@ The `Next Object` is the Location immediately following `Largest Object`, which
 is `{Largest Object.Group, Largest Object.Object + 1}`, or {0, 0} if no content
 has been delivered yet. The `Next Group` is the first Location of the Group
 following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
+`Next Object` and `Next Group` are Locations that do not necessarily
+refer to an Object that exists.
 
 ### Fill Semantics {#fill-semantics}
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -893,11 +893,12 @@ groups. A publisher that does will begin the next group as soon as practical.
 
 ### Location Filters {#location-filters}
 
-A Location filter specifies an inclusive range of Locations.  Only objects
+A Location filter specifies an inclusive range of Locations or an inclusive start
+Location and an open-ended end location.  Only objects
 with Locations within the inclusive range pass the filter.  A Location filter
 is encoded as specified in {{location-filter}}.
 
-Subscribers can specify a Location filter on a subscription indicating to the
+Subscribers can specify a Location filter on a subscription to indicate to the
 publisher which Objects to send.  Subscriptions without a filter pass all
 Objects published or received via upstream subscriptions.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -900,53 +895,12 @@ An object published or received in a subgroup or datagram is
 Some Location filters are defined to be relative to the `Largest Object`
```
suhasHere, line 909:
```suggestion
Some Location filters ((see {{location-filter}}) are defined to be relative to the `Largest Object`
```

suhasHere (approved):

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -744,6 +747,11 @@ processing the message. Largest Object updates when the first byte of an Object
 with a Location larger than the previous value is published or received through
 a subscription.
 
+The `Next Object` is the Location immediately following `Largest Object`, which
+is `{Largest Object.Group, Largest Object.Object + 1}`, or {0, 0} if no content
+has been delivered yet. The `Next Group` is the first Location of the Group
+following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
```
afrind, line 753:
```suggestion
following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
`Next Object` and `Next Group` are Locations that do not necessarily
refer to an Object that exists.
```

draft-ietf-moq-transport.md
```diff
@@ -883,83 +893,35 @@ groups. A publisher that does will begin the next group as soon as practical.
...
 A Location filter specifies an inclusive range of Locations.  Only objects
-with Locations within the inclusive range pass the filter.
+with Locations within the inclusive range pass the filter.  A Location filter
+is encoded as specified in {{location-filter}}.
 
-An object published or received in a subgroup or datagram is
-**subscription-delivered**.  Objects delivered via a fill fetch stream (see
-{{fill-semantics}}) are **fill-delivered**.
+Subscribers can specify a Location filter on a subscription indicating to the
+publisher which Objects to send.  Subscriptions without a filter pass all
```
ianswett: This sentence is a bit awkward, maybe this suggestion is better?

```suggestion
Subscribers can specify a Location filter on a subscription to indicate to the
publisher which Objects to send.  Subscriptions without a filter pass all
```

draft-ietf-moq-transport.md
```diff
@@ -883,83 +893,35 @@ groups. A publisher that does will begin the next group as soon as practical.
...
 ### Location Filters {#location-filters}
 
-Subscribers can specify a Location filter on a subscription indicating to the publisher
-which Objects to send.  Subscriptions without a filter pass all Objects
-published or received via upstream subscriptions.
-
-Fetch requests can also specify a Location filter.  Fetch requests without a filter
-include all Locations from {0, 0} up to `Largest Object` ({{largest-object}}).
-
 A Location filter specifies an inclusive range of Locations.  Only objects
```
```suggestion
A Location filter specifies an inclusive range of Locations or an inclusive start
Location and an open-ended end location.  Only objects
```

ianswett (approved):

PR #1894, opened 2026-09-01
Add Filter Sets subsection with SetID examples

suhasHere:
Extracts the SetID evaluation rules into a dedicated Filter Sets subsection under Range Filters and adds worked examples of AND, OR, mixed conjunctions, and relay aggregation.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -790,11 +790,9 @@ encoding specified in {{range-filter-structure}}.
 An object matches the filter if its value falls within any Range (i.e., Ranges
 are OR'd within a filter parameter).
 
-Each Range Filter parameter carries a SetID, which identifies the set of
-filters it belongs to.  Filter parameters with the same SetID are AND'd;
-distinct SetIDs are OR'd.  The final result is SetID=0 OR SetID=1 OR
-... SetID=255, where each SetID=i is the AND of all filter parameters
-carrying that SetID.
+Each Range Filter parameter carries a SetID, which identifies the filter
+set it belongs to.  Filter sets combine parameters into boolean
+expressions; see {{filter-sets}}.
 
 The Track Property filter parameter MAY appear multiple times in a
 SUBSCRIBE_TRACKS message or REQUEST_UPDATE for it.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -790,11 +790,9 @@ encoding specified in {{range-filter-structure}}.
...
 are OR'd within a filter parameter).
 
-Each Range Filter parameter carries a SetID, which identifies the set of
-filters it belongs to.  Filter parameters with the same SetID are AND'd;
-distinct SetIDs are OR'd.  The final result is SetID=0 OR SetID=1 OR
-... SetID=255, where each SetID=i is the AND of all filter parameters
-carrying that SetID.
+Each Range Filter parameter carries a SetID, which identifies the filter
+set it belongs to.  Filter sets combine parameters into boolean
+expressions; see {{filter-sets}}.
```
ianswett, line 795: I don't think we need to talk about SetID here.  Here's a suggestion on something more functionally oriented, but feel free to write something different.

```suggestion
One or more Range filters can be grouped into a FilterSet, and
an Object passes the Range filters if they pass through
any FilterSet.  For more details on how FilterSets filter Objects
and how they can be updated, see {{filter-sets}}.
```

draft-ietf-moq-transport.md
```diff
@@ -823,6 +821,70 @@ The Object Property Filter can be used to filter Objects with required
 Object Property types and values.  It only filters Object Properties,
 and does not evaluate Track Properties in PUBLISH messages.
 
+### Filter Sets {#filter-sets}
+
+A Range Filter can express boolean combinations across Track or Object
+properties and Object header fields using filter sets.  Each Range
```
ianswett, line 827: I'm wondering if we should use FilterSet in this section, since you mean a specific thing in this context?

ianswett, line 827: I think SetID can't span Track Property Filters and Object Filters, correct?

```suggestion
FilterSets allow Range Filters to match boolean combinations of Track
properties or combinations of Object properties and Object header
fields using filter sets.  SetIDs are not shared between Object Property
Filters and Range Filters on Objects. Each Range
```

draft-ietf-moq-transport.md
```diff
@@ -823,6 +821,70 @@ The Object Property Filter can be used to filter Objects with required
...
+* Filter parameters carrying the same SetID are combined with logical AND.
+* Filter sets (parameters carrying different SetIDs) are combined with
+  logical OR.
+
+The overall match condition for a request is therefore an OR of ANDs:
+`(set 0) OR (set 1) OR ... OR (set 255)`, where each `set i` is the AND
+of every filter parameter tagged with SetID=i.  An Object matches the
+request if any one filter set matches.
+
+A single SetID suffices when all conditions must hold; multiple SetIDs
```
ianswett, line 843: I'd move this above "The evaluation rules are:" because I think it's a nice clarification.

draft-ietf-moq-transport.md
```diff
@@ -823,6 +821,70 @@ The Object Property Filter can be used to filter Objects with required
...
+express an OR of distinct conjunctions that OR'd Ranges on a single
+field cannot.
+
+Examples use the wire encoding from {{range-filter-structure}} and
```
ianswett, line 847: Examples might be useful, but I'm inclined to skip them for now.

I would like a subsection on updating Range Filters

ianswett (commented):
Thanks, this is definitely helpful.  I left some suggestions and I'd still like to see a subsection on updating Range Filters.

ie:
1) Add a second set
2) Remove one of two sets
3) Update a Property Filter within a set

PR #1895, opened 2026-09-02
Add FETCH semantics section

afrind:
Hoisted non-wire format concepts from FETCH Message, FETCH_OK Message, FILL_TIMEOUT and Fetch Streams into a 3.2 Fetch.  Reordered the sentences so they flow logically, and added interior sections:

Fetch Object Delivery
Gaps in a Fetch Stream
Relay Fetch Handling

Editorial wordsmithing throughout, with a notable new paragraph explaining how to figure out descending fetch gaps -- previously I'd been too chicken to attempt this.

I removed a stale MUST regarding relay behavior for unknown state objects that contradicted FILL_TIMEOUT and UNKNOWN elsewhere.

No other normative language was changed.


draft-ietf-moq-transport.md
```diff
@@ -716,12 +716,97 @@ following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
...
+Unless the request included Range Filters (see {{range-filters}}), any gaps in
+the Group and Object IDs in the response stream indicate objects that do not
+exist (see {{object-states}}).  When one or more Range Filters are present, gaps
+indicate objects in the unknown state.
+
+For Ascending Group Order, and Descending Group Order where Start and End
+Location are in the same group, gaps include ranges between the first requested
+object and the first object in the stream; between objects in the stream; and
+between the last object in the stream and the End Location indicated in
+FETCH_OK.
```
afrind, line 769:
```suggestion
For Ascending Group Order, and Descending Group Order where Start and End
Location are in the same group, gaps include ranges between the Start
Location and the first object in the stream; between objects in the stream; and
between the last object in the stream and the End Location indicated in
FETCH_OK.
```

draft-ietf-moq-transport.md
```diff
@@ -716,12 +716,97 @@ following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
...
+
+For Descending Group Order where Start and End Location are in different groups,
+the first expected Object in the stream is `{End Location.Group, 0}` and the
+last expected Object in the stream is `{Start Location.Group, 2^64-1}`.  When
+two consecutive objects A and B are in different groups it represents multiple
+logical gaps: the tail of `A.Group`; all skipped groups; and `{B.Group, 0}`
+through B, exclusive.
```
```suggestion
two consecutive objects A and B are in different groups, a gap represents
multiple logical gaps: the tail of `A.Group`; all skipped groups; and `{B.Group,
0}` through B, exclusive.
```

draft-ietf-moq-transport.md
```diff
@@ -716,12 +716,97 @@ following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
 `Next Object` and `Next Group` are Locations that do not necessarily
 refer to an Object that exists.
 
-## Fetch
+## Fetch {#fetch}
 
-### Fetch State Management
+A FETCH requests pre-existing Objects from a Track between a Start Location and
+an End Location, inclusive.  These are set from a Location Filter (see
```
```suggestion
an End Location, inclusive.  The range is specified by a Location Filter (see
```

draft-ietf-moq-transport.md
```diff
@@ -716,12 +716,97 @@ following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
...
+{{location-filters}}) when present, or default to {0, 0} and `Largest Object`
+({{largest-object}}) respectively.
+
+Objects with Locations larger than the `Largest Object` at the time the request
+is processed will not be retrieved by a FETCH.  The actual end of the FETCH
+response is indicated in the FETCH_OK End Location (see {{message-fetch-ok}}).
```
ianswett, line 728: Why do we have End Location if FETCH returns Largest Object?  We don't have an option for the publisher to truncate earlier, do we?  Having both seems like it's bound to cause confusion about what the actual end is.

In cases the Fetch can be fully satisfied by the cache and there's no need to return Largest Object, then presumably the publisher can satisfy the full range.

I filed #1897

draft-ietf-moq-transport.md
```diff
@@ -716,12 +716,97 @@ following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
...
+FETCH_OK message is sent, but the FETCH_OK MUST NOT be sent until the End
+Location is known.
+
+If no Objects have been published for the track or Start Location is greater
+than the `Largest Object` ({{largest-object}}) the publisher MUST return
+REQUEST_ERROR with error code `INVALID_RANGE`.
+
+### Fetch Object Delivery
+
+The Object Delivery Mode does not apply to fetches.
```
```suggestion
Fetch delivers Objects ordered by their Location and the Object Delivery Mode
does not apply.
```

draft-ietf-moq-transport.md
```diff
@@ -716,12 +716,97 @@ following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
...
+The publisher creates a single unidirectional stream (see {{fetch-streams}})
+that is used to send the Objects.  If no Objects exist in the requested range,
+the publisher opens the unidirectional stream, sends the FETCH_HEADER (see
+{{fetch-header}}) and closes the stream with a FIN.
+
+A publisher MUST send fetched groups in the requested group order (see
+{{group-order}}). Within each group, objects are sent in Object ID order;
+subgroup ID is not used for ordering.
+
+The publisher is responsible for delivering all available Objects in the
```
ianswett: Especially with the caveat of Filters and FILL_TIMEOUT, this is a bit nuanced.  Is this trying to say something important or can it be removed?

draft-ietf-moq-transport.md
```diff
@@ -716,12 +716,97 @@ following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
...
+requested range.
+
+### Gaps in a Fetch Stream
+
+Unless the request included Range Filters (see {{range-filters}}), any gaps in
+the Group and Object IDs in the response stream indicate objects that do not
+exist (see {{object-states}}).  When one or more Range Filters are present, gaps
```
ianswett: What about FILL_TIMEOUT?

draft-ietf-moq-transport.md
```diff
@@ -716,12 +716,97 @@ following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
...
+indicate objects in the unknown state.
+
+For Ascending Group Order, and Descending Group Order where Start and End
```
ianswett, line 765: This paragraph and the next are complicated, can we say something simpler?

We're basically saying "Treat any gaps between the start and the first Object as a gap in the same way you would a gap between two Objects.  Same for the last Object and the end Location."

draft-ietf-moq-transport.md
```diff
@@ -716,12 +716,97 @@ following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
...
+the first expected Object in the stream is `{End Location.Group, 0}` and the
+last expected Object in the stream is `{Start Location.Group, 2^64-1}`.  When
+two consecutive objects A and B are in different groups, a gap represents
+multiple logical gaps: the tail of `A.Group`; all skipped groups; and `{B.Group,
+0}` through B, exclusive.
+
+Gaps at the end of a stream are only detectable when the stream is terminated
+by a FIN.
+
+Rather than leaving a gap, a publisher can mark a range of Objects it will not
```
ianswett: This paragraph feels like it could be more proscriptive, because there's not really an option here, correct?  If you have a timed out gap, you need to use the End of Range indicator?

"When serializing a gap because an Object has status unknown or timed out, an explicit End Of Range indicator () is serialized...

draft-ietf-moq-transport.md
```diff
@@ -716,12 +716,97 @@ following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
...
+serialize with an End of Range indicator (see {{end-of-range}}), identifying
```
ianswett: End of Range indicator is a weird name.  Object Range Gap?

draft-ietf-moq-transport.md
```diff
@@ -716,12 +716,97 @@ following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
...
+those Objects as non-existent, unknown, or timed out.  A publisher SHOULD NOT
+use `End of Non-Existent Range` in a FETCH response except to split a range of
```
ianswett: Does this change with Range Filters, which leave all unspecified Objects in the Unknown state?

ianswett (commented):
It's great to move this text, some suggestions/comments, probably mostly on existing text.

draft-ietf-moq-transport.md
```diff
@@ -716,12 +716,97 @@ following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
...
+Location and the first object in the stream; between objects in the stream; and
+between the last object in the stream and the End Location indicated in
+FETCH_OK.
+
+For Descending Group Order where Start and End Location are in different groups,
+the first expected Object in the stream is `{End Location.Group, 0}` and the
+last expected Object in the stream is `{Start Location.Group, 2^64-1}`.  When
+two consecutive objects A and B are in different groups, a gap represents
+multiple logical gaps: the tail of `A.Group`; all skipped groups; and `{B.Group,
+0}` through B, exclusive.
```
sandarsh: The writeup reads correctly. However, should we write something generic? Why target only these two special cases for gaps, i.e. asc or desc in the same group; and cross-group desc?

draft-ietf-moq-transport.md
afrind: There's one case where it's useful.  If I Fetch groups 0-2000, but the track ends at 1000, I'll see 1000 <FIN>, which will cause me to mark 1000-2000 as a known non-existent range, unless 1000 contains and End Of Track object.  If I later fetch 2000-3000, it has to go upstream.

draft-ietf-moq-transport.md
afrind: I think this will read better at the end of this paragraph below

> A publisher MUST send fetched groups in the requested group order (see
{{group-order}}). Within each group, objects are sent in Object ID order;

draft-ietf-moq-transport.md
afrind: I think we can delete this sentence now.  The below section makes it clear what happens if the publisher leaves a gap -- it is interpreted with specific meaning (Non existent, timeout, or unknown).

draft-ietf-moq-transport.md
afrind: Yes I think we need to combine what's here with the one below: 

> Rather than leaving a gap, a publisher can mark a range of Objects it will not...

draft-ietf-moq-transport.md
afrind: Let's change in a follow up since it impacts the wire format description of this.

afrind: Range State Indicator?

draft-ietf-moq-transport.md
afrind: What happens in this case:

Filtered FETCH for ABCDEF

A, C, E - do no exist
B is filtered out
D exceeds FILL_TIMEOUT
F is in cache

What should the publisher say?

NE U NE T NE F?

Or

<Gap A/B> U (C) T (D) F

draft-ietf-moq-transport.md
afrind: The text is derivable from first principles but I like making it explicit -- if you asked for 10,0 and the first object you see is 10,3, there's a gap there, not just _between_ two objects.

The Descending paragraph is also hard-fought non-obvious knowledge.

draft-ietf-moq-transport.md
```diff
@@ -4652,10 +4678,7 @@ When Serialization Flags indicates an End of Range (e.g. values 0x8C, 0x10C, or
 0x20C), the Group ID and Object ID fields are present.  Subgroup ID, Priority and
 Properties are not present. All Objects with Locations between the last
 serialized Object, if any, and this Location, inclusive, either do not exist
-(when Serialization Flags is 0x8C), are unknown (0x10C), or timed out (0x20C).  A
-publisher SHOULD NOT use `End of Non-Existent Range` in a FETCH response except to
-split a range of Objects that will not be serialized into those that are known
-not to exist and those with unknown or timed out status.
```
suhasHere, line 4658: not sure if we loose this  or explained differently

afrind: Line 784:

> A publisher SHOULD NOT use `End of Non-Existent Range` in a FETCH response
except to split a range of Objects that will not be serialized into those that are
known not to exist and those with unknown or timed out status.

draft-ietf-moq-transport.md
```diff
@@ -716,12 +716,98 @@ following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
...
+
+If a Publisher receives a FETCH with a range that includes one or more Objects
+with unknown state, it can choose to reset the FETCH data stream with
+UNKNOWN_OBJECT_STATUS ({{stream-reset-codes}}), or indicate the range of unknown
+Objects (`End of Unknown Range`, see {{end-of-range}}) and continue serving other
+known Objects.
+
+### Relay Fetch Handling
+
+A relay that has cached objects from the beginning of the range MAY start
```
ianswett, line 798: Isn't this a dupe of Line 732?

draft-ietf-moq-transport.md
```diff
@@ -716,12 +716,98 @@ following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
...
+sending objects immediately in response to a FETCH.  If the upstream FETCH
```
```suggestion
sending objects immediately in response to a FETCH.  If an upstream FETCH
```

ianswett (approved):

draft-ietf-moq-transport.md
afrind: It's not exactly the same.  One is generic for "The publisher" and this is for "A relay that has cached objects".   Since they both have normative MAY, I'm going to leave it for now.

draft-ietf-moq-transport.md
afrind: There's actually another possibility -- it can send FETCH OK and use UNKNOWN to cover any uncached objects.

draft-ietf-moq-transport.md
```diff
@@ -716,12 +716,100 @@ following `Largest Object`, which is `{Largest Object.Group + 1, 0}`.
...
+except to split a range of Objects that will not be serialized into those that are
+known not to exist and those with unknown or timed out status.
+
+If a Publisher receives a FETCH with a range that includes one or more Objects
+with unknown state, it can choose to reset the FETCH data stream with
+UNKNOWN_OBJECT_STATUS ({{stream-reset-codes}}), or indicate the range of unknown
+Objects (`End of Unknown Range`, see {{end-of-range}}) and continue serving
+other known Objects.  When resetting the stream, the publisher can do so
+immediately or after previously sent objects are delivered.  If it has not yet
+sent FETCH_OK, it can also send REQUEST_ERROR.
```
```suggestion
sent FETCH_OK, it can send REQUEST_ERROR instead.
```

ianswett (approved):

draft-ietf-moq-transport.md
ianswett: I found the 'also' a bit confusing in this context.

```suggestion
sent FETCH_OK, it can send REQUEST_ERROR in addition to resetting the stream.
```

PR #1902, opened 2026-09-03
Create .gitattributes

ianswett:
Set the line break formatting to auto to avoid huge diffs

Based on the docs here: https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto


afrind (approved):
I don't really know what this does.  I think the underlying problem is the doc may have some bad lines.

ianswett:
Submitting this didn't fix my existing PR, but we'll see if it helps going forward?

PR #1904, opened 2026-09-04
add warning about fancy setup options (Editorial)

martinduke:
The current text implies you can send requests in parallel with a SETUP with arbitrary options. It's important to have a pointer to the section that adds restrictions about this.

Also, it now just refers to "requests" instead of naming certain request types at random.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1481,7 +1481,7 @@ close the Session with a `PROTOCOL_VIOLATION`. Objects are sent on unidirectiona
 streams.
 
 As such, a client can initiate a MOQT session and requests all in parallel.
-However, there are restrictions on these requests as described in {{setup-message}}.
+However, there are restrictions on these requests as described in {{message-setup}}.
 When this is done before the handshake completes using 0-RTT, the security
 implications described in {{zero-rtt}} apply.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1480,10 +1480,10 @@ begin with any other message type unless negotiated. If they do, the peer MUST
 close the Session with a `PROTOCOL_VIOLATION`. Objects are sent on unidirectional
 streams.
 
-As such, a client can initiate a MOQT session, subscribe, and
-start publishing Objects all in parallel. When this is done before the
-handshake completes using 0-RTT, the security implications described in
-{{zero-rtt}} apply.
+As such, a client can initiate a MOQT session and requests all in parallel.
+However, there are restrictions on these requests as described in {{setup-message}}.
+When this is done before the handshake completes using 0-RTT, the security
+implications described in {{zero-rtt}} apply.
 
 Unidirectional streams containing Objects or bidirectional stream(s) beginning
 with a request message could arrive prior to the control streams, in which case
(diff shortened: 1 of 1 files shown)


martinduke:
Deleted a later paragraph which says related stuff about parameters, but just referring to the SETUP section prevents these from getting out of sync, and points to a broader thing about semantics rather than parameters specifically. For example, I can't send the FOO request that is enabled by a setup option.

draft-ietf-moq-transport.md
```diff
@@ -1480,10 +1480,10 @@ begin with any other message type unless negotiated. If they do, the peer MUST
...
 streams.
 
-As such, a client can initiate a MOQT session, subscribe, and
-start publishing Objects all in parallel. When this is done before the
-handshake completes using 0-RTT, the security implications described in
-{{zero-rtt}} apply.
+As such, a client can initiate a MOQT session and requests all in parallel.
+However, there are restrictions on these requests as described in {{message-setup}}.
+When this is done before the handshake completes using 0-RTT, the security
+implications described in {{zero-rtt}} apply.
```
suhasHere, line 1487: This should be in security consideration section i feel .

draft-ietf-moq-transport.md
```diff
@@ -1496,13 +1496,6 @@ A control stream MUST NOT be closed at the underlying transport layer during the
 session's lifetime.  Doing so results in the session being closed as a
 `PROTOCOL_VIOLATION`.
 
-Prior to receiving the peer's SETUP message, it's unknown what extensions
-a peer will support. Message Parameters requiring negotiation SHOULD NOT
-be used prior to receiving the peer's SETUP message unless the application
-requires the extension or the endpoint knows the peer supports the
-extension. If an unsupported Message Parameter is used, the peer will be
-unable to process it and the session will be terminated. See {{message-params}}.
-
```
suhasHere, line 1505: This still holds true right ?

draft-ietf-moq-transport.md
martinduke: The point of this text is to provide pointers to other existing text in case an implementer is just reading this section.

The stuff in {{message-setup}} is not a security consideration, it's necessary for the protocol to be parsed properly.

If you want {{zero-rtt}} to move to SC that's orthogonal to this PR.

draft-ietf-moq-transport.md
martinduke: Yes, but it repeats the text in {{setup-message}} and will become unaligned if we change things.

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -1480,10 +1480,10 @@ begin with any other message type unless negotiated. If they do, the peer MUST
 close the Session with a `PROTOCOL_VIOLATION`. Objects are sent on unidirectional
 streams.
 
-As such, a client can initiate a MOQT session, subscribe, and
-start publishing Objects all in parallel. When this is done before the
-handshake completes using 0-RTT, the security implications described in
-{{zero-rtt}} apply.
+As such, a client can initiate a MOQT session and requests all in parallel.
```
afrind, line 1483:
```suggestion
As such, a client can initiate a MOQT session and send requests in parallel.
```

afrind: Was the removal of sending objects intentional/redundant?

draft-ietf-moq-transport.md
afrind: All I could find was this sentence:

> An endpoint that
is not offering extensions which modify control message semantics MAY pipeline
other control messages after SETUP without waiting for the peer's SETUP.

That's sort of different than this removal of SHOULD NOT and the consequences?  Did I miss another block?

PR #1588, opened 2026-04-02
Add internationalization statement for moqt URI scheme

ianswett:
🎯 **What:** Replaced the TODO comment regarding the RFC 7595 Section 3.6 internationalization statement with the standard IETF boilerplate text for URI schemes.

💡 **Why:** RFC 7595 dictates that new URI schemes must specify their internationalization and character encoding considerations (how they map to IRIs/RFC3987). This resolves the TODO and completes the `moqt` URI scheme registration template.

✅ **Verification:** Verified that the document builds properly using `make` (`xml2rfc` and `kramdown-rfc` pass without errors).

✨ **Result:** The `draft-ietf-moq-transport.md` specification now contains the proper internationalization rules for the `moqt` URI scheme.

---
*PR created automatically by Jules for task [5263480566833271922](https://jules.google.com/task/5263480566833271922) started by @ianswett*

Fixes: #1680

PR #1909, opened 2026-09-05
Move URI resolution to a separate draft

### LINKED ISSUE #1839: resolution of URI
The draft says 

> When SVCB-compatible records {{?RFC9460}} are published for the `authority`,
> a client MAY use them to learn the server's endpoints and supported ALPN
> protocols before connecting. 
> 

I don't think this is anywhere near specific enough.  The base draft needs to say how URI resolution works and how this ties to TLS certs. I can't deploy a load balancing  and HA strategy for relays unless I know what clients will do. The TLS does not provide much security unless we say how it relates to the URI.

fluffy:
Fixes #1839

An alternative way to do this would be to put the text in the MOQT draft. That approach is sketched out in https://github.com/moq-wg/moq-transport/pull/1901 but after looking at how much is needed, and the fact that the people we need to review this are PKIX and DNS people, I agree with Alan is better to put this in a separate draft as is done in this PR.

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -53,6 +53,7 @@ author:
 normative:
   QUIC: RFC9000
   WebTransport: I-D.ietf-webtrans-http3
+  URI-RESOLV: I-D.jennings-moq-discovery
 
 informative:
   CAT: I-D.ietf-moq-c4m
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1383,27 +1384,8 @@ Types" registry ({{iana-fragment-types}}).
 The default operation for dereferencing a `moqt` URI is to establish a
 MOQT session to the identified server.
 
-The `moqt` URI scheme has the following security considerations:
-
-- The `authority` component is sent in the TLS SNI extension during
-  connection establishment, exposing the target server identity to
-  on-path observers. Encrypted Client Hello (ECH) {{?RFC9580}} can
-  mitigate this exposure.
```
fluffy, line 1391: Note this text was not deleted but just moved to security consideration

suhasHere (approved):

ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -5128,6 +5110,12 @@ subscriber's right to access specific tracks or namespaces. Relays
 that aggregate subscriptions from multiple downstream subscribers MUST
 ensure each subscriber is independently authorized.
 
+The `authority` component of the MOQT URI is sent in the TLS SNI
+extension during connection establishment, exposing the target server
+identity to on-path observers. Encrypted Client Hello (ECH) {{?RFC9580}}
+can mitigate this exposure.
+
+
```
afrind, line 5118:
```suggestion

```

draft-ietf-moq-transport.md
```diff
@@ -1383,27 +1384,8 @@ Types" registry ({{iana-fragment-types}}).
...
-The `moqt` URI scheme has the following security considerations:
-
-- The `authority` component is sent in the TLS SNI extension during
-  connection establishment, exposing the target server identity to
-  on-path observers. Encrypted Client Hello (ECH) {{?RFC9580}} can
-  mitigate this exposure.
-
-- The `path-abempty` and `query` components are visible to the relay
-  that terminates the client's connection.
-
```
afrind, line 1395: This bit didn't make it to security considerations

afrind:
Did you incorporate #1588 in the new draft too?

draft-ietf-moq-transport.md
fluffy: well that's embarrassing ... will fix

PR #1911, opened 2026-09-06
Use Largest Object consistently

afrind:


ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -2958,11 +2958,11 @@ REQUEST_UPDATE Message {
 
 When a subscriber decreases the Start Location of the Location Filter
 (see {{location-filters}}), the Start Location can be smaller than the Track's
-Largest Location, similar to a new Subscription. Including FILL_PARAMETERS
+Largest Object, similar to a new Subscription. Including FILL_PARAMETERS
```
suhasHere, line 2961: Largest Location is fine ,since largest location embeds object too. 
I think we should leave it as-is

draft-ietf-moq-transport.md
```diff
@@ -2958,11 +2958,11 @@ REQUEST_UPDATE Message {
...
 (see {{fill-parameters}}) in the REQUEST_UPDATE causes the publisher to deliver
 the new fill range by opening a new fill fetch stream (see
 {{fill-semantics}}).  FETCH can also be used to retrieve any necessary Objects
-smaller than the current Largest Location.
+with Locations less than or equal to the current Largest Object.
```
suhasHere, line 2965: same comment, either we says "with Locations less than or equal to the location of the current Largest Object." OR leave as-is

suhasHere (changes requested):

draft-ietf-moq-transport.md
afrind: Largest Location is not defined in the document and Largest Object is.

draft-ietf-moq-transport.md
afrind: As-is doesn’t work.  An object cannot be compared to a location - smaller could mean payload length.

afrind: Your proposed text is unnecessarily longer because largest object is a location

vasilvv (approved):
I agree with Suhas that "Largest Location" is better than "Largest Object", but we should first fix existing inconsistencies, and then bikeshed the name.

suhasHere:
@afrind I was surprised that we decided to land this. I thought we could have made proper fix for location vs object when it comes to largest .

afrind:
> @afrind I was surprised that we decided to land this. I thought we could have made proper fix for location vs object when it comes to largest .

I'm sorry I merged this too quickly.  I should have checked directly with you first if you were ok with Victor's suggestion above.

PR #1912, opened 2026-09-06
Add REQUEST_ERROR shorthand names and apply them

### LINKED ISSUE #1722: REQUEST_ERROR name alias.
I like how the PUBLISH_OK is name we use to refer a REQUEST_OK that came back from an PUBLISH. I think this makes is much easier to talk about and explain things. 

I would like to see us do the same things with errors so that we used PUBLISH_ERROR as the alias for a REQUEST_ERROR that came back from PUBLISH request. And do the same for the other requests that generate this.

afrind:
Fixes: #1722

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -557,10 +557,10 @@ initiated and moved to the `Pending` state by either a publisher or a
 subscriber.  A publisher initiates a subscription to a track by
 sending the PUBLISH message.  The subscriber either accepts or rejects
 the subscription using PUBLISH_OK ({{message-request-ok}}) or
-REQUEST_ERROR.  A subscriber
+PUBLISH_ERROR.  A subscriber
 initiates a subscription to a track by sending the SUBSCRIBE message.
 The publisher either accepts or rejects the subscription using
-SUBSCRIBE_OK or REQUEST_ERROR.  Once either of these sequences is
+SUBSCRIBE_OK or SUBSCRIBE_ERROR.  Once either of these sequences is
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2853,6 +2854,11 @@ The REQUEST_ERROR message is sent in response to any request (SUBSCRIBE, FETCH,
 PUBLISH, SUBSCRIBE_NAMESPACE, SUBSCRIBE_TRACKS, PUBLISH_NAMESPACE, TRACK_STATUS,
 REQUEST_UPDATE).
 
+This document uses the shorthand SUBSCRIBE_ERROR, FETCH_ERROR, PUBLISH_ERROR,
```
ianswett: Is this the first time we say this?  Because it's quite a ways after we use these terms, which could be quite confusing.

ianswett (commented):
Looks good, except we describe how we use _ERROR messages very late in the draft, which seems like it could be very confusing?

afrind:
> Looks good, except we describe how we use _ERROR messages very late in the draft, which seems like it could be very confusing?

Yes, I noticed this. It’s somewhat a casualty of our refactor (I think it impacts eg publish_ok already). Any suggestions where to put it? Notational conventions is much lower now too.

ianswett (approved):

draft-ietf-moq-transport.md
afrind: Moved up

suhasHere (approved):

PR #1915, opened 2026-09-06
Preserve the SUBGROUP_HEADER PROPERTIES bit when forwarding

### LINKED ISSUE #1073: Omitting Extension Header Length is a subtle API change
If the subgroup ID is zero, or the first object ID, the relevant bits of the subgroup stream type can be determined on the fly for each hop.

However, to omit the extension header lengths from a stream, I have to know that all objects in the stream will not have extension headers. AFAICT consistent presence of extensions in a subgroup is not a requirement in the object model.

So either:
1) Relay MOQT framers ~never omit the extension header length in a subgroup;
2) There is a new sending API parameter "will never use extensions" when delivering an object for a new subgroup.

afrind:
The PROPERTIES bit applies to every Object on a Subgroup stream, so a relay picks its value when opening the stream downstream, before it has received the rest of the Subgroup. Clearing the bit and then receiving an Object with Properties leaves no legal encoding, since a relay is required to forward Properties.

Tell relays to carry the upstream value over, and to set the bit when they might add Properties of their own.

Fixes: #1073

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2068,7 +2068,7 @@ The PROPERTIES bit in the SUBGROUP_HEADER ({{subgroup-header}}) applies to
 every Object on the stream, so a relay selects its value when it opens the
 stream downstream, before it has received the rest of the Subgroup. A relay
 that does not add Object Properties of its own SHOULD use the value it
-received from upstream. A relay that might add Object Properties MUST set the
+received from upstream. A relay that might add Object Properties needs to set the
 bit to 1; Objects with no Object Properties then encode a Properties Length
 of 0.
 
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2064,6 +2064,14 @@ Object Properties as specified in {{properties}}.
 A relay MUST treat the object payload as opaque.  A relay MUST NOT
 combine, split, or otherwise modify object payloads.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2064,6 +2064,14 @@ Object Properties as specified in {{properties}}.
 A relay MUST treat the object payload as opaque.  A relay MUST NOT
 combine, split, or otherwise modify object payloads.
 
+The PROPERTIES bit in the SUBGROUP_HEADER ({{subgroup-header}}) applies to
+every Object on the stream, so a relay selects its value when it opens the
+stream downstream, before it has received the rest of the Subgroup. A relay
+that does not add Object Properties of its own SHOULD use the value it
+received from upstream. A relay that might add Object Properties MUST set the
```
ianswett: This MUST is sort of odd, because really what we're saying is that you won't be able to add them if you don't set it to 1.  It feels like a 'needs to'?
```suggestion
received from upstream. A relay that might add Object Properties needs to set the
```

ianswett (approved):

afrind:
Since we changed MUST to needs to, this is purely editorial now.

I thought about changing some of the usage of "stream" to "subgroup", but decided it was fine as it was.

afrind:
nvm there's still a new SHOULD - Ed&Min it is

ianswett (changes requested):
After further discussion, I think we'd be better off without this PR, because adding text on how to use the wire format correctly risks being confusing.

PR #1916, opened 2026-09-07
List the allowed parameters in each control message definition

### LINKED ISSUE #1253: Can all control messages clearly say what parameters are allowed
Specifically does FETCH_CANCEL allow authorization tokens ?

afrind:
Each parameter definition states which messages it can appear in, but there was no way to answer the reverse question from a message's own section.  Add a descriptive list of allowed parameters to each control message, splitting REQUEST_OK and REQUEST_UPDATE by the request type they answer or update, and drop the empty Allowed Parameters By Control Message section.

The lists are non-normative; the MAY stays in each parameter definition and the PROTOCOL_VIOLATION MUST stays in Parameter Scope.

Fixes: #1253

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2837,7 +2837,15 @@ REQUEST_OK Message {
 ~~~
 {: #moq-transport-request-ok format title="MOQT REQUEST_OK Message"}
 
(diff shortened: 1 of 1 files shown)


ianswett (approved):
I somehow don't like this as much as I expected to, but it's ok?

suhasHere (approved):

PR #1918, opened 2026-09-07
Clarify that Largest Object can still be arriving

### LINKED ISSUE #976: Clarify terminolgy around objects under delivery and the live edge
As discussed in the interim on 250428 there appear to be lacking a set of terms that correctly describe the following aspects around object delivery:
- Latest (in time) completly delivered object
- Object under delivery
- Next object to be delivered

To my understanding we might need to have seperate terms or qualifiers for if the above language is within one group or within one sub-group.
In addition is there a difference here for this if one consider object ID order? 

`

                                              Current point of delivery
                                                             I
                                                             V
I-- Latest completed delivered object --II-- currently delivered object --II-- next object --I

`

afrind:
Largest Object advances on the first byte of an Object, so it can name an Object that is not yet complete.  Say so where it is defined, state that a FETCH whose range includes it delivers that Object in full, and have the LARGEST_OBJECT parameter refer to the definition rather than describing the value a third way.  Also drop the single undefined use of the term live edge.

Fixes: #976

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -721,7 +721,8 @@ The `Largest Object` is the Object with the largest Location
 ({{location-structure}}) in the Track from the perspective of the publisher
 processing the message. Largest Object updates when the first byte of an Object
 with a Location larger than the previous value is published or received through
-a subscription.
+a subscription.  `Largest Object` therefore identifies an Object that can still
+be arriving.
 
 The `Next Object` is the Location immediately following `Largest Object`, which
 is `{Largest Object.Group, Largest Object.Object + 1}`, or {0, 0} if no content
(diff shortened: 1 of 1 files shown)


ianswett (commented):
I'm not sold on this being useful and it's more text, so I'm inclined towards closing the issue with no action.

afrind:
> I'm not sold on this being useful and it's more text, so I'm inclined towards closing the issue with no action.

I think the only potentially useful bit might be that for a FETCH - the responder is obligated to serve the whole object even if it's partial when FETCH_OK is sent.

suhasHere (approved):

afrind:
CC: @gloinul FYI this is closing your issue but reopen or open a new one if needed.

PR #1920, opened 2026-09-07
Say that NAMESPACE carries a single namespace

### LINKED ISSUE #1394: NAMESPACE messages with multiple sub-namespaces in them?
My understanding of the NAMESPACE message is as follows: Let's say that a subscriber sends a SUBSCRIBE_NAMESPACE request to a publisher, with the namespace set to "foo/bar". The publisher, if it is advertising a more restricted namespace, for example, "foo/bar/baz/goo", can send "baz/goo" in the NAMESPACE response.

What if, however, in my previous example, the publisher is advertising "foo/bar/baz/goo" and "foo/bar/baz/moo"? There is no way for it to send both "baz/goo" and "baz/moo" in the same NAMESPACE message. Are we allowing for multiple NAMESPACE messages?

afrind:
The body of NAMESPACE and NAMESPACE_DONE is a Track Namespace Suffix, which is itself a count-prefixed list of namespace fields, so it is not obvious that the count is fields within one namespace rather than a set of namespaces.  State that each message carries one namespace and that the publisher sends one per match.

Fixes: #1394

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -3416,7 +3416,8 @@ The NAMESPACE message is similar to the PUBLISH_NAMESPACE message, except
 it is sent on the response stream of a SUBSCRIBE_NAMESPACE request.
 All NAMESPACE messages are in response to a SUBSCRIBE_NAMESPACE, so only
 the namespace tuples after the 'Track Namespace Prefix' are included
-in the 'Track Namespace Suffix'.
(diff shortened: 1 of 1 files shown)


ianswett (commented):
I don't think we really need this?  I'd be inclined to close the issue with no action.

afrind:
Closed the issue with no action

PR #1922, opened 2026-09-07
Say how a non-Normal Object encodes Properties in a Subgroup

### LINKED ISSUE #1908: Can an Object in Subgroup streams have Status and Properties?
In 11.2.1 "Object Datagram" it says

If an Object Datagram includes both the STATUS bit and PROPERTIES bit, and the Object Status is not Normal (0x0), the endpoint MUST close the session with a PROTOCOL_VIOLATION, because only Normal Objects can have Properties.

but there is no equivalent text  in 11.3 Subgroup Streams.

I assume this is an oversight?

afrind:
Object Properties already says an Object with a status other than Normal cannot carry Properties, but the SUBGROUP_HEADER PROPERTIES bit applies to every Object on the stream, so such an Object still has the field and needs a length. Say it is 0, which also removes the asymmetry with Object Datagram that made the rule look like it was missing for Subgroups.

Fixes: #1908

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -4419,7 +4419,7 @@ fields are present in the header:
   in all Objects in this Subgroup. When set to 1, the Object Properties structure
   defined in {{object-properties}} is present in all Objects; Objects with no
   properties or non-Normal status set Properties Length to 0. When set to 0, the
-  field is never present.
+  field is never present in this Subgroup.
 
 * The **SUBGROUP_ID_MODE** field (bits 1-2, mask 0x06) is a two-bit field that
   determines the encoding of the Subgroup ID. To extract this value, perform a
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -4424,7 +4418,8 @@ fields are present in the header:
 * The **PROPERTIES** bit (0x01) indicates when the Properties field is present
   in all Objects in this Subgroup. When set to 1, the Object Properties structure
   defined in {{object-properties}} is present in all Objects; Objects with no
-  properties set Properties Length to 0. When set to 0, the field is never present.
+  properties or non-Normal status set Properties Length to 0. When set to 0, the
+  field is never present.
```
ianswett: This may be obvious, but I think it reads a bit more clearly.

```suggestion
  field is never present in this Subgroup.
```

ianswett (approved):

PR #1923, opened 2026-09-07
Clarify who a subscriber can send requests to

### LINKED ISSUE #1701: Clarify: A subscriber MAY send a SUBSCRIBE or FETCH for a track to any publisher
I think we have agreed in discussing all the make before break stuff, that if two publishers publish same namespace, the relay needs to upstream the subscribe to both. 

The spec says 

> A subscriber MAY send a SUBSCRIBE or FETCH for a track to any
> publisher. 

But I think it should say 

> A subscriber MUST send a SUBSCRIBE for a track to all the
> publisher. 

I don't know that we discussed what to do with FETCH so that is a separate issue of what to do with it but it seems to me like it needs to be the same.

afrind:
The sentence about sending to any publisher was read as a rule about how many publishers a request goes to, rather than about which publishers a subscriber is permitted to contact. Say what the permission is in contrast to, add TRACK_STATUS, and note that a Relay is limited to matching publishers, pointing at the rules for selecting among them.

This is qualifying an existing MAY and adding TRACK_STATUS.  I think this was the underlying intent so I marked Editorial but could mark Ed&Min Design too.

Fixes: #1701

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1193,10 +1193,15 @@ subscribers, it is not a full-fledged routing protocol and does not protect
 against loops and other phenomena. In particular, PUBLISH_NAMESPACE SHOULD NOT
 be used to find paths through richly connected networks of relays.
 
-A subscriber MAY send a SUBSCRIBE or FETCH for a track to any publisher. If it
+An End Subscriber is not restricted in which publisher it contacts: it
+MAY send a SUBSCRIBE, FETCH or TRACK_STATUS for a track to any publisher,
+including one from which it has not received a PUBLISH_NAMESPACE. If it
 has accepted a PUBLISH_NAMESPACE with a namespace that exactly matches the
 namespace for that track, it SHOULD only request it from the senders of those
 PUBLISH_NAMESPACE messages.
+
+A Relay is more restricted: it sends these requests only to matching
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1193,11 +1193,16 @@ subscribers, it is not a full-fledged routing protocol and does not protect
 against loops and other phenomena. In particular, PUBLISH_NAMESPACE SHOULD NOT
 be used to find paths through richly connected networks of relays.
 
-A subscriber MAY send a SUBSCRIBE or FETCH for a track to any publisher. If it
+An End Subscriber is not restricted in which publisher it contacts: it
```
ianswett, line 1196: Isn't an End Subscriber typically only connected to one publisher?

afrind: Yeah End Subscriber might be too restrictive?  I was trying to draw a contrast with "A relay..." below.  I can put it back to "A subscriber" or "A non-relay subscriber"?

ianswett:
```suggestion
A non-relay Subscriber that it connected to multiple Publishers
```

afrind: The non-relay subscriber can also connect to different publishers -- your suggestion presupposes it's already connected?

PR #1924, opened 2026-09-07
Shorten Parsing Serialized Names

### LINKED ISSUE #1669: Shorter text on Parsing Serialized Names
trivial editorial nit but ... I think section 1.5.1 could be replaced with, uh nothing, or MUST exactly match syntax in section 1.5 or "MUST match exact syntax in section 1.5 including correct case" and add failing this causes protocol error.

afrind:
The parsing rules restated the encoding rules the section already gives: that literal bytes are output verbatim and everything else becomes a period followed by exactly two lowercase hexadecimal digits. Replace them with the property they exist to preserve, and fold the result into the section that defines the format.

Fixes: #1669

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2370,8 +2370,9 @@ serialized names can be compared without deserializing them. To maintain this
 property, an implementation parsing this format MUST reject a name that does
 not follow the encoding rules exactly, including a period not followed by
 exactly two lowercase hexadecimal digits, or a byte that could have been
-represented literally but was hex-encoded. How an invalid name is handled is
-application-defined.
+represented literally but was hex-encoded.  For example, `.61` is invalid
+because `a` is represented as the literal character `a`. How an invalid name is
+handled is application-defined.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


ianswett (approved):

draft-ietf-moq-transport.md
```diff
@@ -2364,29 +2364,14 @@ is chosen to be generally both filename and URL safe, filename safety is
...
-Implementations that receive an invalid serialized name SHOULD treat it as
-an error. The specific error handling behavior is application-defined.
+Because this format produces exactly one rendering of any given binary value, it
+is bijective: every valid serialized name maps to exactly one binary value, so
+serialized names can be compared without deserializing them. To maintain this
+property, an implementation parsing this format MUST reject a name that does
+not follow the encoding rules exactly, including a period not followed by
+exactly two lowercase hexadecimal digits, or a byte that could have been
+represented literally but was hex-encoded. How an invalid name is handled is
+application-defined.
```
```suggestion
represented literally but was hex-encoded.  For example, `.61` is invalid
because `a` is represented as the literal character `a`. How an invalid name is
handled is application-defined.
```

suhasHere (approved):

vasilvv (approved):

PR #1925, opened 2026-09-07
Define Subscription in Terms and Definitions

### LINKED ISSUE #1226: Use of the word "subscription" in abstract and other sections than 5.1
As discussed in PR #1222, the intro and other sections in the draft might need updating regarding the use of the word "subscription", which should mean "a flow of objects initiated via the SUBSCRIBE or PUBLISH messages".

For example, in the very first instance of "subscription" in the abstract: "MOQT allows a producer of media to publish data and have it consumed via subscription by a multiplicity of endpoints." This sentence gives me an impression that it covers all means of consumption of published data, thus it includes fetches. There are many other places like this in the draft. Therefore, clarifications, or maybe a definition of "subscription", should be made/added, such that readers won't have confusions.

afrind:
The terms list uses subscription to define Publisher and End Subscriber but never defines it, and the definition in Publishing and Receiving Tracks is several hundred lines later. Add an entry that points at it, following the pattern used for Group, Subgroup, Object and Track.

Fixes: #1226

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -255,6 +255,11 @@ Track:
 
 : A track is a collection of groups. See ({{model-track}}).
 
+Subscription:
+
+: An ongoing relationship in which a publisher delivers newly published
+  objects from a track to a subscriber. See ({{subscriptions}}).
+
 ## Stream Management Terms
 
 This document uses stream management terms described in {{?RFC9000, Section


draft-ietf-moq-transport.md
```diff
@@ -255,6 +255,11 @@ Track:
 
 : A track is a collection of groups. See ({{model-track}}).
 
+Subscription:
+
+: An ongoing relationship in which a publisher delivers newly published
```
ianswett, line 262: I still find relationship a bit odd, maybe "request"?

```suggestion
: An ongoing request in which a publisher delivers newly published
```

afrind: If we change that, we should change it in Section 3 also? CC: @vasilvv

afrind: Keeping 'ship for now

vasilvv (approved):

PR #1926, opened 2026-09-07
Note that Client and Server are transport roles only

### LINKED ISSUE #1217: Concepts of Client and Server might need clarifications
In 1.2:

> Client:
> The party initiating a Transport Session.
> 
> Server:
> The party accepting an incoming Transport Session.
> 

Given a Publisher and a Subscriber, if the Subscriber initiates the Transport Session, then it is straightforward that the Subscriber is the Client, and the Publisher is the Server. However, if the Publisher initiates the Transport Session (that is not disallowed, right?), then is it true that the Publisher is the Client while the Subscriber is the Server?

Unless it is disallowed for a Publisher to initiate the Transport Session, there are places in the draft that need to clarified to make things clear. For example, in many (or maybe most) places of the draft, it seems that the term Server is used implying that it is the Publisher, and that would be problematic if the Session is initiated by the Publisher.

afrind:
A Publisher or Subscriber can be either party to the Transport Session, but the terminology did not say so, leaving readers to infer that a Server is a Publisher.

Fixes: #1217

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -179,11 +179,13 @@ Application:
 
 Client:
 
-: The party initiating a Transport Session.
+: The party initiating a Transport Session.  A Client can be a Publisher, a
+  Subscriber, or both.
 
 Server:
 
-: The party accepting an incoming Transport Session.
+: The party accepting an incoming Transport Session.  A Server can be a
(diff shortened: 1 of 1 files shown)


ianswett (approved):
I'm not sure we need this, but it doesn't hurt.

PR #1927, opened 2026-09-07
Scope the auth token cache overflow rule to non-SETUP registrations

### LINKED ISSUE #1271: REGISTER requirements are poorly organized.
IN 9.2.2.1:

> If a registration is attempted which would cause this limit to be exceeded, the receiver MUST termiate the Session with a AUTH_TOKEN_CACHE_OVERFLOW error.

IN 9.3.2.5:

> If a server receives an AUTHORIZATION TOKEN parameter in CLIENT_SETUP with Alias Type REGISTER that exceeds its MAX_AUTH_TOKEN_CACHE_SIZE, it MUST NOT fail the session with AUTH_TOKEN_CACHE_OVERFLOW.

I don't think it's good for one section to say "you MUST always do this"  and have a completely different section say that "you MUST NOT in this very specific circumstance."

At the very least, the sentences above should have hyperlinks to each other. Or we could consolidate this.

afrind:
Authorization Token Compression required terminating the session on a registration that exceeds MAX_AUTH_TOKEN_CACHE_SIZE, while the SETUP option requires not doing that and treating the registration as USE_VALUE instead, because SETUP messages can cross in flight. Neither referenced the other. State the general rule's scope and point at the SETUP case.

Fixes: #1271

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -2503,9 +2503,10 @@ By registering a Token, the sender is requiring the receiver to store the Token
 Alias and Token Value until they are deleted, or the Session ends. The receiver
 can protect its resources by sending a Setup Option defining the
 MAX_AUTH_TOKEN_CACHE_SIZE limit (see {{max-auth-token-cache-size}}) it is
-willing to accept. If a registration is attempted which would cause this limit
-to be exceeded, the receiver MUST terminate the Session with a
-`AUTH_TOKEN_CACHE_OVERFLOW` error.
+willing to accept. If a registration outside of SETUP is attempted which would
+cause this limit to be exceeded, the receiver MUST terminate the Session with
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -2503,9 +2503,10 @@ By registering a Token, the sender is requiring the receiver to store the Token
 Alias and Token Value until they are deleted, or the Session ends. The receiver
 can protect its resources by sending a Setup Option defining the
 MAX_AUTH_TOKEN_CACHE_SIZE limit (see {{max-auth-token-cache-size}}) it is
-willing to accept. If a registration is attempted which would cause this limit
-to be exceeded, the receiver MUST terminate the Session with a
-`AUTH_TOKEN_CACHE_OVERFLOW` error.
+willing to accept. If a registration outside of SETUP is attempted which would
```
```suggestion
willing to accept. If a registration outside of SETUP is attempted that would
```

ianswett (approved):

Issue #1854, opened 2026-08-10
to PUBLISH_NAMESPACE or to not PUBLISH_NAMESPACE


kixelated:
I think this is a major interop issue, and it keeps biting me.

There are implicitly three roles in the moq-transport draft:
- publisher
- relay
- subscriber

While not stated, there's an implicit assumption that:
- **publishers** should send `PUBLISH_NAMESPACE`
- **subscribers** should send `SUBSCRIBE_NAMESPACE`
- **relays** should wait for `PUBLISH_NAMESPACE` and `SUBSCRIBE_NAMESPACE`

The problem is that my library supports:
- publishers connecting directly to subscribers (P2P): results in both messages
- relays connecting to other relays (S2S): results in neither message.

**There's no in-transport mechanism to know if the peer will PUBLISH or SUBSCRIBE.** 

My workaround has been to always send both `PUBLISH_NAMESPACE` and `SUBSCRIBE_NAMESPACE`. But that's not ideal,  and then publishers complain that they get `PUBLISH_NAMESPACE` messages for random shit, as they expected the relay to wait for a `SUBSCRIBE_NAMESPACE`.

The other alternative is to have the application indicate if the peer is a relay just to decide whether to PUBLISH or not. I really don't like this approach.

kixelated:
And I'm focusing on `PUBLISH_NAMESPACE` but this should apply to `PUBLISH` as well. A publisher is expected to `PUBLISH` but a relay is expected to wait for a `SUBSCRIBE`.

afrind:
> While not stated, there's an implicit assumption that:
> 
> publishers should send PUBLISH_NAMESPACE
> subscribers should send SUBSCRIBE_NAMESPACE
> relays should wait for PUBLISH_NAMESPACE and SUBSCRIBE_NAMESPACE

I think this is the wrong framing.

Applications know what they want to do and will send the messages they need to send to relays to make things work.  If your application needs to publish - send PUBLISH (not PUBLISH_NAMESPACE).  If your application wants to wait for subscribers to express interest, send PUBLISH_NAMESPACE.

The same is true for subscribers.  If you know what you want, send SUBSCRIBE.  If you need to do some form of track discovery, send SUBSCRIBE_NAMESPACE.

> publishers connecting directly to subscribers (P2P): results in both messages

Seems like this can be worked out by the application (always PUBLISH after connect?).

> relays connecting to other relays (S2S): results in neither message.

This is explicitly out of scope.

[...]

PR #1932, opened 2026-09-07
Remove leftover MUST forward PUBLISH_NAMESPACE

afrind:
Publisher Interactions said:

  Relays MUST forward PUBLISH_NAMESPACE or PUBLISH messages to all
  matching subscribers.

Both halves of that sentence are gone.

The PUBLISH half was a duplicate. The rule it restated is four paragraphs below it, and is more precise: a relay sends PUBLISH to each subscriber that has sent SUBSCRIBE_TRACKS for the Track's namespace or a prefix thereof, except where the relay is holding a downstream SUBSCRIBE awaiting a publisher for that Track. Nothing is lost by deleting the shorter version.

The PUBLISH_NAMESPACE half was a leftover and did not describe real behavior. A relay that receives an authorized PUBLISH_NAMESPACE records the sender as a source for that prefix, and sends it a SUBSCRIBE for each existing subscription the prefix matches. It advertises namespaces downstream with NAMESPACE in response to a matching SUBSCRIBE_NAMESPACE. Say so alongside the rule for handling a received PUBLISH_NAMESPACE, since the draft never stated it anywhere and implementers were forwarding PUBLISH_NAMESPACE for conformance.

Also split the paragraph that covered PUBLISH_NAMESPACE and PUBLISH handling together, so each message has its own.

Refs: #1854

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1997,8 +1997,7 @@ bar).  It will not match a session with namespace=(foobar).
 
 Relays MUST send SUBSCRIBE messages to all matching publishers. This includes
 matching both Established subscriptions on the Full Track Name and Namespace
-Prefix Matching against published Namespaces.  Relays MUST forward
-PUBLISH_NAMESPACE or PUBLISH messages to all matching subscribers.
+Prefix Matching against published Namespaces.
 
 When a Relay needs to make an upstream FETCH request, it determines the
 available publishers using the same matching rules as SUBSCRIBE. When more than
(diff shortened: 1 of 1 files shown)


ianswett (approved):

suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -2024,15 +2023,19 @@ holding a downstream SUBSCRIBE awaiting a publisher for this Track (see
...
-matching subscription.  When it receives an authorized PUBLISH message for a
+When a relay receives an authorized PUBLISH message for a
 Track that has `Established` downstream subscriptions, it MUST respond with
 PUBLISH_OK.  If at least one downstream subscriber for the Track has
 Forward State=1, the Relay MUST change the Forward State to 1 with
 REQUEST_UPDATE.
 
+When a relay receives an authorized PUBLISH_NAMESPACE for a namespace that
+matches one or more existing subscriptions to other upstream sessions, it MUST
+send a SUBSCRIBE to the publisher that sent the PUBLISH_NAMESPACE for each
```
suhasHere, line 2045:
```suggestion
send a SUBSCRIBE to the publishers that sent the PUBLISH_NAMESPACE for each
```

suhasHere (approved):

draft-ietf-moq-transport.md
afrind: I think the singular matches "receives an authorized"

PR #1940, opened 2026-09-10
Call the forwarding state paused

### LINKED ISSUE #1694: bikeshed forward state name

afrind:
FORWARD itself is unchanged, and there's a new Pausing Subscriptions section for the rest of the draft to point at. The wording comes from #1825, which also removes FORWARD in favor of an empty Location Filter; that part is left for #1352.

Fixes: #1694

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1909,7 +1909,7 @@ that specify otherwise.
 ## Paused Subscription Handling
 
 If one or more downstream subscribers to a track are not paused, the relay
-MUST resume the upstream subscription, if paused, in order to receive and
+MUST have an unpaused upstream subscription, in order to receive and
 forward the requested Objects. When all downstream subscribers are paused, the
 relay chooses whether to pause upstream at its discretion, considering the
 following
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -625,13 +625,6 @@ to a SUBSCRIBE. A subscriber MUST send exactly one PUBLISH_OK
 ({{message-request-ok}}) or PUBLISH_ERROR in response to a PUBLISH. The peer
 SHOULD close the session with a protocol error if it receives more than one.
 
-All `Established` subscriptions have a Forward State which is either 0 or 1.
-The publisher does not send Objects if the Forward State is 0, and does send them
-if the Forward State is 1.  The initiator of the subscription sets the initial
-Forward State in either PUBLISH or SUBSCRIBE.  The subscriber can send
-REQUEST_UPDATE to update the Forward State. Control messages, such as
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

draft-ietf-moq-transport.md
```diff
@@ -1005,11 +1011,12 @@ with an open-ended fill range, which the publisher will end at Largest Object.
...
-REQUEST_UPDATE that carries FILL_PARAMETERS while Forward State is 1.
+REQUEST_UPDATE that carries FILL_PARAMETERS while the subscription is not
+paused (see {{pausing-subscriptions}}).
 
-- FILL_PARAMETERS carried while Forward State is 0 opens no fill fetch stream.
-  Transitioning to Forward State 1 without re-sending FILL_PARAMETERS does not
-  open one either.
+- FILL_PARAMETERS carried while the subscription is paused opens no fill fetch
+  stream.  Resuming the subscription without re-sending FILL_PARAMETERS does
+  not open one either.
```
ianswett, line 1019: This isn't a change, but it's a bit odd.

draft-ietf-moq-transport.md
```diff
@@ -1899,20 +1906,22 @@ A cache MUST store all fields of an Object defined in {{object-header}},
...
-## Forward Handling
+## Paused Subscription Handling
 
-If one or more downstream subscribers to a track have Forward=1, the relay
-MUST set Forward=1 upstream in order to receive and forward the requested
-Objects. When no downstream subscriber has Forward=1, the relay chooses the
-upstream Forward value at its discretion, considering the following
+If one or more downstream subscribers to a track are not paused, the relay
+MUST resume the upstream subscription, if paused, in order to receive and
+forward the requested Objects. When all downstream subscribers are paused, the
```
ianswett, line 1913: This reads a bit more awkwardly than before, maybe this?

```suggestion
If one or more downstream subscribers to a track are not paused, the relay
MUST have an unpaused upstream subscription, in order to receive and
forward the requested Objects. When all downstream subscribers are paused, the
```

ianswett (approved):

draft-ietf-moq-transport.md
afrind: There's a reason -- fill-fetch is for _joining_.  If the live half is paused, your fill-fetch is nonsensical?

The converse is if you are using fill fetch to do replay behind live -- you may be paused but want to spawn fills.  Probably deserves it's own issue.

PR #1946, opened 2026-09-14
Restructure Publisher and Namespace Discovery, Give Subscribe Tracks it's own section

afrind:
- Namespace Prefix Matching gets its own section under Track.
- SUBSCRIBE_TRACKS moves to its own section under Publishing and Receiving
  Tracks — it creates subscriptions rather than discovering namespaces.
- Behavior is hoisted out of the five namespace message sections, which now
  carry only a diagram and fields.
- Publishing now precedes Subscribing, and section 4 is tightened: response
  handling was stated three times, authorization in three places.
- A ladder diagram shows the publisher → relay → subscriber path.
- Editorial cleanup of some prose
- 
sentence-diff (from `draft-21-retro`) against  main: 

2343 of 2379 sentences unchanged.

 * modified 21
 * added 22
 * removed 23

### DIFF
--- a/draft-ietf-moq-transport.md
+++ b/draft-ietf-moq-transport.md
@@ -1234,8 +1234,42 @@ The receiver of a REQUEST_OK or REQUEST_ERROR ought to
 forward the result to the application, so the application can decide which other
 publishers to contact, if any.
 
+The publisher will respond with SUBSCRIBE_NAMESPACE_OK or
+SUBSCRIBE_NAMESPACE_ERROR on the response half of the stream. If the subscriber
+receives any message other than a SUBSCRIBE_NAMESPACE_OK or a
+SUBSCRIBE_NAMESPACE_ERROR as the first message on the response half of the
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-transport.md
```diff
@@ -1203,52 +1303,60 @@ to advertise namespaces within the prefix being discovered. Such a
...
-especially namespace subscriptions,
-to conserve and protect their resources from excessive load.  They MAY
-also impose limits on the number of publishers in a namespace, by rejecting
-or closing namespace subscriptions with the error NAMESPACE_TOO_LARGE, or
-CONFLICTING_FILTERS if too many disjoint filters are requested on downstream
-subscriptions across a large number of subscribers, or PREFIX_OVERLAP if different
-subscribers force an aggregated upstream subscription to overlap.
+~~~
+ Publisher                    Relay                   Subscriber
+     |                          |                          |
```
afrind: Add a PUB_NS before SUB_NS too

